PR #1135, opened 2020-10-23
enhance the skipconnection layer

wyli:
### Description
add/mul modes for the skip connection layer

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/networks/layers/simplelayers.py
+++ b/monai/networks/layers/simplelayers.py
@@ -27,16 +27,38 @@ __all__ = ["SkipConnection", "Flatten", "GaussianFilter", "LLTM", "Reshape", "se
 
 class SkipConnection(nn.Module):
     """
-    Concats the forward pass input with the result from the given submodule.
+    Combine the forward pass input with the result from the given submodule::
+
+        --+--submodule--o--
+          |_____________|
+
+    The available modes are ``"cat"``, ``"add"``, ``"mul"``.
     """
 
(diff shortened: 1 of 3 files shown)


monai/networks/layers/simplelayers.py
```diff
@@ -27,16 +27,38 @@
...
+            submodule: the module defines the trainable branch.
+            dim: the dimension over which the tensors are concatenated.
+                Used when mode is ``"cat"``.
+            mode: ``"cat"``, ``"add"``, ``"mul"``. defaults to ``"cat"``.
+        """
         super().__init__()
         self.submodule = submodule
-        self.cat_dim = cat_dim
+        self.dim = dim
+        self.mode = mode
```
ericspod, line 50: I think the mode strings should be put into constants in `monai/utils/enums.py`. There can then be a proper check in the constructor that a correct mode was selected, raising ValueError if not. The raise in the `forward` method can then be removed.

ericspod (changes requested):

monai/networks/layers/simplelayers.py
wyli: it seems the enum object doesn't work well with torchscript, @Nic-Ma any idea?

ericspod: It should be working, there's references to it in Pytorch commits, but I'm having problems getting it to work as well.

ericspod (approved):

PR #1136, opened 2020-10-23
Rewrite confusion matrix and modify meandice

### LINKED ISSUE #1129: Some metrics don't handle batch and channel dims
**Describe the bug**
Recently, we added several new medical metrics and I found some segmentation metrics don't handle `batch` and `channel` dims which are the aligned data shape in MONAI.
Like `MeanDice` metric, I think we should compute segmentation metrics on [B, C, H, W, <D>] data and return `M x N` results, where M corresponding to batch dim and N corresponding to channel dim. Then compute the average value after 1 epoch.
And it should handle both non-one-hot and one-hot labels.

Please feel free to correct me if I misunderstand anything.

yiheng-wang-nv:

Partially fixes #1129 .

### Description
Rewrite confusion matrix and mean dice, to unify the formats and compute logic.

### Status
**Ready**

### Types of changes


yiheng-wang-nv:
Hi @Nic-Ma , my changes have some conflicts with your last merged commit (https://github.com/Project-MONAI/MONAI/commit/3adee016f6ce522176f7bbaeef11231959677e89), could you please help to see how can I solve these conflicts? Thanks

monai/metrics/seg_metric_utils.py
```diff
@@ -0,0 +1,218 @@
...
+
+import warnings
+from typing import Callable, Optional, Sequence, Union
+
+import torch
+
+from monai.networks import one_hot
+from monai.utils import MetricReduction
+
+
```
Nic-Ma, line 20: I think these `do_onehot`, `do_actication`, and `do_binarization` APIs are duplicated with `Activations` and `AsDiscrete` post transforms.
@wyli , I think maybe we can drop these logic in all the metrics directly to make them more clear? If user is writing PyTorch program with our metrics, they can call the post transforms in validation loop before calling metrics, if writing program based on our workflows, it already contains the `post transforms`.

Thanks.

monai/metrics/seg_metric_utils.py
```diff
@@ -0,0 +1,220 @@
...
+    if not include_background:
+        y = y[:, 1:] if y.shape[1] > 1 else y
+        y_pred = y_pred[:, 1:] if y_pred.shape[1] > 1 else y_pred
+    assert y.shape == y_pred.shape, "data shapes of y_pred and y do not match."
+    y = y.float()
+    y_pred = y_pred.float()
+    return y_pred, y
+
+
+def do_metric_reduction(
```
Nic-Ma, line 163: Could you please change this to `do_reduction()` and move to `utils.py`?
Thanks.

monai/metrics/seg_metric_utils.py
```diff
@@ -0,0 +1,220 @@
...
+    elif bin_mode == "mutually_exclusive":
+        assert input_ndim > 1, "mutually_exclusive is used for multi-class tasks."
+        n_classes = input_data.shape[1]
+        assert n_classes > 1, "mutually_exclusive is used for multi-class tasks."
+        input_data = torch.argmax(input_data, dim=1, keepdim=True)
+        input_data = one_hot(input_data, num_classes=n_classes)
+    return input_data
+
+
+def preprocess_input(
```
Nic-Ma, line 104: I think we can remove all the pre-processing logic for metrics, and ask users to use `post_transforms` before metrics computation. Otherwise, they have duplicated parts and confusing.
@wyli @ericspod, what do you think? If you agree, we can also update examples and tutorials to use `post_transforms`.  Actually, if we changed to use `post_transforms` in examples, the examples are still compatible with MONAI v0.3.0.

Thanks.

monai/metrics/seg_metric_utils.py
```diff
@@ -0,0 +1,220 @@
...
+    # some elements might be Nan (if ground truth y was missing (zeros))
+    # we need to account for it
+    nans = torch.isnan(f)
+    not_nans = (~nans).float()
+    f[nans] = 0
+
+    t_zero = torch.zeros(1, device=f.device, dtype=torch.float)
+    reduction = MetricReduction(reduction)
+
+    if reduction == MetricReduction.MEAN:
```
Nic-Ma, line 190: Could you please help also test the reduction ops for the classification task?
Thanks.

Nic-Ma (commented):
Thanks for your quick update.
Put some minor comments inline.

Thanks.

monai/metrics/seg_metric_utils.py
ericspod: Yes, making things more modular makes sense, putting general purpose functionality into post transforms and expecting users to apply them as appropriate is the way to go with this.

monai/metrics/seg_metric_utils.py
```diff
@@ -0,0 +1,220 @@
...
+            Defaults to True.
+
+    Raises:
+        AssertionError: When ``y_pred`` only has one dimebsion.
+        AssertionError: When ``y`` and ``y_pred`` have different shapes before output.
+
+    """
+
+    # convert y into one-hot version if needed
+    assert y_pred.ndimension() > 1, "y_pred should at least has two dimensions batch and n_classes."
```
ericspod, line 143: In a number of places asserts are being used to check inputs. I recommend that input validation be done with conditional check followed by a raising of a `ValueError`. This is inline with standard practice and is what users expect in terms of the type of exception they would get, plus there's the issue that assertions can be turned off at runtime and input validation checks shouldn't ever be disabled.

monai/metrics/seg_metric_utils.py
```diff
@@ -0,0 +1,220 @@
...
+    if activation is not None:
+        y_pred = do_activation(y_pred, activation=activation)
+    # binarization for y_pred if needed
+    if bin_mode is not None:
+        y_pred = do_binarization(y_pred, bin_mode=bin_mode, bin_threshold=bin_threshold)
+    # remove background if needed
+    if not include_background:
+        y = y[:, 1:] if y.shape[1] > 1 else y
+        y_pred = y_pred[:, 1:] if y_pred.shape[1] > 1 else y_pred
+    assert y.shape == y_pred.shape, "data shapes of y_pred and y do not match."
```
ericspod, line 157: This assert makes more sense, being used to check the correctness of an internal condition of your algorithm.

monai/metrics/seg_metric_utils.py
yiheng-wang-nv: I see, I just removed the preprocess part and update the code. My next step is to add classification support. I will switch to a formal pr after finishing all places.

Nic-Ma:
Hi @ericspod and @wyli ,

Could you please help review this PR again?
Once you guys approved, Yiheng will create another PR to update our examples and tutorials to use `post_transform`.
And merge them together.

Thanks.

monai/metrics/meandice.py
```diff
@@ -9,231 +9,115 @@
...
             Define the mode to reduce computation result of 1 batch data. Defaults to ``"mean"``.
 
-    Raises:
-        ValueError: When ``sigmoid=True`` and ``other_act is not None``. Incompatible values.
-
     """
 
     def __init__(
         self,
         include_background: bool = True,
```
wyli, line 41: How about we have an option `ignore_nans: bool = True`: when it's True, the `__call__` will return the metric (with `nans` replaced by 0) and the result of `non_nans`; when it's False, the `__call__` will directly return the metric without any `nan` replacement.

this is to avoid `self.not_nans = not_nans`, this assignment won't work in a multi-process context

monai/metrics/confusion_matrix.py
```diff
@@ -9,99 +9,307 @@
...
+            y: ground truth to compute the metric. It must be one-hot format and first dim is batch.
+                The values should be binarized.
+        Raises:
+            ValueError: when `y_pred` is not a binarized tensor.
+            ValueError: when `y` is not a binarized tensor.
+            ValueError: when `y_pred` has less than two dimensions.
+        """
+        # check binarized input
+        if not torch.all(y_pred.byte() == y_pred):
+            raise ValueError("y_pred should be a binarized tensor.")
```
wyli, line 78: this could be a warning instead of error, don't think it's a fatal issue here..

wyli (commented):
please see some comments inline

wyli (approved):

PR #1138, opened 2020-10-25
Threadbuffer

ericspod:
### Description
Adds `ThreadBuffer` for generating batches in a separate thread from the main thread. This allows a batch to be generated while the GPU is busy with training, allowing some overlap of execution which should speed up the training process. A example notebook to come in the tutorials repo.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @ericspod ,

Interesting idea, may I know what's the difference with PyTorch DataLoader data parallel?
https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py#L788
I think `DataLoader` already supports `prefetch` in parallel?

Thanks.

ericspod:
The `DataLoader` class itself does prefetch but in a separate process and doesn't allow control of the buffer size. The process based way this works is potentially a problem in Windows as well. This class is generic, allows control of buffer size which may be useful if generating data from remote sources or something else with irregular throughput, is simpler and thread-based, and a generic iterator so could be used to move other iterations into separate threads as well, not just in the context of data loaders.

wyli (approved):
would be nice to support buffer shuffling somehow?

PR #1021, opened 2020-09-11
789 cuda resampling part2

wyli:
related to #789 

### Description
this is an experimental feature, mainly to provide basic code structure and build/test CI environments for the resampling model

- adds cuda/cpp impl. of higher-order interpolation layer (in both 'pull' and 'push' directions -- differentiable)
- by default no cpp/cuda extension installations (unless user enables the flag `BUILD_MONAI=1 python setup.py develop`)
- pypi install won't have the precompiled with this PR
- CI is enabled to test building with a few versions of pytorch+cuda/cpu+linux/macos

the resampling code is adapted from https://github.com/balbasty/nitorch
cc @balbasty @brudfors

### Status
**ready**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [x] New tests added to cover the changes.
- [x] Integration tests passed locally by running `./runtests.sh --codeformat --coverage`.
- [x] In-line docstrings updated.

PR #1140, opened 2020-10-26
1021 Fixes to CPP/CUDA resampling

brudfors:
Fixes #1021.

### Description
* Out-of-bounds checks not working when extrapolate = False.
* Backward pass not returning correct output.
* CPU implementation should use int64 offsets and the CUDA implementation int32.

### Status
**Work in progress**

### Types of changes

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -7,12 +7,12 @@ on:
 jobs:
   cron-gpu:
     container:
-      image: nvcr.io/nvidia/pytorch:20.03-py3
+      image: nvcr.io/nvidia/pytorch:20.03-py3  # CUDA 10.2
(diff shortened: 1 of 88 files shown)


brudfors:
I think that I might not have been on the most up-do-date version of master when I made this pull request, so I think it is 'polluted' a bit. How can I fix this? Thank you

wyli:
> I think that I might not have been on the most up-do-date version of master when I made this pull request, so I think it is 'polluted' a bit. How can I fix this? Thank you

Hi @brudfors, it seems only the changes in `test_ahnet.py` are unwanted. please run ``git checkout origin/master tests/test_ahnet.py`` (assuming `origin/master` is up-to-date), and then commit the relevant files to this PR...

PR #1141, opened 2020-10-26
1021 cuda resampling 1

brudfors:
Fixes #1021 .

### Description
* Out-of-bounds checks not working when extrapolate = False.
* Backward pass not returning correct output.
* CPU implementation should use int64 offsets and the CUDA implementation int32.

### Status
**Work in progress**

### Types of changes


brudfors:
Hi @wyli 

I closed and reopened this pull request as it seemed as if the commit history was a bit messed up too. I think it should all be OK now. Sorry about that.

wyli (approved):
thanks @brudfors!

PR #1147, opened 2020-10-27
Support multi-channel images in ITKReader

### LINKED ISSUE #1007: LoadImage Fails to Load Certain JPEGs
**Describe the bug**
LoadImage fails to correctly load certain JPEGs. The produced array contains streaked information. Test image included:

bad_image
.

**To Reproduce**

```py
im=LoadImage(image_only=True)("bad_image.jpg")
plt.imshow(im[0],cmap="gray")
```

This produces

result

**Expected behavior**
Image should look like the attached image but comes out corrupted when viewed with matplotlib, using `plt.imread` does work.

**Environment (please complete the following information):**
 - Ubuntu
 - Python version 3.7
 - MONAI version 62903b6924e572edfb0d44ce333dd0251e659bfe
 - CUDA/cuDNN version 11, 7.5
 - GPU Titan Xp

thewtex:
The memory layout of itk.Image, like most image processing and
visualization libraries in Python has all pixel's channels
adjacent in memory, i.e. R1G1B1R2G2B2R3G3B3. This layout follows the
memory access patterns important for performance in these use cases.
For deep learning libraries like PyTorch/MONAI, we need/want
channels to be contiguous, i.e. R1R2R3G1G2G3B1B2B3. This layout
follows the memory access patterns import for performance for
many-layered convolutions common in deep learning. We need to change
the NumPy dimensions to result in the desired contiguous channel
layout for PyTorch.

More examples: https://nbviewer.jupyter.org/gist/thewtex/2e9aa95f7e65170082a606f26b49b482

Fixes #1007 .

### Description
Support multi-channel images in ITKReader

### Status
**Ready**

### Types of changes


thewtex:
@aylward @Nic-Ma @hjmjohnson please take a look

monai/data/image_reader.py
```diff
@@ -235,11 +236,28 @@ def _get_array_data(self, img) -> np.ndarray:
...
         Args:
             img: a ITK image object loaded from a image file.
 
         """
-        return itk.array_view_from_image(img, keep_axes=False)
+        channels = img.GetNumberOfComponentsPerPixel()
+        if channels == 1:
+            return itk.array_view_from_image(img, keep_axes=False)
+        else:
+            # The memory layout of itk.Image has all pixel's channels adjacent
```
Nic-Ma, line 251: Hi @ericspod and @thewtex ,

I am not very sure about the dim order in medical images, I think we discussed this topic before for ITK and Nibabel things. Could you please help share some comments here?

Thanks.

hjmjohnson: There is not a consistent dim order for all medical images.

ericspod: That's true, it's not consistent. With Nifti files it's almost always been in my experience WHDT or HWDT, but nothing enforces this.

Nic-Ma: Thanks @hjmjohnson and @ericspod for your comments, so what do you think about this PR?

monai/data/image_reader.py
```diff
@@ -235,11 +236,28 @@ def _get_array_data(self, img) -> np.ndarray:
...
+            # in memory, i.e. R1G1B1R2G2B2R3G3B3. For PyTorch/MONAI, we need
+            # channels to be contiguous, i.e. R1R2R3G1G2G3B1B2B3.
+            arr = itk.array_view_from_image(img, keep_axes=False)
+            dest = list(range(img.ndim))
+            source = dest.copy()
+            end = source.pop()
+            source.insert(0, end)
+            arr_contiguous_channels = np.moveaxis(arr, source, dest)
```
ericspod, line 259: Wouldn't `np.moveaxis(arr,-1,0)` have the same effect?

monai/data/image_reader.py
ericspod: I think this is correct for colour images like the test demonstrates, the issue of ordering of spatial and temporal dimensions might have to be up to the user.

monai/data/image_reader.py
thewtex: I tried that at first, but it does not work as desired.

ericspod: My test is this, which yields something of the right shape but perhaps the array in memory isn't what we want:

```py
x=np.random.rand(10, 20, 3)
dest = list(range(x.ndim))  # [10, 20, 3]
source = dest.copy()
end = source.pop()
source.insert(0, end)
print(x.shape, np.moveaxis(x, source, dest).shape, np.moveaxis(x, -1, 0).shape)
>>> ((10, 20, 3), (3, 10, 20), (3, 10, 20))
```

Another way to shorten the code is construct `source` as `source = np.roll(dest, 1)`.

ericspod (approved):
I think this is a good change for the type of images ITK would load which are know to be multichannel, that is 2D photographic images like jpg or png. For other formats with a channel component the user would have to determine which dimension is which as discussed.

PR #1153, opened 2020-10-28
807 update dice and the relevant loss

### LINKED ISSUE #807: Training instability with Dice Loss/Tversky Loss
I am training a 2D UNet to segment fetal MR images using MONAI and I have been observing some instability in the training when using MONAI Dice loss formulation. After some iteration, the loss jumps up and the network stops learning, as the gradients drop to zero. Here is an example (orange is loss on training set computed over 2D slices, blue is loss on validation computed over 3D volume).
image

After investigating several aspects (using the same deterministic seed), I've narrowed down the issue to the presence of the ```smooth``` term in both the numerator and denominator of the Dice Loss:
```f = 1.0 - (2.0 * intersection + smooth) / (denominator + smooth)```

When using the formulation:
```f = 1.0 - (2.0 * intersection) / (denominator + smooth)```
without the smooth term in the numerator, the training was stable and no longer showed unexpected behaviour:
image
[Note: this experiment was trained for much longer to make sure the jump would not appear later in the training]

The same pattern was observed also for the Tversky Loss, so it could be worth investigating the stability of the losses to identify the best default option.

**Software version**
MONAI version: 0.1.0+84.ga683c4e.dirty
Python version: 3.7.4 (default, Jul  9 2019, 03:52:42)  [GCC 5.4.0 20160609]
Numpy version: 1.18.2
Pytorch version: 1.4.0
Ignite version: 0.3.0

**Training information**
Using MONAI PersistentCache 
2D UNet (as default in MONAI)
Adam optimiser, LR = 1e-3, no LR decay
Batch size: 10

**Other tests**
The following aspects were investigated but did **not** solve the instability issue:
- Gradient clipping
- Different optimisers (SGD, SGD + Momentum)
- Transforming the binary segmentations to a two-channel approach ([background segmentation, foreground segmentation])
- Choosing ```smooth = 1.0``` as default here (https://github.com/MIC-DKFZ/nnUNet/blob/master/nnunet/training/loss_functions/dice_loss.py). However, this made the behaviour even more severe and the jump would happen sooner in the training.

The following losses were also investigated
- Binary Cross Entropy --> stable
- Dice Loss + Binary Cross Entropy --> unstable
- Dice Loss (no smooth at numerator) + Binary Cross Entropy --> stable
- Tversky Loss --> Unstable
- Tversky Loss (no smooth at numerator) --> stable

wyli:
closes #807 

### Description
- adds `smooth_nr`/`smooth_dr` for the numerators and denominators
- adds a flag for the 'batchwise' reduction

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/losses/dice.py
+++ b/monai/losses/dice.py
@@ -26,12 +26,12 @@ class DiceLoss(_Loss):
     Compute average Dice loss between two tensors. It can support both multi-classes and multi-labels tasks.
     Input logits `input` (BNHW[D] where N is number of classes) is compared with ground truth `target` (BNHW[D]).
     Axis N of `input` is expected to have logit predictions for each class rather than being image channels,
-    while the same axis of `target` can be 1 or N (one-hot format). The `smooth` parameter is a value added to the
-    intersection and union components of the inter-over-union calculation to smooth results and prevent divide by 0,
-    this value should be small. The `include_background` class attribute can be set to False for an instance of
-    DiceLoss to exclude the first category (channel index 0) which is by convention assumed to be background.
-    If the non-background segmentations are small compared to the total image size they can get overwhelmed by
-    the signal from the background so excluding it in such cases helps convergence.
+    while the same axis of `target` can be 1 or N (one-hot format). The `smooth_nr` and `smooth_dr` parameters are
+    values added to the intersection and union components of the inter-over-union calculation to smooth results
+    respectively, these values should be small. The `include_background` class attribute can be set to False for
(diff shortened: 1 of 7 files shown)


ericspod:
Looks good to me, @martaranzini ?

martaranzini:
Looks all good to me as well! Perhaps it could be worth adding a test case where the two smooth terms are different, say smooth_nr=0 and smooth_dr=1e-5?

ericspod (approved):

PR #1162, opened 2020-10-30
1149 Fix CI test issues for PyTorch 1.7

### LINKED ISSUE #1149: compatibility with pytorch 1.7
pytorch 1.7 is available, there're some compatibility issues such as https://github.com/Project-MONAI/MONAI/runs/1315919388#step:8:15470

Nic-Ma:
Fixes #1149 .

### Description
This PR updated the CI tests and configs for PyTorch 1.7.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @wyli ,

The root cause of integration test errors is that PyTorch 1.7 updated the `shuffle` logic in `DataLoader` by this PR:
https://github.com/pytorch/pytorch/pull/41682
So my fix is to set `shuffle=False` during training.
BTW, you have 2 commits without signoff, could you please add them?

Thanks.

.github/workflows/pythonapp.yml
```diff
@@ -73,13 +74,13 @@ jobs:
     - if: runner.os == 'windows'
       name: Install torch cpu from pytorch.org (Windows only)
       run: |
-        python -m pip install torch==1.6.0 torchvision==0.7.0 -f https://download.pytorch.org/whl/torch_stable.html
+        python -m pip install torch==1.7.0 torchvision==0.8.0 -f https://download.pytorch.org/whl/torch_stable.html
```
vfdev-5, line 77:
```suggestion
        python -m pip install torch==1.7.0 torchvision==0.8.1 -f https://download.pytorch.org/whl/torch_stable.html
```

.github/workflows/pythonapp.yml
```diff
@@ -73,13 +74,13 @@ jobs:
...
         # min. requirements for windows instances
         python -c "f=open('requirements-dev.txt', 'r'); txt=f.readlines(); f.close(); print(txt); f=open('requirements-dev.txt', 'w'); f.writelines(txt[1:11]); f.close()"
     - name: Install the dependencies
       run: |
-        python -m pip install torch==1.6.0
-        python -m pip install torchvision==0.7.0
+        python -m pip install torch==1.7.0
+        python -m pip install torchvision==0.8.0
```
vfdev-5, line 83:
```suggestion
        python -m pip install torchvision==0.8.1
```

.github/workflows/pythonapp.yml
```diff
@@ -147,27 +148,28 @@ jobs:
...
-            base: "nvcr.io/nvidia/cuda:10.2-devel-ubuntu18.04"
           - environment: PT16+CUDA102
             pytorch: "torch==1.6.0 torchvision==0.7.0"
             base: "nvcr.io/nvidia/cuda:10.2-devel-ubuntu18.04"
           - environment: PT16+CUDA110
             # we explicitly set pytorch to -h to avoid pip install error
             pytorch: "-h"
             base: "nvcr.io/nvidia/pytorch:20.07-py3"
+          - environment: PT17+CUDA102
+            pytorch: "torch==1.7.0 torchvision==0.8.0"
```
vfdev-5, line 167:
```suggestion
            pytorch: "torch==1.7.0 torchvision==0.8.1"
```

.github/workflows/setupapp.yml
```diff
@@ -33,7 +34,7 @@ jobs:
         which python
         python -m pip install --upgrade pip wheel
         python -m pip uninstall -y torch torchvision
-        python -m pip install torch==1.6.0 torchvision==0.7.0
+        python -m pip install torch==1.7.0 torchvision==0.8.0
```
vfdev-5, line 37:
```suggestion
        python -m pip install torch==1.7.0 torchvision==0.8.1
```

.github/workflows/setupapp.yml
```diff
@@ -77,7 +78,7 @@ jobs:
     - name: Install the dependencies
       run: |
         python -m pip install --upgrade pip wheel
-        python -m pip install torch==1.6.0 torchvision==0.7.0
+        python -m pip install torch==1.7.0 torchvision==0.8.0
```
vfdev-5, line 81:
```suggestion
        python -m pip install torch==1.7.0 torchvision==0.8.1
```

vfdev-5 (commented):
Nit: a small comment to update torchvision's version to 0.8.1 as a minor hotfix for 0.8.0.

Nic-Ma:
Add @IsaacYangSLA also to help debug the `shuffle` change in PyTorch 1.7.0 and how to keep `shuffle=True` work for both torch 1.5 and 1.7 in the integration tests.

Thanks.

Nic-Ma:
Hi @ericspod and @wyli ,

I did a quick test to add below lines before DataLoader:
```py
torch.manual_seed(0)
if torch==1.6:
    dataloader = DataLoader(..., shuffle=True)
if torch==1.7
    generator = torch.Generator()
    generator.manual_seed(0)
    dataloader = DataLoader(..., shuffle=True, generator=generator)
```
The shuffle results are still different.
Now I am not sure whether it's necessary to make the integration test do the same shuffle for torch 1.5 and 1.7..
Or maybe we just use different expected results for the tests?
What do you guys think?

Thanks.

Nic-Ma:
Hi @ericspod and @wyli ,

Could you please help share any comments here? Can we only run integration tests for the latest PyTorch?
This may be the easiest way to go and maintain... We have struggled several times in integration tests with new PyTorch release.

Thanks.

wyli:
I think we could maintain different expected results for now. if we couldn't always verify the integration results, then we couldn't claim that the particular version of pytorch is supported.

in the current PR, integration losses are
```
[1.8735464978369938, 1.7890069894729905, 1.7343427884350917, 1.6714839001370083]
```
while previously we have
```
[0.776872731128316, 0.16163671757005582, 0.07500977408449361, 0.04593902905797882]
```
the gap looks significant, any idea?

ericspod:
Is there some difference in the actual code of `DataLoader` between 1.6 and 1.7? We should see if the generated numbers it uses are actually different somehow. 

For the integration losses, the order of batches being generated would impact the results that much if the training didn't go for very long. We don't want to do long training tests so that's fine, I guess if we have no reason to suspect the test is wrong we can have different results for now. 

Perhaps we should have integration tests that don't use `DataLoader` but some simpler way of iterating over batches, or just create one batch and train with it repeatedly. This would be fine since we don't really care what the result is, just that it's consistent.

Nic-Ma:
> I think we could maintain different expected results for now. if we couldn't always verify the integration results, then we couldn't claim that the particular version of pytorch is supported.
> 
> in the current PR, integration losses are
> 
> ```
> [1.8735464978369938, 1.7890069894729905, 1.7343427884350917, 1.6714839001370083]
> ```
> 
> while previously we have
> 
> ```
> [0.776872731128316, 0.16163671757005582, 0.07500977408449361, 0.04593902905797882]
> ```
> 
> the gap looks significant, any idea?

Hi @wyli ,

I think this difference is because now I disabled the shuffle in Dataloader of training as the shuffle logic changed in PyTorch 1.7 we discussed.

Thanks.

Nic-Ma:
Hi @ericspod ,

In the PyTorch source code for `RandomSampler`, it changed the random logic to use a generator:
https://github.com/pytorch/pytorch/blob/v1.7.0/torch/utils/data/sampler.py#L125
I didn't jump into the `torch.randperm()` to debug what the previous version does.

Thanks.

wyli:
> BTW, you have 2 commits without signoff, could you please add them?


Hi @Nic-Ma I signed off by rebasing. thanks!

wyli:
from the meeting minute: we want to support 1.6.1, 1.7.0, nv torch 20.08, nv torch latest, with different expected integration results if needed.

Nic-Ma:
Hi @ericspod and @wyli ,

I updated the PR to use different expected_values for PyTorch 1.6 and 1.7 according to our discussion.
Once all integration tests passed, I will remove the test branch from workflows.

Thanks.

1
```diff
@@ -0,0 +1,43 @@
+Error: Incorrect commandline provided, invalid key: 2.
```
wyli, line 1: please remove this file

wyli (commented):
I think these hard-coded integration files will get complicated and unreadable quickly.
we also discussed a set of answers based on the platform:

create a new `testing_data/integration_answers.py` with, for example:
```py
{"answer1": # for torch1.5, DGX v100
  {"integration_classification_2d": {"losses": 000, "best_metric": 000}, 
   "test_integration_segmentation_3d": {"losses": 000, "best_metric": 000},
}
{"answer2": # for torch1.7, DGX A100
  {"integration_classification_2d": {"losses": 001, "best_metric": 001}, 
   "test_integration_segmentation_3d": {"losses": 001, "best_metric": 001},
}
...
```
then for all the testing files, import all the answers and 
if the testing results match any of the dictionaries we pass the test. what do you think?

Nic-Ma:
Hi @wyli ,

I agree with the idea to add a separate file for the answers, but I am not sure whether it's a good idea to pass if matched anyone of the answers.
Anyway, let's do it in this way first, if have more cases later, we can enrich it again.

Thanks.

wyli:
> Hi @wyli ,
> 
> I agree with the idea to add a separate file for the answers, but I am not sure whether it's a good idea to pass if matched anyone of the answers.
> Anyway, let's do it in this way first, if have more cases later, we can enrich it again.
> 
> Thanks.

setting the tolerance to 1e-2 was already a workaround. now this PR has another workaround on top of the workaround. not sure whether we are doing things properly here.

1
Nic-Ma: removed in the latest commit.
Thanks.

Issue #1172, opened 2020-11-02
TestLLTM failure


rijobro:
**Describe the bug**
`python -m unittest -k TestLLTM` fails on my machine. From https://github.com/Project-MONAI/MONAI/pull/786, I thought it might be fixed with `python setup.py develop`, but error remains:

```
root@rb-monai-0:~/MONAI# python -m unittest -k TestLLTM
EE
======================================================================
ERROR: test_value_0 (tests.test_lltm.TestLLTM)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/conda/lib/python3.8/site-packages/parameterized/parameterized.py", line 530, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/root/MONAI/tests/test_lltm.py", line 33, in test_value
    new_h, new_c = LLTM(**input_param)(x, (h, c))
  File "/opt/conda/lib/python3.8/site-packages/torch/nn/modules/module.py", line 722, in _call_impl
    result = self.forward(*input, **kwargs)
  File "/root/MONAI/monai/networks/layers/simplelayers.py", line 230, in forward
    return LLTMFunction.apply(input, self.weights, self.bias, *state)
  File "/root/MONAI/monai/networks/layers/simplelayers.py", line 186, in forward
    outputs = _C.lltm_forward(input, weights, bias, old_h, old_cell)
  File "/root/MONAI/monai/utils/module.py", line 211, in __getattr__
    raise self._exception
  File "/root/MONAI/monai/utils/module.py", line 171, in optional_import
    pkg = __import__(module)  # top level module
monai.utils.module.OptionalImportError: import monai._C (No module named 'monai._C').

For details about installing the optional dependencies, please visit:
    https://docs.monai.io/en/latest/installation.html#installing-the-recommended-dependencies

======================================================================
ERROR: test_value_cuda_0 (tests.test_lltm.TestLLTM)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/conda/lib/python3.8/site-packages/parameterized/parameterized.py", line 530, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "/root/MONAI/tests/test_lltm.py", line 47, in test_value_cuda
    new_h, new_c = lltm(x, (h, c))
  File "/opt/conda/lib/python3.8/site-packages/torch/nn/modules/module.py", line 722, in _call_impl
    result = self.forward(*input, **kwargs)
[...]

PR #1173, opened 2020-11-02
skip lltm test if code not compiled

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/1172.

### Description
LLTM test shouldn't be run if code hasn't been compiled

### Status
**Ready**

### Types of changes

### DIFF
--- a/tests/test_lltm.py
+++ b/tests/test_lltm.py
@@ -15,44 +15,43 @@ import torch
 from parameterized import parameterized
 
 from monai.networks.layers import LLTM
-from monai.utils import optional_import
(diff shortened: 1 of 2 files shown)


wyli:
perhaps we could have a `skip_if_no_compile` decorator, like https://github.com/Project-MONAI/MONAI/blob/d0b667786e7718c194d6433bd8bf3a550f2ac963/tests/utils.py#L29-L32

rijobro:
Good point, updated.

wyli (approved):

Issue #816, opened 2020-07-24
support of stratified/repeated k-fold splits for cross validation in apps.dataset


wyli:
**Is your feature request related to a problem? Please describe.**
Currently the app level datasets support randomised training/validation splits,
this could be extended to have k-fold splits so that we could implement [nested/non-nested cross-validation](https://scikit-learn.org/stable/auto_examples/model_selection/plot_nested_cross_validation_iris.html#:~:text=Nested%20cross%2Dvalidation%20(CV),its%20(hyper)parameter%20search.&text=To%20avoid%20this%20problem%2C%20nested,%2Fvalidation%2Ftest%20set%20splits)

**Describe the solution you'd like**
integration of optional cross-validation scheme in apps.dataset

**Relevant context**
[scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.StratifiedKFold.html#sklearn.model_selection.StratifiedKFold)
[monai's current simple splits](https://github.com/Project-MONAI/MONAI/blob/ca614e72039d49bfdbcd16bc0bab66017214e9e2/monai/apps/datasets.py#L63)

thanks for the feedback @holgerroth

wyli:
see also https://github.com/Project-MONAI/MONAI/issues/857

PR #1177, opened 2020-11-02
1171 Update CVDecathlonDataset

### LINKED ISSUE #1171: update CVDecathlonDataset to use partition_dataset
As we already added `parittion_dataset` API, need to update the `CVDecathlonDataset` to use it.

Nic-Ma:
Fixes #1171 .

### Description
This PR updated the `CVDecathlonDataset` to leverage the newly added `partition_dataset` API.

### Status
**Ready**

### Types of changes


ericspod (approved):

wyli (changes requested):
we should have a more generic wrapper, which initializes the actual dataset lazily, something like (this should be in the `monai.data` module):
```py
class CrossValidation:
    def __init__(dataset_cls, n_folds=5, seed=0, **dataset_params):
        ...
        self.dataset_cls = dataset_cls
        if self.dataset_cls not implements `_split_datalist`:
            raise ValueError('dataset class not supported')
        self.dataset_params = dataset_params

    def get_partition(self, fold: int, section: str):
        ...
        class _NsplitsDataset(self.dataset_cls):
            def _split_datalist(...):

        return _NsplitsDataset(**self.dataset_params)

# usage
train_ds = CrossValidation(monai.apps.DecathlonDataset, n_folds=3, seed=1234, root_dir="./").get_partition(1, "training")
```

Nic-Ma:
Hi @wyli ,

Thanks for your review, I updated this PR according to your suggestion.
I still put this `CrossValidation` in `monai.apps.datasets` because now we only have `DecathlonDataset` can be used.
When we have more suitable datasets, I agree to move it to `monai.data`.

Thanks.

monai/apps/datasets.py
```diff
@@ -306,41 +312,33 @@ def _split_datalist(self, datalist: List[Dict]) -> List[Dict]:
             return [datalist[i] for i in self.indices]
 
 
-class CVDecathlonDataset:
+class CrossValidation:
     """
-    Cross vaidation dataset based on the general `DecathlonDataset`.
+    Cross vaidation dataset based on the general dataset which must have `_split_datalist` API.
```
wyli, line 317:
```suggestion
    Cross validation dataset based on the general dataset which must have `_split_datalist` API.
```

monai/data/utils.py
```diff
@@ -704,7 +704,7 @@ def partition_dataset_classes(
     return datasets
 
 
-def generate_cross_validation_fold(partitions: Sequence[List], fold_idx: int):
+def generate_cross_validation_fold(partitions: Sequence[List], fold_idx: int) -> Tuple[List, List]:
     """
     Generate cross validation fold based on data partitions and specified fold index.
     Will select the partition with fold index as validaiton dataset and concat others as train dataset.
```
wyli, line 710:
```suggestion
    Will select the partition with fold index as validation dataset and concatenate others as train dataset.
```

monai/apps/datasets.py
```diff
@@ -352,63 +350,41 @@ class CVDecathlonDataset:
...
+        class _NsplitsDataset(self.dataset_cls):  # type: ignore
             def _split_datalist(self, datalist: List[Dict]) -> List[Dict]:
                 if section not in ["training", "validation"]:
                     raise ValueError("section must be training or validation.")
-                length = len(datalist)
-                indices = np.arange(length)
-                self.randomize(indices)
-                fold_indices = split_dataset(nsplits, length)
+                data = partition_dataset(data=datalist, num_partitions=nfolds, shuffle=True, seed=seed)
+                train, val = generate_cross_validation_fold(partitions=data, fold_idx=fold)
```
wyli, line 384: there are a few cross-validation types (https://en.wikipedia.org/wiki/Cross-validation_(statistics)), this is only implementing one of them. 

we need a more flexible API so that 
`cvdataset.get_partition(folds=[0, 2, 3])` will return a subset of the original data consisting of fold 0, 2 and 3. the user could use the subset in a training or validation phase, no need to specify it in `get_partition` interface

monai/apps/datasets.py
```diff
@@ -306,41 +312,33 @@ def _split_datalist(self, datalist: List[Dict]) -> List[Dict]:
             return [datalist[i] for i in self.indices]
 
 
-class CVDecathlonDataset:
+class CrossValidation:
```
wyli, line 315: this looks nice, might need another PR to unify the `apps.dataset` interfaces to support CV.

wyli (commented):
thanks please see some comments inline

monai/apps/datasets.py
Nic-Ma: Yes, if we can unify all the dataset APIs, that would be more clear.
Thanks.

monai/apps/datasets.py
Nic-Ma: should we specify the section (`training` or `validation`) if specify `folds=[0, 2, 3]`?

wyli: no need to specify the section name

Nic-Ma: but if only specify one fold index, will also specify section, right? Otherwise, people need to contact folds for training.

monai/apps/datasets.py
Nic-Ma: updated in the latest commit.
Thanks.

monai/data/utils.py
Nic-Ma: updated in the latest commit.
Thanks.

monai/apps/datasets.py
Nic-Ma: Hi @wyli ,

I reviewed your request and I think it's not easy to achieve, because now we return dataset (subclass of `DecathlonDataset`) directly and it can apply **transforms** to the data. But if you want to return datalists of fold 0, 3, 4 instead of dataset, that's quite different, should we skip the transforms? I think we need a quick online discussion about it.

Thanks.

wyli: it should be minor changes for the interface, previously was:
```py
cvdataset = CrossValidation(DecathlonDataset, nfolds=5, root_dir="./", seed=12345, download=True)
dataset_fold0_train = cvdataset.get_dataset(fold=0, section="training")
# execute training for fold 0 ...
```
more flexible version I'd like to have:
```py
cvdataset = CrossValidation(DecathlonDataset, nfolds=5, root_dir="./", seed=12345, download=True)
dataset_train = cvdataset.get_dataset(folds=[0, 1, 2])
dataset_val = cvdataset.get_dataset(folds=[3])
dataset_infer = cvdataset.get_dataset(folds=[4])
```

Nic-Ma: OK, looks reasonable, I thought you want to get the datalists of fold 0, 3, 4 instead of constructing Datasets.
Will try to update it later.

Thanks.

Nic-Ma: Hi @wyli ,

Updated according to your suggestion, could you please help review it again?
Thanks.

wyli (approved):
thanks, this PR also addresses https://github.com/Project-MONAI/MONAI/issues/816

PR #1181, opened 2020-11-03
[WIP] Checkpoint loader now looks for the latest model, and will resume fro…

Irme:
…m there if found, otherwise it will start training from the beginning.

Fixes # 1180

### Description
Changed the checkpoint_loader handler to resume training from the latest model, or start training if no models are found.

### Status
**Work in progress**
Testing still needs to be done

### Types of changes


Nic-Ma (commented):
Mark "pending" as it's still under discussion.
Thanks.

monai/handlers/checkpoint_loader.py
```diff
@@ -53,9 +54,12 @@ def __init__(
         load_dict: Dict,
         name: Optional[str] = None,
         map_location: Optional[Dict] = None,
+        find_latest: bool = False,
     ) -> None:
         assert load_path is not None, "must provide clear path to load checkpoint."
         self.load_path = load_path
+        assert (find_latest and self.load_path[-3:] != '.pt') or \
+               self.load_path[-3:] == '.pt' and not find_latest, "either provide an exact path, or set find latest to true"
```
rijobro, line 62: This could be simplified to an XOR (one or the other, but not both). In python, this is the `^` operator.

monai/handlers/checkpoint_loader.py
```diff
@@ -79,11 +84,36 @@ def __call__(self, engine: Engine) -> None:
...
+                    if not (key in checkpoint):
+                        checkpoint = {key: checkpoint}
+                Checkpoint.load_objects(to_load=self.load_dict,
+                                        checkpoint=checkpoint)
+                self.logger.info(
+                    f"Restored all variables from {list_model_time[-1][1]}")
+            else:
+                self.logger.info("No models found, resuming normally.")
+        else:
+            checkpoint = torch.load(self.load_path, map_location=self.map_location)
```
rijobro, line 112: There is code duplication here that could be avoided if the loading were moved into its own function.

rijobro:
@Irme I guess the unit test would simply be creating a small network, running it for a few iterations and checking the state of the various variables when it gets loaded back in.

monai/handlers/checkpoint_loader.py
```diff
@@ -79,11 +84,36 @@ def __call__(self, engine: Engine) -> None:
...
+            models = sorted(os.listdir(self.load_path))
+            self.logger.info(
+                "This directory already exisits, looking for previous models.")
+            if len(models) > 1:
+                list_model_time = []
+                times = []
+                for model_ in models:
+                    loc = os.path.join(self.load_path, model_)
+                    edit_time = os.path.getmtime(loc)
+                    times.append(edit_time)
```
ericspod, line 97: `times` isn't used anywhere, could delete?

monai/handlers/checkpoint_loader.py
```diff
@@ -79,11 +84,36 @@ def __call__(self, engine: Engine) -> None:
...
+                    list_model_time.append([edit_time, loc])
+                list_model_time.sort(key=lambda x: x[0])
```
ericspod, line 99: You can extract the most recent model file with `max((os.path.join(self.load_path, m) for m in models), os.path.getmtime)`.

monai/handlers/checkpoint_loader.py
```diff
@@ -79,11 +84,36 @@ def __call__(self, engine: Engine) -> None:
         Args:
             engine: Ignite Engine, it can be a trainer, validator or evaluator.
         """
-        checkpoint = torch.load(self.load_path, map_location=self.map_location)
-        if len(self.load_dict) == 1:
-            key = list(self.load_dict.keys())[0]
-            if not (key in checkpoint):
-                checkpoint = {key: checkpoint}
+        if self.find_latest:
+            models = sorted(os.listdir(self.load_path))
```
ericspod, line 88: `glob.glob(self.load_path+"/*.pt")` would give you a list of full paths to .pt files.

ericspod:
Hi @Irme sorry we haven't been able to circle back to this. I think the ideas we were going for here are covered by bundles now or with experiment management tools. I'm closing this now but we can reopen if you wanted to revisit things. Thanks!

PR #1186, opened 2020-11-04
1184 update dockerfile to not reinstall torch

### LINKED ISSUE #1184: PyTorch 1.6 is not optimized for docker 20.08
**Describe the bug**
In order to avoid PyTorch 1.7 CI test errors, we temporarily forced PyTorch version to >=1.5 and < 1.7, it will uninstall the optimized PyTorch in docker 20.08 and install 1.6.0 instead.
According to @IsaacYangSLA 's training test, the training speed is 2x times slower in latest MONAI docker(reinstall 1.6) because the original torch in the docker was optimized by NVIDIA team.

Need to restore the torch of PyTorch 20.08 docker as the base of MONAI docker.

wyli:
Fixes #1184 

### Description
we prefer the base docker image's torch when building the monai images,
this PR updates the dockerfile to ignore the torch version requirement from the txt file

### Status
**Ready**

### Types of changes


Nic-Ma (approved):
This hack looks good to me.
Maybe we can remove it when the PyTorch 1.7 CI issue fixed, so can you add a "TODO" or "FIXME" tag here?

Thanks.

wyli:
> This hack looks good to me.
> Maybe we can remove it when the PyTorch 1.7 CI issue fixed, so can you add a "TODO" or "FIXME" tag here?
> 
> Thanks.

I actually think this is not a temporary change. we'll never need to pip install torch in the container

PR #1191, opened 2020-11-04
enhance partition utilities

wyli:
enhances the partition utilities, part of #816 

### Status
**Ready**

### Types of changes


wyli:
addressed some corner cases, not sure whether we should use `np.random.RandomState` instead of `seed` for shuffle

Nic-Ma:
> addressed some corner cases, not sure whether we should use `np.random.RandomState` instead of `seed` for shuffle

Hi @wyli ,

Current shuffle logic can work fine cross multiple processes, I am not sure whether np.random.RandomState also work well?

Thanks.

monai/data/utils.py
```diff
@@ -661,67 +668,72 @@ def partition_dataset_classes(
...
+        >>> data = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14]
+        >>> classes = [2, 0, 2, 1, 3, 2, 2, 0, 2, 0, 3, 3, 1, 3]
+        >>> partition_dataset_classes(data, classes, shuffle=False, ratios=[2, 1])
+        [[2, 8, 4, 1, 3, 6, 5, 11, 12], [10, 13, 7, 9, 14]]
 
     """
-    data_len = len(data)
     datasets = list()
+    if not classes:
+        return []
```
Nic-Ma, line 681: I think we should raise an exception directly if no classes provided.
Thanks.

wyli:
> > addressed some corner cases, not sure whether we should use `np.random.RandomState` instead of `seed` for shuffle
> 
> Hi @wyli ,
> 
> Current shuffle logic can work fine cross multiple processes, I am not sure whether np.random.RandomState also work well?
> 
> Thanks.

ok, I was not aware of that, but since each process has a different partition, I guess it's fine to share the randomstate instance? would be great to create a few test cases so that I can understand the context. anyway we shouldn't change the global seed here.

Nic-Ma:
Hi @wyli ,

The multiple process usage is something like this: https://github.com/Project-MONAI/tutorials/blob/master/acceleration/distributed_training/unet_training_smartcache.py#L176
I am not sure whether all the processes can get the exact same shuffle result with `np.random.RandomState`?
Could you please help double confirm it? If yes, others look good to me, we can merge your PR immediately.

Thanks.

wyli:
> Hi @wyli ,
> 
> The multiple process usage is something like this: https://github.com/Project-MONAI/tutorials/blob/master/acceleration/distributed_training/unet_training_smartcache.py#L176
> I am not sure whether all the processes can get the exact same shuffle result with `np.random.RandomState`?
> Could you please help double confirm it? If yes, others look good to me, we can merge your PR immediately.
> Then I can create a PR to update the examples of distributed training.
> 
> Thanks.

i don’t see any issue

Nic-Ma (approved):

Issue #1023, opened 2020-09-11
Deconvolutions improve performance of AH-Net


mmarcinkiewicz:
**Describe the bug**
In the current environment deconvolutions work significantly faster with mixed precision than trilinear interpolation for upsampling. In the current Monai implementation of AH-Net the `upsample_mode` defaults to "trilinear", which is suboptimal.  

**To Reproduce**
Run training with `upsample_mode="trilinear"` and `upsample_mode="transpose"`.

**Expected behavior**
The default `upsample_mode` is changed to `transpose`.

**Screenshots**
Throughput measured on V100 and A100 for a 3D workload.
![image](https://user-images.githubusercontent.com/43240942/92876049-f75a6180-f409-11ea-9bdf-0851fd7dca40.png)

**Environment (please complete the following information):**
 - Ubuntu 18.04
 - Python 3.6
 - MONAI version f1998b72a941d1e5f9578a66dc1c20b01913caab
 - CUDA 11, cuDNN 8.0.3/8.0.4
 - GPUs: V100, A100

**Additional context**
In cuDNN 8.0.3 deconvolutions have Tensor Core support which improves their performance significantly. 
Currently there are issues being tracked for Pytorch and cuDNN which might improve performance. I'll monitor the status and update if needed.

Nic-Ma:
Hi @yiheng-wang-nv ,

I remember you fixed some determinism issue about the upsampling in AHNet, is it related to this ticket?

Thanks.

wyli:
> Hi @yiheng-wang-nv ,
> 
> I remember you fixed some determinism issue about the upsampling in AHNet, is it related to this ticket?
> 
> Thanks.

that's noted in the docstring
https://github.com/Project-MONAI/MONAI/blob/8fc4f5dd5ea4d3726fdd9e001eacf517263625ac/monai/networks/nets/ahnet.py#L353-L355

yiheng-wang-nv:
> Hi @yiheng-wang-nv ,
> 
> I remember you fixed some determinism issue about the upsampling in AHNet, is it related to this ticket?
> 
> Thanks.

FYI, the non-determinism issue is explained in an old [PR](https://github.com/Project-MONAI/MONAI/pull/815). Generally speaking, `torch.nn.functional.interpolate() ` and `torch.nn.MaxPool1d/2d/3d()` use `atomicAdd ` which introduce the non-determinism results. 
This ticket is about the computing speed between `trilinear` and `deconvolution`.

PR #1197, opened 2020-11-06
Refine ahnet to support changeable psp module

### LINKED ISSUE #1165: Make several layers of AHNet to be optional
Currently, the implementation of AHNet aligns with the original paper, which is very heavy.
It may require more than 10 times memory than our UNet.
So I think we can make some deep layers optional, users can select to use them or not with 1 parameter.

yiheng-wang-nv:

Fixes #1165 

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @yiheng-wang-nv ,

Could you please help train with spleen dataset and see the speed up comparing to full structure version?
Thanks.

yiheng-wang-nv:
> Hi @yiheng-wang-nv ,
> 
> Could you please help train with spleen dataset and see the speed & accuracy comparing to the full structure version?
> Thanks.

Sure

wyli (approved):
thanks, please confirm the performance and merge.

Nic-Ma:
I am testing the performance on spleen dataset, will update soon.
Thanks.

Nic-Ma:
I tried to train several times in different cases.
The training accuracy is 0.95 which is OK enough, but the training speed is not expected:
I tested with our AMP notebook: https://github.com/Project-MONAI/tutorials/blob/master/acceleration/automatic_mixed_precision.ipynb
Just replaced the UNet with AHNet.

The AMP training is even **slower** than non-AMP training:
On V100 & PyTorch docker 20.10: AMP speed / step = **0.98s**, non-AMP speed / step = **0.87s**.

And to double confirm this issue is not related to HW or SW env, I also changed back to our UNet and tested again:
AMP speed / step = **0.23s**, non-AMP speed / step = **0.39s**.
It's expected and better than PyTorch docker 20.08.

@yiheng-wang-nv could you please help debug deeper on this issue?

Thanks.

Nic-Ma:
Seems the issue is related to the `upsampling` method.
When I changed to use `transpose`: AMP speed / step = **0.54s**, non-AMP speed / step = **0.91s**.
But `transpose` can't support input size as small as `trilinear`(like [96, 96, 96]), need to discuss it.

Thanks.

wyli:
> Seems the issue is related to the `upsampling` method.
> 
> When I changed to use `transpose`: AMP speed / step = **0.54s**, non-AMP speed / step = **0.91s**.
> 
> But `transpose` can't support input size as small as `trilinear`(like [96, 96, 96]), need to discuss it.
> 
> 
> 
> Thanks.
> 
> 

isn't this known from https://github.com/Project-MONAI/MONAI/issues/1023

Nic-Ma:
Hi @wyli ,

Yes, I think it's same as the test results in that link.
Currently, we prefer to use `transpose` in our project and it's deterministic, `trilinear` can't support determinism.
The only problem is that it requires the input size must be 64x, can't support [96, 96, 96].

Thanks.

yiheng-wang-nv:
> Hi @wyli ,
> 
> Yes, I think it's same as the test results in that link.
> Currently, we prefer to use `transpose` in our project and it's deterministic, `trilinear` can't support determinism.
> The only problem is that it requires the input size must be 64x, can't support [96, 96, 96].
> 
> Thanks.

Hi @Nic-Ma @wyli , I've added `nearest` upsample mode, it is a deterministic interpolation method and the size requirement is the same at other linear interpolation modes.
In order to maintain the whole network's determinism, I used the same kernel size as `transpose` upsample mode does for the `maxpool` layer (as shown from line 390 in ahnet.py):
```
        if upsample_mode in ["transpose", "nearest"]:
            """
            To maintain the determinism, the value of kernel_size and stride should be the same.
            (you can check this link for reference: https://github.com/Project-MONAI/MONAI/pull/815 )
            """
            self.maxpool = pool_type(kernel_size=(2, 2, 2)[-spatial_dims:], stride=2)
        else:
            self.maxpool = pool_type(kernel_size=(3, 3, 3)[-spatial_dims:], stride=2, padding=1)
```

Nic-Ma:
Hi @yiheng-wang-nv ,

Thanks for your update.
Using `nearest` mode, AMP speed / step = **0.52s**, non-AMP speed / step = **0.82s**, looks reasonable.
And the loss value of every step in both AMP and non-AMP is deterministic.
Let me try to train on spleen dataset again and test the accuracy with it.

Thanks.

Nic-Ma (approved):
Looks good to me now.
Thanks.

Nic-Ma:
With `mode=nearest` and `input size=[128, 128, 128]`, `slidingwindow=[192, 192, 192]`, `psp=3`, got: 0.946.

PR #1203, opened 2020-11-07
[#1202] add selectitemd transform and tests

### LINKED ISSUE #1202: Select Items Dictionary Transform
**Problem**
Currently, the only way to remove dictionary keys it to remove them with the `DeleteItemsd` transform. Though this solves the problem it is not the most elegant way to do it. For example, suppose you have numerous different NIfTI files and use the pipeline to combined them into a single channels wise input. In this case, I find that selecting what you want to keep is more elegant (and less prone to error) then listing everything you want to remove.

**The solution**
A transform with the same workings as `DeleteItemsd` but which includes instead of excludes.

nielsdebruin:
Signed-off-by: Niels de Bruin <nielsdebruin@gmail.com>

Fixes #1202 

### Description
Adds a SelectItemsd transform as per #1202

### Status
**Ready**

### Types of changes


wyli (approved):
thanks, this is useful. could you please add the class name to the documentation, after `deleteitemsd` https://github.com/Project-MONAI/MONAI/blob/master/docs/source/transforms.rst#deleteitemsd

nielsdebruin:
Thanks, I have added SelectItemsd to transforms.rst

Issue #1209, opened 2020-11-09
CI/CD pull from dockerhub reaches the rate limit


wyli:
**Is your feature request related to a problem? Please describe.**
"/usr/bin/docker pull projectmonai/monai:latest
  Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit"

https://github.com/Project-MONAI/MONAI/runs/1375688986?check_suite_focus=true#step:2:17

**Additional context**
https://github.com/Project-MONAI/MONAI/blob/fcd72ea201e9599be23114023604c4653e80fc56/.github/workflows/setupapp.yml#L137-L138

Nic-Ma:
Hi @IsaacYangSLA and @YuanTingHsieh ,

Could you please help take a look?
Thanks.

PR #1211, opened 2020-11-10
Temporarily switch to local docker registry

IsaacYangSLA:
Temporarily switch to local docker registry to avoid dockerhub rate limit on self-hosted runners

Fixes https://github.com/Project-MONAI/MONAI/issues/1209

### Description
Push/pull new docker images to/from local registry so self-hosted runners will not be blocked by dockerhub.

### Status
**Ready**

### Types of changes

### DIFF
--- a/.github/workflows/setupapp.yml
+++ b/.github/workflows/setupapp.yml
@@ -138,10 +138,19 @@ jobs:
       env:
         QUICKTEST: True
 
+  local_docker:
+    runs-on: [self-hosted, linux, x64]
+    steps:
+    - uses: actions/checkout@v2
+    - name: docker_build
+      run: |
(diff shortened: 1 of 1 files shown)


wyli:
thanks, could you sign @IsaacYangSLA? I think the cron test could still use the dockerhub, it'll only run once a day.

wyli (approved):

IsaacYangSLA:
> thanks, could you sign @IsaacYangSLA? I think the cron test could still use the dockerhub, it'll only run once a day. we need the tests for the actual dockerhub release content

Thanks, @wyli .  I removed the reference to local registry in cron.yml while keeping it in setupapp.yml.

PR #1221, opened 2020-11-11
add fcnet

rijobro:
### Description
Port FCNet from PTProto.

### Status
**Ready**

### Types of changes


monai/networks/nets/fcnet.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        super().__init__()
+        self.in_channels = in_channels
+        self.out_channels = out_channels
+        self.hidden_channels = list(hidden_channels)
+        self.dropout = dropout
+        self.flatten = nn.Flatten()
+        self.hiddens = nn.Sequential()
+
+        act_name, act_args = split_args(act)
```
wyli, line 44: would be great to use the `ADN` layer for flexibility:
https://github.com/Project-MONAI/MONAI/blob/bd1e008c36adcca64658aa10178321d4cd6854fb/monai/networks/blocks/acti_norm.py#L20-L24

monai/networks/nets/fcnet.py
```diff
@@ -0,0 +1,155 @@
...
+        self.in_channels = in_channels
+        self.out_channels = out_channels
+        self.latent_size = latent_size
+        self.dropout = dropout
+
+        self.encode = nn.Sequential()
+        self.decode = nn.Sequential()
+        self.flatten = nn.Flatten()
+
+        act_name, act_args = split_args(act)
```
wyli, line 98: same here for the `ADN`

monai/networks/nets/fcnet.py
```diff
@@ -0,0 +1,155 @@
...
+        return seq
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = self.flatten(x)
+        x = self.hiddens(x)
+        x = self.output(x)
+        return x
+
+
+class VarFCNet(nn.Module):
```
wyli, line 72: need a small unit test for this

wyli (commented):
thanks! please see some comments inline

monai/networks/nets/fcnet.py
rijobro: Could you be a little more explicit on this one?

wyli: yes, using ADN would make it more flexible with the element-wise operations. this logic
```py
if self.dropout:
    seq.add_module("dropout", nn.Dropout(self.dropout))
seq.add_module("act", self.act)
```
could be replaced by:
```py
seq.add_module('ADN', ADN(dropout=dropout_params, act=act_params, norm=norm_params, ordering=ordering))
```
with user's input:
```py
norm_params=None
dropout_params=...
act_params=...
ordering="AD"
```

monai/networks/nets/fcnet.py
```diff
@@ -0,0 +1,155 @@
...
+            seq.add_module("dropout", nn.Dropout(self.dropout))
+
+        seq.add_module("act", self.act)
+
+        return seq
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = self.flatten(x)
+        x = self.hiddens(x)
+        x = self.output(x)
```
wyli: looks like this is a sequential model, perhaps inheriting `nn.Sequential` is more convenient than `nn.Module`

monai/networks/nets/fcnet.py
rijobro: Ok, updated. Hope it's what you had in mind.

wyli: thanks it looks good to me now

rijobro:
@wyli in think the current error is a [bug](https://github.com/sphinx-doc/sphinx/issues/4685) in sphinx. Or if it's not a bug, I can't figure out what's wrong with the docstring...

monai/networks/nets/fullyconnectednet.py
```diff
@@ -0,0 +1,153 @@
...
+        out_channels: int,
+        hidden_channels: Sequence[int],
+        dropout: Optional[Union[Tuple, str, float]] = None,
+        act: Optional[Union[Tuple, str]] = Act.PRELU,
+        bias: bool = True,
+        adn_ordering: Optional[str] = None,
+    ) -> None:
+        """
+        Defines a network accept input with `in_channels' channels, output of `out_channels' channels, and hidden layers
+        with channels given in `hidden_channels'. If `bias' is True then linear units have a bias term.
```
wyli, line 49: the warnings were because of the syntax 
```suggestion
        Defines a network accept input with `in_channels` channels, output of `out_channels` channels, and hidden layers
        with channels given in `hidden_channels`. If `bias` is True then linear units have a bias term.
```

monai/networks/nets/fullyconnectednet.py
```diff
@@ -0,0 +1,153 @@
...
+    def _get_layer(self, in_channels: int, out_channels: int, bias: bool) -> nn.Sequential:
+        seq = nn.Sequential(nn.Linear(in_channels, out_channels, bias))
+        seq.add_module("ADN", self.adn_layer)
+        return seq
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = self.flatten(x)
+        x = self.hiddens(x)
+        x = self.output(x)
+        return x
```
wyli, line 76: I think if we use `nn.sequential` we could get rid of the `forward` method, something like (not tested):
```suggestion
        super().__init__()
        self.in_channels = in_channels
        self.out_channels = out_channels
        self.hidden_channels = list(hidden_channels)
        self.add_module('flatten', nn.Flatten())
        self.adn_layer = _get_adn_layer(act, dropout, adn_ordering)

        prev_channels = self.in_channels
        for i, c in enumerate(hidden_channels):
            self.add_module("hidden_%i" % i, self._get_layer(prev_channels, c, bias))
            prev_channels = c
        self.add_module('output', nn.Linear(prev_channels, out_channels, bias))

    def _get_layer(self, in_channels: int, out_channels: int, bias: bool) -> nn.Sequential:
        seq = nn.Sequential(nn.Linear(in_channels, out_channels, bias))
        seq.add_module("ADN", self.adn_layer)
        return seq
```

wyli (commented):
thanks, I think we can simplify the implementation. for the docstring fixes I tested locally and works fine

monai/networks/nets/fullyconnectednet.py
rijobro: Thanks, just put this in and local tests seem fine.

wyli (approved):
thanks!

PR #1222, opened 2020-11-11
1189 flexible sliding window device

### LINKED ISSUE #1189: Memory error for sliding_window_inference
Hi MONAI-Team,
thanks for sharing your software development so far!
I tried to adapt the 3D spleen segmentation example to my own dataset. After some adjustments I got the training to run and as a next step tried to run some inference on my test data. The test volume is a 1504x1504x561 CT volume in nifti format.
To do so I used the sliding_window_inference on the CPU as the GPU immediatly runs out of memory.

```
test_transforms = Compose(
    [
        LoadNiftid(keys=["image", "label"]),
        AddChanneld(keys=["image", "label"]),
        Orientationd(keys=["image", "label"], axcodes="RAS"),
        CenterSpatialCropd(keys=["image", "label"], roi_size=[1000, 1000, 561]),
        ToTensord(keys=["image", "label"]),
    ]
)

test_ds = CacheDataset(data=test_files, transform=test_transforms, cache_rate=1, num_workers=1)
test_loader = DataLoader(test_ds, batch_size=1, num_workers=1)

test_device = torch.device("cpu")
model = UNet(
    dimensions=3,
    in_channels=1,
    out_channels=2,
    channels=(16, 32, 64, 128),
    strides=(2, 2, 2, 2),
    num_res_units=2,
    norm=Norm.BATCH,
).to(test_device)
loss_function = DiceLoss(to_onehot_y=True, softmax=True)
optimizer = torch.optim.Adam(model.parameters(), 1e-4)

model.load_state_dict(torch.load(os.path.join(root_dir, "best_metric_model.pth")))
model.eval()

for test_data in test_loader:
    val_inputs, val_labels = (
        test_data["image"].to(test_device),
        test_data["label"].to(test_device),
    )
    print(f"batch_data image: {test_data['image'].shape}")
    print(f"batch_data label: {test_data['label'].shape}")
    roi_size = (64, 64, 64)
    sw_batch_size = 1
    test_outputs = sliding_window_inference(val_inputs, roi_size, sw_batch_size, model)
```

Unfortunately, I get either an out-of-memory error, a data loader error or the kernel dies. In each case the RAM of my machine (360 GB) runs out of memory during the sliding window inference. Therefore, I believe that all three errors are caused by the pilling up of the data in the RAM. 

What am I doing wrong? Did I miss interpreted something? I my understanding the ROI in the sliding window inferer should crop the sample to sub-volumes of the roi_size which then are used for inference by the network. Therefore, the memory footprint should be controlable by the roi_size and should even run on the GPU. But even if the entire sample is stored it should not require more than a couple of GBs (size of the nifti = 1.7 GB).  test_data does only consist of this single sample.
Thank you very much for your help!

wyli:
Fixes #1189

### Description
adds a `sw_device` option to slidingwindowinference

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -482,7 +482,7 @@ def compute_importance_map(
     patch_size: Tuple[int, ...],
     mode: Union[BlendMode, str] = BlendMode.CONSTANT,
     sigma_scale: Union[Sequence[float], float] = 0.125,
-    device: Optional[torch.device] = None,
+    device: Union[torch.device, int, str] = "cpu",
 ) -> torch.Tensor:
     """Get importance map for different weight modes.
 
@@ -506,6 +506,7 @@ def compute_importance_map(
 
     """
     mode = BlendMode(mode)
+    device = torch.device(device)  # type: ignore[arg-type]
     if mode == BlendMode.CONSTANT:
         importance_map = torch.ones(patch_size, device=device).float()
     elif mode == BlendMode.GAUSSIAN:
--- a/monai/inferers/inferer.py
+++ b/monai/inferers/inferer.py
@@ -87,7 +87,13 @@ class SlidingWindowInferer(Inferer):
             Padding mode when ``roi_size`` is larger than inputs. Defaults to ``"constant"``
             See also: https://pytorch.org/docs/stable/nn.functional.html#pad
         cval: fill value for 'constant' padding mode. Default: 0
-
+        sw_device: device for the window data.
+            By default the device (and accordingly the memory) of the `inputs` is used.
+            Normally `sw_device` should be consistent with the device where `predictor` is defined.
+        device: device for the stitched output prediction.
+            By default the device (and accordingly the memory) of the `inputs` is used. If for example
+            set to device=torch.device('cpu') the gpu memory consumption is less and independent of the
+            `inputs` and `roi_size`. Output is on the `device`.
 
     Note:
         ``sw_batch_size`` denotes the max number of windows per network inference iteration,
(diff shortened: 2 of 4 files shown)


Nic-Ma:
Hi @wyli ,

May I know what's the use case for this `sw_device`?
For example, if you want to run SlidingWindow on CPU, you just put the `inputs` and `model` to CPU, if GPU, put to GPU, SlidingWindow module doesn't need to take care of it, right?
I can only imagine 1 use case that the `inputs` is bigger than GPU memory, so you can only put 1 window data to GPU at runtime, is it the case?
If must be needed, could you please change the original `device` parameter to `concat_device` to be consistent?

Thanks.

wyli:
> Hi @wyli ,
> 
> May I know what's the use case for this `sw_device`?
> For example, if you want to run SlidingWindow on CPU, you just put the `inputs` and `model` to CPU, if GPU, put to GPU, SlidingWindow module doesn't need to take care of it, right?
> I can only imagine 1 use case that the `inputs` is bigger than GPU memory, so you can only put 1 window data to GPU at runtime, is it the case?
> If must be needed, could you please change the original `device` parameter to `concat_device` to be consistent?
> 
> Thanks.

yes that's the use case, I dont want to introduce breaking change to "device" for it.

Nic-Ma:
Hi @wyli ,

I feel `sw_device` & `device` is very confusing because the `device` here is only used to concat final results.
Others look good to me.
@ericspod what do you guys think?

Thanks.

wyli:
it's not confusing because device is the primary output device, sw_device is just for the intermediate computation. normally you don't need to care about sw_device

Nic-Ma:
OK, let's don't break the API, but could you please help add a more explicit description in the doc-string of `device`? Make sure user understands it doesn't mean the device for **SlidingWindow computation**?

Thanks.

wyli:
'device' is never used in the window prediction, this pr doesnt change the behaviour of 'device'. I really don't understand your concern here. I've already updated the docstring saying that device is for the stitches output.

Nic-Ma:
OK, let's update it later if user asks questions.
Thanks.

Nic-Ma (approved):

PR #1238, opened 2020-11-16
Torchscript Fixes

### LINKED ISSUE #1160: Add TorchScript support for networks
Now we already fixed some `TorchScript` issues for several networks and added unit tests for them.
Still need to resolve `TorchScript` issues in other networks: `SegResNet & SegResNetVAE`, `DenseNet`, `SENet`, `DynUNet`.

`torch.jit.script()` can work for these networks.

ericspod:
Fixes #1160.

### Description
This introduces a number of changes to networks and their components to be compatible with Torchscript. Every network is being tested, except `DynUNet`, by converting it to a Torchscript script then back to a network before running a forward pass of synthetic data through it. This should ensure that a saved network represents a valid Pytorch object which can be reloaded and used without needing the original codebase. 

The `DynUNet` is not currentl compatible because it relies on iteration through lists of modules or the ListModule class. These are not currently compatible with Torchscript so a different architecture for this class is needed. It's possible to inherit from the existing UNet which is structured differently for this reason, however Torchscript does also have issues with the `super` keyword so a little more design thought is needed to account for the variation in the `DynUNet.forward` method. 

Changes were made to the squeeze and excitation classes to also account for Torchscript's lack of support for `forward`. Other things not supported and removed include the `reverse` function and `Tensor.ndimension` method. 

The return types for some methods of `UNet` were also made more general to accommodate future subtypes which might override these methods to return different subtypes of `nn.Module`. Other classes might need to be modified in a similar way in the future for the same reason.

### Status
**Ready**

### Types of changes


monai/networks/blocks/squeeze_and_excitation.py
```diff
@@ -74,18 +77,24 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        if self.add_residual:
+            result += x
+
+        return result
 
 
 class ResidualSELayer(ChannelSELayer):
     """
-    A "squeeze-and-excitation"-like layer with a residual connection::
+    A "squeeze-and-excitation"-like layer with a residual connection.
```
wyli, line 93: this `::` is required for rendering the plot in fixed-width font

ericspod: Fixed, the warnings from sphinx were cryptic about the issue.

wyli:
thanks, these are non-breaking changes and look good to me. could you please help double-check this PR, it's important for the upcoming release and Clara train @Nic-Ma @yiheng-wang-nv

Nic-Ma (approved):
Hi @ericspod ,

Really nice enhancement!!
Looks good to me, please fix the CI and sign-off then we can merge it.
And I will submit another ticket to keep tracking the TorchScript support for DynUNet.
@yiheng-wang-nv could you please help double confirm this PR?

Thanks.

Nic-Ma:
Submitted ticket https://github.com/Project-MONAI/MONAI/issues/1245 to track the DynUNet support.
Thanks.

yiheng-wang-nv (approved):

wyli:
Hi @ericspod  it's going to be difficult to sign all the commits here, perhaps you could squash locally into one signed commit and then force push? https://stackoverflow.com/questions/25356810/git-how-to-squash-all-commits-on-branch

ericspod:
Squashing things down isn't working. There's some sort of conflict in the history on my fork which should have been overwritten and removed at some point but is now creating conflicts in actual source files if I try to squash things.

wyli:
sure I'll try to squash them and let you sign it. btw it seems there's an issue with window SENet test

wyli:
I couldn't force change this PR... but this works fine:
(generate a patch and then apply the patch to a clean base)
```bash
# git checkout this_pr_branch
git diff origin/master this_pr_branch > ~/Desktop/patch
git reset --hard origin/master
git apply ~/Desktop/patch
git commit -sam 'single commit'
```

ericspod:
There appears to be no difference between `origin/master` and the PR branch `pull/1238/head` so this doesn't end up doing anything as there's nothing to commit.

ericspod:
The Windows test might have timed out, there's no error output just the line stating the process exited with an error condition. The download for the SENet data took 5 minutes so that might be why it's a problem.

wyli:
> There appears to be no difference between `origin/master` and the PR branch `pull/1238/head` so this doesn't end up doing anything as there's nothing to commit.

ok, I shouldn't assume origin/master is the latest upsteam, you can replace all `origin/master` with `09f39dcb84092b07cda480c99644f9f7f8cceab6`  (the current latest upstream hash) in those commands

ericspod:
That still did nothing, there was a patch but the commit and push did nothing.

wyli:
ok... at least you get the patch, I guess the simplest way is to start a new PR with that patch

PR #1250, opened 2020-11-17
Fix confusion matrix issue

### LINKED ISSUE #1244: Issues with compute_confusion_matrix metric
**Describe the bug**
There are 2 minor issues in `compute_confusion_matrix` metric:
1. If reduction mode is `SUMBATCH`, should not do SUM on `not_nans`, otherwise we don't know which class is nan when computing the final result.
2. If `compute_sample` is False, should return the confusion matrix elements of current batch directly, then compute the final metric value for the whole epoch. Current implementation computes metric for every batch, if batch size changed, the metric value will be different.
So in summary, should have 2 modes:
(1) Compute metric for every sample image first, then average or sum the whole epoch to get the final result.
(2) Keep the confusion matrix elements of every image, then average or sum the whole epoch, then compute metric.

yiheng-wang-nv:

Fixes #1244 .

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/handlers/confusion_matrix.py
+++ b/monai/handlers/confusion_matrix.py
@@ -13,7 +13,8 @@ from typing import Callable, Optional, Sequence
 
 import torch
 
-from monai.metrics import ConfusionMatrixMetric
+from monai.metrics import ConfusionMatrixMetric, compute_confusion_matrix_metric
+from monai.metrics.utils import MetricReduction, do_metric_reduction
 from monai.utils import exact_version, optional_import
 
 NotComputableError, _ = optional_import("ignite.exceptions", "0.4.2", exact_version, "NotComputableError")
@@ -31,8 +32,7 @@ class ConfusionMatrix(Metric):  # type: ignore[valid-type, misc] # due to option
         self,
         include_background: bool = True,
         metric_name: str = "hit_rate",
-        compute_sample: bool = True,
-        output_class: bool = False,
+        compute_sample: bool = False,
         output_transform: Callable = lambda x: x,
         device: Optional[torch.device] = None,
     ) -> None:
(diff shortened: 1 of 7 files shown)


Nic-Ma (approved):
Looks good to me.
Thanks.

monai/metrics/confusion_matrix.py
```diff
@@ -92,21 +96,10 @@ def __call__(self, y_pred: torch.Tensor, y: torch.Tensor):
 
         if self.compute_sample:
             confusion_matrix = compute_confusion_matrix_metric(self.metric_name, confusion_matrix)
-            if self.output_class:
-                f, not_nans = do_metric_reduction(confusion_matrix, MetricReduction.MEAN_BATCH)
-            else:
-                f, not_nans = do_metric_reduction(confusion_matrix, MetricReduction.MEAN)
+            f, not_nans = do_metric_reduction(confusion_matrix, self.reduction)
```
wyli, line 99: could you add support for multiple metrics here? for example `metric_name=("sensitivity", "specificity")` it should return `f_0, not_nans_0, f_1, not_nans_1`

wyli (commented):
thanks, please see some comments inline

monai/metrics/confusion_matrix.py
Nic-Ma: Hi @wyli ,

Can we support this feature in `metrics` but not in `metrics ignite handler`?
If returning many items, it's not easy to support distributed data parallel.

Thanks.

wyli: yes, for now let's do it for the metrics. thanks!

monai/handlers/confusion_matrix.py
```diff
@@ -86,20 +94,34 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
...
+            self._num_examples += not_nans
+        else:
+            confusion_matrix = self.confusion_matrix(y_pred, y)
+            confusion_matrix, _ = do_metric_reduction(confusion_matrix, MetricReduction.SUM)
+            self._total_tp += confusion_matrix[0].item()
+            self._total_fp += confusion_matrix[1].item()
+            self._total_tn += confusion_matrix[2].item()
+            self._total_fn += confusion_matrix[3].item()
+
+    @sync_all_reduce("_sum", "_num_examples", "_total_tp", "_total_fp", "_total_tn", "_total_fn")
```
Nic-Ma, line 112: Could you please add a distributed test or at least do it locally to make this `sync_all_reduce` can work fine?
Refer to this test code: https://github.com/Project-MONAI/MONAI/blob/master/tests/test_handler_rocauc_dist.py

Thanks.

Nic-Ma (approved):

Nic-Ma (approved):
Looks good to me now.
Thanks.

PR #1237, opened 2020-11-16
Add integration ChatOps

behxyz:
Fixes #801

### Description
Add integration test ChatOps that is being triggered by mentioning `/integration-test` in the first line of a pull request comment. Upon successful completion of integration tests, it adds a 🚀 reaction to the PR comment.

### Status
**Ready**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh --codeformat --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

PR #1261, opened 2020-11-18
fixes chatops integration tests

wyli:
follow up of https://github.com/Project-MONAI/MONAI/pull/1237

### Description
checkout the pull request ref branch instead of master

### Status
**Ready**

### Types of changes

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -11,7 +11,12 @@ jobs:
       options: --gpus all
     runs-on: [self-hosted, linux, x64, common]
     steps:
-    - uses: actions/checkout@v2
+    # checkout the pull request branch
+    - uses: actions/checkout@v2.1.0
+      with:
+        token: ${{ secrets.PR_MAINTAIN }}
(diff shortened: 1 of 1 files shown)


behxyz:
@wyli all good but it doesn't have to be a specific version `actions/checkout@v2.1.0`. In other tests in MOANI, we are using `actions/checkout@v2` (which is right not at `v2.3.3`), it would be better to be consistent and refer to `v2` tag.

.github/workflows/integration.yml
```diff
@@ -11,7 +11,12 @@ jobs:
       options: --gpus all
     runs-on: [self-hosted, linux, x64, common]
     steps:
-    - uses: actions/checkout@v2
+    # checkout the pull request branch
+    - uses: actions/checkout@v2.1.0
```
behxyz: It would be nice to keep `actions/checkout@v2` here.

behxyz (approved):

.github/workflows/integration.yml
wyli: thanks!

PR #1262, opened 2020-11-19
Enhance prepare_batch callback in workflows

Nic-Ma:
### Description
According to our internal discussion today, it's better to pass `engine` to `prepare_batch` directly, instead of passing `engine.state.device` and `engine.non_blocking`.
It can also support more flexible use cases to leverage other components of `engine` in `prepare_batch` callback

### Status
**Ready**

### Types of changes


wyli:
thanks, this could also be achieved by attaching customised event handlers to the engine, I'm closing this for now. we can revisit this topic if there are specific use cases

PR #1268, opened 2020-11-19
fix windows test_senet CI

rijobro:
### Description
Fix windows test_senet by not reusing networks for different tests.

Also speed up all uses of `test_script_save`, which will be performed with CUDA if available, falling back to the CPU if not.

### Status
**Ready**

### Types of changes

### DIFF
--- a/tests/test_senet.py
+++ b/tests/test_senet.py
@@ -26,13 +26,13 @@ from tests.utils import test_pretrained_networks, test_script_save
 
 device = "cuda" if torch.cuda.is_available() else "cpu"
 
-net_args = {"spatial_dims": 3, "in_channels": 2, "num_classes": 2}
-TEST_CASE_1 = [senet154, net_args]
-TEST_CASE_2 = [se_resnet50, net_args]
-TEST_CASE_3 = [se_resnet101, net_args]
-TEST_CASE_4 = [se_resnet152, net_args]
-TEST_CASE_5 = [se_resnext50_32x4d, net_args]
-TEST_CASE_6 = [se_resnext101_32x4d, net_args]
+NET_ARGS = {"spatial_dims": 3, "in_channels": 2, "num_classes": 2}
+TEST_CASE_1 = [senet154, NET_ARGS]
+TEST_CASE_2 = [se_resnet50, NET_ARGS]
+TEST_CASE_3 = [se_resnet101, NET_ARGS]
+TEST_CASE_4 = [se_resnet152, NET_ARGS]
+TEST_CASE_5 = [se_resnext50_32x4d, NET_ARGS]
+TEST_CASE_6 = [se_resnext101_32x4d, NET_ARGS]
 
 TEST_CASE_PRETRAINED = [se_resnet50, {"spatial_dims": 2, "in_channels": 3, "num_classes": 2, "pretrained": True}]
 
--- a/tests/test_senet.py
+++ b/tests/test_senet.py
@@ -49,7 +49,7 @@ class TestSENET(unittest.TestCase):
     @parameterized.expand([TEST_CASE_1, TEST_CASE_2, TEST_CASE_3, TEST_CASE_4, TEST_CASE_5, TEST_CASE_6])
     def test_script(self, net):
         input_data = torch.randn(2, 2, 64, 64, 64)
-        out_orig, out_reloaded = test_script_save(net.cpu(), input_data)
+        out_orig, out_reloaded = test_script_save(net, input_data)
         assert torch.allclose(out_orig, out_reloaded)
 
 
--- a/tests/test_senet.py
+++ b/tests/test_senet.py
@@ -26,31 +26,34 @@ from tests.utils import test_pretrained_networks, test_script_save
 
 device = "cuda" if torch.cuda.is_available() else "cpu"
 
-TEST_CASE_1 = [senet154(3, 2, 2)]
-TEST_CASE_2 = [se_resnet50(3, 2, 2)]
-TEST_CASE_3 = [se_resnet101(3, 2, 2)]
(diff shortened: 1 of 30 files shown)


wyli:
thanks for looking into this! I couldn't figure out the root cause for the windows test, was about to submit a PR to remove the test as well...

rijobro:
@wyli It seems that the windows test was still failing in my initial commit so i've commented out the pre-trained test to see if that fixes. If not, I'll comment out the other tests until I figure out which is the problematic one.

rijobro:
@wyli @ericspod that seems to have fixed it, the problem was reusing the network for two different tests (`test_senet_shape` and `test_script`).

wyli:
> @wyli @ericspod that seems to have fixed it, the problem was reusing the network for two different tests (`test_senet_shape` and `test_script`).

thanks, there's another issue now with `torch.allclose` for the cuda tensors. I've checked the default `atol=1e-8` which is too strict... `torch.allclose(out_orig, out_reloaded, atol=1e-7)` works fine for me locally

rijobro:
@wyli done. Moved the comparison inside of the `test_script_save` function.

wyli:
ok, looks like 1e-7 is still too small... still need to increase atol

tests/utils.py
```diff
@@ -143,7 +151,15 @@ def test_script_save(net, *inputs, eval_nets=True):
         result1 = net(*inputs)
         result2 = reloaded_net(*inputs)
```
wyli, line 153: not 100% sure if it's relevant, perhaps we need to set the deterministic flag to ensure the reproducibility...
https://pytorch.org/docs/stable/notes/randomness.html
```suggestion
        monai.utils.set_determinism(seed=0)
        result1 = net(*inputs)
        result2 = reloaded_net(*inputs)
        monai.utils.set_determinism(seed=None)
```

ericspod:
I hadn't every encountered the precision issue with `allclose`, either in Torch or Numpy as you have it now. Is this a precision issue with GPU computation?

wyli:
> I hadn't every encountered the precision issue with `allclose`, either in Torch or Numpy as you have it now. Is this a precision issue with GPU computation?

yes but it works fine with pytorch 1.5

rijobro:
I'm getting some pretty wild differences now (e.g., [here](https://github.com/Project-MONAI/MONAI/pull/1268/checks?check_run_id=1426105557#step:6:7741)), so that might indicate a bug. I'll switch the computation back to the CPU tomorrow to see if it's my fault or not.

P.S. I switched from `torch.allclose` to `np.testing.all_close` as this prints the relative and absolute differences, which is obviously helpful during debugging.

tests/test_senet.py
```diff
@@ -26,31 +26,33 @@
...
-TEST_CASE_4 = [se_resnet152(3, 2, 2)]
-TEST_CASE_5 = [se_resnext50_32x4d(3, 2, 2)]
-TEST_CASE_6 = [se_resnext101_32x4d(3, 2, 2)]
+net_args = {"spatial_dims": 3, "in_channels": 2, "num_classes": 2}
+TEST_CASE_1 = [senet154, net_args]
+TEST_CASE_2 = [se_resnet50, net_args]
+TEST_CASE_3 = [se_resnet101, net_args]
+TEST_CASE_4 = [se_resnet152, net_args]
+TEST_CASE_5 = [se_resnext50_32x4d, net_args]
+TEST_CASE_6 = [se_resnext101_32x4d, net_args]
```
YuanTingHsieh, line 35: let's call this NET_ARGS then since it is a constant kind of?
and to avoid warnings below (cuz we have net_args below as well)

tests/utils.py
```diff
@@ -123,27 +123,44 @@ def setUp(self):
...
-
-    return result1, result2
+        set_determinism(seed=None)
+    # When using e.g., VAR, we will produce a tuple of outputs.
+    # Hence, convert all to tuples and then compare all elements.
+    if not isinstance(result1, tuple):
+        retult1 = (result1,)
+        retult2 = (result2,)
```
YuanTingHsieh, line 159: typos here
should be 
result1 = (result1, )
result2 = (result2, )

tests/utils.py
```diff
@@ -123,27 +123,44 @@ def setUp(self):
...
+
+    for i, j in zip(result1, result2):
+        if None not in (i, j):  # might be None
+            np.testing.assert_allclose(i.cpu(), j.cpu())
```
YuanTingHsieh, line 163: when I modify this line to
`np.testing.assert_allclose(i.detach().cpu().numpy(), j.detach().cpu().numpy(), atol=1e-5)`
tests are passed for me

YuanTingHsieh: something like this: https://github.com/YuanTingHsieh/MONAI/commit/81a337428878c6114d3698ad0e2d7907187d95e3

rijobro:
Thanks @YuanTingHsieh!

rijobro:
Worryingly, using the GPU at all requires a relative tolerance of > 0.1. Does this imply that if our network is already on the GPU we can't safely save it and get reliable results? Only for certain OS's it seems.

wyli:
> Worryingly, using the GPU at all requires a relative tolerance of > 0.1. Does this imply that if our network is already on the GPU we can't safely save it and get reliable results when we load it back in? Only for certain OS's it seems.

for a workaround for now, shall we roll back the CPU based torchscript tests, and file a new ticket for the GPU based issue?
currently the windows ci is broken we couldn't check any new PR properly

rijobro:
Sounds good to me, i'll update now. Will have to play around with the `rtol` and `atol` quickly to find reasonable values for the CPU.

rijobro:
all done and GPU issue created: #1273.

wyli (approved):

Issue #921, opened 2020-08-19
PersistentDataset errored out


midmelody:
Regular Dataset works fine 
train_dataset = Dataset(train_images, transform=train_transform)
But upon replacing this line with PersistentDataset as
train_dataset = PersistentDataset(train_images, transform=train_transform, cache_dir=persistent_cache)

Received error " Object of type 'ndarray' is not JSON serializable" at getting training sample, details below:

----------------------------------------------------------------------------------------------------------------------------------------
Traceback (most recent call last):
  File "/media/ziyuexu/Research/Code/Pathology/LAMP/Lamp_Pathology_On_The_Fly_PC/train.py", line 277, in <module>
    train(**input_dict)
  File "/media/ziyuexu/Research/Code/Pathology/LAMP/Lamp_Pathology_On_The_Fly_PC/train.py", line 134, in train
    first_sample = first(train_dataloader)
  File "/media/ziyuexu/Research/Tools/MONAI/MONAI/monai/utils/misc.py", line 41, in first
    for i in iterable:
  File "/home/ziyuexu/anaconda3/envs/camelyon/lib/python3.6/site-packages/torch/utils/data/dataloader.py", line 345, in __next__
    data = self._next_data()
  File "/home/ziyuexu/anaconda3/envs/camelyon/lib/python3.6/site-packages/torch/utils/data/dataloader.py", line 856, in _next_data
    return self._process_data(data)
  File "/home/ziyuexu/anaconda3/envs/camelyon/lib/python3.6/site-packages/torch/utils/data/dataloader.py", line 881, in _process_data
    data.reraise()
  File "/home/ziyuexu/anaconda3/envs/camelyon/lib/python3.6/site-packages/torch/_utils.py", line 395, in reraise
    raise self.exc_type(msg)
TypeError: Caught TypeError in DataLoader worker process 0.
Original Traceback (most recent call last):
  File "/home/ziyuexu/anaconda3/envs/camelyon/lib/python3.6/site-packages/torch/utils/data/_utils/worker.py", line 178, in _worker_loop
    data = fetcher.fetch(index)
  File "/home/ziyuexu/anaconda3/envs/camelyon/lib/python3.6/site-packages/torch/utils/data/_utils/fetch.py", line 44, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
[...]

PR #1269, opened 2020-11-19
1190 adds lmdbdataset

### LINKED ISSUE #1190: persistent cache format
the transformed data items are store using `torch.save`,
which is probably not optimized in terms of file size and access speed for the caching purposes

would be great to explore other cache file formats

context:
https://github.com/Project-MONAI/MONAI/blob/bbc2da59c26a5409c1fec219e112a48495633bbc/monai/data/dataset.py#L210-L212

see also https://github.com/Project-MONAI/MONAI/issues/999

wyli:

Fixes #1190

### Description
adds a persistent dataset based on lmdb

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -24,7 +24,7 @@ from typing import TYPE_CHECKING, Any, Callable, List, Optional, Sequence, Union
 import torch
 from torch.utils.data import Dataset as _TorchDataset
 
-from monai.data.utils import default_hashing
+from monai.data.utils import json_hashing
 from monai.transforms import Compose, Randomizable, Transform, apply_transform
 from monai.utils import MAX_SEED, get_seed, min_version, optional_import
 
@@ -114,7 +114,7 @@ class PersistentDataset(Dataset):
         data: Sequence,
         transform: Union[Sequence[Callable], Callable],
         cache_dir: Optional[Union[Path, str]] = None,
-        hash_func: Callable[..., bytes] = default_hashing,
+        hash_func: Callable[..., bytes] = json_hashing,
     ) -> None:
         """
         Args:
@@ -128,7 +128,7 @@ class PersistentDataset(Dataset):
                 may share a common cache dir provided that the transforms pre-processing is consistent.
                 If the cache_dir doesn't exist, will automatically create it.
             hash_func: a callable to compute hash from data items to be cached.
-                defaults to `monai.data.utils.default_hashing`.
+                defaults to `monai.data.utils.json_hashing`.
         """
         if not isinstance(transform, Compose):
             transform = Compose(transform)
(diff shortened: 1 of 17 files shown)


wyli:
for a quick comparison of the performance on the challenge baseline task:

cache file size: persistent `11GB` vs LMDB `9.2GB`
per epoch loading time: persistent `196s` vs LMDB `175s`

monai/data/dataset.py
```diff
@@ -190,35 +199,149 @@ def _pre_first_random_cachecheck(self, item_transformed):
...
+                        if val is None:
+                            val = self._pre_first_random_transform(deepcopy(item))  # keep the original hashed
+                            val = pickle.dumps(val, protocol=self.pickle_protocol)
+                        txn.put(key, val)
+                    done = True
+                except lmdb.MapFullError:
+                    done, retry = False, retry - 1
+                    size = env.info()["map_size"]
+                    new_size = size * 2
+                    print(f"Resizing the cache database from {int(size) >> 20}MB to {int(new_size) >> 20}MB.")
```
ericspod, line 313: This should instead be sent to the log or as a warning, printing things always makes it hard to have silent scripts.

monai/data/dataset.py
```diff
@@ -190,35 +199,149 @@ def _pre_first_random_cachecheck(self, item_transformed):
...
+                    env.set_mapsize(new_size)
+            if not done:  # still has the map full error
+                size = env.info()["map_size"]
+                env.close()
+                raise ValueError(f"Please increase the map size of lmdb (current size={size}).")
```
ericspod, line 318: The exception message should say that the LMDB map size was reached and that it needs to be increased, so something like `f"LMBD map size reached, increase size above current size of {size}"`

monai/data/utils.py
```diff
@@ -762,3 +764,17 @@ def __init__(self, even_divisible: bool = True, *args, **kwargs):
...
+    """
+
+    Args:
+        item: data item to be hashed
+
+    Returns: the corresponding hash key
+
+    """
+    # TODO: Find way to hash transforms content as part of the cache
+    cache_key = hashlib.md5(json.dumps(item, sort_keys=True).encode("utf-8")).hexdigest()
```
ericspod, line 779: Is hashing the pickle not a viable thing? We're relying on json to encode objects properly.

ericspod (approved):
There's a few possible things to change but it's good to go regardless.

monai/data/utils.py
wyli: thanks, I've added the pickle option, it works fine with numpy arrays, so it'll address https://github.com/Project-MONAI/MONAI/issues/921

PR #1283, opened 2020-11-24
1272 update the generalized Wasserstein Dice loss

### LINKED ISSUE #1272: Update the Generalized Wasserstein Dice loss
The implementation of the Generalized Wasserstein Dice loss that has been merged in MONAI does not correspond exactly to the loss proposed in https://link.springer.com/chapter/10.1007/978-3-319-75238-9_6

I will do the appropriate changes in the MONAI implementation and in the unit tests based on:
https://github.com/LucasFidon/GeneralizedWassersteinDiceLoss

I have used the current implementation from my repo in the BraTS challenge 2020, as an additional test, and it performed very well (and better than the implementation currently in MONAI)
https://arxiv.org/abs/2011.01614

LucasFidon:
Fixes #1272.

### Description
I have updated the Generalized Wasserstein Dice loss and I have improved the documentation for this loss function.

### Status
**Ready**

### Types of changes

### DIFF
--- a/docs/source/losses.rst
+++ b/docs/source/losses.rst
@@ -35,6 +35,14 @@ Segmentation Losses
 .. autoclass:: generalized_dice
     :members:
 
+`GeneralizedWassersteinDiceLoss`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. autoclass:: GeneralizedWassersteinDiceLoss
+    :members:
+
+.. autoclass:: generalized_wasserstein_dice
+    :members:
+
 `FocalLoss`
 ~~~~~~~~~~~
 .. autoclass:: FocalLoss
--- a/monai/losses/__init__.py
+++ b/monai/losses/__init__.py
@@ -17,6 +17,7 @@ from .dice import (
     MaskedDiceLoss,
     dice,
     generalized_dice,
+    generalized_wasserstein_dice,
 )
 from .focal_loss import FocalLoss
 from .tversky import TverskyLoss
--- a/monai/losses/dice.py
+++ b/monai/losses/dice.py
@@ -360,12 +360,12 @@ class GeneralizedDiceLoss(_Loss):
 
 class GeneralizedWassersteinDiceLoss(_Loss):
     """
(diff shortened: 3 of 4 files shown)


wyli (approved):
thanks!

wyli:
came across a unit test error:
https://github.com/Project-MONAI/MONAI/runs/1464637050?check_suite_focus=true#step:3:1819

```
FAIL: test_convergence (tests.test_generalized_wasserstein_dice_loss.TestGeneralizedWassersteinDiceLoss)
  File "/opt/monai/tests/test_generalized_wasserstein_dice_loss.py", line 216, in test_convergence
    self.assertGreater(diff_start, diff_end)
AssertionError: tensor(3.6056) not greater than tensor(3.6056)

```

looks like it's not very easy to replicate it, happened only once in 10+ runs

LucasFidon:
Hi @wyli,

Sorry about that. This test passed on my machine.

The test 'test_convergence' is a regression test that checks if the loss decreases when it is used to train a one-layer neural network on a dummy segmentation task (with one image only).
My guess is that the unpredictable error comes from the random initialization of the one-layer neural network.

Setting the random seed may fix the problem (but not sure how robust it will be).
Another option is to remove 'test_convergence'.

I can do the changes if you let me know which approach you think is best.

wyli:
no problem @LucasFidon, I think it's a nice randomised test, hope it's not some instability of the loss function itself, that's why I put the observations here...

(I wonder if it is possible to make a better assertion based on the optimisation theory?)

LucasFidon:
Thanks @wyli ! If you think there might be any way in which the loss is unstable, that would be interesting to look at specific kind of instability/ies in more detail.

Regarding this test and optimization theory, since the loss function has a smooth derivative (wrt to the parameters of the network), in this example, the loss should decrease at each step if the learning rate is sufficiently small.
(See for example Theorem 4.8 in https://arxiv.org/abs/1606.04838)

wyli:
sure @LucasFidon  I've reduced the learning rate to 0.001 and changed the optimiser to adam, it works locally :) I'll post any further issues here, thanks!

PR #1287, opened 2020-11-24
New simplelayer HilbertTransform implemented, and wrapped in new intensity transform DetectEnvelope.

### LINKED ISSUE #1210: Hilbert transform for envelope detection
Many ultrasound computing pipelines rely on a envelope detection detection algorithm. Having such a feature in MONAI would help ultrasound researchers adopt MONAI.

A GPU enabled, ideally differentiable, PyTorch version of [`scipy.signal.hilbert`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.hilbert.html)

Implementing the Hibert transform if/when needed by using the PyTorch [`fft`](https://pytorch.org/docs/master/torch.html#torch.fft).

**Additional context**
Adding @crnbaker in the discussion

crnbaker:
Fixes #1210 

### Description
A new class `HilbertTransform` has been implemented in `monai/networks/layers/simplelayers.py` using the `torch.fft` module. Another new class `DetectEnvelope`, which wraps `HilbertTransform`, has been implemented in  `monai/transforms/intensity/array.py`. Unit testing of `DetectEnvelope` is implemented in `tests/test_detect_envelope.py`. Documentation has been updated and tested locally. Tests have been passed locally. Commits have been squashed.

### Status
**Ready**

### Types of changes

### Details

Running on a Google CoLab VM with a GPU, `HilbertTransform` is faster than the SciPy equivalent for Tensors with more than about 10^4 elements:
image
Notebook demonstrating this is [here](https://colab.research.google.com/drive/1AbYVL7C7muZ5iPwSzI_riQP3ElZoEvcs?usp=sharing).
I also implemented a Hilbert transform using `torch.nn.functional.conv1d` but its performance was poor compared to the FFT implementation.

### DIFF
--- a/docs/source/networks.rst
+++ b/docs/source/networks.rst
@@ -179,6 +179,11 @@ Layers
 .. autoclass:: GaussianFilter
     :members:
 
+`HilbertTransform`
+~~~~~~~~~~~~~~~~~~
+.. autoclass:: HilbertTransform
+    :members:
+
 `Affine Transform`
 ~~~~~~~~~~~~~~~~~~
 .. autoclass:: monai.networks.layers.AffineTransform
--- a/docs/source/transforms.rst
+++ b/docs/source/transforms.rst
@@ -216,6 +216,12 @@ Intensity
     :members:
     :special-members: __call__
 
+`DetectEnvelope`
+"""""""""""""""""""""
+.. autoclass:: DetectEnvelope
+    :members:
+    :special-members: __call__
+
 IO
 ^^
 
--- a/monai/networks/layers/simplelayers.py
+++ b/monai/networks/layers/simplelayers.py
@@ -17,12 +17,15 @@ import torch.nn.functional as F
 from torch import nn
 from torch.autograd import Function
 
+from monai.config import get_torch_version_tuple
 from monai.networks.layers.convutils import gaussian_1d, same_padding
-from monai.utils import SkipMode, ensure_tuple_rep, optional_import
+from monai.utils import InvalidPyTorchVersionError, SkipMode, ensure_tuple_rep, optional_import
 
 _C, _ = optional_import("monai._C")
+if tuple(int(s) for s in torch.__version__.split(".")[0:2]) >= (1, 7):
+    fft, _ = optional_import("torch.fft")
 
-__all__ = ["SkipConnection", "Flatten", "GaussianFilter", "LLTM", "Reshape", "separable_filtering"]
+__all__ = ["SkipConnection", "Flatten", "GaussianFilter", "LLTM", "Reshape", "separable_filtering", "HilbertTransform"]
 
 
 class SkipConnection(nn.Module):
@@ -128,6 +131,72 @@ def separable_filtering(x: torch.Tensor, kernels: Union[Sequence[torch.Tensor], 
         return conv_type(input=_conv(input_, d - 1), weight=_kernel, padding=_padding, groups=n_chns)
 
     return _conv(x, spatial_dims - 1)
+
+
+class HilbertTransform(nn.Module):
+    """
+    Determine the analytical signal of a Tensor along a particular axis.
+    Requires PyTorch 1.7.0+ and the PyTorch FFT module (which is not included in NVIDIA PyTorch Release 20.10).
+
+    Args:
+        axis: Axis along which to apply Hilbert transform. Default 2 (first spatial dimension).
+        N: Number of Fourier components (i.e. FFT size). Default: ``x.shape[axis]``.
+    """
+
+    def __init__(self, axis: int = 2, n: Union[int, None] = None) -> None:
+
+        if get_torch_version_tuple() < (1, 7):
+            raise InvalidPyTorchVersionError("1.7.0", self.__class__.__name__)
+
+        super().__init__()
+        self.axis = axis
+        self.n = n
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """
+        Args:
+            x: Tensor or array-like to transform. Must be real and in shape ``[Batch, chns, spatial1, spatial2, ...]``.
+        Returns:
+            torch.Tensor: Analytical signal of ``x``, transformed along axis specified in ``self.axis`` using
+            FFT of size ``self.N``. The absolute value of ``x_ht`` relates to the envelope of ``x`` along axis ``self.axis``.
+        """
+
+        # Make input a real tensor
+        x = torch.as_tensor(x, device=x.device if torch.is_tensor(x) else None)
+        if torch.is_complex(x):
+            raise ValueError("x must be real.")
+        else:
+            x = x.to(dtype=torch.float)
+
+        if (self.axis < 0) or (self.axis > len(x.shape) - 1):
+            raise ValueError("Invalid axis for shape of x.")
+
+        n = x.shape[self.axis] if self.n is None else self.n
+        if n <= 0:
+            raise ValueError("N must be positive.")
+        x = torch.as_tensor(x, dtype=torch.complex64)
+        # Create frequency axis
+        f = torch.cat(
+            [
+                torch.true_divide(torch.arange(0, (n - 1) // 2 + 1, device=x.device), float(n)),
+                torch.true_divide(torch.arange(-(n // 2), 0, device=x.device), float(n)),
+            ]
+        )
+        xf = fft.fft(x, n=n, dim=self.axis)
+        # Create step function
+        u = torch.heaviside(f, torch.tensor([0.5], device=f.device))
+        u = torch.as_tensor(u, dtype=x.dtype, device=u.device)
+        new_dims_before = self.axis
+        new_dims_after = len(xf.shape) - self.axis - 1
+        for _ in range(new_dims_before):
+            u.unsqueeze_(0)
+        for _ in range(new_dims_after):
+            u.unsqueeze_(-1)
+
+        ht = fft.ifft(xf * 2 * u, dim=self.axis)
+
+        # Apply transform
+        return torch.as_tensor(ht, device=ht.device, dtype=ht.dtype)
 
 
 class GaussianFilter(nn.Module):
--- a/monai/transforms/intensity/array.py
+++ b/monai/transforms/intensity/array.py
@@ -20,10 +20,11 @@ from warnings import warn
 import numpy as np
 import torch
 
-from monai.networks.layers import GaussianFilter
+from monai.config import get_torch_version_tuple
+from monai.networks.layers import GaussianFilter, HilbertTransform
 from monai.transforms.compose import Randomizable, Transform
 from monai.transforms.utils import rescale_array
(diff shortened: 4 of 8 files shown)


wyli:
looks like torch.fft has some breaking changes in torch1.5 - 1.7, perhaps you can use this pattern as a workaround: https://github.com/Project-MONAI/MONAI/blob/9f51893d162e5650f007dff8e0bcc09f0d9a6680/monai/transforms/spatial/array.py#L51-L58

crnbaker:
> looks like torch.fft has some breaking changes in torch1.5 - 1.7, perhaps you can use this pattern as a workaround:
> 
> https://github.com/Project-MONAI/MONAI/blob/9f51893d162e5650f007dff8e0bcc09f0d9a6680/monai/transforms/spatial/array.py#L51-L58

Thanks @wyli, I'll try that. There are a couple of mypy issues to fix too. Once all the tests (apart from DCO) have passed I'll rebase to a single signed commit.

tvercaut:
@crnbaker Nice graphs! From a quick look at it, the 10^4 performance crossing point refers to the number of pixels across the full image / tensor, rather than the size along the transformed dimension, right?

Looking at the performance crossing point in terms of number of points within an A-line, does the number of lines one has in the image/tensor influence it much?

If the crossing point is always around 200 samples / A-line, I guess it means the GPU implementation should be the best performing one in most real-life applications. Correct?

crnbaker:
Yes, the x axis is number of pixels in the whole tensor. That dataset contained images consisting of 50 A-lines, each image with different numbers of samples per line.

It doesn't appear that the shape of the image affects the results. In the plot below, that "2D" data is the same as the data above, and the "1D" data consists of Tensors containing only a single A-line (the number of samples in each line being higher so that the total size of the Tensors is similar to those in the 2D dataset):
image

I was expecting that the PyTorch FFT module would be doing doing parallelisation under the hood, so that multiple FFTs would be calculated simultaneously. That doesn't appear to be the case, but maybe this can be achieved using batching. Or PyTorch is parallelizing the FFT algorithm itself, which I think is possible but unlikely.

Here's a graph of the (1D) SciPy times divided by the (1D) MONAI GPU times:
image

Our FOH images have approximately 380,000 elements, so we should expect the MONAI Hilbert transform to run approximately 40 times faster than the ScipPy one (at least for the Google CoLab CPU/GPU).

crnbaker:
@wyli I've used the `torch.fft.fft` module which is new in PyTorch 1.7. Pre v1.7 there was the `torch.fft` function, which will be removed in v1.8. I can relatively easily support <1.7 by importing the `torch.fft` function and using that instead, but this would require me to import `get_torch_version_tuple` before the `torch` modules, which would fail Flake8! Any ideas?

wyli:
I think you can declare the type this one https://github.com/Project-MONAI/MONAI/blob/9f51893d162e5650f007dff8e0bcc09f0d9a6680/monai/transforms/spatial/array.py#L49 

you could also put `# pytype: disable=` or `# type: ignore` or `# noqa` to mute the errors for now

crnbaker:
I've created two new functions `fft()` and `ifft()` in `networks/utils` that wrap either `torch.fft()` and `torch.ifft()` or `torch.fft.fft()` and `torch.fft.ifft()` depending on the installed version of PyTorch.

Weirdly, although I'm passing the PT16+CUDA102 and PT17+CUDA102 checks, it's failing the PT17+CUDA110 check because it can't find the `torch.fft` module. It works fine on my PC with `1.7.0+cu110`.

The other failed check is quick-py3 (windows-latest), where something seems to be going wrong in `tests/utils` which isn't a file that I've modified.

I'm not sure where to go from here so any help would be appreciated!

wyli:
the current PR works fine with torch1.6+cu10.2, torch1.7+cu11.1 only if I change the version conditions in the utils to 
```py
if tuple(int(s) for s in torch.__version__.split(".")[0:2]) <= (1, 7):  # line 22
if get_torch_version_tuple() > (1, 7):   # line 248
```

crnbaker:
Is it possible to get the Dockerfiles to allow me set up the containers that are failing so I can debug on my system?

wyli:
that's not yet implemented, for now perhaps you could refer to the configuration which are essentially bash commands https://github.com/Project-MONAI/MONAI/blob/fd24c44c65a2932f13f514ab53eef176d42d3915/.github/workflows/pythonapp.yml#L149 btw the fft API of pytorch is a mess, and apparently the documentation is not quite correct ...

crnbaker:
Thanks. I'm starting to realise that! Also, complex datatypes have only been recently introduced.

crnbaker:
I had another look at this at the end of last week. To debug, I need a system with PyTorch 1.6 and CUDA 11.0. There are no windows binaries available for this combination via pip or conda. There is a Docker image available, but Docker on Windows doesn't support CUDA. So I need a Linux system. I'd use WSL but that doesn't work with CUDA without installing a preview release on Windows... so I think I need to set up dual boot which I'd like to do anyway.

Or as @tvercaut suggested, could we make this feature only available on PyTorch 1.7 somehow?

wyli:
I think we should go for pytorch 1.7+ , in the unit tests we need a new decorator similar to this one, to skip the tests on torch < 1.7  https://github.com/Project-MONAI/MONAI/blob/f01994d47403f920b5a51cfeae75418383d10178/tests/utils.py#L69

crnbaker:
Great, thanks @wyli I'm nearly there with this now. Any suggestions on what exception should be raised if the function is called with PyTorch < 1.7.0? I could defined a new exception, something like:

```
class InvalidPyTorchVersion(Error):
    """Raised when called function or method requires a more recent PyTorch version than that installed."""
    def __init__(self, required_version, name):
        message = f'{name} requires PyTorch version {required_version} or later'
        super().__init__(message)

class HilbertTransform(nn.Module):
    def __init__(self):
        if get_torch_version_tuple() < (1, 7):
            raise InvalidPyTorchVersion('1.7.0', self.__class__.__name__)

```

Is there anywhere in particular this should go?

wyli:
cool, the exception could be in the module util, next to this one: https://github.com/Project-MONAI/MONAI/blob/f01994d47403f920b5a51cfeae75418383d10178/monai/utils/module.py#L108

or perhaps reuse this example: https://github.com/Project-MONAI/MONAI/blob/f01994d47403f920b5a51cfeae75418383d10178/monai/utils/module.py#L142

I'm fine with either

crnbaker:
It turns out the PT17+CU110 check uses an [NVIDIA container image](https://docs.nvidia.com/deeplearning/frameworks/pytorch-release-notes/rel_20-10.html) which is based on a pre-release version of PyTorch 1.7 ([commit 7036e91](https://github.com/pytorch/pytorch/commit/7036e91) from 15th September) that doesn't yet have the FFT module. Apparently the FFT module was added sometime between then and the [1.7.0 release](https://github.com/pytorch/pytorch/releases/tag/v1.7.0) on 27th October.

I suppose others may also be using this NVIDIA image, so I'll need to add in something that not only checks the PyTorch version, but also whether the FFT module is available.

crnbaker:
All sorted. Just failing the flake8 check for some reason... seems to be raising an exception related to the multiprocessing module. Have you seen that before @wyli?

wyli:
> All sorted. Just failing the flake8 check for some reason... seems to be raising an exception related to the multiprocessing module. Have you seen that before @wyli?

interesting, I've never seen this before I'm looking into this...

wyli:
not sure if this is relevant but I got these errors on mac, perhaps remove the shebang line in the file?
```log
flake8
3.8.4 (flake8-bugbear: 20.1.4, flake8-comprehensions: 3.3.0,
flake8-executable: 2.0.4, flake8-pyi: 20.10.0, mccabe: 0.6.1, naming: 0.11.1,
pycodestyle: 2.6.0, pyflakes: 2.2.0) CPython 3.6.10 on Darwin
/Users/wenqili/Documents/MONAI/tests/test_hilbert_transform.py:1:1: EXE001 Shebang is present but the file is not executable.
/Users/wenqili/Documents/MONAI/tests/test_hilbert_transform.py:32:6: N806 variable 'X' in function should be lowercase
/Users/wenqili/Documents/MONAI/tests/test_hilbert_transform.py:37:6: N806 variable 'U' in function should be lowercase
/Users/wenqili/Documents/MONAI/tests/test_hilbert_transform.py:41:10: N806 variable 'U' in function should be lowercase
/Users/wenqili/Documents/MONAI/tests/test_hilbert_transform.py:43:10: N806 variable 'U' in function should be lowercase
1     EXE001 Shebang is present but the file is not executable.
4     N806 variable 'X' in function should be lowercase
5
Check failed!
```

wyli:
I think it's a bug from the flake8 plugin https://github.com/xuhdev/flake8-executable/issues/15  need to make the py files not executable`chmod -x`

crnbaker:
Good to go!

wyli (approved):
thank you!

wyli:
> It turns out the PT17+CU110 check uses an [NVIDIA container image](https://docs.nvidia.com/deeplearning/frameworks/pytorch-release-notes/rel_20-10.html) which is based on a pre-release version of PyTorch 1.7 ([commit 7036e91](https://github.com/pytorch/pytorch/commit/7036e91) from 15th September) that doesn't yet have the FFT module. Apparently the FFT module was added sometime between then and the [1.7.0 release](https://github.com/pytorch/pytorch/releases/tag/v1.7.0) on 27th October.
> 
> I suppose others may also be using this NVIDIA image, so I'll need to add in something that not only checks the PyTorch version, but also whether the FFT module is available.
> 
> EDIT: I'll use `optional_import()` and `@SkipIfNoModule`

I didn't aware of this issue earlier, monai's docker image also uses PyTorch Version 1.7.0a0+7036e91...

crnbaker:
> > It turns out the PT17+CU110 check uses an [NVIDIA container image](https://docs.nvidia.com/deeplearning/frameworks/pytorch-release-notes/rel_20-10.html) which is based on a pre-release version of PyTorch 1.7 ([commit 7036e91](https://github.com/pytorch/pytorch/commit/7036e91) from 15th September) that doesn't yet have the FFT module. Apparently the FFT module was added sometime between then and the [1.7.0 release](https://github.com/pytorch/pytorch/releases/tag/v1.7.0) on 27th October.
> > I suppose others may also be using this NVIDIA image, so I'll need to add in something that not only checks the PyTorch version, but also whether the FFT module is available.
> > EDIT: I'll use `optional_import()` and `@SkipIfNoModule`
> 
> I didn't aware of this issue earlier, monai's docker image also uses PyTorch Version 1.7.0a0+7036e91...

Yes and I've just seen it's failed the Docker check, which seems bizarre considering it passed with the same version of PyTorch here... I'll look into it.

wyli:
no problem I pushed a quick fix https://github.com/Project-MONAI/MONAI/blob/5c75dcb979108657421b86bede1d03e98af82bcf/monai/utils/module.py#L281-L287

PR #1291, opened 2020-11-25
1254 Support to cache N transforms in PersistentDataset

### LINKED ISSUE #1254: persistent dataset to cache first N transforms
currently the persistent dataset caches the outcome up to the first randomised transform,
would be great to have a more flexible option to cache the first N transforms.

this is useful when we want to fix some of the non-randomised transforms and tune the hyperparameters of the others.

Nic-Ma:
Fixes #1254 .

### Description
This PR added support to cache the result of first N transforms in `PersistentDataset` and added unit tests.

### Status
**Ready**

### Types of changes


wyli:
thanks, could you create a new class by inheriting persistentdataset to implement this new feature? the current implementation makes both persistent and lmdb dataset too complicated. sorry I didn't make this clear in the feature request

Nic-Ma:
Hi @wyli ,

But is it useful for `LmdbDataset`? I think it may be better to put in the `PersistentDataset` class as all kinds of caching with "**persistent**" storage can leverage this feature to tune training, so we don't need to implement for other subclass of `PersistentDataset`.
What do you think?

Thanks.

wyli:
the new feature is simple but it's "incorrect" because it ignores the random transforms, I don't think it should be widely used

Nic-Ma:
Hi @wyli ,

OK, I will update to make it into a subclass, if there is a feature request for the LmdbDataset or other subclass of `PersistentDataset`, I think we can refactor to make it in the base class directly.

Thanks.

Nic-Ma:
Hi @wyli ,

Thanks.

wyli (approved):
thanks

PR #1292, opened 2020-11-25
1145 add utility function to set different LR for layers

### LINKED ISSUE #1145: Freeze layers for transfer learning
Is it possible to freeze layers in the UNet?
I want to freeze some layers in order to do transfer learning.

However, when I use model.children, the whole network is printed. I am not able to get specific layers.

So, is this possible, or do you recommand something else for transfer learning?

Thank you. 

Best,
Kirsten

Nic-Ma:
Fixes #1145 .

### Description
This PR implemented a utility function to generate parameter groups with specified LRs.
It can be used to set different LR values for different layers of the network.

### Status
**Ready**

### Types of changes


monai/optimizers/utils.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Utility function to generate parameter groups with different LR values for optimizer.
+
+    Args:
+        layer_matches: a list of callable functions to select network layer groups, input will be the network.
+        lr_values: a list of LR values corresponding to the `layer_matches` functions.
+
+    It's mainly used to set different init LR values for different network elements, for example::
+
+        net = Unet(dimensions=3, in_channels=1, out_channels=3).to(device)
```
wyli, line 27: the example doesn't work
```suggestion
        net = Unet(dimensions=3, in_channels=1, out_channels=3, channels=[2, 2, 2], strides=[1, 1, 1])
```

monai/optimizers/utils.py
```diff
@@ -0,0 +1,44 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Callable, Sequence
+
+import torch
+
+
+def generate_params(network: torch.nn.Module, layer_matches: Sequence[Callable], lr_values: Sequence[float]):
```
wyli, line 17: perhaps the name could be `parameter_groups` or `per_parameter_option`

monai/optimizers/utils.py
```diff
@@ -0,0 +1,44 @@
...
+    if len(layer_matches) != len(lr_values):
+        raise ValueError("length of layer_match callable functions and LR values should be the same.")
+
+    params = list()
+    _layers = list()
+    for func, lr in zip(layer_matches, lr_values):
+        layer_params = func(network).parameters()
+        params.append({"params": layer_params, "lr": lr})
+        _layers.extend(list(map(id, layer_params)))
+    params.append({"params": filter(lambda p: id(p) not in _layers, network.parameters())})
```
wyli, line 42: would be great to have line 42 optional, something like:
```
if include_others:
    params.append({"params": filter(lambda p: id(p) not in _layers, network.parameters())})
```

also, will this work when model has a nested structure, for example `layer_matches=[lambda x: x.model[0][1][0]]`?

monai/optimizers/utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+    It's mainly used to set different init LR values for different network elements, for example::
+
+        net = Unet(dimensions=3, in_channels=1, out_channels=3).to(device)
+        print(net)  # print out network components to select expected items
+        params = generate_params(net, [lambda x: x.model[-1], lambda x: x.model[-2]], [1e-2, 1e-3])
+        optimizer = torch.optim.Adam(params, 1e-4)
+
+    """
+    if len(layer_matches) != len(lr_values):
```
wyli, line 33: could have `ensure_tuple` for both `layer_matches` and `lr_values`?

wyli (commented):
thanks, this is useful for setting layer wise learning rate, please see some comments inline

monai/optimizers/utils.py
Nic-Ma: updated in the latest commit, thanks.

monai/optimizers/utils.py
Nic-Ma: Sure, updated function name to `generate_param_groups` in the latest commit.
Thanks.

monai/optimizers/utils.py
Nic-Ma: Add `include_others` in the latest commit.
It should support nested network, you can take a look at the `TEST_CASE_3` of unit tests.

Thanks.

monai/optimizers/utils.py
Nic-Ma: Sure, added `ensure_tuple` for both in the latest commit.

Thanks.

Nic-Ma:
Hi @wyli ,

Thanks for your review.

Thanks.

monai/optimizers/utils.py
```diff
@@ -0,0 +1,56 @@
...
+    network: torch.nn.Module,
+    layer_matches: Sequence[Callable],
+    lr_values: Sequence[float],
+    include_others: bool = True,
+):
+    """
+    Utility function to generate parameter groups with different LR values for optimizer.
+
+    Args:
+        layer_matches: a list of callable functions to select network layer groups, input will be the network.
```
wyli, line 29: missing the docstring for `network`

monai/optimizers/utils.py
```diff
@@ -0,0 +1,56 @@
...
+        lr_values: a list of LR values corresponding to the `layer_matches` functions.
+        include_others: whether to incude the rest layers as the last group, default to True.
+
+    It's mainly used to set different init LR values for different network elements, for example::
+
+        net = Unet(dimensions=3, in_channels=1, out_channels=3, channels=[2, 2, 2], strides=[1, 1, 1])
+        print(net)  # print out network components to select expected items
+        params = generate_params(net, [lambda x: x.model[-1], lambda x: x.model[-2]], [1e-2, 1e-3])
```
wyli, line 37:
```suggestion
        params = generate_param_groups(net, [lambda x: x.model[-1], lambda x: x.model[-2]], [1e-2, 1e-3])
```

wyli (commented):
thanks, looks good now, some minor issues inline. 

how can I use it to freeze all the convolution weights given a network? I think it's very useful to have this kind of filtering:
```python
net = Unet(dimensions=3, in_channels=1, ...
for n, x in net.named_parameters(): print(n if 'conv.weight' in n else '')
```

so could you add this option? something like:
```py
generate_param_groups(net, name_filter=lambda x: "weight" in x, lr_value=0, include_others=True)
```

monai/optimizers/utils.py
Nic-Ma: Added in the latest commit.
Thanks.

monai/optimizers/utils.py
Nic-Ma: updated in the latest commit.
Thanks.

Nic-Ma:
Hi @wyli ,

Thanks for your review.
I added support for name filter. The output param groups have the same order as `layer_matches` functions, then user can adjust LR values for different param groups according to the order index.

Thanks.

wyli (approved):
thanks, this looks nice!

PR #1293, opened 2020-11-25
Add rank to the engine.state

Nic-Ma:
### Description
This PR is a non-breaking change to our workflows, added `rank` to the `engine.state`, then we don't need to call `dist.get_rank()` repeatedly in distributed data-parallel programs based on workflows.

### Status
**Ready**

### Types of changes


wyli:
hi @vfdev-5, just wanted to check this with you, is this the expected way for an ignite user to resolve the rank information?

vfdev-5:
Hi @wyli , since v0.4.x we start to provide an abstraction on various [distributed frameworks/backends (nccl/gloo for gpu/cpu via native torch, tpu via torch_xla, horovod)](https://pytorch.org/ignite/distributed.html), so user can do the following:
```python
import ignite.distributed as idist

idist.get_rank()
```
which is dispatched to the current backend and its function like `dist.get_rank()`.
So, if user would like to use `ignite.distributed`, we recommend to do `idist.get_rank()`. In other cases, yes, it is possible to put all those distributed communication info into `engine.state`. Hope this could answer your question.

Nic-Ma:
Hi @wyli ,

Do you still have any concern about this PR?
Thanks.

wyli (approved):
thanks, I'm not sure what the use case is but I guess it's ok for now according to the comments

PR #1296, opened 2020-11-26
Add test for printing debug info

rijobro:
### Description
Sorry, it seems I forgot to add the unit test for printing debug info.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/utils/module.py
+++ b/monai/utils/module.py
@@ -236,13 +236,19 @@ def has_option(obj, keywords: Union[str, Sequence[str]]) -> bool:
 
 def get_package_version(dep_name, default="NOT INSTALLED or UNKNOWN VERSION."):
     """
-    Try to load package and get version. Unload once finished. If not found, return `default`.
+    Try to load package and get version. If not found, return `default`.
+
+    If the package was already loaded, leave it. If wasn't previously loaded, unload it.
     """
     dep_ver = default
+    dep_already_loaded = dep_name not in sys.modules
+
     dep, has_dep = optional_import(dep_name)
     if has_dep:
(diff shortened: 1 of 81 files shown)


wyli:
I actually forgot to mention that please add `psutil ` to this doc https://github.com/Project-MONAI/MONAI/blob/master/docs/source/installation.md#installing-the-recommended-dependencies (similar to `tqdm` and `lmdb`)

rijobro:
@wyli I've also updated the bug report to use the debug output, so I'll also need to print OS, python version and CUDA/cuDNN version. 

As such, please don't merge this straight away.

wyli:
sure there's a 'convert to draft' button on this PR page

rijobro:
@wyli thanks, done now.

monai/utils/module.py
```diff
@@ -231,3 +232,17 @@ def has_option(obj, keywords: Union[str, Sequence[str]]) -> bool:
...
+    """
+    Try to load package and get version. Unload once finished. If not found, return `default`.
+    """
+    dep_ver = default
+    dep, has_dep = optional_import(dep_name)
+    if has_dep:
+        if hasattr(dep, "__version__"):
+            dep_ver = dep.__version__
+        del dep
+        del sys.modules[dep_name]
```
wyli, line 247: if `dep_name` is already imported before `get_package_version`, this might cause some surprises...

rijobro: good point, I added in a check to only unload if not already loaded. And added a test to that effect.

wyli (approved):
thanks, passed all tests and also locally verified

PR #1307, opened 2020-12-01
909 Update LoadImage to use Nibabel as default

### LINKED ISSUE #909: image data reader and LoadImage transform refactoring
This is a follow-up of #856 to track the progress of a preliminary IO refactoring.

**Goal**

The main goal of this ticket is to create a file format agnostic wrapper for the image inputs (`LoadImage` and `LoadImaged`).
A typical use case (enhancing the current [`NiftiDataset`](https://github.com/Project-MONAI/MONAI/blob/6a6f52f5f80e5c706f1f0976aa76b56af488d119/monai/data/nifti_reader.py#L21)) is:
```python
train_ds = ImageDataset(images, segs, transform=train_imtrans, seg_transform=train_segtrans)
```
where `images` and `segs` are filenames, `ImageDataset` internally invokes `LoadImage`.

Or using a generic monai Dataset API, specify the first component of `train_imtrans` as a `LoadImage` transform that extracts data array and meta data from the file.

More specifically:

**Reader** （file format-specific component)

1. `[format]reader` are themselves monai data components (users can use the readers without going through the `LoadImage` logic), they call external package APIs
2. `[format]reader` share the same superclass `ImageReader` interface (for handling physical/voxel resolution, orientation...)
```py
class ImageReader:
    def get_data(self, ...):
        # return the data in the designated resolution, orientation, 
        # image channel selector, the relevant coordinates, datatype
    
    def read(self, ...):
        # create an IO object instance or a set of instances
```
3. users can make customized `MyReader` by implementing the `ImageReader` interface
4. the customized `MyReader` could be registered with the `LoadImage` factory
```python
@register_reader(requires='my_external_library')
class MyReader(ImageReader):
    ...
```

**LoadImage**（file format agnostic component)

1. `LoadImage` is a monai transform
2. `LoadImaged` should return the same underlying data representation dictionary regardless of the image format.
3. `LoadImage` may choose a suitable `[format]reader` dynamically at runtime based on the following rules (ordered from high to low priority):
  - user's selection of a specific reader, such as `LoadImage(reader='ITkReader')`
  - try any user-registered readers
  - try resources map based on a global property file and filename extensions:
  ```
  MONAI.IO
  {
      FOO LoadFOO
      PNG  ITK/5.1/LoadITK
      TIFF,TIF  ITK/5.1/LoadITK
      * ITK/5.1/LoadITK
  }
  ```
  - fails early and preserves detailed error messages

**additional requirements**
- should be able to specify versions of external reader to ensure reproducibility
- need to have consistent coordinate systems, resolution conventions

**tasks**
- [x] implement interface `ImageReader`, factory (transforms) `LoadImage` `LoadImaged`
- [x] implement ITKReader, NibabelReader, NumpyReader, PILReader?
- [x] `LoadImage` selecting readers according to the preference list
- [x] new interfaces tutorial https://github.com/Project-MONAI/Tutorials/pull/19
- [x] deprecate `NiftiDataset`, add `ImageDataset`
- [x] deprecate `LoadNifti`, `LoadPNG`, `LoadNumpy`, `LoadDatad`
- [x] revise/check all examples demos

Nic-Ma:
Partly fixes issue #909  .

### Description
This PR implemented below features:
Change the default reader of LoadImage to Nibabel.
Add string support for the reader parameter.
Update all the examples and tutorials to use LoadImage transform.

### Status
**Ready**

### Types of changes


monai/transforms/io/array.py
```diff
@@ -31,38 +32,47 @@
...
             The transform returns image data array if `image_only` is True,
             or a tuple of two elements containing the data array, and the meta data in a dict format otherwise.
 
         """
-        self.default_reader: ITKReader = ITKReader()
+        self.default_reader: NibabelReader = NibabelReader()
         self.readers: List[ImageReader] = list()
         if reader is not None:
+            if isinstance(reader, str):
+                reader = eval(reader + "(*args, **kwargs)")
```
wyli, line 71: this will introduce some security issues, please change self.readers to a dictionary: `self.readers: Dict[str, ImageReader] = {}`, then we can create a reader instance with:
```py
ReaderClass = self.readers[reader]
reader = ReaderClass(*args, **kwargs)
```

Nic-Ma: Sure, updated in the latest commit.
Thanks.

Nic-Ma:
Hi @wyli ,

Thanks for your review, I updated the PR according to your comments.
Could you please help review again?
We will try to update all the examples and tutorials when this PR is ready.

Thanks.

monai/transforms/io/array.py
```diff
@@ -31,38 +32,53 @@
...
             The transform returns image data array if `image_only` is True,
             or a tuple of two elements containing the data array, and the meta data in a dict format otherwise.
 
         """
-        self.default_reader: ITKReader = ITKReader()
+        self.default_reader: NibabelReader = NibabelReader()
         self.readers: List[ImageReader] = list()
         if reader is not None:
+            if isinstance(reader, str):
+                supported_readers = {
```
wyli, line 71: could be combined with `self.readers`?

Nic-Ma: Hi @wyli ,

`self.readers` is used to record all the readers in the `LoadImage` transform, we can register readers to the list and the **order of the list** maps to the reader priorities.
And you can set 2 `ITKReader` with different parameters.

Thanks.

wyli: ok, I think by default if `LoadImage(reader=None)` it should automatically try a set of different reader classes, that was our original requirement.

Nic-Ma: We don't have this logic so far. Acutally, it's not very easy to achieve this requirement, we need to define the default reader priorities list for every kind of image format.
Of course, it sounds like a good idea, now we only use `NibabelReader` as default reader, we can try to enhance it later.

Thanks.

wyli: the idea of feature #909 is to create a file format agnostic wrapper. as a starting point, could we support the logic of: 

```
if nifti file try nibabel
elif jpg/png try pillow
elif npy try loadnumpy
else try itk
```
if you don't have time to implement it I think we should postpone this feature

Nic-Ma: OK, I will try to make the default reader automatically select reader type.
Thanks.

Nic-Ma:
Hi @wyli ,

I updated the PR to automatically select the default reader based on image type.

Thanks.

monai/transforms/io/array.py
```diff
@@ -31,49 +32,68 @@
...
 
         Note:
             The transform returns image data array if `image_only` is True,
             or a tuple of two elements containing the data array, and the meta data in a dict format otherwise.
 
         """
-        self.default_reader: ITKReader = ITKReader()
-        self.readers: List[ImageReader] = list()
+        # set predefined readers as default
+        self.readers: List[ImageReader] = [ITKReader(), NumpyReader(), PILReader(), NibabelReader()]
```
wyli, line 69: i think the default readers should also support `args` and `kwargs`

monai/transforms/io/array.py
```diff
@@ -31,49 +32,68 @@
...
-        Register image reader to load image file and meta data.
+        Register image reader to load image file and meta data, latest registered reader has higher priority.
         Return all the registered image readers.
 
         Args:
             reader: registered reader to load image file and meta data based on suffix,
-                if all registered readers can't match suffix at runtime, use the default ITK reader.
+                if all registered readers can't match suffix at runtime, use the default readers.
 
         """
```
wyli, line 98:
```suggestion
        """
        if not isinstance(reader, ImageReader):
                raise ValueError(f"reader must be an ImageReader instance, but got {type(reader)}.")
```

monai/transforms/io/array.py
```diff
@@ -31,49 +32,68 @@
...
+                    "ITKReader": ITKReader,
+                    "NumpyReader": NumpyReader,
+                }
+                if reader not in supported_readers:
+                    raise ValueError(f"unsupported reader type: {reader}.")
+                reader = supported_readers[reader](*args, **kwargs)
+
+            if not isinstance(reader, ImageReader):
+                raise ValueError(f"reader must be ImageReader object, but got {type(reader)}.")
             self.readers.append(reader)
```
wyli, line 84:
```suggestion
            self.register(reader)
```

monai/transforms/io/array.py
```diff
@@ -93,13 +113,15 @@ def __call__(
 
         """
         if reader is None or not reader.verify_suffix(filename):
-            reader = self.default_reader
             if len(self.readers) > 0:
```
wyli, line 116:
```suggestion
```

monai/transforms/io/array.py
```diff
@@ -93,13 +113,15 @@ def __call__(
...
-                for r in self.readers:
+                for r in reversed(self.readers):
                     if r.verify_suffix(filename):
                         reader = r
                         break
```
wyli, line 120:
```suggestion
            for r in reversed(self.readers):
                if r.verify_suffix(filename):
                    reader = r
                    break
```

wyli (commented):
thanks I put some minor suggestions of using self.register

monai/transforms/io/array.py
Nic-Ma: updated in the latest commit.
Thanks.

monai/transforms/io/array.py
Nic-Ma: updated in the latest commit.
Thanks.

monai/transforms/io/array.py
Nic-Ma: updated in the latest commit.
Thanks.

monai/transforms/io/array.py
Nic-Ma: updated in the latest commit.
Thanks.

Nic-Ma:
Hi @wyli .

Thanks for your suggestions, I updated the PR according to your comments.

Thanks.

wyli (approved):
thanks, all tests upgraded to use the new API and tests work fine

Nic-Ma:
Thanks for your update, I will try to update examples and tutorials soon.

PR #1309, opened 2020-12-01
974 Add IterableDataset and unit tests

### LINKED ISSUE #974: Support of torch.utils.data.IterableDataset (and WebDataset?)
`monai.data.Dataset` mainly supports `torch.utils.data.Dataset` with the `__len__` and `__getitem__` APIs.

would be great to introduce compatible APIs based on `torch.utils.data.IterableDataset`,
the intention is to support the WebDataset https://github.com/pytorch/pytorch/issues/38419 in the future, for efficient I/O of very large datasets

see also https://github.com/tmbdev/webdataset

Nic-Ma:
Fixes #974 .

### Description
This PR added the `IterableDataset` to support loading data from stream.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @wyli and @ericspod ,

Have you guys seen this CI errors before? I didn't figure out why the quick test failed on `mac` and `windows`..
Thanks in advance.

wyli:
I think you could try make `_Stream` a global object, pickle doesn't work well with the local object. also I'm not sure whether windows will work with the multiprocess file accessing

ericspod:
The way that the Dataloader works in Windows is to spawn processes and share the state of the dataset (which includes transforms and other things) through pickling since fork isn't present. This is what the pickle complaint is about, if that object is global then it will be created new for each process when monai is imported.

Nic-Ma:
Cool, thanks to you 2 python experts, I will have a try ASAP.

Nic-Ma:
Hi @wyli and @ericspod ,

Seems it works, thanks for your help!
Could you please help review it now?

Thanks.

monai/data/dataset.py
```diff
@@ -71,6 +72,34 @@ def __getitem__(self, index: int):
...
+    A generic dataset for iterable data source and an optional callable data transform
+    when fetching a data sample.
+    For example, typical input data can be web data stream which can support multi-process access.
+
+    Note that when used with `DataLoader` and `num_workers > 0`, each worker process will have a
+    different copy of the dataset object, need to guarantee process-safe from data source or DataLoader.
+
+    """
+
+    def __init__(self, data: Sequence, transform: Optional[Callable] = None) -> None:
```
wyli, line 86:
```suggestion
    def __init__(self, data: Iterable, transform: Optional[Callable] = None) -> None:
```

monai/data/dataset.py
```diff
@@ -71,6 +72,34 @@ def __getitem__(self, index: int):
...
+        """
+        Args:
+            data: input data source to load and transform to generate dataset for model.
+            transform: a callable data transform on input data.
+        """
+        self.data = iter(data)
```
wyli, line 92: I think this could be `self.data = data`, and then we can get the data  iterator  in the `__iter__`

monai/data/dataset.py
```diff
@@ -71,6 +72,34 @@ def __getitem__(self, index: int):
...
+        self.transform = transform
+
+    def __iter__(self):
```
wyli, line 95:
```suggestion
    def __iter__(self):
        data_iter = iter(data)
```

the webdataset library stores the iterator as `self.source` as well: https://github.com/tmbdev/webdataset/blob/master/webdataset/dataset.py#L498

maybe it's needed,I don't have a use case at hand

wyli (commented):
thanks, perhaps we should start a new py file `iterable_dataset`, the current dataset has 800+ lines already

monai/data/dataset.py
Nic-Ma: updated in the latest commit.
Thanks.

monai/data/dataset.py
Nic-Ma: updated in the latest commit.
Thanks.

monai/data/dataset.py
Nic-Ma: updated in the latest commit.
We can adjust more when more use cases come.

Thanks.

Nic-Ma:
Hi @wyli ,

Thanks for your review. I updated the PR according to your comments.

Thanks.

monai/data/iterable_dataset.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    def __init__(self, data: Iterable, transform: Optional[Callable] = None) -> None:
+        """
+        Args:
+            data: input data source to load and transform to generate dataset for model.
+            transform: a callable data transform on input data.
+        """
+        self.data = data
+        self.transform = transform
```
wyli, line 36:
```suggestion
        self.transform = transform
        self.source = None
```

Nic-Ma: Sure, updated in the latest commit.
Thanks.

wyli (approved):
thanks!

PR #1320, opened 2020-12-03
editing c++/cuda source discovery to allow files in nested directories

### LINKED ISSUE #1319: c++/cuda extention source files not included in build if in nested directory
The build process will currently fail to discover c++/cuda source which is more than a single nested directory deep in the csrc folder. 

The issue is that while we use glob with ** to find the source files, we must add the recursive=True flag to the glob call. 

This has been fixed just need to make the PR.

charliebudd:
Signed-off-by: charliebudd <charles.budd@kcl.ac.uk>

Fixes #1319.

### Description
This small change allows developers to have nested directories in the csrc folder, helping keep code organised. Note that this may look like it removes the ext.cpp from the included source files but this file is picked up on by the source_cpu glob call.   

### Status
**Ready**

### Types of changes


wyli (approved):
thanks!

PR #1324, opened 2020-12-07
Temp fix ITK multi-threads issue

Nic-Ma:
### Description
This PR added temp fix for the multi-threads issue of ITK, referring to @hjmjohnson 's suggestion:
https://github.com/InsightSoftwareConsortium/ITK/issues/2003#issuecomment-739332713

### Status
**Ready**

### Types of changes


monai/data/image_reader.py
```diff
@@ -121,6 +121,10 @@ class ITKReader(ImageReader):
     def __init__(self, **kwargs):
         super().__init__()
         self.kwargs = kwargs
+        if itk.Version.GetITKVersion() < "5.2.1":
```
dzenanz, line 124: This should be 5.2.0, as Hans' fix should make it into 5.2.0.

wyli: perhaps should use numerical comparisons `int(itk.Version.GetITKMajorVersion()) == 5 and int(itk.Version.GetITKMinorVersion()) <= 2 and int(itk.Version.GetITKBuildVersion()) <= 0 `

hjmjohnson: @wyli,  yes!  my code was just psuedocode.  I don't think the above code is valid.

Nic-Ma: Thanks for your suggestions.
Updated in the latest commit.

Thanks.

monai/data/image_reader.py
```diff
@@ -121,7 +121,8 @@ class ITKReader(ImageReader):
     def __init__(self, **kwargs):
         super().__init__()
         self.kwargs = kwargs
-        if itk.Version.GetITKVersion() < "5.2.1":
+        if int(itk.Version.GetITKMajorVersion()) == 5 and int(itk.Version.GetITKMinorVersion()) <= 2 \
+                and int(itk.Version.GetITKBuildVersion()) <= 0:
             # warning the ITK LazyLoading mechanism was not threadsafe until version 5.2.1,
             # a silly call to force loading the required modules before the multi-threading is called
             _ = itk.imread
```
hjmjohnson, line 128: Since this is pretty obscure code, add to the comment:  "requesting access to the itk.imread function triggers the lazy loading of the relevant itk modules before the parallel use of the function."

wyli (approved):
thanks!

monai/data/image_reader.py
```diff
@@ -121,8 +121,11 @@ class ITKReader(ImageReader):
     def __init__(self, **kwargs):
         super().__init__()
         self.kwargs = kwargs
-        if int(itk.Version.GetITKMajorVersion()) == 5 and int(itk.Version.GetITKMinorVersion()) <= 2 \
-                and int(itk.Version.GetITKBuildVersion()) <= 0:
+        if (
+            int(itk.Version.GetITKMajorVersion()) == 5
+            and int(itk.Version.GetITKMinorVersion()) <= 2
```
dzenanz, line 126: Should this be strictly less?

monai/data/image_reader.py
```diff
@@ -121,8 +121,11 @@ class ITKReader(ImageReader):
...
+            and int(itk.Version.GetITKBuildVersion()) <= 0
```
dzenanz, line 127: If so, we don't need this condition.

monai/data/image_reader.py
Nic-Ma: Updated in the latest commit.

monai/data/image_reader.py
Nic-Ma: updated in the latest commit, is it expected now?
Thanks.

dzenanz: Looks good now.

PR #1326, opened 2020-12-07
Adds an image to bounding box transform

wyli:
### Description
a transform that computes nd coordinates of a bounding rectangle from the positive intensities.

### Status
**Ready**

### Types of changes

### DIFF
--- a/docs/source/transforms.rst
+++ b/docs/source/transforms.rst
@@ -105,6 +105,12 @@ Crop and Pad
     :members:
     :special-members: __call__
 
+`BoundingRect`
+""""""""""""""
+.. autoclass:: BoundingRect
+    :members:
+    :special-members: __call__
+
 Intensity
 ^^^^^^^^^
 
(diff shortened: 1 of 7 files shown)


Nic-Ma:
I think the existing `generate_spatial_bounding_box` API may already support this feature?
https://github.com/Project-MONAI/MONAI/blob/master/monai/transforms/utils.py#L508
I think you can enhance the existing API if can't satisfy your request.

Thanks.

wyli:
> Hi @wyli ,
> 
> I think the existing `generate_spatial_bounding_box` API may already support this feature?
> https://github.com/Project-MONAI/MONAI/blob/master/monai/transforms/utils.py#L508
> I think you can enhance the existing API if can't satisfy your request.
> 
> Thanks.

that implementation is 3 times slower than this one. the root cause is that we don't really need to compute 'nonzero_idx = np.nonzero(data)' in order to get the bounding box. I don't have enough time to revise the existing API for now.

Nic-Ma:
OK, let's merge this one first. Then I will try to combine with the existing API soon.

Thanks.

Nic-Ma (approved):

PR #1334, opened 2020-12-09
DynUNet Torchscript Support

### LINKED ISSUE #1245: Add TorchScript support for DynUNet
The DynUNet is not currently compatible because it relies on iteration through lists of modules or the ListModule class. These are not currently compatible with Torchscript so a different architecture for this class is needed. It's possible to inherit from the existing UNet which is structured differently for this reason, however Torchscript does also have issues with the super keyword so a little more design thought is needed to account for the variation in the DynUNet.forward method.

ericspod:
Fixes #1245 .

### Description
Adds Torchscript support to the DynUNet class. This passes the unit tests but may require further inspection by someone more familiar with this network to ensure it functions as expected still.

### Status
**WIP**

### Types of changes


Nic-Ma:
Thanks @ericspod for your quick update on TorchScript support.
@yiheng-wang-nv , could you please help review this PR? Let's make it happen in v0.4 release.

Thanks.

yiheng-wang-nv (approved):
Thanks @ericspod , I have checked the forward consistency for both train and val mode with our current version (I initialized these two versions of the networks with same weights), and the results are correct. I also run a training experiment under the Decathlon Dataset (task 4) and the performance is reasonable.

ericspod:
Thanks guys. There's some cleanup I need to do and run the tests but then the PR should be ready to go.

PR #1339, opened 2020-12-09
fixes params groups

wyli:
### Description
fixes issues of empty generators in generate_param_groups

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/optimizers/utils.py
+++ b/monai/optimizers/utils.py
@@ -61,20 +61,30 @@ def generate_param_groups(
     match_types = ensure_tuple_rep(match_types, len(layer_matches))
     lr_values = ensure_tuple_rep(lr_values, len(layer_matches))
 
+    def _get_select(f):
+        def _select():
+            return f(network).parameters()
+
+        return _select
+
+    def _get_filter(f):
+        def _filter():
+            return filter(f, network.named_parameters())
+
+        return _filter
+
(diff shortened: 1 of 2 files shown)


monai/optimizers/utils.py
```diff
@@ -48,27 +50,31 @@ def generate_param_groups(
...
+        elif ty.lower() == "filter":
             layer_params = filter(func, network.named_parameters())
+            layer_ids = filter(func, network.named_parameters())
         else:
-            raise ValueError(f"unsuppoted layer match type: {ty}.")
+            raise ValueError(f"unsupported layer match type: {ty}.")
 
         params.append({"params": layer_params, "lr": lr})
-        _layers.extend(list(map(id, layer_params)))
+        _layers.extend(list(map(id, layer_ids)))
```
Nic-Ma, line 77: Can we change to `layer_params.deepcopy()` to avoid duplicated `func` and `filter` calls?

Thanks.

Nic-Ma (approved):

monai/optimizers/utils.py
wyli: thanks, I added local functions to return generators, that would work like making fresh copies.

Nic-Ma (approved):
Looks good to me.
Thanks for your quick update.

PR #1369, opened 2020-12-16
Use numpy instead of torch to get random seed

YuanTingHsieh:
Signed-off-by: YuanTingHsieh <yuantinghsieh@gmail.com>

Fixes # .

### Description
For the original way, if we load the model using TorchScript VS load the model using checkpoints.
It will affect the value generated by torch.empty().random_.
If we change to this new way, then this seed will be the same for loading model using either TorchScript or checkpoints.

Note that, even we don't make this change.
The original code is already repeatable. (run several times yield same result)
So this change is not necessary.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/dataloader.py
+++ b/monai/data/dataloader.py
@@ -39,8 +39,7 @@ class DataLoader(_TorchDataLoader):
         if num_workers == 0:
             # when num_workers > 0, random states are determined by worker_init_fn
             # this is to make the behavior consistent when num_workers == 0
-            # torch.int64 doesn't work well on some versions of windows
-            _seed = np.random.randint(np.iinfo(np.uint32).max + 1)
+            _seed = np.random.randint(np.iinfo(np.int32).max + 1)
             set_rnd(dataset, int(_seed))
         if "collate_fn" not in kwargs:
             kwargs.update({"collate_fn": list_data_collate})
--- a/monai/data/dataloader.py
+++ b/monai/data/dataloader.py
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import torch
+import numpy as np
 from torch.utils.data import DataLoader as _TorchDataLoader
 from torch.utils.data import Dataset
 
@@ -40,7 +40,7 @@ class DataLoader(_TorchDataLoader):
             # when num_workers > 0, random states are determined by worker_init_fn
             # this is to make the behavior consistent when num_workers == 0
             # torch.int64 doesn't work well on some versions of windows
-            _seed = torch.empty((), dtype=torch.int32).random_(generator=None).item()
+            _seed = np.random.randint(np.iinfo(np.uint32).max + 1)
             set_rnd(dataset, int(_seed))
         if "collate_fn" not in kwargs:
             kwargs.update({"collate_fn": list_data_collate})


wyli:
Hi @YuanTingHsieh the default multiprocess random seeds are from here: https://github.com/Project-MONAI/MONAI/blob/067875b847605ac97ad49569875288a82b101320/monai/data/utils.py#L226-L227  the current implementation ensures the same seed when `num_workers` is  0 or 1. I don't completely understand your use case, if really necessary, we could add an option `np_pt_rnd` in DataLoader

YuanTingHsieh:
Hi @wyli 

The use case is for verification of the exported model VS trained model using the same pipeline.
If the pipeline contains random transforms and we use MONAI's set_determinism

we do

```
composed_trans = Compose([LoadImaged, SomeRandomTransform])
model = torch.jit.load('model.ts')
data_loader = DataLoader(composed_trans)

# doing some validation
```

VS

```
composed_trans = Compose([LoadImaged, SomeRandomTransform])
model = UNet()
model.load_state_dict(torch.load('model.pt'))
data_loader = DataLoader(composed_trans)

# doing some validation
```

If we set the seed using numpy, then with num_workers == 0 we can guarantee these two results are the same.
But again, I am not sure if we need to make this happen.
@Nic-Ma  can add some comments.

wyli:
ok, in your code example, the cases mainly differ in the model initiailisation 
method 1: 
```model = torch.jit.load('model.ts')```
and method2:
```model = UNet(); model.load_state_dict(torch.load('model.pt'))```

but I don't see why they affect the torch random states, could you explain more?

**Comment by YuanTingHsieh** [comment]
Yeah I don't understand why model init would affect random state either.
But I observe that is what is happening.

You can run the following two codes and check.

```
import torch
import numpy as np

from monai.networks.nets import Unet
from monai.utils import set_determinism

set_determinism(seed=0)
model = Unet(
          dimensions=3,
          in_channels=2,
          out_channels=2,
          channels=[16, 32, 64, 128, 256],
          strides=[2, 2, 2, 2],
          num_res_units=2,
          norm='batch'
        )
model.load_state_dict(torch.load('./models/model.pt'))

torch_seed = torch.empty((), dtype=torch.int32).random_(generator=None).item()
numpy_seed = np.random.randint(np.iinfo(np.int32).max + 1)
print(f'Torch seed is {torch_seed}')
print(f'Numpy seed is {numpy_seed}')

```

On my ubuntu this yields:
Torch seed is 971764478
Numpy seed is 209652396

VS

```
import torch
import numpy as np

from monai.networks.nets import Unet
from monai.utils import set_determinism

set_determinism(seed=0)
model = torch.jit.load('./models/model.ts')

torch_seed = torch.empty((), dtype=torch.int32).random_(generator=None).item()
numpy_seed = np.random.randint(np.iinfo(np.int32).max + 1)
print(f'Torch seed is {torch_seed}')
print(f'Numpy seed is {numpy_seed}')

```

This yields:
Torch seed is 209652396
Numpy seed is 209652396

```

wyli:
thanks for the detailed investigation, I think we couldn't fully resolve this unless we change this pytorch impl., correct? https://github.com/pytorch/pytorch/blob/1a92802bde28286f850b415bfd985515565740c3/torch/utils/data/dataloader.py#L498

YuanTingHsieh:
@wyli I think you are right.
@madil90 @Nic-Ma You can see the discussion

So we can just close this PR then?

Nic-Ma:
Hi @YuanTingHsieh ,

Yes, maybe you can close this PR now, but we need to keep it in mind and try to figure it out someday, or at least make sure we can explain to others why the TorchScript result with the random operation is different.
Thanks very much for your deep analysis, now it's much clear for us.

Thanks.

PR #1371, opened 2020-12-16
adding .venv directory to ignores for flake8 and isort

### LINKED ISSUE #1370: Auto formatting includes '.venv' directory
**Describe the bug**
flake8 and isort are configured to ignore 'venv' not the also common '.venv' directory leading to false fails and corrupted virtual environments 

**To Reproduce**
1. Setup dev environment with a virtual environment named '.venv' 
2. run './runtests.sh --codeformat --nounittests'
3. tests will fail due to files in in the virtual environment folder

**Expected behavior**
virtual environment folder ignored by flake8 and isort

**Additional context**
'.venv' is a common naming convention and is skipped by isorts default settings...
 [https://pycqa.github.io/isort/docs/configuration/options/](url)

charliebudd:
Signed-off-by: charliebudd <charles.budd@kcl.ac.uk>

Fixes #1370 .

### Description
Simply adds '.venv' to the list of ignores. It would be nice to use the VIRTUAL_ENV environment variable which I believe is accessable on linux and windows. However as we use a .cfg file I don't think it can be evaluated when the config is loaded.

I can use the environment variable by directly adding the ignore in runtests.sh however this then overwrites the ignores in the .cfg files

### Status
**Ready**

### Types of changes

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -65,13 +65,13 @@ ignore =
     # N812 lowercase 'torch.nn.functional' imported as non lowercase 'F'
     N812
 per-file-ignores = __init__.py: F401
-exclude = *.pyi,.git,.eggs,monai/_version.py,versioneer.py,venv,_version.py
+exclude = *.pyi,.git,.eggs,monai/_version.py,versioneer.py,venv,.venv,_version.py
 
 [isort]
 known_first_party = monai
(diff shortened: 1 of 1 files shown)


wyli (approved):
thanks!

PR #1375, opened 2020-12-16
1198 efficient differentiable bilateral filter

### LINKED ISSUE #1198: Efficient, differentialble bilateral filter
This would be a useful tool to have in the box, both on its own or part of a larger algorithm such as #315

similar issue for more context... 
https://github.com/kornia/kornia/issues/702

Would be nice to have an implementation which can operate on arbitrary spatial dimensions and channels. 
Due to the high computational load I think dropping down to c++ and cuda could be worth the extra work.

charliebudd:
Fixes #1198 .

### Description
Adds two bilateral filter algorithms to the c++ extention module. Each have both a cpu and a cuda implementation. 
One works by a bruteforce kernel, the other using a more refined approximate solution presented here...[https://graphics.stanford.edu/papers/permutohedral/](url) 
The desired implementation is chosen via a boolean flag at the python interface level.

There are improvements which could be made but this is long overdue and already a very large PR. 
There are also two issues I'm aware of but both are imperceptible at normal scales...
- The approximate cuda implementation shows some small variablity in its output.  
- Both the approximate cpu and approximate cuda implementations show slight asymetry in the filtered output (This may be an artifact of the approximation method)

### Status
**Ready**

### Types of changes


monai/csrc/utils/meta_macros.h
```diff
@@ -0,0 +1,119 @@
+#pragma once
```
wyli, line 1: could you add the copyright sentences to this file

wyli (approved):
thanks @charliebudd this is very nice, would be great to have a short notebook demo

faebstn96:
Hi @charliebudd ,
Thanks for your implementation. Unfortunately, I am failing to get the layer running. When trying to call the forward pass of the bilateral filter with
filter_layer = monai.networks.layers.BilateralFilter(spatial_sigma=5, color_sigma=0.5, fast_approx=True)
filter_layer.apply(tensor_in)  # tensor_in: arbitrary 3,4 or 5 dim torch tensor

I get problems with the save_for_backward function, as it does not accept the filter parameters (I think save_for_backward is actually made for storing the input tensor and not for the parameters):
TypeError: save_for_backward can only save variables, but argument 0 is of type int

Wouldn't it be easier to implement the bilateral filter class as a nn.Module instead of torch.autograd.Function (like, e.g., the Gaussian filter)? Then, one could store the parameters in an instance variable.

Thanks for your help!

faebstn96:
I just recognized, you can just assign the parameters directly to ctx instead of using save_for_backward. E.g., like
ctx.spatial_sigma = spatial_sigma

Then the layer at least does not throw an error.

charliebudd:
Indeed, thanks for pointing out the error. I have made the changes and have drafted a PR #1888. Sorry for the error

faebstn96:
Thank you so much!

Issue #528, opened 2020-06-10
Class API for Visualize Module


abpwrs:
## Is your feature request related to a problem? Please describe.
The current `monai.visualize` module is highly coupled with Tensorboard's `SummaryWriter`. It'd be convenient to have a set of classes designed to process tensors/arrays both inside and outside the training loop to generate common visualizations, like a segmentation overlay.

## Describe the solution you'd like

A restructuring of `monai.visualize` to support out of training loop visualization, and easier creation of custom visualization. I could see this including the following: 

- [ ] Create an abstract `Visualization` base class
- [ ] Refactoring current 2D/3D visualizers to classes that implement `Visualization`
- [ ] Refactor `TensorBoardImageHandler` to support a list of `Visualization` objects
- [ ] Add an `add(viz : Visualization)` method, or the like, to `TensorBoardImageHandler`

## Describe alternatives you've considered
The `monai.visualize` module could also be restructured to be a functional API instead of using classes.

## Additional context
Example usages:

#### Outside training loop
```python
image_data = itk.GetArrayFromImage(itk.imread("image1.nii.gz")
seg_data = itk.GetArrayFromImage(itk.imread("seg1.nii.gz")
segmentation_overlay_visualizer = monai.visualize.SegmentationOverlayVisualization()
viz = segmentation_overlay_visualizer(image_data, seg_data)
```

#### Inside training loop
```python
train_tensorboard_image_handler = TensorBoardImageHandler(
    batch_transform=lambda batch: (batch[0], batch[1]),
    output_transform=lambda output: predict_segmentation(output[0]),
    global_iter_transform=lambda x: trainer.state.epoch
)

# the add function could take in any class that implements the Visualization class
train_tensorboard_image_handler.add(monai.visualize.SegmentationOverlayVisualization())

trainer.add_event_handler(event_name=Events.EPOCH_COMPLETED, handler=train_tensorboard_image_handler)

```

wyli:
thanks, is the variable`viz` an image array in `viz = segmentation_overlay_visualizer(image_data, seg_data)`? if yes then I guess `segmentation_overlay_visualizer` is actually a data transform which does some postprocessing and fuses image/seg data into one?

abpwrs:
Correct, this would be post-processing of the data to overlay the outline of the segmentation onto the image.

wyli:
[...]

PR #1381, opened 2020-12-17
Unify visualisation

rijobro:
Unify visualisation classes.

### Description
Beforehand, occlusion sensitivity was a function in `metrics`, whereas gradcam functionality was in classes inside the `visualize` module. Now, both can be called with similar calls:

```
from monai.networks.nets import densenet121
from monai.visualise import OcclusionSensitivity
from monai.visualise import CAM
model_2d = densenet121(spatial_dims=2, in_channels=1, out_channels=3)

cam = CAM(nn_module=model_2d, target_layers="class_layers.relu", fc_layers="class_layers.out")
result_cam = cam(x=torch.rand((1, 1, 48, 64)))

occ_sens = OcclusionSensitivity(nn_module=model_2d)
result_occ_sens = occ_sens(x=torch.rand((1, 1, 48, 64)), class_idx=None, b_box=[-1, -1, 2, 40, 1, 62])
```

We also had mixed American and British spelling, so changed all instances of `visualize` to `visualise`. Maybe I should have gone the other way, as it made lots of changes...

### Status
**Ready**

### Types of changes


wyli:
I feel we should go for `visualize`... to be consistent with pytorch and numpy

rijobro:
ok, updated.

wyli:
thanks for the clean up! but I think it's too early to define a proper NetVisualizer API, we should keep the module simple for now, and it needs more discussions along with https://github.com/Project-MONAI/MONAI/issues/528

rijobro:
> but I think it's too early to define a proper NetVisualizer API

I can remove it if you'd like but it's only a skeleton at the moment, with a constructor and `__call__` method. Just so that when future visualisations get added, they are more likely to conform (avoid the difference we saw with OccSens and GradCAM).

Up to you, I'll delete if you'd like, but I'm in favour of keeping it.

wyli:
> > but I think it's too early to define a proper NetVisualizer API
> 
> I can remove it if you'd like but it's only a skeleton at the moment, with a constructor and `__call__` method. Just so that when future visualisations get added, they are more likely to conform (avoid the difference we saw with OccSens and GradCAM).
> 
> Up to you, I'll delete the base class if you'd like, but I'm in favour of keeping it.

I feel the benefit of having a `NetVisualizer` class is not very clear at the moment (and the interface is also not well documented), all the other changes look very nice. perhaps @Nic-Ma @ericspod could have more comments.

rijobro:
Ok, I can remove it. 

I also created a `train_mode` context manager (opposite of `eval_mode`) [here](https://github.com/Project-MONAI/MONAI/pull/1381/files#diff-d3917404e925ec1a84e22f35ad430e32d6ca3b0de28bdcbf0fd2a8bf2d2b8777R283). We can use this when calculating the gradient for GradCam, which should avoid the problem you saw here: https://github.com/Project-MONAI/tutorials/pull/83#issuecomment-748003086.

wyli (approved):
thanks!

PR #1384, opened 2020-12-17
with eval_mode(net):

rijobro:
### Description
Current:
```
net.eval()
with torch.no_grad():
    pass
net.train()
```
Proposed:
```
with eval_mode(net):
    pass
```

This is also useful when the `train`/`eval` state of a network could be one or the other, yet we want it to temporarily be in `eval` mode and return to original state at the end (e.g., gradcam).

To avoid conflicts, I haven't updated gradcam or occlusion sensitivity until https://github.com/Project-MONAI/MONAI/pull/1381 is merged. I'll also update the tutorials.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/networks/utils.py
+++ b/monai/networks/utils.py
@@ -13,6 +13,7 @@ Utilities and types for defining networks, these depend on PyTorch.
 """
 
 import warnings
+from contextlib import contextmanager
 from typing import Any, Callable, Optional, Sequence, cast
 
 import torch
@@ -29,6 +30,7 @@ __all__ = [
     "normal_init",
     "icnr_init",
     "pixelshuffle",
+    "eval_mode",
 ]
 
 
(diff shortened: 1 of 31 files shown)


monai/networks/utils.py
```diff
@@ -241,3 +243,36 @@ def pixelshuffle(x: torch.Tensor, dimensions: int, scale_factor: int) -> torch.T
...
+            print(p.training)  # False
+            print(p(t).sum().backward())  # will correctly raise an exception as gradients are calculated
+    """
+
+    # Get original state of network(s)
+    training = [n for n in nets if n.training]
+
+    try:
+        # set to eval mode
+        with torch.no_grad():
```
wyli, line 273: perhaps add some checks for the attributes `hasattr(nets, "train")` `hasattr(nets, "training")`

ericspod: If we don't have the checks the exception will be raised and passed to the caller. Since this will happen on the first line I think it's safe to rely on Pythonic way of deferring checks to runtime. The requirement that the arguments are `nn.Module` instances should let users know what's expected as well.

wyli (approved):

PR #1394, opened 2020-12-23
1378 Add output_dtype to NiftiSaver and SegmentationSaver

### LINKED ISSUE #1378: SegmentationSaver fails if dtype is np.uint8
SegmentationSaver fails with RuntimeError: "inverse_cpu" not implemented for 'Byte' if dtype is np.uint8 in monai.networks.utils line 158. It hapens when the output image that should be written by the Segmentation Saver should be uint8 (which in my eyes is the most meaningfull datatype for saving labelmaps) and an affine transformation is to be applied to rescale the output to the same spacing as the inputimage was. 

It debugged it down to monai.data.nifty_writer.py line 135 wehre the dtype of the SegmentationSaver is set before the transormation is done. This is happening since monai 3.0 and is still the case in monai 4.0. In monai 2.0 the dtype before the call off affine_xform was hardcoded to np.float32 so that the transformation worked out and the desired output dtype was just set at the end of nifti_writer.py. The workaround for now is to not set the output datatype for the SegmentationSaver so it defaults to np.float32, however the output labelmaps get much larger compared to using np.unit8.

**To Reproduce**
Use the SegmentationSaver with some integer dtype, and make it use an affine transformation (eg. rescale to the original spacing of the inputfile with batch_transform=lambda batch: batch["somedictkey_meta_dict"])

**Expected behavior**
Should write uint8 labelmaps

**Environment**
monai 4.0 and 3.0,
works in monai 2.0

Thank you guys once more for the amazing work :-)
Best Eli

Nic-Ma:
Fixes #1378 .

### Description
This PR added `output_dtype` arg to `NiftiSaver` and `SegmentationSaver` to align with `nifti_writer`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/nifti_saver.py
+++ b/monai/data/nifti_saver.py
@@ -37,6 +37,7 @@ class NiftiSaver:
         padding_mode: Union[GridSamplePadMode, str] = GridSamplePadMode.BORDER,
         align_corners: bool = False,
         dtype: Optional[np.dtype] = np.float64,
+        output_dtype: Optional[np.dtype] = np.float32,
     ) -> None:
         """
         Args:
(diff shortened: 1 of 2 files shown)


wyli (approved):
thanks!

PR #1411, opened 2021-01-06
Wildcard import

rijobro:
### Description
Avoid `import *`'s. Add to `runtests.sh` for future additions.

### Status
**Ready**

### Types of changes


wyli:
this is a good point, is it possible to achieve the same by not ignoring F405 flake8 and add some config? https://github.com/Project-MONAI/MONAI/blob/a74fef9bae878bd2826bb6b917f81e097c17bb10/setup.cfg#L64

rijobro:
Yup, looks good. What do you think I should I do for `import *` in `__init__.py` files?

wyli:
> Yup, looks good. What do you think I should I do for `import *` in `__init__.py` files?

this tool suggests to set `__all__` https://lgtm.com/projects/g/Project-MONAI/MONAI/alerts/?mode=tree&ruleFocus=3980091 what do you think?

rijobro:
`import *` is fine when `__all__` is defined, but I think readability is reduced since you don't necessarily know where something was imported from. Personally, I would prefer explicitly importing individual modules. 

But I'm not sure if `import *` is OK inside of an `__init__.py` file...

wyli:
> `import *` is fine when `__all__` is defined, but I think readability is reduced since you don't necessarily know where something was imported from. Personally, I would prefer explicitly importing individual modules.
> 
> But I'm not sure if `import *` is OK inside of an `__init__.py` file...

sure I'm fine with either. I think both are valid for `__init__.py`. @ericspod @Nic-Ma comments?

ericspod:
Wildcard imports in `__init__.py` files is standard if `__all__` is defined in the module being imported. Wildcard imports elsewhere are discouraged by flake8 and other sources. Somehow you want to export from a module the things you want to be part of the public interface, so filtering is done in the `__init__.py` file with selective imports from submodules or with wildcards and `__all__` used in submodules to define what's public. From what I've seen the selective imports isn't as common, it has the effect of defining what should be public in a different place from where things are actually defined. Using `__all__` keeps everything about a definition in one file. 

What MONAI does is a bit of a mix at the moment, what I had implemented before was an `export` decorator to fill the role of `__all__` but I think now we're better off without it, though still using the auto-import `load_submodules` routine. What we don't want is the statement `from monai import *` to import every definition into the current namespace, there needs to be some hierarchy somewhere still.

rijobro:
Couldn't figure out how to set flake8 up so that `import *` was allowed inside of `__init__.py` provided `__all__` was defined in the import module. In the end I simply removed all `import *`.

wyli:
thanks, looks good to me. 

optionally could add a sentence in `CONTRIBUTING.md` about adding new function/class?

(also doesn't setting `per-file-ignores` work? https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-per-file-ignores

https://github.com/Project-MONAI/MONAI/blob/db7bf8984712f3e46ed533611a70c9b8cd0293bd/setup.cfg#L67)

rijobro:
> optionally could add a sentence in CONTRIBUTING.md about adding new function/class?

Will do.

> (also doesn't setting per-file-ignores work? https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-per-file-ignores

We could ignore all `__init__.py` files, but that's not really what we want (we want to flag up an error if `import *` is used with no `__all__`).

wyli (approved):
thanks it looks good to me.

PR #1413, opened 2021-01-06
fixes Dockerfile

wyli:
### Description
fixes docker build error https://github.com/Project-MONAI/MONAI/runs/1658975018#step:3:2335

### Status
**Ready**

### Types of changes


ericspod:
The issue with the comment breaking up a multiline shell command is probably common, I've done it myself. Is there a reason to chain commands together with && rather than have multiple RUN commands? If we had multiple command we can comment them safely, but I assume using && is a common Docker pattern for some reason.

wyli:
yeah I think it minimises the number of layers and size of the image

ericspod:
A possible solution to interspersed comments [isn't exactly pretty](https://stackoverflow.com/a/12797512).

Nic-Ma (approved):

PR #1418, opened 2021-01-07
New simplelayer SavitzkyGolayFilter() and transform SavitzkyGolaySmooth()

### LINKED ISSUE #1155: Savitzky-Golay kernel for data smoothing
[Savitzky-Golay filtering](https://en.wikipedia.org/wiki/Savitzky%E2%80%93Golay_filter) is widely used in fields such as ultrasound imaging and spectral imaging. It is often a prefered option in comparison to Gaussian filtering in such applications.

The Savitzky-Golay filter is a simple discrete convolution filter and should thus be straightforward to integrate in a PyTorch environment such as MONAI. The computation of the kernel wigths only depends on the order of the polynomial and the window size chosen for the underpinning (implicit) least-squares fit.

Exisiting options include computing the filter weights using an external routine, e.g. [`scipy.signal.savgol_coeffs`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.savgol_coeffs.html), and then pass these weigths to PyTorch/MONAI.

**Additional context**
A solution bypassing the dependency on `scipy` would be nice. The kernel computation only involves fairly basic linear algebra, all available in PyTorch.

crnbaker:
Fixes #1155  .

### Description
New simplelayer `SavitzkyGolayFilter()` applies Savitzky Golay filter to a Tensor along a given axis. New array transform `SavitzkyGolaySmooth()` uses `SavitzkyGolayFilter()`. Required change to `separable_filtering()` simplelayer: now has option to use padding modes other than 'zeros' during convolution.

Example: Smoothing a noisy pulse, result compared to `SciPy.signal.savgol_filter`.
image

Performance compared to `SciPy.signal.savgol_filter`:
image

Above plots generated using [this Google CoLab notebook](https://colab.research.google.com/drive/1qG8kdcr9Cpdywy_rWbrB8gJIj3hDenn8?usp=sharing).

### Status
Ready

### Types of changes


monai/networks/layers/simplelayers.py
```diff
@@ -184,23 +195,106 @@ def separable_filtering(x: torch.Tensor, kernels: Union[Sequence[torch.Tensor],
...
-        conv_type = [F.conv1d, F.conv2d, F.conv3d][spatial_dims - 1]
-        return conv_type(input=_conv(input_, d - 1), weight=_kernel, padding=_padding, groups=n_chns)
+
+        if mode == "zeros":  # if zero padding (default), can use functional convolution
+            conv_type = [F.conv1d, F.conv2d, F.conv3d][spatial_dims - 1]
+            return conv_type(input=_conv(input_, d - 1), weight=_kernel, padding=_padding, groups=n_chs)
+        else:
+            _k_len = _kernel.shape.numel()
+            _pad_tup = tuple(_padding)
+            convolver = conv_factory(spatial_dims)(
```
wyli, line 219: `conv_factory` usage is something like:
```py
from monai.networks.layers.factories import Conv

Conv[Conv.CONV, spatial_dims]
```

also why not merge the functional case into the nn.Conv* case?

tests/utils.py
```diff
@@ -110,8 +110,12 @@ def __init__(self, pytorch_version_tuple):
         self.min_version = pytorch_version_tuple
         if has_pkg_res:
             self.version_too_old = ver(torch.__version__) < ver(".".join(map(str, self.min_version)))
-        else:
+        elif len(self.min_version) == 2:
             self.version_too_old = get_torch_version_tuple() < self.min_version
+        elif len(self.min_version) == 3:
+            self.version_too_old = tuple(int(x) for x in torch.__version__.split(".")[:3]) < self.min_version
```
wyli, line 116: perhaps not working for torch version like `1.7.0a0+7036e91`

wyli (commented):
thank you! please see some minor comments inline

monai/networks/layers/simplelayers.py
```diff
@@ -163,18 +164,28 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
             could be a single kernel (duplicated for all dimension), or `spatial_dims` number of kernels.
+        mode (string, optional): padding mode passed to convolution class. ``'zeros'``, ``'reflect'``, ``'replicate'``
+            or ``'circular'``. Default: ``'zeros'``. Modes other than ``'zeros'`` require PyTorch version >= 1.5.1. See
+            torch.nn.Conv1d() for more information.
 
     Raises:
         TypeError: When ``x`` is not a ``torch.Tensor``.
     """
+
+    pytorch_version_pre_1_5_1 = tuple(int(x[0]) for x in torch.__version__.split(".")[:3]) < (1, 5, 1)
```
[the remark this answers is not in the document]
crnbaker: On second thought, I think it would be simpler to only support 1.6+ (rather than 1.5.1+), so that I can use the PyTorch version checking built in to MONAI. That would solve this problem and a few of the others you have highlighted.

tests/utils.py
crnbaker: I will change the new functionality to support only PyTorch 1.6+ instead of 1.5.1+, so these lines will no longer be needed.

monai/networks/layers/simplelayers.py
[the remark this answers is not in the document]
wyli: thanks! I guess we could drop torch 1.5 support at the project level soon...

monai/networks/layers/simplelayers.py
crnbaker: I had some trouble getting the `nn.Conv*` classes to work for all tests, so am now using `nn.functional.pad` to pad the Tensors directly, allowing use of `nn.functional.conv*`. This has added benefit of working with PyTorch versions before 1.5.1, so I have also removed all the version checking.

tests/utils.py
crnbaker: I have rolled back the changes made to this file as they are no longer needed.

monai/networks/layers/simplelayers.py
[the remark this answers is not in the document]
crnbaker: As described below, this version check is no longer needed as the new functional implementation supports 1.5.0. These lines have been removed in the latest push.

wyli (approved):
thanks! would be great to have the benchmarking notebook in https://github.com/Project-MONAI/tutorials

crnbaker:
Cool I will move it there, thanks for approving!

PR #1419, opened 2021-01-07
Improve docker build process

IsaacYangSLA:
COPY only needed

Fix LICENSE

Verify the tests in docker still works

Signed-off-by: Isaac Yang <isaacy@nvidia.com>

Fixes # .

### Description
Added .dockerignore to filter out unnecessary files/folder from the docker build context
Specified files/folders to be added into docker images, instead of all files/folders in current folder.
May still need to figure out which should 'exactly' be included in the release container images.  Developers may love to have tests and docs, but not for runtime users.

### Status
**Ready**

### Types of changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ RUN cp /tmp/requirements.txt /tmp/req.bak \
 # please specify exact files and folders to be copied -- else, basically always, the Docker build process cannot cache
 # this or anything below it and always will build from at most here; one file change leads to no caching from here on...
 
-COPY LICENSE setup.py setup.cfg versioneer.py runtests.sh .gitignore .gitattributes MANIFEST.in ./
+COPY LICENSE setup.py setup.cfg versioneer.py runtests.sh .gitignore .gitattributes README.md MANIFEST.in ./
 COPY tests ./tests
 COPY monai ./monai
 COPY .git ./.git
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -28,7 +27,11 @@ RUN cp /tmp/requirements.txt /tmp/req.bak \
 # TODO: remark for issue [revise the dockerfile #1276](https://github.com/Project-MONAI/MONAI/issues/1276)
 # please specify exact files and folders to be copied -- else, basically always, the Docker build process cannot cache
 # this or anything below it and always will build from at most here; one file change leads to no caching from here on...
-COPY . .
+
+COPY LICENSE setup.py setup.cfg versioneer.py runtests.sh ./
+COPY tests ./tests
+COPY monai ./monai
+COPY .git ./.git
```
wyli, line 34: perhaps we could have`.gitignore`, `.gitattributes` as well?

IsaacYangSLA: I can see .gitattributes interacts with versioneer.py, so it's necessary to add it back.  Additionally, MANIFEST.in is used during python setup.py and I put it back to Dockerfile and removed it from .dockerignore.
I am not 100% sure if .gitignore is used inside container.  However, its content should be pretty stable and won't invalidate cache often, so it's also included, too.
The updated Dockerfile and .dockerignore was just force-pushed.

wyli (approved):
thanks, dockerfile is tested locally

PR #1430, opened 2021-01-11
Update occ sens

rijobro:
Last update to occlusion sensitivity.

### Description
@wyli this finally behaves as I expect and I'll upload the tutorials to match. 

Now you can:
- I left the output without any postprocessing. The absolute values are useful, as they indicate how much 
- Output now has an extra dimension of size N added to the end, where N is the number of classes.
- Also returned is the most probable class. i.e., when part of the image is occluded, does the predicted class change, and if so, to what?

### Status

### Types of changes

### DIFF
--- a/monai/visualize/__init__.py
+++ b/monai/visualize/__init__.py
@@ -9,8 +9,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .visualizer import default_normalizer, default_upsampler  # isort:skip
-from .class_activation_maps import CAM, GradCAM, GradCAMpp, ModelWithHooks
+from .visualizer import default_upsampler  # isort:skip
+from .class_activation_maps import CAM, GradCAM, GradCAMpp, ModelWithHooks, default_normalizer
 from .img2tensorboard import (
     add_animated_gif,
     add_animated_gif_no_channels,
--- a/monai/visualize/__init__.py
+++ b/monai/visualize/__init__.py
@@ -9,7 +9,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .visualizer import default_upsampler  # isort:skip
 from .class_activation_maps import CAM, GradCAM, GradCAMpp, ModelWithHooks, default_normalizer
 from .img2tensorboard import (
     add_animated_gif,
@@ -18,3 +17,4 @@ from .img2tensorboard import (
     plot_2d_or_3d_image,
 )
 from .occlusion_sensitivity import OcclusionSensitivity
+from .visualizer import default_upsampler
--- a/monai/visualize/class_activation_maps.py
+++ b/monai/visualize/class_activation_maps.py
@@ -12,15 +12,30 @@
 import warnings
 from typing import Callable, Dict, Sequence, Union
 
+import numpy as np
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
 
 from monai.networks.utils import eval_mode, train_mode
+from monai.transforms import ScaleIntensity
 from monai.utils import ensure_tuple
-from monai.visualize import default_normalizer, default_upsampler
+from monai.visualize import default_upsampler
 
-__all__ = ["CAM", "GradCAM", "GradCAMpp", "ModelWithHooks"]
+__all__ = ["CAM", "GradCAM", "GradCAMpp", "ModelWithHooks", "default_normalizer"]
+
+
+def default_normalizer(x) -> np.ndarray:
+    """
+    A linear intensity scaling by mapping the (min, max) to (1, 0).
+
+    N.B.: This will flip magnitudes (i.e., smallest will become biggest and vice versa).
+    """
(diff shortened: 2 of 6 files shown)


monai/visualize/__init__.py
```diff
@@ -9,8 +9,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .visualizer import default_normalizer, default_upsampler  # isort:skip
-from .class_activation_maps import CAM, GradCAM, GradCAMpp, ModelWithHooks
+from .visualizer import default_upsampler  # isort:skip
```
wyli, line 12: perhaps remove the isort:skip?

monai/visualize/occlusion_sensitivity.py
```diff
@@ -130,168 +145,182 @@ class OcclusionSensitivity:
...
-                speed the analysis up, which might be useful for larger images.
-                * Min and max are inclusive, so [0, 63, ...] will have size (64, ...).
-                * Use -ve to use 0 for min values and im.shape[x]-1 for xth dimension.
-        Returns:
-            Depends on the postprocessing, but the default return type is a Numpy array.
-            The returned image will occupy the same space as the input image, unless a
-            bounding box is supplied, in which case it will occupy that space. Unless
-            upsampling is disabled, the output image will have voxels of the same size
-            as the input image.
+        :param x: image to test. Should be tensor consisting of 1 batch; spatially can be 2- or 3D.
```
wyli, line 272: the previous docstring style is consistent with the other files

monai/visualize/occlusion_sensitivity.py
```diff
@@ -130,168 +145,170 @@ class OcclusionSensitivity:
...
-        sensitivity_im = sensitivity_im.reshape(tuple(downsampled_im_shape)).unsqueeze(0)
+        for i in range(num_classes):
+            sensitivity_ims[i] = sensitivity_ims[i].reshape(tuple(downsampled_im_shape)).unsqueeze(0)
 
-        return sensitivity_im, output_im_shape
+        return sensitivity_ims, output_im_shape
 
     def __call__(  # type: ignore
-        self, x: torch.Tensor, class_idx: Optional[Union[int, torch.Tensor]] = None, b_box: Optional[Sequence] = None
+        self,
```
wyli, line 267: I thought with non-empty `class_idx` option it'll save some memory here?

wyli (commented):
thanks -- some minor comments in line

monai/visualize/__init__.py
rijobro: The skip is needed unfortunately. Otherwise `visualizer` will be moved to the end (alphabetically) and then there's a circular dependency.

monai/visualize/occlusion_sensitivity.py
rijobro: There's a marginal increase in memory, the output image might be e.g., `64,64,64,3` instead of `64,64,64` for 3D and even smaller for 2D. Since batch size is always 1, the images will always be small, and it seems a shame to discard the information when it's generated anyway during the inference step at no extra cost.

monai/visualize/occlusion_sensitivity.py
rijobro: Ok, will sort this tomorrow.

monai/visualize/class_activation_maps.py
```diff
@@ -12,15 +12,30 @@
...
+import numpy as np
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
 
 from monai.networks.utils import eval_mode, train_mode
+from monai.transforms import ScaleIntensity
 from monai.utils import ensure_tuple
-from monai.visualize import default_normalizer, default_upsampler
+from monai.visualize import default_upsampler
```
wyli, line 23: we could address the isort issue by
```suggestion
from monai.visualize.visualizer import default_upsampler
```

monai/visualize/occlusion_sensitivity.py
```diff
@@ -18,7 +18,7 @@
 import torch.nn as nn
 
 from monai.networks.utils import eval_mode
-from monai.visualize import default_normalizer, default_upsampler
+from monai.visualize import default_upsampler
```
wyli, line 21:
```suggestion
from monai.visualize.visualizer import default_upsampler
```

rijobro: With that modification: `ModuleNotFoundError: No module named 'monai.visualize.visualizer.visualizer'; 'monai.visualize.visualizer' is not a package`

wyli: I checked out this branch and this seems to work fine from my side: 
```py
# from monai.visualize import default_upsampler
from monai.visualize.visualizer import default_upsampler
```

rijobro: Apologies, you're right!

wyli (approved):
thanks!

PR #1474, opened 2021-01-20
Add more Events for MONAI workflow engines

Nic-Ma:
### Description
This PR registered more Events to ignite engine for MONAI workflow trainer and evaluator.
Users can write more powerful handlers based on it.
If user has very special `iteration` computation, they can still use `iteration_update()` callback.

### Status
**Ready**

### Types of changes


wyli (changes requested):
thanks this needs more discussions and tests

Nic-Ma:
Hi @wyli and @yanchengnv ,

Thanks for your reminder, I added test cases in the integration tests.
I think it's ready for review now.

Thanks.

Nic-Ma:
And BTW, this PR refers to ignite example: https://github.com/pytorch/ignite/blob/master/ignite/engine/events.py#L146
Thanks for @vfdev-5 's help.

vfdev-5:
@Nic-Ma thanks for mentioning ! Just FYI, currently we have undocumented requirements for custom events if using filters with them: https://github.com/pytorch/ignite/issues/1555
Hopefully, this will be fixed in the next release v0.4.3 and in v0.5.0 user wont need to increment state attribute counter: 
```
def train_step(engine, batch):
    engine.fire_event(TestEvents.TEST_EVENT)
    engine.state.test_event += 1
```

Nic-Ma:
Hi @vfdev-5 ,

Thanks for your tips, I didn't add `event_to_attr` dict to the `register_events()` call because I think we don't have count attributes to keep in State for these 3 Events. So we don't need `engine.state.test_event += 1`, right?

Thanks.

monai/engines/trainer.py
```diff
@@ -153,14 +157,20 @@ def _iteration(self, engine: Engine, batchdata: Dict[str, torch.Tensor]):
         if self.amp and self.scaler is not None:
             with torch.cuda.amp.autocast():
                 predictions = self.inferer(inputs, self.network, *args, **kwargs)
```
wyli, line 159: this won't support all the use cases, for example, we might need an accumulated trainer like this: https://gist.github.com/thomwolf/ac7a7da6b1888c2eeac8ac8b9b05d3d3#file-gradient_accumulation-py

monai/engines/utils.py
```diff
@@ -9,11 +9,41 @@
...
+    Addtional Events engine can register and trigger in the iteration process.
+    Refer to the example in ignite: https://github.com/pytorch/ignite/blob/master/ignite/engine/events.py#L146
+    These Events can be triggered during training iteration:
+    `PREDICT_COMPLETED` is the Event when `network(image, label)` completed.
+    `LOSS_COMPLETED` is the Event when `loss(pred, label)` completed.
+    `BACKWARD_COMPLETED` is the Event when `loss.backward()` completed.
+
+    """
+
+    PREDICT_COMPLETED = "predict_completed"
```
wyli, line 44: could be `FORWARD_COMPLETED` to be consistent with the backward?

monai/engines/workflow.py
```diff
@@ -119,44 +119,53 @@ def set_sampler_epoch(engine: Engine):
...
+            metrics.update(add_metrics)
+        for name, metric in metrics.items():
+            metric.attach(self, name)
+
+    def _register_handlers(self, handlers):
+        handlers_ = ensure_tuple(handlers)
+        for handler in handlers_:
+            handler.attach(self)
+
+        @self.on(Events.EPOCH_COMPLETED)
```
wyli, line 161: this could be a generic event handler for best metric/loss

monai/engines/workflow.py
```diff
@@ -119,44 +119,53 @@ def set_sampler_epoch(engine: Engine):
...
+
+    def _register_metrics(self, k_metric, add_metrics):
+        if not isinstance(k_metric, dict):
+            raise TypeError(f"key_metric must be None or a dict but is {type(key_metric).__name__}.")
+        self.state.key_metric_name = list(k_metric.keys())[0]
+        metrics = k_metric
+        if add_metrics is not None and len(add_metrics) > 0:
+            if not isinstance(add_metrics, dict):
+                raise TypeError(
+                    f"additional_metrics must be None or a dict but is {type(additional_metrics).__name__}."
```
wyli, line 150: undefined variable `additional_metrics`, there are similar issues detected here: https://deepsource.io/gh/Project-MONAI/MONAI/run/6757e0d3-951f-4bec-9fc4-e24a51982044/python/

wyli (commented):
thanks please see some comments inline, we'll discuss this next monday

monai/engines/workflow.py
Nic-Ma: Thanks for your tips, fixed in the latest commit.

monai/engines/workflow.py
Nic-Ma: Sorry for my mistake, this callback should be in the `_register_metrics()` function instead of handlers, because it depends on the `state.best_metric` and `state.best_metric_epoch` which only exist in MONAI workflows, and should only work when metrics registered.
I updated this part in the latest commit.

Thanks.

monai/engines/utils.py
Nic-Ma: Sure, updated in the latest commit.
Thansk.

monai/engines/trainer.py
Nic-Ma: Hi @wyli ,

You are right, we don't suppose to use `Handler` for any cases, just want to cover most common cases and recommend users to use below 3 levels API:
```
1. Handlers based on rich Events -- try to cover most of the common cases, that's what this PR want to do.
2. iteration_update -- write own arbitrary iteration logic.
3. Extend SupervisedTrainer -- Do the most advanced features that current SupervisedTrainer can't satisfy.
```
For your example, user can write their own `optimizer` or `iteration_update`.

Thanks.

Nic-Ma:
Hi @wyli ,

Thanks for your review.

Thanks.

monai/engines/trainer.py
```diff
@@ -147,23 +151,32 @@ def _iteration(self, engine: Engine, batchdata: Dict[str, torch.Tensor]):
             kwargs: Dict = {}
         else:
             inputs, targets, args, kwargs = batch
+        # put iteration outputs into engine.state
+        engine.state.output = output = {Keys.IMAGE: inputs, Keys.LABEL: targets}
```
wyli, line 155: the `_iteration` method is extended to support additional events (such as FORWARD_COMPLETED). line 155 updates the output dictionary `engine.state.output` before this method returns. is this a reasonable usage of ignite? @vfdev-5 could you please help review, many thanks!

vfdev-5: @wyli Yes, you are correct about pointing out that: `engine.state.output` will be overwritten when `_iteration` will return `output` again. Here, however as far as I understand the idea is to make `engine.state.output` accessible in handlers attached to your custom events: `IterationEvents.FORWARD_COMPLETED`, `IterationEvents.LOSS_COMPLETED` etc.
The same idea can be also realized by passing directly the data as : 
```python
def _iteration(...):

    output[Keys.PRED] = self.inferer(inputs, self.network, *args, **kwargs)
    engine.fire_event(IterationEvents.FORWARD_COMPLETED, output) 
    output[Keys.LOSS] = self.loss_function(output[Keys.PRED], targets).mean()
    engine.fire_event(IterationEvents.LOSS_COMPLETED, output)
...
def on_forward_completed(engine, output):
    batch = engine.state.batch  # your input batch
    assert Keys.PRED in output
    ...

def on_loss_completed(engine, y_pred, loss):
    batch = engine.state.batch  # your input batch
    assert Keys.PRED in output
    assert Keys.LOSS in output
    ...
```
In this case, you do not need to write `output` to the state knowing that it will be overwritten in the end...

Nic-Ma: Hi @vfdev-5 ,

We added more events in the iteration, and want to **unify the data sharing** with handlers through `engine`.
For example, if user want to add some additional operations to `predictions` before loss computation and also add the noise to `engine.state.output`:
```py
class AddNoise:
    def attach(self, engine):
        engine.add_event_handler(Events.FORWARD_COMPLETED, self)

    def __call__(self, engine):
        noise = torch.rand(...)
        engine.state.output[Keys.PREDS] += noise
        engine.state.output[noise] = noise
```
And I think `engine.state.output` will not be overwritten at the end of this iteration as we return this `output` directly.
Your example code also can work for this case, but user needs to remember **different args for different event callbacks** in the handler, I feel that's not a unified API, so I want to always pass `engine` as handler arg, it's much easier for me to explain the API usage to users.
I am writing a detailed document to help users quickly understand how to use MONAI workflows which are based on ignite. Users who never used ignite complained many questions about the data-sharing part, especially not clear "what is exactly stored in the engine?" "what can I get in handler callback?", etc. We need to define a standard data-sharing method instead of ad-hoc solution for every single program.

Thanks.

vfdev-5: Thanks the explanation! Yeah, makes sense to have a unified API. Just a nit with this approach is that user also should remember what is already added to `state.output`.

> And I think engine.state.output will not be overwritten at the end of this iteration as we return this output directly.

Yes, by overwritten I wanted to say that the ref to `output` will be reassigned in the end to `output`... Anyway, yes, it is possible to do that.

> Users who never used ignite complained many questions about the data-sharing part, especially not clear "what is exactly stored in the engine?" "what can I get in handler callback?", etc. We need to define a standard data-sharing method instead of ad-hoc solution for every single program.

This is very interesting feedback! Could you please share more feedback on the API such that we could improve our docs 👍

Nic-Ma: Hi @vfdev-5 ,

Thanks for your comments.
I am writing the document to make a standard data-sharing method based on `ignite.engine`, and will make a sheet to describe every element in `engine` and especially the items in `engine.state`, then users just need to check this sheet to read/write expected data in their handlers or add their own data to `engine.state`.
And we mainly recommend users to write their own handlers to do **customized logic** if the built-in `SupervisedTrainer` and existing handlers can't satisfy their reuqests.
The main purpose of this design to make the `data-sharing` through workflows much more straight-forward.
Sure, I will share the document in MONAI pages when it's ready.

Thanks.

wyli: thanks for the discussion! still I feel setting `engine.state.output` twice (within `_iteration` and as the return value of `_iteration`) is a bit confusing and looks like a workaround solution. but I also could find a better way without making large changes. so I'll approve this PR for now.

Hi Nic, perhaps we should share the document in https://github.com/Project-MONAI/MONAI/tree/master/docs/source as it's important

wyli (approved):
thanks, this PR adds a few events to each iteration. just wanted to note that `state.output` is used to cache information across those events. not sure if this is the best way of using `state.output` but seems like a viable workaround according to @Nic-Ma

Nic-Ma:
Thanks @wyli ,

And @vfdev-5 , as we discussed in the above thread, I think `saving data to engine.state.output` is acceptable solution so far, referring to ignite conrib: https://github.com/pytorch/ignite/blob/master/ignite/contrib/engines/tbptt.py#L111.
But please let me know if this design is against your ignite design or any future plan, then we can update it.

Thanks.

PR #1489, opened 2021-01-22
CopyToDevice transform

rijobro:
Will enable (dictionary example given, array is the same):

```
transforms = Compose([
    LoadImaged(keys=["image","label"]),
    ToTensord(keys=["image","label"]),
    CopyToDeviced(keys=["image","label"], device=device),
])

for batch_data in train_loader:
    im, label = batch_data["image"], batch_data["label"]
```
removing the need for:
```
for batch_data in train_loader:
    im, label = batch_data["image"].to(device), batch_data["label"].to(device)
```

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/__init__.py
+++ b/monai/transforms/__init__.py
@@ -254,6 +254,7 @@ from .utility.array import (
     TorchVision,
     ToTensor,
     Transpose,
+    CopyToDevice,
 )
 from .utility.dictionary import (
     AddChanneld,
@@ -316,6 +317,9 @@ from .utility.dictionary import (
     ToTensord,
     ToTensorD,
     ToTensorDict,
+    CopyToDeviced,
+    CopyToDeviceD,
+    CopyToDeviceDict,
 )
 from .utils import (
     apply_transform,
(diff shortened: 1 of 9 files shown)


Nic-Ma:
Hi @rijobro ,

Thanks for your good idea.
I have a question In my simple mind: if moving data to GPU in a transform in every process, it will batch together in DataLoader later and the batch data is not contiguous in GPU memory, maybe the performance is worse than "moving to GPU" after batching? Just curious, maybe you can provide some test results to compare later.

Thanks.

rijobro:
Thanks @Nic-Ma I hadn't thought of that. I'll have a look!

rijobro:
Hi @Nic-Ma was looking into this, but it seems it's a bad idea, since each of the subprocesses from the DataLoader would have to share GPU memory across processes. See [here](https://discuss.pytorch.org/t/runtimeerror-cannot-re-initialize-cuda-in-forked-subprocess-to-use-cuda-with-multiprocessing-you-must-use-the-spawn-start-method/14083) and [here](https://discuss.pytorch.org/t/not-using-multiprocessing-but-getting-cuda-error-re-forked-subprocess/54610).

Hence, I'll close this PR.

wyli:
but I think `copy_to_device` is still a useful utility, isn't it?

rijobro:
that functionality still exists in the lr_finder PR, I could split it into its own PR if you'd like. But I think if we're about to merge the lr_finder PR, there's not much point.

PR #1490, opened 2021-01-22
remove decorator double negative

rijobro:
Tiny PR, just removes double negative from test decorators.

### Status
**Ready**

### Types of changes


wyli (approved):
thanks!

PR #1509, opened 2021-01-26
Implement FROC metric

yiheng-wang-nv:

Implement FROC metric which is used in https://camelyon16.grand-challenge.org/Evaluation/.

### Status
**Hold**

### Types of changes


yiheng-wang-nv:
Hi @behxyz , in this MR, I remove the part of compute evaluation mask and ITC, since they are only focusing on the challenge based data format, I'm not sure if they should be added into MONAI. Could you please help to review this MR and if needed, let's discuss about the next step for the implementation. Thanks!

behxyz (changes requested):
@yiheng-wang-nv Very neat code, I liked it! I proposed few changes. On the other hand, we should think of what else is needed to make this metric usable in MONAI, and maybe starting from the pathology use case. Any thought?

monai/metrics/froc.py
```diff
@@ -0,0 +1,121 @@
...
+    for thresh in all_probs[1:]:
+        total_fps.append((fp_probs >= thresh).sum())
+        total_tps.append((tp_probs >= thresh).sum())
+    total_fps.append(0)
+    total_tps.append(0)
+    total_fps = np.asarray(total_fps) / float(num_of_samples)
+    total_sensitivity = np.asarray(total_tps) / float(num_of_tumors)
+
+    interp_sens = np.interp(eval_thresholds, total_fps[::-1], total_sensitivity[::-1])
+    return np.mean(interp_sens)
```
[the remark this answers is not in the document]
yiheng-wang-nv: Thanks, I agree.

yiheng-wang-nv:
Hi @Nic-Ma @wyli , I met an error like this:
```
    if torch.is_tensor(probs):
        probs = probs.detach().cpu().numpy()
```
The above part got the mypy error:
`monai/metrics/froc.py:52:17: error: Item "ndarray" of "Union[ndarray, Tensor]" has no attribute "detach"  [union-attr]`
However, it used mypy 0.8, which is the same as my local environment, and I did not get this error locally. Could you please help to take a look at this issue? Thanks!

Nic-Ma:
> Hi @Nic-Ma @wyli , I met an error like this:
> 
> ```
>     if torch.is_tensor(probs):
>         probs = probs.detach().cpu().numpy()
> ```
> 
> The above part got the mypy error:
> `monai/metrics/froc.py:52:17: error: Item "ndarray" of "Union[ndarray, Tensor]" has no attribute "detach" [union-attr]`
> However, it used mypy 0.8, which is the same as my local environment, and I did not get this error locally. Could you please help to take a look at this issue? Thanks!
> 
> The code is in [here](https://github.com/Project-MONAI/MONAI/pull/1509/files#diff-a59bbef295e948153e12e70ae60cc159cc24be1ed37ee08c1131fc3eaf74500aR51).

2 points:
1. you can try to use `isinstance(probs, torch.Tensor)` instead of `torch.is_tensor()` and try to submit PR again.
2. The reason why your local test is OK: your numpy is old, if you update to the latest numpy, you will get the error in mypy.

Thanks.

yiheng-wang-nv:
> > Hi @Nic-Ma @wyli , I met an error like this:
> > ```
> >     if torch.is_tensor(probs):
> >         probs = probs.detach().cpu().numpy()
> > ```
> > 
> > 
> > The above part got the mypy error:
> > `monai/metrics/froc.py:52:17: error: Item "ndarray" of "Union[ndarray, Tensor]" has no attribute "detach" [union-attr]`
> > However, it used mypy 0.8, which is the same as my local environment, and I did not get this error locally. Could you please help to take a look at this issue? Thanks!
> > The code is in [here](https://github.com/Project-MONAI/MONAI/pull/1509/files#diff-a59bbef295e948153e12e70ae60cc159cc24be1ed37ee08c1131fc3eaf74500aR51).
> 
> 2 points:
> 
> 1. you can try to use `isinstance(probs, torch.Tensor)` instead of `torch.is_tensor()` and try to submit PR again.
> 2. The reason why your local test is OK: your numpy is old, if you update to the latest numpy, you will get the error in mypy.
> 
> Thanks.

Thanks Nic,

My running environment is the latest MONAI docker, and the numpy version is 1.19.1, I upgraded it into 1.19.5 but still cannot get the same error.

wyli:
the latest numpy 1.20 doesn't support python 3.6 which is the default python of the docker image...

monai/metrics/froc.py
```diff
@@ -0,0 +1,137 @@
...
+import numpy as np
+import torch
+
+
+def compute_fp_tp_probs(
+    probs: Union[np.ndarray, torch.Tensor],
+    ycorr: Union[np.ndarray, torch.Tensor],
+    xcorr: Union[np.ndarray, torch.Tensor],
+    evaluation_mask: Union[np.ndarray, torch.Tensor],
+    isolated_tumor_cells: Optional[List] = None,
```
wyli, line 24: this `isolated_tumor_cells` option is too task-specific, could be `target_labels`?

monai/metrics/froc.py
```diff
@@ -0,0 +1,137 @@
...
+            tp_probs[i - 1] = probs[np.where(hittedlabel == i)].max()
+
+    num_of_tumors = max_label - len(isolated_tumor_cells)
+    return fp_probs, tp_probs, num_of_tumors
+
+
+def compute_froc_curve_data(
+    fp_probs: Union[np.ndarray, torch.Tensor],
+    tp_probs: Union[np.ndarray, torch.Tensor],
+    num_of_tumors: int,
```
wyli, line 82: num_of_tumors -> num_of_targets

monai/metrics/froc.py
```diff
@@ -0,0 +1,137 @@
...
+    num_of_samples: int,
```
wyli, line 83: it's unclear what `num_of_samples` is. The docstring should be more generic... @behxyz could you please help update the descriptions here?

wyli (commented):
thanks, I put some comments on the docstring and variable names

monai/metrics/froc.py
behxyz: if we change `sample` to `image`, it would be self-explanatory, e.g. `num_images`

monai/metrics/froc.py
```diff
@@ -0,0 +1,137 @@
...
+):
+    """
+    This function is modified from the official evaluation code of
+    `CAMELYON 16 Challenge <https://camelyon16.grand-challenge.org/>`_, and used to compute
+    the required data for plotting the Free Response Operating Characteristic (FROC) curve.
+
+    Args:
+        fp_probs: an array that contains the probabilities of the false positive detections for all
```
behxyz, line 91: I believe it would be clearer to change:
"for all samples that to be computed" --> "for all images under evaluation"

behxyz (changes requested):
Most of the suggestions are related to name changing to make it more generic and not specific to pathology use cases. The general flow seems to be generic enough but I need to find a couple of use cases to verify. I will add my comments once I find the use cases to check.

behxyz (approved):
It looks good to me. Well done, @yiheng-wang-nv! 
After this PR, it's time to make this FROC work for pathology inference pipeline.

Issue #1512, opened 2021-01-27
Shape mismatch ValueError when using RandGaussianNoised with more than one key


LucasFidon:
**Describe the bug**
When using RandGaussianNoised during training with more than one key the error "ValueError: shape mismatch: objects cannot be broadcast to a single shape" is returned.

**To Reproduce**
I have encountered the error when using RandGaussianNoised in my training pipeline for BraTS.

I have in my composition of training transforms:
`RandGaussianNoised(keys=["flair", "t1", "t1ce", "t2"], mean=0., std=0.1, prob=0.15),`

For this example, in https://github.com/Project-MONAI/MONAI/blob/8207e1e2a3555ddc3fe938e058552651900dc951/monai/transforms/intensity/dictionary.py#L113
we obtain `self.mean = (0.,0,0,0)`.

This creates a mismatch between the shape of `self.mean` and `im_shape` in https://github.com/Project-MONAI/MONAI/blob/8207e1e2a3555ddc3fe938e058552651900dc951/monai/transforms/intensity/dictionary.py#L120
where `im_shape=(1,128,128,128)`.

**A Simple Solution**
By replacing https://github.com/Project-MONAI/MONAI/blob/8207e1e2a3555ddc3fe938e058552651900dc951/monai/transforms/intensity/dictionary.py#L113
by `self.mean = mean` the error does not appear anymore.

**Expected behavior**
The reason why the tests pass with the current version of the code is that the test https://github.com/Project-MONAI/MONAI/blob/master/tests/test_rand_gaussian_noised.py only contains examples with `len(keys)==1`.

In addition, in https://github.com/Project-MONAI/MONAI/blob/8207e1e2a3555ddc3fe938e058552651900dc951/monai/transforms/intensity/dictionary.py#L131-L133
the same sample of Gaussian noise is used for all the keys.
I would have expected that different noise samples would be used with different keys.
Is it the intended behavior for this transformation?

I can work on a pull request if you can confirm which behavior you would like to see here.

Thank you,
Lucas

saruarlive:
@LucasFidon,
Based on the bug, so it handles a single dictionary key, right!
If you stack ["flair", "t1", "t1ce", "t2"] images in one tensor, (4, :, : , : ) before, then use the RandGaussianNoised class using one dictionary key, does it raise error then?

LucasFidon:
Hi @saruarlive 
Yes, that seems to be a good workaround.
[...]

PR #1530, opened 2021-02-01
1512 rand_gaussian_noised with multiple keys

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/1512.

### Description
Create a list of noise realisations, with length equal to the number of keys. The noise realisations can all use the same mean, or a mean can be specified for each of the keys. The unit test has been updated to run for multiple labels.

@LucasFidon Does this suitably solve your problem? 

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/intensity/dictionary.py
+++ b/monai/transforms/intensity/dictionary.py
@@ -35,7 +35,7 @@ from monai.transforms.intensity.array import (
     ShiftIntensity,
     ThresholdIntensity,
 )
-from monai.utils import dtype_torch_to_numpy, ensure_tuple_size
+from monai.utils import dtype_torch_to_numpy, ensure_tuple_rep, ensure_tuple_size
 
 __all__ = [
     "RandGaussianNoised",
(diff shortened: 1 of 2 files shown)


LucasFidon:
Hi @rijobro 
Yes, this solves #1512, thank you for looking at this!

monai/transforms/intensity/dictionary.py
```diff
@@ -110,27 +110,29 @@ def __init__(
     ) -> None:
         super().__init__(keys)
         self.prob = prob
-        self.mean = ensure_tuple_size(mean, len(self.keys))
+        self.mean = ensure_tuple_size(mean, len(self.keys), mean)
```
Nic-Ma, line 113: I think for this case we should use `ensure_tuple_rep(mean, len(self.keys))`, right?
For example: https://github.com/Project-MONAI/MONAI/blob/master/monai/transforms/utility/dictionary.py#L294

Thanks.

rijobro: Yup, thanks!

wyli (approved):
thanks!

PR #1535, opened 2021-02-01
1533 Fix distributed data parallel issue in ClassificationSaver

### LINKED ISSUE #1533: ClassificationSaver can't save all the data from ranks into CSV file
**Describe the bug**
When doing distributed training, `ClassificationSaver` runs on different ranks and write to the same CSV file, it causes a random output in the file and sometimes even only 1 rank can open the file.

**Expected behavior**
Expect `all_gather` image path and model output data, then only write CSV file in the specified rank.

Nic-Ma:
Fixes #1533 .

### Description
This PR fixed the file saving issue of ClassificationSaver in distributed data parallel mode.

### Status
**Ready**

### Types of changes


monai/handlers/classification_saver.py
```diff
@@ -77,6 +85,12 @@ def __call__(self, engine: Engine) -> None:
         Args:
             engine: Ignite Engine, it can be a trainer, validator or evaluator.
         """
-        meta_data = self.batch_transform(engine.state.batch)
-        engine_output = self.output_transform(engine.state.output)
-        self.saver.save_batch(engine_output, meta_data)
+        _meta_data = self.batch_transform(engine.state.batch)
+        if "filename_or_obj" in _meta_data:
```
ericspod, line 89: The string "filename_or_obj" is used in a number of places here and in other code. This should be stored somewhere as a constant value with documentation stating what it's used to store.

monai/handlers/classification_saver.py
```diff
@@ -67,7 +75,7 @@ def attach(self, engine: Engine) -> None:
             self.logger = engine.logger
         if not engine.has_event_handler(self, Events.ITERATION_COMPLETED):
             engine.add_event_handler(Events.ITERATION_COMPLETED, self)
-        if not engine.has_event_handler(self.saver.finalize, Events.COMPLETED):
+        if self._expected_rank and not engine.has_event_handler(self.saver.finalize, Events.COMPLETED):
             engine.add_event_handler(Events.COMPLETED, lambda engine: self.saver.finalize())
```
ericspod, line 79: Should () be at the end of `self.saver.finalize()`?

ericspod (commented):
I mentioned some small things only. According to the code coverage output the tests aren't being run under multi GPU with Pytorch 1.7 so `string_list_all_gather` isn't being tested but it looks ok.

monai/handlers/classification_saver.py
Nic-Ma: Hi @ericspod ,

The handler callback registered here is a lambda function: `lambda engine: self.saver.finalize()`, not the `finalize` function itself.

Thanks.

ericspod: Oops I missed that!

ericspod (approved):

monai/handlers/classification_saver.py
Nic-Ma: Thanks for your suggestion, let me change to some `constant` value.

Nic-Ma: updated in the latest commit.
Thanks.

wyli:
@ericspod the code coverage is inaccurate in our current setting as the multiprocess executions are not tracked properly...I'll create an issue

PR #1548, opened 2021-02-04
WSI reader

### LINKED ISSUE #1504: WSI image reader
Whole slide imaging (WSI) in pathology generates huge images which makes is inefficient to load the whole image into memory and extract a small patch for training.

There are libraries that are able to more efficiently extract and load small patches into memory from very large images. OpenSlide is the most popular and CuClaraImage is another under-development library. We should have a WSIreader that uses either of them under the hood.

behxyz:
Fixes #1504

### Description
Training models for pathology use cases requires loading patches from Whole Slide Imaging (WSI) scans. These images are enormous in size and usually much larger than user's available RAM (#1504). This PR provide a solution by implementing whole slide image readers for CuImage and OpenSlide.

### Types of changes


behxyz:
@wyli, @Nic-Ma 
The `mypy` [complains](https://github.com/Project-MONAI/MONAI/pull/1548/checks?check_run_id=1851266513#step:7:285) are for the existing code and I think they are due to the third-party libraries update. For instance, the first one is about `np.eye`, which I suspect it is because of the interface file ( `__init__.pyi`) that has been added after `numpy==1.20` where the return type of `np.eye` is defined as `Any`. Have you encounter this in any other place in MONAI?

```
monai/data/image_reader.py:252:9: error: Returning Any from function declared to return "ndarray"  [no-any-return]
286
monai/data/image_reader.py:280:13: error: Returning Any from function declared to return "ndarray"  [no-any-return]
287
monai/data/image_reader.py:393:9: error: Returning Any from function declared to return "ndarray"  [no-any-return]
```

behxyz:
> @wyli, @Nic-Ma
> The `mypy` [complains](https://github.com/Project-MONAI/MONAI/pull/1548/checks?check_run_id=1851266513#step:7:285) are for the existing code and I think they are due to the third-party libraries update. For instance, the first one is about `np.eye`, which I suspect it is because of the interface file ( `__init__.pyi`) that has been added after `numpy==1.20` where the return type of `np.eye` is defined as `Any`. Have you encounter this in any other place in MONAI?
> 
> ```
> monai/data/image_reader.py:252:9: error: Returning Any from function declared to return "ndarray"  [no-any-return]
> 286
> monai/data/image_reader.py:280:13: error: Returning Any from function declared to return "ndarray"  [no-any-return]
> 287
> monai/data/image_reader.py:393:9: error: Returning Any from function declared to return "ndarray"  [no-any-return]
> ```

@wyli @Nic-Ma, what do you suggest here?

monai/data/image_reader.py
```diff
@@ -244,7 +249,7 @@ def _get_affine(self, img) -> np.ndarray:
         affine = np.eye(direction.shape[0] + 1)
```
ericspod, line 249: Here if you declare `affine` to be type `np.ndarray` (ie. `affine: np.ndarray = np.eye(...)` which looks unnecessary) does this sort the mypy issue?

behxyz:
I changed the scope of this PR to whole slide image reader only.

monai/data/image_reader.py
```diff
@@ -24,21 +24,25 @@
 from .utils import is_supported_format
 
 if TYPE_CHECKING:
+    import cuimage
     import itk  # type: ignore
     import nibabel as nib
+    import openslide
```
wyli, line 30: please update the package requirements in `requirements-dev.txt`, `docs/requirements.txt`, `setup.cfg`

tests/test_openslide_reader.py
```diff
@@ -0,0 +1,89 @@
...
+    @skip_if_quick
+    def test_read_patches(self, filename, patch_info, expected_img):
+        self.camelyon_data_download(filename)
+        reader = WSIReader("OpenSlide")
+        img_obj = reader.read(filename)
+        img = reader.get_data(img_obj, **patch_info)
+        self.assertTupleEqual(img.shape, expected_img.shape)
+        self.assertIsNone(assert_array_equal(img, expected_img))
+
+    def camelyon_data_download(self, filename):
```
wyli, line 77: is it possible to embed a few artificial tiny images as test cases instead? ideally within a few kilobytes

wyli (commented):
thanks, please see some initial comments inline, I'll try it out with an EA container soon...

tests/test_openslide_reader.py
behxyz: I haven't found a tiny tiff file that can test WSIReader properly. However, I found a smaller histopathology TIFF file (~200MB) with a faster download (~30s).

monai/data/image_reader.py
```diff
@@ -626,3 +630,143 @@ def _get_spatial_shape(self, img) -> np.ndarray:
...
+        level: int = 0,
+        dtype: DtypeLike = np.uint8,
+    ):
+        # reverse the order of dimensions for size and location to be compatible with image shape
+        size = size[::-1]
+        location = location[::-1]
+        region = img_obj.read_region(location=location, size=size, level=level)
+        if self.wsi_reader_name == "openslide":
+            region = region.convert("RGB")
+        # convert to numpy
```
Nic-Ma, line 736: I would suggest to convert the data shape in the following `EnsureChannelFirst` transform later, don't hard-code here. @wyli @rijobro , what do you think?

monai/data/image_reader.py
```diff
@@ -626,3 +630,143 @@ def _get_spatial_shape(self, img) -> np.ndarray:
...
+        region = np.asarray(region, dtype=dtype)
+        # cuCalaraImage/OpenSlide: (H x W x C) -> torch image: (C X H X W)
+        region = region.transpose((2, 0, 1))
+        return region
```
Nic-Ma, line 740: Could you please help also add the meta data? Refer to the PILReader: https://github.com/Project-MONAI/MONAI/blob/master/monai/data/image_reader.py#L598

monai/data/image_reader.py
```diff
@@ -626,3 +630,143 @@ def _get_spatial_shape(self, img) -> np.ndarray:
...
+                img.shape = (img.dimensions[1], img.dimensions[0], 3)
+            img_.append(img)
+
+        return img_ if len(filenames) > 1 else img_[0]
+
+    def get_data(
+        self,
+        img_obj,
+        location: Tuple[int, int] = (0, 0),
+        size: Optional[Tuple[int, int]] = None,
```
Nic-Ma, line 687: Suggest to use `spatial_size`, otherwise, it may be confusing whether it contains `channel` dim.

Nic-Ma (commented):
Thanks for your quick update.
I put several minor comments inline, others look good to me.
And please add this new object to the "__init__.py" of `mode/data/` and docs/sources/data.

Thanks.

monai/data/image_reader.py
wyli: this channel last format is a very common default, perhaps we could embed an `EnsureChannelFirst` here (or for all the readers)? otherwise the user almost always need to add an `EnsureChannelFirst`

Nic-Ma: Hi @wyli ,

Interesting idea, we can do it later as it's a non-breaking feature, `EnsureChannelFirst` transform will not cause an error if the data is already channel-first in readers.

Thanks.

Nic-Ma: But for current PR, I think it should not change the data shape here.
If we want to use `EnsureChannelFirst`, we should apply in the base class.

Thanks.

behxyz: I think for any reader that we are implementing, we should make sure that it is channel first, to follow the convention throughout the readers.

monai/data/image_reader.py
behxyz: metadata from OpenSlide and cuCalraImage are not structured in the same way so they results in a very different output which makes the output of WSIReader depend on the library to use. I believe it would be better to have a separate PR for that once the development of cuClaraImage become stable.

monai/data/image_reader.py
behxyz: I kind of agree here but since both OpenSlide and cuClaraImage use `size` to define this input argument, I found using another name confusing for people who has ML in pathology background and has worked with WSI in the past.

monai/data/image_reader.py
```diff
@@ -626,3 +630,143 @@ def _get_spatial_shape(self, img) -> np.ndarray:
...
+            data: file name or a list of file names to read.
+
+        """
+        img_: List = []
+
+        filenames: Sequence[str] = ensure_tuple(data)
+        for name in filenames:
+            img = self.wsi_reader(name)
+            if self.wsi_reader_name == "openslide":
+                img.shape = (img.dimensions[1], img.dimensions[0], 3)
```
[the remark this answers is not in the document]
wyli: thanks, the doc suggests RGBA (https://openslide.org/api/python/#openslide.OpenSlide.associated_images), I think we should make it clear in the docstring if we only support RGB

monai/data/image_reader.py
Nic-Ma: Hi @behxyz ,

I agree to add well-structured metadata later, but for this very first version, would be useful to add at least the `spatial_shape` and `original_channel_dim`. Some other transforms or handlers may need this metadata for additional logic.
You can refer to the PILReader: https://github.com/Project-MONAI/MONAI/blob/master/monai/data/image_reader.py#L599
And don't transpose axis here, use `EnsureChannelFirst` transform in the transform chain instead.
If we decide to make all the readers output `channel-first`, I will submit another PR to support it in the reader base class.
Others of this PR look good to me now.

Thanks very much.

behxyz: @Nic-Ma,
giving more thought into this, I found some confusion here. When we are talking about `spatial_shape` and `original_channel_dim`, is it about the output image (which is a subregion of WSI) or the whole slide image?

What is the use for this? Because if it is about the output image, in this case `original_channel_dim` is meaningless because since it depends on the how we are extracting it, and if it is out the original whole slide image, `spatial_shape` becomes useless.

behxyz: For now, I put the `spatial_shape` and `original_channel_dim` for extracted region, which makes the most sense.

Nic-Ma (approved):
Looks good to me.
Thanks for your quick update!

PR #1561, opened 2021-02-08
Prepare Batch

behxyz:
Extend default prepare batch in Trainer

### Description
This PR implements the abstract `PrepareBatch` and concrete classes:
- PrepareBatchDefault
- PrepareBatchExtraInput
- PrepareBatchShuffle

### Status
Ready

### Types of changes

### DIFF
--- a/monai/engines/prepare_batch.py
+++ b/monai/engines/prepare_batch.py
@@ -1,4 +1,3 @@
-
 from abc import ABC, abstractmethod
 from typing import Dict, Optional, Sequence, Tuple, Union
 
@@ -21,6 +20,7 @@ class PrepareBatch(ABC):
         otherwise, it should update the batchdata dict and return it directly.
 
     """
+
     @abstractmethod
     def __call__(
         self,
@@ -38,6 +38,7 @@ class PrepareBatchDefault(PrepareBatch):
     it should be the last component in the prepare_batch chain.
 
     """
+
     def __call__(
         self,
         batchdata: Dict[str, torch.Tensor],
@@ -60,6 +61,7 @@ class PrepareBatchExtraInput(PrepareBatch):
             dict to the network(**kwargs).
 
     """
+
     def __init__(self, extra_keys: Union[str, Sequence[str], Dict[str, str]]) -> None:
         self.extra_keys = extra_keys
 
@@ -100,8 +102,7 @@ class PrepareBatchShuffle(PrepareBatch):
         non_blocking: bool = False,
     ) -> Union[Tuple[torch.Tensor, Optional[torch.Tensor]], torch.Tensor]:
 
-        images = batchdata["image"].to(device=device, non_blocking=non_blocking)
-        labels = batchdata["label"].to(device=device, non_blocking=non_blocking)
+        images, labels = default_prepare_batch(batchdata, device, non_blocking)
         idx_rand = torch.randperm(len(labels)).to(device=device, non_blocking=non_blocking)
 
         return images[idx_rand], labels[idx_rand]
--- a/monai/engines/prepare_batch.py
+++ b/monai/engines/prepare_batch.py
@@ -101,8 +101,9 @@ class PrepareBatchShuffle(PrepareBatch):
         device: Optional[Union[str, torch.device]] = None,
         non_blocking: bool = False,
     ) -> Union[Tuple[torch.Tensor, Optional[torch.Tensor]], torch.Tensor]:
-
         images, labels = default_prepare_batch(batchdata, device, non_blocking)
-        idx_rand = torch.randperm(len(labels)).to(device=device, non_blocking=non_blocking)
+        idx_rand = torch.randperm(len(images)).to(device=device, non_blocking=non_blocking)
 
-        return images[idx_rand], labels[idx_rand]
+        if labels is not None:
+            return images[idx_rand], labels[idx_rand]
+        return images[idx_rand], None
--- a/monai/engines/prepare_batch.py
+++ b/monai/engines/prepare_batch.py
@@ -1,0 +1,107 @@
+
+from abc import ABC, abstractmethod
+from typing import Dict, Optional, Sequence, Tuple, Union
+
+import torch
+
+from ..utils import ensure_tuple
+from .utils import default_prepare_batch
+
+
+class PrepareBatch(ABC):
+    """
+    Base class of customized prepare_batch in the trainer or evaluator workflows.
+    It takes the data of current batch, target device and non_blocking flag as input.
+    Default to run the default_prepare_batch of MONAI, which is consistent with ignite:
+    https://github.com/pytorch/ignite/blob/v0.4.2/ignite/engine/__init__.py#L28.
+
+    Note: if it is designed to be the last component in the prepare_batch chain,
+        it should return either (image, label(optional)) or (image, label(optional), tuple, dict).
+        the `tuple` and `dict` here are the `*args **kwargs` parameters for the network.
+        otherwise, it should update the batchdata dict and return it directly.
+
+    """
+    @abstractmethod
+    def __call__(
+        self,
+        batchdata: Dict[str, torch.Tensor],
+        device: Optional[Union[str, torch.device]] = None,
+        non_blocking: bool = False,
+    ):
+        raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
+
+
+class PrepareBatchDefault(PrepareBatch):
+    """
+    Default prepare_batch method to return `image` and `label` only,
+    it's consistent with MONAI `default_prerpare_batch` API.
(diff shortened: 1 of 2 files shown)


behxyz:
I'll update this PR to make them all handlers.

Nic-Ma:
Thanks @behxyz , handler is existing API in MONAI and ignite, and very straight forward.
Recommend to use it for most cases.

Thanks.

behxyz:
The [unittests](https://github.com/Project-MONAI/MONAI/pull/1561/checks?check_run_id=1856851001#step:10:7762), involving randomness, are failing despite `set_determinism(0)`. I have tested on several machines and docker images (ubuntu and macos) with different version of pytorch and all worked fine. Any thought why it is failing here? @wyli @Nic-Ma

wyli:
I think there's a bug with the global random state, before that's confirmed/fixed, please use the `setUp` and `tearDown` to set seed like https://github.com/Project-MONAI/MONAI/blob/db831ae096bec76371bb8c2e3db226d947a5ac06/tests/test_bounding_rectd.py#L28

Nic-Ma:
Hi @wyli and @behxyz ,

I think maybe it's better to implement the `shuffle` feature with ignite handler instead of `batch_prepare`.
handler is more flexible and easier to investigate with other components, some draft code:
```py
from typing import TYPE_CHECKING, Dict, Optional
import torch
from monai.utils import exact_version, optional_import, ensure_tuple
from monai.config import KeysCollection
Events, _ = optional_import("ignite.engine", "0.4.2", exact_version, "Events")
if TYPE_CHECKING:
    from ignite.engine import Engine
else:
    Engine, _ = optional_import("ignite.engine", "0.4.2", exact_version, "Engine")


class PatchShuffler:
    """
    When every Dataset generates a list of image patches, DataLoader only shuffles at high level.
    This handler is used to totally shuffle all the patches for a iteration.
    Expected input data should be dictionary.

    """
    def __init__(self, keys: KeysCollection):
        self.keys = ensure_tuple(keys)

    def attach(self, engine: Engine) -> None:
        engine.add_event_handler(Events.ITERATION_STARTED, self)

    def __call__(self, engine: Engine) -> None:
        idx_rand = torch.randperm(length(engine.state.batch[self.keys[0]]))
        for k in self.keys:
            engine.state.batch[k] = engine.state.batch[k][idx_rand]
```

What do you think?

Thanks.

behxyz:
I think implementing prepare_batch as handlers could be confusing. Since the workflow already have prepare_batch and anyways it will call the default one on the batches which convert them from dict to tuple. Now if we make a prepare batch handler as well. First of all, we should check and see if it is called before or after self.prepare_batch. Second of all, anyways it changes the input/output type of prepare_batch and we cannot reuse them in their original place. Third of all, if it is being called after self.prepare_batch, it will be tuple in/tuple out, and we can’t know what they are without checking the pre transforms.

wyli:
thanks @behxyz, this PR proposes to shuffle within each batch, could you provide a paper reference for this approach for clarity?

my impression of ignite is that we should make use of the event handling system whenever possible instead of expanding `prepare_batch`. 

@vfdev-5 could you please comment on this if you have time? our goal here is to shuffle a training batch before the model.forward(batch): https://github.com/behxyz/MONAI/blob/prepare_batch/monai/engines/prepare_batch.py#L104-L109  which is done as a `prepare_batch` callable. But it could also be implemented as a handler https://github.com/Project-MONAI/MONAI/pull/1561#issuecomment-775963186 which one is preferred from ignite's perspective?

vfdev-5:
@wyli thanks for mentioning! 

Yes, we provide a way to dynamically preprocess the input batch (please, see the note [here](https://pytorch.org/ignite/engine.html#ignite.engine.engine.Engine.run)) and this can be done with a configurable handler:
```python
trainer = ...

@trainer.on(Events.ITERATION_STARTED(custom_event_filtering))
def switch_batch(engine):
    engine.state.batch = preprocess_batch(engine.state.batch)
```
such that user could easily turn on/off this preprocessing or apply with a certain frequency...

Hope this could help to decide from your side.

vfdev-5:
I'm curious in which case it is interesting to shuffle things ? AFAIK in most of the cases, loss functions average across the batch dimension and this shouldn't change gradients, right ?

wyli:
> I'm curious in which case it is interesting to shuffle input batch ? AFAIK in most of the cases, loss functions average across the batch dimension and this shouldn't change gradients, right ?

I'd like to know more details as well :) @behxyz

behxyz:
That's good point! I believe this is what inherited from one of our pathology use cases, [here](https://github.com/baidu-research/NCRF/blob/a8657eed4ea105ec80bb1a003a5afb464e1ce99f/wsi/bin/train.py#L57). I am not convinced that this shuffling has any effect either.

vfdev-5:
There can be a practical use-case, for example, when using N devices and when a batch is sort of data unbalanced (e.g. [C1, C1, ..., C1, C2, C2]). Then we'd like to interleave data such that each process (and models with BN) sees more or less balanced data...

wyli:
yes, perhaps to avoid bias in a fixed data splitting in the multi gpu case

behxyz:
Right, especially because in that use case the the patches are extracted in a sequence from tumor or normal region, so although the whole batch is balances any sub-batch is highly imbalanced.

wyli:
I guess to do it properly there should be a buffering stage before model forwarding to allow for sufficient shuffling...

behxyz:
> ```python
> trainer = ...
> 
> @trainer.on(Events.ITERATION_STARTED(custom_event_filtering))
> def switch_batch(engine):
>     engine.state.batch = preprocess_batch(engine.state.batch)
> ```

@vfdev-5, @wyli 
In MONAI SupervisedTrainer:
https://github.com/Project-MONAI/MONAI/blob/8d70890a4b39486109912dfdd9a03e198c7c4736/monai/engines/trainer.py#L147
we already apply a `prepare_batch` function, so `engine.state.batch` is already a processed batch and the type is tuple here, right? or the handler is being triggered before this line?

behxyz:
> I guess to do it properly there should be a patch buffering stage before model.forward to allow for sufficient shuffling...

@wyli, could you elaborate on that?

wyli:
> > I guess to do it properly there should be a patch buffering stage before model.forward to allow for sufficient shuffling...
> 
> 
> 
> @wyli, could you elaborate on that?

yes, even if we shuffle at the image level, the patch level batch might be biased toward a specific subject at each iteration. we can construct a patch buffer and shuffle the buffer to mix patches from different images

behxyz:
> > > I guess to do it properly there should be a patch buffering stage before model.forward to allow for sufficient shuffling...
> > 
> > 
> > @wyli, could you elaborate on that?
> 
> yes, even if we shuffle at the image level, the patch level batch might be biased toward a specific subject at each iteration. we can construct a patch buffer and shuffle the buffer to mix patches from different images

Oh right! similar to having a shuffling buffer before batching images in other use cases.

Nic-Ma:
@wyli @behxyz , I think we are aligned on the direction of this PR now?
Thanks.

behxyz:
I have communicated with other researchers about this and they confirmed we already discussed. It shouldn't have any effect on single-gpu training but it is required for multi-gpu training because since the model has BN layers. We are going to explore replacing BN layers by SyncBN layers for our pathology use case but it requires tweaking any models that is being used.

However, in general, I believe shuffling batches is needed in MONAI for people who are using the torchvision models on multi-gpu with any systematic order of data in their batch. What do you think? @wyli @vfdev-5

vfdev-5:
@behxyz i'd say this could be helpful in ignite too :)

ericspod:
For randomized transforms we use [Randomizable](https://github.com/Project-MONAI/MONAI/blob/af1ffd60d377fa2fc3cf8c43e8057b1ac52559b4/monai/transforms/compose.py#L76) to provide a mechanism for setting the random state object. This is Numpy rather than Pytorch but I think the intent makes sense, we should have a way to set the random state for `PrepareBatchShuffle` or at least its seed separately from a global state as set with `set_determinism`. Should we also have an example and a test of using these with an actual Trainer object?

wyli:
> I guess to do it properly there should be a patch buffering stage before model.forward to allow for sufficient shuffling...

pytorch 1.8 comes with a `BufferedShuffleDataset` https://github.com/pytorch/pytorch/blob/v1.8.0/torch/utils/data/dataset.py#L257 perhaps could resolve this

behxyz:
> > I guess to do it properly there should be a patch buffering stage before model.forward to allow for sufficient shuffling...
> 
> pytorch 1.8 comes with a `BufferedShuffleDataset` https://github.com/pytorch/pytorch/blob/v1.8.0/torch/utils/data/dataset.py#L257 perhaps could resolve this

Thank you @wyli! I'll take a look into it.

PR #1514, opened 2021-01-28
[WIP] 365 Inverse spatial transforms

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/365.

Initial proposition for inverting deterministic spatial transforms.

### Status
**Work in progress**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh --codeformat --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

**To do**

- [x] Implement `inverse` method for a deterministic spatial transform (`SpatialPadd`)
- [x] Implement `inverse` for `Compose` class
- [x] Create serialised representation of transform (`append_applied_transforms` and `get_input_args`)
- [x] Implement for a random class (to check we can safely store random parameters)
- [x] Test with `CacheDataset` and `DataLoader` to check for multiprocessing problems
- [x] Check GPU vs CPU
- [x] Check numpy vs. tensor
- [x] Check inclusion of batch dimension
- [ ] Test with transforms that return lists (e.g., `RandSpatialCropSamplesd`)
- [ ] Add to `transforms.rst`

**Invertible classes**

- [x] Compose
- [x] SpatialPadd
- [x] BorderPadd
- [x] DivisiblePadd
- [x] SpatialCropd
- [x] CenterSpatialCropd
- [x] CropForegroundd
- [x] ResizeWithPadOrCropd
- [x] Spacingd
- [x] Orientationd
- [x] Rotate90d
- [x] Resized
- [x] Flipd
- [x] Rotated
- [x] Zoomd
- [x] RandSpatialCropd
- [ ] RandSpatialCropSamplesd
- [ ] RandWeightedCropd
- [ ] RandCropByPosNegLabeld
- [x] RandRotate90d
- [x] RandAffined
- [x] Rand2DElasticd
- [x] Rand3DElasticd
- [x] RandFlipd
- [x] RandRotated
- [x] RandZoomd

**Not applicable**
- BoundingRectd -- doesn't modify the image, but adds a dictionary entry containing the bounding box.

PR #1569, opened 2021-02-09
1567 add deletechanneld

leotam:
Signed-off-by: Leo Tam <leot@nvidia.com>

Fixes # .

### Description
Add DeleteChannel inverse operation to RepeatChannel via slicing. I tried squashing the commits which hilarious duplicated them instead. Anyways if it passes review, can fight git more. 

### Status

### Types of changes

### DIFF
--- a/monai/transforms/utility/dictionary.py
+++ b/monai/transforms/utility/dictionary.py
@@ -31,6 +31,7 @@ from monai.transforms.utility.array import (
     CastToType,
     ConvertToMultiChannelBasedOnBratsClasses,
     DataStats,
+    DeleteChannel,
     FgBgToIndices,
     Identity,
     LabelToMask,
@@ -52,6 +53,7 @@ __all__ = [
     "AsChannelLastd",
     "AddChanneld",
     "RepeatChanneld",
+    "DeleteChanneld",
     "SplitChanneld",
     "CastToTyped",
     "ToTensord",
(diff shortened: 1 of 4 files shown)


wyli:
thanks, adding the new class to this file will address some of the ci/cd errors https://github.com/Project-MONAI/MONAI/blob/master/monai/transforms/__init__.py#L258

monai/transforms/utility/dictionary.py
```diff
@@ -52,6 +53,7 @@
     "AsChannelLastd",
     "AddChanneld",
     "RepeatChanneld",
+    "DeleteChanneld",
```
rijobro, line 56: The `D` and `Dict` aliases need to go in here as well

leotam: Oops thought the mapping DeleteChannelD = DeleteChannelDict = DeleteChanneld did that already. Very tricksy, fixed it now

rijobro: Yeah it's not obvious. Thanks!

rijobro:
I find the name slightly confusing. I would expect `DeleteChannel` to have argument `idx: Union[Sequence[int], int]`, deleting the channels at the specified index/indices. Since this is the inverse of `RepeatChannel` I think the name should match that more closely, e.g., `RemoveRepeatedChannel`.

Alternatively, an `inverse` method could be added to https://github.com/Project-MONAI/MONAI/pull/1514. I've only concentrated on spatial transformations so far, but there's no reason it has to be limited to these use cases (the base class is `InvertibleTransform`).

leotam:
That's probably clearer to fit the use case. As for adding the inverse method to your spatial transforms PR, the hesitation would be it bloats the scope of that PR as it's a nontrivial expansion of scope, and makes it harder to test.

Discussing tests, what's this windows-latest CPU test failing? There's no clear error message. @wyli any helpful comments for these two items?

rijobro:
I think the relevant part of the Windows error is [here](https://github.com/Project-MONAI/MONAI/pull/1569/checks?check_run_id=1892008761#step:10:7666). It's trying to use pretrained `FCN`, but failing to download the pretrained network.

wyli:
agree that this transforms is removing channels according to the 'repeated' pattern, I think RemoveRepeatedChannel is more appropriate...
we could have a separate PR to inverse the repeat channel transform, we need https://github.com/Project-MONAI/MONAI/pull/1514 to provide some generic interfaces

wyli (approved):
thanks!

PR #1573, opened 2021-02-09
Add base class for TensorBoard handlers

Nic-Ma:
### Description
This PR added the base class for TensorBoard handlers.
We want to extend more TensorBoard handlers soon.

### Status
**Ready**

### Types of changes


wyli:
thanks, might need some `writer.close()` in the base class, we need concrete implementations to verify the multihandler use case

Nic-Ma:
> thanks, might need some `writer.close()` in the base class, we need concrete test cases to verify the multihandler use case

Thanks @wyli for your review. what do you mean `multihandler use case`?

wyli:
> > thanks, might need some `writer.close()` in the base class, we need concrete test cases to verify the multihandler use case
> 
> Thanks @wyli for your review. what do you mean `multihandler use case`?

I mean sharing the same summary_writer instance for multiple handlers. I guess we need some tests for that.. perhaps as a part of an integration test...

Nic-Ma:
> > > thanks, might need some `writer.close()` in the base class, we need concrete test cases to verify the multihandler use case
> > 
> > 
> > Thanks @wyli for your review. what do you mean `multihandler use case`?
> 
> I mean sharing the same summary_writer instance for multiple handlers. I guess we need some tests for that.. perhaps as a part of an integration test...

OK, sure, I will add it.
And I think we don't have the `writer.close()` logic in the 2 existing TensorBoard handlers, where do you expect to add it?

Thanks.

wyli:
> OK, sure, I will add it.
> And I think we don't have the `writer.close()` logic in the 2 existing TensorBoard handlers, where do you expect to add it?
> 
> Thanks.

I had an issue earlier that tensorboard server not being able to load if the event file writer is not closed properly, perhaps try to close when any writer before the last event during training

Nic-Ma:
Hi @wyli ,

For the `tensorboardhandler` attached to `trainer`, we can close writer when `Events.COMPLETED` triggered, but for the `tensorboardhandler` attached to `validator`, it completes many times for every validation  during training, we should always keep the event file open until training completed.
So it's hard to detect the "last event" for validator.
Maybe we need to add the `close()` logic in the `__del__()` function of TensorBoardHandler?

Thanks.

Nic-Ma:
Hi @wyli ,

I added the `close()` function referring to the TensorBoaderLogger in ignite: https://github.com/pytorch/ignite/blob/master/ignite/contrib/handlers/tensorboard_logger.py#L167
And also updated our unit tests.

Thanks.

wyli (approved):

PR #1596, opened 2021-02-19
Modify dynunet forward function

### LINKED ISSUE #1564: DynUnet get_feature_maps() issue when using distributed training
**Describe the bug**
If we use `DistributedDataParallel` to wrap the network, calling the 'get_feature_maps' function will raise the following error:

Screen Shot 2021-02-08 at 6 51 01 PM

It is common to return multiple variables in a network's forward function, not only for this case, but also for other situations such as for calculating triplet loss in metric learning based tasks. Therefore, here I think we should re-consider about what to return in the forward function of DynUnet.

Let me think about it and submit a PR then we can discuss it later.
@Nic-Ma @wyli @rijobro

yiheng-wang-nv:

Fixes #1564 .

### Description
Hi @rijobro @wyli @Nic-Ma @ericspod , after the discussion, I did some changes for the forward function of DynUNet. If we all used the list based return format, the default sliding window inferrer cannot work, thus I decided to return a single tensor for both train and eval modes. This change solves the DDP issue, and meet the restriction of TorchScript.

The change is that in deep supervision mode, all feature maps will be interpolated into the same size as the last feature map, and then stack together as a single tensor.

Therefore, in the loss calculation step, in the original implementation, the ground truth will be interpolated into each feature map's size, and then do the calculation. In this PR's implementation, the ground truth will be calculated with each interpolated feature map. These two ways have a little difference, but according to my simple test, the performance for task 04 will not be reduced.

Do you think we can change in this way?

### Status
**Work in progress**

### Types of changes

### DIFF
--- a/monai/networks/nets/dynunet.py
+++ b/monai/networks/nets/dynunet.py
@@ -14,6 +14,7 @@ from typing import List, Optional, Sequence, Union
 
 import torch
 import torch.nn as nn
+from torch.nn.functional import interpolate
 
 from monai.networks.blocks.dynunet_block import UnetBasicBlock, UnetOutBlock, UnetResBlock, UnetUpBlock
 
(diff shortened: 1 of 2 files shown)


wyli (approved):
thanks, to me this solution looks nice. just wanted to note that `interpolate` takes some additional arguments https://pytorch.org/docs/stable/nn.functional.html#torch.nn.functional.interpolate but they are not tunable in this PR. perhaps we could expand the API to support those if we have further feature requests

Nic-Ma (approved):
Looks good to me.
Please also update the DynUNet tutorial notebook.

Thanks.

yiheng-wang-nv:
> Looks good to me.
> Please also update the DynUNet tutorial notebook.
> 
> Thanks.

Thanks @Nic-Ma @rijobro and @wyli for the review. I submitted a PR to update the tutorial as you can see [here](https://github.com/Project-MONAI/tutorials/pull/126)

PR #1599, opened 2021-02-19
1568 Enhance tests to use self random state

Nic-Ma:
Fixes #1568 .

### Description
This PR enhances the 3D and 2D random images in tests to use an independent random state.

### Status
**Ready**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh --codeformat --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

PR #1607, opened 2021-02-21
Revert "1568 Enhance tests to use self random state (#1599)"

wyli:
This reverts commit 181f633f847ae1594a627f123931827b8e6d2995.

fixes #1599 

### Description
#1599 still generates nondeterministic results

### Status
**Ready**

### Types of changes


wyli:
i'm not 100% sure if this is related to #1599 but it was more stable before this. please look into it if you have time https://github.com/Project-MONAI/MONAI/runs/1947486945?check_suite_focus=true#step:6:8667

Nic-Ma:
I think maybe we should set a different random seed for every test case, so every test will be deterministic.
Anyway, let's merge this PR to avoid CI failures first.

Thanks.

Nic-Ma (approved):

PR #1612, opened 2021-02-22
1575 Add SaveImage and SaveImaged transforms

### LINKED ISSUE #1575: Add SaveNifti transform to debug transform chain
Need to add a SaveNifti transform based on the NiftiSaver for debug purpose. May also add SavePNG transform.

Nic-Ma:
Fixes #1575 .

### Description
This PR added the `SaveImage` and `SaveImaged` transforms to save data in both pre-transforms and post-transforms.
This can be a useful debugging tool and output tool, like the existing `DataStats` transform.

### Status
**Ready**

### Types of changes


monai/transforms/io/array.py
```diff
@@ -129,3 +133,90 @@ def __call__(
...
+            )
+        elif output_ext == ".png":
+            self.saver = PNGSaver(
+                output_dir=output_dir,
+                output_postfix=output_postfix,
+                output_ext=output_ext,
+                resample=resample,
+                mode=InterpolateMode(mode),
+                scale=scale,
+            )
```
rijobro, line 215: else raise error?

Nic-Ma: Sure, added in the latest commit.
Thanks.

rijobro (approved):
Looks good!

monai/data/png_writer.py
```diff
@@ -76,6 +76,10 @@ def write_png(
         else:
             raise ValueError(f"Unsupported scale: {scale}, available options are [255, 65535]")
 
+    # PNG data must be int number
+    if data.dtype != np.uint8 and data.dtype != np.uint16:
```
wyli, line 80: this is a good comment? https://deepsource.io/gh/Project-MONAI/MONAI/run/c6544bdf-660c-4bd1-8192-f184f59f7f0f/python/PYL-R1714

Nic-Ma: Hi @wyli ,

I used `data.dtype in (np.uint8, np.uint16)` before, but it can't pass `mypy`. Let me add a ignore.

Thanks.

monai/transforms/io/dictionary.py
```diff
@@ -106,4 +110,91 @@ def __call__(self, data, reader: Optional[ImageReader] = None):
...
+            scale=scale,
+            dtype=dtype,
+            output_dtype=output_dtype,
+            save_batch=save_batch,
+        )
+
+    def __call__(self, data):
+        d = dict(data)
+        for key in self.keys:
+            meta_data = d[f"{key}_{self.meta_key_postfix}"] if self.meta_key_postfix is not None else None
```
SachidanandAlle, line 194: if the image is loaded through numpy, can it still support saving as nifti files?

Nic-Ma: I think it's should work, just set `meta_key_postfix=None`, you can have a try.

Thanks.

SachidanandAlle: can't we not force the user to pass `meta_key_postfix=None`.  That way, user can use LoadImaged for both Numpy/Nifti.  I guess somehow NumpyReader doesn't create the meta_data 

Instead you can change it to
`meta_data = d.get(f"{key}_{self.meta_key_postfix}") if self.meta_key_postfix is not None else None`

Nic-Ma: Hi Sachi,

`NumpyReader` also creates meta_data:
https://github.com/Project-MONAI/MONAI/blob/master/monai/data/image_reader.py#L512
So I think we have meta_data by default.

Thanks for your suggestions.

Issue #876, opened 2020-08-09
Monai segmentation. Unable to use datasets besides Spleen.


dlabella29:
**Issue
Unable to use other datasets besides Task09_Spleen from medicaldecathlon when running the spleen_segmentation_3d script.
I believe the issue is with the SpacingD, pixdim transform, but could be elsewhere.
The other organ dataset has same file tree structure.
I also tried the other dataset with resampled .nii images and labels to 1x1x1.

*Note: I imported into Pycharm and ran as .py file.
Spleen dataset runs well with Dice > 0.94 after 600 epochs.

**Steps to reproduce the behavior:
1. Open the spleen_segmenatation_3d.py script.

2 working for spleen: Download and reference the Task09_Spleen dataset from medicaldecathalon.
    data_root = r'/home/USER/PycharmProjects/MONAI/Task09_Spleen'
    train_images = sorted(glob.glob(os.path.join(data_root, 'imagesTr', '*.nii.gz')))
    train_labels = sorted(glob.glob(os.path.join(data_root, 'labelsTr', '*.nii.gz')))

2 not working for other organ:
    data_root = r'/home/USER/PycharmProjects/MONAI/Task_other_organs'
    train_images = sorted(glob.glob(os.path.join(data_root, 'imagesTr', '*.nii.gz')))
    train_labels = sorted(glob.glob(os.path.join(data_root, 'labelsTr', '*.nii.gz')))

3 working for spleen: Use the default transforms:
        train_transforms = Compose([        
        LoadNiftid(keys=['image', 'label']),
        AddChanneld(keys=['image', 'label']),
        Spacingd(keys=['image', 'label'], pixdim=(1.5, 1.5, 2), mode=('bilinear', 'nearest')),
        Orientationd(keys=['image', 'label'], axcodes='RAS'),
        ScaleIntensityRanged(keys=['image'], a_min=-57, a_max=164, b_min=0.0, b_max=1.0, clip=True),
        CropForegroundd(keys=['image', 'label'], source_key='image'),
        RandCropByPosNegLabeld(keys=['image', 'label'], label_key='label', spatial_size=(96, 96, 96), pos=1,neg=1, num_samples=4, image_key='image', image_threshold=0),
        ToTensord(keys=['image', 'label'])
])

[...]

PR #1617, opened 2021-02-23
RandCropByPosNegLabel(d) to return 1 element if num_samples==1

rijobro:
### Description
If `RandCropByPosNegLabel` or `RandCropByPosNegLabeld` is called with `num_samples==1`, then return just an element, rather than a list containing 1 element. This is useful if followed by something that doesn't expect a list of results or debugging (e.g., `DataStatsd`).

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/croppad/array.py
+++ b/monai/transforms/croppad/array.py
@@ -574,7 +574,7 @@ class RandCropByPosNegLabel(Randomizable, Transform):
         image: Optional[np.ndarray] = None,
         fg_indices: Optional[np.ndarray] = None,
         bg_indices: Optional[np.ndarray] = None,
-    ) -> List[np.ndarray]:
+    ) -> Union[List[np.ndarray], np.ndarray]:
         """
         Args:
             img: input data to crop samples from based on the pos/neg ratio of `label` and `image`.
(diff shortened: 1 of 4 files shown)


rijobro:
addresses https://github.com/Project-MONAI/MONAI/issues/876#issuecomment-784006128.

Nic-Ma:
@rijobro thanks for raising this up with a PR.
But according to our experience with `DynUNet`, different return data types maybe not a good idea, users usually expect consistent output `shape` and `type` and develop the following logic.
@wyli @ericspod , what do you guys think?

Thanks.

wyli:
I don't have a preference here but I think `DataStats` could be improved to at least print some information for any datatype...

rijobro:
@Nic-Ma Good point. Although all these changes would be handled by the DataLoader, so I think that changing the output type wouldn't affect us in the same way as the output of `DynUNet`, because the batch created at the end will be just the same.

@wyli although we could change the `DataStatsd`, that was just one example. Presumably any transform that expects a dictionary but receives a list of dictionaries will have the same problem.

Nic-Ma:
Hi @rijobro ,

Actually, if a transform generates a list of dict instead of dict, the following transforms will execute on every item of the list.
Check this: https://github.com/Project-MONAI/MONAI/blob/master/monai/transforms/utils.py#L385
So I think it will not affect the `DataStat` transform.
Please feel free to correct me if I misunderstand your question.

Thanks.

rijobro:
Ah, ok thanks @Nic-Ma ! I think I'll close this for now then.

PR #1621, opened 2021-02-23
make random affine matrix

rijobro:
create random affine matrix (will be required for inverse transform)

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -151,6 +151,19 @@ def make_nifti_image(array, affine=None):
     return image_name
 
 
+def make_rand_affine(ndim: int = 3, random_state: Optional[np.random.RandomState] = None):
+    """Create random affine transformation (with values == -1, 0 or 1)."""
+    rs = np.random if random_state is None else random_state
+
+    vals = rs.choice([-1, 1], size=ndim)
+    positions = rs.choice(range(ndim), size=ndim, replace=False)
+    af = np.zeros([ndim + 1, ndim + 1])
+    af[ndim, ndim] = 1
+    for i, (v, p) in enumerate(zip(vals, positions)):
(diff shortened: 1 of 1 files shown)


wyli:
thanks, need some constraints about the invertibility?

rijobro:
No, I don't use these for inverting anything, I just use it to write nifti images for the purposes of testing. I need them to have affines along the lines of:

```
 0, 1,  0, 0
-1, 0,  0, 0
 0, 0, -1, 0
 0, 0,  0, 1
```
At any rate, this is trivially invertible.

tests/utils.py
```diff
@@ -151,6 +151,17 @@ def make_nifti_image(array, affine=None):
     return image_name
 
 
+def make_rand_affine(ndim: int = 3):
+    """Create random affine transformation (with values == -1, 0 or 1)."""
+    vals = np.random.choice([-1, 1], size=ndim)
```
wyli, line 156: would be great to support random generator e.g.: https://github.com/Project-MONAI/MONAI/blob/master/monai/data/synthetic.py#L51-L55

tests/utils.py
```diff
@@ -151,6 +151,17 @@ def make_nifti_image(array, affine=None):
...
+    positions = np.random.choice([0, 1, 2], size=ndim, replace=False)
```
wyli, line 157: do you mean `[0, 1]` when ndim==2?

tests/utils.py
rijobro: done!

tests/utils.py
rijobro: thanks, solved with `range(ndim)`

wyli (approved):

PR #1622, opened 2021-02-23
Compose len

rijobro:
Get the number of transformations in the compose operator (required for inverse transforms)

### DIFF
--- a/monai/transforms/compose.py
+++ b/monai/transforms/compose.py
@@ -231,6 +231,10 @@ class Compose(Randomizable, Transform):
                     f'Transform "{tfm_name}" in Compose not randomized\n{tfm_name}.{type_error}.', RuntimeWarning
                 )
 
+    def __len__(self):
+        """Return number of transformations."""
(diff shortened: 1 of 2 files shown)


monai/transforms/compose.py
```diff
@@ -231,6 +231,10 @@ def randomize(self, data: Optional[Any] = None) -> None:
                     f'Transform "{tfm_name}" in Compose not randomized\n{tfm_name}.{type_error}.', RuntimeWarning
                 )
 
+    def __len__(self):
+        """Return number of transformations."""
+        return sum(len(t) if isinstance(t, Compose) else 1 for t in self.transforms)
```
wyli, line 236: this is doing flattening and then get length, shall we have a `Compose.flatten()` and `len` just does a simple len(transform)? not sure about the specific use cases, I'm just looking at this API...

wyli (approved):

monai/transforms/compose.py
rijobro: Thanks, have added a `flatten()` method.

PR #1623, opened 2021-02-23
move transforms out of compose file

rijobro:
refactor the Transform, MapTransform and Randomizable to live in transform.py instead of compose.py.

Needed for inverse transform PR.


wyli:
this is potentially a breaking change in many cases, could you confirm @Nic-Ma?

rijobro:
If it would create many breaking changes, we could always alias the Transform, Randomizable, and MapTransform back into compose.py.

I made these changes because the amount of code in these classes grew so it made sense to put them into their correct file (Transform isn't of type Compose). 

I can do without this PR if it's deemed more effort than necessary, but I think it makes things cleaner.

rijobro:
Hi @wyli @Nic-Ma, backwards compatibility is now supported and tested, see here: https://github.com/Project-MONAI/MONAI/pull/1623/files#diff-5c752a2f5479d567fcd2f5dcaede165633debe9615906cb876fd2d3f552275dbR170-R171.

wyli (approved):
thanks, this is a non-breaking module rearrangement

PR #1624, opened 2021-02-23
decollate batch

rijobro:
decollate a batch of data


rijobro:
this requires https://github.com/Project-MONAI/MONAI/pull/1625

wyli (approved):
thanks! this could also be a packaged as a post-processing transform? https://github.com/Project-MONAI/MONAI/blob/master/monai/transforms/post/dictionary.py#L33

rijobro:
I've added the `Decollated` as a post transform, good idea.

Nic-Ma:
Hi @rijobro ,

Thanks for developing this new feature, may I know what's the typical use cases?

rijobro:
Hi @Nic-Ma of course. After the data loader, the data are collated. Images that were `C,H,W,[D]` become `B,C,H,W,[D]`. Similarly metadata are merged (e.g., a list of the transforms applied to each image).

To perform the inverse transforms on images, I want to convert them back from `B,C,H,W,[D]` and collated metadata to a list containing `B` elements, each element containing `C,H,W,[D]` images, and only the metadata associated with that image. 

My particular use case is for inverse transforms of a batch of images, but I'm sure this isn't the only use case for this functionality (for example, `NiftiSaver.save_batch` could be rewritten to use it).

Nic-Ma:
OK, I see, thanks for your explanation.

wyli:
not sure why it gives a circular import error in the container https://github.com/Project-MONAI/MONAI/runs/1977893666?check_suite_focus=true#step:3:32 I couldn't replicate the issue locally

rijobro:
Hm, so all the different CI builds were OK, but the docker one is causing a problem? Is this using master branch (since it's already merged)?

PR #1625, opened 2021-02-23
unify progress bar for datasets

rijobro:
unify arguments for progress bar in datasets.

### DIFF
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -51,14 +51,16 @@ class Dataset(_TorchDataset):
          },                           },                           }]
     """
 
-    def __init__(self, data: Sequence, transform: Optional[Callable] = None) -> None:
+    def __init__(self, data: Sequence, transform: Optional[Callable] = None, progress: bool = True) -> None:
         """
         Args:
             data: input data to load and transform to generate dataset for model.
             transform: a callable data transform on input data.
+            progress: whether to display a progress bar.
         """
         self.data = data
         self.transform = transform
+        self.progress = progress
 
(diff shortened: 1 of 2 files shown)


wyli (approved):
Thanks!

monai/data/dataset.py
```diff
@@ -51,14 +51,16 @@ class Dataset(_TorchDataset):
          },                           },                           }]
     """
 
-    def __init__(self, data: Sequence, transform: Optional[Callable] = None) -> None:
+    def __init__(self, data: Sequence, transform: Optional[Callable] = None, progress: bool = True) -> None:
```
wyli, line 54: perhaps the `progress` is not a fundamental property of a `Dataset`, so it shouldn't be in this base class, what do you think? for a user of the `Dataset` API, it's difficult to understand this option without looking at the cache dataset and some of the concrete implementations... 

@Nic-Ma @rijobro   sorry I should have had this discussion earlier during my review, somehow I missed it...

rijobro: I think you're right, sorry about that. A couple of choices:

1. New middleman class. `DatasetWithProgress`. `PersistentDataset` and `CacheDataset` would inherit from it.
2. Update current documentation for base class: `progress: whether to display a progress bar **(if relevant)**.`
3. Revert this PR.

wyli: thanks, I vote for option 1... @Nic-Ma thoughts?

Nic-Ma: I agree with you guys, `progress` should not be in the base class.

Thanks.

PR #1626, opened 2021-02-23
pad_list_data_collate

rijobro:
A collator, which prior to performing the running collator, pads the images to make sure that sizes match for a given batch.

This is done by calculating the maximum size in each dimension, and then padding all images in the batch to that size.

This introduced a slight circular dependency, because the spatialpadd is now used inside the data level. Solving it required all the seemingly unrelated changes to other files.


tests/test_pad_collation.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import List, Tuple
+
+import numpy as np
+from parameterized import parameterized
+
+from monai.data.utils import pad_list_data_collate
+from monai.transforms import RandRotate90d, RandRotated, RandSpatialCropd, RandZoomd
+from monai.utils import set_determinism
+
+set_determinism(seed=0)
```
wyli, line 22: is it possible to move this into `setUp`? e.g. https://github.com/Project-MONAI/MONAI/blob/db831ae096bec76371bb8c2e3db226d947a5ac06/tests/test_bounding_rect.py#L28-L32

monai/data/utils.py
```diff
@@ -240,7 +241,57 @@ def list_data_collate(batch: Sequence):
...
+        if "stack expects each tensor to be equal size" in re_str:
+            re_str += (
+                "\nMONAI hint: if your transforms intentionally create images of different shapes, creating your "
+                + "`DataLoader` with `collate_fn=pad_list_data_collate` might solve this problem (check its "
+                + "documentation)."
+            )
+        raise RuntimeError(re_str)
+
+
+def pad_list_data_collate(batch: Sequence):
```
wyli, line 257: this mainly support `batch` with a list of dictionary, shall we support list of list as well? the underlying `default_collate` is quite powerful: https://github.com/pytorch/pytorch/blob/master/torch/utils/data/_utils/collate.py#L43

wyli (commented):
thanks! maybe we have some issues with the overall utility module arrangements...please see some comments inline

tests/test_pad_collation.py
rijobro: Done!

monai/data/utils.py
rijobro: Although this is doable, I can't immediately see the use case. Currently I only want to decollate for inverse transformations which means the format is list of dictionaries. Would you accept this?

```python
if isinstance(batch[0], list):
    raise NotImplementedError("decollation has only currently been implemented for list of dictionaries. Please raise a PR if you feel this functionality would be useful")
```

wyli: sure, sounds great

rijobro: In the end it was pretty simple to allow for list of lists/tuples, so I added that too.

wyli: thanks, forgot to say that we should support different padding mode here...

rijobro: Good point, I've just added padding options

wyli (approved):
thanks, this looks nice

PR #1634, opened 2021-02-24
1633 fix string length > 1024 issue in all_gather

### LINKED ISSUE #1633: MetricsSaver Bug when filename is long
**Describe the bug**
When using the `MetricsSaver` handler, a list of filenames will be joined as a string first (see [here](https://github.com/Project-MONAI/MONAI/blob/master/monai/handlers/metrics_saver.py#L111)), and then send to write the metric report.

In distributed mode, the way to join all filenames into a string is directly connect all filenames by the setted delimiter (see [here](https://github.com/Project-MONAI/MONAI/blob/db2dbb07379e3dff47777fef8c9ac66671457916/monai/handlers/utils.py#L88)), and all filenames need to be gathered via ignite's `all_gather` function first. However, this method has a limitation that if the length of the string to be gathered is larger than 1024, the string will be truncated and only keep the first 1024 characters (the ignite source code of this function is in [here](https://github.com/pytorch/ignite/blob/master/ignite/distributed/comp_models/base.py#L90)).

Therefore, if the filename string is truncated, the number of metrics will be different from the number of filenames.

It is usual to get a larger than 1024 length string, for instance, a filename on my current working dataset is:
`'/workspace/data/medical/Task04_Hippocampus/imagesTr/hippocampus_033.nii.gz'` that has the length 74, thus more than 14 samples will introduce the bug.

Hi @Nic-Ma @wyli , we may need to modify the metric saving way, and avoid gathering the string (tensor or float types do not have the limitation, see [here](https://github.com/pytorch/ignite/blob/master/ignite/distributed/comp_models/base.py#L142)). What do you think?

Nic-Ma:
Fixes #1633 .

### Description
This PR fixed the string list all gather issue when length > 2014. Changed to all_gather every item separately instead of joining to a long string.

### Status
**Ready**

### Types of changes


monai/handlers/utils.py
```diff
@@ -85,27 +85,38 @@ def evenly_divisible_all_gather(data: torch.Tensor) -> torch.Tensor:
...
-    if idist.get_world_size() <= 1:
+    world_size = idist.get_world_size()
+    if world_size <= 1:
         return strings
 
-    _joined = delimiter.join(strings)
+    result: List[List[str]] = [[] for _ in range(world_size)]
+    # get length of strings
+    length = len(strings)
+    all_lens = idist.all_gather(length)
```
wyli, line 104: thanks, would be nice to have some filename like test cases

Nic-Ma: Hi @wyli ,

We already have the tests to cover filenames all-gather in multi-gpu:
https://github.com/Project-MONAI/MONAI/blob/master/tests/test_handler_metrics_saver_dist.py#L89

Thanks.

yiheng-wang-nv (approved):
Thanks Nic, I did some test locally and it works fine.

monai/handlers/utils.py
```diff
@@ -85,27 +85,38 @@ def evenly_divisible_all_gather(data: torch.Tensor) -> torch.Tensor:
     return torch.cat([data[i * max_len : i * max_len + l, ...] for i, l in enumerate(all_lens)], dim=0)
 
 
-def string_list_all_gather(strings: List[str], delimiter: str = "\t") -> List[str]:
+def string_list_all_gather(strings: List[str]) -> List[str]:
     """
     Utility function for distributed data parallel to all gather a list of strings.
+    Note that if the item in `strings` is longer than 1024 chars, it will be truncated to 1024.
```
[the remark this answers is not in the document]
Nic-Ma: Sure, updated in the latest commit.
Thanks.

monai/handlers/utils.py
wyli: the existing tests can't detect bug #1634

Nic-Ma: OK, do you mean to add a test that every single string in the list is longer than 1024 or just the total length of the list > 1024 as the bug?

Thanks.

wyli: thanks I'm more interested in the latter case but I think you can test both

Nic-Ma: Add test case for the later.

Nic-Ma: Added test cases to cover both.

Nic-Ma:
Hi @wyli ,

The CI failed due to some CUDA driver missing issue, have you seen this before?

Thanks.

PR #1635, opened 2021-02-24
enhance random range parameters

rijobro:
Allow min and max for random range-based parameters (e.g., amount of rotation in `RandRotate`)

From the new documentation:

If element `i` is iterable, then `uniform[-rotate_range[i][0], rotate_range[i][1])` will be used to generate the rotation parameter for the ith dimension. If not, `uniform[-rotate_range[i], rotate_range[i])` will be used. This can be altered on a per-dimension basis. E.g., `((0,3), 1, ...)`: for dim0, rotation will be in range `[0, 3]`, and for dim1 `[-1, 1]` will be used. Setting a single value will use `[-x, x]` for dim0 and nothing for the remaining dimensions.

Backwards compatibility is ensured because the old behaviour used to do `uniform[-rotate_range[i], rotate_range[i])`, and this PR just adds the possibility to do `uniform[-rotate_range[i][0], rotate_range[i][1])`.


wyli (approved):
thanks!

PR #1639, opened 2021-02-25
Randomizable constructor

rijobro:
Adds randomizable constructor.

### Description
This allows the parent class to store `prob` and `_do_transform`. 

I'll need this for the inverse transformation PR.

### Status
**Ready**


rijobro:
@wyli any chance you could have a look at this? Hoping to get it merged so I can move onto the next set of PRs.

wyli:
> @wyli any chance you could have a look at this? Hoping to get it merged so I can move onto the next set of PRs.

thanks I'll need more time to check the details as there's no unit test to cover the changes...

`Randomizable` is a general one, for example it's also used for shuffling the dataset https://github.com/Project-MONAI/MONAI/blob/master/monai/apps/datasets.py#L136 perhaps we could subclass it for the randomised transforms?

rijobro:
> thanks I'll need more time to check the details as there's no unit test to cover the changes...

Sure, my metric was just that none of the previous unit tests were broken. Slightly worrying that the code format wasn't picking up any warnings with this one... https://github.com/Project-MONAI/MONAI/blob/master/monai/apps/datasets.py#L136

rijobro:
@wyli updated to add a `RandomizableTransform` class.

monai/transforms/transform.py
```diff
@@ -77,7 +65,6 @@ def set_random_state(
         self.R = np.random.RandomState()
         return self
 
-    @abstractmethod
     def randomize(self, data: Any) -> None:
```
wyli, line 68: shall we support None here and in the `RandomizableTransform`?  (if it doesn't break anything...)
```suggestion
    def randomize(self, data: Optional[Any] = None) -> None:
```

wyli (approved):
thanks, it looks good to me

monai/transforms/transform.py
rijobro: I shouldn't have enabled automatic merging! I would quite like the default of `data` for like `Randomizable` and `RandomizableTransform` to default to `None`. Currently I have to use a lot of `super().randomize(None)`, but I'd rather have `super().randomize()`. However, this is incompatible with certain child classes. E.g., `RandFlip` takes `np.ndarray` as argument, which is incompatible with `None`.

The alternative is to switch all child classes from e.g., `np.ndarray` to `Optional[np.ndarray]`. What do you think?

wyli: thanks! it's just a very minor issue, I think the current one is also good, I forgot the mypy errors...

PR #1658, opened 2021-02-26
1651 implement RegUNet

kate-sann5100:
Signed-off-by: kate-sann5100 <yiwen.li@st-annes.ox.ac.uk>

Fixes # 1651.

### Description
Implements RegUNet which will serve as the parent class of the LocalNet and GlobalNet

### Status

### Types of changes

### DIFF
--- a/monai/networks/nets/regunet.py
+++ b/monai/networks/nets/regunet.py
@@ -1,0 +1,249 @@
+# Copyright 2020 - 2021 MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import List, Optional, Tuple, Union
+
+import torch
+from torch import nn
+
+from monai.networks.blocks.regunet_block import (
+    RegistrationDownSampleBlock,
+    RegistrationExtractionBlock,
+    RegistrationResidualConvBlock,
+    get_conv_block,
+    get_deconv_block,
+)
+
+
+class RegUNet(nn.Module):
+    """
+    Class that implements an adapted UNet. This class also serve as the parent class of LocalNet and GlobalNet
+
+    Reference:
+        O. Ronneberger, P. Fischer, and T. Brox,
+        “U-net: Convolutional networks for biomedical image segmentation,”,
+        Lecture Notes in Computer Science, 2015, vol. 9351, pp. 234–241.
+        https://arxiv.org/abs/1505.04597
+
+    Adapted from:
+        DeepReg (https://github.com/DeepRegNet/DeepReg)
+    """
+
+    def __init__(
+        self,
+        spatial_dims: int,
+        in_channels: int,
+        num_channel_initial: int,
+        depth: int,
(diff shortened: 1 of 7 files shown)


wyli:
is it possible to unify this network and [the existing unet](https://github.com/Project-MONAI/MONAI/blob/master/monai/networks/nets/unet.py)? otherwise we should highlight the key differences between them, so that the users wouldn't get confused... could you help us on this @ericspod?

ericspod:
The way our basic UNet is structured is around template methods called by the constructor when putting together what is a tower of layers. Each layer consists of a down sampling block, and skip connect, and the upsampling block. The path of data thus flows through the downsampling block, down to the next layer, when back up from that layer through the upsampling block. Think of squeezing the U structure of UNet into a tower. The important thing is the class represents the abstract structure and relies on methods to provide the blocks, for example [_get_down_layer](https://github.com/Project-MONAI/MONAI/blob/380f042dec9c55ba1a5ab241ae6b1a7b3e2b07fb/monai/networks/nets/unet.py#L109) defines the block for one layer of the down path. 

The way to adapt other UNet classes is to subclass our basic UNet and provide overrides for these methods, then in the constructor call the inherited constructor to create the structure. If the arguments passed to them aren't sufficient for what you're doing you can set some other state in the before calling the inherited constructor for the methods to pick up on. If other outputs are needed, for example DynUNet collecting outputs throughout the upsampling path, then some additional structure would need to be added to collect these outputs, I've been meaning to have a go at adapting that class but it's still being worked on.

With your RegUNet you have a lot of that already factored out, the difference is that you're explicitly handling the forward pass through the U structure, keeping track of the skip connections manually. I think this class can be adapted relatively easily by overriding the methods from the UNet to do what you have currently in the `build_*` methods which should allow you to avoid having a custom `forward` method entirely.

kate-sann5100:
I have drawn the architectures of `RegUNet`, `DynUNet`, `UNet`, and `BasicUNet` in the following diagram.
Unet

ericspod:
What I would suggest architecturally is to wrap the decode layers in a simple class which captures the output of the forward pass as a local variable. Once a forward pass through all the layers has been done a loop can go through these and collect the outputs which in RegUNet you sum together, in SynUNet these would be passed through the superheads. I think this will work and be compatible with Torchscript.

monai/networks/nets/regunet.py
```diff
@@ -0,0 +1,249 @@
...
+        # [depth - 1, ..., min_extract_level]
+        for i, (decode_deconv, decode_conv) in enumerate(zip(self.decode_deconvs, self.decode_convs)):
+            # [depth - 1, depth - 2, ..., min_extract_level]
+            decoded = decode_deconv(decoded)
+            if self.concat_skip:
+                decoded = torch.cat([decoded, skips[-i - 1]], dim=1)
+            else:
+                decoded = decoded + skips[-i - 1]
+            decoded = decode_conv(decoded)
+            outs.append(decoded)
```
mathpluscode, line 246: this line is a bug, It has been fixed in https://github.com/DeepRegNet/DeepReg/blob/main/deepreg/model/backbone/u_net.py#L556

in short, the decoded should be appended in a reverse order

kate-sann5100:
@ericspod 
That sounds great. Do you want to write the SynUnet or I am happy to write it?

monai/networks/nets/regunet.py
kate-sann5100: The extraction block takes care of the reversion in order.

https://github.com/Project-MONAI/MONAI/blob/6979a6708caed1932655be02baa411b6caf4842c/monai/networks/blocks/regunet_block.py#L262-L268

wyli (approved):
thanks! we've agreed that it'll take more iterations to unify the different net APIs. (this PR supports torchscript with a unit test)

PR #1663, opened 2021-02-26
Inverse transformation

rijobro:
Main PR for inverse transformations.

Try to keep changes to `__call__` minimal. Just append the current transformation to a list of transformations. And then add an `inverse` method to spatial transformations.

I also needed to return some extra information from certain `__call__`'s, for example affine matrices. backwards compatibility should be ensured by defaulting the option to return this to `False`.

(After this there is just batch inversion and test time augmentation!)


requirements-dev.txt
```diff
@@ -30,3 +30,4 @@ Sphinx==3.3.0
 recommonmark==0.6.0
 sphinx-autodoc-typehints==1.11.1
 sphinx-rtd-theme==0.5.0
+vtk
```
ericspod, line 33: We should really avoid adding large dependencies like this if it can be avoided.

rijobro: This is for the inverse of non-rigid transformations. I wrote them in both vtk and simple itk, so perhaps we could convert the simple itk code to itk and avoid another import? I think I'd need some help though

rijobro: Other options are simply not providing the inverse for non-rigid transformations (torchvision does this) or not adding it as a dependency (an exception is raised if the user tries to do the inverse but vtk/simpleitk is missing) and allowing the user to download it if desired.

ericspod: I think adding it as an optional dependency with `optional_import` is the way to do that at least for now.

rijobro: I only ever import vtk with `optional_import`, but I think that's true of everything in `requirements-dev.txt`.

monai/transforms/inverse_transform.py
```diff
@@ -0,0 +1,223 @@
...
+from monai.utils import optional_import
+
+sitk, has_sitk = optional_import("SimpleITK")
+vtk, has_vtk = optional_import("vtk")
+vtk_numpy_support, _ = optional_import("vtk.util.numpy_support")
+
+__all__ = ["InvertibleTransform", "NonRigidTransform"]
+
+
+class InvertibleTransform(Transform):
```
wyli, line 28: would be great to have more documentation for clarity:
- the "undo stack" implementation is for each `key` of data?
-  the `info` in `data[key_transform]` should be compatible with the default collate?
- it's using `id()` so I guess it's not persistent, shall we support a persistent undo stack?
(I feel the method names could be `push_transforms`, `pop_transforms`...)

monai/transforms/compose.py
```diff
@@ -136,3 +139,17 @@ def __call__(self, input_):
...
+            raise RuntimeError("Inverse method only available for dictionary transforms")
+        d = deepcopy(dict(data))
+        if keys:
+            keys = ensure_tuple(keys)
+
+        # loop backwards over transforms
+        for t in reversed(self.transforms):
+            # check if transform is one of the invertible ones
+            if isinstance(t, InvertibleTransform):
+                d = t.inverse(d, keys)
```
wyli, line 154: will we have a case that we want to modify keys as well, `d, keys = t.inverse(d, keys)`?

wyli (commented):
many thanks, this is a concrete working example! shall we split it into smaller PRs: 
1. the base APIs: compose, invertible transform (and please add independent unit tests)
2. the basic transform implementations (crop/pad/flip etc.) and the unit tests
3. the nonrigid transforms and tests

I put some dicussions for item 1 for now...

monai/transforms/compose.py
```diff
@@ -136,3 +139,17 @@ def __call__(self, input_):
         for _transform in self.transforms:
             input_ = apply_transform(_transform, input_)
         return input_
+
+    def inverse(self, data, keys: Optional[Tuple[Hashable, ...]] = None):
+        if not isinstance(data, Mapping):
+            raise RuntimeError("Inverse method only available for dictionary transforms")
+        d = deepcopy(dict(data))
+        if keys:
+            keys = ensure_tuple(keys)
```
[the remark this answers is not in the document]
rijobro: Which assumptions? I checked for dictionary data because only dictionary transforms have been made invertible. Would you prefer something like this:
```python
if sum(isinstance(t, InvertibleTransforms) for t in self.flatten().transforms) == 0:
    warning.warn("inverse has been called but no invertible transforms have been supplied")
```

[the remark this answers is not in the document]
wyli: sorry I didn't make it clear, I mean all the input validadtions, the method could be:
```py
def inverse(self, data, keys: Optional[Tuple[Hashable, ...]] = None):
       d = deepcopy(data)
       for t in reversed(self.transforms):
            # check if transform is one of the invertible ones
            if isinstance(t, InvertibleTransform):
                d = t.inverse(d, keys)
        return d
```

(btw are `flatten()` and `__len__` needed here? I could find where they are used...)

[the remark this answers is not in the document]
rijobro: Ok, but if a user has array transforms, they will call the `inverse` method but their data will remain unchanged and they won't receive any explanation as to why.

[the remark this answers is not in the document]
wyli: I see, I was thinking about making`Compose.__call__` and `Compose.inverse` mostly symmetric... that would be nice :)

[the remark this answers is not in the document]
rijobro: Alright, I'll remove the checks and add the suggested warning. That way, apart from the warning they're symmetric. Is that OK?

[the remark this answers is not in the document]
wyli: sure, thanks! still, `Compose.__call__` works fine with both array and dict, but why does `Compose.inverse` have the concept of `data` and `keys`? `InvertibleTransform` perhaps should inherit `MapTransform` since it's handling dictionaries? what do you think?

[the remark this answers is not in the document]
rijobro: I see your objection. I wanted to potentially be able to call the inverse on a subset of the keys. E.g., the forward pass contains image and label, then you infer a segmentation and only want to inverse the segmentation using the label's transformations. 

I think it's definitely a good idea to change `InvertibleTransform` to `InvertibleMapTransform` and have it inherit from `MapTransform`. However, I don't think that will solve this problem...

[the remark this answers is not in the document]
rijobro: Ok, I've removed the concept of keys from the inverse method, including `Compose.inverse`. I've replaced it with a withable [here](https://github.com/Project-MONAI/MONAI/pull/1663/files#diff-96cd15307195a8d855e3fb33c35abaa18afaecc182874751af4eb3144726d158R712):
```python
data = {"image": np.arange(16, dtype=float).reshape(1, 4, 4)}
t = SpatialPadd(["image", "label"], 10, allow_missing_keys=False)
_ = t(data)  # would raise exception
with AllowMissingKeysMode(t):
    _ = t(data)  # OK!
```
This means I can do the inverse with a subset of keys. Might also be useful in other circumstances!

What do you think? I'll put it into its own PR if you approve.

monai/transforms/inverse.py
```diff
@@ -0,0 +1,223 @@
...
+        self,
+        data: dict,
+        key: Hashable,
+        extra_info: Optional[dict] = None,
+        orig_size: Optional[Tuple] = None,
+    ) -> None:
+        """Append to list of applied transforms for that key."""
+        key_transform = str(key) + "_transforms"
+        info = {
+            "class": self.__class__.__name__,
```
Nic-Ma, line 61: May I know what this "class" is used for?

monai/transforms/inverse.py
```diff
@@ -0,0 +1,223 @@
...
+    def append_applied_transforms(
+        self,
+        data: dict,
+        key: Hashable,
+        extra_info: Optional[dict] = None,
+        orig_size: Optional[Tuple] = None,
+    ) -> None:
+        """Append to list of applied transforms for that key."""
+        key_transform = str(key) + "_transforms"
+        info = {
```
Nic-Ma, line 60: Should we use constant values for these keys as they are used in every transform?
Maybe self.ID, self.ORIG_SIZE, self.EXTRA_INFO?

monai/transforms/inverse.py
rijobro: It's useful for printing a list of invertible transforms that have been applied, e.g.: `for i in data["image_transforms"]: print(i["class"])`

Nic-Ma:
Thanks for the great work on this big task, the overall design looks good to me.
I have a question in my simple mind: If we save `original_affine` and `original_spatial_size` for all kinds of images in `LoadImage` transform, then **always update** `affine` and `spatial_size` in every spatial transform, can we inverse the data based on the `current affine + original affine` and `current spatial size + original spatial size`?
Please feel free to correct me if I misunderstand it.

Thanks.

Nic-Ma (approved):
Looks good to me as the first version of inverse transforms.
Thanks.

wyli (changes requested):
thanks! we'll merge https://github.com/Project-MONAI/MONAI/pull/1720 first and then back to this

rijobro:
@wyli do you have comments on this one or should I merge when CI is green? No problem if you need time to check it over.

wyli (changes requested):
I'm looking into the details, it may take some time especially for the spatial ones, do you want to have a separate PR to merge the `croppad` related first? they look straightforward. otherwise I'll put all my comments here gradually... (I may push some test cases to this branch if that's ok for you)

wyli:
major feature merged, the remaining tasks are covered by other tickets or PRs, I'm closing this

Issue #1662, opened 2021-02-26
Add option for transforms to skip missing keys instead of raising exception


rijobro:
**Is your feature request related to a problem? Please describe.**
In certain use cases, data might be missing, and this is acceptable. In such cases, we want to allow the user to skip transforms.

rijobro:
Also need to update documentation: https://github.com/Project-MONAI/MONAI/blob/master/monai/transforms/transform.py#L176-L182

PR #1676, opened 2021-03-02
[1662] Allow missing keys

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/1662.

### Description
Option to not raise exception if keys are missing during transform.

This is my current proposition. I've only done it for a few classes to get opinions before continuing with rest.

The generator in the base class to minimise code duplication.

### Status
**Work in progress**

### Types of changes


monai/transforms/croppad/dictionary.py
```diff
@@ -108,15 +109,16 @@ def __init__(
...
         """
-        super().__init__(keys)
+        super().__init__(keys, allow_missing_keys)
         self.mode = ensure_tuple_rep(mode, len(self.keys))
         self.padder = SpatialPad(spatial_size, method)
 
     def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.ndarray]:
         d = dict(data)
-        for key, m in zip(self.keys, self.mode):
+        for key, m in self.generator(d, self.mode):
```
rijobro, line 121: @wyli @Nic-Ma this is the approach if `zip` was previously used.

monai/transforms/transform.py
```diff
@@ -224,3 +225,19 @@ def __call__(self, data):
 
         """
         raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
+
+    def generator(
```
wyli, line 229: perhaps we could have a more intuitive function name here...

wyli (commented):
thanks, I think this a nice direction to go... @jansoe would be great to have your comments as well. I was also thinking of using a special key in `keys` such as `None` or `*` to represent the `allow_missing_keys` flag... but maybe that's more confusing APIs

jansoe (commented):
Very cool, I think this will make pipelines much more flexible.

wyli (approved):
thank you, it looks good to me

PR #1679, opened 2021-03-03
1500-update grid dataset to support patch level transforms

### LINKED ISSUE #1500: GridPatchDataset to support patch level transformations
the current `GridPatchDataset` supports transformations at image level https://github.com/Project-MONAI/MONAI/blob/2eada01efd9ddbcf8d9befe78b6ed67ce0cb454f/monai/data/grid_dataset.py#L26-L38

this ticket looks for options of patch level transforms to be applied before yielding the patches

context:
https://github.com/Project-MONAI/MONAI/discussions/1473

wyli:

Fixes #1500

### Description
revised grid patch dataset to support patch level transforms and coordinates outputs.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/grid_dataset.py
+++ b/monai/data/grid_dataset.py
@@ -32,7 +32,7 @@ class PatchIter:
     def __init__(
         self,
         patch_size: Sequence[int],
-        start_pos: Sequence[int],
+        start_pos: Sequence[int] = (),
         mode: Union[NumpyPadMode, str] = NumpyPadMode.WRAP,
         **pad_opts: Dict,
     ):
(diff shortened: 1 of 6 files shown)


Nic-Ma (approved):
Thanks for the enhancement.
Looks good to me.
@ericspod do you have any more comments?

Thanks.

monai/data/utils.py
```diff
@@ -209,8 +217,11 @@ def iter_patch(
     # patches which are only in the padded regions
     iter_size = tuple(s + p for s, p in zip(arr.shape, patch_size_))
 
-    for slices in iter_patch_slices(iter_size, patch_size_, start_pos_padded):
-        yield arrpad[slices]
+    for coords in iter_patch_coordinates(iter_size, patch_size_, start_pos_padded):
+        slices = tuple(slice(coord[0], coord[1]) for coord in coords)
+        # compensate original image padding
+        coords_no_pad = tuple((coord[0] - p, coord[1] - p) for coord, p in zip(coords, patch_size_))
```
ericspod, line 223: If `iter_patch_coordinates` yielded `slice` objects as before you could pull out the `start` and `stop` members instead of `coords[0]` and `coords[1]`, this would be clearer about what's going on and not change the behaviour of `iter_patch_coordinates` from what it was.

monai/data/grid_dataset.py
```diff
@@ -20,64 +20,143 @@
 from monai.transforms import apply_transform
 from monai.utils import NumpyPadMode, ensure_tuple
 
-__all__ = ["PatchDataset", "GridPatchDataset"]
+__all__ = ["PatchDataset", "GridPatchDataset", "PatchIter"]
 
 
-class GridPatchDataset(IterableDataset):
+class PatchIter:
```
ericspod, line 26: It's worth noting this whole class could be substituted with use of `partial` with `iter_patch` like:

```python
patch_iter=partial(iter_patch, patch_size=..., start_pos=..., copy_back=False, mode=..., **pad_opts)
```

wyli: yes I was using the partial function, it seems it's not compatible with the multiprocessing loader, the error message said it's not a picklable object

ericspod: It's going to depend on the arguments passed into partial, I can get it to pickle with simple values and a simple function but something more complex or locally declared would fail.

Issue #1677, opened 2021-03-02
in write_nifti an additional dimension is added to the output image


michalhg:
In the write_nifti function that exports the 3D segmentation prediction to a nifty image file, and addition dimension is added (either for channel or batch) which makes it unreadable using ITK-SNAP or Slicer.
I propose the following solution 

 results_img = nib.Nifti1Image(data.squeeze().astype(output_dtype), to_affine_nd(3, target_affine)) # squeeze channel dim

for row 105 in row nifty_writer.py file.

rijobro:
Hi @michalhg, have you looked into `NiftiSaver`, which has the `save_batch` method?

rijobro:
Or if you also want to remove the channel dimension, how about `Compose([SquezeDim, SaveImage])`?

michalhg:
I am using save_batch, but in the end it calls write_nifti without removing the channel dim. 
regarding your solution, should I avoid save_batch, and just call Compose([SquezeDim, SaveImage]) ?

rijobro:
Hi, I looked into this a bit more today, and have suggested [this PR](https://github.com/Project-MONAI/MONAI/pull/1680), allowing the user to squeeze any trailing singleton dimensions, being the channel dimension in your case.

wyli:
addressed by https://github.com/Project-MONAI/MONAI/pull/1680

PR #1680, opened 2021-03-03
nifti saver squeeze dims

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/discussions/1678.

### Description
Currently `NiftiSaver` (therefore default `SaveImage` and `SaveImaged`) saves images as `(H,W,D,C)`, even if input image had no channel dimension, or was only a 2D image.

This PR allows users to squeeze trailing singleton dimensions. If C==1, image will be saved as `(H,W,D)`. If D is also == 1, image will be saved as `(H,W)`. If D==1, but C>1, image will still be saved as `(H,W,D,C)` (to avoid channel dimension being interpreted as depth).

~~I also changed the saved path from `/some/path/filename/filename.nii.gz` to `/some/path/filename.nii.gz`, as I assumed the former was a mistake. I can undo this if it was intentional.~~ EDIT: I reverted this for backwards compatibility, but I still find it ugly.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/nifti_saver.py
+++ b/monai/data/nifti_saver.py
@@ -27,6 +27,8 @@ class NiftiSaver:
     Typically, the data can be segmentation predictions, call `save` for single data
     or call `save_batch` to save a batch of data together. If no meta data provided,
     use index from 0 as the filename prefix.
+
+    NB: image should include channel dimension: [B],C,H,W,[D].
     """
 
     def __init__(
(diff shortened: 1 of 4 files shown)


rijobro:
fixes https://github.com/Project-MONAI/MONAI/issues/1677.

wyli (approved):
thanks! looks good to me. (there's a similar workaround in niftynet https://github.com/NifTK/NiftyNet/blob/935bf4334cd00fa9f9d50f6a95ddcbfdde4031e0/niftynet/io/misc_io.py#L441-L450 proposed by @hjmjohnson )

wyli:
for the file path, I think it's originally from Clara Train where we want to save some other relevant predictions along with each `.nii.gz` file

PR #1689, opened 2021-03-04
Fix senet pretrained weights issue

### LINKED ISSUE #1688: Senet pretrained weights issue
**Describe the bug**
In #1120 , some layer names in  `Convolution` class have changed which lead to loading pre-trained weights unsuccessfully for Senet.

I'll submit a PR to fix this issue, and add the consistency test case to prevent the same issues in the future.

yiheng-wang-nv:

Fixes #1688 .

### Description

Except for the re-implementation of load pre-trained weights function, a new test case is added. For the same input tensor, the output between the torchvision's densenet and our version should be close (since we use nn.Linear for the fc layer, and Cadene's version uses conv layer with kernel size =1, it brings a few difference).

### Status
**Ready**

### Types of changes


wyli (approved):
thanks!

PR #1699, opened 2021-03-05
Raise import error for openslide and cuimage

behxyz:
### Description
In order for WSIReader to be used in [LoadImage](https://github.com/Project-MONAI/MONAI/blob/232e8a524c7074ceae4c45b756bfddc67772c05f/monai/transforms/io/array.py#L93), optional import error in the `__init__` of `WSIReader` was skipped, so a secondary exception was raised instead, which is not informative and does not help the user to know the cause. This PR fixes this issue by raising the appropriate import error when `WSIReader.read` is being called.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/image_reader.py
+++ b/monai/data/image_reader.py
@@ -645,7 +645,7 @@ class WSIReader(ImageReader):
         if self.reader_lib == "openslide":
             if has_osl:
                 self.wsi_reader = openslide.OpenSlide
-                print("> OpenSlide is being used.")                
+                print("> OpenSlide is being used.")
         elif self.reader_lib == "cuclaraimage":
             if has_cux:
                 self.wsi_reader = cuimage.CuImage
--- a/monai/data/image_reader.py
+++ b/monai/data/image_reader.py
@@ -639,13 +639,13 @@ class WSIReader(ImageReader):
 
     """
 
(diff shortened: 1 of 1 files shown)


monai/data/image_reader.py
```diff
@@ -672,6 +672,11 @@ def read(self, data: Union[Sequence[str], str, np.ndarray], **kwargs):
             data: file name or a list of file names to read.
 
         """
+        if (self.reader_lib == "openslide") and (not has_osl):
```
wyli, line 675: could have these logics even earlier for example between line 646 and 652?

behxyz: Actually no, Nic is instantiating this class in LoadImage, so the error shouldn't be raised in `__init__` (it used to be raise without having `has_osl` and `has_cux`).  The first place that it can be raised here.

I agree that the best place would be when the class is instantiating but for that purpose `LoadImage` need to be changed: https://github.com/Project-MONAI/MONAI/blob/6bc94b52572745534897cba4339e12aa18030dfa/monai/transforms/io/array.py#L93

wyli: sure you're right! 

(@Nic-Ma I'm approving this for now, perhaps we need another iteration to revise the base API?)

wyli (approved):
thanks!

PR #1714, opened 2021-03-08
with allow_missing_keys_mode

rijobro:
Allows keys to be temporarily be missing from map transform using `with`:

```python
data = {"image": np.arange(16, dtype=float).reshape(1, 4, 4)}
t = SpatialPadd(["image", "label"], 10, allow_missing_keys=False)
_ = t(data)  # would raise exception
with allow_missing_keys_mode(t):
    _ = t(data)  # OK!
```

This will be useful for doing the inverse of a segmented volume, since you probably won't need to do the inverse of the image key, only the label key.

Required `apply_transform` to be moved into `transform.py` to avoid circular import.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/utils.py
+++ b/monai/transforms/utils.py
@@ -9,12 +9,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from contextlib import contextmanager
 import itertools
-from monai.transforms.transform import MapTransform
-from monai.transforms.compose import Compose
 import random
 import warnings
+from contextlib import contextmanager
 from typing import Callable, List, Optional, Sequence, Tuple, Union
 
 import numpy as np
@@ -22,6 +20,8 @@ import torch
 
 from monai.config import DtypeLike, IndexSelection
 from monai.networks.layers import GaussianFilter
+from monai.transforms.compose import Compose
+from monai.transforms.transform import MapTransform
 from monai.utils import ensure_tuple, ensure_tuple_rep, ensure_tuple_size, fall_back_tuple, min_version, optional_import
 
 measure, _ = optional_import("skimage.measure", "0.14.2", min_version)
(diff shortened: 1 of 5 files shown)


monai/transforms/utils.py
```diff
@@ -730,3 +708,42 @@ def map_spatial_axes(
...
+        with allow_missing_keys_mode(t):
+            _ = t(data)  # OK!
+    """
+    if isinstance(transform, MapTransform):
+        transforms = [transform]
+    elif isinstance(transform, Compose):
+        # Only keep contained MapTransforms
+        transforms = [t for t in transform.flatten().transforms if isinstance(t, MapTransform)]
+    else:
+        transforms = []
```
Nic-Ma, line 736: I would suggest to raise exception instead of silently "pass" the code, because users maybe not clear how to use this API and pass wrong transforms.

monai/transforms/utils.py
```diff
@@ -730,3 +708,42 @@ def map_spatial_axes(
...
+    .. code-block:: python
+
+        data = {"image": np.arange(16, dtype=float).reshape(1, 4, 4)}
+        t = SpatialPadd(["image", "label"], 10, allow_missing_keys=False)
+        _ = t(data)  # would raise exception
+        with allow_missing_keys_mode(t):
+            _ = t(data)  # OK!
+    """
+    if isinstance(transform, MapTransform):
+        transforms = [transform]
```
Nic-Ma, line 731: Maybe you can use `ensure_tuple` to also support a list of `MapTransform`?

monai/transforms/utils.py
rijobro: Thanks, changed to raise an exception!

monai/transforms/utils.py
rijobro: Thanks, solved by converting an iterable of transforms to a Compose.

Nic-Ma (approved):
Looks good to me.
Thanks.

PR #1730, opened 2021-03-09
remove printing msg

wyli:

### Description
removes the debugging info from WSIreader, to be consistent with the other readers

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/image_reader.py
+++ b/monai/data/image_reader.py
@@ -645,11 +645,9 @@ class WSIReader(ImageReader):
         if self.reader_lib == "openslide":
             if has_osl:
                 self.wsi_reader = openslide.OpenSlide
-                print("> OpenSlide is being used.")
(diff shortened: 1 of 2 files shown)


ericspod (approved):

behxyz:
@wyli, it was more of a user info than a debugging info. Unlike the other readers, this reader uses different libraries so it might be useful to inform the user which one is being used.

wyli:
@behxyz I was trying to load a nifti file using the `loadimaged` API, it printed 'openslide is being used' which is quite confusing

behxyz:
@wyli, 
oh you are right! that's very confusing and shouldn't happen. It is because WSIReader is being instantiated in LoadImage, which I think we should change at some point, but at the meantime I'll add the info somewhere deeper into the class.

Issue #1710, opened 2021-03-08
GradCam giving inconsistent results


rijobro:
**Describe the bug**
When GradCam is called with `class_idx=None`, it should use the index of the class inferred by the network. If the network correctly classifies the image, the result should therefore be the same as `class_idx=y_true`. However, this is being reported as not being the case ([link](https://github.com/Project-MONAI/MONAI/discussions/1604#discussioncomment-435580_)). Seems to be a bug introduced since 0.4.0.

PR #1734, opened 2021-03-10
[1710] Gradcam eval

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/1710.

### Description
The problem was that for `class_idx=None`, the `argmax` was used in `training_mode`, resulting in different inferred classes compared to `eval_mode`. Secondly, the model was getting silently updated. Solved by using `self.model.eval()` but not using `with torch.no_grad()` or `with eval_mode(self.model)`.

I've added a new test to check that when class_idx=None` the inferred model is the same as that using `eval_mode`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/visualize/class_activation_maps.py
+++ b/monai/visualize/class_activation_maps.py
@@ -17,7 +17,6 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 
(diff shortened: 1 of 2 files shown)


wyli (approved):
thanks!

PR #1736, opened 2021-03-10
improve decollate test for set_determinism

rijobro:
Signed-off-by: Richard Brown <33289025+rijobro@users.noreply.github.com>

### DIFF
--- a/tests/test_decollate.py
+++ b/tests/test_decollate.py
@@ -73,7 +73,7 @@ class TestDeCollate(unittest.TestCase):
             transforms = Compose([LoadImaged("image"), transforms])
 
         dataset = CacheDataset(data, transforms, progress=False)
-        loader = DataLoader(dataset, batch_size=batch_size, shuffle=False, num_workers=0)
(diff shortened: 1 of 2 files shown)


tests/test_decollate.py
```diff
@@ -82,7 +73,7 @@ def test_decollation(self, _, data, batch_size=2, num_workers=2):
             transforms = Compose([LoadImaged("image"), transforms])
 
         dataset = CacheDataset(data, transforms, progress=False)
-        loader = DataLoader(dataset, batch_size=batch_size, shuffle=True, num_workers=num_workers)
+        loader = DataLoader(dataset, batch_size=batch_size, shuffle=False, num_workers=0)
```
wyli, line 76: does the previous value cause errors?

wyli (approved):

tests/test_decollate.py
rijobro: Shuffling should cause a problem, since it's compared with images directly from the dataset which aren't shuffled. I'll revert the `num_workers` since that doesn't cause any problems.

PR #1737, opened 2021-03-10
Croppad inverse transform

rijobro:
Croppad inverse transforms. All lossless (thus all the 0's in `test_inverse.py`).

### Status
**Ready**

### Types of changes


rijobro:
@wyli as requested in https://github.com/Project-MONAI/MONAI/pull/1737.

monai/transforms/croppad/dictionary.py
```diff
@@ -293,11 +359,36 @@ def __init__(
...
+            orig_size = np.array(transform[InverseKeys.ORIG_SIZE.value])
+            current_size = np.array(d[key].shape[1:])
+            pad_to_start = np.floor((orig_size - current_size) / 2)
+            # in each direction, if original size is even and current size is odd, += 1
+            pad_to_start[np.logical_and(orig_size % 2 == 0, current_size % 2 == 1)] += 1
+            pad_to_end = orig_size - current_size - pad_to_start
+            pad = np.empty((2 * len(orig_size)), dtype=np.int32)
+            pad[0::2] = pad_to_start
+            pad[1::2] = pad_to_end
+            inverse_transform = BorderPad(pad.tolist())
```
Nic-Ma, line 382: For crop transforms, maybe we can consider to add inverse_pad args if users don't want to use default value to pad back.

Nic-Ma (approved):
Looks good to me.
Thanks.

monai/transforms/croppad/dictionary.py
rijobro: Yeah we discussed this at the time to have kwargs for the inverse arguments. Would you be OK to add that as a future feature request?

monai/transforms/croppad/dictionary.py
```diff
@@ -232,11 +257,30 @@ def __init__(
...
 
+    def inverse(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.ndarray]:
+        d = deepcopy(dict(data))
+
+        for key in self.key_iterator(d):
+            transform = self.get_most_recent_transform(d, key)
+            # Create inverse transform
+            orig_size = np.array(transform[InverseKeys.ORIG_SIZE.value])
+            current_size = np.array(d[key].shape[1:])
+            roi_start = np.floor((current_size - orig_size) / 2)
```
Nic-Ma, line 272: Could you please help add more tests to cover the corner cases that `np.floor()` may shift 1 pixel for the ROI position?

monai/transforms/croppad/dictionary.py
```diff
@@ -347,12 +439,48 @@ def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.nda
...
+            pad_to_start = np.empty((len(orig_size)), dtype=np.int32)
+            pad_to_end = np.empty((len(orig_size)), dtype=np.int32)
+            if random_center:
+                for i, _slice in enumerate(transform[InverseKeys.EXTRA_INFO.value]["slices"]):
+                    pad_to_start[i] = _slice[0]
+                    pad_to_end[i] = orig_size[i] - _slice[1]
+            else:
+                current_size = d[key].shape[1:]
+                for i, (o_s, c_s) in enumerate(zip(orig_size, current_size)):
+                    pad_to_start[i] = pad_to_end[i] = (o_s - c_s) / 2
```
Nic-Ma, line 467: Same comment to make sure the size correct for odd numbers.

monai/transforms/croppad/dictionary.py
rijobro: I tried to cover all cases by (for each spatial dimension) having odd and even sized input, and cropping/padding by even amounts. On top of that, I have added 1D test cases for all odd/even combinations of input image and amount of cropping/padding.

rijobro: https://github.com/Project-MONAI/MONAI/pull/1737/files#diff-2f0622a3053b0bd6e89b040d75c4abd375d54611a311604b0cbaaa87f9b1d51eR56-R71

monai/transforms/croppad/dictionary.py
```diff
@@ -268,11 +313,31 @@ def __init__(
...
 
+    def inverse(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.ndarray]:
+        d = deepcopy(dict(data))
+
+        for key in self.key_iterator(d):
+            transform = self.get_most_recent_transform(d, key)
+            # Create inverse transform
+            orig_size = transform[InverseKeys.ORIG_SIZE.value]
+            pad_to_start = np.array(self.cropper.roi_start)
+            pad_to_end = orig_size - self.cropper.roi_end
```
wyli, line 328: the forward part of this has some tolerances for too large cropping values  https://github.com/Project-MONAI/MONAI/blob/master/monai/transforms/croppad/array.py#L239  (taking max over the values). Do we care about this kind of details in this PR? I feel it's gonna be lots of efforts

rijobro: I don't think we need to take care of these, since they're taken care of in the forward direction. I.e., in line 239 of the link you sent, `self.roi_end` is updated. This means that the amount of cropping is performed with `self.roi_end` and then the amount of padding in the inverse uses the same value.

wyli: I was trying to add this use case to `test_inverse`:
```py
TESTS.append(
    (
        "SpatialCropd 2d",
        "2D",
        0,
        SpatialCropd(KEYS, [49, 51], [390, 89]),
    )
)
```
the forward works fine, but in the inverse:
```
  File "/usr/local/anaconda3/envs/py37/lib/python3.7/site-packages/parameterized/parameterized.py", line 533, in standalone_func
  File "/Documents/MONAI/tests/test_inverse.py", line 301, in test_inverse
    fwd_bck = t.inverse(fwd_bck)
  File "/Documents/MONAI/monai/transforms/croppad/dictionary.py", line 333, in inverse
    d[key] = inverse_transform(d[key])
  File "/Documents/MONAI/monai/transforms/croppad/array.py", line 152, in __call__
    raise ValueError(f"self.spatial_border must contain only nonnegative ints, got {spatial_border}.")
ValueError: self.spatial_border must contain only nonnegative ints, got (0, -289, 7, 4).

```

rijobro: In [this commit](https://github.com/Project-MONAI/MONAI/pull/1737/commits/25c5daad158d4d7e3398f286d18c153c7d31b7f4), I addressed your concern by modifying the forward pass of `BorderPadd`. -ve padding values are set to 0. What do you think? Should be backwards compatible since previously it would raise an exception.

rijobro: I've also added extra edge cases for the other croppad. Modifying the forward of `BorderPadd` allows for little changes to the inverse methods.

rijobro:
@wyli if you're happy with the PR now, would you mind enabling auto-merge?

wyli (approved):
thanks, this looks good to me!

wyli:
there's an assertion error that occurs randomly...
```
  File "/__w/MONAI/MONAI/tests/test_inverse.py", line 291, in check_inverse
    self.assertLessEqual(mean_diff, acceptable_diff)
AssertionError: 0.47619047619047616 not less than or equal to 0
```
- https://github.com/Project-MONAI/MONAI/runs/2104731201?check_suite_focus=true#step:5:6353
- https://github.com/Project-MONAI/MONAI/runs/2103101765?check_suite_focus=true#step:6:6416

Failed: RandSpatialCropd1D odd Compose. Mean diff = 0.47619047619047616 (expected <= 0), unmodified diff: 2.0952380952380953

rijobro:
This could happen if the part that is randomly cropped includes non-zero voxels. But given that the seed is used during `setUp`, I would have thought that if it passed the test once, it would always pass?

wyli:
yes, strange that I couldn't replicate the issue locally on Linux or mac os. I've seen the errors from the CI a few times, the mean diff is always 0.4761 though

PR #1745, opened 2021-03-11
1742 Add DistributedWeightedRandomSampler

### LINKED ISSUE #1742: Need DistributedWeightedRandomSampler
Currently, PyTorch provides `WeightedRandomSampler` and `DistributedSampler` for the `sampler` arg of `DataLoader`, but users can't get both features together.

Add `DistributedWeightedRandomSampler` to resample data based on weights after the `distributed sampling`.

Nic-Ma:
Fixes #1742 .

### Description
This PR added the `DistributedWeightedRandomSampler` to enhance MONAI samplers.

### Status
**Ready**

### Types of changes


monai/data/utils.py
```diff
@@ -945,6 +948,55 @@ def __init__(self, even_divisible: bool = True, *args, **kwargs):
             self.total_size = data_len
 
 
+class DistributedWeightedRandomSampler(DistributedSampler):
```
ericspod, line 951: This file is now getting rather large, we should look later into splitting out the definitions here into other categories if that makes sense but for now maybe the samples should go into `samplers.py` in this directory.

monai/data/utils.py
```diff
@@ -945,6 +948,55 @@ def __init__(self, even_divisible: bool = True, *args, **kwargs):
...
+        self.generator = generator
+
+    def __iter__(self):
+        indices = list(super().__iter__())
+        num_samples = self.num_samples_per_rank if self.num_samples_per_rank is not None else self.num_samples
+        weights = torch.as_tensor([self.weights[i] for i in indices], dtype=torch.double)
+        # sample based on the provided weights
+        rand_tensor = torch.multinomial(weights, num_samples, self.replacement, generator=self.generator)
+
+        return iter([indices[i] for i in rand_tensor.tolist()])
```
ericspod, line 997: This could be `return (indices[i] for i in rand_tensor)` or maybe more clearly

```python
for in in rand_tensor:
    yield indices[i]
```

monai/data/utils.py
```diff
@@ -945,6 +948,55 @@ def __init__(self, even_divisible: bool = True, *args, **kwargs):
...
+    """
+
+    def __init__(
+        self,
+        weights: Sequence[float],
+        num_samples_per_rank: Optional[int] = None,
+        replacement: bool = True,
+        generator: Optional[torch.Generator] = None,
+        even_divisible: bool = True,
+        *args,
```
ericspod, line 981: It's getting very lengthy but it would much better for the position and keyword arguments here and in the inherited sampler to be explicitly declared rather than using args/kwargs. From a user's perspective it's not clear what these arguments are supposed to be without reading deeper into our documentation and then having to dig up Pytorch documentation as well. This would also allow us to put our position arguments after those required from the Pytorch type which I feel makes more sense, ie. additional arguments come after those "inherited" from the supertype's constructor.

monai/data/utils.py
Nic-Ma: OK, sounds good.

monai/data/utils.py
Nic-Ma: Hi @ericspod ,

Thanks for your suggestion.
The reason why we used `args, kwargs` for MONAI `DistributedSampler` and `DataLoader` it that:
Now we support PyTorch 1.5 - 1.8, but the args of `DistributedSampler` and `DataLoader` are slightly different in different PyTorch versions.
Maybe we can try to inherit the args when we drop support for 1.5 and 1.6 later.

Thanks.

monai/data/utils.py
Nic-Ma: sure, updated in the latest commit.
Thanks.

Nic-Ma:
Hi @ericspod ,

Thanks for your review.

Thanks.

monai/data/samplers.py
```diff
@@ -0,0 +1,93 @@
...
+        for example, input data: [1, 2, 3, 4, 5], rank 0: [1, 3, 5], rank 1: [2, 4].
+        args: additional arguments for `DistributedSampler` super class.
+        kwargs: additional arguments for `DistributedSampler` super class.
+
+    More information about DistributedSampler, please check:
+    https://github.com/pytorch/pytorch/blob/master/torch/utils/data/distributed.py
+
+    """
+
+    def __init__(self, even_divisible: bool = True, *args, **kwargs):
```
ericspod, line 35: The common arguments between versions 1.5 and 1.8 are `dataset`, `num_replicas`, `rank`, `shuffle`. I think these should be declared in this constructor and `**kwargs` used to pass `seed` and `drop_last` if these are to be used with a note in the docstring stating that's what it's there for. We can drop the use of `*args` this way.

ericspod (changes requested):
I think it's worth the effort to clean up the arguments in the constructor still but otherwise it looks good.

monai/data/samplers.py
Nic-Ma: Sure, I updated the PR according to your comments.
Thanks.

Nic-Ma:
Hi @ericspod ,

Thanks for your review.
I have clean the args, could you please help review it again?

Thanks.

wyli (approved):

Nic-Ma:
Hi @wyli and @ericspod ,

The CI tests conflicted with the master branch so CUDA out of memory. Passed now.

Thanks.

wyli (approved):

wyli:
I rerun the tests a few times, works fine now, could you help review it again? @ericspod

ericspod (approved):

PR #1759, opened 2021-03-12
1756 update ignite version to 0.4.4

### LINKED ISSUE #1756: update to use ignite 0.4.4
there're some enhancements since 0.4.2 based on monai's feedback.
shall we use the latest for monai v0.5 @Nic-Ma?

https://github.com/pytorch/ignite/releases

Nic-Ma:
Fixes #1756 .

### Description

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @vfdev-5 ,

Thanks for your great work on ignite project.
May I double confirm why you guys changed the default value of arg `device` from `None` to `CPU` in metrics?

Thanks.

vfdev-5:
Hi @Nic-Ma , this is a BC-breaking change since 0.4.3 to explicitly manage metric's storage device and reduce the time of moving from CPU to GPU in DDP for reduction etc. Please, see https://github.com/pytorch/ignite/issues/1082

Nic-Ma:
Hi @vfdev-5 ,

Thanks for your explanation!
Could you please also help review this PR?

Thanks.

monai/handlers/iteration_metric.py
```diff
@@ -77,7 +77,7 @@ def update(self, output: Sequence[torch.Tensor]) -> None:
         score = self.metric_fn(y_pred, y)
         if isinstance(score, (tuple, list)):
             score = score[0]
-        self._scores.append(score)
+        self._scores.append(score.clone().to(self._device))
```
vfdev-5, line 80: I think you can check the device of `score` and optionally pass it to `self._device` if mismatch. Otherwise cloning and sending can be expensive for each iteration. What do you think, @Nic-Ma ?

monai/handlers/roc_auc.py
```diff
@@ -17,8 +17,8 @@
 from monai.metrics import compute_roc_auc
 from monai.utils import Average, exact_version, optional_import
 
-idist, _ = optional_import("ignite", "0.4.2", exact_version, "distributed")
-EpochMetric, _ = optional_import("ignite.metrics", "0.4.2", exact_version, "EpochMetric")
+idist, _ = optional_import("ignite", "0.4.4", exact_version, "distributed")
+EpochMetric, _ = optional_import("ignite.metrics", "0.4.4", exact_version, "EpochMetric")
 
 
 class ROCAUC(EpochMetric):  # type: ignore[valid-type, misc]  # due to optional_import
```
vfdev-5, line 24: @Nic-Ma do you think it would be more interesting to improve and adapt our ROCAUC implementation or this one is very specific for your use-cases ?

vfdev-5 (commented):
Overall PR's impressions is good. I left few questions though (1 is more or less unrelated).

monai/handlers/iteration_metric.py
Nic-Ma: Hi @vfdev-5 ,

I agree with you, so I think I don't need to update this line at all, just keep the original:
```py
self._scores.append(score)
```

vfdev-5: Well, in this case, `self._device` wont be used, right ?
If user sets `device="cpu"`, but function returns "cuda", internal storage will be "cuda" and not cpu as required by the user ?

monai/handlers/roc_auc.py
Nic-Ma: Hi @vfdev-5 ,

There are 3 reasons that I didn't use the AUC implementation in ignite:
1. The ignite version is based on `sklearn.metrics`, we don't want to involve this dependency, it's self-contained here.
2. We need to support the case that predictions at different ranks have different lengths for distributed evaluation:
https://github.com/Project-MONAI/MONAI/blob/master/monai/handlers/roc_auc.py#L91
3. The AUC metric of ignite is in contrib folder, we don't want to involve your contrib code so far.

Thanks.

monai/handlers/iteration_metric.py
Nic-Ma: Hi @vfdev-5 ,

What's the expected behavior in ignite metrics? What is this `device` arg used for?
I refer to this code:
https://github.com/pytorch/ignite/blob/master/ignite/metrics/epoch_metric.py#L102

Thanks.

Nic-Ma: Do you mean to add another arg to optionally pass `device` to `self._device`?

Thanks.

vfdev-5: Expected behaviour is to put metric internal tensors to `self._device`. On the line to you cite here, we clone y_pred and y as we do not manage the lifetime of those tensors (and user can potentially modify inplace them after...).
In your case, score is a tensor from user-defined function and its lifetime is managed by your class. I think you can just simply do `score.to(self._device)` (and use if to check if eventually need to put to `self._device`).

Nic-Ma: Hi @vfdev-5 ,

I updated the PR according to your suggestion.
I think PyTorch will not copy the Tensor if the device of `score` is same as `self._device`:
https://pytorch.org/docs/stable/tensors.html#torch.Tensor.to

Thanks.

vfdev-5: Yes, I think so.

wyli (approved):
thanks!

PR #1761, opened 2021-03-13
BUG: MedNISTDataset returns differently sized datasets for different seeds

adamaji:
### Description
Fixes an issue with MedNISTDataset where choosing a different random seed returns a differently sized dataset. 
Reproduce by running

```
data0 = monai.apps.datasets.MedNISTDataset(root_dir="./tmp", transform=[], section="test", download=False, seed=0)
data42 = monai.apps.datasets.MedNISTDataset(root_dir="./tmp", transform=[], section="test", download=False, seed=42)
assert len(data0) != len(data42)
```

Instead of seeing whether a chosen random value falls between the train/val/test split serially, this PR uses the same kind of implementation for choosing random indices all at once like how it's done in `DecathlonDataset`.

Also modified the existing MedNIST test to use the full dataset length for comparison and to test dataset lengths for different seeds.

It seems like https://github.com/Project-MONAI/tutorials/blob/master/2d_classification/mednist_tutorial.ipynb may also need these changes.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/apps/datasets.py
+++ b/monai/apps/datasets.py
@@ -11,7 +11,7 @@
 
 import os
 import sys
-from typing import Any, Callable, Dict, List, Optional, Sequence, Union
+from typing import Callable, Dict, List, Optional, Sequence, Union
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


wyli (approved):
thanks!

PR #1687, opened 2021-03-04
1685-upgrade base image to 21.02

wyli:
Fixes #1685

### Description
- tested https://github.com/Project-MONAI/MONAI/actions/runs/619529045
- added docstring about building with a different base
- docker build cmd tested locally

### Status
**Ready**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [x] New tests added to cover the changes.
- [x] Quick tests passed locally by running `./runtests.sh --quick`.
- [x] In-line docstrings updated.

PR #1762, opened 2021-03-13
TEST: Fix failing pytype check for predict_segmentation

adamaji:
### Description
It seems like 0bd86b1 (https://github.com/Project-MONAI/MONAI/pull/1687/commits/67b2b1d77bb7f2209f633a8a2be04e1d79401eae) introduces changes which don't pass the pytype check running `./runtests.sh -f -u --net --coverage`. It's not clear to me why removing the cast in that commit should do this since the `.int()` should still return a `torch.Tensor`? But these changes make it so that pytype passes. Alternatively could do 
```
return torch.Tensor((logits >= threshold).int())
```
in the function but not sure what the original commit was trying to do by removing the cast.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/networks/utils.py
+++ b/monai/networks/utils.py
@@ -71,9 +71,7 @@ def slice_channels(tensor: torch.Tensor, *slicevals: Optional[int]) -> torch.Ten
     return tensor[slices]
 
 
-def predict_segmentation(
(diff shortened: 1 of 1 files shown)


adamaji:
cc @wyli for review

wyli (approved):
thanks, the previous change was to address the pytype issue after upgrading to use torch 1.8.

PR #1765, opened 2021-03-15
1755 improve focal loss

### LINKED ISSUE #1755: Unify the input of loss components
So far, the input format of focal loss is different from dice loss, in order to make it more convenient to use, I think it should be modified according to dice loss's input format.

yiheng-wang-nv:

Fixes #1755  .

### Description
Except for unifying the input parameters, some changes are made to make the class more generic:

- Modify the type of weight parameter form torch.Tensor to a sequence or a single value. It is helpful if users need to input the parameters via a json file.

### Status
**Ready**

### Types of changes


monai/losses/focal_loss.py
```diff
@@ -108,8 +136,8 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
         i = i.view(i.size(0), i.size(1), -1)
         t = t.view(t.size(0), t.size(1), -1)
 
-        # Compute the log proba (more stable numerically than softmax).
-        logpt = F.log_softmax(i, dim=1)
+        # Compute the log proba.
+        logpt = torch.log(torch.clamp(i, self.smooth_log, 1.0))
```
wyli, line 142: the previous impl. should be more stable, I think we should use that API...

monai/losses/focal_loss.py
```diff
@@ -29,15 +31,21 @@ class FocalLoss(_WeightedLoss):
 
     def __init__(
         self,
+        include_background: bool = True,
+        to_onehot_y: bool = False,
         gamma: float = 2.0,
-        weight: Optional[torch.Tensor] = None,
+        weight: Optional[Union[Sequence[int], int]] = None,
```
wyli, line 37: the type should be float  or tensor (in line 119)

monai/losses/focal_loss.py
```diff
@@ -9,16 +9,18 @@
...
 import torch.nn.functional as F
-from torch.nn.modules.loss import _WeightedLoss
+from torch.nn.modules.loss import _Loss
 
+from monai.networks import one_hot
 from monai.utils import LossReduction
 
 
-class FocalLoss(_WeightedLoss):
+class FocalLoss(_Loss):
```
wyli, line 23: why changing this base class? I think the previous one has better support of torchscript

monai/losses/focal_loss.py
```diff
@@ -57,80 +65,73 @@ def __init__(
...
+    def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
         """
         Args:
-            logits: the shape should be BCH[WD].
-                where C (greater than 1) is the number of classes.
-                Softmax over the logits is integrated in this module for improved numerical stability.
-            target: the shape should be B1H[WD] or BCH[WD].
-                If the target's shape is B1H[WD], the target that this loss expects should be a class index
-                in the range [0, C-1] where C is the number of classes.
+            input: the shape should be BNH[WD], where N is the number of classes.
```
wyli, line 77: if softmax is enabled by default, it should be clarified here..

monai/losses/focal_loss.py
```diff
@@ -57,80 +65,73 @@ def __init__(
...
-
-        # Compute the log proba (more stable numerically than softmax).
-        logpt = F.log_softmax(i, dim=1)  # N,C,H*W
-        # Keep only log proba values of the ground truth class for each voxel.
-        if target.shape[1] == 1:
-            logpt = logpt.gather(1, t.long())  # N,C,H*W => N,1,H*W
-            logpt = torch.squeeze(logpt, dim=1)  # N,1,H*W => N,H*W
+        # Change the shape of input and target to B x N x num_voxels.
+        i = i.view(i.size(0), i.size(1), -1)
+        t = t.view(t.size(0), t.size(1), -1)
```
wyli, line 110: the previous implementation has the unsqueeze for classification, any reason for removing that?

tests/test_seg_loss_integration.py
```diff
@@ -25,8 +25,8 @@
     [DiceLoss, {"to_onehot_y": True, "squared_pred": True, "batch": True}, {}],
     [DiceLoss, {"to_onehot_y": True, "sigmoid": True}, {}],
     [DiceLoss, {"to_onehot_y": True, "softmax": True}, {}],
-    [FocalLoss, {"gamma": 1.5, "weight": torch.tensor([1, 2])}, {}],
-    [FocalLoss, {"gamma": 1.5}, {}],
+    [FocalLoss, {"to_onehot_y": True, "gamma": 1.5, "weight": torch.tensor([1, 2])}, {}],
```
wyli, line 28: could you add a case of `to_onehot_y=False`

tests/test_focal_loss.py
```diff
@@ -22,7 +22,7 @@
 class TestFocalLoss(unittest.TestCase):
     def test_consistency_with_cross_entropy_2d(self):
         # For gamma=0 the focal loss reduces to the cross entropy loss
-        focal_loss = FocalLoss(gamma=0.0, reduction="mean")
+        focal_loss = FocalLoss(to_onehot_y=True, gamma=0.0, reduction="mean", weight=1.0)
```
wyli, line 25: could you add a case of `to_onehot_y=False`

wyli (commented):

monai/losses/focal_loss.py
yiheng-wang-nv: In some cases, like a 2 class classification task, a common way is to input a single `alpha` (as the paper shown [here](https://arxiv.org/pdf/1708.02002.pdf)). That's why I change the type to `Optional[Union[Sequence[float], float, torch.Tensor]]`. As for `_WeightedLoss `, the type of `weight` can only be `torch.Tensor`, thus I changed it into `_Loss`. What do you think?

monai/losses/focal_loss.py
yiheng-wang-nv: Hi @wyli , here I wish to unify the format between focal loss and other loss functions, thus I added `to_onehot_y` to ensure that the target will be changed into one-hot format and in this way, there is no need to handle the situation that `target.shape[1]` is 1.

wyli (approved):
thanks!

PR #1766, opened 2021-03-15
Add __len__ for DistributedWeightedRandomSampler

Nic-Ma:
### Description
This PR added the `__len__()` API for `DistributedWeightedRandomSampler`, refer to the PyTorch source code:
https://github.com/pytorch/pytorch/blob/master/torch/utils/data/sampler.py#L189

### Status
**Ready**

### Types of changes


monai/data/samplers.py
```diff
@@ -107,16 +107,18 @@ def __init__(
...
         weights = torch.as_tensor([self.weights[i] for i in indices], dtype=torch.double)
         # sample based on the provided weights
-        rand_tensor = torch.multinomial(weights, num_samples, self.replacement, generator=self.generator)
+        rand_tensor = torch.multinomial(weights, self.num_samples_per_rank, self.replacement, generator=self.generator)
 
         for i in rand_tensor:
             yield indices[i]
+
+    def __len__(self):
+        return self.num_samples_per_rank
```
wyli, line 124: not sure how we achieve `replacement=False` in a distributed setting, could you add a test for https://github.com/pytorch/pytorch/blob/master/torch/utils/data/sampler.py#L164 or we could remove the `False` option...

wyli (approved):
thanks!

monai/data/samplers.py
Nic-Ma: Sounds reasonable.
Removed `replacement` arg in the latest commit.

Thanks.

Issue #1749, opened 2021-03-11
Move inverse methods to array classes


rijobro:
Inverse methods currently sit in the dictionary transforms. This is because we ideally need the dictionary so that we can keep a track of the applied transforms to be able to apply the inverse in the correct order. 

However, there's no reason that the inverse can't live inside the array class, and each dictionary class calls its relevant inverse.

follow-up of https://github.com/Project-MONAI/MONAI/issues/1515

wyli:
thanks, this feature is now on the dev branch.

PR #1767, opened 2021-03-15
lossless inverse

rijobro:
Remaining lossless inverse transformations (croppad has already been merged: #1737).

This includes:

- Flipd
- RandFlipd
- RandAxisFlipd
- Orientationd (only lossless for orthogonal)
- Rotate90d
- RandRotate90d

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/dictionary.py
+++ b/monai/transforms/spatial/dictionary.py
@@ -282,7 +282,7 @@ class Orientationd(MapTransform, InvertibleTransform):
             orig_axcodes = nib.orientations.aff2axcodes(orig_affine)
             inverse_transform = Orientation(
                 axcodes=orig_axcodes,
-                as_closest_canonical=self.ornt_transform.as_closest_canonical,
+                as_closest_canonical=False,
                 labels=self.ornt_transform.labels,
             )
             # Apply inverse
--- a/monai/transforms/spatial/dictionary.py
+++ b/monai/transforms/spatial/dictionary.py
@@ -22,13 +22,11 @@ import numpy as np
 import torch
 
 from monai.config import DtypeLike, KeysCollection
-from monai.networks.layers import AffineTransform
 from monai.networks.layers.simplelayers import GaussianFilter
-from monai.transforms.croppad.array import CenterSpatialCrop, SpatialPad
-from monai.transforms.inverse import InvertibleTransform, NonRigidTransform
+from monai.transforms.croppad.array import CenterSpatialCrop
+from monai.transforms.inverse import InvertibleTransform
 from monai.transforms.spatial.array import (
     Affine,
-    AffineGrid,
     Flip,
     Orientation,
     Rand2DElastic,
@@ -52,6 +50,9 @@ from monai.utils import (
     fall_back_tuple,
 )
 from monai.utils.enums import InverseKeys
+from monai.utils.module import optional_import
+
+nib, _ = optional_import("nibabel")
 
 __all__ = [
     "Spacingd",
--- a/monai/transforms/spatial/dictionary.py
+++ b/monai/transforms/spatial/dictionary.py
@@ -15,16 +15,20 @@ defined in :py:class:`monai.transforms.spatial.array`.
 Class names are ended with 'd' to denote dictionary-based transforms.
 """
 
+from copy import deepcopy
 from typing import Any, Dict, Hashable, Mapping, Optional, Sequence, Tuple, Union
 
 import numpy as np
 import torch
 
 from monai.config import DtypeLike, KeysCollection
+from monai.networks.layers import AffineTransform
 from monai.networks.layers.simplelayers import GaussianFilter
-from monai.transforms.croppad.array import CenterSpatialCrop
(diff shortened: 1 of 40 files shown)


monai/transforms/spatial/dictionary.py
```diff
@@ -259,13 +265,36 @@ def __call__(
...
+        d = deepcopy(dict(data))
+        for key in self.key_iterator(d):
+            transform = self.get_most_recent_transform(d, key)
+            # Create inverse transform
+            meta_data = d[transform[InverseKeys.EXTRA_INFO.value]["meta_data_key"]]
+            orig_affine = transform[InverseKeys.EXTRA_INFO.value]["old_affine"]
+            orig_axcodes = nib.orientations.aff2axcodes(orig_affine)
+            inverse_transform = Orientation(
+                axcodes=orig_axcodes,
+                as_closest_canonical=self.ornt_transform.as_closest_canonical,
```
wyli, line 285: this `as_closest_canonical` option should always be `False` here.  to verify: 
```py
TESTS.append(
    (
        "Orientationd 3d",
        "3D",
        0,
        Orientationd(KEYS, "LAS", as_closest_canonical=True),
    )
)
```

tests/test_inverse.py
```diff
@@ -207,6 +213,82 @@
...
+    )
+)
+
+TESTS.append(
+    (
+        "RandRotate90d 3d",
+        "3D",
+        0,
+        # For data loader, output needs to be same size, so input must be square/cubic
+        SpatialPadd(KEYS, 110),
```
wyli, line 287:
```suggestion
```

tests/test_inverse.py
```diff
@@ -207,6 +213,82 @@
...
+    )
+)
+
+TESTS.append(
+    (
+        "Orientationd 3d",
+        "3D",
+        0,
+        # For data loader, output needs to be same size, so input must be square/cubic
+        SpatialPadd(KEYS, 110),
```
wyli, line 258:
```suggestion
```

monai/transforms/spatial/dictionary.py
```diff
@@ -286,11 +315,31 @@ def __init__(
...
+            self.push_transform(d, key)
             d[key] = self.rotator(d[key])
         return d
 
+    def inverse(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.ndarray]:
+        d = deepcopy(dict(data))
+        for key in self.key_iterator(d):
+            _ = self.get_most_recent_transform(d, key)
+            # Create inverse transform
+            spatial_axes = self.rotator.spatial_axes
```
wyli, line 327: if we always use the undo stack to store `spatial_axes` and `k` instead of `self.rotator`, there's no need to  call `check_transforms_match` for the instance id. perhaps it makes the entire impl. more robust?

monai/transforms/spatial/dictionary.py
```diff
@@ -337,6 +386,27 @@ def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Mapping[Hashable, np.
...
+
+    def inverse(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.ndarray]:
+        d = deepcopy(dict(data))
+        for key in self.key_iterator(d):
+            transform = self.get_most_recent_transform(d, key)
+            # Check if random transform was actually performed (based on `prob`)
+            if transform[InverseKeys.DO_TRANSFORM.value]:
+                # Create inverse transform
+                num_times_rotated = transform[InverseKeys.EXTRA_INFO.value]["rand_k"]
+                num_times_to_rotate = 4 - num_times_rotated
```
wyli, line 400: I guess this line 400 would be merged into the array transforms https://github.com/Project-MONAI/MONAI/issues/1749

wyli (commented):
thanks I put some minor comments inline

monai/transforms/spatial/dictionary.py
rijobro: Are you suggesting I do it now? Or in the future?

monai/transforms/spatial/dictionary.py
rijobro: Good spot, thanks! I've included that as a test.

monai/transforms/spatial/dictionary.py
rijobro: Hm, I think the check still needs to happen. Otherwise a different transformation would be popped. Also, imagine `Rotate90d` followed by a crop. If the two are done in the wrong order, the resulting image would be the incorrect size.

monai/transforms/spatial/dictionary.py
wyli: sorry i'm just making a reference here to the ticket for future work...

wyli (approved):
thanks looks good to me. @Nic-Ma could you help confirm?

rijobro:
@Nic-Ma If it's ok, I'm going to enable auto-merge so I can get on with next PR once this is merged. If you don't have the chance to review before it's merged, I'll open a new PR with your suggested modifications.

PR #1769, opened 2021-03-15
reduce randomly cropped amount to appease CI

rijobro:
Hopefully fixes [CI problem](https://github.com/Project-MONAI/MONAI/pull/1737#issuecomment-798900063) by reducing the number of voxels that are randomly cropped.

### Status
**Ready**

### Types of changes

### DIFF
--- a/tests/test_inverse.py
+++ b/tests/test_inverse.py
@@ -30,6 +30,7 @@ from monai.transforms import (
     DivisiblePadd,
     InvertibleTransform,
     LoadImaged,
+    Randomizable,
     RandSpatialCropd,
     ResizeWithPadOrCrop,
     ResizeWithPadOrCropd,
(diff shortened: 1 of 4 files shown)


wyli:
thanks, I'm looking into this as well, I'll push to this branch if I can locate the issue...

tests/test_inverse.py
```diff
@@ -309,6 +314,8 @@ def test_inverse(self, _, data_name, acceptable_diff, *transforms):
 
         # Apply forwards
         for t in transforms:
+            if isinstance(t, Randomizable):
+                t.set_random_state(seed=get_seed())
```
wyli, line 318: I think the issue is here, the randomised transforms are not deterministic unless the random state set.  this is a bit confusing because it's done in the `compose` by default... https://github.com/Project-MONAI/MONAI/blob/master/monai/transforms/compose.py#L100

Nic-Ma (approved):
Looks good to me.
If CI tests passed 2 times, we can merge it.

Thanks.

PR #1775, opened 2021-03-16
1757 Enhance CheckpointLoader to restore max_epochs

### LINKED ISSUE #1757: Continue training past checkpoint's max epochs
Currently, there is no clean way, to continue training past its initial epochs due to the fact that the state of the trained is being loaded by CheckpointLoader.

A quick solution would be to assume that if this case appears the code of the user will account for those things (aka just gamma decay, etc) and the CheckpointLoader should only modify the max_epochs after loading the checkpoint data and issue a Warning through a logger.

None

**Additional context**
None

Nic-Ma:
Fixes #1757 .

### Description
This PR fixed the issue that if loading the state_dict of engine itself, it will overwrite current `max_epochs` value.

### Status
**Ready**

### Types of changes


Nic-Ma:
@vfdev-5 , could you please help review this PR?

Thanks in advance.

monai/handlers/checkpoint_loader.py
```diff
@@ -80,5 +80,11 @@ def __call__(self, engine: Engine) -> None:
         """
         checkpoint = torch.load(self.load_path, map_location=self.map_location)
 
+        # save current max epochs setting in the engine, don't overwrite it if larger than max_epochs in checkpoint
+        prior_max_epochs = engine.state.max_epochs
         Checkpoint.load_objects(to_load=self.load_dict, checkpoint=checkpoint)
+        if engine.state.epoch > prior_max_epochs:
+            raise ValueError("current max epochs setting is smaller than epoch count in the checkpoint.")
```
vfdev-5, line 87: Maybe a nit, a small test entering this line could increase your codecov.

vfdev-5 (approved):
Looks good to me!

monai/handlers/checkpoint_loader.py
Nic-Ma: Hi @vfdev-5 ,

Thanks for your review.
I added more unit tests to cover it.

Thanks.

danieltudosiu:
The current error message is ambiguous and does not really help the user debug without a dive in the MONAI codebase. I would change the error message to be more descriptive from:

```
raise ValueError("current max epochs setting is smaller than epoch count in the checkpoint.")
```

to

```
raise ValueError(f"The epoch count found in the checkpoint ({engine.state.epoch}) is larger than the engine.state.max_epochs of the constructed engine prior loading the checkpoint ({prior_max_epochs}). If you want to further train this checkpoint please construct the trainer with a max_epochs bigger than the epoch count of the checkpoint and bigger or equal with its max_epochs. If you want to use it for inference please use the same max_epochs you used during training.")
```

It might be too stuffy, please have a go at trimming it down if you feel it needs it.

ericspod:
> raise ValueError(f"The epoch count found in the checkpoint ({engine.state.epoch}) is larger than the engine.state.max_epochs of the constructed engine prior loading the checkpoint ({prior_max_epochs}). If you want to further train this checkpoint please construct the trainer with a max_epochs bigger than the epoch count of the checkpoint and bigger or equal with its max_epochs. If you want to use it for inference please use the same max_epochs you used during training.")

A little more concise:

```python
ValueError(f"Epoch count in checkpoint ({engine.state.epoch}) is larger than the `engine.state.max_epochs` of engine ({prior_max_epochs}). To further train from checkpoint construct trainer with `max_epochs` larger than checkpoint's epoch count. To use for inference `max_epochs` must be the same as checkpoint.")
```

monai/handlers/checkpoint_loader.py
```diff
@@ -80,5 +80,14 @@ def __call__(self, engine: Engine) -> None:
         """
         checkpoint = torch.load(self.load_path, map_location=self.map_location)
 
+        # save current max epochs setting in the engine, don't overwrite it if larger than max_epochs in checkpoint
+        prior_max_epochs = engine.state.max_epochs
         Checkpoint.load_objects(to_load=self.load_dict, checkpoint=checkpoint)
+        if engine.state.epoch > prior_max_epochs:
+            raise ValueError(f"Epoch count ({engine.state.epoch}) in checkpoint is larger than \
```
ericspod, line 87: I think this will end up printing the message across different lines with the space included. The line should be broken up with multiple string literals to avoid this:

```python
raise ValueError(f"Epoch count ({engine.state.epoch}) in checkpoint is larger than "
               f"the `engine.state.max_epochs` ({prior_max_epochs}) of engine. To further train from checkpoint, "
                "construct trainer with `max_epochs` larger than checkpoint's epoch count. "
                "To use checkpoint for inference, no need to load state_dict for the engine.")
```

Nic-Ma: sounds good.

ericspod (approved):

Issue #1515, opened 2021-01-28
Invertible transforms


rijobro:
**Is your feature request related to a problem? Please describe.**
It would be useful to be able to invert certain transforms (notably spatial transforms). In this way, you could, for example, perform a segmentation on a cropped volume of the image, and then after inference you could inverse the cropping such that when saving to file, the segmentation can be superposed on the original image in an external viewer.

**Describe the solution you'd like**
After initial discussions, we've come up with this proposal:
- Have a new class `InvertibleTransforms`. Classes that inherit from this have an `invert` method implemented
- When calling invertible transforms, add a serialisation of this transform to the image
    - for this reason, we'll only work with dictionary data initially, as metadata can simply be added as another key in the dictionary
- The compose method will call the inverse methods in the reversed order (last in, first out).
- We will initially focus on spatial transforms, as these features are most desired. However, there is no reason not to expand this to non-spatial transforms if desired

Current progress is here: https://github.com/Project-MONAI/MONAI/pull/1514.

see also: https://github.com/Project-MONAI/MONAI/issues/365

tvercaut:
For the record, it would be good to ensure that the current design would also allow for easier test-time augmentation as per #718

wyli:
mostly addressed, the remaining concrete tasks tracked here https://github.com/Project-MONAI/MONAI/milestone/4, I'm closing this

PR #1778, opened 2021-03-16
Inverse Zoomd and RandZoomd

rijobro:
Inverse `Zoomd` and `RandZoomd`

#1515

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/dictionary.py
+++ b/monai/transforms/spatial/dictionary.py
@@ -1167,7 +1167,7 @@ class RandRotated(RandomizableTransform, MapTransform):
         return d
 
 
-class Zoomd(MapTransform):
+class Zoomd(MapTransform, InvertibleTransform):
     """
     Dictionary-based wrapper of :py:class:`monai.transforms.Zoom`.
 
@@ -1213,6 +1213,7 @@ class Zoomd(MapTransform):
         for key, mode, padding_mode, align_corners in self.key_iterator(
             d, self.mode, self.padding_mode, self.align_corners
         ):
+            self.push_transform(d, key)
             d[key] = self.zoomer(
                 d[key],
                 mode=mode,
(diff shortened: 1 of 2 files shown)


monai/transforms/spatial/dictionary.py
```diff
@@ -1311,6 +1338,29 @@ def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.nda
...
         return d
 
+    def inverse(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.ndarray]:
+        d = deepcopy(dict(data))
+        for key, mode, padding_mode, align_corners in self.key_iterator(
+            d, self.mode, self.padding_mode, self.align_corners
+        ):
+            transform = self.get_most_recent_transform(d, key)
+            # Create inverse transform
+            zoom = np.array(transform[InverseKeys.EXTRA_INFO.value]["zoom"])
```
wyli, line 1348: I think `_do_transform` flag should be checked somewhere in this method?

rijobro: thanks, updated

wyli (approved):
thanks!

monai/transforms/spatial/dictionary.py
```diff
@@ -1338,6 +1362,8 @@ def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.nda
         self.randomize()
         d = dict(data)
         if not self._do_transform:
+            for key in self.keys:
+                self.push_transform(d, key, extra_info={"zoom": self._zoom})
```
wyli, line 1366: perhaps extra_info `self._zoom` could have different sizes (on/off _do_transform) in a batch loader, then they are not compatible with the collate https://github.com/Project-MONAI/MONAI/issues/1798

rijobro: Yeah, you're right. You can see there's a few `ensure_tuple(self._zoom)` that come a few lines later, so they just need to be moved ahead.

PR #1782, opened 2021-03-16
Inverse Rotated

rijobro:
Inverse `Rotated` and `RandRotated`.

#1515

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -421,7 +421,7 @@ class Rotate(Transform):
         self.padding_mode: GridSamplePadMode = GridSamplePadMode(padding_mode)
         self.align_corners = align_corners
         self.dtype = dtype
-        self.rotation_matrix = None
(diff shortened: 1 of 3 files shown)


monai/transforms/spatial/array.py
```diff
@@ -482,7 +484,10 @@ def __call__(
             torch.as_tensor(np.ascontiguousarray(transform).astype(_dtype)),
             spatial_size=output_shape,
         )
-        return np.asarray(output.squeeze(0).detach().cpu().numpy(), dtype=np.float32)
+        output_np = np.asarray(output.squeeze(0).detach().cpu().numpy(), dtype=np.float32)
+        if return_rotation_matrix:
```
wyli, line 488: how about having a property `self.rot_mat = transform` instead of modifying the returns?

wyli (approved):
thanks!

PR #1783, opened 2021-03-16
cron tutorial notebooks

rijobro:
Run tutorial notebooks as cron job. Not sure if github actions syntax is correct.

### Status
**Ready**

### Types of changes

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -149,7 +149,7 @@ jobs:
         export CUDA_VISIBLE_DEVICES=$(python -m tests.utils)
         echo $CUDA_VISIBLE_DEVICES
         cd /opt
-        git clone --depth 1 https://github.com/Project-MONAI/tutorials.git  # latest commit
+        git clone --depth 1 --branch master --single-branch https://github.com/Project-MONAI/tutorials.git  # latest commit of master branch
         cd tutorials
         python -m pip install -r requirements.txt
     - name: Run tutorial notebooks
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -1,133 +1,136 @@
 name: crons
 
 on:
-  schedule:
-    - cron: "0 2 * * *"  # at 02:00 UTC
+#  schedule:
+#    - cron: "0 2 * * *"  # at 02:00 UTC
+  push:
+    branches:
+      - cron_tutorials
 
 jobs:
-  cron-gpu:
-    if: github.repository == 'Project-MONAI/MONAI'
-    container:
-      image: nvcr.io/nvidia/pytorch:20.03-py3  # CUDA 10.2
-      options: "--gpus all"
-    runs-on: [self-hosted, linux, x64, common]
-    strategy:
-      matrix:
-        pytorch-version: [1.5.1, 1.6.0, 1.7.1, latest]
-    steps:
-    - uses: actions/checkout@v2
-    - name: Install the dependencies
-      run: |
-        which python
-        python -m pip install --upgrade pip wheel
-        python -m pip uninstall -y torch torchvision
-        if [ ${{ matrix.pytorch-version }} == "latest" ]; then
-          python -m pip install torch torchvision
-        elif [ ${{ matrix.pytorch-version }} == "1.5.1" ]; then
-          python -m pip install torch==1.5.1 torchvision==0.6.1
-        elif [ ${{ matrix.pytorch-version }} == "1.6.0" ]; then
-          python -m pip install torch==1.6.0 torchvision==0.7.0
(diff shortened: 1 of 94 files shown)


.github/workflows/cron.yml
```diff
@@ -128,3 +128,26 @@ jobs:
...
+        python -m pip install ${{ matrix.pytorch }}
+        python -m pip install -r requirements-dev.txt
+    - name: Checkout tutorials and install their requirements
+      run: |
+        git clone https://github.com/Project-MONAI/tutorials.git
+        cd tutorials
+        python -m pip install -r requirements.txt
+    - name: Run tutorial notebooks
+      run: |
+        sh ./runtests.sh
```
ericspod, line 153: All our scripts are bash scripts so this should be `bash` in case `sh` isn't an alias to it. Also is the script called `runtests.sh` in tutorials? There's a script file there called `runner.sh` which refers to itself as `runtests.sh`.

rijobro: Thanks. Updated to using `$(pwd)/runner.sh` (`$(pwd)` is used for other bash scripts). Also made a PR for `runner.sh` referring to itself as `runtests.sh`.

ericspod (approved):

wyli (changes requested):
thanks I'll test it and get back to this PR

wyli:
the script's syntax is fine now, but some of the notebooks are not ready https://github.com/Project-MONAI/MONAI/runs/2125064870?check_suite_focus=true#step:6:154

rijobro:
Not sure how to resolve this, I don't see it locally: `ERROR: Unexpected bus error encountered in worker. This might be caused by insufficient shared memory (shm).`

wyli:
> Not sure how to resolve this, I don't see it locally: `ERROR: Unexpected bus error encountered in worker. This might be caused by insufficient shared memory (shm).`

I see, I'll try to set `--shm-size` of the docker container tomorrow

.github/workflows/cron.yml
```diff
@@ -128,3 +128,31 @@ jobs:
...
+        python -m pip install --upgrade pip wheel
+        python -m pip install -r requirements-dev.txt
+        BUILD_MONAI=0 python setup.py develop  # install monai
+    - name: Checkout tutorials and install their requirements
+      run: |
+        nvidia-smi
+        export CUDA_VISIBLE_DEVICES=$(python -m tests.utils)
+        echo $CUDA_VISIBLE_DEVICES
+        cd /opt
+        git clone --depth 1 https://github.com/Project-MONAI/tutorials.git  # latest commit
```
rijobro, line 152: Marginally quicker:
```suggestion
        git clone --depth 1 --branch master --single-branch https://github.com/Project-MONAI/tutorials.git  # latest commit of master branch
```

rijobro:
Tests made it further this time, but still running out of memory: `RuntimeError: DataLoader worker (pid 11596) is killed by signal: Bus error. It is possible that dataloader's workers are out of shared memory. Please try to raise your shared memory limit.`

rijobro:
@wyli Committing from a github suggestion seems to make DCO fail...

wyli:
now having a different issue with the tensorboard magic line: https://github.com/Project-MONAI/MONAI/runs/2130772126?check_suite_focus=true#step:6:298 any idea?

rijobro:
There's [info online](https://bugzilla.redhat.com/show_bug.cgi?id=1827745) about `cgi.escape` being deprecated in favour of `html.escape`:

> cgi.escape which was deprecated in Python 3.7 and removed in 3.8.

Not sure what that means for solving this problem. Perhaps using python3.7? Or maybe it would be solved with `pip install -U <package>`, but I'm not sure which package(s) need to be updated -- jupyter?

wyli:
updating tensorboard resolves that issue. now we have a testing data issue https://github.com/Project-MONAI/MONAI/runs/2132482457?check_suite_focus=true#step:6:317 need to host the zip file for testing?

rijobro:
@ericspod The current error comes from data in the [covid notebook](https://github.com/Project-MONAI/tutorials/blob/master/modules/interpretability/covid_classification.ipynb) not being automatically download-able from Kaggle. Do you think we could host a copy on Zenodo or somewhere else?

ericspod:
Yeah that's fine, it's easier that way for sure.

wyli:
I feel we could merge this after addressing https://github.com/Project-MONAI/tutorials/issues/144, so that these nightly tests would at least give us some useful information about all the tutorials...

wyli (approved):
thanks, this is a nice follow-up on https://github.com/Project-MONAI/tutorials/issues/101, nightly tutorial tests will be scheduled using this repo's runner https://github.com/Project-MONAI/MONAI/actions/runs/661446080/workflow

PR #1792, opened 2021-03-17
1790 Fix the type error of ScaleIntensityd

### LINKED ISSUE #1790: `ScaleIntensityd` input data type is not consistent with `ScaleIntensity`
In `ScaleIntensityd`, according to the doc string:
```
    If `minv` and `maxv` not provided, use `factor` to scale image by ``v = v * (1 + factor)``.
```
The parameters `minv` and `maxv` can be `None` that are the same as `ScaleIntensity`.
However, the types for these two parameters are limited to `float`.
 It will lead to an error that if users need to use `factor`, the parameter will not be used actually since `minv` and `maxv` have values, the corresponding codes are as follow:
```
        if self.minv is not None and self.maxv is not None:
            return np.asarray(rescale_array(img, self.minv, self.maxv, img.dtype))
        if self.factor is not None:
            return np.asarray(img * (1 + self.factor), dtype=img.dtype)
```
Let me submit a simple PR to modify them into `Optional[float]`.

yiheng-wang-nv:

Fixes #1790 .

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/intensity/array.py
+++ b/monai/transforms/intensity/array.py
@@ -148,7 +148,8 @@ class ScaleIntensity(Transform):
         Args:
             minv: minimum value of output data.
             maxv: maximum value of output data.
-            factor: factor scale by ``v = v * (1 + factor)``.
+            factor: factor scale by ``v = v * (1 + factor)``. In order to use
+                this parameter, please set `minv` and `maxv` into None.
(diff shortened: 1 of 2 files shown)


wyli (approved):
thanks!

PR #1794, opened 2021-03-17
Test time augmentations

### LINKED ISSUE #718: Test-Time-Augmentation(TTA)
Hi,
    I wonder if you can add some tools for test-time-augmentation, especially in semantic segmentation tasks.
Thank you

rijobro:
Test time augmentations.

closes #718

### Status
**Ready**

### Types of changes


rijobro:
requires #1795.

monai/data/test_time_augmentation.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def is_transform_rand(transform):
+    if not isinstance(transform, Compose):
+        return isinstance(transform, Randomizable)
+    # call recursively for each sub-transform
+    return any(is_transform_rand(t) for t in transform.transforms)
+
+
+class TestTimeAugmentation:
```
ericspod, line 35: We should have a docstring explaining what this class does and what the parameters are.

monai/data/test_time_augmentation.py
```diff
@@ -0,0 +1,116 @@
...
+        # check that the transform has at least one random component
+        if not is_transform_rand(self.transform):
+            raise RuntimeError(
+                type(self).__name__
+                + " requires a `Randomizable` transform or a"
+                + " `Compose` containing at least one `Randomizable` transform."
+            )
+
+    def __call__(
+        self, data: Dict[str, Any], num_examples=10, image_key="image", label_key="label", return_full_data=False
```
ericspod, line 59: Instead of the string literals "image" and "label" we can get the values from `CommonKeys`.

monai/data/test_time_augmentation.py
```diff
@@ -0,0 +1,116 @@
...
+    ):
+        d = dict(data)
+
+        # check num examples is multiple of batch size
+        if num_examples % self.batch_size != 0:
+            raise ValueError("num_examples should be multiple of batch size.")
+
+        # generate batch of data of size == batch_size, dataset and dataloader
+        data_in = [d for _ in range(num_examples)]
```
ericspod, line 68: `[d] * num_examples` would be equivalent.

monai/data/test_time_augmentation.py
```diff
@@ -0,0 +1,116 @@
...
+
+            # do model forward pass
+            batch_output = self.inferrer_fn(batch_images)
+            if isinstance(batch_output, torch.Tensor):
+                batch_output = batch_output.detach().cpu()
+            if isinstance(batch_output, np.ndarray):
+                batch_output = torch.Tensor(batch_output)
+
+            # check binary labels are extracted
+            if not all(torch.unique(batch_output.int()) == torch.Tensor([0, 1])):
```
ericspod, line 91: This class will only work with segmentation networks then? Could it theoretically make sense for other tasks like image-to-image transformation? The returned values may make less sense perhaps.

tests/test_testtimeaugmentation.py
```diff
@@ -0,0 +1,134 @@
...
+                AsDiscrete(threshold_values=True),
+                KeepLargestConnectedComponent(applied_labels=1),
+            ]
+        )
+
+        def inferrer_fn(x):
+            return post_trans(model(x))
+
+        tt_aug = TestTimeAugmentation(transforms, batch_size=5, num_workers=0, inferrer_fn=inferrer_fn, device=device)
+        mean, std = tt_aug(test_data)
```
ericspod, line 122: Doesn't this return four values: mode, mean, std, vvc?

monai/data/test_time_augmentation.py
rijobro: I've removed the check. We have the `return_full_data` if the user doesn't want to use the default metrics (e.g., computing the mode probably doesn't make sense for floating point numbers).

rijobro:
@wyli do you think you could provide some insight into [this error](https://github.com/Project-MONAI/MONAI/pull/1794/checks?check_run_id=2165223454#step:9:45)? I can't see how any of the files I've modified could cause this.

Also I'm getting [a flake error](https://github.com/Project-MONAI/MONAI/pull/1794/checks?check_run_id=2165222905#step:7:219). This comes from:
```python
    if isinstance(transform, Compose):
        # call recursively for each sub-transform
        return any(is_transform_rand_invertible(t) for t in transform.transforms)
```

When the if statement is satisfied, transform must therefore be of type `Compose` and therefore has the attribute `.transforms`. I thought flake was smart enough to figure this out? It does locally, at least.

wyli:
Hi @rijobro  for the first error, I think it's because `monai.data` is now depending on `ignite`, this is unexpected (the user should be able to use `monai.data` without installing ignite.

to replicate the issue in a system without ignite:
```
>>> import monai.data
  File "/opt/monai/monai/data/__init__.py", line 37, in <module>
    from .test_time_augmentation import TestTimeAugmentation
  File "/opt/monai/monai/data/test_time_augmentation.py", line 21, in <module>
    from monai.engines.utils import CommonKeys
  File "/opt/monai/monai/engines/__init__.py", line 12, in <module>
    from .evaluator import EnsembleEvaluator, Evaluator, SupervisedEvaluator
  File "/opt/monai/monai/engines/evaluator.py", line 17, in <module>
    from monai.engines.utils import CommonKeys as Keys
  File "/opt/monai/monai/engines/utils.py", line 33, in <module>
    class IterationEvents(EventEnum):
  File "/opt/monai/monai/utils/module.py", line 232, in __getattr__
    raise self._exception
  File "/opt/monai/monai/utils/module.py", line 191, in optional_import
    pkg = __import__(module)  # top level module
monai.utils.module.OptionalImportError: from ignite.engine import EventEnum (No module named 'ignite').
```

wyli:
I couldn't replicate the pytype error either... perhaps have a `# type: ignore`?

monai/data/test_time_augmentation.py
```diff
@@ -0,0 +1,180 @@
...
+from monai.transforms.compose import Compose
+from monai.transforms.inverse import InvertibleTransform
+from monai.transforms.transform import RandomizableTransform
+from monai.transforms.utils import allow_missing_keys_mode
+from monai.utils.enums import InverseKeys
+
+__all__ = ["TestTimeAugmentation"]
+
+
+def is_transform_rand_invertible(transform):
```
wyli, line 31: I feel according to this function's name, it checks the flag, should always return a boolean, shouldn't raise an error...

rijobro:
@wyli ok so the ignite must be because I'm importing `CommonKeys` which sits in engines. It should probably be moved to `enums.py` so that all modules can access it.

wyli:
> @wyli ok so the ignite must be because I'm importing `CommonKeys` which sits in engines. It should probably be moved to `enums.py` so that all modules can access it.

agreed, I think they are generic properties. cc @Nic-Ma this would be a breaking change

rijobro:
Hi @wyli, current tests are failing: https://github.com/Project-MONAI/MONAI/pull/1794/checks?check_run_id=2169625488#step:11:5093. But device is set with `device = "cuda" if torch.cuda.is_available() else "cpu"`. Any idea what's going on?

wyli:
> Hi @wyli, current tests are failing: https://github.com/Project-MONAI/MONAI/pull/1794/checks?check_run_id=2169625488#step:11:5093. But device is set with `device = "cuda" if torch.cuda.is_available() else "cpu"`. Any idea what's going on?

strange, how do I get the corresponding code of that log?

rijobro:
Don't think you can do it from the logs, but you can obviously get it from the changed files in the PR. Here's the problematic `test_testtimeaugmentations.py`: https://github.com/Project-MONAI/MONAI/blob/e02a67ce4f94a81027d7006976aea2a5601e92d4/tests/test_testtimeaugmentation.py

wyli:
> Don't think you can do it from the logs, but you can obviously get it from the changed files in the PR. Here's the problematic `test_testtimeaugmentations.py`: https://github.com/Project-MONAI/MONAI/blob/e02a67ce4f94a81027d7006976aea2a5601e92d4/tests/test_testtimeaugmentation.py

the log corresponds to 6b488c https://github.com/Project-MONAI/MONAI/pull/1794/checks?check_run_id=2169625488#step:3:472 which is the commit before [this device update](https://github.com/Project-MONAI/MONAI/pull/1794/commits/e02a67ce4f94a81027d7006976aea2a5601e92d4), and the latest test doesn't show the error, correct?

monai/data/test_time_augmentation.py
```diff
@@ -0,0 +1,178 @@
...
+        invertibles = np.array([isinstance(t, InvertibleTransform) for t in ts])
+        # check at least 1 random
+        if sum(randoms) == 0:
+            raise RuntimeError(
+                "Requires a `Randomizable` transform or a `Compose` containing at least one `Randomizable` transform."
+            )
+        # check that whenever randoms is True, invertibles is also true
+        for r, i in zip(randoms, invertibles):
+            if r and not i:
+                raise RuntimeError(
```
wyli, line 109: this could be a warning because the random factors may be in the model,

wyli (approved):
thanks!

Issue #1791, opened 2021-03-17
CuFileDriver cannot be open


wyli:
**Describe the bug**
running `nvidia-smi` gives a warning `CuFileDriver cannot be open. Falling back to use POSIX file IO APIs.` 
https://github.com/Project-MONAI/MONAI/runs/2130671421?check_suite_focus=true#step:6:294

**To Reproduce**
appeared in several recent testing results, but all tests finished normally

wyli:
could you help look into this? @Nic-Ma I saw this today but don't know the root cause

Nic-Ma:
Hi @behxyz ,

I think you know something about this issue, could you please help take a look?

Thanks in advance.

behxyz:
Hi @wyli and @Nic-Ma,
It is not related to `nvidia-smi` and running `nvidia-smi` does not give this warning. This is an expected message from `cuCIM`. Since our docker does not have `libcufile` for GPU Direct Storage, it falls back to use POSIX IO, so it gives a warning. It doesn't have any effect on the functionality of our codes, so nothing to worry about. I talked to the developer and it can be removed in the next release but the implication of removing this message needs to be examined first.

wyli:
ok, do you know which line is triggering this warning? there are many repeated warning messages during the tests. if it's from cuCIM, is it possible to mute it unless we actually use the relevant module?

behxyz:
This message comes from C++ level of cuCIM, and we cannot silence it. It is produced right after `import cucim` in `image_reader`. Since there are many tests for image_reader, we will see it multiple times.
It's a warning that we should skip for now and maybe request to remove this warning from `cucim` library.

wyli:
we should move the cucim import to line 652 in the reader as a workaround? https://github.com/Project-MONAI/MONAI/blob/master/monai/data/image_reader.py#L652

wyli:
the notebooks are giving too many outputs
```
Executing:  22%|██▏       | 5/23 [00:06<00:24,  1.37s/cell]
Warning: :  39%|███▉      | 9/23 [00:17<00:30,  2.20s/cell][warning] CuFileDriver cannot be open. Falling back to use POSIX file IO APIs.
Warning:  CuFileDriver cannot be open. Falling back to use POSIX file IO APIs.
Warning:  CuFileDriver cannot be open. Falling back to use POSIX file IO APIs.
Warning:  CuFileDriver cannot be open. Falling back to use POSIX file IO APIs.
[...]

PR #1796, opened 2021-03-17
Remove WSIReader from LoadImage

behxyz:
### Description
Since we are not using WSIReader in the same way that we use other image reader, I removed it from LoadImage to avoid any potential  issues that it may cause.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/io/array.py
+++ b/monai/transforms/io/array.py
@@ -19,7 +19,7 @@ import numpy as np
 import torch
 
 from monai.config import DtypeLike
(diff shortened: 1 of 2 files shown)


behxyz:
The problem with the warning message( #1791) will be resolved independent of this PR, so I'm fine if we merge this PR or not.

Nic-Ma:
Please also update the doc-string of `LoadImaged` transform.

Thanks.

wyli (approved):

Issue #1798, opened 2021-03-17
PR #1778 causes error


behxyz:
**Describe the bug**
PR #1778 has broken a Clara Train MMAR that uses RandomZoomD.

```
Original Traceback (most recent call last):
  File "/workspace/MONAI/monai/data/utils.py", line 257, in list_data_collate
    return default_collate(data)
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 73, in default_collate
    return {key: default_collate([d[key] for d in batch]) for key in elem}
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 73, in <dictcomp>
    return {key: default_collate([d[key] for d in batch]) for key in elem}
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 83, in default_collate
    return [default_collate(samples) for samples in transposed]
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 83, in <listcomp>
    return [default_collate(samples) for samples in transposed]
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 73, in default_collate
    return {key: default_collate([d[key] for d in batch]) for key in elem}
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 73, in <dictcomp>
    return {key: default_collate([d[key] for d in batch]) for key in elem}
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 73, in default_collate
    return {key: default_collate([d[key] for d in batch]) for key in elem}
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 73, in <dictcomp>
    return {key: default_collate([d[key] for d in batch]) for key in elem}
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 81, in default_collate
    raise RuntimeError('each element in list of batch should be of equal size')
RuntimeError: each element in list of batch should be of equal size

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/worker.py", line 202, in _worker_loop
    data = fetcher.fetch(index)
[...]

PR #1801, opened 2021-03-18
[1798] fix RandZoomd collation

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/1798.

### Description
`self._zoom` was handled differently depending on `self._do_transform`. Modified to be the same, such that they can be collated together. 

Also add testing for all invertible random transforms.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/inverse.py
+++ b/monai/transforms/inverse.py
@@ -9,10 +9,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import torch
 from typing import Dict, Hashable, Optional, Tuple
 
 import numpy as np
+import torch
 
 from monai.transforms.transform import RandomizableTransform, Transform
 from monai.utils.enums import InverseKeys
--- a/monai/transforms/inverse.py
+++ b/monai/transforms/inverse.py
@@ -94,7 +94,7 @@ class InvertibleTransform(Transform):
             return
         # basic check if multiprocessing uses 'spawn' (objects get recreated so don't have same ID)
         if (
-            torch.multiprocessing.get_start_method(allow_none=True) == "spawn"
+            torch.multiprocessing.get_start_method(allow_none=False) == "spawn"
             and transform[InverseKeys.CLASS_NAME.value] == self.__class__.__name__
         ):
             return
--- a/monai/transforms/inverse.py
+++ b/monai/transforms/inverse.py
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import sys
+import torch
 from typing import Dict, Hashable, Optional, Tuple
 
 import numpy as np
@@ -92,8 +92,11 @@ class InvertibleTransform(Transform):
         """Check transforms are of same instance."""
         if transform[InverseKeys.ID.value] == id(self):
             return
-        # basic check if windows because of multiprocessing differences (objects get recreated so don't have same ID)
-        if sys.platform == "win32" and transform[InverseKeys.CLASS_NAME.value] == self.__class__.__name__:
+        # basic check if multiprocessing uses 'spawn' (objects get recreated so don't have same ID)
+        if (
+            torch.multiprocessing.get_start_method(allow_none=True) == "spawn"
+            and transform[InverseKeys.CLASS_NAME.value] == self.__class__.__name__
+        ):
             return
         raise RuntimeError("Should inverse most recently applied invertible transform first")
 
--- a/monai/transforms/inverse.py
+++ b/monai/transforms/inverse.py
@@ -93,7 +93,7 @@ class InvertibleTransform(Transform):
         if transform[InverseKeys.ID.value] == id(self):
             return
         # basic check if windows because of multiprocessing differences (objects get recreated so don't have same ID)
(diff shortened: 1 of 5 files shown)


rijobro:
@behxyz, do you think you could try this PR? I'm fairly confident it will have solved the problem.

tests/test_inverse_collation.py
```diff
@@ -0,0 +1,91 @@
...
+        # num workers = 0 for mac
+        num_workers = 2 if sys.platform != "darwin" else 0
+
+        dataset = CacheDataset(self.data, transform=transform, progress=False)
+        loader = DataLoader(
+            dataset,
+            batch_size=self.batch_size,
+            shuffle=False,
+            num_workers=num_workers,
+            collate_fn=pad_list_data_collate,
```
wyli, line 80: I think test `list_data_collate` should be tested here or as part of the test_inverse? https://github.com/Project-MONAI/MONAI/blob/master/tests/test_inverse.py#L541   also, not sure whether https://github.com/Project-MONAI/MONAI/blob/master/monai/data/utils.py#L260 needs an update to cover the https://github.com/Project-MONAI/MONAI/issues/1798 error mesages

rijobro: I used `pad_list_data_collate` because my input image wasn't cubic, so random flips etc result in images of different sizes. I'll switch to using a cubic image (100,100,100), and then I can use `list_data_collate`.

What do you suggest the error message be updated to?

wyli: change the condition to a more general: `if "equal size" in re_str:`? 

I think the current `pad_list_data_collate` is nice, perhaps add another class/methods for `list_data_collate `

wyli: if the last transform is `ResizeWithPadOrCropd`, then the input doesn't have to be cubic...

rijobro: Ok, now testing both default collation (`list_data_collate`) and `pad_list_data_collate`. Reverted to non-cubic image and added `ResizeWithPadOrCropd` when `list_data_collate` is used.

monai/transforms/inverse.py
```diff
@@ -89,8 +90,12 @@ def push_transform(
 
     def check_transforms_match(self, transform: dict) -> None:
         """Check transforms are of same instance."""
-        if transform[InverseKeys.ID.value] != id(self):
-            raise RuntimeError("Should inverse most recently applied invertible transform first")
+        if transform[InverseKeys.ID.value] == id(self):
+            return
+        # basic check if windows because of multiprocessing differences (objects get recreated so don't have same ID)
+        if sys.platform == "win32" and transform[InverseKeys.CLASS_NAME.value] == self.__class__.__name__:
+            return
```
wyli, line 97: more specific check (need `import torch`):
```suggestion
        # basic check if multiprocessing uses 'spawn' (objects get recreated so don't have same ID)
        if (
            torch.multiprocessing.get_start_method(allow_none=True) == "spawn"
            and transform[InverseKeys.CLASS_NAME.value] == self.__class__.__name__
        ):
            return
```

monai/transforms/inverse.py
```diff
@@ -9,6 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
```
wyli, line 12:
```suggestion
import torch
```

wyli (approved):
thanks for the quick fix!

rijobro:
@wyli do you understand the error? https://github.com/Project-MONAI/MONAI/pull/1801/checks?check_run_id=2141079751#step:11:5166

It seems that the following code is being executed on the windows and osx test:

```python
# skip this test if multiprocessing uses 'spawn', as the check is only basic anyway
if torch.multiprocessing.get_start_method(allow_none=True) == "spawn":
    # Check that error is thrown when inverse are used out of order.
    t = SpatialPadd("image", [10, 5])
    with self.assertRaises(RuntimeError):
        t.inverse(forwards[-1])
```

rijobro:
Seems to be unrelated CUDA error: https://github.com/Project-MONAI/MONAI/pull/1801/checks?check_run_id=2141451226#step:6:7682

wyli:
> @wyli do you understand the error? https://github.com/Project-MONAI/MONAI/pull/1801/checks?check_run_id=2141079751#step:11:5166
> 
> It seems that the following code is being executed on the windows and osx test:
> 
> ```python
> # skip this test if multiprocessing uses 'spawn', as the check is only basic anyway
> if torch.multiprocessing.get_start_method(allow_none=True) == "spawn":
>     # Check that error is thrown when inverse are used out of order.
>     t = SpatialPadd("image", [10, 5])
>     with self.assertRaises(RuntimeError):
>         t.inverse(forwards[-1])  # <-- windows and osx shouldn't reach here, but they do, causing error
> ```

I've changed `allow_none` to False, it'll set a specific start method otherwise I think it's `None`...

behxyz:
> @behxyz, do you think you could try this PR? I'm fairly confident it will have solved the problem.

Sure, let me check this PR before merging.

behxyz:
It works perfectly! Thank you, @rijobro, for the quick fix.

Issue #1601, opened 2021-02-19
Grabcut Implementation


charliebudd:
Grabcut (presented [here](https://cvg.ethz.ch/teaching/cvl/2012/grabcut-siggraph04.pdf)) is a method of segmenting a reigion based on a sparse initial labeling. 

A color model is employed (GMM is common) to determine color distibutions of the labeled reigions. These distributions are then used to asign proberbilities to unlabled regions of the image. Graphcut is then employed to optimize the boundries of the reigions based on the morphological boundries of the original image. This process may be iterated on until convergence.

This will be useful in MONAI for both interactive segmentation and as a post-processing step for segmentation algorithms. 

I see this being split into three PRs. GMM, Graphcut, and finally Grabcut. Each would provide useful functionality induvidually.

tvercaut:
For the record, a good starting point in terms of implementation is the one from @luiscarlosgph 
https://github.com/luiscarlosgph/grabcut
Itself mostly borrowing from that in OpenCV + code from old CUDA samples

tvercaut:
A few more components are now available, with some starting to be used in MONAILabel.

@masadcv implemented a wrapper for OpenCV's graphcut algorithm here:
https://github.com/masadcv/numpymaxflow
There is also a wrapper for pytorch directly but it faces some issues related to packaging (https://github.com/masadcv/torchmaxflow). Ideally, this would be integrated in MONAI proper and the packaging issue addressed in MONAI.

It also seems that the JIT-based compilation used for the GMM code here may not solve all packaging issues, see:
https://github.com/Project-MONAI/MONAILabel/pull/763#discussion_r862096992
Until this is solved, would it be possible to have a pure-python fallback for the GMM? An implementation for it can for example be borrowed from https://github.com/ldeecke/gmm-torch

masadcv:
I may be able to help integrate torchmaxflow into MONAI (if this is something of interest here). 
As @tvercaut pointed out, this may first need packaging to be addressed such that C++ based torch implementations can be packaged and deployed within MONAI.

We have already seen significance of using maxflow algorithm in MONAI Label (https://github.com/Project-MONAI/MONAILabel/pull/731) and I am sure it will be useful for other segmentation processing applications as well.

tvercaut:
[...]

PR #1802, opened 2021-03-18
JIT compiled gaussian mixture model (GMM)

charliebudd:
Introduces a new feature, gaussian mixture modeling. The algorithm takes some initial labeling and uses a mixture of gaussians to approximate each classes distribution in the feature space. Each unlabled element is then asigned a label based on which classes distibution it fits best. 

This is implemented as a c++/cuda extention, but currently it is only implemented in cuda. It may be run with any spatial dimension, channel count, class count, and mixture component count (number of gaussians in each classes mixture).

This PR also introduces JIT compilation for c++/cuda extentions. The source code for this extention is not compiled when monai is installed. When an end user comes to create a GMM python object for the first time, the code will be compiled. This allows us to take parameters passed to the GMM python object and use these as compile time constants in the build allowing the compiler to provide more optimized binaries. A new build is created and stored for each configuration the user asks for. Once a configuration has been built, the JIT system will provide this binary for all subsequent requests.

see #1601 

### Status
**Ready**

### Types of changes


wyli:
thanks!  I think if we could address these and all integration tests work fine then the PR could go into v0.5... 
- currently clang-format is used for `monai/csrc` when we run `./runtests.sh --clangformat (https://github.com/Project-MONAI/MONAI/blob/master/runtests.sh#L137) what's the best way to apply the same for `monai/_extensions`?
- there's a chance that the build dir `path.join(module_dir, "build", build_name)` is not writable by the pip install users, need a fallback option here...
- I think we need to expose all the options of pytorch `load` to `load_module` or via sys. env variable.
- `monai/_extensions` is also a python module path, perhaps rename it to `monai/extensions` or `monai/ext`?

@rijobro @Nic-Ma  @ericspod what do you think?

charliebudd:
Thanks for raising these @wyli. 

clang formatting: yep I can look into adding the extentions directory to the clang formatting step. 

build directory write access: I'm unsure on this, the current build system writes the _C.so object into the monai directory after the build just fine. Is there a distinction I'm missing?

load args: I can see the use of passing additional compiler args beyond just defines as I have implemented. But I'm not sure on ALL the args though.

"_extensions" dir: I used the underscore to signify that it is not intended as a part of our public api. But I'm happy to change it to just extensions if that would be preferred.

wyli:
build dir: the current _C.so is built by setup.py or pip in a temporary dir (`./` or `/tmp/`), and then copied to site-packages during installing

charliebudd:
Well bit of luck I tested it out because I wasn't including the source files in package_data for the distutils setup command (maybe we should explore running our CI in more user like environments?). Anyway, after pip installing monai from my branch I can successfully build the extentions at runtime both in a venv and system install.

wyli:
> Well bit of luck I tested it out because I wasn't including the source files in package_data for the distutils setup command (maybe we should explore running our CI in more user like environments?). Anyway, after pip installing monai from my branch I can successfully build the extentions at runtime both in a venv and system install.

you are right, the current CI doesn't provide enough test cases for packaging and releasing. considering this PR introduces a new top-level module and new JIT capability -- we don't have enough time to test it properly, we should postpone this to v0.6...

monai/_extensions/loader.py
```diff
@@ -0,0 +1,104 @@
+# Copyright 2020 - 2021 MONAI Consortium
```
wyli, line 1: perhaps this could be a generic API in `monai.utils.module`?

monai/_extensions/loader.py
```diff
@@ -0,0 +1,104 @@
...
+from shutil import rmtree
+from sys import platform
+from time import sleep
+
+from torch import cuda
+from torch.utils.cpp_extension import load
+
+dir_path = path.dirname(path.realpath(__file__))
+lock_max_wait = 30
+lock_check_interval = 5
```
wyli, line 24: could these `dir_path`, `lock_max_wait` be arguments of `load_module(...)`, so that `load_module` could be a generic API for MONAI users to compile their own extensions anywhere on PYTHONPATH?

wyli (commented):
thanks -- I put some comments on the `load_module` API. now we have the packaging tests, it's time to revisit this PR and merge it soon...

monai/_extensions/loader.py
charliebudd: I've updated the timeout code and added the timeout limit as an argument to the `load_module` function, This way it can account for extentions with longer compile times where needd. I dont think this function would be useful to MONAI users as its just a wrapper for torch's `cpp_extention.load` which reduces the functionality.

wyli: thanks, the `build_dir` is currently hard-coded as a subfolder of the `module_dir`, is it really needed? [`torch.utils.cpp_extension.load`](https://pytorch.org/docs/stable/cpp_extension.html#torch.utils.cpp_extension.load) has an option of setting the `TORCH_EXTENSIONS_DIR` environment variable. Shall we preserve that behaviour?

monai/_extensions/loader.py
```diff
@@ -0,0 +1,82 @@
...
+    Args:
+        module_name: Name of the module to load. Must match the name of the relevant source directory in the _extensions directory.
+        defines: Dictionary containing names and values of compilation defines.
+        verbose_build: Set to true to enable build logging.
+        build_timeout: Time in seconds before the build will throw an exception to prevent hanging.
+    """
+
+    # Ensuring named module exists in _extensions directory.
+    module_dir = path.join(dir_path, module_name)
+    assert path.exists(module_dir), f"No extention module named {module_name}"
```
wyli, line 48: the assert could be replaced by a proper exception: https://deepsource.io/gh/Project-MONAI/MONAI/run/4102d428-fc5e-4e0e-87a8-26f5490d4cda/python/BAN-B101/

wyli (approved):
thanks, apart from the minor comments and ci errors, the module looks good to me

monai/_extensions/loader.py
charliebudd: good point, changed it

PR #1805, opened 2021-03-18
1804 add pretrain options

### LINKED ISSUE #1804: Add pretrain option for DenseNet and SENet
So far, Class DenseNet and Senet do not have an option to load pretrained weights for specific arch. Users need to call a function, such as `densenet121` and `se_resnet50` to load pretrained weights. It may not be very convenient for some cases that only `Class` type are supported to call.
Let me submit a PR to improve it.

yiheng-wang-nv:

Fixes #1804 .

### Description

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/networks/nets/densenet.py
+++ b/monai/networks/nets/densenet.py
@@ -220,8 +220,9 @@ class DenseNet(nn.Module):
         if arch in model_urls.keys():
             model_url = model_urls[arch]
         else:
-            error_msg = "only densenet121, densenet169 and densenet201 are supported to load pretrained weights."
-            raise AssertionError(error_msg)
+            raise ValueError(
+                "only 'densenet121', 'densenet169' and 'densenet201' are supported to load pretrained weights."
+            )
         pattern = re.compile(
             r"^(.*denselayer\d+)(\.(?:norm|relu|conv))\.((?:[12])\.(?:weight|bias|running_mean|running_var))$"
         )
--- a/monai/networks/nets/densenet.py
+++ b/monai/networks/nets/densenet.py
@@ -115,6 +115,11 @@ class DenseNet(nn.Module):
         bn_size: multiplicative factor for number of bottle neck layers.
             (i.e. bn_size * k features in the bottleneck layer)
         dropout_prob: dropout rate after each dense layer.
+        pretrained: whether to load ImageNet pretrained weights when `spatial_dims == 2`.
+            In order to load weights correctly, Please ensure that the `block_config`
+            is consistent with the corresponding arch.
+        pretrained_arch: the arch name for pretrained weights.
+        progress: If True, displays a progress bar of the download to stderr.
     """
 
     def __init__(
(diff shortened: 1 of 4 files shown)


monai/networks/nets/densenet.py
```diff
@@ -190,43 +198,48 @@ def __init__(
...
+        model_urls = {
+            "densenet121": "https://download.pytorch.org/models/densenet121-a639ec97.pth",
+            "densenet169": "https://download.pytorch.org/models/densenet169-b2777c0a.pth",
+            "densenet201": "https://download.pytorch.org/models/densenet201-c1103571.pth",
+        }
+        if arch in model_urls.keys():
+            model_url = model_urls[arch]
+        else:
+            error_msg = "only densenet121, densenet169 and densenet201 are supported to load pretrained weights."
+            raise AssertionError(error_msg)
```
Nic-Ma, line 224: Suggest to use:
```py
raise ValueError("only 'densenet121', 'densenet169' and 'densenet201' are supported to load pretrained weights.")
```

monai/networks/nets/senet.py
```diff
@@ -176,6 +183,65 @@ def __init__(
...
+            "se_resnext101_32x4d": "http://data.lip6.fr/cadene/pretrainedmodels/se_resnext101_32x4d-3b2fe3d8.pth",
+        }
+        if arch in model_urls.keys():
+            model_url = model_urls[arch]
+        else:
+            error_msg = (
+                "only senet154, se_resnet50, se_resnet101,  se_resnet152, se_resnext50_32x4d "
+                + "and se_resnext101_32x4d are supported to load pretrained weights."
+            )
+            raise AssertionError(error_msg)
```
Nic-Ma, line 208: suggest to use:
```py
raise ValueError(
    "only senet154, se_resnet50, se_resnet101,  se_resnet152, se_resnext50_32x4d, \
    and se_resnext101_32x4d are supported to load pretrained weights."
)
```

Nic-Ma (approved):
Looks good to me, except some minor comments.

Thanks.

monai/networks/nets/senet.py
yiheng-wang-nv: Hi Nic, I just found that if using this kind of line breaks, the printed results will introduce extra spaces. During my latest PR (#1914) I fixed it.

Nic-Ma: Hi @ericspod ,

Could you please help share some comments here? Maybe I misunderstood your previous comments in my PR? You suggested this print: https://github.com/Project-MONAI/MONAI/blob/master/monai/transforms/io/array.py#L147

Thanks.

ericspod: This will replicates the issue since the indentation gets included with the string:

```python
if True:
    s="Hello \
    world"
print(s)  # prints "Hello     world"
```

However this will work as expected, the important difference is to close the strings off with " so the newline escape becomes essentially an instruction to concat strings together:

```python
if True:
    s="Hello "\
    "world"
print(s)  # prints "Hello world"
```

Nic-Ma: I see, thanks.

Issue #315, opened 2020-04-27
Fully-connected 3D CRF


LucasFidon:
**Is your feature request related to a problem? Please describe.**
3D CRF have been widely used as post-processing in segmentation applications.

**Describe the solution you'd like**
We have made publically available an implementation of a fully-connected 3D CRF in PyTorch:
https://github.com/SamuelJoutard/Permutohedral_attention_module

This could be added directly to MONAI.

The 3D CRF itself is based on the existing implementation in NiftyNet, and it relies on an efficient implementation of gaussian filtering in cuda. Our implementation is several times faster than the NiftyNet implementation.

wyli:
looks really nice, perhaps could start with the pure pytorch version? we haven't figured out how to package/distribute cuda kernels in monai. (have you tried to contribute the kernels to pytorch directly?)

LucasFidon:
@wyli yes starting with the pure PyTorch version in the folder PAM sounds good.
It would only require to change one of the import in the crf.py file.

As a motivation for future integration of the cuda code, I have measured that the CRF with cuda version is 5x faster than the pure PyTorch version on my machine at inference for a volume 136x136x136.

Regarding the cuda kernels integration I will let @SamuelJoutard reply as he is the expert on this.

SamuelJoutard:
@wyli No I did not try to submit the kernel to Pytorch because the HashTable is not dynamically allocated in terms of memory (the kernel operates on a pre-allocated "relatively large" amount of memory) and this could be improved in my opinion. 

In addition to the speed-up @LucasFidon  mentioned, the Cuda version properly handles collisions in the HashTable which could marginally improve the results.

tvercaut:
This issue hinges on #785

wyli:
as #785 is now fixed, do you want to revisit this ticket? @LucasFidon @SamuelJoutard?
the cpp/cuda codes are organised in https://github.com/Project-MONAI/MONAI/tree/master/monai/csrc

currently (as of [monai v0.3](https://github.com/Project-MONAI/MONAI/tree/0.3.0)) in a git-cloned MONAI codebase, running:
- `BUILD_MONAI=1 python setup.py develop` will compile all the CPP/CUDA code in `monai/csrc`
- `sh runtests.sh --clean` will uninstall and clean up the *.so files
- `sh runtests.sh --clangformat` will run auto code formatting

tvercaut:
@charliebudd-kcl will be able to help.
[...]

Issue #94, opened 2020-02-19
Port deepmedic


wyli:
a subtask of #78, port deepmedic

ref:
https://niftynet.readthedocs.io/en/dev/niftynet.network.deepmedic.html

tvercaut:
This might warrant a separate issue but it's probably worth highlignting that the deepmedic paper relies of a post-processing by a dense CRF. The dense CRF is implemented through a permutohedral lattice approach. Although deemedic itself does not rely on backpropagating through the CRF, other papers do (e.g. CRF-RNN) and they also rely on the same permutohedral lattice approach.

As such, it would be good to have a generic permutohedral lattice / CRF-RNN module in MONAI.

A pytorch implementation of the CRF-RNN can be found here but note that the permutohedral lattice does not run on GPU there:
https://github.com/sadeepj/crfasrnn_pytorch/tree/master/crfasrnn
https://github.com/sadeepj/crfasrnn_pytorch/blob/master/crfasrnn/permutohedral.cpp

The corresponding NiftyNet module is here but lacks some GPU ops and cannot back-propagate through features:
https://niftynet.readthedocs.io/en/dev/niftynet.layer.crf.html

An alternative implementation of a GPU-based backpropagatable permutohedral lattice can be found here:
https://github.com/SamuelJoutard/Permutohedral_attention_module

This might also be interesting:
https://github.com/MiguelMonteiro/permutohedral_lattice

brosscle:
Hello,
I would like to use and play with the network BLAST (https://github.com/biomedia-mira/blast-ct) which strongly relies on the Deepmedic architecture. Since I discovered MONAI a few days ago, I would like to know if any of you advanced in the portage of Deepmedic in MONAI. I guess I can spend some time and contribute if the portage is not already completed.
Thanks !

wyli:
Hi @brosscle currently there's no plan from the core dev team to port this architectures, please feel free to submit a PR.

for your (or anyone interested in contributing) reference, there are a few examples about using the relevant monai APIs:
- notebook: https://github.com/Project-MONAI/MONAIBootcamp2020/blob/master/day1notebooks/lab3_networks.ipynb
- basic unet example: https://github.com/Project-MONAI/MONAI/blob/master/monai/networks/nets/basic_unet.py
- a simple unit test for the network class: https://github.com/Project-MONAI/MONAI/blob/master/tests/test_basic_unet.py

[...]

PR #1806, opened 2021-03-18
Conditional Random Field

charliebudd:
Adds a new feature, Conditional Random Field (CRF). Combines message passing with a class compatability convolution into an iterative process designed to successively minimise the energy of the class labeling. 

see #94 and #315

### Status
**Ready**

### Types of changes


wyli:
/black

thanks I'll run some tests locally and get back to this PR

tests/test_crf_cpu.py
```diff
@@ -0,0 +1,450 @@
...
+    @parameterized.expand(TEST_CASES)
+    def test(self, test_case_description, params, input, features, expected):
+
+        # Create input tensors
+        input_tensor = torch.from_numpy(np.array(input)).to(dtype=torch.float, device=torch.device("cpu"))
+        feature_tensor = torch.from_numpy(np.array(features)).to(dtype=torch.float, device=torch.device("cpu"))
+
+        # apply filter
+        crf = CRF(*params)
+        output = crf(input_tensor, feature_tensor).cpu().numpy()
```
wyli, line 443: So how to use it in backprop? I'm trying to do this, but get some typeerrors (`TypeError: phl_filter(): incompatible function arguments. The following argument types are supported:
    1. (arg0: at::Tensor, arg1: at::Tensor) -> at::Tensor`)
```suggestion
        input_tensor.requires_grad = True
        input_tensor.retain_grad()
        output = crf(input_tensor, feature_tensor)
        output.sum().backward()
        output = output.cpu().numpy()
```

charliebudd: A good point, I think the PHLFilter's back propergation is only implemented for one of the input tensors, which is an oversight. I'll get working on the fix, I can either add the changes to this or make a new PR?

wyli: the current impl. is enough for creating this tutorial https://github.com/NifTK/NiftyNet/blob/dev/demos/crf_as_rnn/crf_as_rnn_inference_demo.ipynb shall we merge this first? but it should raise a NotImpelmentedError when we run backprop

charliebudd: I had intended to move on to investigating the back propergation after it was in so that suits me. Ill make it raise an error for now and open a new PR for the fix.

monai/networks/blocks/crf.py
```diff
@@ -0,0 +1,140 @@
...
+        # expanding matrix to kernel
+        compatability_kernel = _expand_matrix_to_kernel(
+            compatability_matrix, spatial_dim, self.compatability_kernel_range
+        )
+
+        # choosing convolution function
+        conv = [conv1d, conv2d, conv3d][spatial_dim - 1]
+
+        # seting up output tensor
+        output_tensor = softmax(input_tensor, dim=1)
```
wyli, line 103: just for clarification... I remember there were some normalisation factors https://github.com/SamuelJoutard/Permutohedral_attention_module/blob/master/CRF/crf.py#L103, probably from https://github.com/cvlab-epfl/densecrf/blob/c16e3c50d214ab8b4727c0d471f888240020d757/libDenseCRF/densecrf.cpp#L421, but couldn't recall the details... are they implemented here?

wyli (approved):
thanks! cc @SamuelJoutard could you please help review this one? I'll merge it as an initial version, but we'll have follow-up, would be great to get your inputs

monai/networks/blocks/crf.py
charliebudd: This was a conscious choice. In my implementation I found them to have little effect on the output. But worth noting as a divergence from previous implementations.

PR #1813, opened 2021-03-19
PadListDataCollate transform

rijobro:
Addresses https://github.com/Project-MONAI/MONAI/pull/1795#discussion_r597585725.

part of #1515 

### Description
Create transform that in the forward direction takes a batch of data and centrally pads any tensors to the maximum size of each dimension of all images in the batch.

The inverse acts on a dictionary containing the data for 1 dataset. 

I find the asymmetry between the forward and inverse a little ugly, but I can't really see any other way about it. The only reason we need the whole batch in the forward direction is so that we can find out the required padding size. However, in the inverse direction, we want to deal with them individually so that we can take advantage of multiprocessing.

I should note that the asymmetry is mostly hidden from the user who wouldn't call this directly. Instead, they would pass the forward as the `collate_fn` for a `DataLoader`, and then the `inverse` would be called automatically when doing batch inversion.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -337,64 +337,25 @@ def pad_list_data_collate(
     mode: Union[NumpyPadMode, str] = NumpyPadMode.CONSTANT,
 ):
     """
-    Same as MONAI's ``list_data_collate``, except any tensors are centrally padded to match the shape of the biggest
-    tensor in each dimension.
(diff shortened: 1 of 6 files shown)


wyli (approved):
thanks!

PR #1817, opened 2021-03-22
remove .value from InverseKeys enum

rijobro:
`InverseKeys` enum only ever gets used as string. So cleaner to not inherit from enum, therefore not needing `.value` throughout code.

### Status
**Ready**

### Types of changes


wyli (approved):
thanks!

PR #1825, opened 2021-03-22
refactor CommonKeys

rijobro:
Move `CommonKeys` to `enums.py` so that it can be used by the whole MONAI codebase. Addresses problem in this PR: https://github.com/Project-MONAI/MONAI/pull/1794#issuecomment-804013257.

### DIFF
--- a/monai/apps/deepgrow/interaction.py
+++ b/monai/apps/deepgrow/interaction.py
@@ -13,9 +13,9 @@ from typing import Callable, Dict, Sequence, Union
 import torch
 
 from monai.engines import SupervisedEvaluator, SupervisedTrainer
-from monai.engines.utils import CommonKeys
 from monai.engines.workflow import Events
 from monai.transforms import Compose
+from monai.utils.enums import CommonKeys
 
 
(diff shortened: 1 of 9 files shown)


wyli (approved):
thanks, ~this is a breaking change moves `CommonKeys` from `monai.engines.utils` to `monai.utils.enums`~ this is non-breaking

rijobro:
How is this a breaking change? Do you think many people would be importing CommonKeys in their own code?

wyli:
yes, people are building customised workflows similar to deepgrow  https://github.com/Project-MONAI/MONAI/pull/1825/files#diff-41ea322a0a24b98b7e0366343405c3e4e61716b2639f0acbaecea2079e74d127R18

rijobro:
I see. Well at least the old import still works: `from monai.engines.utils import CommonKeys` (because it still gets imported into `monai/engines/utils.py`).

wyli:
you're right, this is non-breaking, and we have access to `CommonKeys` even with the min. system setup (previously we can't have that)

PR #1826, opened 2021-03-23
Jupyter followup

ericspod:
### Description
Adds unit tests for MetricLogger and the plotting functions in jupyter_utils.py. These are tested by testing `ThreadContainer.plot_status` which will call all the plot functions currently present. The test involves saving a plot to a file and comparing against a stored file, this nails down what the plot is expected to create but doesn't test whether the plot is correct. 

### Status
**Ready**

### Types of changes


wyli (approved):
thanks!

PR #1830, opened 2021-03-23
fixes tutorial issue 150

wyli:

Fixes https://github.com/Project-MONAI/tutorials/issues/150

### Description
added some 2D testing cases

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/dictionary.py
+++ b/monai/transforms/spatial/dictionary.py
@@ -1285,7 +1285,7 @@ class RandRotated(RandomizableTransform, MapTransform, InvertibleTransform):
         d = dict(data)
         if not self._do_transform:
             for key in self.keys:
-                self.push_transform(d, key, extra_info={"rot_mat": np.eye(4)})
+                self.push_transform(d, key, extra_info={"rot_mat": np.eye(d[key].ndim)})
             return d
         angle: Union[Sequence[float], float] = self.x if d[self.keys[0]].ndim == 3 else (self.x, self.y, self.z)
         rotator = Rotate(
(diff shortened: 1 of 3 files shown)


monai/transforms/spatial/dictionary.py
```diff
@@ -1285,7 +1285,7 @@ def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.nda
         d = dict(data)
         if not self._do_transform:
             for key in self.keys:
-                self.push_transform(d, key, extra_info={"rot_mat": np.eye(4)})
+                self.push_transform(d, key, extra_info={"rot_mat": np.eye(d[key].ndim)})
```
rijobro, line 1288: Should this be ndim+1? Not sure if it's purely rotation matrix or also translation component.

rijobro: Sorry, thought about this a bit more, and it makes sense.

tests/test_inverse_collation.py
```diff
@@ -53,6 +53,19 @@
...
 
+TESTS_2D = [
+    (t.__class__.__name__ + (" pad_list_data_collate" if collate_fn else " default_collate"), t, collate_fn, 2)
+    for collate_fn in [None, pad_list_data_collate]
+    for t in [
+        RandFlipd(keys=KEYS, spatial_axis=[1]),
+        RandAxisFlipd(keys=KEYS),
+        RandRotate90d(keys=KEYS, spatial_axes=(0, 1)),
+        RandZoomd(keys=KEYS, prob=0.5, min_zoom=0.5, max_zoom=1.1, keep_size=True),
+        RandRotated(keys=KEYS, range_x=np.pi),
```
rijobro, line 64: By default probability is 0.1, so for batch size of 7, there's no guarantee of a mix of `True` and `False` for `_do_transform`. Should probably just set `prob=0.5` for all transforms.

rijobro (approved):

monai/transforms/spatial/dictionary.py
wyli: yes it's with the channel dimension...

tests/test_inverse_collation.py
wyli: good point, I'll revise...

PR #1842, opened 2021-03-24
1822 Implement Prob NMS

### LINKED ISSUE #1822: Prob Non-maximal Suppression Implementation
@behxyz has provided the startup code about the implementation of Prob Non-maximal Suppression ( #1800 ), I will continue to that work and finish this feature.

yiheng-wang-nv:

Fixes #1822 .

### Status
**Ready**

### Types of changes


yiheng-wang-nv:
Hi @behxyz , could you please help to check the calculation first. I just tested locally, and will add formal test cases and doc string later.

Compared with the startup code in #1800 , this version enhances:

1. Optimize the calculation via using more numpy based methods, now for a same 2D input probability map, the calculation speed is around 3 times faster.
2. Remove skimage dependency.
3. Supports any dimension inputs.

monai/utils/radial_nms.py
```diff
@@ -0,0 +1,55 @@
...
+    def __call__(self, probs_map: Union[np.ndarray, torch.Tensor], level: int):
+        if self.sigma > 0:
+            if isinstance(probs_map, np.ndarray):
+                probs_map = torch.as_tensor(probs_map, dtype=torch.float)
+            device = (
+                torch.device("cuda")
+                if (probs_map.device == "cuda" and torch.cuda.is_available())
+                else torch.device("cpu:0")
+            )
+            self.filter.to(device)
```
wyli, line 27: I think this could be simplified to 
```suggestion
            self.filter.to(probs_map)
```

behxyz: Exactly, this way it uses whatever device that `prob_maps` is residing on.

monai/utils/radial_nms.py
```diff
@@ -0,0 +1,55 @@
...
+        while np.max(probs_map) > self.prob_threshold:
+            max_idx = np.unravel_index(probs_map.argmax(), probs_map_shape)
+            prob_max = probs_map[max_idx]
+            max_idx_arr = np.asarray(max_idx)
+            coord_wsi = ((max_idx_arr + 0.5) * resolution).astype(int)
+            outputs.append([prob_max] + list(coord_wsi))
+
+            # achieve min index for each dimension
+            idx_min_range = max_idx_arr - self.radius
+            idx_min_range[idx_min_range < 0] = 0
```
behxyz, line 44: It is clearer to use `np.clip(idx_min_range, 0)` instead.

monai/utils/radial_nms.py
```diff
@@ -0,0 +1,55 @@
...
+
+            # achieve max index for each dimension
+            idx_upper_bound = np.asarray(probs_map_shape)
+
+            idx_max_range = max_idx_arr + self.radius
+            idx_max_range[idx_max_range > idx_upper_bound] = idx_upper_bound[idx_max_range > idx_upper_bound]
```
behxyz, line 50: It is clearer and more concise to use `np.clip(idx_max_range, probs_map_shape)` instead.

monai/utils/radial_nms.py
```diff
@@ -0,0 +1,55 @@
...
+            max_idx_arr = np.asarray(max_idx)
+            coord_wsi = ((max_idx_arr + 0.5) * resolution).astype(int)
+            outputs.append([prob_max] + list(coord_wsi))
+
+            # achieve min index for each dimension
+            idx_min_range = max_idx_arr - self.radius
+            idx_min_range[idx_min_range < 0] = 0
+
+            # achieve max index for each dimension
+            idx_upper_bound = np.asarray(probs_map_shape)
```
behxyz, line 47: This will be unnecessary if using `np.clip`

monai/utils/radial_nms.py
yiheng-wang-nv: Thanks! I took https://github.com/Project-MONAI/MONAI/blob/master/tests/test_gaussian_filter.py#L41 for reference. Do I need to modify that place too?

wyli: I don't think they are the same, that one is to select based on a string, here is based on an existing variable...

behxyz (changes requested):
Well done, @yiheng-wang-nv! 
Apart from some minor changes, all looks good to me. So please add docstrings and unnittests, and move on from this to implement how to actually use it in conjunction with other components during inference.

behxyz:
@yiheng-wang-nv, maybe we should change the name from RadialNMS to PixelNMS or something. Because although we are receiving the radius as an input, we are suppressing over a square (or a cube) and not a circle (or a sphere). Unless, we use a radial base to suppress the pixels within the given radius.

yiheng-wang-nv:
> @yiheng-wang-nv, maybe we should change the name from RadialNMS to PixelNMS or something. Because although we are receiving the radius as an input, we are suppressing over a square (or a cube) and not a circle (or a sphere). Unless, we use a radial base to suppress the pixels within the given radius.

Agree, do you think we should also change the name of `radius`?

monai/utils/pixel_nms.py
```diff
@@ -0,0 +1,83 @@
...
+            prob_max = probs_map[max_idx]
+            max_idx_arr = np.asarray(max_idx)
+            coord_wsi = ((max_idx_arr + 0.5) * resolution).astype(int)
+            outputs.append([prob_max] + list(coord_wsi))
+
+            idx_min_range = (max_idx_arr - self.pixel_dis).clip(0, None)
+            idx_max_range = (max_idx_arr + self.pixel_dis).clip(None, probs_map_shape)
+            # for each dimension, set values during index ranges to 0
+            slices = tuple([slice(idx_min_range[i], idx_max_range[i]) for i in range(self.spatial_dims)])
+            probs_map[slices] = 0
```
wyli, line 81: if `probs_map` is a numpy array it'll be modified, but if it's a torch tensor it won't be modified (because of line 64), shall we make a copy of `probs_map` to avoid in-place modifications?

monai/utils/pixel_nms.py
```diff
@@ -0,0 +1,83 @@
...
+            probs_map = self.filter(probs_map)
+
+        if isinstance(probs_map, torch.Tensor):
+            probs_map = probs_map.detach().cpu().numpy()
+
+        probs_map_shape = probs_map.shape
+        resolution = pow(2, resolution_level)
+
+        outputs = []
+        while np.max(probs_map) > self.prob_threshold:
```
wyli, line 70: if `self.prob_threshold < 0` this may run into an infinite loop

monai/utils/pixel_nms.py
```diff
@@ -0,0 +1,83 @@
...
+            max_idx = np.unravel_index(probs_map.argmax(), probs_map_shape)
+            prob_max = probs_map[max_idx]
+            max_idx_arr = np.asarray(max_idx)
+            coord_wsi = ((max_idx_arr + 0.5) * resolution).astype(int)
```
wyli, line 74: this `resolution` step seems to be pathology specific, shall we remove it from this generic utility? also the name of the method seems to suggest 2D (pixels but not voxels), perhaps it could be `ProbNMS` because it only works for probability scores

wyli (commented):
thanks I put some comments inline

monai/utils/pixel_nms.py
yiheng-wang-nv: Hi @behxyz , what do you think? Remove `resolution` should be fine and we can use some post processing method to do it right?

behxyz: Hi, removing the resolution requires the user to manually change the resolution before calling it and then change it back after calling it, which prone to error. Maybe we should put it as an option or so.

wyli: I feel putting a resolution option wouldn't resolve any issue without having access to the full meta data of the input slide...we can add a resolution based wrapper in the `monai.apps.pathology` module, but this generic module should be decoupled from that concept. what do you think?

behxyz: @yiheng-wang-nv @wyli 
Although it doesn't require the access to meta data, it needs the knowledge of the resolution the inference has performed at. So I agree, we can make this one a generic `ProbNMS` without resolution and add another one for pathology in `monai.apps.pathology` to handle.

monai/utils/pixel_nms.py
```diff
@@ -0,0 +1,83 @@
...
+    Return:
+        a list of selected lists, where inner lists contain probability and coordinates.
+        For example, for 3D input, the inner lists are in the form of [probability, x, y, z].
+
+    """
+
+    def __init__(
+        self,
+        spatial_dims: int,
+        sigma: Union[Sequence[float], float, Sequence[torch.Tensor], torch.Tensor] = 0,
```
behxyz, line 36:
```suggestion
        sigma: Union[Sequence[float], float, Sequence[torch.Tensor], torch.Tensor] = 0.0,
```

monai/utils/pixel_nms.py
```diff
@@ -0,0 +1,83 @@
...
+    Performs pixel based non-maximum suppression (NMS) on the probabilities map via
+    iteratively selecting the coordinate with highest probability and then move it as well
+    as its surrounding values. The remove range is determined by the parameter `pixel_dis`.
+    If multiple coordinates have the same highest probability, only one of them will be
+    selected.
+
+    Args:
+        spatial_dims: number of spatial dimensions of the input probabilities map.
+        sigma: the standard deviation for gaussian filter.
+            It could be a single value, or `spatial_dims` number of values. Defaults to 0.
```
behxyz, line 20:
```suggestion
            It could be a single value, or `spatial_dims` number of values. Defaults to 0.0.
```

tests/test_pixel_nms.py
```diff
@@ -0,0 +1,107 @@
...
+
+import unittest
+
+import numpy as np
+import torch
+from parameterized import parameterized
+
+from monai.utils import PixelNMS
+
+probs_map_1 = np.random.rand(100, 100).clip(0, 0.5)
```
behxyz, line 20:
```suggestion
probs_map_1 = np.random.rand(100, 100)
```
Is there any reason not to keep the probability maps between 0 and 1? In the case of this NMS function, each pixel/voxel is an independent variable.

behxyz:
@yiheng-wang-nv
Could you please finalize this PR without `resolution` and add another pathology specific one with `resolution` in `monai/apps/pathology/`?

monai/utils/pixel_nms.py
```diff
@@ -0,0 +1,83 @@
...
+        For example, for 3D input, the inner lists are in the form of [probability, x, y, z].
+
+    """
+
+    def __init__(
+        self,
+        spatial_dims: int,
+        sigma: Union[Sequence[float], float, Sequence[torch.Tensor], torch.Tensor] = 0,
+        prob_threshold: float = 0.5,
+        pixel_dis: int = 24,
```
behxyz, line 38:
```suggestion
        box_size: int = 48,
```
To make this NMS more generic, could you please change ``pixel_dis` to `box_size` and use a multi-dimension (the same number of dimension as `spatial_dims`). So it would allow for ` box_size = (10, 20)`, ` box_size = (10, 20, 30)`, and if one number is provided makes it tuple of length `spatial_dims`,

behxyz: Just be aware that `box_size` at each dimension is twice as `radius`.

monai/apps/pathology/utils.py
```diff
@@ -0,0 +1,64 @@
...
+            coord_wsi = ((max_idx_arr + 0.5) * resolution).astype(int)
+            outputs.append([prob_max] + list(coord_wsi))
+
+            idx_min_range = (max_idx_arr - self.box_lower_bd).clip(0, None)
+            idx_max_range = (max_idx_arr + self.box_upper_bd).clip(None, probs_map_shape)
+            # for each dimension, set values during index ranges to 0
+            slices = tuple([slice(idx_min_range[i], idx_max_range[i]) for i in range(self.spatial_dims)])
+            probs_map[slices] = 0
+
+        return outputs
```
wyli, line 64: is it possible to do something like (perhaps need to remove `astype(int)` in `ProbNMS`:
```suggestion
        resolution = pow(2, resolution_level)
        outputs = ProbNMS.__call__(self, resolution)
        # update outputs with resolution here
        return outputs
```

yiheng-wang-nv: thus still add a parameter `resolution ` in `ProbNMS ` right?

wyli: sorry that's a typo, I update it

wyli (approved):
thanks!

behxyz (approved):

PR #1845, opened 2021-03-25
Fully convolutional models based on TorchVision models

behxyz:
### Description
This PR build fully convolutional models based on any compatible `torchvision` model and convert the adaptive average pooling layer to a 7x7 average pooling. This model allows for using pre-trained `torchvision` models with any number of output classes, and most importantly enable a much faster inference in use cases like pathology where the input images are huge.

### Status
**Ready**

### Types of changes

- [x ] Documentation updated, tested `make html` command in the `docs/` folder.


monai/networks/nets/torchvision_fc.py
```diff
@@ -0,0 +1,43 @@
...
+        model = getattr(models, model_name)(pretrained=pretrained)
+
+        # remove last fully connected (FC) layer and adaptive avg pooling
+        self.features = torch.nn.Sequential(*list(model.children())[:-2])
+
+        # add 7x7 avg pooling (in place of adaptive avg pooling)
+        self.pool = torch.nn.AvgPool2d(7, stride=1)
+
+        # add 1x1 conv (it behaves like a FC layer)
+        self.fc = torch.nn.Conv2d(model.fc.in_features, n_classes, kernel_size=(1, 1))
```
wyli, line 30: perhaps if the n_class is the same as the pretrained, we should also initialise the conv kernels using the pretrained model fc layers?

monai/networks/nets/torchvision_fc.py
```diff
@@ -0,0 +1,43 @@
...
+models, _ = optional_import("torchvision.models")
+
+
+class TorchVisionFullyConvModel(torch.nn.Module):
+    """
+    Customize TorchVision models to replace fully connected layer by convolutional layer.
+
+    Args:
+        model_name: name of any torchvision with adaptive avg pooling and fully connected layer at the end.
+            Default to "resnet18".
```
wyli, line 14: what are the supported options? perhaps provide a few examples other than `resnet18`

monai/networks/nets/torchvision_fc.py
```diff
@@ -0,0 +1,43 @@
...
+        x = self.features(x)
+
+        # apply 7x7 avg pooling
+        x = self.pool(x)
+
+        # apply last 1x1 conv layer that act like fc layer
+        x = self.fc(x)
+
+        # remove the color channel
+        x = x.squeeze(1)
```
wyli, line 42: this is treating `n_classes==1` differently from `n_classes > 1`, shall we make all these cases consistent here?

monai/networks/nets/torchvision_fc.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def __init__(self, model_name: str = "resnet18", n_classes: int = 1, pretrained: bool = False):
+        super().__init__()
+        model = getattr(models, model_name)(pretrained=pretrained)
+
+        # remove last fully connected (FC) layer and adaptive avg pooling
+        self.features = torch.nn.Sequential(*list(model.children())[:-2])
+
+        # add 7x7 avg pooling (in place of adaptive avg pooling)
+        self.pool = torch.nn.AvgPool2d(7, stride=1)
```
wyli, line 27: the pooling size should be configurable

wyli (commented):

wyli (approved):
thanks! (the `ValueError` could be `NotImplementedError`, but feel free to keep the current one)

monai/networks/nets/torchvision_fc.py
wyli: this could be left for future work

behxyz: It's an easy fix but I thought to treat different class numbers equally. Also in MONAI people would use 1000 classes rarely if at all.

wyli: sure, agreed, I didn't know that all of models are of 1000 classes

PR #1863, opened 2021-03-26
1862 Fix checkpoint read permission

### LINKED ISSUE #1862: Checkpoint handler save files with wrong permissions
**Describe the bug**
We always run programs in dockers, usually it's root user, and found an issue that: pure PyTorch program saves model(torch.save) and has the read permission for non-root users, but ignite Checkpoint handler saves model and only has read permission for root users.
Reproduce this issue by below 2 MONAI examples:
1. https://github.com/Project-MONAI/tutorials/blob/master/3d_segmentation/torch/unet_training_dict.py
The saved model will be:
```
-rw-r--r-- 1 root root 19253862 Mar 26 02:08 best_metric_model_segmentation3d_dict.pth
```
2. https://github.com/Project-MONAI/tutorials/blob/master/3d_segmentation/ignite/unet_training_dict.py
The saved model will be:
```
-rw------- 1 root root 57738580 Mar 26 01:54 net_checkpoint_50.pt
```

Nic-Ma:
Fixes #1862 .

### Description
This PR fixed the read permission issue of checkpoint, based on suggestion from @vfdev-5 in https://github.com/pytorch/ignite/issues/1876.

### Status
**Ready**

### Types of changes


monai/handlers/checkpoint_saver.py
```diff
@@ -116,7 +116,8 @@ class _DiskSaver(DiskSaver):
             """
 
             def __init__(self, dirname: str, filename: Optional[str] = None):
-                super().__init__(dirname=dirname, require_empty=False)
+                # set `atomic=False` as `atomic=True` may save files without read permission for non-root users
+                super().__init__(dirname=dirname, require_empty=False, atomic=True)
```
rijobro, line 120: Comment (`atomic=False`) doesn't seem to correspond to code (`atomic=True`)?

Nic-Ma: Sorry for my typo, I changed the code to double confirm and forgot to change back...

monai/handlers/checkpoint_saver.py
```diff
@@ -116,7 +116,8 @@ class _DiskSaver(DiskSaver):
             """
 
             def __init__(self, dirname: str, filename: Optional[str] = None):
-                super().__init__(dirname=dirname, require_empty=False)
+                # set `atomic=False` as `atomic=True` may save files without read permission for non-root users
```
vfdev-5, line 119: More exactly, it gives read/write permission to the user who saved the file, without group/others read permission, i.e. `-rw-------`, not related to root or not-root user...

wyli (approved):
thanks, we'll revert this after the upstream fix release

monai/handlers/checkpoint_saver.py
Nic-Ma: Hi @vfdev-5 , thanks for your review. I updated the comments.

PR #1885, opened 2021-03-29
1821 add slicing index and negative index for datasets

### LINKED ISSUE #1821: flexible indexing for CacheDataset
currently the cachedataset doesn't work well with negative index and slicing, e.g.:
```py
train_data = MedNISTDataset(root_dir="./", section="training", download=True, transform=None)
train_data[1:3]
```
result:
```traceback
Loading dataset:   0%|          | 0/47164 [00:00<?, ?it/s]Verified 'MedNIST.tar.gz', md5: 0bc7306e7427e00ad1c5526a6677552d.
file ./MedNIST.tar.gz exists, skip downloading.
extracted file ./MedNIST exists, skip extracting.
Loading dataset: 100%|██████████| 47164/47164 [00:00<00:00, 139781.09it/s]
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-22-24e3dd26d1a0> in <module>()
      1 train_data = MedNISTDataset(root_dir="./", section="training", download=True, transform=None)
----> 2 train_data[1:3]

/usr/local/lib/python3.7/dist-packages/monai/data/dataset.py in __getitem__(self, index)
    531 
    532     def __getitem__(self, index):
--> 533         if index >= self.cache_num:
    534             # no cache for this index, execute all the transforms directly
    535             return super(CacheDataset, self).__getitem__(index)

TypeError: '>=' not supported between instances of 'slice' and 'int'
```
it seems `train_data[-1]` is also not working properly.

supporting the usual sequence indexing and slicing syntax

context:
https://github.com/Project-MONAI/MONAI/blob/5151b15aaadc90bf578729d85e8011faae5d1836/monai/data/dataset.py#L533

Nic-Ma:
Fixes #1821 .

### Description
This PR enhanced datasets to support slicing index and negative index.

### Status
**Ready**

### Types of changes


monai/data/dataset.py
```diff
@@ -66,19 +68,24 @@ def __init__(self, data: Sequence, transform: Optional[Callable] = None) -> None
...
-            data = apply_transform(self.transform, data)
+    def _transform(self, index: int):
+        return apply_transform(self.transform, self.data[index]) if self.transform is not None else self.data[index]
 
-        return data
+    def __getitem__(self, index: int):
+        if isinstance(index, slice):
+            start, stop, step = index.indices(len(self))
+            index_list = list(range(start, stop, step))
+            return [self._transform(i) for i in index_list]
```
wyli, line 78: I thought about the same but logically this should return a subset dataset `Dataset(data=self.data[slice], transform=transform)`. what would be the expected behaviour for the cached version? now I'm not sure... cc @rijobro  @ericspod

Nic-Ma: @rijobro @ericspod , could you please help share some comments here?

Thanks in advance.

ericspod: We should use [Subset](https://pytorch.org/docs/stable/data.html#torch.utils.data.Subset) for getting subsets of existing datasets, I don't think we should necessarily support slice in this way.

Nic-Ma: Hi @ericspod ,

Sounds good, let me update the PR to return a `Subset` instead.
Thanks.

Nic-Ma:
Hi @wyli and @ericspod ,

I updated the PR according to both of your comments.

Thanks in advance.

monai/data/dataset.py
```diff
@@ -74,8 +76,8 @@ def _transform(self, index: int):
     def __getitem__(self, index: int):
```
ericspod, line 76: The type of `index` and the return type do need to be adjusted now. The documentation should also state that the return value relies on whether `index` is a single value or a slice.

ericspod: We should also note explicitely that our behaviour is not diverging from Pytorch's `Dataset` class in this specific way, up to now the two were substitutable.

Nic-Ma: Sure, let me update the doc-strings.
Thanks.

Nic-Ma: Hi @ericspod , added doc-strings for this enhancement: https://github.com/Nic-Ma/MONAI/blob/1821-add-slicing-datasets/monai/data/dataset.py#L48

Thanks.

ericspod (approved):

monai/data/dataset.py
```diff
@@ -66,19 +70,25 @@ def __init__(self, data: Sequence, transform: Optional[Callable] = None) -> None
...
-        if self.transform is not None:
-            data = apply_transform(self.transform, data)
+    def _transform(self, index: int):
+        return apply_transform(self.transform, self.data[index]) if self.transform is not None else self.data[index]
 
-        return data
+    def __getitem__(self, index: Union[int, slice]):
+        if isinstance(index, slice):
+            start, stop, step = index.indices(len(self))
+            indices = list(range(start, stop, step))
```
wyli, line 79: perhaps no need to call `list()`

Nic-Ma: Thanks for your update!

PR #1893, opened 2021-03-29
Pathology Probability Map Generator

behxyz:
### Description
This PR implements the handler requires to get pathology patch inference outputs, generate probability maps, and save them.

### Status
**Work in progress**

### Types of changes


monai/apps/pathology/handlers.py
```diff
@@ -0,0 +1,99 @@
...
+from monai.utils import exact_version, optional_import
+
+Events, _ = optional_import("ignite.engine", "0.4.4", exact_version, "Events")
+if TYPE_CHECKING:
+    from ignite.engine import Engine
+else:
+    Engine, _ = optional_import("ignite.engine", "0.4.4", exact_version, "Engine")
+
+
+class ProbMapGenerator:
```
wyli, line 16: generator is also a generic type in python, shall we rename this to ProbMapSaver?

monai/apps/pathology/handlers.py
```diff
@@ -0,0 +1,99 @@
...
+        self.num_done_images: int = 0
+        self.num_images: int = 0
+
+    def attach(self, engine: Engine) -> None:
+        """
+        Args:
+            engine: Ignite Engine, it can be a trainer, validator or evaluator.
+        """
+
+        self.num_images = len(engine.data_loader.dataset.data_list)
```
wyli, line 47: would `len(engine.data_loader.dataset)` work?

monai/apps/pathology/handlers.py
```diff
@@ -0,0 +1,99 @@
...
+
+class ProbMapGenerator:
+    """
+    Event handler triggered on completing every iteration to save the probability map
+    """
+
+    def __init__(
+        self,
+        output_dir: str = "./",
+        name: Optional[str] = None,
```
wyli, line 24: shall we have a `dtype` arg for the saved data precision? and `output_postfix` like https://github.com/Project-MONAI/MONAI/blob/43f482c4ace9cbbc1da50d994edf389ec17f1096/monai/handlers/segmentation_saver.py#L53

monai/apps/pathology/handlers.py
```diff
@@ -0,0 +1,99 @@
...
+        self.logger.info(f"Inference of '{name}' is done [{self.num_done_images}/{self.num_images}]!")
+        del self.prob_map[name]
+        del self.counter[name]
+        del self.level[name]
+
+    def finalize(self, engine: Engine):
+        if self.counter:
+            raise RuntimeError(f"Counter: {self.counter}")
+        else:
+            self.logger.info(f"Probability map is created for {self.num_done_images}/{self.num_images} images!")
```
wyli, line 99: Unnecessary else after `raise`
```suggestion
        self.logger.info(f"Probability map is created for {self.num_done_images}/{self.num_images} images!")
```

wyli (commented):
thanks!

monai/apps/pathology/handlers.py
behxyz: Actually, the `if` is not necessary anymore because it should happen only if there is an implementation error, which will cause an error somewhere else, so I removed it.

monai/apps/pathology/handlers.py
wyli: ok then I think the dataset should provide an interface such as `get_num_images` otherwise this line is coupled with the dataset impl. (btw is `images` the same as `samples`?)

yiheng-wang-nv (approved):
Thanks, it looks fine.

PR #1907, opened 2021-03-31
Fix stdshift type error

yiheng-wang-nv:

### Description
This PR fixes the type error in StdShiftIntensity transform.

### Status
**Ready**

### Types of changes


Nic-Ma (approved):
Looks good to me.
Could you please help add a test case to cover this new arg `dtype`?
For example, input data is int or float64, expected output is float32?

Thanks.

wyli (approved):
thanks

PR #1909, opened 2021-03-31
Update CuCIM

behxyz:
### Description
CuCIM 0.18.2 fixs upside-downed image for TIFF file if the image is not RGB & tiled image with JPEG/Deflate-compressed tiles.
- Use slow path if the image is not RGB & tiled image with JPEG/Deflate-compressed tiles.
Show an error message if the out-of-boundary cases are requested with the slow path.
- ValueError: Cannot handle the out-of-boundary cases for a non-RGB image or a non-Jpeg/Deflate-compressed image.

I have tested it and it works as expected now.

### Status
**Ready**

### Types of changes

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -4,7 +4,7 @@ pytorch-ignite==0.4.4
 numpy>=1.17
 itk>=5.0
 nibabel
-cucim==0.18.1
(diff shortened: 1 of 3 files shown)


wyli (approved):
thanks!

PR #1912, opened 2021-03-31
[WIP] Adding Stain Normalization Transform for Pathology

nsrivathsa:
Fixes # .

### Description
Added a Transform for CuPy-based Stain Normalization for Pathology.

### Status
Work in progress

### Types of changes


monai/apps/pathology/transforms.py
```diff
@@ -0,0 +1,182 @@
...
+    Performs stain deconvolution of the source image to obtain the stain concentration matrix
+    for the image. Then, performs the inverse Beer-Lambert transform to recreate the
+    patch using the target H&E stain matrix provided. If no target stain provided, a default
+    reference stain is used. Similarly, if no maximum stain concentrations are provided, a
+    reference maximum stain concentrations matrix is used.
+
+    Args:
+        tli: (optional) transmitted light intensity
+        alpha: (optional) tolerance for the pseudo-min and pseudo-max
+        beta: (optional) Optical Density (OD) threshold for transparent pixels
```
behxyz, line 93: The same as lines 14-16.

monai/apps/pathology/transforms.py
```diff
@@ -0,0 +1,182 @@
...
+            img: RGB image to perform stain deconvolution of
+
+        Return:
+            he: H&E OD matrix for the image (first column is H, second column is E, rows are RGB values)
+        """
+        # reshape image
+        img = img.reshape((-1, 3))
+
+        # calculate optical density
+        od = -cp.log((img.astype(cp.float) + 1) / self.tli)
```
behxyz, line 42:
```python
        od = -cp.log((img.astype(cp.float) + 1) / self.tli)
```
Shouldn't the value of `img` clipped by `self.tli`? 
I understand that if `self.tli` is the true transmitted light intensity, there shouldn't be any value beyond that but since it is an input parameter, it would make sense to avoid negative values for absorption (`img` values larger than `self.tli`), which violates the law of conservation of energy. ;)

monai/apps/pathology/transforms.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # remove transparent pixels
+        od_hat = od[~cp.any(od < self.beta, axis=1)]
```
behxyz, line 45:
```suggestion
        absorbance_hat = absorbance[cp.any(absorbance > self.beta, axis=1)]
```
Just changing `<` to `>` to save an operation on all values (`~`).

requirements.txt
```diff
@@ -1,2 +1,3 @@
 torch>=1.5
 numpy>=1.17
+cupy
```
behxyz, line 3: please mention the version.

behxyz (commented):
Well done! It's a great job! 
I left few comments but once you add the unittests, I will review it again. Thanks

monai/apps/pathology/transforms.py
```diff
@@ -0,0 +1,182 @@
+# modified from sources:
+# - Original implementation from Macenko paper in Matlab: https://github.com/mitkovetta/staining-normalization
+# - Implementation in Python: https://github.com/schaugf/HEnorm_python
+# - Link to Macenko et al., 2009 paper: http://wwwx.cs.unc.edu/~mn/sites/default/files/macenko2009.pdf
+import cupy as cp
```
behxyz, line 5: Please make this an optional import, so that if it does not exist, it doesn't affect other components. Like this:
https://github.com/Project-MONAI/MONAI/blob/99bca69f26f9f6a76e60f2ce7848e83470547833/monai/apps/pathology/handlers.py#L22-L25

requirements.txt
behxyz, line 3: Please remove this dependency from here. We will add it to the images with CUDA later.

behxyz (commented):
I left few comments on `cupy`. Please let me know if you have any question on that.

monai/apps/pathology/transforms.py
nsrivathsa: Oh, thank you for pointing this out! Also changing it to `cupy.all()` instead of `cupy.any()` in this case, to maintain the logic :)

behxyz (commented):
Hi @nsrivathsa, thanks for addressing all the comments. So far so good! Have you had a chance to work on the test cases?

behxyz:
closing in favor of #1998

PR #1878, opened 2021-03-28
Tumor FROC Evaluation

behxyz:
Fixes #1872 

### Description
It a reimplementation of #1873 that
- Uses WSIReader to support both OpenSlide and CuCIM
- Updates class, methods, parameters, and variables's names to be more representative
- Restructures the process for more modularity
- Receives the probability map of inference outputs
- Applies non-maximal suppression
- Adds level and pixel_pacing to the input
- Removes redundant calculations
- Rewrites compute_evaluation_mask, now is compute_multi_instance_mask
- Rewrites compute_itc_list, now is compute_isolated_tumor_cells

### Status
**Ready**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick --unittests`.
- [x] In-line docstrings updated.
- [x] Documentation updated, tested `make html` command in the `docs/` folder.

PR #1913, opened 2021-04-01
followup of #1878

wyli:
update tests (`TEST_CASE_5` duplicated), remove json loading (https://github.com/Project-MONAI/MONAI/pull/1878#discussion_r605015813)

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/apps/pathology/metrics.py
+++ b/monai/apps/pathology/metrics.py
@@ -9,8 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import json
-from typing import Dict, List, Tuple, Union
+from typing import Dict, List, Tuple
 
 import numpy as np
 
@@ -51,7 +50,7 @@ class LesionFROC:
 
     def __init__(
         self,
-        data: Union[List[Dict], str],
+        data: List[Dict],
         grow_distance: int = 75,
         itc_diameter: int = 200,
         eval_thresholds: Tuple = (0.25, 0.5, 1, 2, 4, 8),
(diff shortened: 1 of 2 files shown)


behxyz:
Hi @wyli, why are you removing the json loading part? For pathology, since this metric relies on all the ground truth mask and all the probability masks, it should be called independently after all the inferences are done, and they need be defined in a json fie.

wyli:
> Hi @wyli, why are you removing the json loading part? For pathology, since this metric relies on all the ground truth mask and all the probability masks, it should be called independently after all the inferences are done, and they need be defined in a json fie.

yes before the data working group provides a concrete solution we shouldn't couple the lesion evaluation module with json encoding/decoding logic. if it's really needed please write a separate json loader. This class should only work with `Sequence[Dict]`. my previous comment was somehow ignored... (https://github.com/Project-MONAI/MONAI/pull/1878#discussion_r605015813)

wyli:
thanks for the update @behxyz! there's an example of deepgrow with a demo script to show a task-specific json loader https://github.com/Project-MONAI/tutorials/blob/55006d748626f6af59567a295725ccbd3bdd6bef/deepgrow/ignite/train.py#L315

behxyz:
> > Hi @wyli, why are you removing the json loading part? For pathology, since this metric relies on all the ground truth mask and all the probability masks, it should be called independently after all the inferences are done, and they need be defined in a json fie.
> 
> yes before the data working group provides a concrete solution we shouldn't couple the lesion evaluation module with json encoding/decoding logic. if it's really needed please write a separate json loader. This class should only work with `Sequence[Dict]`. my previous comment was somehow ignored... ([#1878 (comment)](https://github.com/Project-MONAI/MONAI/pull/1878#discussion_r605015813))

OK, I'll add a separate json reader till we have an agreed upon solution. Would this be the topic of data WG or pathology WG?

behxyz:
I updated the test case number to be in an order and committed to your PR.

behxyz (approved):

PR #1914, opened 2021-04-01
1904 Implement dice_focal loss

### LINKED ISSUE #1904: Need to add Early-Stop handler
As ignite already has `early-stop` handler, we can develop the MONAI enhanced version based on it:
https://github.com/pytorch/ignite/blob/master/ignite/handlers/early_stopping.py
It's an existing feature in Clara pipeline before.

yiheng-wang-nv:

Fixes #1904 .

### Description
In this PR, dice focal loss has been implemented.
In addition, I improved some places for other features:

- added weight parameters `lambda_dice` and `lambda_ce` in `DiceCELoss` (include unittest)
- added values check for the parameter `weight` in `FocalLoss` (include unittest)
- fixed a string issue in `SENet` (the long string with line breaks will introduce extra spaces when printing).

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/losses/focal_loss.py
+++ b/monai/losses/focal_loss.py
@@ -45,7 +45,9 @@ class FocalLoss(_Loss):
             weight: weights to apply to the voxels of each class. If None no weights are applied.
                 This corresponds to the weights `\alpha` in [1].
                 The input can be a single value (same weight for all classes), a sequence of values (the length
-                of the sequence should be the same as the number of classes).
+                of the sequence should be the same as the number of classes, if not ``include_background``, the
+                number should not include class 0).
+                The value/values should be no less than 0. Defaults to None.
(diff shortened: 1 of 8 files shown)


wyli (approved):
thanks!

monai/losses/focal_loss.py
```diff
@@ -122,6 +127,13 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
                 class_weight = torch.as_tensor([self.weight] * i.size(1))
             else:
                 class_weight = torch.as_tensor(self.weight)
+                if class_weight.size(0) != i.size(1):
+                    raise ValueError(
+                        "the length of the weight sequence should be the same as the number of classes. "
+                        + "If `include_background=False`, the number should not include class 0."
```
wyli, line 133: I think you can omit the `+ `

PR #1948, opened 2021-04-05
1947 Enhance load decathlon datalist API

### LINKED ISSUE #1947: Prepend path to "mask"
`decathlon_dataloader.py` only prepend `base_dir` path to "image" and if it's segmentation to "label". However, if we have a "mask" entry it is not prepended by `base_dir`.
https://github.com/Project-MONAI/MONAI/blob/master/monai/data/decathlon_datalist.py#L65-L68

```
            if k == "image" or k == "mask":
                item[k] = _compute_path(base_dir, v)
            elif is_segmentation and k == "label":
                item[k] = _compute_path(base_dir, v)
```

Nic-Ma:
Fixes #1947 .

### Description
This PR added support to join base dir for additional keys.

### Status
**Ready**

### Types of changes


monai/data/decathlon_datalist.py
```diff
@@ -17,34 +17,42 @@
...
         base_dir: the base directory of the dataset.
         element: file path(s) to append to directory.
+        check_path: if `True`, only compute when the result is an existing path.
 
     Raises:
         TypeError: When ``element`` contains a non ``str``.
         TypeError: When ``element`` type is not in ``Union[list, str]``.
 
     """
+    def _join_path(base_dir: str, item: str):
```
behxyz, line 42:
```suggestion

    def _join_path(base_dir: str, item: str):
```

behxyz (approved):

monai/data/decathlon_datalist.py
```diff
@@ -63,9 +72,12 @@ def _append_paths(base_dir: str, is_segmentation: bool, items: List[Dict]) -> Li
             raise TypeError(f"Every item in items must be a dict but got {type(item).__name__}.")
         for k, v in item.items():
             if k == "image":
-                item[k] = _compute_path(base_dir, v)
+                item[k] = _compute_path(base_dir, v, check_path=False)
             elif is_segmentation and k == "label":
```
behxyz, line 76: I believe it is better to remove `is_segmentation` because it might be confusing since "label" are processing either here or in the next if (depending on this flag) and by removing. Also, what if we are doing segmentation but using patch-wise classification, then the labels are still values and not masks.

behxyz (changes requested):

PR #1954, opened 2021-04-06
allow slicing with SpatialCrop(d)

rijobro:
Allow the use of a list of `slice` when cropping. This allows indexing from the back or `None` to not crop in that direction of that dimension. As requested here: https://github.com/Project-MONAI/MONAI/discussions/1929 and https://github.com/Project-MONAI/MONAI/discussions/1935.

I decided to add `slices` rather than modifying `roi_start` to accept -ve numbers, for example, as this would break backwards compatibility (and doesn't allow for `None`).

### Description
Transform can now be called with e.g., `SpatialCrop(roi_slices=[slice(-10, None), slice(None)])` to take the last 10 elements in the first dimension, and leave the second dimension untouched.

Implement the inverse using the slices rather than `roi_start` and `roi_end`, as was done previously.

PR is backwards compatible.

### Status
**Ready**

### Types of changes


wyli:
could this be used to support https://github.com/Project-MONAI/MONAI/issues/1993?

rijobro:
@wyli do you know if the line I deleted in the most [recent commit](https://github.com/Project-MONAI/MONAI/pull/1954/commits/c4a0072f318a964761b9bc7e143d595eeb5c7d95) was required? `box_start` and `box_end` are both defined a few lines earlier.

wyli:
> @wyli do you know if the line I deleted in the most [recent commit](https://github.com/Project-MONAI/MONAI/pull/1954/commits/c4a0072f318a964761b9bc7e143d595eeb5c7d95) was required? `box_start` and `box_end` are both defined a few lines earlier.

I think those are needed because `SpatialCrop` will refine those initial values, perhaps we can still keep `self.roi_start` and `self.roi_end` in `SpatialCrop`?

wyli:
similarly the current inverse logic uses the original box_start/end, we might have some issues...
https://github.com/Project-MONAI/MONAI/blob/5f47407416c8391d9267efb4d7fe5275697374aa/monai/transforms/croppad/dictionary.py#L602-L604

rijobro:
The inverse of `CropForegroundd` is fine since the limits get calculated by `generate_spatial_bounding_box`. I've updated the deepgrow transform to get the start and end bounding box from the `.start` and `.stop` of slice. 

If possible, I'd rather not keep `self.roi_start` and `self.roi_end`. It would make life easier in the short-run, but given that the bounding box is a less general form than `slice`, keeping both would require a lot if `if`/`else` statements.

wyli (approved):
thanks!

PR #1967, opened 2021-04-07
[WIP] Gradient clipping

danieltudosiu:
### Description
Added a basic gradient clipping draft for the SupervisedTrainer. The implementation choice must be discussed.

### Status
Work in progress

### Types of changes

### DIFF
--- a/monai/engines/trainer.py
+++ b/monai/engines/trainer.py
@@ -106,6 +106,9 @@ class SupervisedTrainer(Trainer):
         amp: bool = False,
         event_names: Optional[List[Union[str, EventEnum]]] = None,
         event_to_attr: Optional[dict] = None,
+        gradient_clip_val: Optional[float] = None,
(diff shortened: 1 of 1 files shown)


danieltudosiu:
@wyli / @Nic-Ma I need your opinion about how to attack this one. 

In my mind, each Trainer class must have its own gradient clipping function since the number of optimizers and scalers can vary.

What do you think of the logic draft? Should I write it for the GAN one as well (I observed the GAN one is not on par with the SupervisedTrainer)?

I think the following solutions do not work:
- backward_hook
    - "inf" norm will not be usable since it needs all model's parameters
- Ignite Handler
    - It will need to be created after the creation of the engine to be able to access the scaler and amp attributes of the trainer.

wyli:
closing this, as preferred solutions are described in https://github.com/Project-MONAI/MONAI/discussions/3892#discussioncomment-2325221

PR #1970, opened 2021-04-08
1866 Add TransformInverter handler

### LINKED ISSUE #1866: Add ignite handler to inverse transforms
I would suggest to develop a ignite handler to inverse spatial transforms, @rijobro what do you think?

follow-up of https://github.com/Project-MONAI/MONAI/issues/1515

Nic-Ma:
Fixes #1866 .

### Description
This PR added the ignite handler to invert transforms for model output data.

### Status
**Ready**

### Types of changes


rijobro:
Thanks for looking into this @Nic-Ma. I notice an issue here that I hadn't thought about previously. 

During training/validation/testing, there will typically be a ground truth label attached, and we'll want to use those transforms for the inverse.

At deployment, we'll probably only be using images (no labels). And typically for labels, we'll use nearest-neighbour interpolation, whereas for images we'll use something else (e.g., linear). So if we do the inverse as you've suggested in this PR, we'll probably be using non-nearest neighbour interpolation on our labels.

I wonder if this would be solved by simply doing `AsDiscrete` at the end. A little messy. cc @wyli @ericspod.

Nic-Ma:
Hi @rijobro ,

Thanks for your review, there are another 2 minor issues about inverse logic:
1. You didn't implement the `inverse` for `ToTensorD` transform, so it will run with `Tensor` data, fixed in this PR.
2. The `BatchInverseTransform` requires the transform must be `InvertibleTransform` while the `allow_missing_keys_mode` requires `Compose` or `MapTransform`, it will cause some typehints issue, I ignored in the PR.

About the sampling question you proposed, it's a good point, let me add another arg `as_discrete`.

Thanks.

rijobro (approved):
test is basic, but I suppose testing shape is sufficient here, as voxel values are tested in test_inverse.py. Thanks!

Have you tested this personally to check it works as you expect?

Nic-Ma:
Hi @rijobro ,

Thanks for your review, I already tested it with the workflow example:
https://github.com/Project-MONAI/tutorials/blob/master/modules/engines/unet_evaluation_dict.py

And I will write another PR to solve the `interpolation mode` issue soon.

Thanks.

wyli:
thanks @rijobro chat with Nic offline that we plan to update to store/use the inverse parameters such as `mode` and `align_corners` in  `EXTRA_INFO`

PR #1978, opened 2021-04-08
1977 Support different inverse interpolation mode

### LINKED ISSUE #1977: Need to support different interpolation mode in inverse transforms
During inverse transform, we may need to use different interpolation mode.

Nic-Ma:
Fixes #1977 

### Description
This PR added support to use different interpolation mode in inverse transforms.

### Status
**Ready**

### Types of changes


Nic-Ma:
This PR depends on: https://github.com/Project-MONAI/MONAI/pull/1970

Thanks.

rijobro:
Could the same be achieved with a withable (which wouldn't adding `mode` to `EXTRA_INFO`)?

```python
with nearest_neighbour_transforms(*transforms):
    # if Compose, get list (same logic as in allow_missing_keys_mode)
    all_transforms = ...

    # get sublist of all transforms that contain `mode`
    mode_transforms = [i for i in all_transforms is hasattr(i, mode)]

    # record previous values
    orig_modes = [i.mode for i in mode_transforms]

    try:
        # Set all to nearest_neighbour
        for i in mode_transforms:
            i.mode = 'nearest-neighbour'
        yield
    finally:
        # Revert
        for t, o_s in zip(mode_transforms, orig_modes):
            t.allow_missing_keys = o_s
```

Nic-Ma:
Hi @rijobro ,

Thanks for your suggestion.
Actually, @wyli and I discussed this approach yesterday, the benefit of my current solution is to save `mode`, `align_corners`, etc. into the dict which only **affects current input data**. It's dangerous to modify `self.mode` property in the transform instances and especially thread-safe related issues. We may support `inverting` several data **in parallel** later, some of them may expect `nearest` mode but some may not.
And we can also use this saved `mode` information for some future features.
@wyli what do you think?

Thanks.

rijobro:
Ok that sounds like good motivation for doing it your way. You've put the logic of modifying `EXTRA_INFO` inside of the ignite handler, but it could be put somewhere else for more general use. For example, we could allow the inverse method to accept a dictionary that could replace items in `EXTRA_INFO`. I think that would have the same effect as your current code but be accessible for users not using ignite:

```python
def inverse(self, data, **extra_info):
    d = deepcopy(dict(data))
        # replace any elements in d["EXTRA_INFO"] with **extra_info
        ...
```

and then we'd call it with `transforms.inverse(data, {"mode": "nearest"}`

What do you think?

Nic-Ma:
Hi @rijobro ,

Thanks for your suggestion, do you mean to add the `replacement` logic for every `ItertibleTransform`?

Nic-Ma:
Hi @rijobro ,

I think we can make the logic into a util function, then users can use it without ignite, what do you think?

Thanks.

wyli:
> Ok that sounds like good motivation for doing it your way. You've put the logic of modifying `EXTRA_INFO` inside of the ignite handler, but it could be put somewhere else for more general use. For example, we could allow the inverse method to accept a dictionary that could replace items in `EXTRA_INFO`. I think that would have the same effect as your current code but be accessible for users not using ignite:
> 
> ```python
> def inverse(self, data, **extra_info):
>     d = deepcopy(dict(data))
>     for key in self.key_iterator(d):
>         # replace any elements in d["EXTRA_INFO"] with **extra_info
>         ...
> ```
> 
> and then we'd call it with `transforms.inverse(data, {"mode": "nearest"})`
> 
> What do you think?

I feel this is necessary for general usability, but we should do it in a separate PR as a new feature implementation? perhaps we don't have time for this new feature in v0.5...

Nic-Ma:
Hi @rijobro and @wyli ,

I already changed the replacement logic to util function and added unit tests.
Now users can use it without the `ignite` workflows.

Thanks in advance.

wyli (approved):

rijobro:
@Nic-Ma looks great, thanks!

PR #1992, opened 2021-04-12
TTA progress bar, use torch for mode, add label meta_data

rijobro:
### Description
A few updates to TTA:

- Progress bar
- Use torch for computing mode (quicker than numpy)
- Include the `meta_dict` as this is required for the inverse of `Orientationd`, `Spacingd`, etc.

### Status
**Ready**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [x] Quick tests passed locally by running `./runtests.sh --quick --unittests`.
- [x] In-line docstrings updated.
- [x] Documentation updated, tested `make html` command in the `docs/` folder.

PR #1997, opened 2021-04-12
include meta when inverting

wyli:

followup of #1992

### Status
**Ready**

### Types of changes

### DIFF
--- a/tests/test_handler_transform_inverter.py
+++ b/tests/test_handler_transform_inverter.py
@@ -32,6 +32,7 @@ from monai.transforms import (
     RandZoomd,
     ResizeWithPadOrCropd,
     ScaleIntensityd,
+    Spacingd,
     ToTensord,
 )
 from monai.utils.misc import set_determinism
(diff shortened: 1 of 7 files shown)


tests/test_handler_transform_inverter.py
```diff
@@ -48,6 +49,7 @@ def test_invert(self):
             [
                 LoadImaged(KEYS),
                 AddChanneld(KEYS),
+                Spacingd(KEYS, pixdim=(1.1, 1.01, 0.9), mode=["bilinear", "nearest"], dtype=np.float32),
```
wyli, line 52: quite annoying we have the shape quantization error as I mentioned before. 
the current test case works fine, but if I change the case to:
```suggestion
                Spacingd(KEYS, pixdim=(1.2, 1.01, 0.9), mode=["bilinear", "nearest"], dtype=np.float32),
```
output:
```
F
FAIL: test_invert (__main__.TestTransformInverter)
  File "/Documents/MONAI/tests/test_handler_transform_inverter.py", line 99, in test_invert
    self.assertTupleEqual(i.shape, (1, 100, 101, 107))
AssertionError: Tuples differ: torch.Size([1, 101, 101, 107]) != (1, 100, 101, 107)

First differing element 1:
101
100

- torch.Size([1, 101, 101, 107])
+ (1, 100, 101, 107)

Ran 1 test in 7.454s

```
I'll try to address it...

cc @rijobro @Nic-Ma

Nic-Ma (approved):

PR #1998, opened 2021-04-13
[WIP] Adding Stain Normalization Transform for Pathology

nsrivathsa:
### Description
Adds a Transform for CuPy-based Stain Normalization for Pathology.

### Status
**Work in progress**

### Types of changes


behxyz (commented):
Well done, @nsrivathsa! I left few comments inline. 
In general, can you think of edge cases in physical world for test cases as well, like the situation that we expect to have only one stain so the concentration of the other one is zero.
Also it would be great to have input images that are not uniformly filled. 
Keep in mind that writing solid unittests are very important to ensure the correctness of our code and it might take more time than implementing the original functions and classes.

tests/test_pathology_transforms.py
```diff
@@ -0,0 +1,202 @@
...
+from parameterized import parameterized
+
+from monai.apps.pathology.transforms import ExtractStainsMacenko
+from monai.utils import exact_version, optional_import
+
+cp, has_cp = optional_import("cupy", "8.6.0", exact_version)
+
+# input pixels all transparent and below the beta absorbance threshold
+EXTRACT_STAINS_TEST_CASE_1 = [
+    cp.full((3, 2, 3), 240),
```
behxyz, line 23: Calling cupy outside of the testing class scope is troublesome when it is not available. You should put all of them into a function and call it from within the test class. You can find an example here:
https://github.com/Project-MONAI/MONAI/blob/e0035b108899e70084b988b6bb78d707313f493d/tests/test_lesion_froc.py#L305-L306

behxyz:
@nsrivathsa I see that [lines 47-75](https://github.com/Project-MONAI/MONAI/pull/1998/files#diff-f260e09d7ecae12bededcace9c6083c97daa65c33c87d74ecfa5696ea91fbd8bR47-R75) for stain extraction are being repeated in line 143-171 for stain normalization. Can you modify the code so that it reuses stain extraction in stain normalization?

drbeh:
A variant of this PR is merged.

PR #1999, opened 2021-04-13
Update transform inverter for postfix

Nic-Ma:
### Description
This PR updated the `TransformInverter` handler to align the postfix format with other components.

### Status
**Ready**

### Types of changes


monai/handlers/transform_inverter.py
```diff
@@ -104,7 +104,11 @@ def __call__(self, engine: Engine) -> None:
 
             transform_info = engine.state.batch[transform_key]
             if nearest_interp:
-                convert_inverse_interp_mode(trans_info=transform_info, mode="nearest", align_corners=None)
+                transform_info = convert_inverse_interp_mode(
+                    trans_info=deepcopy(transform_info),
```
Nic-Ma, line 108: Maybe several output_keys share the same batch_key during inverting, for example, all items use the transform_info of `label`, should deepcopy if we modify it.

wyli (approved):
thanks

monai/handlers/transform_inverter.py
wyli: thanks, we agreed that we should create some followup test cases

PR #2060, opened 2021-04-21
2044 Add check for label data type

### LINKED ISSUE #2044: ignore `label` in the workflow if it's not tensor
During evaluation, now we use `prepare_batch` to extract `image` and `label` fields if existing the key. But maybe we don't want to load `label` for inference and didn't apply transforms, so need to ignore it for this case.

Nic-Ma:
Fixes #2044 .

### Description
Currently, we always move `label` data to device, but it can't work if the label is not a `Tensor`, this PR added type check.

### Status
**Ready**

### Types of changes


wyli (approved):

Nic-Ma:
The `test_integration_workflow` test passed locally.
Thanks.

PR #2065, opened 2021-04-21
2018 Save inverted meta dict into state.batch

### LINKED ISSUE #2018: Missing inverse for `LoadImageD` transform
**Describe the bug**
As the `LoadImageD` transform may convert input data and `affine` here:
https://github.com/Project-MONAI/MONAI/blob/master/monai/data/image_reader.py#L389
Need to support inverse operations.
Otherwise, we can't get proper results to compare with the input image if set `as_closest_canonical=True`.

Nic-Ma:
Fixes #2018 .

### Description
As we can't support `inverse` for `LoadImageD` transform, this PR changed to save the inverted meta dict into `engine.state.batch`, then we can use this meta data to resample before saving.

### Status
**Ready**

### Types of changes


wyli (approved):
thanks!

Issue #2087, opened 2021-04-26
DeepGrow transform broken as `SpatialCrop` changed properties


Nic-Ma:
**Describe the bug**
The CropTransform of DeepGrow app depends on the `SpatialCrop` transform, but recent PR changed it, so this line can't work anymore:
https://github.com/Project-MONAI/MONAI/blob/master/monai/apps/deepgrow/transforms.py#L481

Nic-Ma:
Hi @rijobro ,

Your previous PR: https://github.com/Project-MONAI/MONAI/pull/1954 changed the properties of `SpatialCrop` transform, which is used in the DeepGrow module.
Could you please help take a look and update `SpatialCrop` or update `DeepGrow` transform?
And it's better to add 1 more unit test to cover this case in the DeepGrow tests. @SachidanandAlle 

Thanks in advance.

PR #2088, opened 2021-04-26
update box_start/end for deepgrow

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/2087.

### Description
Update method for getting start and stop of bounding box from `slices`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/apps/deepgrow/transforms.py
+++ b/monai/apps/deepgrow/transforms.py
@@ -478,8 +478,8 @@ class SpatialCropForegroundd(MapTransform):
 
         if np.all(np.less(current_size, self.spatial_size)):
             cropper = SpatialCrop(roi_center=center, roi_size=self.spatial_size)
-            box_start = cropper.roi_start
-            box_end = cropper.roi_end
+            box_start = np.array([s.start for s in cropper.slices])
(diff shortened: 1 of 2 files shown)


rijobro:
@Nic-Ma hopefully this addresses that issue, but I guess the code is untested as the deepgrow unit tests passed prior to these changes. Have you got example code that would previously cause an error? Hopefully we can turn a simplified version of that into a new unit test.

Nic-Ma:
Hi @rijobro ,

Thanks very much for your quick fix!
This case only run when `if np.all(np.less(current_size, self.spatial_size))`, and the unit test of this transform doesn't cover it:
https://github.com/Project-MONAI/MONAI/blob/master/tests/test_deepgrow_transforms.py#L357
Could you please help add 1 case to test your PR?

Thanks in advance.

rijobro:
The added test passes, but I'm not really sure if it's correct or not. Could you check?

Nic-Ma (approved):
Looks good to me.
@SachidanandAlle , could you please also help review it?

Thanks.

SachidanandAlle (approved):
LGTM

PR #2106, opened 2021-04-28
only pip install cucim if on linux

rijobro:
### Description

Only pip install cucim if on linux. The requirements-dev.txt, as it currently stands, causes an error on non-linux machines, meaning that following requirements don't get installed.

### Status
**Ready**

### Types of changes


wyli (approved):
thanks! didn't know this syntax

rijobro:
For interest: https://www.python.org/dev/peps/pep-0508/#environment-markers

Issue #2041, opened 2021-04-19
Branches restructuring


wyli:
### Is your feature request related to a problem? Please describe.
Currently, the MONAI repo has
- `master` branch: accepting pull requests, new features and/or bug fixes -- full CI/CD tests, nightly tests.
- `releases/*` branch: branching off from the master, no direct commits allowed -- releasing/packaging related tests
- contributors' pull request `forks`: code change proposals -- quick pre-merge tests.

This approach is simple and efficient. But it has the problem that there is no place for the new features that are scheduled beyond the current release (they only live in some of the contributors' forks).

### Describe the solution you'd like

As we discussed in the dev meeting, we'd like to follow a more rigorous releasing approach.

- `main` branch: for production-ready features, milestone releases -- releasing/packaging related tests
- `dev` branch: for all new features -- full CI/CD, nightly tests
- `releasing/*` branch: a subset of new features from `dev`, to be released, could have additional bug fixes for releasing
- `forks` of the contributors: new feature proposals -- quick pre-merge tests

So, the new workflow would be:
The contributors add new features to `dev` via pull request.
The maintainers plan and build a `releasing/*` branch (from `dev`). 
The maintainers merge the production-ready `releasing/*` to `main` and tag `main` for a new milestone.
The users download and use the package built from the new release upon `main`.

Tasks:
-  [x] create a `dev` branch with proper CI/CD
-  [x] double-check the offline backup of the repo
-  [x] conclude existing PRs
-  [x] create a `main` branch for milestones
-  [x] update contributors guide
-  [x] revise the CI/CD workflows
-  [x] deprecate the `master` branch
-  [x] deprecate the `releases/*` they are temporary branches, tagged commits are available at `main`

cc @ericspod 

update:
![Screenshot 2021-04-28 at 19 12 33](https://user-images.githubusercontent.com/831580/116452595-cc683500-a855-11eb-8d59-8567cb58f83b.png)

vfdev-5:
@wyli interesting suggested approach! Curious to see how it works in practice and if it really helps to solve the above issue without headaches of what to merge etc.

In the first approach: `master - releases/*`, you also can merge all features to the master and prepare releases as a new branch `vX.Y.Z` starting from somewhere on the master or previous release branch and cherry-picking from master or reverting existing commits. I'd say the idea is almost the same, but branches names are inverted a bit and there is additional `main` in the second approach...

wyli:
[...]

PR #2107, opened 2021-04-28
2041 - re-enable dev branch tests

wyli:

part of #2041

### Description
enable all dev branch related tests

### Status
**Ready**

### Types of changes

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,10 +4,10 @@ name: docker
 # docker_test_latest: test the latest internal docker image (has flake)
 # docker_test_dockerhub: test the latest dockerhub release (no flake)
 on:
-  # master only docker deployment and quick tests
+  # dev only docker deployment and quick tests
   push:
     branches:
-      - master
+      - dev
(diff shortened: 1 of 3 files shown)


ericspod:
Shouldn't we be running at least some of the CI/CD tests for both `dev` and `main` branches? In theory nothing should be committed to main that wasn't tested but it wouldn't hurt maybe to double check that. Also anywhere that MONAI is installed directly from the repo needs to refer to `dev` branch I think otherwise the main branch will be tested instead.

wyli:
> Shouldn't we be running at least some of the CI/CD tests for both `dev` and `main` branches? In theory nothing should be committed to main that wasn't tested but it wouldn't hurt maybe to double check that. Also anywhere that MONAI is installed directly from the repo needs to refer to `dev` branch I think otherwise the main branch will be tested instead.

yes, this is the part for`dev` branch only, this will be used for our daily pull request handling, would be great to get this running first. I'm working on the details of the `main` branch and the releasing related ones, they will be in another few PRs.

ericspod (approved):

PR #2119, opened 2021-04-30
create meta data key if necessary

rijobro:
Not all images will have been read from file, meaning `img_meta_data` won't necessarily be present. The user may still want to change voxel spacing, in which case the original affine is assumed to be identity. However, previously an error would be thrown if the meta data wasn't present. Here, we create it if necessary.

### Status
**Ready**

### Types of changes


wyli (approved):
thanks!

Issue #2116, opened 2021-04-30
Collate error on the key 'img_transforms' of dictionary data.


ambroslins:
**Describe the bug**
While using the `CacheDataset` and multiple workers and the `persistent_workers` option for the `DataLoader` following exception is raised in the second epoch:

```
  File "C:\Users\Ambros\Dev\monai-issue\env\lib\site-packages\monai\data\utils.py", line 277, in list_data_collate
    raise RuntimeError(re_str)
RuntimeError: each element in list of batch should be of equal size
Collate error on the key 'img_transforms' of dictionary data.

MONAI hint: if your transforms intentionally create images of different shapes, creating your `DataLoader` with `collate_fn=pad_list_data_collate` might solve this problem (check its documentation).

````
[see full output](https://gist.github.com/ambroslins/9b079eac659c3483c6fceba5fcee8859#file-output-txt)

**To Reproduce**
I wrote a [minimal example](https://gist.github.com/ambroslins/9b079eac659c3483c6fceba5fcee8859) based on [unet_training_dict.py](https://github.com/Project-MONAI/tutorials/blob/master/3d_segmentation/torch/unet_training_dict.py) from the tutorials repository.

1. Go to  [https://gist.github.com/ambroslins/9b079eac659c3483c6fceba5fcee8859#file-unet_training_dict-py](https://gist.github.com/ambroslins/9b079eac659c3483c6fceba5fcee8859#file-unet_training_dict-py) and download `unet_training_dict.py`
2. Install [environment](https://gist.github.com/ambroslins/9b079eac659c3483c6fceba5fcee8859#file-environment-txt)
3. Run command: `python unet_training_dict.py`

**Expected behavior**
I don't expect this expection to be raised, the same code worked fine on monai version 0.4.0.

**Environment**
[pip freeze](https://gist.github.com/ambroslins/9b079eac659c3483c6fceba5fcee8859#file-environment-txt)
```
> python -c 'import monai; monai.config.print_debug_info()'
================================
Printing MONAI config...
================================
MONAI version: 0.5.2
Numpy version: 1.20.2
Pytorch version: 1.8.1+cu111
MONAI flags: HAS_EXT = False, USE_COMPILED = False
MONAI rev id: feb3a334b7bbf302b13a6da80e0b022a4cf75a4e

Optional dependencies:
Pytorch Ignite version: NOT INSTALLED or UNKNOWN VERSION.
Nibabel version: 3.2.1
[...]

PR #2121, opened 2021-04-30
fix cachedataset w persistent workers problem by deep copying

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/2116.

### Description
When using `CacheDataset` combined with `persistent_workers=True` in the `DataLoader`, it seems that the `<key>_transforms` are duplicated for the first random transform onwards.

So for a transform of `Compose([Spacingd, RandAffined, ToTensord])`, then on the second pass through the data, we would have `img_transforms` as: `Spacing, RandAffined, ToTensord, RandAffined, ToTensord`. This implies a shallow copy at some point. 

I've put the copy in the `CacheDataset` but I worry that this may cause unnecessary copying of the data. Having said that, I can't see where else to put it... I'm open to ideas. 

### Status
**Ready**

### Types of changes


tests/test_cachedataset_persistent_workers.py
```diff
@@ -0,0 +1,50 @@
...
+    DataLoader,
+)
+from monai.transforms import (
+    Compose,
+    RandAffined,
+    Spacingd
+)
+
+
+class TestTransformsWCacheDatasetAndPersistentWorkers(unittest.TestCase):
```
Nic-Ma, line 26: Please add this test to the min_tests list to skip.

Nic-Ma (commented):
Looks good to me.
Could you please help run the `CacheDataset` tutorial twice with / without `deepcopy` to make sure it will not affect the training speed and memory usage?
https://github.com/Project-MONAI/tutorials/blob/master/acceleration/dataset_type_performance.ipynb

Thanks.

tests/test_cachedataset_persistent_workers.py
Nic-Ma: Hi @rijobro ,

I think this PR is important to enhance the `CacheDataset`, can you please help run some tests for speed and memory,  then let's merge it?

Thanks.

rijobro: Sure I'll have a look now.

rijobro: Confirmed as same execution time and same results on the spleen_segmentation_3d notebook.

Nic-Ma:
Hi @rijobro ,

Do you have update on this PR?

Thanks.

rijobro:
Sorry, put the comment in a resolved conversation. All looks good to me, I'll just have a look at the github action error.

rijobro:
@wyli @Nic-Ma I don't understand the output of the flake8 check. I re-ran the job just in case, but still don't know what's wrong. Any ideas?

wyli:
looks like an issue with the latest pytype https://github.com/google/pytype/issues/ 909

monai/data/dataset.py
```diff
@@ -580,7 +580,7 @@ def _transform(self, index: int):
         for _transform in self.transform.transforms:
             if start_run or isinstance(_transform, Randomizable) or not isinstance(_transform, Transform):
                 start_run = True
-                data = apply_transform(_transform, data)
+                data = apply_transform(_transform, deepcopy(data))
```
wyli, line 583: I feel the deepcopy should happen after line 577？

Nic-Ma: I would suggest below code to only deepcopy once:
```py
for _transform in self.transform.transforms:
    if start_run or isinstance(_transform, Randomizable) or not isinstance(_transform, Transform):
        if not start_run:
            start_run = True
            data = deepcopy(data)
        data = apply_transform(_transform, data)
```
@rijobro @wyli , what do you think?

Thanks.

rijobro: @Nic-Ma good suggestion, I did something similar using `enumerate` so we can delete `start_run`. I checked that the unit test passes with the modified code, and fails when I comment it out, so the deep copy is still working as intended.

Nic-Ma: Hi @rijobro , the `start_run` var is used to mark the **first** random transform, so we only need to deepcopy once, then run all the random transforms based on the copied data.

Thanks.

rijobro: ah I understand, sorry about that. reverted back to using `start_run`.

Nic-Ma (approved):

Nic-Ma:
BTW, I tested this PR in our `CacheDataset` tutorial:
https://github.com/Project-MONAI/tutorials/blob/master/acceleration/dataset_type_performance.ipynb

1. There is no **speed** or **memory usage** difference with / without `deepcopy`.
2. With "persistent_workers=True", the training speed of every epoch is about 1 second faster.

Thanks.

Issue #2124, opened 2021-04-30
RandAffine/RandAffined skipping the resampling if do_transform is False


wyli:
**Is your feature request related to a problem? Please describe.**
when do_transform is False and the spatial_size isn't set, the resampling step of RandAffine could be skipped for better performance, for example:
https://github.com/Project-MONAI/MONAI/blob/fd9081bb369827f56853d184139408630ba37ea2/monai/transforms/spatial/array.py#L1441-L1444

Nic-Ma:
Thanks for your suggestion, let me try to optimize it later.

Nic-Ma:
Welcome a contribution if anyone is interested.

Thanks.

PR #2151, opened 2021-05-06
skip RandAffine if possible

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/2124.

### Description
Skip RandAffine if possible.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -1437,6 +1437,10 @@ class RandAffine(RandomizableTransform):
                 See also: https://pytorch.org/docs/stable/nn.functional.html#grid-sample
         """
         self.randomize()
+        # if not doing transform and spatial size not defined, nothing to do
+        if not self._do_transform and not spatial_size and not self.spatial_size:
+            return torch.Tensor(img) if self.resampler.as_tensor_output else np.array(img)
+
         sp_size = fall_back_tuple(spatial_size or self.spatial_size, img.shape[1:])
         if self._do_transform:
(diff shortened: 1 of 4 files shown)


rijobro:
When debugging, I put a print statement to verify that the first unit test skips the resampling, as expected.

wyli:
thanks, need something like
```py
        if not self.rand_affine.do_transform and self.rand_affine.spatial_size is None:
            for idx, key in enumerate(self.keys):
                if self.rand_affine.resampler.as_tensor_output:
                    d[key] = torch.tensor(key)
                elif torch.is_tensor(d[key]):                    
                    d[key] = d[key].cpu().numpy()
            return d
```
for `RandAffined`

need to double-check the output data type, make sure that the PR doesn't change the dtype for integer `image`

rijobro:
also need to think about the inverse for `RandAffined`...

rijobro:
@wyli why would the dtype change if the image isn't modified? Because of the torch/numpy conversion?

wyli:
the root cause is that the current resampling requires floats...https://github.com/Project-MONAI/MONAI/blob/41e62e544685e87a14b88d7fdca229dc4191836d/monai/transforms/spatial/array.py#L1240-L1242

if we bypass the resample step, I feel the datatype might be different, but I haven't checked all the details...

wyli (approved):
thanks!

wyli:
the dev branch had its 1024th merged pull request 🎉

rijobro:
that was quick, it's only a few weeks old! 😅

PR #2153, opened 2021-05-06
Pytype fix

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/pull/2121#issuecomment-833551183.

### Description
Avoid most recent pytype (2021.05.04) due to https://github.com/google/pytype/issues/909.

### Status
**Ready**

### Types of changes

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -21,7 +21,8 @@ pycodestyle
 pyflakes
 black
 isort
-pytype>=2020.6.1, <=2020.12.23
+attrs<21.1.0
+pytype>=2020.6.1
 mypy>=0.790
 ninja
 torchvision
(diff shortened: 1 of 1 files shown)


wyli:
according to the pytype ticket, looks like it's actually the attrs issue...  https://github.com/python-attrs/attrs/releases

rijobro:
Ok. So we add in `attrs<21.1.0` before pytype and then remove it once their issue is closed?

wyli:
> Also looks like the `attrs` bug has already [been fixed](https://github.com/python-attrs/attrs/issues/752), so the problem will solve itself when they do their next release.

shall we close this and wait?

wyli:
sounds good

PR #2162, opened 2021-05-08
2097 CoordConv transform

### LINKED ISSUE #2097: CoordConv transform
CoordConv gives networks an idea of where they are in an input by adding additional channels to the input that simply contain the spatial coordinate system. It has been shown to improve performance in various computer vision tasks, [paper here](https://arxiv.org/abs/1807.03247).

This could be simply implemented as a transform. If you think this would be useful I can make a PR to add.

marksgraham:
Fixes #2097  .

### Description
Implements a transformation that appends spatial coordinates to the input as new channels, inspired by the  [CoordConv paper](https://arxiv.org/abs/1807.03247). Have found this improves the performance of my segmentation networks.

### Status
**Ready**

### Types of changes

- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`. **I get a pytype error that I can't resolve - help appreciated**


wyli:
thanks, for the typing error, a simple workaround would be removing ` -> Union[np.ndarray, torch.Tensor]` in the newly added classes

monai/transforms/spatial/array.py
```diff
@@ -1700,3 +1701,50 @@ def __call__(
...
+            )
+        if 0 in self.spatial_channels:
+            raise ValueError("cannot add CoordConv channel for dimension 0, as 0 is channel dim.")
+
+        spatial_dims = img.shape[1:]
+        # pre-allocate memory
+        coord_channels = np.ones((len(self.spatial_channels), *spatial_dims)).astype(img.dtype)
+
+        for i, dim in enumerate(self.spatial_channels):
+            ones = np.ones((1, *spatial_dims))
```
rijobro, line 1742: Could the calculation of the coordinates be done as:

```python
shape = img.shape[1:]
num_channels = img.shape[0]
coords = sum(np.ogrid[tuple(slice(0, i) for i in shape)])
coords = np.repeat(coords[None], repeat=num_channels, axis=0)
```

rijobro: You could remove the `for s in spatial_channels` by requiring use input to be 0-based on the spatial dimensions (e.g., `(0,1,2)` for `(x,y,z)`).

monai/transforms/spatial/array.py
```diff
@@ -1700,3 +1701,51 @@ def __call__(
...
+            )
+        if 0 in self.spatial_channels:
+            raise ValueError("cannot add AddCoordinateChannels channel for dimension 0, as 0 is channel dim.")
+
+        spatial_dims = img.shape[1:]
+        num_spatial_dims = len(spatial_dims)
+        coord_channels = np.mgrid[tuple(slice(0, i) for i in spatial_dims)]
+        # normalise so coordinates lie between [-0.5,0.5]
+        normalisation = np.expand_dims(np.array(spatial_dims) - 1, axis=tuple(np.arange(1, num_spatial_dims + 1)))
+        coord_channels = coord_channels / normalisation - 0.5
```
rijobro, line 1746: if you want results between -0.5 and 0.5, you could use something meshgrid?

```suggestion
        coord_channels = np.array(np.meshgrid(*tuple(np.linspace(-0.5, 0.5, s) for s in spatial_dims)))
```

marksgraham:
Thanks for comments. I have:
- Fixed the typing error (I think!)
- Used @rijobro's cleaner code for generating the channels. I've decided to stick with the `spatial_channels` argument starting from 1 rather than 0 but can change if you feel stronlg they should start from 0. I looked for some precedent in other transforms but couldn't find any.
- Finally, I've renamed the transform to `AddCoordinateChannels`, and provided a bit of motivation for when you might want to use it in the docs. It isn't correct to call it CoordConv because CoordConv changes the convolution layer itself to append spatial channels for each convolution, while my code only appends these channels to the input data as a transform. I think my version has particular use for medical imaging examples where e.g. you might train on patches of 3D volumes that are all aligned to a common space. Appending the coordinate frame to the patch might allow the network to understand where the patch is in the image - CoordConv only appends the local reference coordinates of the input feature map.

rijobro:
Thanks, looks great. Since the output is a concatenation of the channels and the original image, in the unit test it might be worth checking that the first part of the image matches the original image?

I suggested another update to the calculation of the coords (sorry for playing a bit of code golf!) if you want to add that too.

marksgraham:
Thanks - I've taken your suggestion for meshgrid (had to add `indexing='ij'` to get the first two dimensions right), and added that unit test.

rijobro:
Looks good to me, happy to merge once the checks are passing!

wyli (approved):
thanks! @marksgraham @rijobro

PR #2163, opened 2021-05-09
Added Gibbs transform

yanielc:
### Description

Added a Gibbs artifact transform to model Gibbs artifacts in 3D MRI data. The transform has the property to be applied randomly as a naturalistic data augmentation tool.  Gibbs artifacts are one of the typical sources of noise in MRI data.  I added both the transform to the intensity/array.py file and its dictionary partner to the intensity/dictionary.py.

For general information on Gibbs artifacts, please refer to:
 https://pubs.rsna.org/doi/full/10.1148/rg.313105115
 https://pubs.rsna.org/doi/full/10.1148/radiographics.22.4.g02jl14949

### Status
**Work in progress**

### Types of changes


rijobro:
@yanielc thanks! Do you think you could tidy up the git diff? I'm not sure why, but the changes from [this PR](https://github.com/Project-MONAI/MONAI/pull/2144) seem to be present. Perhaps simply clicking "Update branch" will solve it.

yanielc:
> Would you also be able to attach a before and after photo in this thread so we can visually check the transform?

Hi @rijobro, here one can see the transform applied on a sample image from the BraTS dataset.
Gibbs_example

Leftmost is the original volume, then separate applications of the transform with different strengths.

yanielc:
> @yanielc thanks! Do you think you could tidy up the git diff? I'm not sure why, but the changes from [this PR](https://github.com/Project-MONAI/MONAI/pull/2144) seem to be present. Perhaps simply clicking "Update branch" will solve it.

Sure thing, I can try. I also noticed afterwards that my pull request had stuff from other PR you mentioned. I'll try your suggestion.

monai/transforms/intensity/array.py
```diff
@@ -1137,3 +1130,120 @@ def __call__(self, img: np.ndarray) -> np.ndarray:
...
+                you need to transform.
+        alpha: Parametrizes the intensity of the Gibbs noise filter applied. Takes
+            values in [0,1] with alpha = 0 acting as the identity mapping. 
+            If a length-2 list is given as [a,b] then the value of alpha will be 
+            sampled uniformly from the interval [a,b]. 0 <= a <= b <= 1.
+        prob: probability of applying the transform. 
+
+    """
+
+    def __init__(self, alpha: Union[float, List[float]] = 0.5, prob: float = 0.1,) -> None:
```
rijobro, line 1157: Think most MONAI transforms will put the probability at the front. Guess you can also delete the trailing comma.

monai/transforms/intensity/array.py
```diff
@@ -1137,3 +1130,120 @@ def __call__(self, img: np.ndarray) -> np.ndarray:
...
+
+        assert prob <= 1 and prob >= 0, "prob must take values in [0,1]."
```
rijobro, line 1159: delete this since you have the if/else statement

monai/transforms/intensity/dictionary.py
```diff
@@ -1020,6 +1017,58 @@ def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.nda
         return d
 
 
+class RandGibbsNoised(RandomizableTransform, MapTransform):
```
rijobro, line 1020: Do you think you could make this into an `InvertibleTransform`?

monai/transforms/intensity/array.py
yanielc: ok, I'll fix it.

monai/transforms/intensity/dictionary.py
```diff
@@ -1020,6 +1017,58 @@ def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.nda
...
+            given as [a,b] then the value of alpha will be sampled uniformly from the 
+            interval [a,b]. 
+        allow_missing_keys: don't raise exception if key is missing.
+
+    """
+
+    def __init__(
+        self,
+        prob: float = 0.1,
+        keys: KeysCollection = "image",
```
rijobro, line 1047:
```suggestion
        keys: KeysCollection,
        prob: float = 0.1,
```
think all MONAI map transforms are keys first, with no default value.

yanielc: ok, thanks

monai/transforms/intensity/dictionary.py
```diff
@@ -1020,6 +1017,58 @@ def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.nda
...
+    def __call__(self, data: Mapping[Hashable, torch.tensor]) -> Dict[Hashable, torch.tensor]:
+
+        d = dict(data)
+        self.randomize(None)
+
+        if not self._do_transform:
+            return d
+        else:
+            for key in self.key_iterator(d):
+                gibbs = RandGibbsNoise(self.alpha, 1.0)
```
rijobro, line 1067: I'm not sure this logic will work. You'll be creating a different random instance each time and so the image and label will have two different transforms applied to them. I think you'll want to have something in the constructor that creates a single transform: `self.transform = RandGibbsNoise(self.alpha, 1.0)`. And then in `randomize` you'll want to call `self.transform.randomize`.

rijobro: You could also simplify the if else logic to `if self._do_transform:  for key in ...` and then outdented you can have `return d` to reduce duplication (similar to what I suggested for the array transform).

monai/transforms/intensity/array.py
```diff
@@ -1137,3 +1130,120 @@ def __call__(self, img: np.ndarray) -> np.ndarray:
...
+        if type(alpha) == list:
+            assert alpha[1] <= 1 and alpha[0] >= 0, "alpha must take values in [0,1]."
```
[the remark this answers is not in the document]
rijobro: yes, to avoid a cryptic "index out of bound" error if length is less than 2.

monai/transforms/intensity/dictionary.py
yanielc: You are correct. If the user gives `alpha=[a,b]` then the two instances of `RandGibbsNoise` will have different `alpha` values.

Thanks for that catch!

rijobro: Yeah, sounds like separate deterministic and random versions may be necessary, sorry about that!

monai/transforms/intensity/array.py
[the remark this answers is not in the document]
yanielc: Sound good! Will add.

monai/transforms/intensity/array.py
yanielc: Ok, will remove. The __init__ of the `RandomizableTransform ` superclass also has the line`self.prob = min(max(prob, 0.0), 1.0)` which caps any erroneous values of `prob`. My reasoning was to alert the user that he/she had a typo inputting the probability value as oppose to just letting that typo be mapped to 0 or 1 by the min-max.

monai/transforms/intensity/dictionary.py
```diff
@@ -194,14 +198,7 @@ def __init__(
...
         RandomizableTransform.__init__(self, global_prob)
-        self.rand_rician_noise = RandRicianNoise(
-            prob,
-            mean,
-            std,
-            channel_wise,
-            relative,
-            sample_std,
-        )
+        self.rand_rician_noise = RandRicianNoise(prob, mean, std, channel_wise, relative, sample_std,)
```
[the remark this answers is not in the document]
yanielc: I'm removing the comma. I think that runtests --autofix made that edit.

monai/transforms/intensity/array.py
rijobro: Sorry I realise my comment didn't make sense, I mistook `prob` for `alpha` so thought it was code duplication. Anyway, as you say, the `RandomizableTransform` base class should take care of that.

monai/transforms/intensity/dictionary.py
yanielc: I think technically it is possible, but it would be a very bloated transform. The way the transform works is by masking out the father points in k-space. So to make it invertible, it would need to save a 3D image in k-space (as large as the input scan) for each scan that it is applied on.

monai/transforms/intensity/array.py
```diff
@@ -1137,3 +1131,141 @@ def __call__(self, img: np.ndarray) -> np.ndarray:
...
+
+
+    Args:
+        alpha: Parametrizes the intensity of the Gibbs noise filter applied. Takes
+            values in [0,1] with alpha = 0 acting as the identity mapping. 
+    """
+
+    def __init__(self, alpha: Union[float, List[float]] = 0.5) -> None:
+
+        assert alpha <= 1 and alpha >= 0, "alpha must take values in [0,1]."
```
rijobro, line 1203: if alpha is list?

monai/transforms/intensity/dictionary.py
```diff
@@ -1020,6 +1017,70 @@ def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.nda
...
+        return d
+
+    def randomize(self) -> None:
+        """
+        (1) Set random variable to apply the transform.
+        (2) Get alpha from uniform distribution.
+        """
+        super().randomize(None)
+        if type(self.alpha) == list:
+            self.alpha = self.R.uniform(self.alpha[0], self.alpha[1])
```
rijobro, line 1081: since you overwrite `self.alpha`, the randomisation will only work the first time. How about 

```suggestion
        return self.R.uniform(self.alpha[0], self.alpha[1]) if isinstance(self.alpha, list) else self.alpha
```

and then you call it with

```python
alpha = self.randomize()
transform = GibbsNoise(alpha)
```

You'd need to do the same for the array version

monai/transforms/intensity/dictionary.py
rijobro: Yeah makes sense, the information is lost the same as it is with any other crop, just in the k-space instead.

monai/transforms/__init__.py
```diff
@@ -121,6 +123,9 @@
     RandGaussianSmoothd,
     RandGaussianSmoothD,
     RandGaussianSmoothDict,
+    RandGibbsNoised,
```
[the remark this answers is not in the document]
yanielc: I could add it, but by setting prob = 1 and giving specifying alpha (as opposed to giving an interval for alpha) in `RandGibbsNoised` the user obtains deterministic behavior.

Let me know what you think.

monai/transforms/intensity/array.py
rijobro: ah, alpha won't be a list here. In which case you need to remove `List[float]` from the constructor type.

yanielc: `GibbsNoise` is fully deterministic so it takes one alpha (a float). I have relegated the option of `alpha = [a,b]` for the random versions `RandGibbsNoise` and `RandGibbsNoised`. In that case the if the user gives the interval [a,b], the value of alpha will be sampled uniformly from [a,b], and passed to the instatiation of `GibbsNoise` in the call method by `transform = GibbsNoise(self.alpha)`.

monai/transforms/intensity/array.py
```diff
@@ -1137,3 +1131,141 @@ def __call__(self, img: np.ndarray) -> np.ndarray:
...
+
+    def shift_fourier(self, x: torch.tensor) -> torch.tensor:
+        """
+        Applies fourier transform and shifts its output.
+        Only the the last three dimensions get transformed: (x,y,z)-directions.
+        
+        Args: 
+            x (torch.tensor): tensor to fourier transform.
+        """
+        return torch.fft.fftshift(torch.fft.fftn(x, dim=(-3, -2, -1)), dim=(-3, -2, -1))
```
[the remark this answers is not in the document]
rijobro: Is it not possible to have it be generic to both 2 and 3D? For example, hard-coded `(-3, -2, -1)` could be replaced with `tuple(range(-nd,0))` where `nd` is the number of spatial dimensions of the image.

yanielc:
> Aside from the remaining comments, unit tests are needed for all 4 classes (random/deterministic and array/dictionary). Should use a variety of inputs (alpha as float vs. list) etc. Should also be tested on both 2d and 3d images with a variety of number of channels.

Sounds good, I will commit the changes possibly tomorrow and I will include the unit tests.

yanielc:
> Aside from the remaining comments, unit tests are needed for all 4 classes (random/deterministic and array/dictionary). Should use a variety of inputs (alpha as float vs. list) etc. Should also be tested on both 2d and 3d images with a variety of number of channels.

Hi Richard, the commit is ready for review. I added the unit tests as well.

monai/transforms/intensity/array.py
```diff
@@ -1137,3 +1131,164 @@ def __call__(self, img: np.ndarray) -> np.ndarray:
...
+
+    def __init__(self, prob: float = 0.1, alpha: Union[float, List[float]] = 0.5, dim: int = 3) -> None:
+
+        if type(alpha) == list:
+            if alpha[1] > 1 or alpha[0] < 0:
+                raise AssertionError("alpha must take values in the interval [0,1]")
+            if alpha[0] > alpha[1]:
+                raise AssertionError("When alpha = [a,b] we need a < b.")
+            if len(alpha) != 2:
+                raise AssertionError("If alpha is a list, then its length must be 2.")
```
rijobro, line 1167: length check would have to come first to avoid index out of bounds error
```suggestion
            if len(alpha) != 2:
                raise AssertionError("If alpha is a list, then its length must be 2.")
            if alpha[1] > 1 or alpha[0] < 0:
                raise AssertionError("alpha must take values in the interval [0,1]")
            if alpha[0] > alpha[1]:
                raise AssertionError("When alpha = [a,b] we need a < b.")
```

monai/transforms/intensity/array.py
```diff
@@ -1137,3 +1131,164 @@ def __call__(self, img: np.ndarray) -> np.ndarray:
...
+        else:
+            if alpha > 1 or alpha < 0:
+                raise AssertionError("alpha must take values in the interval [0,1].")
+
+        if dim not in [2, 3]:
+            raise AssertionError(("Set dim = 2 or dim = 3 depending on spatial" " dimensions of image."))
```
rijobro, line 1173:
```suggestion
            raise AssertionError("Set dim = 2 or dim = 3 depending on spatial dimensions of image.")
```

monai/transforms/intensity/array.py
```diff
@@ -1137,3 +1131,164 @@ def __call__(self, img: np.ndarray) -> np.ndarray:
...
+
+        self.alpha = alpha
+        self.dim = dim
+        self.sampled_alpha = None  # stores last alpha sampled by randomize()
+        RandomizableTransform.__init__(self, prob=prob)
+
+    def __call__(self, img: torch.tensor) -> torch.tensor:
+
+        # randomize application and possibly alpha
+        self.sampled_alpha = self.randomize()
```
rijobro, line 1183: I'm not sure that we'd want to store the most recent alpha as this might lead to multiprocessing problems. Is that correct @Nic-Ma ?

yanielc: My initial thinking to store the alpha was in case the user wanted to know which alpha was sampled and applied to the data.

yanielc:
@rijobro, the check "GPU-quick-py3 (PT16+CUDA102)" fails because it uses Pytorch 1.6.0 which does not carry the Fourier modules found in `torch.fft` which came out more recently. The checks which use newer versions of Pytorch run ok. Do you have a recommendation here?

rijobro:
Is the numpy version older? We could switch to using that if it existed in numpy>=1.17 (which is our minimum numpy version)

rijobro:
I've also made some suggested changes and created a PR into your branch: https://github.com/yanielc/MONAI/pull/1.

monai/transforms/intensity/array.py
rijobro: I think you're right. Plus, `randomize` doesn't allow us to return anything so we don't have much choice anyway!

yanielc:
> Is the numpy version older? We could switch to using that if it existed in numpy>=1.17 (which is our minimum numpy version)

Yes,  just checked and numpy >=1.17 carries the same fourier functions I am using (forward, backwards transforms, and k-space centering). I can rewrite that part of the code with numpy.

rijobro:
Ok, merge my PR first (if you're happy of course) to avoid any merge conflicts.

yanielc:
> Ok, merge my PR first (if you're happy of course) to avoid any merge conflicts.

Just did :)

yanielc:
> Ok, merge my PR first (if you're happy of course) to avoid any merge conflicts.

I will test more on my side, but your changes look really nice. Thank you. 

One thing I was assuming, perhaps unnecessarily, was that the number of leading dimensions was arbitrary, say (..., channel, H, W) for 2D and (..., channel, H, W, D) for 3D. Hence my mask method was a bit less direct.

monai/transforms/intensity/array.py
```diff
@@ -1137,3 +1131,158 @@ def __call__(self, img: np.ndarray) -> np.ndarray:
...
+
+    """
+
+    def __init__(self, alpha: float = 0.5, as_tensor_output: bool = True) -> None:
+
+        if alpha > 1 or alpha < 0:
+            raise AssertionError("alpha must take values in the interval [0,1].")
+        self.alpha = alpha
+        self.as_tensor_output = as_tensor_output
+        self._device = "cpu"
```
rijobro, line 1223: Did you add this or did I? Doesn't seem we're using it.

rijobro:
Looks really good, just need to get all the checks green and then I think this is good to go. Thanks!

rijobro:
> One thing I was assuming, perhaps unnecessarily, was that the number of leading dimensions was arbitrary, say (..., channel, H, W) for 2D and (..., channel, H, W, D) for 3D. Hence my mask method was a bit less direct.

I think all of our transforms assume `(C, H, W, [D])` with no batch. I think this is a safe assumption here as otherwise all our other transforms would fall over.

monai/transforms/intensity/array.py
yanielc: I did.  The operations in `GibbsNoise` are now done with numpy because we saw failures due to Pytorch 1.6 lacking some of the fourier functions. So I save the device of the input tensor in `_device` before going to numpy. Then I make sure to put the tensor to the its original device at the end of `__call___` via `return torch.Tensor(img).to(self._device) if self.as_tensor_output else img`.

rijobro: Ah, I see. Great.

yanielc:
> Looks really good, just need to get all the checks green and then I think this is good to go. Thanks!

There are currently two tests failing. One is **build-docs**. The interesting thing is that when I run `make html` locally, I no longer see any warnings or errors. So I am a bit uncertain what to do about the error it shows on the repository:

/home/runner/work/MONAI/MONAI/monai/transforms/intensity/array.py:docstring of monai.transforms.intensity.array.GibbsNoise.shift_fourier:1: WARNING: duplicate object description of monai.transforms.GibbsNoise.shift_fourier, other instance in transforms, use :noindex: for one of them

The other error is simpler. It's **flake8**. When I run it locally it wants to reformat more files than the ones I have edited it. Can I let it do that?

Thank you for all guidance!

docs/source/transforms.rst
```diff
@@ -262,6 +262,30 @@ Intensity
...
 
+`GibbsNoise`
+""""""""""""""
+.. autoclass:: GibbsNoise
+    :members:
+    :special-members: __call__
+
+`RandGibbsNoise`
+"""""""""""""""""
+.. autoclass:: GibbsNoise
```
rijobro, line 273:
```suggestion
.. autoclass:: RandGibbsNoise
```

rijobro:
@wyli @Nic-Ma @ericspod – do you think we could revisit the use of DCO? I committed from a [github suggestion](https://github.com/Project-MONAI/MONAI/pull/2163#discussion_r632400085), which apparently doesn't do the DCO (so not very useful), but I can't force push since it's not my branch...

wyli:
> @wyli @Nic-Ma @ericspod – do you think we could revisit the use of DCO? I committed from a [github suggestion](https://github.com/Project-MONAI/MONAI/pull/2163#discussion_r632400085), which apparently doesn't do the DCO (so not very useful), but I can't force push since it's not my branch...

there is a discussion for the DCO here https://github.com/probot/dco/issues/141

yanielc:
@rijobro, almost all green marks! I think you saw there is just one missing sign-off from a previous commit of yours.

rijobro (approved):

rijobro:
tell me when you're happy. I'll have to manually set the DCO check to pass, and then enable auto-merge.

yanielc:
> tell me when you're happy. I'll have to manually set the DCO check to pass, and then enable auto-merge.

Ok, I think I am done with changes on my end.

PR #2165, opened 2021-05-10
2145 Extract invert logic from `TransformInverter` handler

### LINKED ISSUE #2145: extract main logic of TransformInverter into utils
Need to extract the main logic of `TransformInverter` handler into utils, then not only ignite users can use this feature.

Nic-Ma:
Fixes #2145 .

### Description
This PR extracted the main logic of `TransformInverter` to a post transform. So non-ignite users can freely use this feature in the regular PyTorch program.
It also moved the `BatchInverseTransform` from "monai/data/" to "monai/transforms/" as it's a transform.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @rijobro ,

I slightly changed the `_BatchInverseDataset` to avoid circle-import, I didn't figure out why the TTA unit tests failed...
If adding "self.XXX = transform" in the init of `_BatchInverseDataset`, the unit tests can pass (XXX can be any word).
Could you please help take a look? Maybe this is some potential logic in your TTA module that I ignored?

Thanks in advance.

Nic-Ma:
Paste the error log:
```
ERROR: test_single_transform (tests.test_testtimeaugmentation.TestTestTimeAugmentation)
  File "/home/runner/work/MONAI/MONAI/monai/data/utils.py", line 264, in list_data_collate
    ret[k] = default_collate([d[k] for d in data])
  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py", line 55, in default_collate
    return torch.stack(batch, 0, out=out)
TypeError: expected Tensor as element 1 in argument 0, but got numpy.ndarray

During handling of the above exception, another exception occurred:

  File "/home/runner/work/MONAI/MONAI/tests/test_testtimeaugmentation.py", line 147, in test_single_transform
    tta(self.get_data(1, (20, 20)))
  File "/home/runner/work/MONAI/MONAI/monai/data/test_time_augmentation.py", line 183, in __call__
    inv_batch = inverter(inferred_dict)
  File "/home/runner/work/MONAI/MONAI/monai/transforms/inverse_batch_transform.py", line 92, in __call__
    return first(inv_loader)
  File "/home/runner/work/MONAI/MONAI/monai/utils/misc.py", line 71, in first
    for i in iterable:
  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 517, in __next__
    data = self._next_data()
  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 557, in _next_data
    data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/torch/utils/data/_utils/fetch.py", line 47, in fetch
    return self.collate_fn(data)
  File "/home/runner/work/MONAI/MONAI/monai/data/utils.py", line 288, in list_data_collate
    raise TypeError(re_str)
TypeError: expected Tensor as element 1 in argument 0, but got numpy.ndarray
Collate error on the key 'label' of dictionary data.

MONAI hint: if your transforms intentionally create mixtures of torch Tensor and numpy ndarray, creating your `DataLoader` with `collate_fn=pad_list_data_collate` might solve this problem (check its documentation).
```

rijobro:
I just tried the current state of your PR and `test_testtimeaugmentation.py` passes. Have you solved the problem? Or do I need to change something to be able to reproduce the problem?

Nic-Ma:
Hi @rijobro ,

I think the root cause is that: the 2 failed unit tests of TTA missed `ToTensorD()` transform, so it means that we invert a Tensor data in a random transform, if `do_transform`, it will convert to `numpy array`, if not, still be `Tensor`, so sometimes can't **collate a batch**.
As it's random behavior, adding `self.transform = transform` may affect the random result and if I change `prob=0.2`, it also failed.
Now I just changed `prob=1.0` in the TTA unit tests.
I think maybe we can change the `inverse` logic to convert Tensor -> numpy no matter do_transform is True or not, 
Or we change the TTA logic to convert `Tensor -> numpy` before inverse, now it's `numpy -> Tensor`.
what do you think?

Thanks.

rijobro:
I see. Perhaps best to implement both, since the inverse of `ToTensor` is `ToNumpy`, so that makes sense. Also inverse transforms will always need to be preceded by a `ToTensor` logic.

Nic-Ma:
Hi @rijobro ,

Thanks for your help, yes, `ToTensorD` is supposed to be used in all MONAI transform chains, so the test failure is not a regular use case.
But any way, we can enhance the `numpy` and `Tensor` logic to make it more clear later.
This PR mainly decoupled `invert transform` with `ignite`, users can use it as post transform in regular program.

Thanks.

rijobro (approved):
sounds good, thanks for this

Nic-Ma:
Hi @vfdev-5 ,

In this PR, I temporarily combined `state.batch` and `state.output` to construct a `unique dict` for the dict-based transform which is decoupled from ignite, just as we discussed in the ignite ticket before.
Could you please double confirm whether this PR has any potential issues?

Thanks.

monai/handlers/transform_inverter.py
```diff
@@ -114,42 +112,17 @@ def __call__(self, engine: Engine) -> None:
...
-                transform_key: transform_info,
-            }
-            meta_dict_key = f"{batch_key}_{self.meta_key_postfix}"
-            if meta_dict_key in engine.state.batch:
-                segs_dict[meta_dict_key] = engine.state.batch[meta_dict_key]
-
-            with allow_missing_keys_mode(self.transform):  # type: ignore
-                inverted = self.inverter(segs_dict)
+        # combine `batch` and `output` to temporarily act as 1 dict for post transform
+        data = dict(engine.state.batch)
```
vfdev-5, line 116: I think, we have to update the docs saying that the engine should have `state.batch` and `state.output` as dicts, otherwise it will fail here for any engine (e.g. those where batch is a tuple of 2 items).

vfdev-5:
@Nic-Ma I left a small comment in the code. Otherwise, I do not see any other issues with this approach.

Thanks for pinging!

monai/handlers/transform_inverter.py
Nic-Ma: Hi @vfdev-5 ,

Thanks for your review!
Yes, you made a good point, I added more doc-string in the latest commit.

Thanks.

PR #2166, opened 2021-05-10
enhance norm and act layers

### LINKED ISSUE #2164: Support **kwargs for some pytorch args of networks
For networks like `SegResNet`, we cannot change parameters for normalization layers (such as set `track_running_stats=True` when we use BatchNorm3d). Therefore, the implementation should be more generic.

yiheng-wang-nv:

Fixes #2164 .

### Description
Enhance norm and act layers for segresnet and dynunet.

### Status
**Ready**

### Types of changes


monai/networks/nets/segresnet.py
```diff
@@ -213,30 +192,10 @@ def __init__(
...
-        in_channels: int = 1,
-        out_channels: int = 2,
-        dropout_prob: Optional[float] = None,
-        norm_name: str = "group",
-        num_groups: int = 8,
-        use_conv_final: bool = True,
-        blocks_down: tuple = (1, 2, 2, 4),
-        blocks_up: tuple = (1, 1, 1),
-        upsample_mode: Union[UpsampleMode, str] = "nontrainable",
+        **kwargs,
```
Nic-Ma, line 195: Actually, I think maybe it would be more clear to leave the args for users to check directly.

monai/networks/blocks/segresnet_block.py
```diff
@@ -9,28 +9,31 @@
...
-    if norm_name == "group":
-        norm = Norm[norm_name](num_groups=num_groups, num_channels=in_channels)
-    else:
-        norm = Norm[norm_name, spatial_dims](in_channels)
-    if norm.bias is not None:
-        nn.init.zeros_(norm.bias)
-    if norm.weight is not None:
-        nn.init.ones_(norm.weight)
-    return norm
+def get_norm_layer(spatial_dims: int, in_channels: int, norm: Union[Tuple, str]):
```
Nic-Ma, line 22: Can we make it to be a `utils function` and share it between networks? I see several networks have this duplicated logic now.

monai/networks/blocks/dynunet_block.py
```diff
@@ -31,9 +32,8 @@ class UnetResBlock(nn.Module):
         out_channels: number of output channels.
         kernel_size: convolution kernel size.
         stride: convolution stride.
-        norm_name: [``"batch"``, ``"instance"``, ``"group"``]
-            feature normalization type and arguments. In this module, if using ``"group"``,
-            `in_channels` should be divisible by 16 (default value for ``num_groups``).
+        norm_name: feature normalization type and arguments.
```
Nic-Ma, line 35: Should we still keep the doc-string for users: "if using ``"group"``, `in_channels` should be divisible by 16 (default value for ``num_groups``)."?

monai/networks/nets/segresnet.py
ericspod: I would generally have arguments set out in full rather than hiding a lot of things behind args/kwargs. Users rely on the signature and docstring of a constructor to know how to create an object, it's much harder when that detail is hidden.

ericspod (commented):
This is a good improvement and using the factories the way I think we should. There's a few comments to address but then we can approve.

monai/networks/blocks/dynunet_block.py
yiheng-wang-nv: Now we leave the parameter for users to type, and we do not have the ``num_groups `` parameter in the class, thus I suggest remove this part.

wyli (approved):
thanks! this PR introduces breaking changes for `SegResNet`, removed `norm_name` and `num_groups` parameters, added `act` and `norm` parameter. but the default behaviour doesn't change.

PR #2171, opened 2021-05-10
initial network consistency test

rijobro:
Suggested format of testing network consistency. Uses a git submodule to store the data so as not to clog up the main repository. Currently this submodule sits in my user area, but this would obviously be migrated to Project-MONAI.

The idea would be that if the submodule is present, it'll loop over each of the folders in the submodule, load the network (required creating a network factory), and checking consistency of forward pass.

### Status
**Ready**

### Types of changes

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -281,7 +281,7 @@ jobs:
         python -m pip list
     - name: Run quick tests (GPU)
       run: |
-        git clone --depth 1 --branch master --single-branch \
+        git clone --depth 1 \
           https://github.com/Project-MONAI/MONAI-extra-test-data.git /MONAI-extra-test-data
         export MONAI_EXTRA_TEST_DATA="/MONAI-extra-test-data"
         nvidia-smi
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -260,7 +260,8 @@ jobs:
           libboost-test-dev \
           libgoogle-glog-dev \
           libjsoncpp-dev \
-          cmake && \
+          cmake \
+          git && \
         rm -rf /var/lib/apt/lists/* && \
         export PYTHONIOENCODING=utf-8 LC_ALL=C.UTF-8 && \
         rm -f /usr/bin/python && \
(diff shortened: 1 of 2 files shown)


rijobro:
Not using a git submodule (I agree they can be tricky) means that we either abandon this idea, store the data in this repository, or store it somewhere else. What do you think? 

The data for the 3d unet was ~7.5MB, so for all networks we might be looking at something like ~100MB.

rijobro:
@ericspod suggested (offline) simply git cloning the test repository, rather than having it as a submodule. Seems like an easy alternative @wyli

wyli:
thanks, perhaps we can start a new repo under project-monai to host the test data with git-lfs?

rijobro:
That sounds like a good plan. I hadn't heard of git lfs before!

rijobro:
@wyli unrelated error, the added test passes: https://github.com/Project-MONAI/MONAI/pull/2171/checks?check_run_id=2574279151#step:6:2046.

.github/workflows/pythonapp.yml
```diff
@@ -280,6 +281,8 @@ jobs:
         python -m pip list
     - name: Run quick tests (GPU)
       run: |
+        git clone https://github.com/Project-MONAI/MONAI-extra-test-data.git /MONAI-extra-test-data
```
wyli, line 284: could have git clone with depth 1 to avoid fetching all the history?

tests/test_network_consistency.py
```diff
@@ -0,0 +1,69 @@
...
+        # Load data
+        loaded_data = torch.load(data_path)
+
+        # Load json from file
+        json_file = open(json_path)
+        model_params = json.load(json_file)
+        json_file.close()
+
+        # Create model
+        factory = NetworkFactory()
```
wyli, line 55: could we use these simple commands for now:
```py
import monai.networks.nets as nets
nets.__dict__['UNet']
```
I feel the network factory idea should have a separate discussion

wyli (approved):
thanks, I think this is a good initial example, we could merge it if it's decoupled from the network factory idea.

tests/test_network_consistency.py
rijobro: good suggestion, thanks!

Issue #2161, opened 2021-05-08
CUDA_VISIBLE_DEVICES environ setting not work in ver. 0.5.0


ChenglongWang:
**Describe the bug**
I'm using `os.environ["CUDA_VISIBLE_DEVICES"] = '0'` to set available GPU instead of specifing device like `cuda:0`. 
Everthing works fine until updating to ver. 0.5.0.  Maybe some GPU-related operations were added in the latest version during import phase.


**To Reproduce**
1. Use one [tutorial](https://github.com/Project-MONAI/tutorials/blob/master/2d_segmentation/torch/unet_training_array.py) for example. 
2. Default behavior: All GPUs are detected and only one gpu is actually used
3. Add `os.environ["CUDA_VISIBLE_DEVICES"] = '0'` after `def main(tempdir):` on line 31.
  - ver. 0.4.x : Only GPU-0 is detected and used.
  - ver. 0.5.0 : All GPUs are detected.


**Environment**
MONAI version: 0.5.0
Numpy version: 1.19.4
Pytorch version: 1.6.0
MONAI flags: HAS_EXT = False, USE_COMPILED = False
MONAI rev id: 2707407fed8c78ccb1c18d1e994a68580457219e

Optional dependencies:                                                                                                                                                                                                                                                        Pytorch Ignite version: 0.4.4
Nibabel version: 3.2.1
scikit-image version: 0.17.2  
Pillow version: 8.0.1
Tensorboard version: 2.5.0a20201221 
gdown version: NOT INSTALLED or UNKNOWN VERSION.                                                                                                                                                                                                                              TorchVision version: 0.7.0
ITK version: 5.1.2
tqdm version: 4.54.1
lmdb version: NOT INSTALLED or UNKNOWN VERSION.
psutil version: 5.7.3

rijobro:
Also I think it's unlikely that MONAI is the cause of your problem, since MONAI doesn't (and I don't think ever has) use the variable `CUDA_VISIBLE_DEVICES`. It can be set with `monai.config.deviceconfig.set_visible_devices(0)`, but this will just call `os.environ["CUDA_VISIBLE_DEVICES"] = '0'`. The rest gets handled by pytorch.

If you think I've missed something, would it be possible to create a minimum example that highlights your point? Preferably using all the same versions for the dependencies and simply switching between `0.4.x` and `0.5.x`? Thanks.

rijobro:
[...]

PR #2174, opened 2021-05-11
fix set visible devices

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/2161.

### Description
In TestTimeAugmentation, I called `torch.cuda` as a default argument to a constructor. This meant that the following code previously worked and now didn't:

```python
import os
import monai
os.environ["CUDA_VISIBLE_DEVICES"] = '0'
```

I fixed the problem and also created a unit test to avoid it happening in the future. The trick is that we shouldn't call `torch.cuda.` during `import monai`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/test_time_augmentation.py
+++ b/monai/data/test_time_augmentation.py
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Tuple, Union
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


monai/data/test_time_augmentation.py
```diff
@@ -89,7 +89,7 @@ def __init__(
         batch_size: int,
         num_workers: int,
         inferrer_fn: Callable,
-        device: Optional[Union[str, torch.device]] = "cuda" if torch.cuda.is_available() else "cpu",
+        device: Optional[Union[str, torch.device]] = None,
```
Nic-Ma, line 92: I think maybe it's more clear to just use ` cpu` as the default value, `auto-select` logic is easy to cause unexpected behavior. You can refer to the ignite approach:
https://github.com/pytorch/ignite/blob/master/ignite/metrics/metric.py#L207
Which just use:
```py
device: Union[str, torch.device] = torch.device("cpu")
self.device = device
```
What do you think?
Others look good to me.

Thanks.

Nic-Ma (approved):

PR #2178, opened 2021-05-12
2177 Add max_roi_size to RandSpatialCrop

### LINKED ISSUE #2177: Add max_roi_size support in `RandSpatialCrop`
Currently, when we do spatial crop, we can select a random size in (roi_size, image_size) where roi_size is the min_size. So we also need to support a max roi_size.

Nic-Ma:
Fixes #2177 .

### Description
This PR added support to limit the max random crop size in `RandSpatialCrop`.

### Status
**Ready**

### Types of changes


rijobro:
Do you think it would look nicer if, instead of adding `max_roi_size`, we modified `roi_size` to also take `Sequence[Sequence[int]]` i.e., `roi_size=((10,10,10), (20, 20, 20))`? It's just a shame that we have `roi_size` and `max_roi_size`, where `roi_size` actually means `min_roi_size`.

monai/transforms/croppad/array.py
```diff
@@ -342,9 +351,13 @@ class RandSpatialCropSamples(Randomizable):
     It will return a list of cropped images.
 
     Args:
-        roi_size: if `random_size` is True, the spatial size of the minimum crop region.
-            if `random_size` is False, specify the expected ROI size to crop. e.g. [224, 224, 128]
+        roi_size: if `random_size` is True, it specifies the minimum crop region.
+            if `random_size` is False, it specifies the expected ROI size to crop. e.g. [224, 224, 128]
+            If its components have non-positive values, the corresponding size of input image will be used.
         num_samples: number of samples (crop regions) to take in the returned list.
+        max_roi_size: if `random_size` is True and `roi_size` specified the min crop region size, `max_roi_size`
```
yiheng-wang-nv, line 358: `roi_size` specifies

monai/transforms/croppad/array.py
```diff
@@ -298,15 +298,23 @@ class RandSpatialCrop(Randomizable):
         roi_size: if `random_size` is True, it specifies the minimum crop region.
             if `random_size` is False, it specifies the expected ROI size to crop. e.g. [224, 224, 128]
             If its components have non-positive values, the corresponding size of input image will be used.
+        max_roi_size: if `random_size` is True and `roi_size` specified the min crop region size, `max_roi_size`
+            can specify the max crop region size. if None, default to the input image size.
```
yiheng-wang-nv, line 302: defaults to

monai/transforms/croppad/array.py
```diff
@@ -342,9 +351,13 @@ class RandSpatialCropSamples(Randomizable):
...
 
     Args:
-        roi_size: if `random_size` is True, the spatial size of the minimum crop region.
-            if `random_size` is False, specify the expected ROI size to crop. e.g. [224, 224, 128]
+        roi_size: if `random_size` is True, it specifies the minimum crop region.
+            if `random_size` is False, it specifies the expected ROI size to crop. e.g. [224, 224, 128]
+            If its components have non-positive values, the corresponding size of input image will be used.
         num_samples: number of samples (crop regions) to take in the returned list.
+        max_roi_size: if `random_size` is True and `roi_size` specified the min crop region size, `max_roi_size`
+            can specify the max crop region size. if None, default to the input image size.
```
yiheng-wang-nv, line 359: defaults to

monai/transforms/croppad/dictionary.py
```diff
@@ -410,6 +410,9 @@ class RandSpatialCropd(Randomizable, MapTransform, InvertibleTransform):
         roi_size: if `random_size` is True, it specifies the minimum crop region.
             if `random_size` is False, it specifies the expected ROI size to crop. e.g. [224, 224, 128]
             If its components have non-positive values, the corresponding size of input image will be used.
+        max_roi_size: if `random_size` is True and `roi_size` specified the min crop region size, `max_roi_size`
+            can specify the max crop region size. if None, default to the input image size.
```
yiheng-wang-nv, line 414: defaults to

monai/transforms/croppad/dictionary.py
```diff
@@ -410,6 +410,9 @@ class RandSpatialCropd(Randomizable, MapTransform, InvertibleTransform):
         roi_size: if `random_size` is True, it specifies the minimum crop region.
             if `random_size` is False, it specifies the expected ROI size to crop. e.g. [224, 224, 128]
             If its components have non-positive values, the corresponding size of input image will be used.
+        max_roi_size: if `random_size` is True and `roi_size` specified the min crop region size, `max_roi_size`
```
yiheng-wang-nv, line 413: specifies

monai/transforms/croppad/dictionary.py
```diff
@@ -499,9 +505,13 @@ class RandSpatialCropSamplesd(Randomizable, MapTransform):
...
         keys: keys of the corresponding items to be transformed.
             See also: monai.transforms.MapTransform
-        roi_size: if `random_size` is True, the spatial size of the minimum crop region.
-            if `random_size` is False, specify the expected ROI size to crop. e.g. [224, 224, 128]
+        roi_size: if `random_size` is True, it specifies the minimum crop region.
+            if `random_size` is False, it specifies the expected ROI size to crop. e.g. [224, 224, 128]
+            If its components have non-positive values, the corresponding size of input image will be used.
         num_samples: number of samples (crop regions) to take in the returned list.
+        max_roi_size: if `random_size` is True and `roi_size` specified the min crop region size, `max_roi_size`
+            can specify the max crop region size. if None, default to the input image size.
```
yiheng-wang-nv, line 513: defaults to

Nic-Ma:
Hi @rijobro ,

Thanks for your suggestion, but the problem is that:
1. `roi_size` only acts as "min" size when `random_size=True`, if False, it's the fixed roi size to crop.
2. We also support to set **only 1 value for all the spatial dims** in `roi_size`, for example: `roi_size=128`, if we put `max_roi_size` input `roi_size` as you suggested, when crop a 2D image, what `roi_size=(128, 192)` means? It can be roi=128 and max_roi_size=192, or real roi_size=(128, 192), right?

What do you think?

Thanks.

monai/transforms/croppad/dictionary.py
```diff
@@ -434,7 +439,8 @@ def __init__(
     def randomize(self, img_size: Sequence[int]) -> None:
         self._size = fall_back_tuple(self.roi_size, img_size)
         if self.random_size:
-            self._size = [self.R.randint(low=self._size[i], high=img_size[i] + 1) for i in range(len(img_size))]
+            max_size = img_size if self.max_roi_size is None else fall_back_tuple(self.max_roi_size, img_size)
+            self._size = [self.R.randint(low=self._size[i], high=max_size[i] + 1) for i in range(len(img_size))]
```
yiheng-wang-nv, line 443: Errors will happen if `low > high`, thus I suggest add a comparison between `roi_size` and `max_roi_size` in `__init__`, and could you please add include one more test case to cover this situation.

yiheng-wang-nv (commented):
Thanks Nic, I added some comments for this PR.

rijobro:
@Nic-Ma Ok makes sense. I suppose we can't change the behaviour of `roi_size` either, because of backwards compatibility. If we were designing it now, I would have `roi_size`, `min_roi_size` and `max_roi_size` where you can specify the first one or the latter two. Ah well, looks good! 

I'll approve but agree with @yiheng-wang-nv 's suggested changes.

rijobro (approved):

Nic-Ma:
Hi @rijobro and @yiheng-wang-nv ,

Thanks for your review, I will try to update the PR to address all the comments.

monai/transforms/croppad/dictionary.py
Nic-Ma: added check for min / max ROI size.

PR #2181, opened 2021-05-12
More efficient SmartCacheLoader

cgrain:
### Description

On [this website](https://sagivtech.com/2017/09/19/optimizing-pytorch-training-code/) it is advised to delete tensors explicitly (with `del`) when the program is done with it. Applying that, I have made a small change in SmartCacheLoader that reflects that. I have also added a small unit test that ensures that the updated cache is indeed what it needs to be. 

### Status
**Ready**

### Types of changes

### DIFF
--- a/tests/test_smartcachedataset.py
+++ b/tests/test_smartcachedataset.py
@@ -70,6 +70,38 @@ class TestSmartCacheDataset(unittest.TestCase):
                         self.assertIsInstance(dataset[15]["image"], str)
                 dataset.shutdown()
 
+    def test_update_cache(self):
+       # Given
+        test_data = [{"image": f"test_image{i}.nii.gz",
+                        "label": f"test_image{i}.nii.gz"
+                        } for i in range(40)] 
+        dataset = Smart_New_Dataset(
+            data=test_data,
+            transform=None,
+            replace_rate=0.2,
+            cache_num=10,
+            num_init_workers=4,
+            num_replace_workers=4,
+            shuffle=False,
+        )
+        dataset.start()
+        start_num = int(0.2 * 10)
+        remain_num = int((1-0.2) * 10)
+        
+        old_cache = copy.deepcopy(dataset._cache)
+        # When
+        with dataset._update_lock:
+            replacements = copy.deepcopy(dataset._replacements)
+        dataset.update_cache()
+        new_cache = dataset._cache 
+        kept_cache = old_cache[start_num:]
+        # Then
+        for string1, string2 in zip(kept_cache, new_cache[0:remain_num]):
+            assert(string1 == string2)
+        for string_new, string_replacement in zip(replacements, new_cache[remain_num:]): 
+            assert(string_new == string_replacement)
+        
+    
     def test_shuffle(self):
         test_data = [{"image": f"test_image{i}.nii.gz"} for i in range(20)]
         dataset = SmartCacheDataset(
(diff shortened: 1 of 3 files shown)


rijobro:
Hi, I read through the blog you linked, and guess you're referring to this paragraph:

> Free up memory using del
This is a common pitfall for new PyTorch users, and we think it isn’t documented enough.
After you’re done with some PyTorch tensor or variable, delete it using the python del operator to free up memory.

In the blog post there is no data to back up this claim. As such, I think we'd need to see some numbers that your implementation uses less memory than the current one.

rijobro:
Although I agree that the current implementation is a little ugly, it seems to have unnecessary for loops. How about:

```python
self._cache = self._cache[:remain_num] + self._replacements[:self._replace_num]
```

@wyli

cgrain:
Thank you for your quick reply! 

Your question for some data is very reasonable, and I have only some anecdotal data unfortunately. I have actually implemented my method in my own project, and instead of SLURM killing my job because I was out of memory, the job ran without problems. Of course, I have not implemented nor tested your method. I will see if I can create a test that shows the difference in data use. The problem to check that thoroughly is that the garbage collector of python (which is not completely determenistic) can either be quick, which will result in a small difference, or slow, which will result in higher memory spikes.

rijobro:
Sounds good, I'm curious to see what your benchmarking gives. My worry is that if we can't trust the garbage collector here, then we can't trust it anywhere in the code base and will have to delete everything everywhere. My default position is therefore to trust the garbage collector!

wyli:
did a quick test it seems `test2` is good:
```py
import random

def test1(cache_num=100, replace_num=20):
  _cache = list(range(cache_num))
  replace = [random.random() for x in range(replace_num)]
  remain_num: int = cache_num - replace_num
  for i in range(remain_num):
    _cache[i] = _cache[i + replace_num]
  for i in range(replace_num):
    _cache[remain_num + i] = replace[i]
  return _cache

def test2(cache_num=100, replace_num=20):
  _cache = list(range(cache_num))
  replace = [random.random() for x in range(replace_num)]

  del _cache[:replace_num]
  _cache.extend(replace)
  return _cache

def test3(cache_num=100, replace_num=20):
  _cache = list(range(cache_num))
  replace = [random.random() for x in range(replace_num)]
  remain_num: int = cache_num - replace_num
  _cache = _cache[replace_num:replace_num + remain_num] + replace[:replace_num]
  return _cache
```

Screenshot 2021-05-12 at 13 45 25

wyli:
thanks @cgrain, could you address this DCO issue by signing the commits?  https://github.com/Project-MONAI/MONAI/pull/2181/checks?check_run_id=2565463758

cgrain:
Thanks @wyli, DCO should be signed now!

tests/test_smartcachedataset.py
```diff
@@ -70,6 +71,35 @@ def test_shape(self, replace_rate, num_replace_workers, transform):
                         self.assertIsInstance(dataset[15]["image"], str)
                 dataset.shutdown()
 
+    def test_update_cache(self):
+        # Given
+        test_data = [{"image": f"test_image{i}.nii.gz", "label": f"test_image{i}.nii.gz"} for i in range(40)]
+        dataset = Smart_New_Dataset(
```
wyli, line 77: Smart_New_Dataset is undefined here

wyli (approved):

PR #2194, opened 2021-05-13
2180 Add RandScaleCrop transform

### LINKED ISSUE #2180: Add `RandScaleCrop` transform and inverse operation
It would be useful to randomly crop images with specified scale arg as min & max size, for example: (0.3, 0.6). Then images with different sizes will use different crop size range.

Nic-Ma:
Fixes #2180 .

### Description
This PR added the `RandScaleCrop` transform to crop images with different sizes but same scale ratio.

### Status
**Ready**

### Types of changes


rijobro:
Should we have a deterministic version? I know we can have deterministic behaviour by setting the scale, but it won't be cached by `CacheDataset` for example.

Nic-Ma:
> Should we have a deterministic version? I know we can have deterministic behaviour by setting the scale, but it won't be cached by `CacheDataset` for example.

Sounds interesting, let me try to implement a `ScaleCrop` transform in another PR next week.

Thanks.

monai/transforms/croppad/array.py
```diff
@@ -345,6 +346,52 @@ def __call__(self, img: np.ndarray):
...
+        self.max_roi_scale = max_roi_scale
+
+    def __call__(self, img: np.ndarray):
+        """
+        Apply the transform to `img`, assuming `img` is channel-first and
+        slicing doesn't apply to the channel dim.
+        """
+        img_size = img.shape[1:]
+        ndim = len(img_size)
+        self.roi_size = [int(r * s) for r, s in zip(ensure_tuple_rep(self.roi_scale, ndim), img_size)]
```
wyli, line 387: need to double check this `int(r*s)` if img_size is small for example (192, 192, 5), and scale is 0.1

monai/transforms/croppad/array.py
```diff
@@ -303,7 +304,7 @@ class RandSpatialCrop(Randomizable):
             if its components have non-positive values, the corresponding size of input image will be used.
         random_center: crop at random position as center or the image center.
         random_size: crop with random size or specific size ROI.
-            The actual size is sampled from `randint(roi_size, img_size)`.
+            if True, the actual size is sampled from `randint(roi_size, max_roi_size)`.
```
wyli, line 307: the `randint` API from  numpy doesn't include the higher value, the interval is [low, high), could you help check the related edge cases..

wyli (commented):
thanks, I put minor comments inline

monai/transforms/croppad/array.py
Nic-Ma: Hi @wyli , I already added doc-string to indicate this case:
"If its components have non-positive values or very small scale value causes the crop size to be zero, will use `1.0` instead, which means the input image size."

Thanks.

monai/transforms/croppad/array.py
Nic-Ma: sure, updated in the latest commit.
Thanks.

wyli (approved):
thanks!

PR #2195, opened 2021-05-13
add network factory

rijobro:
Add network factory. Required renaming `LayerFactory` to the more general `ObjectFactory`, and moving it to a more suitable location.

### Status
**Ready**

### Types of changes


tests/test_network_factory.py
```diff
@@ -0,0 +1,54 @@
...
+        {"dimensions": 3, "in_channels": 4, "out_channels": 2, "channels": (4, 8, 16, 32), "strides": (2, 4, 1)},
+    ],
+    [DenseNet, "DenseNet", {"spatial_dims": 2, "in_channels": 4, "out_channels": 2}],
+]
+
+
+class TestNetworkFactory(unittest.TestCase):
+    @parameterized.expand(TESTS)
+    def test_network_factory(self, class_name, name, args):
+        net = NetworkFactory.get_constructor(name, **args)
```
rijobro, line 31: If anyone could help, I can't figure out how to do `NetworkFactory[name](**kwargs)` or `NetworkFactory[name, **kwargs]` or something similar to construct the object the network with kwargs.

Nic-Ma: Hi @ericspod , could you please help here as you defined the other factories before?

Thanks.

monai/networks/nets/autoencoder.py
```diff
@@ -16,8 +16,10 @@
 
 from monai.networks.blocks import Convolution, ResidualUnit
 from monai.networks.layers.factories import Act, Norm
+from monai.networks.nets.net_factory import NetworkFactory
 
 
+@NetworkFactory.factory_function("AutoEncoder")
```
ericspod, line 22: This decorator is meant for callable objects that return the type, so these shouldn't applied to a class. For example we have 

```python
@Dropout.factory_function("alphadropout")
def alpha_dropout_factory(_dim):
      return nn.AlphaDropout
```

The decorator is applied to a function in this case but it could be to a class to act as another factory whose `__call__` method would have to return newly created objects. That's not what `AutoEncoder` is so we need a different mechanism, either adding a function like the example or we add a new decorator to the factories that accept class types as arguments and store a callable factory function into the factory.

monai/utils/factories.py
```diff
@@ -0,0 +1,151 @@
...
+    def factory_function(self, name: str) -> Callable:
+        """
+        Decorator for adding a factory function with the given name.
+        """
+
+        def _add(func: Callable) -> Callable:
+            self.add_factory_callable(name, func)
+            return func
+
+        return _add
```
ericspod, line 107: I'd suggest this decorator method for class types:

```python
def factory_type(self, name: Optional[str]=None) -> Callable:
    """
    Decorator for class types, adding a factory function taking no arguments. This is stored under `name` or the class'
    name if not provided.
    """
    def _add(cls: type) -> type:
        self.add_factory_callable(name or cls.__name__, lambda: cls)
        return cls

    return _add
```

This would be used as 

```python
fact=ObjectFactory()

@fact.factory_type()
class Foo:
    def __init__(self,a,b):
        pass

foo_inst = fact["Foo"](a=1,b=2)
```

I *think* that'll work but I haven't tried it.

rijobro:
Hi @ericspod suggestions worked. Can now use `NetworkFactory["UNet"](**kwargs)`. Thanks. Any other comments?

wyli:
Is it possible to implement the idea in a separate module, so that we keep the readability of the network classes (without introducing a decorator to each network)
for example `networks/registry.py`
```py
from . import NetworkFactory
import monai

NetworkFactory.add(alias=('UNet', 'unet', 'Unet'), mod=monai.networks.nets.UNet)
NetworkFactory.add(alias=('AHNet', 'Ahnet'), mod=monai.networks.nets.AHNet)

```

in the future we may have a network with a specific config
```py
# an instance of the network
NetworkFactory.add(alias=('UNet32', 'unet32', 'Unet32'), mod=monai.networks.nets.UNet, args={'channels': (32, 32, 32)})
```

PR #2204, opened 2021-05-17
2203 Add SaveClassifcation transform

### LINKED ISSUE #2203: Develop post transform to save classification results
Need to implement the saving logic as a post transform, which is not coupled with ignite.

Nic-Ma:
Fixes #2203 .

### Description
This PR added the `SaveClassification` transform to decouple with ignite handler, non-breaking change.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @wyli @rijobro ,

To follow the same API patten, I used `meta_key_postfix` arg in this transform, but actually, the `pred` data usually doesn't have the corresponding `pred_meta_dict`. So to use this transform, we must use `CopyItemD` to copy the meta data from `image_meta_dict` to `pred_meta_dict` first (Just what the unit test did).
Maybe it's better to use `meta_key` as arg in this transform? So users can set `meta_key="image_meta_dict"`.
Actually, the benefit of `meta_key_postfix` is to execute on **several keys** together, all the `keys` have the same `meta_key_postfix`. This `SaveClassificationd` transform can only work for 1 key because it will save data into a single file, multiple keys will generate a mass in the file, so no need this benefit.

We have the same problem with `SaveImageD` transform, but it can save several keys together, I am not sure whether we need to change it.

What do you think?

Thanks.

Nic-Ma:
I added both `meta_key` and `meta_key_postfix` args after offline discussion.
@wyli @rijobro , could you please help review it?

Thanks.

monai/handlers/classification_saver.py
```diff
@@ -134,6 +138,6 @@ def _finalize(self, engine: Engine) -> None:
 
         # save to CSV file only in the expected rank
         if idist.get_rank() == self.save_rank:
-            saver = CSVSaver(self.output_dir, self.filename, self.overwrite)
+            saver = CSVSaver(self.output_dir, self.filename, self.overwrite) if self.saver is None else self.saver
```
rijobro, line 141: The following would be slightly shorter
```suggestion
            saver = self.saver or CSVSaver(self.output_dir, self.filename, self.overwrite)
```

rijobro (approved):

monai/data/csv_saver.py
```diff
@@ -26,44 +25,45 @@ class CSVSaver:
...
             filename: name of the saved CSV file name.
-            overwrite: whether to overwriting existing CSV file content. If we are not overwriting,
-                then we check if the results have been previously saved, and load them to the prediction_dict.
+            overwrite: whether to overwriting existing CSV file content, if True, will clear the file before saving.
+                otherwise, will apend new content to the CSV file.
 
         """
         self.output_dir = output_dir
         self._cache_dict: OrderedDict = OrderedDict()
         if not (isinstance(filename, str) and filename[-4:] == ".csv"):
```
wyli, line 44: this filename extension constraint is restrictive, do we really need this check? perhaps a warning?

monai/handlers/classification_saver.py
Nic-Ma: sure, updated in the latest commit.
Thanks.

monai/data/csv_saver.py
Nic-Ma: sure, changed to a warning in the latest commit.
Thanks.

monai/data/csv_saver.py
```diff
@@ -26,44 +26,54 @@ class CSVSaver:
...
+                otherwise, will apend new content to the CSV file.
+            flush: whether to write the cache data to CSV file immediately when `save_batch` and clear the cache.
+                default to False.
 
         """
         self.output_dir = output_dir
         self._cache_dict: OrderedDict = OrderedDict()
         if not (isinstance(filename, str) and filename[-4:] == ".csv"):
-            raise AssertionError("filename must be a string with CSV format.")
+            warnings.warn("filename is not a string with CSV format.")
```
wyli, line 54: CSV filename is not a string ends with '.csv'

PR #2214, opened 2021-05-19
2208 Add `meta_keys` arg to the components

### LINKED ISSUE #2208: Add `meta_key` arg to enhance the `meta_key_postfix` arg
Currently, we use `meta_key_postfix` to extract meta data from dict, but sometimes it may not have corresponding meta key for the specified key, like `pred` and `pred_meta_dict`.
So we need an approach to specify `meta_key` directly.

Nic-Ma:
Fixes #2208 .

### Description
The `meta_key_postfix` arg can help to use a same `postfix` to get meta data of different keys, but sometimes we may want to use key A's meta data for key B, so also need a more explicit way to specify the `meta_keys`.

### Status
**Ready**

### Types of changes


monai/apps/deepgrow/transforms.py
```diff
@@ -702,10 +722,11 @@ def __call__(self, data):
         # update bounding box in case it was corrected by the SpatialCrop constructor
         box_start = np.array([s.start for s in cropper.slices])
         box_end = np.array([s.stop for s in cropper.slices])
-        for key in self.key_iterator(d):
+        for key, meta_key, meta_key_postfix in self.key_iterator(d, self.meta_keys, self.meta_key_postfix):
             if not np.array_equal(d[key].shape[1:], original_spatial_shape):
                 raise RuntimeError("All the image specified in keys should have same spatial shape")
-            meta_key = f"{key}_{self.meta_key_postfix}"
+            if meta_key is None:
```
ericspod, line 728: Elsewhere you have `meta_key = meta_key or f"{key}_{meta_key_postfix}"` which I think is correct to use. The `or` operator would be a problem if the left-hand side was something that overrode certain operators but using it to only check for `None` should be fine.

monai/transforms/post/dictionary.py
```diff
@@ -500,10 +508,26 @@ def __init__(
...
+            post_func,
+        ) in self.key_iterator(
+            d,
+            self.orig_keys,
+            self.meta_keys,
+            self.meta_key_postfix,
+            self.nearest_interp,
+            self.to_tensor,
+            self.device,
+            self.post_func,
```
ericspod, line 528: For legibility it might help to put the arguments to `key_iterator` into a tuple on the line above and expand it here so that the `for` statement is on one line.

ericspod (commented):
Just a few comments but otherwise looks good.

monai/transforms/post/dictionary.py
Nic-Ma: Hi @ericspod ,

Thanks for your suggestion, I tried to use:
```py
iterators = (d, self.orig_keys, self.meta_keys, self.meta_key_postfix, self.nearest_interp, self.to_tensor, self.device, self.post_func)
for (
    key, orig_key, meta_key, meta_key_postfix, nearest_interp, to_tensor, device, post_func
) in self.key_iterator(iterators):
```
The first line is still more than 120 chars, maybe I misunderstood your idea?

Thanks.

Nic-Ma: So now I just changed to below format in the latest commit:
```py
for (
    key, orig_key, meta_key, meta_key_postfix, nearest_interp, to_tensor, device, post_func
) in self.key_iterator(
    d,
    self.orig_keys,
    self.meta_keys,
    self.meta_key_postfix,
    self.nearest_interp,
    self.to_tensor,
    self.device,
    self.post_func,
):
```
What do you think?

monai/apps/deepgrow/transforms.py
Nic-Ma: Sure, updated in the latest commit.

Thanks.

Nic-Ma:
Hi @ericspod ,

Thanks very much for your review.

Thanks in advance.

monai/transforms/post/dictionary.py
ericspod: It's better but something like this might be more readable:

```python
iterators = (d, self.orig_keys, self.meta_keys, self.meta_key_postfix, self.nearest_interp, self.to_tensor, self.device, self.post_func)
for vars in self.key_iterator(*iterators):
    key, orig_key, meta_key, meta_key_postfix, nearest_interp, to_tensor, device, post_func = vars
```

Having the `for` on one line might be easier to read, up to you if you want to change it.

ericspod (approved):

monai/transforms/post/dictionary.py
Nic-Ma: I see, but the problem is that below line of your example code is more than 120 chars:
```py
iterators = (d, self.orig_keys, self.meta_keys, self.meta_key_postfix, self.nearest_interp, self.to_tensor, self.device, self.post_func)
```
So we still need to break it into lines...

Thanks.

Nic-Ma:
Hi @ericspod ,

This feature can be helpful if we want to use the meta data which was not generated from current key, for example:
If we use `SaveImageD` transform to save the model output with key `pred`, there is no meta data with key `pred_meta_dict` which has postfix `meta_dict`, instead, we should use the `image_meta_dict` to extract meta data. So we need a way to explicitly specify the key of expected meta data.

Thanks.

Issue #2213, opened 2021-05-18
Random Intensity transforms for test-time augmentation


han-liu:
Hi,

Test-time augmentation is a helpful tool to use but it seems that currently only spatial transformations are included for TTA. Intensity transformations such as Random contrast are not invertible transforms and thus cannot be used in current TTA implementation. Is there a way that I can use intensity transforms as TTA? Thanks.

han-liu:
@rijobro @tvercaut Thanks for the clarification and I think what @tvercaut proposed here is a great feature to be added to TTA.

rijobro:
@tvercaut to take the example of `GaussianNoise`, the inverse of the spatial component of that transform is identity, because obviously adding Gaussian noise doesn't modify the image spatially. Having said that, I think it would be complicated for us to therefore say that a pseudo of the inverse can be given as identity. Because at that point, we're saying that the most important component of the inverse is the spatial component. This may or may not be true, but I'm not sure we really want to make a decision either way. 

A potential solution might be to have something like:

```python
def inverse(self, x):
    x = self.spatial_inverse(x)
    x = self.intensity_inverse(x)
    # any other types of inverse
    ...
```
But I think this starts to get messy and is probably overkill for what we want. And no, we don't have different inverses depending on the network type.

PR #2217, opened 2021-05-19
inverse adjust contrast

rijobro:
Starts to address https://github.com/Project-MONAI/MONAI/issues/2213.

### Description
Add inverse for `AdjustContrastd` and `RandAdjustContrastd`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/intensity/array.py
+++ b/monai/transforms/intensity/array.py
@@ -623,7 +623,7 @@ class AdjustContrast(Transform):
     """
     Changes image intensity by gamma. Each pixel/voxel intensity is updated as::
 
-        x = ((x - min) / intensity_range) ^ gamma * intensity_range + min
+        x' = ((x - min) / intensity_range) ^ gamma * intensity_range + min
 
     Args:
         gamma: gamma value to adjust the contrast as function.
@@ -642,6 +642,16 @@ class AdjustContrast(Transform):
         img_min = img.min()
         img_range = img.max() - img_min
         return np.power(((img - img_min) / float(img_range + epsilon)), self.gamma) * img_range + img_min
+
+    @staticmethod
+    def inverse(img: Union[np.ndarray, torch.Tensor], old_min: float, old_max: float, gamma: float):
+        """Inverse method. Pass the min and max of the original image, as well as the
+        gamma that was used. Inverse is:
+
+            x = old_min + old_range * ((x' - old_min) / old_range) ** (1/gamma)
+        """
+        _range = old_max - old_min
+        return old_min + _range * ((img - old_min) / _range) ** (1 / gamma)
 
 
 class RandAdjustContrast(RandomizableTransform):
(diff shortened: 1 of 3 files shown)


wyli (approved):
thanks! there might be an implicit int->float casting in the forward, not sure whether we should consider that in the inverse, perhaps could refactor that later

Nic-Ma:
Hi @rijobro ,

I have a concern about this PR:
If we use `AdjustContrastd` in the transform chain for `image` during inference, and want to invert `pred` based on the transform_info of `image`. How to avoid inverting this transform but invert other spatial transforms?

Thanks in advance.

rijobro:
The applied transforms are stored in a list, so we could simply manipulate the list as we need. Untested example might be like this:

```python
def remove_transforms(transforms: list, to_be_removed: list):
    return [t for t in transforms if t["class"] not in to_be_removed]

transforms = remove_transforms(data["image_transforms"], ["AdjustContrastd"])
```

We could potentially formalise this using a withable or putting it in the Inverter logic:


```python
ts = data["image_transforms"]
with InvertibleTransform.skipInverses(ts, ["AdjustContrastd"]):
    ts.inverse(data)
```

Nic-Ma:
Hi @rijobro ,

Thanks for your proposal, but is there any use case to avoid only 1 of same transforms?
For example:
```py
Resized(keys="img", spatial_size=[128, 128, 128]),
RandSpatialCrop(keys="img", random_size=True),
Resized(keys="img", spatial_size=[64, 64, 64]),
```
And only skip the first `Resized`.
So in summary, I am not sure we should skip transforms by **class name** or instance **unique ID**.
@wyli @ericspod , what do you think? Or any other ideas?
We also need to think about how to enhance the `Invertd` transform for this feature request.

Thanks.

rijobro:
I think there could be lots of different reasons for wanting to do the inverse of some, but not all of the applied transforms, so there won't be any "one size fits all" solution.

I don't think we often use multiple instances of the same transform, but If the user wanted to skip the first `Resized`, then they could just do it via indexing: `d["image_transforms] = d["image_transforms][1:]`. Random transforms are appended even if `do_transform` was false (below threshold probability), meaning the length of invertible transforms will remain constant and indexing can be safely used.

Nic-Ma:
Hi @rijobro ,

Could you please help create a PR to enhance the `Invertd` transform for this issue?
Inverting is heavily used in our internal projects now, once users use `AdjustContrastd` transform, it will cause issue..

Thanks in advance.

rijobro:
Glad to hear that the inverse functionality is being used! Sure, I'll create a draft PR and we can discuss if the solution is acceptable there.

Nic-Ma:
Cool, sounds good!
Thanks.

wyli:
perhaps could have a `SpatialTransform` interface? because in many cases we want to invert the spatial ones only.

rijobro:
Similar to my suggestion here: https://github.com/Project-MONAI/MONAI/issues/2213#issuecomment-846906915.

I wonder if we're getting a little too complicated -- what about transforms that are a mix of spatial and intensity, what about invertible transforms that are neither spatial or intensity (we don't have any yet, but may in the future).

Would it be worth scheduling a meeting for our transforms? We may want to unify certain things to make them more future-proof.

wyli:
sure @rijobro let's have a Teams chat soon...

PR #1952, opened 2021-04-05
remove unused RandomizableTransform

wyli:
### Description
fixes a few usage of `RandomizableTransform`:
- a few transforms inherit `RandomizableTransform` but doesn't initialise it (addresses "`__init__` method from the base class not called PYL-W0231")
- a few transforms have `super().__init__(prob=1.0, do_transform=True)`, effectively not using the `RandomizableTransform`, therefore change to inherit `Randomizable` for clarity
- all cache-based datasets now check isinstance `Randomizable` instead of  `Randomizable` as the defining property
- partly fixes a few "Attribute defined outside `__init__` PYL-W0201"

### Status
**Ready**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [x] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [x] Quick tests passed locally by running `./runtests.sh --quick --unittests`.
- [x] In-line docstrings updated.
- [x] Documentation updated, tested `make html` command in the `docs/` folder.

PR #2219, opened 2021-05-19
array transforms inherit Transform

wyli:

minor fixes for the array transforms, follow-up of https://github.com/Project-MONAI/MONAI/pull/1952

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/apps/deepgrow/transforms.py
+++ b/monai/apps/deepgrow/transforms.py
@@ -61,7 +61,7 @@ class FindAllValidSlicesd(Transform):
         return d
 
 
-class AddInitialSeedPointd(Randomizable):
+class AddInitialSeedPointd(Randomizable, Transform):
     """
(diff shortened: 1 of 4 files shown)


Nic-Ma (approved):

rijobro:
All of these should have probable inherited from `RandomizableTransform`, as opposed to `Randomizable` and `Transform` independently. I'll create a follow up PR.

wyli:
but the concepts of `do_transform` and `prob` from `RandomizableTransform` are not needed in these transforms, I feel in heriting that API is not needed

rijobro:
Ah I see. Ok, thanks!

rijobro:
Although perhaps for the sake of continuity across all of our random transforms, it would be nice if they also had a `prob` argument. This wouldn't be backwards compatible though, so probably not worth it.

PR #2222, opened 2021-05-20
2026 Make all the data temporarily available for post transforms

### LINKED ISSUE #2026: Provide all available data in `engine.state.output` of workflows

Hi @ericspod @wyli @rijobro ,

The problem and idea is that:
currently, MONAI post transforms(independent with ignite) execute on the `engine.state.output` dict, the post transforms can't get some useful information in `engine.state.batch`, for example, the meta_dict of images or some transform information we put in the `engine.state.batch`. As @vfdev-5 and I discussed in the ignite ticket: https://github.com/pytorch/ignite/issues/1923, I think we need to shallow-copy the data of `engine.state.batch` into `engine.state.output`, then `post transforms` and `handlers` can get all the available data from only 1 dict.
The data flow can run through all the components and no need to write 2 lambda funs (`batch_transform` and `output_transform`) in handlers, we can use `keys` to get expected data directly.
I didn't find any benefit to split the data flow into `state.batch` and `state.output` in MONAI.
For the first step, I plan to change this line as aligned with @vfdev-5 :
https://github.com/Project-MONAI/MONAI/blob/master/monai/engines/trainer.py#L160
into:
```py
engine.state.output = dict(engine.state.batch)
engine.state.output.update({Keys.IMAGE: inputs, Keys.LABEL: targets})
```
Then we can use `engine.state.output` in post transforms and handlers as the overall data flow, and no need to get data from `engine.state.batch` anymore.

Do you guys have any comments?

Thanks in advance.

Nic-Ma:
Fixes #2026 .

### Description
This PR tried to temporarily combine `batch` and `output` to make all the data available for post transforms.

### Status
**Ready**

### Types of changes


monai/engines/workflow.py
```diff
@@ -162,7 +162,20 @@ def _register_post_transforms(self, posttrans: Callable):
 
         @self.on(IterationEvents.MODEL_COMPLETED)
         def run_post_transform(engine: Engine) -> None:
```
wyli, line 164: could you make it a post transform handler? I feel in that way we could support arbitrary execution order...

Nic-Ma: I see your point.
But I feel that is a big breaking change for current MONAI workflow, it will remove the `post_transform` arg of MONAI engines, need to update many tutorials or tests, and break all the NVIDIA Clara 4.0 products which are built based on current MONAI workflows API.
Maybe we can discuss within MONAI & Clara teams first then make a separate PR?
This PR is just an enhancement, will not break anything so far.
What do you think?

Thanks.

wyli: I think `post_transform` is a wrong terminology here anyway... it should be postprocessing

wyli: if it's a handler, we can attach it to engine by default so that it's non-breaking, what do you think?

Nic-Ma: Hi @wyli ,

Or maybe we can keep current post transform logic and develop a new handler called `PostTransform` or `PostProcessing`, users can still use current post transforms in MONAI engines directly, or set it to `None` and use the new `PostTransform` handler. Or even use both.
What do you think?

Thanks.

wyli: sounds good to me, thanks!

Nic-Ma: Thanks, I created a separate ticket to track that feature: https://github.com/Project-MONAI/MONAI/issues/2223, will make a PR for it when our handler discussion with ignite team is ready.
And we may also add `MetricHandler` for current MONAI metrics, because ignite metrics are handlers in theory, but it's more complicated to achieve and based on ignite `Metric` base class. We can discuss it later.

About the `post transform` naming, some other DL papers or docs also use it:
https://www.sciencedirect.com/science/article/abs/pii/S1568494620309261
https://www.khronos.org/opengl/wiki/Post_Transform_Cache
Maybe @ericspod @rijobro can help share more comments, I am not sure about it.

Thanks.

wyli: in that context, probably post-transform means some operations applied after the transforms. in our case, it should be some operations applied after the model forward, in our case it is "post_model_forward_transform"

monai/engines/workflow.py
```diff
@@ -162,7 +162,20 @@ def _register_post_transforms(self, posttrans: Callable):
...
-            engine.state.output = apply_transform(posttrans, engine.state.output)
+            if isinstance(engine.state.batch, dict) and isinstance(engine.state.output, dict):
```
wyli, line 165: this could be a utility as we discussed

Nic-Ma: Sure, updated in the latest commit.
Thanks.

wyli (approved):
thanks!

PR #2225, opened 2021-05-20
2224 decouple handler from ignite

### LINKED ISSUE #2224: Make handlers to decouple with ignite APIs
According to our design principle, almost all the MONAI components can be easily integrated into other workflows or PyTorch programs except for handlers.
As handlers contain many important features, users who are using other workflows also want to integrate them.
After analysis with ignite team: https://github.com/pytorch/ignite/issues/2000, actually it's possible to extract ignite related logic from handlers as utils and make most of the handlers to be independent from ignite, except for some handlers that are based on ignite handlers or ignite distributed data parallel.
So I plan to implement based on the discussion and start from `inference` related handlers.

Nic-Ma:
Fixes #2224 .

### Description
This PR implemented the proposal to decouple handlers from ignite.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @wyli and @vfdev-5 ,

In this proposal, users can register new events to the ignite engine and develop handlers for them, but how to fire some customized event in a handler?
Sample code in the previous ignite stype handler:
```py
class TestHandler:
    def attach(self, engine):
        engine.add_event_handler("foo_event", self.test1)
        engine.add_event_handler("gar_event", self.test2)
        engine.add_event_handler(Events.EPOCH_COMPLETED, self)

    
    def __call__(self, engine):
        engine.fire_event("foo_event")
        engine.fire_event("gar_event")

    def test1(self, engine):
        pass
    
    def test2(self, engine):
        pass

```

Thanks.

Nic-Ma:
Hi @vfdev-5 ,

Thanks for your support, you are right about the proposal.
But when I double check our existing handler features, I found that this proposal can't support the use case that handler **fires other event in it** anymore, just like your example: https://github.com/pytorch/ignite/blob/master/ignite/engine/engine.py#L186
So I am not sure whether we have a solution for this case or we don't need to support it anymore.

Thanks.

Nic-Ma:
Hi @wyli ,

Do you have any comments or ideas about this feature? Actually, I never see a user really fire customized events in his customized handler. Maybe we don't need to support it anymore?

Thanks.

Nic-Ma:
Hi @wyli ,

Do you know it always imports the `IterationEvents` even no program uses it?

Thanks.

Nic-Ma:
Hi @wyli @ericspod @rijobro ,

After offline discussion, as we never see any use case that fires a new event in a handler, let's drop this feature.
So this PR is ready for review now.
More details about the design discussion are available at ignite ticket: https://github.com/pytorch/ignite/issues/2000.

Thanks.

wyli (changes requested):
thanks, I think we should discuss about the details more

ericspod:
Hi @Nic-Ma I think a lot of what we were doing here is in MONAI now or we've changed ideas, shall we close this one?

Nic-Ma:
> Hi @Nic-Ma I think a lot of what we were doing here is in MONAI now or we've changed ideas, shall we close this one?

Hi @ericspod ,

After some previous discussions, I think we don't have clear requirements for this feature request and haven't got a perfect solution so far, maybe let's just keep it open?

Thanks.

ericspod:
> Hi @ericspod ,
> 
> After some previous discussions, I think we don't have clear requirements for this feature request and haven't got a perfect solution so far, maybe let's just keep it open?
> 
> Thanks.

Sure we'll keep it open then and come back to it later.

PR #2227, opened 2021-05-20
1871 thread-safety for the cache-based datasets

### LINKED ISSUE #1871: cachedataset useability
this feature request is a follow up of https://github.com/Project-MONAI/MONAI/pull/1836. ideally multiple preprocessing threads would 1. read from a shared cache (storing non random transform outcomes) and 2. have independent transform instances for each of them.  2 is not currently implemented as a result the transform instance properties might be modified by multiple threads simultaneously and causing issues #1030 #1836.

the caching logic could be improved so that it is easier to make customised transforms for the cache based datasets.

wyli:
Fixes #1871

### Description
this PR addresses cache-based dataset multi-thread preloading issues by:
- adding a `monai.transforms.ThreadUnsafe` API, the unsafe transforms will be deepcopied in the cache computation.
- revise all transforms to be either thread-safe or inheriting `ThreadUnsafe`
- updated persistent dataset although it's not using multi-threading, mainly to keep the behaviour consistent with the cache-based ones
 
all the randomisables are not thread-safe, but we only use them with multi-process dataloader, there's no multi-thread issue.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -668,10 +668,10 @@ class SmartCacheDataset(Randomizable, CacheDataset):
             self._cache = self._fill_cache()
         if self.cache_num >= len(data):
             warnings.warn(
-                "cache_num is greater or equal than dataset length, " "fall back to regular monai.data.CacheDataset."
+                "cache_num is greater or equal than dataset length, fall back to regular monai.data.CacheDataset."
             )
         if replace_rate <= 0:
-            raise ValueError("replace_rate must be greater than 0, otherwise, " "please use monai.data.CacheDataset.")
+            raise ValueError("replace_rate must be greater than 0, otherwise, please use monai.data.CacheDataset.")
 
         self.num_replace_workers: Optional[int] = num_replace_workers
         if self.num_replace_workers is not None:
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -30,7 +30,7 @@ from torch.utils.data import Dataset as _TorchDataset
 from torch.utils.data import Subset
 
 from monai.data.utils import first, pickle_hashing
-from monai.transforms import Compose, Randomizable, Transform, apply_transform
+from monai.transforms import Compose, Randomizable, ThreadUnsafe, Transform, apply_transform
 from monai.utils import MAX_SEED, get_seed, min_version, optional_import
 
 if TYPE_CHECKING:
@@ -186,7 +186,8 @@ class PersistentDataset(Dataset):
             # execute all the deterministic transforms
             if isinstance(_transform, Randomizable) or not isinstance(_transform, Transform):
                 break
-            item_transformed = apply_transform(_transform, item_transformed)
+            _xform = deepcopy(_transform) if isinstance(_transform, ThreadUnsafe) else _transform
+            item_transformed = apply_transform(_xform, item_transformed)
         return item_transformed
 
     def _post_transform(self, item_transformed):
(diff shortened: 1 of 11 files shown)


monai/transforms/croppad/array.py
```diff
@@ -158,9 +158,9 @@ def __call__(self, img: np.ndarray, mode: Optional[Union[NumpyPadMode, str]] = N
         spatial_border = tuple(max(0, b) for b in spatial_border)
 
         if len(spatial_border) == 1:
-            data_pad_width = [(spatial_border[0], spatial_border[0]) for _ in range(len(spatial_shape))]
+            data_pad_width = [(spatial_border[0], spatial_border[0]) for _ in spatial_shape]
         elif len(spatial_border) == len(spatial_shape):
-            data_pad_width = [(spatial_border[i], spatial_border[i]) for i in range(len(spatial_shape))]
+            data_pad_width = [(sp, sp) for sp, _ in zip(spatial_border, spatial_shape)]
```
ericspod, line 163: `data_pad_width = [(sp,) * 2 for sp in spatial_border[:len(spatial_shape)]]` 

Is that clearer or actually worse?

ericspod (approved):

Nic-Ma:
Thanks for the enhancement for CacheDataset, it's heavily used in NVIDIA Clara project, I will try to review this PR as soon as possible.

monai/transforms/transform.py
```diff
@@ -71,14 +73,29 @@ def _log_stats(data, prefix: Optional[str] = "Data"):
...
+    cautiously in a multi-thread context.
+
+    This type is typically used by :py:class:`monai.data.CacheDataset` and
+    its extensions, where the transform cache is built with multiple threads.
+    """
+
+    pass
+
+
+class Randomizable(ABC, ThreadUnsafe):
```
Nic-Ma, line 89: May I know why Randomizable transforms are thread-unsafe?
I think "writing class member variables at runtime" causes the `thread-unsafe` issue, right?

Thanks.

monai/data/dataset.py
```diff
@@ -180,13 +180,13 @@ def _pre_transform(self, item_transformed):
...
-        if not isinstance(self.transform, Compose):
-            raise ValueError("transform must be an instance of monai.transforms.Compose.")
-        for _transform in self.transform.transforms:
+        for _transform in self.transform.transforms:  # type:ignore
             # execute all the deterministic transforms
             if isinstance(_transform, Randomizable) or not isinstance(_transform, Transform):
                 break
-            item_transformed = apply_transform(_transform, item_transformed)
+            # this is to be consistent with CacheDataset even though it's not in a multi-thread situation.
+            _xform = deepcopy(_transform) if isinstance(_transform, ThreadUnsafe) else _transform
```
Nic-Ma, line 188: I think `deepcopy` a class instance maybe a little bit dangerous as some kinds of class properties may not support pickle? For example:
```py
import numpy as np

class Transform:
    def __init__(self):
        self.np = np

a = A()
a_copy = copy.deepcopy(a)
```
Got error: `can’t pickle module objects error`

So maybe we need a test method to guarantee all the exiting and future `thread-unsafe` transforms can support `deepcopy`?

Thanks.

Nic-Ma (commented):
Thanks for the enhancement, put several comments inline.
Others look good to me.

Thanks.

monai/transforms/transform.py
wyli: I think calling `self.R` would implicitly change its state here, so it's not thread-safe.

monai/data/dataset.py
wyli: for now we only have one such threadunsafe transform... 
https://github.com/Project-MONAI/MONAI/blob/2fb4b2170506da73effd82f4acab5fedb30722e5/monai/transforms/spatial/array.py#L491

let me think about this, and perhaps will create a ticket about it. Thanks!

monai/transforms/transform.py
Nic-Ma: Could you please help enhance the doc-string to describe this `self.R` point?

Thanks.

monai/data/dataset.py
Nic-Ma: I know we don't have many thread-unsafe transforms so far, but need a test or method to make sure future thread-unsafe transforms can support `deepcopy`.

Thanks.

Issue #1901, opened 2021-03-30
Invert patch-based transforms


wyli:
invert the patch based transforms such as 
- RandSpatialCropSamplesd
 - RandWeightedCropd
 - RandCropByPosNegLabeld

(these transforms return lists)

followup of #1515

wyli:
feature implemented as inverting each item of the list of patch samples respectively

PR #2230, opened 2021-05-21
Invert patch-based transforms

rijobro:
Adds inverse for:
- RandSpatialCropSamplesd
- RandWeightedCropd
- RandCropByPosNegLabeld

Fixes https://github.com/Project-MONAI/MONAI/issues/1901. 
cc @diazandr3s.

### Status
**Ready**

### Types of changes


monai/transforms/croppad/dictionary.py
```diff
@@ -15,6 +15,7 @@
 Class names are ended with 'd' to denote dictionary-based transforms.
 """
 
+from contextlib import nullcontext
```
wyli, line 18: this module is not in python 3.6...

rijobro: Thanks, I've modified to not use it.

wyli (approved):
thanks!

rijobro:
@Nic-Ma do you think this will this cause any problems with your current workflows?

Nic-Ma (approved):
Looks good to me.
Could you please help add a comment or doc-string to highlight that: we can't automatically `batch-invert` the entire transform chain if contains patch-based transform, because the output will be a list?

Thanks.

PR #2237, opened 2021-05-24
Save image print log

rijobro:
### Description
Allow disabling verbose print when saving images.

### Status
**Ready**

### Types of changes


wyli (approved):
thanks!

PR #2242, opened 2021-05-24
enhance downloading APIs

wyli:

- enhance the download APIs with better default path values
- remove duplicated check_hash in `download_and_extract`

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/apps/utils.py
+++ b/monai/apps/utils.py
@@ -181,7 +181,7 @@ def extractall(
         hash_val: expected hash value to validate the compressed file.
             if None, skip hash validation.
         hash_type: 'md5' or 'sha1', defaults to 'md5'.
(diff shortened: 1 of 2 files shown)


monai/apps/utils.py
```diff
@@ -37,6 +39,12 @@
 ]
 
 
+def _basename(p):
+    """get the last part of the path (removing the trailing slash if it exists)"""
+    sep = os.path.sep + (os.path.altsep or "") + "/ "
```
Nic-Ma, line 44: Please double confirm whether it works fine for windows?
Thanks.

Nic-Ma (approved):
Looks good to me except for minor comments.

Thanks.

monai/apps/utils.py
wyli: sure, and the '/' is mainly for getting filename from the URLs

Issue #561, opened 2020-06-13
Additional Metric Interface to compute quantities over dataset


mibaumgartner:
**Is your feature request related to a problem? Please describe.**
Some metrics need to be computed over the whole dataset. To name a few metrics:

Classification: ROC/AUC
Segmentation: Dice (when training on patches the dice can not be computed for background patches -> need to compute intermediate values and save them)
Detection: mAP, FROC

For most of these examples it is not feasible to store the target and the prediction over many iterations (e.g. storing all the ground truth and predicted segmentations), so the computation needs to be splitted in multiple parts. E.g. storing the tp, fp, ... per sample for the later computation (similar things can be done for the detection metrics).

**Describe the solution you'd like**
Create a functional interface and a class interface for the metrics (probably related to https://github.com/Project-MONAI/MONAI/issues/497).

```python
class BaseMetric:
    ...

    def __call__(self, ...):
        # call functional interface to compute metric
```

```python
class PartialMetric(BaseMetric):
    ...

    def add_sample(self, ...):
        # compute intermediate results to store for later computation

    def compute(self, ....):
         # could implement something additional based on the intermediate values
        # or directly compute the target metric
```

**Describe alternatives you've considered**
//

**Additional context**
This Issue is probably related to/to be tackled after https://github.com/Project-MONAI/MONAI/issues/497 .

After discussing the details I would like to implement this feature (also including some of the metrics mentioned above, which we could discuss separately )  :)

Nic-Ma:
Hi @mibaumgartner ,

Thanks for your suggestion, please go ahead and submit a PR for it if you like.
For #497 , you can also implement the requirements in your PR, I think it's a minor modification.
This is a slightly big change, we can work together if you need some help, for example: you can just implement metrics for the `class` API and keep the original functional API. Then I can help to update all the usages of metrics and all the examples, integration tests, etc.

Thanks.

mibaumgartner:
Hi @Nic-Ma ,
thank you for your quick response, I'll work on the PR today :)

PR #565, opened 2020-06-14
[WIP] 561 dataset/partial metrics

mibaumgartner:
Fixes #561 #497  .

### Description
This pull request is intended to provide a class based API for metric evaluation and add additional functionality to some metrics which can be evaluated on previously stored intermediate values.

### Status
Work in progress. The current status is intended as an early draft of the targeted functionality (so some docstrings are not perfect and tests are missing). Would be great to get feedback for the following points:
- General structure and design of the changed hierarchy inside the metrics package
- Design of the exemplary ROCAUC class API (even though the full advantage of partial metrics can not be seen here 🙈 ) 
    - It is not always possible to call the functional method inside the `evaluate` function because most functions will store an intermediate representation of the data (e.g. for the Dice metric would store tp, fp .. which can not be passed to the functional interface). How should this be handled to reduce duplicate code?
- Which metrics to include here (I would propose ROCAUC and Dice)

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or new feature that would cause existing functionality to change)
- [ ] New tests added to cover the changes
- [ ] Docstrings/Documentation updated

Issue #497, opened 2020-06-06
Add class API for metrics


Nic-Ma:
**Is your feature request related to a problem? Please describe.**
If we have class API for metrics, it makes easier to set all training components before training loop.
Get 3 feedback from Andriy:
1. Add class API for the metrics functions.
2. Change `torch.tensor(float(“nan”)).to(y_o.float())` to
`torch.tensor(float("nan"), device=y_o.device)`
3. Add `dim=` to several PyTorch calls.

Nic-Ma:
Mark: @myron is working on a PR to resolve this issue.

myron:
So, here is my attempt 
https://github.com/Project-MONAI/MONAI/pull/574

plz check

wyli:
fixed by https://github.com/Project-MONAI/MONAI/pull/574

PR #2244, opened 2021-05-25
[WIP] 2220 Decollate batch into list of tensors after model forward

Nic-Ma:
### Description
This PR added support to handle list of Tensor in post transforms.

### Status
**Work in progress**

### Types of changes


Nic-Ma:
Hi @ericspod @wyli @rijobro ,

In this drat PR, I tried to enhance the `Activationd` transform to support a list of Tensor.
If you have no concern about this solution, I will try to enhance all the other post transforms and utility transforms.
Or if you have any better idea, please feel free to share some sample code here.

Thanks in advance.

ericspod:
It looks fine to me. Do we have a particular use case for this?

Nic-Ma:
> It looks fine to me. Do we have a particular use case for this?

Hi @ericspod ,

Thanks for your review.
When we try to invert the transforms, the results usually get a list of channel-first Tensor with different size. So we need to support it in the following transforms. For example:
```py
post_transforms = Compose([
    Activations(keys="pred"),
    CopyItemsd(keys="pred", times=1, names="inverted_pred"),
    Inverted(keys="inverted_pred", nearest_interp=False),
    AsDiscreted(keys=["pred", "inverted_pred"]),
    SaveImaged(keys=["pred", "inverted_pred"]),
])
```

Thanks.

monai/transforms/post/array.py
```diff
@@ -83,6 +84,9 @@ def __call__(
             ValueError: When ``self.other=None`` and ``other=None``. Incompatible values.
 
         """
+        if isinstance(img, (tuple, list)):
+            return [self(img=i.unsqueeze(0), sigmoid=sigmoid, softmax=softmax, other=other).squeeze(0) for i in img]
```
rijobro, line 88: This looks good. Can we think of any way to abstract this out to a base level to reduce repetition when we implement it in other post-processing transforms?

ericspod:
@rijobro mentioned abstracting the concept out rather than inserting it into every transform everywhere. Earlier we had discussed transforms representing diverging paths in the transform sequence, for example something like [OneOf](https://github.com/albumentations-team/albumentations/blob/a3f40853b33de19f8c64f5d977190a6cdd0c786c/albumentations/core/composition.py#L244) where  one of the transforms passed to the object as an argument is randomly chosen to be applied. Perhaps a wrapping transform that takes a list of tensors as you want and applies each in the list to the sequence of transforms it was given then either returns the list or collates the list. This would look like:

```python
post_transforms = Compose([
    ApplyToSeqd(keys="pred", collate=False, [
        Activations(keys="pred"),
        CopyItemsd(keys="pred", times=1, names="inverted_pred"),
        Inverted(keys="inverted_pred", nearest_interp=False),
        AsDiscreted(keys=["pred", "inverted_pred"]),
        SaveImaged(keys=["pred", "inverted_pred"]),
    ])  # output is dictionary with lists of tensors for values
])
```

monai/transforms/post/array.py
Nic-Ma: If we can think of a way to abstract this feature to base class, it would be better, but I didn't get so far...

Thanks.

Nic-Ma:
Hi @rijobro and @ericspod ,

So in summary, now we have 4 draft ideas to solve this problem:
1. Add recursive call in every post transform to handle a list of tensor (Nic)
2. Extract (1) to the base class, so don't need to change every transform, haven't got how to implement it (Richard)
3. Decollate a batch tensor to a list of tensor, then all the transforms and metrics handle every tensor in a for-loop (Wenqi)
4. Make a wrapper transform to handle the list of tensor, also can support other cases like `OneOf` (Eric)

Thanks.

rijobro:
Good summary, let's discuss tomorrow.

Nic-Ma:
Thanks for this great online discussion.
We all agreed to go with `decollate` way.
It's a big change to many modules, I will try to implement it soon.

Thanks.

wyli:
> Thanks for this great online discussion.
> We all agreed to go with `decollate` way.
> It's a big change to many modules, I will try to implement it soon.
> 
> Thanks.

thanks, if needed please create a branch in the codebase dedicated to this feature, it'll make it easier to manage the release of this feature

Nic-Ma:
Hi @wyli ,

Thanks for your suggestions.
I already split the task into steps.
If can't clearly submit PR for steps, I will create a branch.

Thanks.

Nic-Ma:
I already updated all the necessary post / IO / utility transforms to be `channel-first`, will try to update MONAI engines and handlers in the next steps.

Thanks.

wyli:
> I already updated all the necessary post / IO / utility transforms to be `channel-first`, will try to update MONAI engines and handlers in the next steps.
> 
> Thanks.

thanks, I think we'll need full integration tests before merging this feature... these pre-merge tests won't be reliable enough for this PR.

Nic-Ma:
Hi @wyli ,

I will trigger integration tests when all the things are ready.

Thanks

Nic-Ma:
Hi @wyli @ericspod @rijobro ,

I am adjusting the metrics interface, after decollating, there are several options:
1. Input a `list of channel-first Tensor` for `y_pred` and `y`.
2. Input only 1 `channel-first Tesnor` for `y_pred` and `y`, do for-loop for 1 batch outside metrics, just same as transforms. But some metrics may need to compute on a batch data or the whole data together, like `AUC`, etc.

I prefer to unify the metrics API to below signature, which was discussed in PR https://github.com/Project-MONAI/MONAI/pull/565:
```py
class Metric:
    def update(self, y_pred: Union[torch.Tensor, Sequence[torch.Tensor]], y: Union[torch.Tensor, Sequence[torch.Tensor]]):
        # call for a decollated batch data usually when iteration completed
        # compatible with `batch-first tensor` and `list of channnel-first tensor`
        for pred_, y_ in zip(y_pred, y):
            self._add_sample(pred_, y_)
    
    def _add_sample(self, y_pred: torch.Tensor, y: torch.Tensor):
        # execute logic for 1 sample of the batch
        pass

    def compute(self):
        # call for final computation usually when epoch completed
        pass
```
What do you guys think?

Thanks.

Nic-Ma:
After a quick online discussion, I would prepare a separate PR to unify the basic metrics API.
It should support both `batch-first Tensor` and `list of channel-first Tensor`.
And will close ticket #561 #497 .

Thanks.

Nic-Ma:
As this is a big feature update, I created a branch feature/2220-decollate in Project-MONAI repo.
I will merge this PR into that branch and create a new PR to dev branch.

Thanks.

PR #2246, opened 2021-05-25
RandFlip arbitrary combination of axes

### LINKED ISSUE #2245: RandFlip operates on all axes or none
For data augmentation, you might want to flip your images. You might want to flip it around any of the dimensions, either all at the same time or with some combination. The current code, however, only allows for flipping across all or none of the dimensions. 

The current solution is therefore to chain multiple transforms together, but I feel a single transform should be able to give this functionality. 

Current workaround:

```python
RandFlipd(["image", "label"], prob=0.5, spatial_axis=0),
RandFlipd(["image", "label"], prob=0.5, spatial_axis=1),
```

```python
def randomize(data):
    if isinstance(self.spatial_axis, int):
        self.sampled_spatial_axis = self.spatial_axis
    else:
        sample = np.random.randint(2, size=len(self.sampled_spatial_axis))
        self.sampled_spatial_axis = np.array(self.sampled_spatial_axis)[sample].tolist()
```

rijobro:
Fixes #2245.

**Whilst the API remains the same, the output will be changed, so this is a breaking change.**

### Description
Randomly flip arbitrary combination of axes. 

Previously:

```python
RandFlipd(["image", "label"], prob=0.5, spatial_axis=0),
RandFlipd(["image", "label"], prob=0.5, spatial_axis=1),
```

Now:

```python
RandFlipd(["image", "label"], prob=0.5, spatial_axis=[0,1]),
```

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -773,7 +773,17 @@ class RandFlip(RandomizableTransform):
 
     def __init__(self, prob: float = 0.1, spatial_axis: Optional[Union[Sequence[int], int]] = None) -> None:
         RandomizableTransform.__init__(self, prob)
-        self.flipper = Flip(spatial_axis=spatial_axis)
+        self.spatial_axis = spatial_axis
+        self.sampled_spatial_axis: Optional[Union[Sequence[int], int]] = None
+
+    def randomize(self, _) -> None:
+        super().randomize(None)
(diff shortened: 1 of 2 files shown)


wyli:
I feel we can have a flag to preserver the previous "all or none" behaviour when the user provides multiple axis indices.

PR #2252, opened 2021-05-26
Invert some keys

rijobro:
Invert some but not all transforms.

To skip the inverse of `RandAdjustContrastd`, for example, this can be done with:

```python
d_fwd = transforms(d)
d_fwd_inv = transforms.inverse_with_omissions(d, "RandAdjustContrastd")
```

Skipped transforms will be added to a list in the dictionary e.g., `image_skipped_inverses`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/compose.py
+++ b/monai/transforms/compose.py
@@ -13,6 +13,7 @@ A collection of generic interfaces for MONAI transforms.
 """
 
 import warnings
+from copy import deepcopy
 from typing import Any, Callable, Optional, Sequence, Union
 
 import numpy as np
@@ -28,6 +29,7 @@ from monai.transforms.transform import (  # noqa: F401
     apply_transform,
 )
 from monai.utils import MAX_SEED, ensure_tuple, get_seed
+from monai.utils.enums import InverseKeys
 
 __all__ = ["Compose"]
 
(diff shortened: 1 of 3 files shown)


Nic-Ma:
Hi @rijobro ,

Thanks for the quick PR. 
I have several comments here:
1. Maybe we should support more complicated cases, not only skip.  So I would suggest to enhance the `inverse()` API directly instead of bringing new APIs. And as we are still not clear about all the use cases, I think we can provide the basic feature to just specify a list of transforms to invert by class name or index. If no list, invert all. 
2. Could you please also help enhance the `Invertd` post transform? Almost all the users invert by this API. 
3. As `Invertd` is heavily used now, the AdjustContrast transform will cause invert ing error, we must fix this blocking issue ASAP. 

Thanks.

wyli:
shall we revert https://github.com/Project-MONAI/MONAI/pull/2217 temporarily? we could include that after properly merging this PR.

rijobro:
Sure, we can revert #2217, although I don't fully understand why we can't just temporarily manipulate `image_transforms` and `label_transforms`. If you remove `AdjustContrastd` from these lists, then they won't be inverted.

rijobro:
@Nic-Ma I thought about modifying the `inverse` method, but the problem is that the interface should be the same for all `InvertibleTransform` classes, including `Compose` and all the others. The idea of skipping, however, only really makes sense for the `Compose` class, which is why I created a new method for it. Happy to discuss this tomorrow, though.

rijobro:
Also @Nic-Ma I can't really modify `Invertd` until we make a decision here, since `Invertd` calls `BatchInverseTransform`, which eventually calls the `inverse` method. So first we need to agree on the `inverse` API before we can do `Invertd`.

Nic-Ma:
Sure, let's discuss it online. 

Thanks.

PR #2261, opened 2021-05-26
Bringing CRF implementation inline with literature

charliebudd:
In the current version the compatability transform is extended to take into account neighbouring voxels. While this produces smoother results, it may well over smooth. While possibly worth some investigation it should be implemented as in the paper for now. This PR brings it inline with the literature. 

Signed-off-by: chaliebudd <charles.budd@kcl.ac.uk>

### Status
**Ready**

### Types of changes


monai/networks/blocks/crf.py
```diff
@@ -112,12 +96,14 @@ def forward(self, input_tensor: torch.Tensor, reference_tensor: torch.Tensor):
             # combining filter outputs
             combined_output = self.bilateral_weight * bliateral_output + self.gaussian_weight * gaussian_output
 
-            # compatibility convolution
-            combined_output = pad(combined_output, 2 * spatial_dim * [padding], mode="replicate")
-            compatibility_update = conv(combined_output, compatibility_kernel)
+            # optionally running a compatability transform
+            if self.compatability_matrix != None:
```
rijobro, line 100:
```suggestion
            if self.compatability_matrix is not None:
```

rijobro (approved):

monai/networks/blocks/crf.py
charliebudd: Good point, thanks

charliebudd:
Thoughts on how to handle this? 
https://github.com/Project-MONAI/MONAI/pull/2261/checks?check_run_id=2684150676#step:7:338

Issue #2231, opened 2021-05-21
Default to using Pytorch APIs for `monai.transforms` (ETA TBD)


wyli:
**Is your feature request related to a problem? Please describe.**
Based on our discussions, this ticket proposes to implement the MONAI transforms preferably using the Pytorch APIs instead of the current mainly numpy-based solutions.

wyli:
as an initial step, the transform should support both tensor and ndarray types, and return value should have the same type as input.

rijobro:
I had a thought about this. `LoadImage` would still return numpy and therefore all transforms that follow would be doing numpy<->torch conversions. So, to make the most of this suggested plan of returning the image in the same type as the input, we would want to move the `ToTensor` as early as possible, e.g.:

```python
t = Compose([
    LoadImage,
    ToTensor,
    AllOtherTransforms,
    ...
])
```

wyli:
because this is a significant feature, how about we create a feature branch for it in the codebase? when it's entirely ready we merge it into dev. otherwise we keep the feature branch up-to-date with the dev branch. all of us can collaborate on the feature branch.

rijobro:
If I do this as a branch on the Project-MONAI fork, will I have to get an approved code review every time I want to commit? If so, I think I'll just do it as a branch on my fork.

wyli:
that's also fine, just two thoughts: 
- from your fork , you wouldn't be able to trigger the full self-hosted ci pipelines (I feel we may need a full test before merging)
- we (at least @Nic-Ma and I) could help build this feature together, submitting PRs to a MONAI branch is easier than submitting them to your fork...

rijobro:
Ok. Could we create a branch that has CI/CD testing, but doesn't require code reviews? Potentially also doesn't require the tests to pass before pushing to that particular branch (just to speed things up, we'd have to be careful not to break things of course).

rijobro:
Great. I can't create branches on the main repository, could you take care of that?

wyli:
sure, created `feature/2231-transforms` (https://github.com/Project-MONAI/MONAI/tree/feature/2231-transforms) I'll push update ci/cd config to include this `feature/*` pattern

Spenhouet:
This is really a necessary change. I just wanted to create a discussion on this since I find my self constantly switching with `ToTensord` and `ToNumpyd`.

> Alternatively, we modify LoadImage to return torch.Tensor, but this would be a breaking change.

[...]

PR #2263, opened 2021-05-27
torch default for Flip

rijobro:
### Description
Starts to address: https://github.com/Project-MONAI/MONAI/issues/2231.

To base class convert to/from numpy as necessary, and then convert back to the original at the end. Use torch for the main functionality.

This is an example case for the rest of the transforms

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/transform.py
+++ b/monai/transforms/transform.py
@@ -14,7 +14,7 @@ A collection of generic interfaces for MONAI transforms.
 
 import logging
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Dict, Generator, Hashable, Iterable, List, Optional, Tuple
+from typing import Any, Callable, Dict, Generator, Hashable, Iterable, List, Optional, Tuple, Union
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


monai/transforms/transform.py
```diff
@@ -197,6 +205,32 @@ def __call__(self, data: Any):
...
+    def pre_conv_data(
+        self, data: TransformTypes.Images, required_type: type = torch.Tensor
+    ) -> Tuple[TransformTypes.Images, type]:
+        """Convert to torch/numpy, as required. Also return the original state so that after the transform,
+        the data can be reverted to its original type.
+        """
+        orig_type = type(data)
+        assert orig_type in (torch.Tensor, np.ndarray)
+
+        if orig_type is np.ndarray and required_type is torch.Tensor:
```
ericspod, line 217: We can put this conversion block into a separate utility function somewhere:

```python
def convert_data_type(data, current_type, required_type):
    if current_type is np.ndarray and required_type is torch.Tensor:
        return torch.Tensor(data)
    elif current_type is torch.Tensor and required_type is np.ndarray:
        return data.detach().cpu().numpy()

    return data
```

PR #2291, opened 2021-06-01
561 497 Provide metrics base APIs

### LINKED ISSUE #561: Additional Metric Interface to compute quantities over dataset

Nic-Ma:
### Description
This PR added base class for the metrics.
And add support for `list of channel-first Tensor`.
It's a non-breaking change.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @vfdev-5 ,

This PR added the metric base class and support a list of tensors. Tried to align with your new metric API design.

Thanks.

monai/config/type_definitions.py
```diff
@@ -67,3 +68,10 @@
 # Generic type which can represent either a numpy.ndarray or a torch.Tensor
 # Unlike Union can create a dependence between parameter(s) / return(s)
 NdarrayTensor = TypeVar("NdarrayTensor", np.ndarray, torch.Tensor)
+
+
+TensorList = Union[torch.Tensor, Sequence[torch.Tensor]]
```
vfdev-5, line 73: nit: seeing `TensorList` I think I would not expect it to be a Tensor itself but only a list of tensor.
However, I agree that there can be difficult to find an appropriate name.

monai/metrics/metric.py
```diff
@@ -0,0 +1,75 @@
...
+
+from abc import ABC, abstractmethod
+from typing import Any, Optional
+
+import torch
+
+from monai.config import TensorList
+
+
+class Metric(ABC):
```
vfdev-5, line 20: @Nic-Ma seems like you already support list of tensors here. Maybe, I'm missing the context here. Could you please explain how this will be working with ignite metrics ? Thanks

monai/config/type_definitions.py
Nic-Ma: Maybe use `TensorOrList`?

monai/metrics/metric.py
Nic-Ma: Actually, we have 2 levels of metrics in MONAI:
1. PyTorch metrics, this `Metric` base class is for these metrics.
2. ignite metrics, we have many ignite metrics under the handlers/ folder, they are wrapper of 1, but inherit ignite Metric.

This PR added support for `list of Tensor` in all the MONAI metrics, but if your ignite metrics can't support it, we can't use ignite metrics directly.

Thanks.

Nic-Ma:
Hi @wyli @rijobro @ericspod ,

This PR is ready for review now.

Thanks.

monai/config/type_definitions.py
Nic-Ma: I updated in the latest commit.
Thanks.

monai/metrics/hausdorff_distance.py
```diff
@@ -18,18 +18,20 @@
...
 __all__ = ["HausdorffDistanceMetric", "compute_hausdorff_distance", "compute_percent_hausdorff_distance"]
 
 
-class HausdorffDistanceMetric:
+class HausdorffDistanceMetric(Metric):
     """
     Compute Hausdorff Distance between two tensors. It can support both multi-classes and multi-labels tasks.
     It supports both directed and non-directed Hausdorff distance calculation. In addition, specify the `percentile`
     parameter can get the percentile of the distance.
     Input `y_pred` (BNHW[D] where N is number of classes) is compared with ground truth `y` (BNHW[D]).
```
yiheng-wang-nv, line 31: Input `y_pred` is compared with ground truth `y`.

monai/metrics/hausdorff_distance.py
```diff
@@ -18,18 +18,20 @@
...
     `y_preds` is expected to have binarized predictions and `y` should be in one-hot format.
     You can use suitable transforms in ``monai.transforms.post`` first to achieve binarized values.
-
+    `y_preds` and `y` can also be a list of Tensor with shape: [CHW[D]].
```
yiheng-wang-nv, line 34: `y_pred` and `y` can be a list of `channel-first` Tensor (CHW[D]) or a `batch-first` Tensor (BCHW[D]).

monai/metrics/meandice.py
```diff
@@ -10,15 +10,17 @@
...
 from monai.utils import MetricReduction
 
+from .metric import Metric
 
-class DiceMetric:
+
+class DiceMetric(Metric):
     """
     Compute average Dice loss between two tensors. It can support both multi-classes and multi-labels tasks.
     Input `y_pred` (BNHW[D] where N is number of classes) is compared with ground truth `y` (BNHW[D]).
```
yiheng-wang-nv, line 26: same comments in `monai/metrics/hausdorff_distance.py`

monai/metrics/metric.py
```diff
@@ -0,0 +1,84 @@
...
+            if y is not None:
+                ret_ = [self._apply(p_.detach().unsqueeze(0), y_.detach().unsqueeze(0)) for p_, y_ in zip(y_pred, y)]
+            else:
+                ret_ = [self._apply(p_.detach().unsqueeze(0), None) for p_ in y_pred]
+            # concat the list of results
+            if isinstance(ret_[0], torch.Tensor):
+                ret = torch.cat(ret_, dim=0)
+            elif isinstance(ret_[0], (list, tuple)) and all([isinstance(i, torch.Tensor) for i in ret_[0]]):
+                # if _apply() returned not only 1 Tensor, concat them separately
+                ret = [torch.cat([k[i] for k in ret_], dim=0) for i in range(len(ret_[0]))]
```
yiheng-wang-nv, line 57: Here what if `len(ret_) > 1`?

monai/metrics/regression.py
```diff
@@ -10,23 +10,38 @@
...
 from monai.metrics.utils import do_metric_reduction
 from monai.utils import MetricReduction
 
+from .metric import Metric
+
+
+class RegressionMetric(Metric):
+    """
+    Base class for regression metrics.
+    Input `y_pred` (BCHW[D] where C is number of channels) is compared with ground truth `y` (BCHW[D]).
```
yiheng-wang-nv, line 28: same comments in `monai/metrics/hausdorff_distance.py`

tests/test_handler_rocauc.py
```diff
@@ -34,6 +34,8 @@ def test_compute(self):
         y = torch.Tensor([[0], [1]])
         y_pred = act(y_pred)
         y = to_onehot(y)
+        # test a list of channel-first tensors
+        y_pred, y = list(y_pred), list(y)
         auc_metric.update([y_pred, y])
```
yiheng-wang-nv, line 39: In `monai/handlers/roc_auc.py`, the type hint for `update` is:
`def update(self, output: Tuple[torch.Tensor, torch.Tensor]`, here the input is `List[torch.Tensor, torch.Tensor]`, should the type be changed into `Union[Tuple[torch.Tensor, torch.Tensor], List[torch.Tensor, torch.Tensor]]` ?

yiheng-wang-nv (commented):
Thanks for the changes, I left some comments above.

monai/metrics/hausdorff_distance.py
Nic-Ma: updated in the latest commit.
Thanks.

monai/metrics/metric.py
```diff
@@ -0,0 +1,84 @@
...
+            y: the ground truth to compute, must be a list of `channel-first` Tensor
+                or a `batch-first` Tensor.
+
+        """
+        ret: Union[torch.Tensor, Sequence[torch.Tensor]]
+        if isinstance(y_pred, (list, tuple)) or isinstance(y, (list, tuple)):
+            # if y_pred or y is a list of channel-first data, add batch dim and compute metric
+            ret_: List[torch.Tensor]
+            if y is not None:
+                ret_ = [self._apply(p_.detach().unsqueeze(0), y_.detach().unsqueeze(0)) for p_, y_ in zip(y_pred, y)]
```
wyli, line 49: a few questions about the base class:
- how to support multiprocessing `_apply` in parallel? this class shouldn't provide the concrete implementation of the list processing, for example, we may have subclasses to do:
```
with multiprocess.pool as p:
  p.map(self._apply, zip(y_pred, y))
```

- `_apply` is unclear to me, perhaps `_compute` is better?

- I feel `reduce` could be `aggregate`, reduce is a parallel computing concept, it doesn't suit all our use cases such as computing ROC AUC.

- the postprocesses after line 52 including `ret = torch.cat(ret_, dim=0)` look like a kind of reduce/aggregate as well, why not implement them in `self.reduce`?

- does this abstraction support the pathology froc metric? https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/pathology/metrics.py#L34

monai/metrics/hausdorff_distance.py
Nic-Ma: updated in the latest commit.
Thanks.

monai/metrics/meandice.py
Nic-Ma: updated in the latest commit.
Thanks.

wyli (commented):
thanks, I put some initial comments, the metric base classes may need more work, let me check the evaluation working group discussions and think about this

monai/metrics/regression.py
Nic-Ma: updated in the latest commit.

Thanks.

monai/metrics/metric.py
Nic-Ma: Sorry I didn't get your question, ret_ is a list.

tests/test_handler_rocauc.py
ericspod: `List[torch.Tensor, torch.Tensor]` will complain about too many arguments since it's used to definite an arbitrary length list of items, so `List[T]` is a list of zero or more `T`. For `Tuple[T,T]` this states a tuple of precisely two `T` objects.

monai/metrics/metric.py
ericspod: * Reduction needs to be done across processes in a multi-GPU situation as well, see [Ignite's facility](https://github.com/pytorch/ignite/blob/dce97dfeacf1516e78b1cd9c6a37e57a1228ceed/ignite/metrics/metric.py#L519) for doing this. 
* `_compute` does sound better. 
* `aggregate` is a better term, I had thought `reduce` was a parallel operation

tests/test_handler_rocauc.py
Nic-Ma: Hi @yiheng-wang-nv ,

I used `output: Tuple[torch.Tensor, torch.Tensor]` because ignite defined the API and args. It actually can support tuple and list.

Thanks.

monai/metrics/confusion_matrix.py
```diff
@@ -43,14 +45,10 @@ class ConfusionMatrixMetric:
             Except for input only one metric, multiple metrics are also supported via input a sequence of metric names, such as
             ("sensitivity", "precision", "recall"), if ``compute_sample`` is ``True``, multiple ``f`` and ``not_nans`` will be
             returned with the same order as input names when calling the class.
-        compute_sample: if ``True``, each sample's metric will be computed first. If ``False``, the confusion matrix for each image
-            (the output of function ``get_confusion_matrix``) will be returned. In this way, users should achieve the confusion
-            matrixes for all images during an epoch and then use ``compute_confusion_matrix_metric`` to calculate the metric.
-            Defaults to ``False``.
+        compute_sample: when reducing, if ``True``, each sample's metric will be computed first. If ``False``,
+            compute reduction first, defaults to ``False``.
```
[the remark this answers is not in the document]
Nic-Ma: updated in the latest commit.
Thanks.

monai/handlers/iteration_metric.py
```diff
@@ -116,17 +104,15 @@ def compute(self) -> Any:
...
+            result = self.metric_fn.reduce(_scores)
+            result = result[0] if isinstance(result, (list, tuple)) else result
 
         if ws > 1:
             # broadcast result to all processes
             result = idist.broadcast(result, src=0)
 
         return result.item() if isinstance(result, torch.Tensor) else result
 
-    def _reduce(self, scores) -> Any:
```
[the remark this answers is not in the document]
Nic-Ma: sure, changed in the latest commit.

Thanks.

monai/metrics/metric.py
```diff
@@ -0,0 +1,84 @@
...
+        accumulate the original `y_pred` and `y`, then execute on the accumulated data.
+
+        Args:
+            y_pred: the model prediction data to compute, must be a list of `channel-first` Tensor
+                or a `batch-first` Tensor.
+            y: the ground truth to compute, must be a list of `channel-first` Tensor
+                or a `batch-first` Tensor.
+
+        """
+        ret: Union[torch.Tensor, Sequence[torch.Tensor]]
```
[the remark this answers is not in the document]
Nic-Ma: updated in the latest commit.

Thanks.

monai/metrics/metric.py
Nic-Ma: Hi @ericspod and @wyli ,

Thanks very much for your comments and suggestions, about the questions:
1. I added `_compute_list()` protected method, so a subclass can easily re-implement it with multi-threads.
2. Changed `_apply()` to `_compute()`.
3. Changed `reduce()` to `aggregate()` and updated all the tests.
4. I added this concat logic in `call()` is to make the output of the function to be a `batch-first` Tensor for both `list of channel-first tensor` and `batch-first tensor` inputs. I think we should not return `list of batch-first tensor` which is not a supported data shape in our design. Or maybe we can split this post-processing logic into a separate function?
5. I checked the FROC metric, seems it computes on some numpy array data in files which was computed by `y_pred` and `y` before. So the computation is actually only the `aggregation` part. Once we finalized the `Metric` base class, I can update it if necessary.

What do you think?

Thanks.

Nic-Ma: Hi @ericspod @wyli @rijobro , do you have any more comments on this PR?

Thanks in advance.

Nic-Ma:
As this is a very big feature update, I created a new branch: **feature/561-metrics-api** in the Project-MONAI repo.
I will merge this PR into that branch and create a new PR to dev branch.

Thanks.

PR #2299, opened 2021-06-03
Move IterationEvents into workflow

Nic-Ma:
### Description
This PR moved the `IterationEvents` registration into `Workflow` as we need to register `decollate` logic to IterationEvents.MODEL_COMPLETED event.
It also helped to remove several duplicated logic.

### Status
**Ready**

### Types of changes


yiheng-wang-nv (approved):

monai/engines/workflow.py
```diff
@@ -137,16 +137,14 @@ def set_sampler_epoch(engine: Engine):
...
-        if event_names is not None:
-            if not isinstance(event_names, list):
+        event_names = [IterationEvents] if event_names is None else event_names + [IterationEvents]
+        for name in event_names:
+            if isinstance(name, str):
+                self.register_events(name, event_to_attr=event_to_attr)
+            elif issubclass(name, EventEnum):
+                self.register_events(*name, event_to_attr=event_to_attr)
+            else:
                 raise ValueError("event_names must be a list or string or EventEnum.")
```
wyli, line 147: when event_names is not a list `event_names + [IterationEvents]` would fail already, so line 147 is not used. why there's no unit test for this PR

Nic-Ma: Hi @wyli ,

This feature is tested in the integration test:
https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_integration_workflows.py#L159
Abou the list type, we also didn't have the check in previous code, you are right, let me try to add the sanity check for `event_names` type in other PR later.

Thanks.

Nic-Ma: Added sanity check in this PR: https://github.com/Project-MONAI/MONAI/pull/2303
Thanks.

PR #2308, opened 2021-06-04
2304 enhances image dataset to handle meta

### LINKED ISSUE #2304: Spacing transforms scale pixel dimensions or resample to specified pixel dimensions
Hello! 

Is the Spacing transform meant to resample a given volume to a specified the pixel dimensions? I have NiFTI images with different voxel dimensions and am trying to resample all of them to (2mm, 2mm, 2mm) using the following code. \

```python 

fpaths = ["img0.nii", "img1.nii"]
transform = Compose([
        AsChannelFirst(channel_dim=-1),
        Spacing(pixdim=(2, 2, 2), mode="bilinear", diagonal=True),
        ScaleIntensity(minv=0.0, maxv=1.0),
        ToTensor(),
    ])
check_ds = ImageDataset(image_files=fpaths, transform=transform)

for i, img in enumerate(check_ds):

    print(check_ds[i][0].shape)

```

```img0.nii``` has a pixel size of (2mm, 2mm, 2mm) to begin with so it should remain unchanged by the transform, but its dimensions were shrunk by a factor of 2 from (128, 128, 90) to (64, 64, 46) ```img1.nii``` has a pixel size of (2mm, 2mm, 3.27mm), its first two dimensions should remain unchanged whereas the third dimension should expand to account for the reduced pixel size. However, its dimensions were reduced from (128, 128, 47) to (64, 64, 24). It appears that the Spacing transform only scales the pixel dimensions by a factor of 2 instead of resampling them to the specified dimensions. Is this the normal behavior? 

I am using Mona 0.5.2 and you can find the aforementioned NiFTI files here (https://drive.google.com/drive/folders/1k__dds_31elP-zkelCNSguWIDrHPTDe5?usp=sharing)

Thank you for looking into this.

wyli:

Fixes #2304

### Description
adds a `transform_with_metadata` to ImageDataset

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/image_dataset.py
+++ b/monai/data/image_dataset.py
@@ -26,7 +26,8 @@ class ImageDataset(Dataset, Randomizable):
     for the image and segmentation arrays separately.
     The difference between this dataset and `ArrayDataset` is that this dataset can apply transform chain to images
     and segs and return both the images and metadata, and no need to specify transform to load images from files.
-
+    For more information, please see the image_dataset demo in the MONAI tutorial repo,
+    https://github.com/Project-MONAI/tutorials/blob/master/modules/image_dataset.ipynb
     """
 
     def __init__(
@@ -37,6 +38,7 @@ class ImageDataset(Dataset, Randomizable):
         transform: Optional[Callable] = None,
         seg_transform: Optional[Callable] = None,
         image_only: bool = True,
+        transform_with_metadata: bool = False,
         dtype: DtypeLike = np.float32,
         reader: Optional[Union[ImageReader, str]] = None,
         *args,
@@ -53,6 +55,7 @@ class ImageDataset(Dataset, Randomizable):
             transform: transform to apply to image arrays
             seg_transform: transform to apply to segmentation arrays
             image_only: if True return only the image volume, otherwise, return image volume and the metadata
+            transform_with_metadata: if True, the metadata will be passed to the transforms whenever possible.
             dtype: if not None convert the loaded image to this data type
             reader: register reader to load image file and meta data, if None, will use the default readers.
                 If a string of reader name provided, will construct a reader object with the `*args` and `**kwargs`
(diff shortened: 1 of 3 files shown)


Nic-Ma:
Hi @wyli ,

Thanks for your enhancement for ImageDataset.
I think the reason we kept `ImageDataset` after developed `ArrayDataset` is that: `ImageDataset` loads image data and meta_data in this dataset instead of transforms, so it can apply following typical transform chain only for the image data and still return both transformed image data and the meta_data.
For the use case you provided in this PR, it requires every transform in the chain must accept the meta_data arg, then why not use `ArrayDataset` directly?
And to support mixed transforms that some can accept meta_data and some not, users can refer to this example:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/dataset.py#L926
What do you think?

Thanks.

wyli:
Hi @Nic-Ma, this API is used by multiple tutorials, but it's half baked. here I enhanced it and made a demo to show the usage. https://github.com/Project-MONAI/tutorials/pull/222

Nic-Ma:
Hi @wyli ,

Thanks for your tutorial, may I know why we don't use `ArrayDataset` for the case in the tutorial? Maybe you can change to show `ArrayDataset` in the tutorial?
I think we only use `ImageDataset` in the case that we need the meta_data for the following operations(like saving files) and all the transforms don't need meta_data, otherwise, `ArrayDataset` may be a better choice.

Thanks.

wyli:
with this enhancement the `imagedataset` has better handling of the metadata compared with the arraydataset, now it's a good complimentary to our dataset APIs

Nic-Ma:
Hi @wyli ,

I don't quite understand "ImageDataset has better handling of the metadata compared with the ArrayDataset", could you please help share more details?
I think we discussed to use `ArrayDataset` and delete `ImageDataset`, but `ImageDataset` has 1 benefit that can get both image data and meta_data but only run transform chain on the image data, so we keep it.

Thanks.

wyli:
the current imagedataset doesn't work with the meta data as reported in #2304, this PR fixes the issue.

Nic-Ma:
OK, I see your point now.
So you mean to support the case that: all the transforms in the Compose chain expect (img, meta) input and `ArrayDataset` can't support this case if using `Compose`, right?
That makes sense.

Thanks for your explanation.

Nic-Ma (approved):

PR #2309, opened 2021-06-04
optional `strict_check` for EnsureChannelFirst(d)

wyli:

### Description
adds a boolean flag `strict_check` to the `EnsureChannelFirst`

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/utility/array.py
+++ b/monai/transforms/utility/array.py
@@ -16,7 +16,8 @@ https://github.com/Project-MONAI/MONAI/wiki/MONAI_Design
 import logging
 import sys
 import time
-from typing import Callable, Dict, List, Optional, Sequence, Tuple, Union
+import warnings
+from typing import Callable, List, Mapping, Optional, Sequence, Tuple, Union
 
 import numpy as np
 import torch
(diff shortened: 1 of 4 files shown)


Nic-Ma:
Hi @wyli ,

Thanks for your enhancement for `EnsureChannelFirstd`.
But I think if no `meta data` provided or no `original_channel_dim` found, `EnsureChannelFirstd` can't work as expected at all, so we should raise the exception instead of just a warning.
Maybe you have some use cases for it, could you please help share it for further discussion?

Thanks.

wyli:
this is useful when customising preprocessors and composes, e.g. I was experimenting with this example, using ensure channel and changing the flags.
https://github.com/Project-MONAI/MONAI/blob/c1e06f5ada1e66a544965136d5029c0bf1f4d510/monai/data/dataset.py#L931-L943

Nic-Ma:
Thanks for your explanation.
I feel when users use `EnsureChannelFirstd` transform, they expect it to automatically guarantee a `channel-first` data, if not, raise exception instead of "can't work and return the original data" with a warning.
For experiment purpose, maybe you can easily comment this transform when you change the code to test something?
Anyway, it's not a breaking change, if @ericspod and @rijobro agree with it, I am OK to approve it.

Thanks.

wyli:
if that's the original intention then the previous function signature is also wrong `meta_dict: Optional[Dict] = None`, it shouldn't accept optional meta_dict

Nic-Ma:
Hi @wyli ,

Thanks for your finding.
I agree to change the signature from `meta_dict: Optional[Dict] = None` to `meta_dict: Dict` as the original check in the function already strongly require it to be a `dict`. Maybe I made the mistake when developing this transform.
Could you please help change this PR to update the signature of the function?

Thanks.

wyli:
> Hi @wyli ,
> 
> Thanks for your finding.
> I agree to change the signature from `meta_dict: Optional[Dict] = None` to `meta_dict: Dict` as the original check in the function already strongly require it to be a `dict`. Maybe I made the mistake when developing this transform.
> Could you please help change this PR to update the signature of the function?
> 
> Thanks.

this PR addressed a previous error, added more flexibility, kept the previous behaviour. I really don't see an issue here.

Nic-Ma:
But I think maybe the previous issue is that: the API signature is not defined clearly?
@ericspod @rijobro what do you think?

Thanks.

Nic-Ma (approved):
Since @ericspod @rijobro don't have any comments and this PR is a non-breaking enhancement, I am OK to merge it.

Thanks.

PR #2313, opened 2021-06-05
573 masked loss wrapper

### LINKED ISSUE #573: Adding a loss mask?
Are there any plans to add an easy way to include "loss mask" to `Dataset` class for segmentation task? As what I understood, it only takes `img` and `seg`?

For example I have such CT volumes with the following "windowing" that could be masked away

image

And in general the following behavior could be a nice feature

https://github.com/deepmind/deepmind-research/blob/cf0a9ca46058a61e8b66bc8708cd35ac77209bc6/hierarchical_probabilistic_unet/geco_utils.py#L100

wyli:

Fixes #573

### Description
adds a generic loss masking class

### Status
**Ready**

### Types of changes

### DIFF
--- a/docs/source/losses.rst
+++ b/docs/source/losses.rst
@@ -82,9 +82,14 @@ Registration Losses
     :members:
 
 Loss Wrappers
---------------
+-------------
 
 `MultiScaleLoss`
-~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


yiheng-wang-nv (approved):
Thanks, I don't see any mistakes here.

PR #2314, opened 2021-06-06
561 497 Provide metrics base APIs

### LINKED ISSUE #303: Evaluation subpackage and metric implementations
I'd like to use this issue to discuss the structure of the evaluation part of MONAI, essentially what is currently `monai.metrics`. I plan to do the following in a PR, but obviously would like to coordinate with you guys.

1. Rename to `monai.evaluation` to make it more generic, at first there will be modules
    1. `metrics` (Implementations of metrics)
    2. `util`(For now a confusion matrix helper to hold TP, FP, FN, TN and a wrapper to convert metrics into a `ignite.metrics.Metric`)
2. Implement the most commonly used metrics
    1. Confusion matrix based metrics (see https://en.wikipedia.org/wiki/Confusion_matrix for an overview)
    2. Hausdorff distance (95), medpy has an implementation
    3. Surface distance, medpy has an implementation
    4. Check medpy, NiftyNet, Clara Train for other metrics (see also #85)

I'm assuming you also have ideas what that package should look like and requirements that I'm not aware of. Would love to hear your thoughts! I'm also a bit biased towards segmentation, I'm guessing there are special metrics for other tasks that should be integrated.

### LINKED ISSUE #561: Additional Metric Interface to compute quantities over dataset

Nic-Ma:
Fixes #561
Part of #497  
Closes #303 
Closes #565

### Description
This PR added base class for the metrics.
And add support for list of channel-first Tensor.
It's a breaking change.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/metrics/confusion_matrix.py
+++ b/monai/metrics/confusion_matrix.py
@@ -17,10 +17,10 @@ import torch
 from monai.metrics.utils import do_metric_reduction, ignore_background
 from monai.utils import MetricReduction, ensure_tuple
 
-from .metric import Metric
+from .metric import IterationMetric
 
 
-class ConfusionMatrixMetric(Metric):
+class ConfusionMatrixMetric(IterationMetric):
     """
     Compute confusion matrix related metrics. This function supports to calculate all metrics mentioned in:
     `Confusion matrix <https://en.wikipedia.org/wiki/Confusion_matrix>`_.
@@ -49,6 +49,7 @@ class ConfusionMatrixMetric(Metric):
             if ``False``, compute reduction on the confusion matrices first, defaults to ``False``.
         reduction: {``"none"``, ``"mean"``, ``"sum"``, ``"mean_batch"``, ``"sum_batch"``,
             ``"mean_channel"``, ``"sum_channel"``}
+        get_not_nans: whether to return the `not_nans` count, if True, aggregate() returns (metric, not_nans).
 
     """
 
@@ -58,14 +59,16 @@ class ConfusionMatrixMetric(Metric):
         metric_name: Union[Sequence[str], str] = "hit_rate",
         compute_sample: bool = False,
         reduction: Union[MetricReduction, str] = MetricReduction.MEAN,
+        get_not_nans: bool = False,
     ) -> None:
         super().__init__()
         self.include_background = include_background
         self.metric_name = ensure_tuple(metric_name)
         self.compute_sample = compute_sample
         self.reduction = reduction
+        self.get_not_nans = get_not_nans
 
-    def _compute(self, y_pred: torch.Tensor, y: Optional[torch.Tensor] = None):
+    def _compute(self, y_pred: torch.Tensor, y: torch.Tensor):
         """
         Args:
             y_pred: input data to compute. It must be one-hot format and first dim is batch.
@@ -98,7 +101,7 @@ class ConfusionMatrixMetric(Metric):
             include_background=self.include_background,
         )
 
-    def aggregate(self, data: torch.Tensor):
+    def aggregate(self, data: Optional[torch.Tensor] = None):
         """
         Execute reduction for the confusion matrix values, the `data` usually is a Tensor of shape [BC4],
         Where, the third dimension represents the number of true positive, false positive, true negative
@@ -106,6 +109,7 @@ class ConfusionMatrixMetric(Metric):
         to the batch size and C equals to the number of classes that need to be computed.
 
         """
+        data = self._synced_scores if data is None else data
         results = []
         for metric_name in self.metric_name:
             if self.compute_sample:
@@ -115,7 +119,8 @@ class ConfusionMatrixMetric(Metric):
                 f, not_nans = do_metric_reduction(data, self.reduction)
                 f = compute_confusion_matrix_metric(metric_name, f)
             results.append(f)
-            results.append(not_nans)
+            if self.get_not_nans:
+                results.append(not_nans)
         return results
 
 
--- a/monai/metrics/confusion_matrix.py
+++ b/monai/metrics/confusion_matrix.py
@@ -101,12 +101,11 @@ class ConfusionMatrixMetric(CumulativeIterationMetric):
             include_background=self.include_background,
         )
 
-    def aggregate(self):
+    def aggregate(self):  # type: ignore
         """
         Execute reduction for the confusion matrix values.
 
         """
-        super().aggregate()
         data = self.get_buffer()
         if not isinstance(data, torch.Tensor):
             raise ValueError("the data to aggregate must be PyTorch Tensor.")
--- a/monai/metrics/confusion_matrix.py
+++ b/monai/metrics/confusion_matrix.py
@@ -10,17 +10,17 @@
 # limitations under the License.
 
 import warnings
-from typing import Optional, Sequence, Union
+from typing import Sequence, Union
 
 import torch
 
 from monai.metrics.utils import do_metric_reduction, ignore_background
 from monai.utils import MetricReduction, ensure_tuple
 
-from .metric import IterationMetric
+from .metric import CumulativeIterationMetric
 
 
-class ConfusionMatrixMetric(IterationMetric):
+class ConfusionMatrixMetric(CumulativeIterationMetric):
     """
     Compute confusion matrix related metrics. This function supports to calculate all metrics mentioned in:
     `Confusion matrix <https://en.wikipedia.org/wiki/Confusion_matrix>`_.
@@ -61,14 +61,14 @@ class ConfusionMatrixMetric(IterationMetric):
         reduction: Union[MetricReduction, str] = MetricReduction.MEAN,
         get_not_nans: bool = False,
     ) -> None:
-        super().__init__()
+        super().__init__(buffer_num=1)
         self.include_background = include_background
         self.metric_name = ensure_tuple(metric_name)
         self.compute_sample = compute_sample
         self.reduction = reduction
         self.get_not_nans = get_not_nans
 
-    def _compute(self, y_pred: torch.Tensor, y: torch.Tensor):  # type: ignore
+    def _compute_tensor(self, y_pred: torch.Tensor, y: torch.Tensor):  # type: ignore
         """
         Args:
             y_pred: input data to compute. It must be one-hot format and first dim is batch.
@@ -101,15 +101,13 @@ class ConfusionMatrixMetric(IterationMetric):
             include_background=self.include_background,
         )
 
-    def aggregate(self, data: Optional[torch.Tensor] = None):  # type: ignore
+    def aggregate(self):
         """
-        Execute reduction for the confusion matrix values, the `data` usually is a Tensor of shape [BC4],
-        Where, the third dimension represents the number of true positive, false positive, true negative
-        and false negative values for each channel of each sample within the input batch. Where, B equals
-        to the batch size and C equals to the number of classes that need to be computed.
+        Execute reduction for the confusion matrix values.
 
         """
-        data = self._synced_scores if data is None else data
+        super().aggregate()
+        data = self.get_synced_tensors()
         if not isinstance(data, torch.Tensor):
             raise ValueError("the data to aggregate must be PyTorch Tensor.")
 
--- a/monai/metrics/confusion_matrix.py
+++ b/monai/metrics/confusion_matrix.py
@@ -61,7 +61,7 @@ class ConfusionMatrixMetric(CumulativeIterationMetric):
         reduction: Union[MetricReduction, str] = MetricReduction.MEAN,
         get_not_nans: bool = False,
     ) -> None:
-        super().__init__(buffer_num=1)
+        super().__init__()
         self.include_background = include_background
         self.metric_name = ensure_tuple(metric_name)
         self.compute_sample = compute_sample
@@ -107,7 +107,7 @@ class ConfusionMatrixMetric(CumulativeIterationMetric):
 
         """
         super().aggregate()
-        data = self.get_synced_tensors()
+        data = self.get_buffer()
         if not isinstance(data, torch.Tensor):
             raise ValueError("the data to aggregate must be PyTorch Tensor.")
 
--- a/monai/metrics/confusion_matrix.py
+++ b/monai/metrics/confusion_matrix.py
@@ -68,7 +68,7 @@ class ConfusionMatrixMetric(IterationMetric):
         self.reduction = reduction
         self.get_not_nans = get_not_nans
 
-    def _compute(self, y_pred: torch.Tensor, y: torch.Tensor):
+    def _compute(self, y_pred: torch.Tensor, y: torch.Tensor):  # type: ignore
         """
         Args:
             y_pred: input data to compute. It must be one-hot format and first dim is batch.
@@ -101,7 +101,7 @@ class ConfusionMatrixMetric(IterationMetric):
             include_background=self.include_background,
         )
 
-    def aggregate(self, data: Optional[torch.Tensor] = None):
+    def aggregate(self, data: Optional[torch.Tensor] = None):  # type: ignore
         """
         Execute reduction for the confusion matrix values, the `data` usually is a Tensor of shape [BC4],
         Where, the third dimension represents the number of true positive, false positive, true negative
@@ -110,6 +110,9 @@ class ConfusionMatrixMetric(IterationMetric):
 
         """
         data = self._synced_scores if data is None else data
+        if not isinstance(data, torch.Tensor):
+            raise ValueError("the data to aggregate must be PyTorch Tensor.")
+
         results = []
         for metric_name in self.metric_name:
             if self.compute_sample:
--- a/monai/metrics/confusion_matrix.py
+++ b/monai/metrics/confusion_matrix.py
@@ -10,15 +10,17 @@
 # limitations under the License.
 
 import warnings
-from typing import Sequence, Union
+from typing import Optional, Sequence, Union
 
 import torch
 
 from monai.metrics.utils import do_metric_reduction, ignore_background
-from monai.utils import MetricReduction
+from monai.utils import MetricReduction, ensure_tuple
+
+from .metric import Metric
 
 
-class ConfusionMatrixMetric:
+class ConfusionMatrixMetric(Metric):
     """
     Compute confusion matrix related metrics. This function supports to calculate all metrics mentioned in:
     `Confusion matrix <https://en.wikipedia.org/wiki/Confusion_matrix>`_.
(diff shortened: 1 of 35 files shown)


Nic-Ma:
Hi @wyli ,

Thanks for providing the detection metrics example:
https://github.com/rafaelpadilla/Object-Detection-Metrics/blob/master/lib/Evaluator.py#L24
I checked the code and seems we can implement it based on this `Metric`, something like:
```py
class PascalVOCMetric(Metric):
    def __init__(self, IOUThreshold=0.5, method="EveryPointInterpolation"):
        self.IOUThreshold = IOUThreshold
        self.method = method
        self._ groundtruths = []
        self._detections = []
        self._classes = []

    # here the args and arg type don't match base class
    def _compute(self, bb):
        if bb.getBBType() == BBType.GroundTruth:
            self._groundtruths.append([
                bb.getImageName(),
                bb.getClassId(), 1,
                bb.getAbsoluteBoundingBox(BBFormat.XYX2Y2)
            ])
        else:
            self._detections.append([
                bb.getImageName(),
                bb.getClassId(),
                bb.getConfidence(),
                bb.getAbsoluteBoundingBox(BBFormat.XYX2Y2)
            ])
        if bb.getClassId() not in classes:
            classes.append(bb.getClassId())

    def aggregate(self, data):
        # final metric computation progress

    def reset(self):
        self._ groundtruths = []
        self._detections = []
        self._classes = []
```
I got several questions with current Metric design:
1. python doesn't support `overload` for different arg list, how to define the API for both the metrics with `pred_y` & `y` and the metrics with only `pred_y`? The data type of args may be also unknown. Can we define a API interface with uncertain length args? `def __call__(...)`?
2. Although this PR doesn't include member variables in the Metric base class, I agree we should add a `reset()` API to unify the logic for member variables.
3. Should we include the `multi-gpus` distributed syncing logic in `Metric` or put this task to the upper layer or let users to do it? I have a draft idea is to add a `_sync(self, data)` interface in `Metric`, will be called at the beginning of `aggregate()`, every metric class can implement the real `_sync()` logic for itself.
4. Should we keep the Metric base class simple, and make several subclasses for different categories of metrics?

Thanks.

wyli:
This PR should try to close https://github.com/Project-MONAI/MONAI/issues/303 #565 #561 with a lightweight solution, perhaps also try to be compatible with a few existing packages: 
- https://github.com/PyTorchLightning/metrics/blob/master/torchmetrics/metric.py
- https://github.com/loli/medpy/tree/master/medpy/metric

So, I think we can have `reset` and `_sync` API in a simple base class.  could you finish this PR, and then we can get feedback from the working group.

Nic-Ma:
Hi @wyli ,

Thanks for your sharing.
I investigated several other projects and got some ideas, let me try to update this PR ASAP.

Thanks.

Nic-Ma:
Hi @wyli @ericspod ,

I updated the PR based on your comments and some further investigation, could you please help review it again?

Thanks in advance.

monai/metrics/metric.py
```diff
@@ -0,0 +1,214 @@
...
+from abc import ABC, abstractmethod
+from typing import Any, List, Optional
+
+import torch
+
+from monai.config import TensorOrList
+from monai.utils import evenly_divisible_all_gather
+
+
+class Metric(ABC):
```
wyli, line 21: thanks, I rearranged a few APIs based on your impl. https://gist.github.com/wyli/0733d7965412092244e08f237bbb8205 is this feasible? in this case we'll have iterationMetric for the low-level impl. and cumulative interface might be useful for some of the metrics

Nic-Ma: Hi @wyli ,

Thanks very much for your review and suggestions, let me re-think it and try to adjust it again tomorrow.

Nic-Ma: BTW, I checked your sample code, may I know what's the difference between `IterationMetric` and `EpochMetric`? In current PR, the difference is the `y_pred` vs `scores`.

Thanks.

wyli: I amended the script, `EpochMetric` becomes a generic wrapper of IterationMetric, but ROCAUC is not compatible with the wrapper, because it aggregate differently

Nic-Ma: Hi @wyli ,

I checked your latest script, making `Cumulative` a separate class is a good idea. But I think making `EpochMetric` a wrapper is not very clear, every time users want to use MeanDice, do they need to create both MeanDiceMetric and this wrapper? And maybe different metrics have different aggregation methods(take confusion_matrix as example), we can't use the same wrapper?
Maybe we should not distinguish MeanDice and AUC, below script is slightly changed from your proposal and tried to define unified base classes for all metrics:
```py
class Metric:
  def __call__(self, ...):
    # iteration-level compute

class Cumulative:
    # utility class for cumulative metrics
  def __init__(self, buffer_count=1)
    # initialize metric buffer
    self._buffers = [[] for _ in buffer_count]
    self._synced_tensors = [torch.tensor(0) for _ in buffer_count]
    self.buffer_count = buffer_count

  def add(self, *data):
      for i, d in enumerate(data):
          self._buffers[i].append(d)

  def reset(self, ...):
    # reset the metric buffer
    self._buffers = [[] for _ in self.buffer_count]
    self._synced_tensors = [torch.tensor(0) for _ in self.buffer_count]

  def aggregate(self, ...):
    # compute result based on buffer (build by multiple __call__)
    self._sync()
    
  def _sync(self, ...):
    # sync up distributed metric when aggregate
    self._synced_tensors = [evenly_divisible_all_gather(data=torch.cat(b, dim=0), concat=True) for b in self._buffers]

  def get_synced_tensors(self):
    # need to get raw metric details to generate reports
    return self._synced_tensors

### concrete classes ###
class CumulativeMetric(Metric, Cumulative):
  def __call__(self, ...):
    # compute based on _compute_*
    if isinstance(list):
      ret = self._compute_list()
    else:
      ret = _compute_tensor()
    self.add(ret)

  def _compute_list(self, ...):
    # per iteration results on list data

  def _compute_tensor(self, ...):
    # per iteration results on tensor data

class ROCAUCMetric(CumulativeMetric):
  def __init__(self, ...):
      super().__init__(buffer_count=2)

  def _compute_tensor(self, ...):

  def aggregate(self, ...):
      super().aggregate()
      y_pred, y = self._synced_tensors[0], self._synced_tensors[1]
      ...

class MeanDiceMetric(CumulativeMetric):
  def __init__(self, ...):
      super().__init__(buffer_count=1)

  def _compute_tensor(self, ...):

  def aggregate(self, ...):
      super().aggregate()
      scores = self._synced_tensors[0]
      ...
```
What do you think?

Thanks.

wyli: sure I updated the script to unify the parts

Nic-Ma: Cool, I think we almost aligned on the APIs of metrics, I will update the PR today based on our discussion.

Thanks.

Nic-Ma: Updated the base classes in the latest commit.
Thanks for the discussion.

wyli: thanks the base classes look good to me, I'll review/revise the details if you fix the tests

Nic-Ma: Sure, I will try to update all the tests ASAP.

Thanks.

Nic-Ma:
Hi @wyli ,

I have totally updated the PR based on the latest design, could you please help review it again?

Thanks in advance.

monai/metrics/metric.py
```diff
@@ -0,0 +1,199 @@
...
+        self._sync()
+
+    def _sync(self):
+        """
+        Sync up distributed data when aggregating.
+
+        """
+        self._synced_tensors = [evenly_divisible_all_gather(torch.cat(b, dim=0), concat=True) for b in self._buffers]
+
+    def get_synced_tensors(self):
```
wyli, line 162: this method name could be `read_buffer` and it should call `self._sync`.

monai/metrics/metric.py
```diff
@@ -0,0 +1,199 @@
...
+        cum.reset()
+
+    Args:
+        buffer_num: the number of buffers to create to cumulate PyTorch Tensors,
+            usually every Tensor should map to a buffer.
+
+    """
+
+    def __init__(self, buffer_num: int = 1):
+        self.buffer_num = buffer_num
```
wyli, line 122: instead of defining `buffer_num`, it could be automatically determined at the first time `self.add`?

wyli (commented):
thanks, please see some comments for the details. would be great to have some more docstrings in the base class, but we can also do that in follow up PRs

monai/metrics/metric.py
Nic-Ma: Sounds like a good idea, let me have a try soon.
Thanks.

monai/metrics/metric.py
Nic-Ma: OK, then it also needs a flag `self._synced: bool = False` to avoid duplicated syncing, considering the case that we may need to read metric details several times to generate reports.

Thanks.

monai/metrics/metric.py
```diff
@@ -0,0 +1,199 @@
...
+        self.reset()
+
+    def reset(self):
+        """
+        Reset buffers to cumulate tensors and the synced results.
+
+        """
+        self._buffers: List[List[torch.Tensor]] = [[] for _ in range(self.buffer_num)]
```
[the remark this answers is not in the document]
Nic-Ma: OK, let me add a self._buffers = None in the init().

monai/metrics/metric.py
```diff
@@ -0,0 +1,199 @@
...
+                or a `batch-first` Tensor.
+            y: the ground truth to compute, must be a list of `channel-first` Tensor
+                or a `batch-first` Tensor.
+
+        """
+        ret: TensorOrList
+        if isinstance(y_pred, (list, tuple)) or isinstance(y, (list, tuple)):
+            # if y_pred or y is a list of channel-first data, add batch dim and compute metric
+            ret_: List[torch.Tensor] = self._compute_list(y_pred, y)
+            # concat the list of results
```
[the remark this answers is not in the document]
Nic-Ma: Hi @wyli ,

The concat logic is to guarantee the output shape of ret is `BCHW[D]`, otherwise it's `list of batch-first` which is against our principle that data in metrics should be `BCHW[D]` or `list of channel-first`.
You are right, we should do it the `_compute_list()` func, let me move it.

Thanks.

monai/metrics/metric.py
```diff
@@ -0,0 +1,199 @@
...
+
+import torch
+
+from monai.config import TensorOrList
+from monai.utils import evenly_divisible_all_gather
+
+
+class Metric(ABC):
+    """
+    Base class of all Metrics inerface.
```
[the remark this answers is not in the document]
Nic-Ma: Sure, let me try to add more docstrings in the base class.
Thanks.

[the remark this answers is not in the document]
Nic-Ma: enhanced doc-strings in the latest commit.
Thanks.

monai/metrics/metric.py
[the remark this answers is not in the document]
Nic-Ma: Moved to `_compute_list()` in the latest commit.
Thanks.

monai/metrics/metric.py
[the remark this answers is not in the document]
Nic-Ma: Added in the latest commit.
Thanks.

monai/metrics/metric.py
Nic-Ma: Updated in the latest commit.
Thanks.

monai/metrics/metric.py
Nic-Ma: Changed to automatically detect buffer size in the latest commit.
Thanks for the suggestion.

Thanks.

Nic-Ma:
Hi @wyli ,

I updated the PR based on your comments and enhanced the doc-string of base classes.
I can refine the doc-strings again according to future feedback from users and metrics working group.

Thanks.

monai/metrics/metric.py
```diff
@@ -0,0 +1,215 @@
...
+    def reset(self):
+        """
+        Reset the buffers for cumulative tensors and the synced results.
+
+        """
+        self._buffers = None
+        self._synced_tensors = None
+        self._synced = False
+
+    def add(self, *data: torch.Tensor):
```
wyli, line 137: I think every time `self.add` is called, it also should set `self._synced=False`?

Nic-Ma: Makes sense, added in the latest commit.
I think we only need to support 2 regular use cases: (1) add, add, add, aggregate, reset. (2) add, add, aggregate, add, add, aggregate, add, add, ...

Thanks.

monai/metrics/metric.py
```diff
@@ -0,0 +1,216 @@
...
+        """
+        self._sync()
+
+    def _sync(self):
+        """
+        All gather the buffers across distributed ranks for aggregating.
+        Every buffer will be concatenated as a PyTorch Tensor.
+
+        """
+        self._synced_tensors = [evenly_divisible_all_gather(torch.cat(b, dim=0), concat=True) for b in self._buffers]
```
wyli, line 174: may need to check `torch.distributed.is_available()` and `torch.distributed.is_initialized()`. Is this interface also supposed to be used for the case of single machine, multiprocessing? if not, how to do multiprocessing metrics computations? (for the example in line 112-118, `cum.add` may run in parallel to speed up the metrics computation)

wyli (commented):
thanks, just wanted to confirm that the use case of single machine multiprocesing is also covered by the design? I put detailed comments inline

monai/metrics/confusion_matrix.py
```diff
@@ -92,27 +95,34 @@ def __call__(self, y_pred: torch.Tensor, y: torch.Tensor):
...
-                raise ValueError("the sequence should at least has on metric name.")
-            results = []
-            for metric_name in self.metric_name:
-                sub_confusion_matrix = compute_confusion_matrix_metric(metric_name, confusion_matrix)
+    def aggregate(self):
+        """
+        Execute reduction for the confusion matrix values.
+
+        """
+        super().aggregate()
```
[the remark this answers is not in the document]
Nic-Ma: Currently, we only have the `sync` logic in `super().aggregate()`, so yes, it equals to call `get_buffer()` directly, but I think we may add more logic in the aggregate of base class, so I still call `super().aggregate()` here.

monai/metrics/metric.py
Nic-Ma: Hi @wyli ,

1. We already checked distributed status in `evently_divisible_all_gather()`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/utils/misc.py#L393
2. I added `torch.distributed.is_available()` check in the latest commit.
3. This interface is designed for `multi-processing` with the `sync` operation across processes, to speed up metrics computation, we usually split data to ranks by `DistributedSampler` before an epoch, every process then run its own data part according to the example in line 112 - 118. everytime we call `add()` in a process, it's a iteration, I think no need to call `add()` in parallel again, you can increase the batch size if you feel GPU is not full. I enhanced the doc-string of this interface for `multi-processing` in the latest commit.

Thanks.

wyli: hi @Nic-Ma I'm not talking about training here (so not quite related to 'epoch' or 'batch'). For example, if we have 1000 model predictions already stored in nifti files, and 1000 corresponding ground truth files, how to efficiently compute 1000 dice scores and return an mean and std over the 1000 scores? the 1000 dice computations are independent to each other, a evaluation suite would use multiple processes to speed up the whole computation. my question is how to do it with these APIs?

wyli: looks nice, could you please make a tutorial for this?

Nic-Ma: Sure, I can make a tutorial for it when this PR merged.

Thanks.

wyli (approved):
thanks!

monai/metrics/confusion_matrix.py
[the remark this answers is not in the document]
wyli: perhaps the base class's `aggregate()` should do nothing, otherwise we have 3 ways to run `_sync`: `self._sync`, `self.aggregate`, `self.get_buffer`, but I couldn't find the corresponding use cases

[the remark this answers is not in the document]
Nic-Ma: Hi @wyli ,

I think:
1. `self._sync()` is private function which is not supposed to be called by user.
2. `self.aggregate()` is the main function to compute values, it's better to automatically sync first, otherwise, users need to always call `sync` by themselves. And I can't find a use case that we should NOT sync before aggregation.
3. `self.get_buffer()` is a util function to get the metrics details or other use cases.

Thanks.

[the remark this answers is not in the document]
wyli: agree that _sync wouldnt be used by the end user, but it could be used by subclasses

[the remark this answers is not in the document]
Nic-Ma: Hi @wyli ,

Thanks for your suggestion, sounds make sense.
Let me try to update it.

Thanks.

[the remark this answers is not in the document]
wyli: sure, I think the base class aggregate should be left not implemented, the subclasses should use _sync and get_buffer to implement it, that's what is implemented as well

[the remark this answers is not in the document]
Nic-Ma: yes, I plan to update like that.
Thanks.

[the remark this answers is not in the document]
Nic-Ma: Is there a use case that users just use this class to accumulate tensors across ranks?
```py
cum = Cumulative()
cum.add(...)
cum.add(...)
data = cum.get_buffer()
cum.reset()
```
If so, we should set `pass` to `aggregate()` instead of `implemented`.

Thanks.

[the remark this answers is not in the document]
wyli: sure, I'm fine with either, depending on whether "do nothing" is considered a way of aggregation...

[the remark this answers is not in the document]
Nic-Ma: updated to `implemented` in the latest commit.

Thanks.

PR #2331, opened 2021-06-08
2330 Enhance all_gather APIs for regular PyTorch programs

### LINKED ISSUE #2330: Enhance the all_gather APIs for non-ignite programs
Currently, we have 2 enhanced `all_gather()` APIs for distributed data parallel, they are based on ignite APIs.
As they can be also used to sync data in regular PyTorch programs for metrics, classification report, etc. we need to adjust them based on PyTorch dist APIs.
And also need to fix the ignite 1024 length issue when all_gather strings.

Nic-Ma:
Fixes #2330 .

basic warnings for #2335 

### Description
This PR adjusted the `evenly_divisible_all_gather` and `string_list_all_gather` APIs to support regular PyTorch programs. 
Thanks for @vfdev-5 's help, referring to the basic idea of ignite `all_gather(string)`.
And resolved the max_length=1024 issue of ignite v0.4.4.

### Status
**Ready**

### Types of changes


monai/utils/misc.py
```diff
@@ -352,6 +356,91 @@ def copy_to_device(
     return obj
```
wyli, line 358: have you checked that the methods are equivalent as the previous ignite APIs? https://github.com/pytorch/ignite/tree/master/ignite/distributed looks like the get device logic is different

monai/utils/misc.py
```diff
@@ -352,6 +356,91 @@ def copy_to_device(
...
+    world_size = dist.get_world_size() if dist.is_initialized() else 1
+    if world_size <= 1:
+        return data
+
+    device = get_dist_device()
+    orig_device = data.device
+    data = data.to(device)
+    # tensor must have batch dimension
+    if data.ndimension() == 0:
+        data = data.unsqueeze(0)
```
wyli, line 401: if the original data is unsqueezed and `concat` is False, it seems the results will be wrong

monai/utils/misc.py
```diff
@@ -352,6 +356,91 @@ def copy_to_device(
...
+    # remove the padding items
+    output = [o[:l, ...].to(orig_device) for o, l in zip(output, all_lens_)]
+
+    return torch.cat(output, dim=0) if concat else output
+
+
+def string_list_all_gather(strings: List[str], delimiter: str = "\t") -> List[str]:
+    """
+    Utility function for distributed data parallel to all gather a list of strings.
+    Refer to the idea of ignite `all_gather(string)`.
```
wyli, line 425: perhaps put a link here to ignite for future reference

monai/handlers/utils.py
```diff
@@ -55,71 +48,6 @@ def stopping_fn(engine: Engine):
     return stopping_fn
 
 
-def evenly_divisible_all_gather(data: torch.Tensor) -> torch.Tensor:
```
wyli, line 58: I think this is a breaking change, but the PR's description says non-breaking?

wyli (commented):
thanks, please see some comments inline. there are some utility functions moved from `monai.handlers.utils` to `monai.utils`, we need [a deprecation method](https://github.com/Project-MONAI/MONAI/issues/2335) for this type of change as well @ericspod, any suggestions for this PR or shall we have another PR for the deprecating module?

monai/handlers/utils.py
Nic-Ma: You are right, the default value of `concat` is a breaking change, sorry I forgot to change the PR check box.
Updated the PR description.

Thanks.

monai/utils/misc.py
Nic-Ma: Sure, added in the latest commit.

Thanks.

monai/utils/misc.py
Nic-Ma: Thanks for your good point.
Updated in the latest commit based on the API assumption:
_The input data of every rank should have the same number of dimensions, only the first dim can be different._

Thanks.

monai/utils/misc.py
Nic-Ma: The ignite `device` API is: https://github.com/pytorch/ignite/blob/master/ignite/distributed/comp_models/native.py#L283
I think if user set `backend = GLOO`, the device should be `CPU` instead of `CUDA:x`, what do you think?

Thanks.

Nic-Ma:
Hi @wyli ,

For this kind of deprecated API, maybe I should not delete the previous implementation, just add a warning in the function:
```
warnings.warn("evenly_divisible_all_gather had been moved to monai.utils module, will deprecate this API in MONAI v0.7.", DeprecationWarning)
```
Or any other suggestions? Or some overall advanced method to handle it?

Thanks.

ericspod:
For now you can add the warning but I'm going to have a PR shortly on a more involved mechanism.

Nic-Ma:
Hi @ericspod ,

Sounds good!
I added the warning message to the deprecated APIs.

Thanks.

monai/utils/misc.py
```diff
@@ -352,6 +356,95 @@ def copy_to_device(
...
+    if dist.is_initialized():
+        backend = dist.get_backend()
+        if backend == "nccl" and torch.cuda.is_available():
+            return torch.device(f"cuda:{torch.cuda.current_device()}")
+        elif backend == "gloo":
+            return torch.device("cpu")
+    return None
+
+
+def evenly_divisible_all_gather(data: torch.Tensor, concat: bool = False):
```
wyli, line 376: if `concat` default to True, will we have better backward compatibility?

Nic-Ma: I am OK to set it default to True, actually, I am not very sure the `False` value. Let me update it.

Thanks.

monai/utils/misc.py
wyli: ok, I haven't looked into the details I think we can update it later if needed

Nic-Ma: sure, sounds good!
Thanks.

wyli (approved):
thanks, this PR also had a workaround for https://github.com/Project-MONAI/MONAI/issues/2335

PR #2337, opened 2021-06-08
Spike (Herringbone) artifact transform

yanielc:
### Description
Implementation of a transform to apply _k_-space spike artifacts on MRI images. Two versions of the transform were added to the transforms/intensity/array.py file: a deterministic and a random version which serves as a naturalistic data augmentation tool.

Spike artifacts is one of the types of artifacts which appear naturally during data acquisition. More information can be found in these  two sources:
[AAPM/RSNA physics tutorial for residents: fundamental physics of MR imaging.](https://pubmed.ncbi.nlm.nih.gov/16009826)
[Body MRI artifacts in clinical practice: A physicist's and radiologist's perspective.](https://doi.org/10.1002/jmri.24288)

~I can add the dictionary-style counterparts as well.~ [update: this is now in place]

Example of application on brain MRI images:

Screen Shot 2021-06-08 at 9 53 29 PM

### Status
**ready**

### Types of changes


monai/transforms/intensity/array.py
```diff
@@ -1292,3 +1293,288 @@ def _apply_mask(self, k: np.ndarray) -> np.ndarray:
...
+        if len(img.shape) < 3:
+            raise AssertionError("Image needs a channel direction.")
+        if isinstance(self.loc[0], int) and len(img.shape) == 4 and len(self.loc) == 2:
+            raise AssertionError("Input images of dimension 4 need location tuple to be length 3 or 4")
+        if isinstance(self.loc[0], Sequence) and len(img.shape) == 4 and min(map(lambda x: len(x), self.loc)) == 2:
+            raise AssertionError("Input images of dimension 4 need location tuple to be length 3 or 4")
+
+        # convert to ndarray to work with np.fft
+        if isinstance(img, torch.Tensor):
+            self._device = img.device
```
wyli, line 1372: just a minor comment here -- the instance variables such as `self._device` and `self.k_intensity` are not thread-safe, could you please make them local, something like `_device = img.device` or inherit `ThreadUnsafe` like https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/spatial/array.py#L392. it's fine for the `RandomizableTransform`, as `ThreadUnsafe` is one of the base classes of it.

monai/transforms/intensity/array.py
```diff
@@ -28,6 +28,7 @@
     PT_BEFORE_1_7,
     InvalidPyTorchVersionError,
     dtype_torch_to_numpy,
+    ensure_tuple,
     ensure_tuple_rep,
     ensure_tuple_size,
 )
```
Nic-Ma, line 34: Please add the new transforms in `__all__`.

Nic-Ma:
Hi @yanielc ,

Thanks for your implementation of this new transform. I have 2 minor comments:
1. Could you please help add the new transforms to the docs page: https://github.com/Project-MONAI/MONAI/blob/dev/docs/source/transforms.rst
2. Do you plan to also implement the `dict` version of these transforms?

Thanks.

monai/transforms/intensity/array.py
yanielc: Ok, will do!

yanielc:
> Hi @yanielc ,
> 
> Thanks for your implementation of this new transform. I have 2 minor comments:
> 
> 1. Could you please help add the new transforms to the docs page: https://github.com/Project-MONAI/MONAI/blob/dev/docs/source/transforms.rst
> 2. Do you plan to also implement the `dict` version of these transforms?
> 
> Thanks.

Hi @Nic-Ma,

I plan to add the dictionary style versions in this pull request as well. I just wanted to get some feedback before adding them all. 

I will add add the four transforms to the documentation pagess.

Thanks.

rijobro:
There are a few transforms now that operate in k-space (including your previous Gibbs transform). Do you think you could abstract out the FFT and iFFT to a common function to reduce code duplication?

You may also remember that in your Gibbs PR, you originally used torch for the FFT/iFFT, but I recommended that we switch to numpy because the functionality didn't exist in older versions of PT. I've since learnt that elsewhere in MONAI (e.g., [HilbertTransform](https://github.com/Project-MONAI/MONAI/blob/dev/monai/networks/layers/simplelayers.py#L350)), we use the torch version and then in the testing, we skip if the functionality [isn't present](https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_detect_envelope.py#L114-L115). Might be worth bearing in mind if we unify all FFT/iFFT functionality.

yanielc:
> There are a few transforms now that operate in k-space (including your previous Gibbs transform). Do you think you could abstract out the FFT and iFFT to a common function to reduce code duplication?
> 
> You may also remember that in your Gibbs PR, you originally used torch for the FFT/iFFT, but I recommended that we switch to numpy because the functionality didn't exist in older versions of PT. I've since learnt that elsewhere in MONAI (e.g., [HilbertTransform](https://github.com/Project-MONAI/MONAI/blob/dev/monai/networks/layers/simplelayers.py#L350)), we use the torch version and then in the testing, we skip if the functionality [isn't present](https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_detect_envelope.py#L114-L115). Might be worth bearing in mind if we unify all FFT/iFFT functionality.

I was thinking the same about a possible super class for the Fourier transforms. I would not mind writing the code and rewriting the Gibbs and Spikes transforms to inherit from the new class. To compartmentalize the work more, can I make those changes on a different pull request right after this PR is done?

Thanks for the heads up on skipping the functionality option for the older Pytorch tests.

yanielc:
Hi @Nic-Ma and other reviewers, just letting you know that the PR is ready for review.

wyli (approved):
thanks! (sorry for the delay)

PR #2373, opened 2021-06-15
2372 Add from_engine to simplify the lambda in metrics

### LINKED ISSUE #2372: Add utility to simplify metric handlers API
Currently, users must set `output_transform = lambda output: output["pred"], output["label"]` for metrics.
We have received much feedback from users that the `lambda` function is hard to understand.
Thanks for the discussion with @vfdev-5 in ignite ticket:
https://github.com/pytorch/ignite/pull/2055#issuecomment-861380281
I will try to add utility to simplify the usage of the metrics with the dictionary.

Nic-Ma:
Fixes #2372 .

### Description
This PR added `from_dict` utility to simplify the lambda arg of ignite metrics, based on discussion:
https://github.com/pytorch/ignite/pull/2055#issuecomment-861380281

### Status
**Ready**

### Types of changes


monai/handlers/utils.py
```diff
@@ -228,3 +229,24 @@ def _compute_op(op: str, d: np.ndarray):
                     f.write(f"class{deli}{deli.join(ops)}\n")
                     for i, c in enumerate(np.transpose(v)):
                         f.write(f"{class_labels[i]}{deli}{deli.join([f'{_compute_op(k, c):.4f}' for k in ops])}\n")
+
+
+def from_dict(keys: Sequence[str]):
```
wyli, line 234: from_dict is unclear to me, I think it should be `from_engine` or `from_engine_state`. I suppose this could also be used for `engine.state.batch`, why is it only described for the output? I think if this function doesn't have a good name and docstring, it'll be worse than a lambda function

Nic-Ma: Hi @wyli ,

Thanks for your suggestions!
Yes, it can also be used with `engine.state.batch`, and we can even enhance it to extract keys from a list of dict after decollacting.
Let me try to update the name and doc-string.

Thanks.

Nic-Ma: Updated in the latest commit.

Thanks.

monai/handlers/utils.py
```diff
@@ -228,3 +229,28 @@ def _compute_op(op: str, d: np.ndarray):
...
+
+        metric = MeanDice(
+            include_background=False,
+            output_transform=from_engine(["pred", "label"])
+        )
+
+    """
+
+    def _wrapper(output: Dict):
+        return tuple(output[k] for k in keys)
```
wyli, line 254: this could be `ensure_tuple(keys)`, so that it supports single string as the key

wyli (approved):

monai/handlers/utils.py
Nic-Ma: sure, updated in the latest commit.

Thanks.

PR #2379, opened 2021-06-16
2369 refactoring decollate

### LINKED ISSUE #2369: Should detach the Tensor before decollate_batch
**Describe the bug**
Currently, we decollate a Tensor data(like: pred) from batch to list without `detach()` first.
I will enhance it soon.

wyli:

Fixes #2369

### Description
simplified the decollate logic

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -335,15 +335,17 @@ def decollate_batch(batch, detach=True):
             batch = batch.detach()
         if batch.ndim == 0 and detach:
             return batch.item()
-        out_list = list(torch.unbind(batch, dim=0))
+        out_list = torch.unbind(batch, dim=0)
         if out_list[0].ndim == 0 and detach:
             return [t.item() for t in out_list]
-        return out_list
+        return list(out_list)
     if isinstance(batch, Mapping):
         _dict_list = {key: decollate_batch(batch[key]) for key in batch}
         return [dict(zip(_dict_list, item)) for item in zip(*_dict_list.values())]
-    if isinstance(batch, Iterable) and not isinstance(next(iter(batch)), (str, bytes)):
-        return [list(item) for item in zip(*(decollate_batch(b) for b in batch))]
+    if isinstance(batch, Iterable):
+        item_0 = first(batch)
+        if isinstance(item_0, Iterable) and not isinstance(item_0, (str, bytes)):
+            return [list(item) for item in zip(*(decollate_batch(b) for b in batch))]
     return batch
 
 
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -356,7 +356,7 @@ def decollate_batch(batch, detach=True):
             # Not running the usual list decollate here:
             # don't decollate ['test', 'test'] into [['t', 't'], ['e', 'e'], ['s', 's'], ['t', 't']]
             # torch.tensor(0) is iterable but iter(torch.tensor(0)) raises TypeError: iteration over a 0-d tensor
-            return list(decollate_batch(b, detach) for b in batch)
+            return [decollate_batch(b, detach) for b in batch]
         return [list(item) for item in zip(*(decollate_batch(b, detach) for b in batch))]
     raise NotImplementedError(f"Unable to de-collate: {batch}, type: {type(batch)}.")
 
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -327,19 +327,23 @@ def decollate_batch(batch, detach=True):
 
     Args:
         batch: data to be de-collated.
-        detach: whether to detach the tensors.
+        detach: whether to detach the tensors. Scalars tensors will be detached into number types
+            instead of torch tensors.
     """
     if isinstance(batch, torch.Tensor):
         if detach:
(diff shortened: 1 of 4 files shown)


monai/data/utils.py
```diff
@@ -315,63 +315,38 @@ def decollate_batch(data: Union[dict, list, torch.Tensor], batch_size: Optional[
...
+        if detach:
+            batch = batch.detach()
+        if batch.ndim == 0 and detach:
+            return batch.item()
+        out_list = torch.unbind(batch, dim=0)
+        if out_list[0].ndim == 0 and detach:
+            return [t.item() for t in out_list]
+        return list(out_list)
+    if isinstance(batch, Mapping):
+        _dict_list = {key: decollate_batch(batch[key]) for key in batch}
```
Nic-Ma, line 343: Should we add `detach=detach` in the recursive call?

monai/data/utils.py
```diff
@@ -315,63 +315,38 @@ def decollate_batch(data: Union[dict, list, torch.Tensor], batch_size: Optional[
...
+        return [dict(zip(_dict_list, item)) for item in zip(*_dict_list.values())]
+    if isinstance(batch, Iterable):
+        item_0 = first(batch)
+        if isinstance(item_0, Iterable) and not isinstance(item_0, (str, bytes)):
```
Nic-Ma, line 347: Just to confirm: so here you suppose every item in the `batch` should be Iterable?

Nic-Ma (commented):
Thanks for refactoring the `decollate` logic, I put a minor comment about the `detach=detach` arg in the recursive call, others look good to me.

Thanks.

monai/data/utils.py
wyli: that's a good one, I'll update soon

monai/data/utils.py
Nic-Ma: Because I am thinking about how to decollate `engine.state.output` which contains `"loss"： XXX`, which is a scalar number without batch dim?

Thanks.

wyli: could you make a test case to show the usage? I'll update the code, currently it wouldn't decollate

Nic-Ma: I mean for example:
```py
engine.state.output = {
    "image": torch.tensor([[[1, 2]], [[3, 4]]]),
    "label": torch.tensor([[[5, 6]], [[7, 8]]]),
    "pred": torch.tensor([[[9, 10]], [[11, 12]]]),
    "loss": 0.56,
}
```
I think your current code can handle `image`, `label`, `pred` then make a list of dictionaries, but how do you handle the `loss` in this case? Just copy to every dictionary in the decollated list? like:
```
[
    {"image": torch.tensor([[1, 2]]), "label": torch.tensor([[5, 6]]), "pred": torch.tensor([[9, 10]]), "loss": 0.56},
    {"image": torch.tensor([[3, 4]]), "label": torch.tensor([[7, 8]]), "pred": torch.tensor([[11, 12]]), "loss": 0.56},
]
```
How about if `engine.state.output` is a list and the **first item is loss**?
```py
engine.state.output = [0.56, torch.tensor([[[1, 2]], [[3, 4]]]), torch.tensor([[[5, 6]], [[7, 8]]]), torch.tensor([[[9, 10]], [[11, 12]]])]
```

Thanks.

wyli: ok, I feel these are out of the scope of this utility, this function will only do the 'inverse' according to pytorch's 'default_collate'
mainly:
```py
>>> import torch
>>> from torch.utils.data._utils.collate import default_collate
>>> from monai.data.utils import decollate_batch
>>> default_collate([[0, 0]])
[tensor([0]), tensor([0])]
>>> decollate_batch([torch.tensor([0]), torch.tensor([0])])
[[0, 0]]
>>> 
>>> default_collate([0])
tensor([0])
>>> decollate_batch(torch.tensor([0]))
[0]

```

Nic-Ma: Sure, this PR is good enough for the current situation, let's discuss the scalar case later in my PR: https://github.com/Project-MONAI/MONAI/pull/2315.

Thanks.

Nic-Ma (approved):

monai/data/utils.py
```diff
@@ -330,23 +330,35 @@ def decollate_batch(batch, detach=True):
...
-    return batch
+        if (
+            not isinstance(item_0, Iterable)
+            or isinstance(item_0, (str, bytes))
+            or (isinstance(item_0, torch.Tensor) and item_0.ndim == 0)
+        ):
+            # Not running the usual list decollate here:
+            # don't decollate ['test', 'test'] into [['t', 't'], ['e', 'e'], ['s', 's'], ['t', 't']]
+            # torch.tensor(0) is iterable but iter(torch.tensor(0)) raises TypeError: iteration over a 0-d tensor
+            return list(decollate_batch(b, detach) for b in batch)
```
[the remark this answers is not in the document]
wyli: mainly to detach the case of `[tensor(1), tensor(2), tensor(3)]` if needed, otherwise this line could be `list(batch)`

PR #2389, opened 2021-06-18
Update crop and pad doc-strings for the ROI size

Nic-Ma:
### Description
According to the feedback from Yucheng(our internal team), this PR enhanced the doc-strings of all the pad and crop transforms to make it clear about the output data size.

### Status
**Ready**

### Types of changes


monai/transforms/croppad/array.py
```diff
@@ -214,6 +215,9 @@ def __call__(self, img: np.ndarray, mode: Optional[Union[NumpyPadMode, str]] = N
 class SpatialCrop(Transform):
     """
     General purpose cropper to produce sub-volume region of interest (ROI).
+    If a dimension of the expected ROI size is bigger than the input image size, will not crop that dimension.
+    So the cropped result may be smaller than expected ROI, and the cropped results of several images may not
+    have exactly same shape.
```
yiheng-wang-nv, line 220: ... be smaller than **the** expected ROI, and the cropped results of several images may not have exactly **the** same shape.

Same for other docstring.

monai/transforms/croppad/array.py
```diff
@@ -272,9 +278,13 @@ def __call__(self, img: Union[np.ndarray, torch.Tensor]):
 class CenterSpatialCrop(Transform):
     """
     Crop at the center of image with specified ROI size.
+    If a dimension of the expected ROI size is bigger than the input image size, will not crop that dimension.
+    So the cropped result may be smaller than expected ROI, and the cropped results of several images may not
+    have exactly same shape.
 
     Args:
         roi_size: the spatial size of the crop region e.g. [224,224,128]
+            if a dimension of ROI size is bigger than image size, will not crop that dimension of the image.
```
yiheng-wang-nv, line 287: Hi @Nic-Ma , do you think that we can add an example here?
e.g. 
```
example = torch.randn([3, 3, 3, 3])
cropper = CenterSpatialCrop(roi_size=[4, 2, 2])
print(cropper(example).shape)
#  torch.Size([3, 3, 2, 2])
```

monai/transforms/croppad/array.py
Nic-Ma: Sure, updated in the latest commit.

Thanks.

monai/transforms/croppad/array.py
Nic-Ma: added example in the latest commit.

Thanks.

yiheng-wang-nv (approved):

Issue #2391, opened 2021-06-18
pytype doesn't work on macOS


wyli:
**Describe the bug**
`./runtests.h --pytype` gives an error on macOS, looks like it's a pytype issue https://github.com/google/pytype/issues/661

<details>

```
pytype
2021.06.17
Computing dependencies
Analyzing 180 sources with 8 local dependencies
ninja: Entering directory `.pytype'
[1/120] check monai.data.dataloader
FAILED: /Users/testing/Documents/MONAI/.pytype/pyi/monai/data/dataloader.pyi 
/usr/local/anaconda3/envs/py37/bin/python -m pytype.single --disable pyi-error --imports_info /Users/testing/Documents/MONAI/.pytype/imports/monai.data.dataloader.imports --module-name monai.data.dataloader -V 3.7 -o /Users/testing/Documents/MONAI/.pytype/pyi/monai/data/dataloader.pyi --analyze-annotated --check-parameter-types --nofail --precise-return --protocols --quick /Users/testing/Documents/MONAI/monai/data/dataloader.py
ERROR:pytype.imports_map_loader Invalid imports_map entries (checking from root dir: /Users/testing/Documents/MONAI/.pytype)
ERROR:pytype.imports_map_loader   file does not exist: '/Users/testing/Documents/MONAI/.pytype/pyi/monai/data/dataloader.pyi' (mapped from 'monai/data/dataloader')
ERROR:pytype.imports_map_loader   file does not exist: '/Users/testing/Documents/MONAI/.pytype/pyi/monai/data/image_reader.pyi' (mapped from 'monai/data/image_reader')
ERROR:pytype.imports_map_loader   file does not exist: '/Users/testing/Documents/MONAI/.pytype/pyi/monai/data/nifti_saver.pyi' (mapped from 'monai/data/nifti_saver')
ERROR:pytype.imports_map_loader   file does not exist: '/Users/testing/Documents/MONAI/.pytype/pyi/monai/data/nifti_writer.pyi' (mapped from 'monai/data/nifti_writer')
ERROR:pytype.imports_map_loader   file does not exist: '/Users/testing/Documents/MONAI/.pytype/pyi/monai/data/png_saver.pyi' (mapped from 'monai/data/png_saver')
ERROR:pytype.imports_map_loader   file does not exist: '/Users/testing/Documents/MONAI/.pytype/pyi/monai/data/png_writer.pyi' (mapped from 'monai/data/png_writer')
[...]

PR #2392, opened 2021-06-18
2391 skip pytype script on macos

wyli:

temp. workaround for #2391, could revert this update once it's fixed

### Status
**Ready**

### Types of changes


Nic-Ma (approved):
Thanks for the quick fix, looks good to me.
@madil90 could you please help double confirm it?

Thanks in advance.

PR #2408, opened 2021-06-22
2407 Add support to set number classes for TorchVision models

### LINKED ISSUE #2407: Add support to change class number of TorchVision models
Hogler raised this feature request internally, need to support TorchVison models with specified number of classes.

Nic-Ma:
Fixes #2407 .

### Description
This PR added the wrapper class to finetune 2D classification models, support to change the output classes of FC layer.

### Status
**Ready**

### Types of changes


monai/networks/nets/finetune_fc.py
```diff
@@ -0,0 +1,85 @@
...
+# limitations under the License.
+
+from typing import Tuple, Union
+
+import torch
+
+
+class FinetuneFC(torch.nn.Module):
+    """
+    Wrapper to customize the fully connected layer of 2D classification model or replace it by convolutional layer.
```
wyli, line 19: this class should support 3D models as well. the `args` descriptions could be more generic

monai/networks/nets/finetune_fc.py
```diff
@@ -0,0 +1,85 @@
...
+        use_conv: bool = False,
+        pool_size: Union[int, Tuple[int, int]] = (7, 7),
+        pool_stride: Union[int, Tuple[int, int]] = 1,
+    ):
+        super().__init__()
+        layers = list(model.children())
+
+        # check if the model is compatible
+        if not str(layers[-1]).startswith("Linear"):
+            raise ValueError("input model does not have a Linear layer at the end.")
```
wyli, line 47: perhaps this check is not needed, the final layers might be some activation/norm layers

monai/networks/nets/finetune_fc.py
```diff
@@ -0,0 +1,85 @@
...
+        orig_fc = layers[-1]
+        self.fc: Union[torch.nn.Linear, torch.nn.Conv2d]
+        if use_conv:
+            if not str(layers[-2]).startswith("AdaptiveAvgPool2d"):
+                raise ValueError("input model does not have a AdaptiveAvgPool2d layer next to the end.")
```
wyli, line 52: this check is not needed for this generic impl.

monai/networks/nets/finetune_fc.py
```diff
@@ -0,0 +1,85 @@
...
+
+            # remove the last Linear layer (fully connected) and the adaptive avg pooling
+            self.features = torch.nn.Sequential(*layers[:-2])
+            # add 7x7 avg pooling (in place of adaptive avg pooling)
+            self.pool = torch.nn.AvgPool2d(kernel_size=pool_size, stride=pool_stride)
```
wyli, line 57: this should be implemented with the Pool network factory, perhaps  we could also add `get_pool_layer` similar to https://github.com/Project-MONAI/MONAI/blob/075bccd161062e9894f9430714ba359f162d848c/monai/networks/layers/utils.py#L20
also if `pool_param` is None, skip adding this layer

monai/networks/nets/finetune_fc.py
```diff
@@ -0,0 +1,85 @@
...
+            # add 1x1 conv (it behaves like a FC layer)
+            self.fc = torch.nn.Conv2d(
+                in_channels=orig_fc.in_features,  # type: ignore
+                out_channels=n_classes,
+                kernel_size=(1, 1),
+            )
+        else:
+            # remove the last Linear layer (fully connected)
+            self.features = torch.nn.Sequential(*layers[:-1])
+            # replace the out_features of FC layer
```
wyli, line 67: may need an adaptive pooling layer here so that we can always flatten properly

wyli (commented):

monai/networks/nets/finetune_fc.py
Nic-Ma: if FC is not the last layer, how do we detect the FC layer?

Thanks.

wyli: the wrapper is to remove the existing final layers, and construct new ones, as long as `self.features = torch.nn.Sequential(*layers[:-2])` produces some feature maps in shape BxCxSpatialdims it should be ok.

Nic-Ma: Hi @wyli ,

But if the last several layers are uncertain, how can we make sure the feature maps is [:**-2**]?
For example, if has activation layer, I think we should use [:-3]?
And should we also add the activation layer after the modified FC layer if original model has it?
I feel maybe it's not easy to support the case that FC is not last layer, or maybe I didn't get your point clearly?

Thanks.

wyli: this is a low-level component, this class could be configured to do two things: (1) replace the last two layers with a 'pooling + conv'; or (2) replace the last layer with a 'linear'. whether the input `model` is suitable for this component should be decided by the upper-level components and probably depending on the applications... we can add an optional ADN layer at the end but it could also be a postprocessing step, so it's fine to omit that for now.

Nic-Ma: Sure, I will try to update it ASAP today.
Thanks for your suggestions.

monai/networks/nets/finetune_fc.py
Nic-Ma: Hi @wyli ,

I adapted the code from:
https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L243
I think maybe even with `AdaptiveAvgPool`, we still need to flatten the tensor?

Thanks.

monai/networks/nets/finetune_fc.py
Nic-Ma: sure, updated in the latest commit.

Thanks.

monai/networks/nets/finetune_fc.py
Nic-Ma: added support for 3D and more generic cases in the latest commit.
Thanks.

monai/networks/nets/finetune_fc.py
Nic-Ma: updated in the latest commit.
Thanks.

monai/networks/nets/finetune_fc.py
Nic-Ma: Sure, added `get_pool_layer()` utils in the latest commit.

Thanks.

Nic-Ma:
Hi @wyli ,

Thanks for your great suggestions, I updated the PR based on your comments.

Thanks in advance.

monai/networks/nets/finetune_fc.py
```diff
@@ -0,0 +1,93 @@
...
+                # add 7x7 avg pooling
+                self.pool = get_pool_layer(
+                    name=("avg", {"kernel_size": pool_size, "stride": pool_stride}),
+                    spatial_dims=dim,
+                )
+            # add 1x1 conv (it behaves like a FC layer)
+            self.fc = Conv[Conv.CONV, dim](
+                in_channels=orig_fc.in_features,  # type: ignore
+                out_channels=n_classes,
+                kernel_size=(1, 1) if dim == 2 else (1, 1, 1),
```
wyli, line 71: `kernel_size=1` covers the  cases

monai/networks/nets/finetune_fc.py
```diff
@@ -0,0 +1,93 @@
...
+        pool_stride: if using convolutional layer to replace the FC layer, it defines the stride for `AvgPool`,
+            default to 1.
+        bias: the bias value when replacing FC layer. if False, the layer will not learn an additive bias,
+            default to True.
+
+    """
+
+    def __init__(
+        self,
+        model: torch.nn.Module,
```
wyli, line 44: need options of `in_channels` if `orig_fc.in_features` not available

monai/networks/nets/finetune_fc.py
```diff
@@ -0,0 +1,93 @@
...
+# limitations under the License.
+
+from typing import Optional, Tuple, Union
+
+import torch
+
+from monai.networks.layers import Conv, get_pool_layer
+
+
+class FinetuneFC(torch.nn.Module):
```
wyli, line 19: name could be `AdaptiveFC`? I think this one is  now generic and could be used in use cases other than fine tuning

monai/networks/nets/finetune_fc.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    Wrapper to customize the fully connected layer of a classification model or replace it by convolutional layer.
```
wyli, line 21:
```suggestion
    Wrapper to customize the final fully connected layer of a model or replace it by a convolutional layer.
```

monai/networks/nets/finetune_fc.py
```diff
@@ -0,0 +1,93 @@
...
+    This module expects the output of `model layers[0: -2]` is a feature map with shape [B, C, spatial dims],
+    then replace the model's last two layers with a `pooling + conv`, or replace the last layer with a `linear`.
+
+    Args:
+        model: PyTorch model with fully connected layer at the end, support both 2D and 3D models.
```
wyli, line 26:
```suggestion
        model: a PyTorch model, support both 2D and 3D models.
```

monai/networks/nets/finetune_fc.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+
+    def __init__(
+        self,
+        model: torch.nn.Module,
+        n_classes: int = 1,
+        dim: int = 2,
+        use_conv: bool = False,
+        pool_size: Optional[Union[int, Tuple[int, int]]] = 7,
+        pool_stride: Optional[Union[int, Tuple[int, int]]] = 1,
```
wyli, line 49: this API should be more generic now
```suggestion
        pool = ("avg", {"kernel_size": 7, "stride": 1})
```

wyli (commented):
thanks, please see some comments inline

monai/networks/nets/finetune_fc.py
wyli: in this case, shape [B, C, spatials] will be flattened to `B, Cxspatials`... I think if we have an adaptive layer, it'll always be flattened to `B, C`

monai/networks/nets/finetune_fc.py
```diff
@@ -0,0 +1,93 @@
...
+        super().__init__()
+        layers = list(model.children())
+
+        orig_fc = layers[-1]
+        self.fc: Union[torch.nn.Linear, torch.nn.Conv2d, torch.nn.Conv3d]
+        self.pool = None
+        if use_conv:
+            # remove the last Linear layer (fully connected) and the adaptive avg pooling
+            self.features = torch.nn.Sequential(*layers[:-2])
+            if pool_size is not None and pool_stride is not None:
```
wyli, line 61: perhaps if pool is None, self.eatures should use one more layer from the original one,  `layers[:-1]`

monai/networks/nets/finetune_fc.py
Nic-Ma: changed to `NetAdapter` in the latest commit.
Thanks.

monai/networks/nets/finetune_fc.py
Nic-Ma: sure, added in the latest commit.
Thanks.

monai/networks/nets/finetune_fc.py
Nic-Ma: Added adaptivepooling in the latest commit.
Thanks.

monai/networks/nets/finetune_fc.py
Nic-Ma: sure, updated in the latest commit.
Thanks.

monai/networks/nets/finetune_fc.py
Nic-Ma: sure, updated in the latest commit.
Thanks.

monai/networks/nets/finetune_fc.py
Nic-Ma: sure, updated in the latest commit.
Thanks.

monai/networks/nets/finetune_fc.py
Nic-Ma: updated in the latest commit.
Thanks.

monai/networks/nets/finetune_fc.py
Nic-Ma: updated in the latest commit.
Thanks.

Nic-Ma:
Hi @wyli ,

Thanks for your cool suggestions.
I updated this PR according to your comments.

Thanks.

wyli (approved):
thanks, this looks nice!

monai/networks/layers/utils.py
```diff
@@ -92,3 +92,26 @@ def get_dropout_layer(name: Union[Tuple, str, float, int], dropout_dim: Optional
         drop_name, drop_args = split_args(name)
     drop_type = Dropout[drop_name, dropout_dim]
     return drop_type(**drop_args)
+
+
+def get_pool_layer(name: Union[Tuple, str], spatial_dims: Optional[int] = 1):
+    """
+    Create a pooling layer instance.
+
+    For example, to create normalization layers:
```
wyli, line 101: Docstring needs an update here

monai/networks/nets/torchvision_fc.py
```diff
@@ -9,66 +9,56 @@
...
 
 models, _ = optional_import("torchvision.models")
 
 
-__all__ = ["TorchVisionFullyConvModel"]
+__all__ = ["TorchVisionFCModel"]
 
 
-class TorchVisionFullyConvModel(torch.nn.Module):
+class TorchVisionFCModel(NetAdapter):
```
wyli, line 23: This might break an existing pathology MMAR, need to create an alias for the old name?

monai/networks/nets/torchvision_fc.py
```diff
@@ -9,66 +9,56 @@
...
         self,
         model_name: str = "resnet18",
         n_classes: int = 1,
-        pool_size: Union[int, Tuple[int, int]] = (7, 7),
-        pool_stride: Union[int, Tuple[int, int]] = 1,
+        dim: int = 2,
+        in_channels: Optional[int] = None,
+        use_conv: bool = False,
+        pool: Optional[Tuple[str, Dict[str, Any]]] = ("avg", {"kernel_size": 7, "stride": 1}),
+        bias: bool = True,
```
wyli, line 52: I think the layer type checks could be enabled here

Nic-Ma: Do you mean to check that the last layer or both the last FC layer and pooling layer as before?

Thanks.

monai/networks/nets/torchvision_fc.py
Nic-Ma: I feel no need to create a alias, all the MMAR should totally update when adapting to a new Clara version later.
What do you think?

Thanks.

monai/networks/nets/torchvision_fc.py
Nic-Ma: BTW, is there any Torchvision classification model doesn't use FC as the final layer and we also want to support it?
If so, we should not check the layer type.

Thanks.

monai/networks/nets/torchvision_fc.py
wyli: but we want to support mmar downloading in the next monai release... merging this PR will break the dev test at the moment https://github.com/Project-MONAI/MONAI/blob/c20f70bf8c36c09630f7add95f823dbbcf1e9a73/tests/test_mmar_download.py#L87
maybe it's a good opportunity to think about backward comp. now

monai/networks/layers/utils.py
Nic-Ma: sorry for the typo, updated in the latest commit.
Thanks.

monai/networks/nets/torchvision_fc.py
wyli: yes, this model is from `getattr(models, model_name)`, so we can always check them

monai/networks/nets/torchvision_fc.py
Nic-Ma: OK, let me create an alias class for it.
Thanks for letting me know the potential issue.

monai/networks/nets/torchvision_fc.py
Nic-Ma: Added logic to check the final FC layer in the latest commit.
Thanks.

monai/networks/nets/torchvision_fc.py
Nic-Ma: Added back `TorchVisionFullyConvModel` in the latest commit for compatibility.
Will remove it in v0.7.0 when the MMAR updated to use new `TorchVisionFCModel` API.

Thanks.

Nic-Ma:
Hi @wyli ,

Thanks for pointing out the potential issue, I updated the PR according to your comments.

Thanks in advance.

monai/networks/nets/torchvision_fc.py
```diff
@@ -9,18 +9,68 @@
...
+            dim=dim,
+            in_channels=in_channels,
+            use_conv=use_conv,
+            pool=pool,
+            bias=bias,
+        )
+
+
+@deprecated(since="0.6.0", version_val="0.7.0", msg_suffix="please consider to use `TorchVisionFCModel` instead.")
 class TorchVisionFullyConvModel(torch.nn.Module):
```
wyli, line 74: could implement the current `TorchVisionFullyConvModel ` with the new `TorchVisionFCModel` class? it's also to test that the new API is more flexible

Nic-Ma: Hi @wyli ,

Sure, updated to use the new `TorchVisionFCModel` class in the latest commit.

Thanks in advance.

tests/test_net_adatpter.py
```diff
@@ -0,0 +1,65 @@
+# Copyright 2020 - 2021 MONAI Consortium
```
wyli, line 1: there's a typo in the filename. could you make the test cases smaller? it's currently slow on CPU

Nic-Ma: Sure, that may because the newly added 3D test cases are too big, I changed the input image size in the latest commit.

Thanks.

Nic-Ma:
> /integration-test

Hi @wyli ,

It seems that integration tests can not be triggered today, something related to the new Blossom CI?

Thanks.

wyli (approved):

Nic-Ma:
> > /integration-test
> 
> Hi @wyli ,
> 
> It seems that integration tests can not be triggered today, something related to the new Blossom CI?
> 
> Thanks.

I think @IsaacYangSLA resolved the integration issue by adding missing package to requirements-dev.txt yesterday.
@IsaacYangSLA please feel free to correct me if I misunderstood anything.

Thanks for the quick fix!

PR #2416, opened 2021-06-22
2415 - add json config model support for mmar loading

### LINKED ISSUE #2415: `load_from_mmar` to support model config from the mmar json
**Describe the bug**
part of #2273 
the current loader supports loading a config dictionary from `model.pt` (and assumes `model.pt` contains the dict).
this should be enhanced to support loading from json config, in case that `model.pt` doesn't contain the full model specification.

wyli:

Fixes #2415 

### Description
adds support of model config from json

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/apps/mmars/mmars.py
+++ b/monai/apps/mmars/mmars.py
@@ -138,7 +138,7 @@ def load_from_mmar(
     # loading with `torch.load`
     model_dict = torch.load(model_file, map_location=map_location)
     if weights_only:
-        return model_dict.get("model", model_dict)  # model_dict["model"] or model_dict directly
+        return model_dict.get(model_key, model_dict)  # model_dict[model_key] or model_dict directly
 
     # 1. search `model_dict['train_config]` for model config spec.
     model_config = _get_val(dict(model_dict).get("train_conf", {}), key=model_key, default={})
@@ -168,7 +168,11 @@ def load_from_mmar(
         model_module, model_name = model_config.get("path", ".").rsplit(".", 1)
         model_cls, has_cls = optional_import(module=model_module, name=model_name)
         if not has_cls:
-            raise ValueError(f"Could not load model config {model_config.get('path', '')}.")
+            raise ValueError(
+                f"Could not load MMAR model config {model_config.get('path', '')}, "
+                f"Please make sure MMAR's sub-folders in '{model_dir}' is on the PYTHONPATH."
+                "See also: https://docs.nvidia.com/clara/clara-train-sdk/pt/byom.html"
+            )
     else:
         raise ValueError(f"Could not load model config {model_config}.")
 
@@ -180,7 +184,7 @@ def load_from_mmar(
     else:
         model_inst = model_cls()
     if pretrained:
-        model_inst.load_state_dict(model_dict.get("model", model_dict))
+        model_inst.load_state_dict(model_dict.get(model_key, model_dict))
     print("\n---")
     print(f"For more information, please visit {item[Keys.DOC]}\n")
     return model_inst
--- a/monai/apps/mmars/mmars.py
+++ b/monai/apps/mmars/mmars.py
@@ -16,6 +16,7 @@ See Also:
     - https://docs.nvidia.com/clara/clara-train-sdk/pt/mmar.html
 """
 
+import json
 import os
 import warnings
 from typing import Mapping
@@ -89,7 +90,15 @@ def download_mmar(item, mmar_dir=None, progress: bool = True):
     return model_dir
 
 
-def load_from_mmar(item, mmar_dir=None, progress: bool = True, map_location=None, pretrained=True, weights_only=False):
+def load_from_mmar(
+    item,
+    mmar_dir=None,
+    progress: bool = True,
+    map_location=None,
+    pretrained=True,
+    weights_only=False,
+    model_key: str = "model",
+):
     """
     Download and extract Medical Model Archive (MMAR) model weights from Nvidia Clara Train.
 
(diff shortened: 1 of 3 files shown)


monai/apps/mmars/mmars.py
```diff
@@ -126,27 +138,65 @@ def load_from_mmar(item, mmar_dir=None, progress: bool = True, map_location=None
...
+    if not model_config:
+        # 2. search json CONFIG_FILE for model config spec.
+        json_path = os.path.join(model_dir, item.get(Keys.CONFIG_FILE, "config_train.json"))
+        with open(json_path) as f:
+            conf_dict = json.load(f)
+        conf_dict = dict(conf_dict)
+        model_config = _get_val(conf_dict, key=model_key, default={})
+    if not model_config:
+        # 3. search `model_dict` for model config spec.
+        model_config = _get_val(dict(model_dict), key=model_key, default={})
```
Nic-Ma, line 154: May I know what this case is for? I think Clara `model.pt` doesn't have model config in it directly.

Thanks.

monai/apps/mmars/mmars.py
```diff
@@ -126,27 +138,65 @@ def load_from_mmar(item, mmar_dir=None, progress: bool = True, map_location=None
...
+
+    # parse `model_config` for model class and model parameters
+    if model_config.get("name"):  # model config section is a "name"
         model_name = model_config["name"]
         model_cls = monai_nets.__dict__[model_name]
-    else:  # model config section is a "path"
+    elif model_config.get("path"):  # model config section is a "path"
         # https://docs.nvidia.com/clara/clara-train-sdk/pt/byom.html
-        model_module, model_name = model_config.get("path", "").rsplit(".", 1)
+        model_module, model_name = model_config.get("path", ".").rsplit(".", 1)
```
Nic-Ma, line 168: Could you please help add a warning or some other notification that the model specified by "path" may be a BYOC component in the custom of MMAR, probably can't find it by `optional_import`.

Thanks.

Nic-Ma (commented):
Thanks for the enhancement, I put some minor comments.
Others look good to me.

Thanks.

monai/apps/mmars/mmars.py
```diff
@@ -126,27 +138,65 @@ def load_from_mmar(item, mmar_dir=None, progress: bool = True, map_location=None
...
     model_dict = torch.load(model_file, map_location=map_location)
     if weights_only:
-        return model_dict["model"]
-
-    # TODO: search for the module based on model name?
-    if not model_dict.get("train_conf", ""):
-        raise ValueError("The MMAR configuration does not have a 'train_conf' section.")
-    model_config = model_dict["train_conf"]["train"]["model"]
-    if model_config.get("name", ""):  # model config section is a "name"
+        return model_dict.get("model", model_dict)  # model_dict["model"] or model_dict directly
```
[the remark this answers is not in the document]
wyli: quite a few MMARs have this dictionary in `model.pt`:
```
{
"model": torch_module_parameters,
"train_conf": {"model": {"name": "resnet"}}
}
```
'model' is used to represent two different concepts here, one is the parameters, the other is the config.
here I'd like to search for `{"model": {"name": "resnet"}}`, but not `"model": torch_module_parameters`.

I'm not familiar with the relevant config, is there any rules for defining those keys that I can check here?

monai/apps/mmars/mmars.py
wyli: if the MMAR folder is on the `PYTHONPATH`, we'll be able to load it, or did I miss something here?

monai/apps/mmars/mmars.py
wyli: this is the last guess for the model config location, in case that `model_dict` doesn't have a `train_conf` section, but have something similar, for example a `training_conf` section.

monai/apps/mmars/mmars.py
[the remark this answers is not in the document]
Nic-Ma: Thanks for the explanation, I see your point.
Actually, both cases are hard-code to `model` in clara now, I feel you can use `model_key` for both cases.
We may change the naming in future if want to support GAN, etc.

Thanks.

monai/apps/mmars/mmars.py
Nic-Ma: OK, I see.
Clara doesn't support this case so far. And if it is, how can I specify the `training_conf` key in this case?

Thanks.

wyli: the `_get_val` is a search in the dictionary for `model_key`, the idea is to not assuming the `train_conf` or `training_conf`

Nic-Ma: OK, that makes sense!
Thanks.

monai/apps/mmars/mmars.py
Nic-Ma: It supposes that the `custom` folder is in your PYTHONPATH, when you try to run the MMAR in Clara docker, it will automatically add `custom` to PYTHONPATH.
see the `set_env.sh` in:
https://ngc.nvidia.com/catalog/models/nvidia:med:clara_pt_deepgrow_3d_annotation/files
So I think MONAI users just download the MMAR locally(don't run the MMAR) and try to build a model instance, they don't have `custom` in the PYTHONPATH.

Thanks.

wyli: sure, let me update the error message

monai/apps/mmars/mmars.py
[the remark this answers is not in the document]
wyli: ok I'll update

Nic-Ma (approved):
Thanks for the quick update.
Looks good to me.

PR #2423, opened 2021-06-23
2422 Enhance the pixdim of spacing transform

Nic-Ma:
Fixes #2422 .

### Description
This PR added support to use original pixdim in `Spacing` transform, also fixed a minor issue when using 1 single float number for all the dims. According to the test case:
https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_spacing.py#L151
it should expand the float number to all dims, but we actually used 1.0 or the original pixdim to pad for this case:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/spatial/array.py#L174

### Status
**Ready**

### Types of changes


monai/transforms/spatial/array.py
```diff
@@ -91,7 +91,10 @@ def __init__(
     ) -> None:
         """
         Args:
-            pixdim: output voxel spacing.
+            pixdim: output voxel spacing. if providng a single number, will use it for all dimensions.
+                if the components of the `pixdim` are non-positive values, the transform will use the
+                corresponding components of the origial pixdim, which is computed from the `affine`
+                matrix of input image.
```
yiheng-wang-nv, line 97: In old version, if `pixdim` is a sequence, it can has larger length than `sr`, but now we have the restriction: `out_d = ensure_tuple_rep(self.pixdim, sr)`. Therefore, I suggest here to add more docstring about the length of it. Like:
If providing a sequence, please ensure that its length equals to the spatial size of input image.

monai/transforms/spatial/array.py
```diff
@@ -91,7 +91,10 @@ def __init__(
     ) -> None:
         """
         Args:
-            pixdim: output voxel spacing.
+            pixdim: output voxel spacing. if providng a single number, will use it for all dimensions.
```
yiheng-wang-nv, line 94: providing

monai/data/utils.py
```diff
@@ -517,13 +519,14 @@ def zoom_affine(affine: np.ndarray, scale: Sequence[float], diagonal: bool = Tru
...
-    if np.any(scale_np <= 0):
-        raise ValueError("scale must contain only positive numbers.")
+
     d = len(affine) - 1
+    # compute original pixdim
+    norm = np.sqrt(np.sum(np.square(affine), 0))[:-1]
     if len(scale_np) < d:  # defaults based on affine
-        norm = np.sqrt(np.sum(np.square(affine), 0))[:-1]
         scale_np = np.append(scale_np, norm[len(scale_np) :])
-    scale_np = scale_np[:d]
```
yiheng-wang-nv, line 526: If we remove this line, when `len(scale_np) > d`, error happens in `fall_back_tuple(scale_np, norm)`

monai/transforms/spatial/array.py
Nic-Ma: sure, updated in the latest commit.
Thanks.

monai/transforms/spatial/array.py
Nic-Ma: updated in the latest commit.
Thanks.

monai/data/utils.py
Nic-Ma: Thanks for pointing this, updated in the latest commit.

Nic-Ma: And BTW, as we require `length of pixdim = input size`, there should be no case that `len(scale_np) > d` anymore.

Thanks.

yiheng-wang-nv (approved):

monai/transforms/spatial/array.py
```diff
@@ -173,11 +179,11 @@ def __call__(
             affine_ = np.eye(sr + 1, dtype=np.float64)
         else:
             affine_ = to_affine_nd(sr, affine)
-        out_d = self.pixdim[:sr]
-        if out_d.size < sr:
-            out_d = np.append(out_d, [1.0] * (out_d.size - sr))
-        if np.any(out_d <= 0):
-            raise ValueError(f"pixdim must be positive, got {out_d}.")
+        if not issequenceiterable(self.pixdim):
+            out_d = ensure_tuple_rep(self.pixdim, sr)
```
wyli, line 183: by default we should assume pixdim 1.0 instead of `ensure_tuple_rep`

Nic-Ma: So I think you mean we should treat `pixdim=0.5` and `pixdim=[0.5]` as the same? And if applying on a 3D data, it should map to `pixdim=[0.5, 1.0, 1.0]` instead of `pixdim=[0.5, 0.5, 0.5]` or `pixdim=[0.5, <image pixdim[1]>, <image pixdim[2]>]`?

Thanks.

Nic-Ma: Hi @wyli , could you please help confirm which the expected behavior is? Then I will update the PR ASAP.

Thanks.

Nic-Ma: updated in the latest commit.
Thanks.

Issue #2426, opened 2021-06-23
Multiple transform return values are not properly passed as parameters for next transform


Spenhouet:
**Describe the bug**

`transform.apply_transform` does currently not handle the case where the return value of a transform call includes multiple return values.

```
if isinstance(data, (list, tuple)) and map_items:
    return [transform(item) for item in data]
```

It just passes the return values of the last transform as tuple.

**To Reproduce**
1. Compose the `LoadImage` and `Spacing` transforms
2. `LoadImage` will return the array and meta_data
3. `Spacing` expects both parameters but will fail

**Expected behavior**
The parameters are unpacked:

```
if isinstance(data, (list, tuple)) and map_items:
    return [transform(*items) for items in data]
```

It might be necessary to either always return a list from all transforms or to make a check here to not unpack something that should not be unpacked.
The compose could accept a `unpack` boolean flag which is provided to the `apply_transform` function.

**Environment**

```
================================
Printing MONAI config...
================================
MONAI version: 0.5.2
Numpy version: 1.20.1
Pytorch version: 1.6.0+cu101
MONAI flags: HAS_EXT = False, USE_COMPILED = False
MONAI rev id: feb3a334b7bbf302b13a6da80e0b022a4cf75a4e

Optional dependencies:
Pytorch Ignite version: NOT INSTALLED or UNKNOWN VERSION.
Nibabel version: 3.2.1
scikit-image version: 0.17.2
Pillow version: 8.1.0
Tensorboard version: 2.2.2
gdown version: NOT INSTALLED or UNKNOWN VERSION.
TorchVision version: NOT INSTALLED or UNKNOWN VERSION.
ITK version: NOT INSTALLED or UNKNOWN VERSION.
tqdm version: 4.61.1
lmdb version: NOT INSTALLED or UNKNOWN VERSION.
psutil version: NOT INSTALLED or UNKNOWN VERSION.

For details about installing the optional dependencies, please visit:
    https://docs.monai.io/en/latest/installation.html#installing-the-recommended-dependencies


================================
Printing system config...
================================
`psutil` required for `print_system_info`

================================
Printing GPU config...
================================
Num GPUs: 1
Has CUDA: True
CUDA version: 10.1
cuDNN enabled: True
cuDNN version: 7604
Current device: 0
Library compiled for CUDA architectures: ['sm_37', 'sm_50', 'sm_60', 'sm_61', 'sm_70', 'sm_75', 'compute_37']
[...]

PR #2437, opened 2021-06-24
Add parameter unpacking for transforms

Spenhouet:
Signed-off-by: Sebastian Penhouet <sebastian.penhouet@airamed.de>

Fixes https://github.com/Project-MONAI/MONAI/issues/2426

### Description
Transforms can return any number of values. This pull request enables that these return values are passed into the next transform in a Compose as parameters (unpacked arguments).
This is especially useful in the case where the input and output of transforms always matches like in the case of always passing along the image array and the meta data dict.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -1031,7 +1031,7 @@ class NPZDictItemDataset(Dataset):
         self,
         npzfile: Union[str, IO],
         keys: Dict[str, str],
-        transform: Optional[Callable] = None,
+        transform: Optional[Callable[..., Dict[str, Any]]] = None,
         other_keys: Optional[Sequence[str]] = (),
     ):
         self.npzfile: Union[str, IO] = npzfile if isinstance(npzfile, str) else "STREAM"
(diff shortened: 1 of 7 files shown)


Spenhouet:
mypy issues:
```
monai/engines/utils.py:144:16: error: Incompatible types in assignment (expression has type "Union[List[Dict[Any, Any]], Dict[Any, Any]]", variable has type "Dict[Any, Any]")  [assignment]
monai/data/image_dataset.py:127:38: error: "None" object is not iterable  [misc]
monai/data/dataset.py:1062:20: error: Incompatible types in assignment (expression has type "Union[List[Dict[str, Any]], Dict[str, Any]]", variable has type "Dict[str, Any]")  [assignment]
```

Strange that I did not get them on my local development. I will look into it.

Nic-Ma:
Hi @Spenhouet ,

Thanks for your quick PR,  I think `mypy` usually has big difference between versions, please update your local mypy to `0.902` and try again.

Thanks.

Spenhouet:
@wyli @Nic-Ma I fixed the mypy complains (and also the usage in `ImageDataset`).

wyli (approved):
thanks!

Spenhouet:
Btw. I did set the `unpack` flag to `False` by default to not break any current behavior.

monai/data/dataset.py
```diff
@@ -1058,10 +1058,15 @@ def __len__(self):
...
-        if self.transform is not None:
-            data = apply_transform(self.transform, data)
+        if not self.transform:
+            return data
 
-        return data
+        transformed_data = apply_transform(self.transform, data)
+
+        if not isinstance(transformed_data, dict):
+            raise AssertionError("With a dict supplied to apply_transform a single dict return is expected.")
```
Nic-Ma, line 1067: Hi @Spenhouet ,

Thanks very much for your quick PR!
I think this assert is not correct here, some MONAI sampling transforms return a list of dictionaries instead of a single dict.
I will submit a quick PR to fix it soon, CC @wyli @ericspod @rijobro .
Others of this PR look good to me.

Thanks.

PR #2441, opened 2021-06-24
no need for default orig_size stored for invertible transforms

rijobro:
### Description
If an object didn't have the attribute `shape`, we stored `orig_size` as `None`. However, `None` can't be collated into a batch, so if the object doesn't have the attribute `shape`, then don't store that information.

Fixes https://github.com/Project-MONAI/MONAI/discussions/2414.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/inverse.py
+++ b/monai/transforms/inverse.py
@@ -76,7 +76,7 @@ class InvertibleTransform(Transform):
         info = {
             InverseKeys.CLASS_NAME: self.__class__.__name__,
             InverseKeys.ID: id(self),
-            InverseKeys.ORIG_SIZE: orig_size or (data[key].shape[1:] if hasattr(data[key], "shape") else None),
+            InverseKeys.ORIG_SIZE: orig_size or (data[key].shape[1:] if hasattr(data[key], "shape") else -1),
         }
         if extra_info is not None:
             info[InverseKeys.EXTRA_INFO] = extra_info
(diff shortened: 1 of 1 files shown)


monai/transforms/inverse.py
```diff
@@ -76,8 +76,11 @@ def push_transform(
         info = {
             InverseKeys.CLASS_NAME: self.__class__.__name__,
             InverseKeys.ID: id(self),
-            InverseKeys.ORIG_SIZE: orig_size or (data[key].shape[1:] if hasattr(data[key], "shape") else None),
         }
+        if orig_size is not None:
+            InverseKeys.ORIG_SIZE = orig_size
+        elif hasattr(data[key], "shape"):
+            InverseKeys.ORIG_SIZE = data[key].shape[1:]
```
wyli, line 83: missing the `else` here for the -1?

rijobro: sorry, i updated the docs. There's no reason to have a default (None or -1). If the object doesn't have `shape` then none of that batch will either. So no need to add anything. If shape doesn't exist in the forward direction, the orig_size won't be used in the inverse direction, either.

wyli (approved):
thanks!

monai/transforms/inverse.py
rijobro: ...or that's what I thought until the CI/CD failed...

PR #2444, opened 2021-06-25
2426 Enhance the assert of transform results

### LINKED ISSUE #2426: Multiple transform return values are not properly passed as parameters for next transform

Nic-Ma:
Fixes #2426 .

### Description
This PR is a followup of #2437 , it enhanced the assert of transform results to support the sampling transforms, like:
`RandSpatialCropSamplesD`, `RandCropByPosNegLabelD`, etc.

### Status
**Ready**

### Types of changes


monai/data/dataset.py
```diff
@@ -1061,12 +1061,11 @@ def _transform(self, index: int):
...
             return data
 
-        transformed_data = apply_transform(self.transform, data)
+        result = apply_transform(self.transform, data)
 
-        if not isinstance(transformed_data, dict):
-            raise AssertionError("With a dict supplied to apply_transform a single dict return is expected.")
-
-        return transformed_data
+        if isinstance(result, dict) or (isinstance(result, list) and isinstance(result[0], dict)):
```
yiheng-wang-nv, line 1066: This line can be simplified into:
```
if isinstance(result, (dict, list)) and isinstance(result[0], dict)):
```

Nic-Ma: Hi @yiheng-wang-nv ,

Thanks for your review, I think your suggestion may not be correct if the `result` is a dict, result[0] is not dict.

yiheng-wang-nv: Oh, sorry you're right.

yiheng-wang-nv (approved):

PR #2475, opened 2021-06-29
2474 Deprecate SegmentationSaver and TransformInverter handlers

### LINKED ISSUE #2474: Mark SegmentationSaver and TransformInverter handlers as `deprecated`
As the `SaveImaged` and `Invertd` transforms are more flexible than the 2 handlers, and the output of the 2 handlers can't align with the latest dataflow design(combine `batch` and `output`). I would suggest to deprecate the handlers in MONAI v0.7.

Nic-Ma:
Fixes #2474 .

### Description
This PR added `deprecated` decorator to `SegmentationSaver` and `TransformInverter` handlers as we already have better options.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @ericspod ,

Thanks for developing the `deprecated` API, I tried to use it in this PR.
But I got a minor issue that I can't see the warning message when call the deprecated function in **multi-processing** case. Something wrong with my env settings?
Maybe you can help verify locally?
You can download this PR and just run below program in a PyTorch distributed program:
```py
from monai.handlers.utils import evenly_divisible_all_gather
        print(evenly_divisible_all_gather(["tests"]))
```
It can print out the warning message if running in a regular single process program.

Thanks.

ericspod:
I think you're using incorrect arguments unless the ones you provided were for testing. The `version_val` argument is for the current version of MONAI, you'd provide a value to this during testing so that comparisons can be made against a chosen version that doesn't change. Leaving that argument alone will use the current MONAI version which is what you want I think. The `removed` argument is for specifying which version the definition will be removed in, that is what I think you want to be "0.7.0". 

Also for the `msg_suffix` value, this should be a full correct English sentence since it will be appended after a full stop. Unless the first word is a proper name/term that doesn't use normal capitalization rules, the sentence should begin with a capital letter.

For the multiprocessing this relates to the mechanism I added to ensure the warning is show only once. A global dictionary stores definitions that were called/instantiated and have printed the warning,  on subsequent calls/instantiations the warning is skipped. This won't necessarily work across processes if this dictionary wasn't filled in before spawning subprocesses, so each process will start with an empty dictionary and each should issue a warning. This will always happen in Windows which lacks fork semantics. If the stdout/stderr for subprocesses is suppressed or routed elsewhere you will never see the warning. 

However looking at the tests, for `evenly_divisible_all_gather` for example, I don't see how this would impact the lack of message. I think it might be a combination of factors between the value for the `version_val`, multiprocessing, and the fact that if you test a function multiple times you'd only get the warning on the first test so subsequent ones will fail if checking a warning was printed. The `version_val` argument can only be set when the decorator is applied which means when you try to test decorated functions the result you get will change as the version of MONAI increments. Perhaps these should changed to expect an exception raise when the version of MONAI is past the `removed` version (ie. anything at or above 0.7.0).

Nic-Ma:
Hi @ericspod ,

Thanks for your detailed explanation! May I ask what the `since` version I should set in the dev branch now?
Actually, it will only be marked as `deprecated` since v0.6.0, but as we haven't released v0.6.0, if I use 0.6.0 now, it will raise an exception instead of a deprecated warning.

Thanks.

ericspod:
If you use a `since` value in the future it should do nothing. I think I might have missed a test for this case however.

Nic-Ma:
> If you use a `since` value in the future it should do nothing. I think I might have missed a test for this case however.

I have updated the PR according to your comments to use the `removed` arg and capital word.
What do you think we should do next? You will update the `since` or something?

Thanks.

wyli:
Just a reminder that I pushed the tag `0.6.0.rc1` to the repo. So if you fetch the tags to your local copy, the dev version will be `0.6.0.rc1+commits` instead of the previous `0.5.3+commits`, this might change the deprecation message results

Nic-Ma:
Hi @ericspod ,

Thanks for your suggestions! Maybe I misunderstood your points. So is it correct in the PR now?
```py
@deprecated(since="0.6.0", removed="0.7.0", msg_suffix="XXX")
```

Nic-Ma:
Seems the CI test failed due to below error:
```
  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/monai/networks/nets/torchvision_fc.py", line 71, in <module>
    @deprecated(since="0.6.0", removed="0.7.0", msg_suffix="Please consider using `TorchVisionFCModel` instead.")
  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/monai/utils/deprecated.py", line 87, in deprecated
    is_deprecated = since is not None and version_leq(since, version_val)
  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/monai/utils/deprecated.py", line 61, in version_leq
    return l < r
TypeError: '<' not supported between instances of 'int' and 'str'
```

Thanks.

ericspod:
> Hi @ericspod ,
> 
> Thanks for your suggestions! Maybe I misunderstood your points. So is it correct in the PR now?
> 
> ```python
> @deprecated(since="0.6.0", removed="0.7.0", msg_suffix="XXX")
> ```

Yes that should be right.

ericspod:
> Seems the CI test failed due to below error:
> 
> ```
>   File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/monai/networks/nets/torchvision_fc.py", line 71, in <module>
>     @deprecated(since="0.6.0", removed="0.7.0", msg_suffix="Please consider using `TorchVisionFCModel` instead.")
>   File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/monai/utils/deprecated.py", line 87, in deprecated
>     is_deprecated = since is not None and version_leq(since, version_val)
>   File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/monai/utils/deprecated.py", line 61, in version_leq
>     return l < r
> TypeError: '<' not supported between instances of 'int' and 'str'
> ```
> 
> Thanks.

It doesn't like versions of the form "0.6.0rc0" because "0rc0" part won't parse. I had assumed that versions would always be three  numbers separated by "." followed optionally by "." and other text that's ignored. A version of the form "0.6.0.rc0" would work otherwise we can figure out how to split the version string by "rc#" if it occurs.

Nic-Ma:
Hi @ericspod ,

Maybe `pkg_resource.parse_version` can support it?
https://github.com/Project-MONAI/MONAI/blob/dev/tests/utils.py#L117

Thanks.

ericspod:
> Hi @ericspod ,
> 
> Maybe `pkg_resource.parse_version` can support it?
> https://github.com/Project-MONAI/MONAI/blob/dev/tests/utils.py#L117
> 
> Thanks.

I hadn't want to rely on `pkg_resource` since it isn't a base dependency. I'm working on a fix right now though and I'l have a PR shortly.

Nic-Ma:
Cool! Thanks for your quick update!
CC @wyli .

Thanks.

monai/networks/nets/torchvision_fc.py
```diff
@@ -68,7 +68,7 @@ def __init__(
         )
 
 
-@deprecated(since="0.6.0", version_val="0.7.0", msg_suffix="please consider using `TorchVisionFCModel` instead.")
```
wyli, line 71: this mistake actually makes the dev branch ci/cd work fine...

Nic-Ma:
Hi @ericspod ,

Cool! Thanks for your quick update.
I merged master, could you please help review this PR again?

Thanks.

ericspod (approved):
I'm good with it if @wyli is too.

wyli (approved):
thanks, perhaps you could also add some docstring for things that were not clear to you?

Nic-Ma:
Sure, enhanced the doc-strings to make it more clear.

Thanks.

Nic-Ma:
Hi @ericspod @wyli ,

Do you know why the GPU tests failed? I tried to run several times.

Thanks in advance.

wyli:
looks like it's related to version comparison, I'm looking into this

Nic-Ma:
> looks like it's related to version comparison, I'm looking into this

Can your PR https://github.com/Project-MONAI/MONAI/pull/2485 fix it?
If yes, we can wait until merging your PR first.

Thanks.

wyli:
> > looks like it's related to version comparison, I'm looking into this
> 
> Can your PR #2485 fix it?
> If yes, we can wait until merging your PR first.
> 
> Thanks.

yes, tested #2485, it is good now. I didn't address https://github.com/Project-MONAI/MONAI/pull/2485#discussion_r661007654 but I think it's good enough for v0.6..

PR #2482, opened 2021-06-29
Deprecated fix

### LINKED ISSUE #2479: enhance the version comparison utililty
version comparison is needed for a few cases:
- https://github.com/Project-MONAI/MONAI/blob/2085a49820be146ad953ece23041540a237e08b6/tests/utils.py#L133-L136
- https://github.com/Project-MONAI/MONAI/blob/2085a49820be146ad953ece23041540a237e08b6/monai/utils/module.py#L276-L279
- https://github.com/Project-MONAI/MONAI/blob/2085a49820be146ad953ece23041540a237e08b6/monai/utils/deprecated.py#L55

ericspod:

Fixes #2479 .

### Description
This fixes an issue with version strings and replaces some functionality reliant on an optional package with an internal solution. Version strings should contain integers separated by ".", any component not containing an integer is broken into an initial integer component and the rest. The integers from this are compared when comparing versions, so "0.6.0rc0" will be treated the same as "0.6.0". Git version suffixes are ignored. 

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/utils/deprecated.py
+++ b/monai/utils/deprecated.py
@@ -48,7 +48,7 @@ def version_leq(lhs, rhs):
             m = re.match("(\d+)(.*)", val)
             if m is not None:
                 val = m.groups()[0]
-                
+
             return int(val)
         except ValueError:
             return val
@@ -92,7 +92,7 @@ def deprecated(
     is_deprecated = since is not None and version_leq(since, version_val)
     is_removed = removed is not None and version_leq(removed, version_val)
     is_not_yet_deprecated = since is not None and version_val != since and version_leq(version_val, since)
-    
+
     if is_not_yet_deprecated:
         return lambda obj: obj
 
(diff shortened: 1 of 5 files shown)


monai/utils/deprecated.py
```diff
@@ -44,6 +45,10 @@ def version_leq(lhs, rhs):
     def _try_cast(val):
         val = val.strip()
         try:
+            m = re.match("(\\d+)(.*)", val)
```
wyli, line 48: we should prefer `optional_import("pkg_resources", name="parse_version")` if it is installed... this `version_leq` will be used in many cases, better to make it as robust as possible

wyli (approved):
thanks! if you don't have time to add `optional_import("pkg_resources", name="parse_version")`, I'll do it later today

ericspod:
This should sort the optional dependency on pkg_resources now

PR #2485, opened 2021-06-29
2479 - enhance version compare

wyli:

### Description
- update to prefer pkg_resources if it's installed
- move `version_leq` to utils.module
- fixes an issue the following code doesn't throw errors:
```py
from monai.utils.deprecated import deprecated
deprecated(since="0.6", removed="0.8", version_val="0.9")(lambda x: x)(1)
```
- add tests

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/utils/deprecated.py
+++ b/monai/utils/deprecated.py
@@ -21,8 +21,6 @@ from .. import __version__
 
 __all__ = ["deprecated", "deprecated_arg", "DeprecatedError"]
 
-warned_set = set()
-
 
 class DeprecatedError(Exception):
     pass
@@ -30,13 +28,9 @@ class DeprecatedError(Exception):
 
 def warn_deprecated(obj, msg):
     """
-    Issue the warning message `msg` only once per process for the given object `obj`. When this function is called
-    and `obj` is not in `warned_set`, it is added and the warning issued, if it's already present nothing happens.
+    Issue the warning message `msg`.
     """
-    global warned_set
-    if obj not in warned_set:  # ensure warning is issued only once per process
-        warned_set.add(obj)
-        warnings.warn(msg, category=DeprecationWarning, stacklevel=2)
+    warnings.warn(msg, category=DeprecationWarning, stacklevel=2)
 
 
 def deprecated(
--- a/monai/utils/deprecated.py
+++ b/monai/utils/deprecated.py
@@ -23,7 +23,6 @@ from .. import __version__
 __all__ = ["deprecated", "deprecated_arg", "DeprecatedError"]
 
 warned_set = set()
-warned_lock = Lock()
 
 
 class DeprecatedError(Exception):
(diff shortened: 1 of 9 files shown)


monai/utils/deprecated.py
```diff
@@ -10,19 +10,19 @@
...
 from typing import Optional
 
+from monai.utils.module import version_leq
+
 from .. import __version__
 
 __all__ = ["deprecated", "deprecated_arg", "DeprecatedError"]
 
 warned_set = set()
-warned_lock = Lock()
```
wyli, line 25: looks like this variable was not in use? @ericspod

monai/utils/deprecated.py
```diff
@@ -34,40 +34,12 @@ def warn_deprecated(obj, msg):
     Issue the warning message `msg` only once per process for the given object `obj`. When this function is called
     and `obj` is not in `warned_set`, it is added and the warning issued, if it's already present nothing happens.
     """
+    global warned_set
```
wyli, line 37: I try to add a `global` here but not sure what would be the intended behaviour? @ericspod

monai/utils/deprecated.py
ericspod: No those were used by `warn_deprecated` to ensure warnings were issued only once.

Nic-Ma: Hi @ericspod , maybe it's OK to always show the warning message when calling the API?
We can refer to other repos about handling this case.

Thanks.

ericspod: The issue is if you used a deprecated function in a loop or in a transform pipeline, it will get repeated a lot. We see this with warnings from Numpy or Scipy and it can be annoying.

wyli: true, not sure if it could be addressed by filtering https://docs.python.org/3/library/warnings.html#the-warnings-filter

ericspod: It would require somehow reacting to when a warning comes up for the first time and filtering then, but you need some mechanism to detect that.

tests/test_version_leq.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+# from pkg_resources
+torture = """
+    0.80.1-3 0.80.1-2 0.80.1-1 0.79.9999+0.80.0pre4-1
+    0.79.9999+0.80.0pre2-3 0.79.9999+0.80.0pre2-2
+    0.77.2-1 0.77.1-1 0.77.0-1
+    """
+
+TEST_CASES = (
```
Nic-Ma, line 36: Great test cases. I think it covers all the cases we have in MONAI so far, @ericspod what do you think?

Thanks.

Nic-Ma (approved):
Looks good to me.
@ericspod could you please help double confirm it?

Thanks.

tests/test_version_leq.py
ericspod: > 
> 
> Looks good to me.
> @ericspod could you please help double confirm it?
> 
> Thanks.

It's fine by me, except note that this removes the code for preventing the duplicate warnings.

PR #2487, opened 2021-06-30
2467 Add support to save checkpoint with smaller metric

### LINKED ISSUE #2467: Inflexible score function of CheckpointSaver
https://github.com/Project-MONAI/MONAI/blob/9f9b63a7dc56d8b31d4fc8d34aae77865ba95947/monai/handlers/checkpoint_saver.py#L148-L160

Currently the `CheckpointSaver` saves the check point that has the largest score value. However, this is not desirable when a lower score indicates better performance, e.g. validation loss. We can either allow the user to pass a custom score function (as in the ignite.handlers), or add a flag to control returning `round(engine.state.metrics[metric_name], 4)` or ` - round(engine.state.metrics[metric_name], 4)`

Nic-Ma:
Fixes #2467 .

### Description
This PR added support to save checkpoint with smaller metric value, because some error-like metric is `smaller is better`.

### Status
**Ready**

### Types of changes


monai/engines/evaluator.py
```diff
@@ -53,6 +53,8 @@ class Evaluator(Workflow):
             engine.state.metrics when epoch completed. key_val_metric is the main metric to compare and save the
             checkpoint into files.
         additional_metrics: more Ignite metrics that also attach to Ignite Engine.
+        metric_cmp_fn: function to compare current key metric with previous best key metric value.
+            it must accept 2 args (current_metric, previous_best) and return a bool result. default to `greater than`.
```
vfdev-5, line 57:
```suggestion
            It must accept 2 args (current_metric, previous_best) and return a bool result. Default to `greater than`.
```
Same for others below. 

 I think the docstring is missing the explanation on what to return and how it will taken into account by the code. For example, if returns True, what will happen.

Nic-Ma:
integration test `tests/test_integration_workflows.py` passed locally.

Thanks.

monai/engines/evaluator.py
Nic-Ma: Hi @vfdev-5 , thanks for your review! I updated according to your comments.

monai/handlers/checkpoint_saver.py
```diff
@@ -62,6 +62,8 @@ class CheckpointSaver:
             typically, it's used to resume training and compare current metric with previous N values.
         key_metric_greater_or_equal: if `True`, the latest equally scored model is stored. Otherwise,
             save the the first equally scored model. default to `False`.
+        key_metric_score_sign: sign of the score: `1.0` or `-1.0`. For error-like metrics, e.g. smaller is better,
```
wyli, line 65: This could be an int or Boolean flag

wyli (approved):

monai/handlers/checkpoint_saver.py
Nic-Ma: Thanks for your review, updated doc-string in the latest commit.

monai/handlers/checkpoint_saver.py
```diff
@@ -62,6 +62,9 @@ class CheckpointSaver:
             typically, it's used to resume training and compare current metric with previous N values.
         key_metric_greater_or_equal: if `True`, the latest equally scored model is stored. Otherwise,
             save the the first equally scored model. default to `False`.
+        key_metric_score_sign: sign of the score, available value: `[1.0, -1.0, 1, -1]`. For error-like metrics,
```
wyli, line 65: I think this should be a flag indicating whether to flip the sign of the metric value, if it is true or positive, the sign of the metric value won't be flipped, if it's false or negative, the metric value will be flipped. It's not about supporting 1.0 -1.0 1 -1 here.

Nic-Ma: Hi @wyli ,

Thanks for your explanation. I tried to align with ignite API: https://github.com/pytorch/ignite/blob/master/ignite/handlers/checkpoint.py#L585
I am OK with both solutions, your idea seems more robust, no need to check the arg value.

Thanks.

Nic-Ma: Let me tried to update it to bool flag in this ignite related PR together: https://github.com/Project-MONAI/MONAI/pull/2492

Thanks.

wyli: That docstring seems to suggest that it'll ensure a positive or negative value, but the actual implemention is about 'flip or not'. Perhaps the ignite docstring needs an update as well...

Nic-Ma: Hi @vfdev-5 , what do you think about this arg? Maybe ignite can also change it to be a bool flag?

Thanks.

Nic-Ma: @wyli , updated in the latest commit.
Thanks.

vfdev-5: > what do you think about this arg? Maybe ignite can also change it to be a bool flag?

Hi @Nic-Ma , this is a good question. Let me see with the team about that.

PR #2492, opened 2021-07-01
2468 Add ignite dist to support more backends

### LINKED ISSUE #2468: Support more backends in distributed operations
The current version of MONAI utilizes `torch.distributed` for `all_gather`, `all_reduce` etc., e.g. in **line 409** below. However, this will raise an Error when I launch the parallel programs in `ignite.distributed.Parallel` context with `horovod` or `xla` backend: `RuntimeError: Default process group has not been initialized, please make sure to call init_process_group.` Is it better to replace all the `torch.distributed` operations with `ignite.distributed` ones? They have naive support for all the backends.

https://github.com/Project-MONAI/MONAI/blob/1a707f0226bfe10c000cc54ac87b35d50973e221/monai/utils/misc.py#L359-L423

Nic-Ma:
Fixes #2468 .

### Description
This PR added dist implementation based on ignite APIs to support more kinds of backends.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @wyli ,

As we don't have Horovod, XLA, etc. in the dist CI tests, I locally tested both PyTorch and ignite implementation based on the native GLOO and NCCL backends.

Thanks.

monai/handlers/checkpoint_saver.py
```diff
@@ -62,9 +62,9 @@ class CheckpointSaver:
             typically, it's used to resume training and compare current metric with previous N values.
         key_metric_greater_or_equal: if `True`, the latest equally scored model is stored. Otherwise,
             save the the first equally scored model. default to `False`.
-        key_metric_score_sign: sign of the score, available value: `[1.0, -1.0, 1, -1]`. For error-like metrics,
-            e.g. smaller is better, a negative score sign should be used (objects with larger score are retained).
-            default to `1.0`.
+        key_metric_negative_sign: whether adding a negative sign to the metric score to compare metrics,
```
vfdev-5, line 65: @Nic-Ma I think this change is from another PR ?

Nic-Ma: Hi @vfdev-5 , yes, it's to address this comment: https://github.com/Project-MONAI/MONAI/pull/2487#discussion_r662082802.

Thanks.

monai/utils/dist.py
```diff
@@ -0,0 +1,151 @@
...
+            # if all the input data doesn't have batch dim, unbind to a list of 0-dim Tensors
+            return list(torch.unbind(output, dim=0))
+        return [output[i * max_len : i * max_len + l, ...] for i, l in enumerate(all_lens)]
+
+    output: List[torch.Tensor]
+    if dist.is_available() and dist.is_initialized():
+        if dist.get_world_size() <= 1:
+            return data
+        output = _torch_all_gather(data=data)
+    elif has_ignite:
```
wyli, line 116: I think we should prefer ignite if it's installed otherwise fall back to torch native?

Nic-Ma: Hi @wyli , thanks for your suggestion.
The issue is that: there are some cases that if you don't init your dist program with ignite APIs, the ignite dist calls will not work, for example: https://github.com/pytorch/ignite/issues/2048#issuecomment-858331410.
But PyTorch dist APIs can work if initialized with ignite APIs.
So I think we should use PyTorch dist APIs by default, if users want to use ignite implementation, they should totally use ignite dist APIs in their program.
What do you think?

Thanks.

wyli: I'm not sure about the idea here. @vfdev-5 could you please help confirm it?

vfdev-5: Hi @Nic-Ma @wyli , the issue pytorch/ignite#2048 revealed an issue we have in the way how ignite DiskSaver class interacts with distributed group created by user using pytorch in a single rank. In few words the issue is the following:
```python
dist.init_process_group(...)

# ignite is not aware of created DDP group

if dist.get_rank() == 0:
    # - Here only 1 process passes
    
    # - Now user executes something that calls ignite.distributed the first time 
    # where ignite internally calls a collective op => execution hangs
    
    # - To avoid that we propose to call `idist.sync()` right after `dist.init_process_group(...)`

dist.destroy_process_group()
```

> The issue is that: there are some cases that if you don't init your dist program with ignite APIs, the ignite dist calls will not work

If this method is called by all processes there wont be any issue. And according to the name `evenly_divisible_all_gather` it will be called by all DDP processes...

Nic-Ma: Hi @vfdev-5 , thanks for your explanation!
So I think you mean we can change as @wyli suggested? Prefer ignite, if not installed, fall back to torch native?

Thanks.

vfdev-5: For me both variants should work here and I have no particular preference for any. But, yes, we can change as @wyli suggested.

Thanks

Nic-Ma: Hi @wyli , I have changed the PR to prefer to ignite implementation.

Thanks in advance.

wyli (approved):
thanks

PR #2508, opened 2021-07-02
2506 Fix strict_shape in CheckointLoader

### LINKED ISSUE #2506: A potential bug in CheckpointLoader
There might a potential bug in following lines:

https://github.com/Project-MONAI/MONAI/blob/ce45640b2aab6e094e3c8af9ea49634497df9fb6/monai/handlers/checkpoint_loader.py#L95-L104

If we 
1. set `CheckpointLoader.strict_shape = False` 
2. load other objects besides model's `state_dict` from checkpoint (e.g. when `load_dict = {"model": model, "optimizer": optimizer}`),

the `obj` might be `model` or `optimizer` in Line 103, while `checkpoint` is always a dict containing keys `"model"` and `"optimizer"`. If the obj is `optimizer`, then [this line](https://github.com/Project-MONAI/MONAI/blob/ce45640b2aab6e094e3c8af9ea49634497df9fb6/monai/networks/utils.py#L370) will raise an Error since `optimizer` is not iterable. If the obj is `model`, `{dst_prefix}{s}`([see here](https://github.com/Project-MONAI/MONAI/blob/ce45640b2aab6e094e3c8af9ea49634497df9fb6/monai/networks/utils.py#L377)) will never be matched, since `s` is just `"model"` or `"optimizer"`. Thus, the `state_dict` from check point will never be loaded.

Although the conditions to trigger this error is barely fulfilled in practice (e.g. when we do transfer learning and the checkpoint happens to contain some non-iterable objects like tensors, optimizer, etc.), it is better to be aware of this issue.

You can run the following test script to see the detailed error message:
```python3
from ignite.engine import Engine
from torch.optim import SGD
import torch
import torch.nn as nn
import tempfile
from monai.handlers import CheckpointLoader
import os.path as osp


def main(output_dir):
	# build dummy data set and data loader
	x = torch.randn(10, 2)
	y = torch.randn(10)
	train_set = torch.utils.data.TensorDataset(x, y)
	train_loader = torch.utils.data.DataLoader(
		train_set, 
		batch_size=2, 
		num_workers=0, 
		shuffle=False)

	# build dummy model, optimizer and engine
	model = nn.Linear(2, 1, bias=False)
	optimizer = SGD(model.parameters(), lr=0.001)

	def update_model(engine, batch):
		x_, y_ = batch
		optimizer.zero_grad()
		loss = (model(x_) - y_).sum()
		loss.backward()
		optimizer.step()
		return loss.item()

	trainer = Engine(update_model)

	# save a checkpoint for loading
	to_save = {'model': model.state_dict(), 'optimizer': optimizer.state_dict()}
	to_load = {'model': model, 'optimizer': optimizer}
	ckpt = torch.save(to_save, osp.join(tmp_dir, "dummy.pth"))

	load_handler = CheckpointLoader(osp.join(tmp_dir, 'dummy.pth'), to_load, strict_shape=False)
	load_handler.attach(trainer)

	trainer.run(train_loader, max_epochs=1)


if __name__ == '__main__':
	with tempfile.TemporaryDirectory() as tmp_dir:
		main(tmp_dir)
```

Nic-Ma:
Fixes #2506 .

### Description
This PR fixed the strick_shape arg issue for non-model items.

### Status
**Ready**

### Types of changes


monai/handlers/checkpoint_loader.py
```diff
@@ -53,6 +54,11 @@ class CheckpointLoader:
             This can be useful advanced feature for transfer learning. users should totally
             understand which layers will have different shape. default to `True`.
 
+    Note: if `strict_shape=False`, will only load checkpoint for `torch.nn.Module` and skip other
```
wyli, line 57: This flag is a bit confusing, I think the name could be "nn_modules_only" or "nn_mod_only"? But if this is the case, do we support nn.module in a nested dict, such as {"models": {"test": {"model": nn_parameters}}}

Nic-Ma: Hi @wyli , the main point here is that: `strict_shape=False` can only work with model, if model layers have different shape, we should not load `state_dict` for the related optimizer and lr_sheduler, etc. anymore.
What do you think a better name for this arg?

Thanks.

wyli: I'm not sure if I understand the current `strict` and `strict_shape` correctly, could you document more in the unit test for the test cases, or @sandylaker could you please help review this pull request?

Nic-Ma: Hi @wyli ,

Sure, I enhanced the unit tests with more check and doc-strings.
Actually, I think if loading checkpoint for a model with different shape layers, usually it's not a continuous training, so should not load the checkpoint of optimizer, lr_scheduler, etc.
And if model **has different shape layers**, we should not load other related things, which may cause an unexpected error as the `state_dict` of these components are not pure `Tensor` like nn.Module.
What do you think?

Thanks.

sandylaker: @wyli 

1. The `strict` flag checks if the model's keys match the state dictionary's keys. Here the state dictionary is loaded from the check point. For example, If `model = nn.Linear(2, 1, bias=False)` and the `state_dict` to be loaded is from `nn.Linear(2, 1, bias=True)`, then setting `strict = False` allows for loading such `state_dict`, otherwise `torch` will complain about unmatched keys, see [here](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.load_state_dict). It is also to notice that this flag is actually passed to `ignite.handlers.Checkpoint.load_state_dict` (see [here](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.load_state_dict)) at the first place. In this function, `strict` has only effect to `nn.Module`, `DP` and `DDP` wrappers. For other classes like `Optimizer`, `_LRscheduler`, their `load_state_dict` does not accept extra flags. In this case, the `strict` flag is simply ignored.
2. The `strict_shape` forces the tensors of matched keys to have the same shape. Thus, when doing transfer learning, it is better to set `strict_shape=False`. E.g., one might hope to fine-tune a model on CIFAR-10 with pre-trained weights on ImageNet. In this case, the shape of last linear layer is different.

To make the docs more comprehensive, it would be great if a link to `ignite.Checkpoint.load_state_dict` or `nn.Module.load_state_dict` is added. Thus, the user can refer to these links when confused.

monai/handlers/checkpoint_loader.py
```diff
@@ -92,15 +98,22 @@ def __call__(self, engine: Engine) -> None:
...
-            # single object and checkpoint is directly a state_dict
-            if len(self.load_dict) == 1 and k not in checkpoint:
-                checkpoint = {k: checkpoint}
+        k, _ = list(self.load_dict.items())[0]
+        # single object and checkpoint is directly a state_dict
+        if len(self.load_dict) == 1 and k not in checkpoint:
+            checkpoint = {k: checkpoint}
 
-            # skip items that don't match data shape
+        # skip items that don't match data shape
```
wyli, line 106: this comment says skipping the shape mismatches, but line 110 is actually checking the type, not the shape.  is there a test case or use case that `strict=True, strict_shape=False`?

Nic-Ma: Hi @wyli ,

Thanks for your reminder, this is a previous comment line that I forgot to update.
Updated in the latest commit and also added check for `strict=True, strict_shape=False`.

Thanks.

monai/handlers/checkpoint_loader.py
Nic-Ma: Hi @sandylaker ,

Thanks for your detailed explanation!
Sure, I added the 2 links in the latest commit.

Thanks.

monai/handlers/checkpoint_loader.py
```diff
@@ -92,15 +105,22 @@ def __call__(self, engine: Engine) -> None:
...
-            # skip items that don't match data shape
+        if not self.strict_shape:
+            pop_items: List[str] = []
             for k, obj in self.load_dict.items():
-                checkpoint[k] = copy_model_state(obj, checkpoint, inplace=False)[0]
+                if isinstance(obj, (torch.nn.Module)):
+                    # skip items that don't match key name or data shape
+                    checkpoint[k] = copy_model_state(obj, checkpoint, inplace=False)[0]
+                else:
+                    warnings.warn("`strict_shape` is False, load checkpoint for model, skip others in `load_dict`.")
```
wyli, line 120: so, is this warning message correct here, should it be `strict` instead of `strict_shape`?

monai/handlers/checkpoint_loader.py
```diff
@@ -92,15 +105,22 @@ def __call__(self, engine: Engine) -> None:
...
+            checkpoint = {k: checkpoint}
 
-            # skip items that don't match data shape
+        if not self.strict_shape:
+            pop_items: List[str] = []
             for k, obj in self.load_dict.items():
-                checkpoint[k] = copy_model_state(obj, checkpoint, inplace=False)[0]
+                if isinstance(obj, (torch.nn.Module)):
+                    # skip items that don't match key name or data shape
+                    checkpoint[k] = copy_model_state(obj, checkpoint, inplace=False)[0]
```
sandylaker, line 118: @Nic-Ma according to the docstring of `copy_model_state`, the values of `obj` will be overwritten by the ones from `checkpoint` when their keys match (also see [here](https://github.com/Project-MONAI/MONAI/blob/ce45640b2aab6e094e3c8af9ea49634497df9fb6/monai/networks/utils.py#L385-L386)). So for example, if the current model is `nn.Linear(4096, 10)`, and the corresponding Tensor from checkpoint has shape `(4096, 1000)`, the returned `state_dict` will have a tensor with shape `(4096, 1000)` right? Is this the desirable behavior? In most cases, we hope to skip loading the unmatched weights rather than let them overwrite the current model's weights.

wyli: that's the expected behaviour because the user explicitly provides a `mapping` dict here for the model keys

monai/handlers/checkpoint_loader.py
Nic-Ma: Hi @wyli ,

As the `Note` in doc-string: "if `strict_shape=False`, will only load checkpoint for `torch.nn.Module` and skip other items in the `load_dict`."
We only skip others when `strict_shape` is False, optimizer, lr_scheduler, etc. can support `strict=False`, it's PyTorch behavior.

Thanks.

monai/handlers/checkpoint_loader.py
wyli: thanks @sandylaker, @Nic-Ma, I was confused by two things, 
1. "model keys" are the keys in `dict(nn.modules.named_parameters())`, not the keys of `load_dict` such as `load_dict={"net": net2, "opt": opt2}`, 

2. in the current implementation,  `strict_shape=False` is actually implemented as "don't check anything", the `strict` flag is ignored... I thought `strict_shape=False` means the keys should match, but allow for shape mismatch.

monai/handlers/checkpoint_loader.py
wyli: yeah, `strict_shape=False` actually means ignore the `strict` flag, try to load all the nn.module.parameters from the checkpoint as long as the shapes match and keys match, and quietly ignore any mismatched keys, mismatched shapes... to me this is a confusing option

monai/handlers/checkpoint_loader.py
Nic-Ma: Hi @wyli ,

1. Sure, let me try to enhance the doc-string to make the `model keys` more clear in `strict` arg.
2. Actually, we can't allow "keys must match but shape mismatch", because the `strict` check for keys is in PyTorch itself when **applying** checkpoint, if the shape mismatch, it will throw an exception. So can we enhance the doc-string of `strict_shape` to make it clear to users that it will ignore `strict=True`?

Thanks.

monai/handlers/checkpoint_loader.py
Nic-Ma: Sure, let's discuss it in the other thread.
Thanks.

monai/handlers/checkpoint_loader.py
wyli: I think you put a warning there already. Maybe it's fine for now

wyli (approved):
Thanks, I think the API could be improved, but I guess that is beyond this bug fix PR

monai/handlers/checkpoint_loader.py
Nic-Ma: Enhanced doc-strings in the latest commit.
Thanks.

PR #2522, opened 2021-07-05
2511 Add EnsureType transform

### LINKED ISSUE #2511: Add `EnsureTensor` transform
Currently, we have the `ToTensor` transform, will add a more flexible transform `EnsureTensor` which sounds more clear and better to support `structured` or `nested` data.

Nic-Ma:
Fixes #2511 .

### Description
This PR added the `EnsureType` and `EnsureTyped` transforms.

### Status
**Ready**

### Types of changes


monai/transforms/utility/array.py
```diff
@@ -316,6 +319,42 @@ def __call__(self, img) -> torch.Tensor:
...
+
+    def __call__(self, data):
+        """
+        Args:
+            data: input data can be PyTorch Tensor, numpy array, list, dictionary, int, float, bool, str, etc.
+                will ensure Tensor, Numpy array, float, int, bool as Tensors, strings and objects keep the original.
+                for dictionay, list or tuple, ensure every item as a Tensor if applicable.
+
+        """
+        if isinstance(data, torch.Tensor):
```
wyli, line 337: could you create utility functions for this and the inverse? then the transforms could use the utils

Nic-Ma: Hi @wyli ,

Sure, Thanks for your review.
I updated to utils in the latest commit.

Thanks in advance.

wyli (dismissed):
Thanks!

Nic-Ma:
Hi @wyli ,

Thanks for your review.
BTW, there is one logic slightly different from `ToTensor` transform: if passing a list like: `[1, 2, 3, 4]`,  `EnsureTensor` will convert it to `[torch.Tensor(1), torch.Tensor(2), torch.Tensor(3), torch.Tensor(4)]`, while `ToTensor` will convert it to `torch.Tensor([1, 2, 3, 4])`. I prefer to the behavior of `EnsureTensor` because users may also pass `[[1], [2, 3], 4]` that can't convert to 1 Tensor array, what do you think?

Thanks.

wyli:
> Hi @wyli ,
> 
> Thanks for your review.
> BTW, there is one logic slightly different from `ToTensor` transform: if passing a list like: `[1, 2, 3, 4]`, `EnsureTensor` will convert it to `[torch.Tensor(1), torch.Tensor(2), torch.Tensor(3), torch.Tensor(4)]`, while `ToTensor` will convert it to `torch.Tensor([1, 2, 3, 4])`. I prefer to the behavior of `EnsureTensor` because users may also pass `[[1], [2, 3], 4]` that can't convert to 1 Tensor array.
> 
> Thanks.

sure I feel the transform shouldn't change the data structure, but does it mean we shouldn't deprecate `ToTensor`?

monai/transforms/utility/dictionary.py
```diff
@@ -443,6 +448,38 @@ def inverse(self, data: Mapping[Hashable, Any]) -> Dict[Hashable, Any]:
...
+        for key in self.key_iterator(d):
+            self.push_transform(d, key)
+            d[key] = self.converter(d[key])
+        return d
+
+    def inverse(self, data: Mapping[Hashable, Any]) -> Dict[Hashable, Any]:
+        d = deepcopy(dict(data))
+        for key in self.key_iterator(d):
+            # Apply inverse
+            d[key] = tensor_to_numpy(d[key])
```
rijobro, line 477: The inverse isn't necessarily back to numpy, it depends on the input type.

monai/transforms/utility/dictionary.py
```diff
@@ -409,6 +413,7 @@ def __call__(
         return d
 
 
+@deprecated(since="0.6.0", removed="0.7.0", msg_suffix="Please consider using `EnsureTensord` transform instead.")
```
rijobro, line 416: Why do we need to deprecate this? If so, I think that implies that we should also be deprecating ToNumpy, ToPIL, ToCupy etc

Nic-Ma:
Hi @wyli ,

That's a good point, yes, I think we should keep `ToTensor` just like the `AddChannel` and `EnsureChannelFirst`.
Let me update the PR.

Thanks.

monai/transforms/utility/dictionary.py
Nic-Ma: Hi @rijobro ,

Here we only convert Tensor back to numpy, not all kinds of data, if scalar value, convert to regular number. Just the inverse operation of this transform.

Thanks.

rijobro:
I think we could have more general solutions. We have `LoadImage` instead of `LoadNifti`, so we could also have `ToType`. This would take the output type as argument (`Numpy`, `Tensor`) etc. and could also have the option to do recursively or not. This would solve your problem, whilst not creating a confusing amount of classes that might leave users unsure as to which they should be using. What do you think?

monai/transforms/utility/dictionary.py
Nic-Ma: Yes, you are making a good point.
Let me remove the deprecated tag.

Thanks.

Nic-Ma:
Hi @rijobro ,

I think the reason we add this `EnsureTensor` transform is because almost every MONAI program needs to use `ToTensor` transform in preprocessing and postprocessing now, but `ToNumpy` is only used in very few cases.
So we want to enhance the transform to convert data to Tensor.

Thanks.

Nic-Ma:
@wyli @rijobro @ericspod ,

So all you guys prefer to `EnsureType(dtype: str)` now?
If so, I will try to update the PR ASAP and write the `decollate` document based on it for v0.6 release tomorrow.

Thanks in advance.

Nic-Ma:
Or maybe we can have both `ToType` and `EnsureTensor` transforms?
For most cases, users just need to put `EnsureTensor` in their pre-transforms for `collate` and post-transforms after `decollate`.

Thanks.

wyli:
> Hi @wyli @rijobro @ericspod ,
> 
> So all you guys prefer to `EnsureType(dtype: str)` now?
> If so, I will try to update the PR ASAP and write the `decollate` document based on it for v0.6 release tomorrow.
> 
> Thanks in advance.

I think we could have a `EnsureType(dtype='tensor')` API? with an initial and default support for 'tensor'. this should be minor change for the current PR, we can expand it if time permits...

ericspod:
An `EnsureType` transform and dict version makes sense. We could do a simple version for now that just substitutes in for `ToTensor` but think about defining it later to convert to and from  Pytorch, Numpy, and native Python types as the case allows.

Nic-Ma:
Hi @wyli @ericspod ,

Plan sounds good, let me try to update this PR today.

Thanks.

Nic-Ma:
Hi @wyli @rijobro @ericspod ,

Thanks for your great suggestions.

Thanks in advance.

monai/transforms/utility/array.py
```diff
@@ -316,6 +323,36 @@ def __call__(self, img) -> torch.Tensor:
...
+    def __call__(self, data):
+        """
+        Args:
+            data: input data can be PyTorch Tensor, numpy array, list, dictionary, int, float, bool, str, etc.
+                will ensure Tensor, Numpy array, float, int, bool as Tensors or numpy arrays, strings and
+                objects keep the original. for dictionay, list or tuple, ensure every item as expected type
+                if applicable.
+
+        """
+        return convert_to_tensor(data) if self.dtype == "tensor" else convert_to_numpy(data)
```
rijobro, line 353: should probably throw something if self.dtype is unknown. Also should probably be `type`, not `dtype` as the latter is normally e.g., `np.float32`.

monai/transforms/utils.py
```diff
@@ -837,3 +843,90 @@ def compute_divisible_spatial_size(spatial_shape: Sequence[int], k: Union[Sequen
...
+    elif isinstance(data, (float, int, bool)):
+        return torch.as_tensor(data)
+    elif isinstance(data, dict):
+        return {k: convert_to_tensor(v) for k, v in data.items()}
+    elif isinstance(data, list):
+        return [convert_to_tensor(i) for i in data]
+    elif isinstance(data, tuple):
+        return tuple([convert_to_tensor(i) for i in data])
+
+    return data
```
wyli, line 877: I think this should raise an error, otherwise it's not 'ensure'
```suggestion
    raise ValueError
```

rijobro (approved):

monai/transforms/utility/array.py
Nic-Ma: Hi @rijobro , I think `type` is python internal operator we can't use it as an arg, right?

Thanks.

monai/transforms/utils.py
Nic-Ma: Hi @wyli ,

Thanks for your review.
This `return data` is to support recursive operation for the `string` or objects in a dict or list. For example, in `meta dict` we can convert the meta_dict: `meta_dict = {"img": np.array([1, 2, 3]), "filename_or_obj": "./workspace/data/img.nii"}`.
What do you think?

Thanks.

wyli: Ok, perhaps we should clarify this in the ensuretype transform. This utility looks fine to me.

monai/transforms/utility/array.py
rijobro: Ok, how about `data_type`?

Nic-Ma: @wyli @ericspod , what do you guys think?
Thanks.

wyli: sure, data_type looks good... or simply `t`: `EnsureType(t='tensor')`

monai/transforms/utils.py
Nic-Ma: Sure, enhanced the doc-string the latest commit.
Thanks.

monai/transforms/utility/array.py
Nic-Ma: Thanks for your discussion.
`t` seems too simple and a little bit confusing, changed to `data_type` in the latest commit.

Thanks.

Nic-Ma:
Hi @wyli ,

Seems the GPU tests are running for 3 hours, is there any issue or expected behavior?

Thanks.

wyli:
I think it's because I pushed the 0.6.0rc2 tag as requested by Sachi, this has triggered full tests...

PR #2531, opened 2021-07-06
update lmdbdataset

wyli:

this PR moves the lmdb `_fill_cache_start_reader` to the dataset's constructor, otherwise multiprocess lmdb writing may slow down the loading a lot

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -428,8 +428,8 @@ class LMDBDataset(PersistentDataset):
 
     def _fill_cache_start_reader(self, show_progress=True):
         """
-        Write the LMDB cache. py-lmdb doesn't have good support of concurrent write
-        this method could be used with multiple processes, but it may have an impact on the performance.
+        Check the LMDB cache and write the cache if needed. py-lmdb doesn't have a good support for concurrent write.
+        This method can be used with multiple processes, but it may have a negative impact on the performance.
 
         Args:
             show_progress: whether to show the progress bar if possible.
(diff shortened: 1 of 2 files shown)


Nic-Ma:
Hi @wyli and @madil90 ,

May I know why you guys put several comments `/build` here? Is this PR ready for review now?

Thanks.

wyli:
> Hi @wyli and @madil90 ,
> 
> May I know why you guys put several comments `/build` here? Is this PR ready for review now?
> 
> Thanks.

sorry, we were testing the premerge, they are irrelevant. this is ready for review, I tested it in a training tutorial, on both NGC and google colab.

monai/data/dataset.py
```diff
@@ -422,43 +423,56 @@ def set_data(self, data: Sequence):
 
         """
         super().set_data(data=data)
-        self._read_env = None
+        self._env = None
+        self._read_env = self._fill_cache_start_reader(show_progress=self.progress)
```
Nic-Ma, line 427: So it will behave like `CacheDataset` that runs the caching before training instead of runtime?
We use `multi-threads` in `CacheDataset` to accelerate the caching progress, as you said the multi-processing perf is not good, is it possible to support multi-threads like `CacheDataset`?

Thanks.

wyli: sure, I'll try that soon

wyli: I looked into this, LMDB is single-writer multi-reader by default https://github.com/jnwatson/py-lmdb/issues/114, so I think the current solution is good for now.

Nic-Ma (approved):
Sure, could you please help add the `single-writer & multi-reader` feature of LMDB to the doc-string? Just in case users will ask the same question when the initial caching is very slow.
Others look good to me.

Thanks.

PR #2551, opened 2021-07-07
OneOf Transform

### LINKED ISSUE #1847: Monai equivalent to tio.OneOf transform? (ETA 8/13)
sometimes it is useful to have multiple augmentation paths

I am a fan of tio.OneOf
https://torchio.readthedocs.io/_modules/torchio/transforms/augmentation/composition.html#OneOf

currently I use the tio function
**Additional context**
I find it so useful that I think it should be part of monai core.

lyndonboone:
### Description
Added OneOf transform (subclass of Compose) similar to same name transform in TorchIO

Intended to close #1847 

### Status
**ready** (inverse impl. for the dict-based inputs)

### Types of changes


monai/transforms/compose.py
```diff
@@ -168,3 +168,64 @@ def inverse(self, data):
...
+
+    def __init__(
+        self,
+        transforms: Optional[Sequence[Callable]] = None,
+        weights: Optional[Sequence[float]] = None,
+        map_items: bool = True,
+        unpack_items: bool = False,
+    ) -> None:
+        if transforms is None:
+            transforms = []
```
rijobro, line 195: Would there be any utility in having an empty list of transforms? I would have thought it would be best to throw an error at this point.

rijobro: I see that you've copied this from the `Compose` constructor. In which case, better to use `super().__init__()`.

monai/transforms/compose.py
```diff
@@ -168,3 +168,64 @@ def inverse(self, data):
...
+            if len(transforms) == 0:
+                weights = []
+            else:
+                weights = [1.0 / len(transforms)] * len(transforms)
+        if len(weights) != len(transforms):
+            raise AssertionError("transforms and weights should be same size if both specified as sequences.")
+        self.weights = ensure_tuple(self._normalize_probabilities(weights))
+        self.map_items = map_items
+        self.unpack_items = unpack_items
+        self.set_random_state(seed=get_seed())
```
rijobro, line 207: all of this sits in parent constructor

monai/transforms/compose.py
lyndonboone: Hi @rijobro , thanks for the helpful comments. I just made a new commit that uses `super().__init__()` which should clean things up hopefully

monai/transforms/compose.py
lyndonboone: Should be resolved in new commit. Thanks!

rijobro:
Thanks, the update looks good. Some tests are required though before the PR can be merged.

I think it would also be really useful to have a go at the inverse method. It should be fairly straightforward. When the method gets called, you just need to store the index of the randomly selected transform. In the inverse direction, you use `self.transforms[index].inverse()`. Not all transforms have inverse methods, of course, so there would need to be some `isinstance(self.transform[index], InvertibleTransform)` logic there.

lyndonboone:
Thanks! I'll try to write some unit tests for it. Would it be best to reference the ones in `test_compose.py`?

I can definitely try to have a go at the inverse method as well. Will give it a try and update the PR when finished

rijobro:
`test_compose` might not be the best to use as reference, perhaps just a normal transform would be better. Something like `test_rand_flip`, for example. You'll want to supply a few transforms and then ensure that if called enough times, they get called at a reasonable percentage (e.g., if you have weightings of 1, 1, 2, you would expect to have 0.25, 0.25, 0.5 splits. You can create blank transforms so that it is very quick, and you can easily run 100 times. That's just one example of a test, I'm sure there are many other sanity checks.

Nic-Ma:
Hi @lyndonboone @rijobro ,

I think the `flatten()` API of this `OneOf` compose is not correct, if using base class's `flatten()`, it will convert `OneOf` to `Compose`, right?

Thanks.

rijobro:
Good point. Do you think we should just raise a `NotImplementedError`? The alternatively is recursively working through list and if we have nested `OneOf` transforms, we would need to multiply the probabilities of each depth. Sounds like more work than it's worth.

e.g.,
```
OneOf:
    w=0.5, Flip
    w=0.5, OneOf:
        w=0.25, Rotate
        w=0.75, Resize
```
would become
```
OneOf:
    w=0.5, Flip
    w=0.125, Rotate
    w=0.375, Resize
```

Nic-Ma:
Hi @rijobro ,

Thanks for sharing the idea.
I am working in this PR: https://github.com/Project-MONAI/MONAI/pull/2667 to add `RandCompose`, I just return `self` in `flatten()`, what do you think about it? Maybe you can help review the draft PR?

Thanks in advance.

Nic-Ma:
Hi @lyndonboone ,

May I know your plan for this PR? Do you have some bandwidth to complete it in a short time?

Thanks in advance.

lyndonboone:
Hi @Nic-Ma , 

My apologies, I got a little tied up with some other things over the past few weeks. I should be able to get back to this early next week if that works.

Nic-Ma:
Hi @lyndonboone ,

Cool, plan sounds good!
Thanks for your contribution.

lyndonboone:
Hi @Nic-Ma , @rijobro ,

Just catching up with the discussion. I was wondering what to do with the `flatten` transform. Is it decided that we just raise `NotImplementedError`? If so, what should we do with the `__len__` method it inherits since it uses `flatten`?

Also, do you think we should use `**kwargs` in the constructor for any arguments it shares with `Compose`?

rijobro:
Hi @lyndonboone, I just added the `flatten` method, so we get the `len` method for free. I also added the start of the unit tests, feel free to flesh those out. 

P.S. I wanted to create a PR to your branch but accidentally pushed straight to it. Sorry about that! Feel free to review the modifications and make any changes you might want.

rijobro:
also i don't think we should do the **kwargs. We have quite a lot of inheritance in our transforms and it would get quite confusing to use kwargs for all of those, since the user would have to check through the documentation for each of the inherited classes.

lyndonboone:
Oh awesome, thanks! No worries at all.

I'll try to have a go at the unit tests and inverse method

monai/transforms/compose.py
```diff
@@ -168,3 +168,82 @@ def inverse(self, data):
...
+            weights = weights / weights.sum()
+            return list(weights)
+
+    def flatten(self):
+        transforms = []
+        weights = []
+        for t, w in zip(self.transforms, self.weights):
+            # if nested, probability is the current weight multiplied by the nested weights,
+            # and so on recursively
+            if isinstance(t, OneOf):
```
Nic-Ma, line 222: @lyndonboone @rijobro , could you please help add some `flatten` unit tests for the case:
```py
Compose([
    transform1,
    transform2,
    OneOf([
        transform3,
        transform4,
        Compose([
            transform5,
            transform6,
        ]),
    ]),
])
```

Thanks.

rijobro: Do we agree that the flatten for the example you've given should remain unchaned?

rijobro: Done - https://github.com/Project-MONAI/MONAI/pull/2551/files#diff-8520f0d748c797ad52f7f72cf05073177776fe1c66b5ac82c4ff64a9936e2e1aR83-R95.

monai/transforms/compose.py
```diff
@@ -168,3 +168,82 @@ def inverse(self, data):
...
+            return input_
+        else:
+            index = self.R.multinomial(1, self.weights).argmax()
+            self.index = index
+            _transform = self.transforms[index]
+            input_ = apply_transform(_transform, input_, self.map_items, self.unpack_items)
+            return input_
+
+    def inverse(self, data):
+        index = self.index
```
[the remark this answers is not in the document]
Nic-Ma: Hi @rijobro ,

About this inverse comment, do you agree to save the index info in the data dict?

Thanks.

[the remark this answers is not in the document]
wyli: Hi @Nic-Ma, do you mean reusing an instance of `OneOf` class multiple times?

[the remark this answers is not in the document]
wyli: OneOf is a subclass of ThreadUnsafe, looks like it's fine here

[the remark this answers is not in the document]
Nic-Ma: Hi @wyli @rijobro ,

The issue is that: when we run the `__call__` of OneOf in multi-processing DataLoader, it recorded different random index to `self.index` in **every subprocess**, when we run `inverse()` later, the `self.index` of main process is actually `None`.
It can be easily reproduced by:
```py
def test_dataloader(self):
    trans = OneOf([Inv(), NonInv()])
    dataset = Dataset(data=[1, 2, 3, 4], transform=trans)
    dataloader = DataLoader(dataset=dataset, batch_size=2, num_workers=2)
    for d in dataloader:
        print(d)
        print("self.index =", trans.index)
        inverted_d = [trans.inverse(i) for i in decollate_batch(d)]
        print(inverted_d)
```
The output is:
```
tensor([2.0000, 2.0135])
self.index = None
[2.0, 2.013509511947632]
tensor([4, 5])
self.index = None
[4, 5]
```

Thanks.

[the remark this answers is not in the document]
Nic-Ma: Hi @wyli , @rijobro ,

What do you think about this issue now? I am not sure whether I described the issue clearly.

Thanks.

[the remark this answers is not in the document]
rijobro: I've updated the inverse to make it threadsafe - https://github.com/Project-MONAI/MONAI/pull/2551/files#diff-1b439819c9b8b0015db06f96bb75d0ff8525dea5a86f06c40ae76f2a00821af4R232-R244

neuronflow:
very cool to see my feature request become reality, thanks for your efforts! 🚀

monai/transforms/compose.py
```diff
@@ -168,3 +169,101 @@ def inverse(self, data):
...
+                # get the index of the applied OneOf transform
+                index = self.get_most_recent_transform(data, key)[InverseKeys.EXTRA_INFO]["index"]
+                # and then remove the OneOf transform
+                self.pop_transform(data, key)
+        if index is None:
+            raise RuntimeError("No invertible transforms have been applied")
+
+        # if applied transform is not InvertibleTransform, throw error
+        _transform = self.transforms[index]
+        if not isinstance(_transform, InvertibleTransform):
```
Nic-Ma, line 265: Seems we use `warning` instead of `exception at other places:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/inverse_batch_transform.py#L47

monai/transforms/compose.py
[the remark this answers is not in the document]
Nic-Ma: Hi @rijobro ,

Thanks for the quick update, look good to me now!

PR #2555, opened 2021-07-08
2554 Add channel dim if necessary

### LINKED ISSUE #2554: `AsDiscrete` needs to add channel dim for scalar label values
An question raised from the internal team:
Some PyTorch loss functions (like `torch.nn.CrossEntropyLoss`) require the label data has no channel dim, but postprocessing transforms and metrics require the label data with channel dim if multiple-classes. Need to add channel dim in post processing if ndim=0.

Actually, we had such feature in previous versions:
https://github.com/Project-MONAI/MONAI/blob/0.5.3/monai/transforms/post/array.py#L95

`VoteEnsemble` already considers the no-channel data:
https://github.com/Project-MONAI/MONAI/blob/0.6.0rc2/monai/transforms/post/array.py#L413

I think maybe we deleted it by mistake when developing the `decollate batch`.

Nic-Ma:
Fixes #2554 .

### Description
This PR added the channel dim if passing a scalar label tensor, also added the unit tests to cover it.

### Status
**Ready**

### Types of changes


Nic-Ma:
Just realized that `one_hot` utility already has a similar logic to add channel dim internally:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/networks/utils.py#L74
Updated the PR to remove the redundant logic in One-Hot and keep the unit tests for it.

Thanks.

yiheng-wang-nv (approved):

PR #2576, opened 2021-07-12
2562 Add support to not save data in a separate folder

### LINKED ISSUE #2562: Save segmentation results in the same folder
Feature requested in discussion: https://github.com/Project-MONAI/MONAI/discussions/2069#discussioncomment-978827.
@jnc13:
Since it is a related question to this discussion, is there a way to avoid having a new directory being created for each segmentation being saved? Ideally I would have the segmentation files just saved to the output directory rather than in separate folders:

/output/S01_seg.nii.gz
/output/S05_seg.nii.gz
etc

Thanks

Nic-Ma:
Fixes #2562 .

### Description
According to user's feedback, this PR enhanced the `nifti_saver` and `png_saver` to optionally save every image in a separate folder or not.

### Status
**Ready**

### Types of changes


monai/data/utils.py
```diff
@@ -712,16 +716,17 @@ def create_file_basename(
...
 
-    if len(postfix) > 0:
-        # add the sub-folder plus the postfix name to become the file basename in the output path
-        output = os.path.join(subfolder_path, filename + "_" + postfix)
-    else:
-        output = os.path.join(subfolder_path, filename)
+    if separate_folder:
+        output = os.path.join(output, filename)
+
+    if not os.path.exists(output):
```
ericspod, line 725:
```python
if not os.path.exists(output):
    os.makedirs(output)
````
-> `os.makedirs(output, exist_ok=True)`

**Comment by ericspod** [comment]
I would suggest the default value should be `False` for this option to preserve existing behaviour, I personally would probably tend not to use this feature.

**Comment by Nic-Ma** [comment]
> I would suggest the default value should be `False` for this option to preserve existing behaviour, I personally would probably tend not to use this feature.

Hi @ericspod ,

Thanks for your review and comments!
Actually, the current existing behavior in MONAI v0.6 equals `True` here, that's why users asked to support saving all images in the same folder.

Thanks.

File: monai/data/utils.py (hunk 1 above)
Comment (Nic-Ma) [reply monai/data/utils.py]: Sure, updated in the latest commit.
Thanks.

**Comment by Nic-Ma** [comment]
Hi @ericspod ,

Thanks for your review.
Could you please help review again?

Thanks in advance.

File: monai/data/utils.py [hunk 2]
```diff
@@ -712,16 +716,16 @@ def create_file_basename(
...
+    # output folder path will be original name without the extension
+    output = os.path.join(folder_path, filedir_rel_path)
 
-    if len(postfix) > 0:
-        # add the sub-folder plus the postfix name to become the file basename in the output path
-        output = os.path.join(subfolder_path, filename + "_" + postfix)
-    else:
-        output = os.path.join(subfolder_path, filename)
+    if separate_folder:
+        output = os.path.join(output, filename)
```
wyli, line 723: not sure if this `filename` should also have the `postfix`?

ericspod: That will be used in the next lines for the subfolders that go in the folder referred to as `output` here.

ericspod (approved):

monai/data/utils.py
Nic-Ma: Hi @wyli @ericspod ,

I think you mean to save as: "output/image_seg/image_seg.nii", right? The existing bahvior in MONAI is "output/image/image_seg.nii".
Maybe let's change to "output/image_seg/image_seg.nii"?

Thanks.

PR #2584, opened 2021-07-12
2571 add DecollateBatch handler

### LINKED ISSUE #2571: No Post-processing Without Decollate!
**Describe the bug**
https://github.com/Project-MONAI/MONAI/blob/3edff7449984f87c371ac87c99244e6b00e97a92/monai/engines/workflow.py#L167-L171

Not all the metrics accepting decollated tensors, for instance `ignite.metrics.Accuracy` expects individual tensors, so disabling the entire post-processing transformation when `decollate=False` is limiting and in fact it has broken the pathology pipeline.

**Expected behavior**
The post-processing being applied regardless of `decollate` status.

Nic-Ma:
Fixes #2571 .

### Description
This PR added the `DecollateBatch` handler to provide a more flexible operation to decollate the batch.

### Status
**Ready**

### Types of changes


monai/handlers/decollate_batch.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def attach(self, engine: Engine) -> None:
+        """
+        Args:
+            engine: Ignite Engine, it can be a trainer, validator or evaluator.
+        """
+        if self.event == "MODEL_COMPLETED":
+            engine.add_event_handler(IterationEvents.MODEL_COMPLETED, self)
+        else:
+            engine.add_event_handler(Events.ITERATION_COMPLETED, self)
```
wyli, line 52: should the `PostProcessing` handler have the same logic of choosing the event to attach?

or perhaps this one  is not needed, the user can have `monai.handlers.PostProcessing(transform=Decollated)`

monai/handlers/decollate_batch.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def __call__(self, engine: Engine) -> None:
+        """
+        Args:
+            engine: Ignite Engine, it can be a trainer, validator or evaluator.
+        """
+        # replicate the scalar values to make sure all the items have batch dimension, then decollate
+        engine.state.batch = decollate_batch(rep_scalar_to_batch(engine.state.batch), detach=True)
```
wyli, line 60: I think `rep_scalar_to_batch` and `detach=True` should be configurable,
whether to apply to `batch` or `output` or both should also be configurable.

not sure if we want to support the decollate of a specific dictionary key if `batch` and `output` are dictionaries

wyli (commented):
thanks, I put suggestions of refactoring. while looking at the `Decollated` transform,
https://github.com/Project-MONAI/MONAI/blob/cbc8941d0de04f52fa3a51dc57b5e3406516ef57/monai/transforms/inverse_batch_transform.py#L107
any idea why we don't allow for decollating a few user-specified keys? it might be useful for quickly decollating a few items from `data`, if `decollate_batch(rep_scalar_to_batch(...))` is not flexible enough for the users

monai/handlers/decollate_batch.py
Nic-Ma: Hi @wyli ,

I also added the logic to choose event in PostProcessing handler in this PR.
But I think that's a good idea to use `Decollated` transform directly, any other use-case to define this handler?

Thanks.

monai/handlers/decollate_batch.py
Nic-Ma: Thanks for your suggestions.
Actually, when I was developing `rep_scalar_to_batch`, I also thought about only decollating several keys of the dict data. But I didn't get how to save the results, currently, we decollate a dict to a list of dicts. Any more thoughts about how to save the results?

Thanks.

monai/handlers/decollate_batch.py
Nic-Ma: I think maybe we can't put `Decollated` in a transform chain of `monai.handlers.PostProcessing` because `engine_apply_transform` will temporarily combine 2 dictionaries and output 2 dictionaries, it can't support the `Decollated` transform which output a list of dictionaries, unless we customize a logic in `Postprocessing` handler for `Decollated` transform.
Let's keep to use a handler for `decollate` first?

Thanks.

wyli: sure, we can refactor this later

monai/handlers/decollate_batch.py
Nic-Ma: OK, I can add this logic to specify keys. Seems this is the best way so far, I will try to make it clear in the doc-string that it will **discard the data** with keys not in the list. It's different from other dict transforms which always keep the not specified keys.
Anyway, as we just released it in v0.6, we can keep refactoring the `decollate` design according to users' feedback, like what we did for the `Invertd` feature.

Thanks.

Nic-Ma: Hi @wyli ,

I feel maybe we don't need `rep_scalar_to_batch` to be configurable, because it will fail if decollating data with scalar tensors and didn't specify the keys.
What do you think?

Thanks.

Nic-Ma: updated in the latest commit.
Thanks.

Nic-Ma:
Hi @wyli ,

Thanks for your review.

Thanks in advance.

monai/handlers/decollate_batch.py
wyli: sure @Nic-Ma, we can keep it hard-coded, maybe in the future we will have cases where we don't want to replicate the scalars. I agree that we can refactor it later...

wyli (approved):
thanks, these are non-breaking changes that improve flexibility.

monai/handlers/decollate_batch.py
Nic-Ma: Thanks, removed `rep_scalar` arg in the latest commit.

PR #2596, opened 2021-07-14
2595 Add support for the set_to_none arg in optimizer

### LINKED ISSUE #2595: Add set_to_none=False for optimizer in workflow
After Pytorch 1.7, Optimizer adds a parameter `set_to_none`. When setting to True, it may increase the performance.
Some references:

https://pytorch.org/docs/stable/generated/torch.optim.Optimizer.zero_grad.html

https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html

Nic-Ma:
Fixes #2595 .

### Description
This PR added support to set the `set_to_none` arg for optimizer in our workflows.

### Status
**Ready**

### Types of changes


yiheng-wang-nv (approved):

monai/engines/trainer.py
```diff
@@ -185,7 +189,11 @@ def _compute_pred_loss():
             engine.fire_event(IterationEvents.LOSS_COMPLETED)
 
         self.network.train()
-        self.optimizer.zero_grad()
+        if get_torch_version_tuple() < (1, 7, 0):
```
wyli, line 192: `get_torch_version_tuple` always a tuple of length 2. but (1,7) < (1,7,0) is probably not what you want here... please change this to `if get_torch_version_tuple() < (1, 7)`.  could you also help change this https://github.com/Project-MONAI/MONAI/blob/025dee7c89a37675f4f1c3d30173eb5f0d288c3b/monai/handlers/utils.py#L115, although that line seems to have no bug

Nic-Ma: Sure, thanks for your tips, this is an interesting corner case.
I updated in the latest commit.

Thanks.

wyli:
Hi @Nic-Ma, there's a change conflict here in case you didn't see it

Nic-Ma:
Hi @wyli ,

Seems our PyTorch 1.7 CI test is not running with official PyTorch 1.7(our version is 1.7.0a0+8deb4fe)?
There is no `set_to_none` arg in it.

Thanks.

Nic-Ma:
Hi @wyli ,

Should I change this line to use PyTorch 1.7.1 version instead?
https://github.com/Project-MONAI/MONAI/blob/dev/.github/workflows/pythonapp.yml#L224

Thanks.

wyli:
> Hi @wyli ,
> 
> Should I change this line to use PyTorch 1.7.1 version instead?
> https://github.com/Project-MONAI/MONAI/blob/dev/.github/workflows/pythonapp.yml#L224
> 
> Thanks.

that is supposed to test nvidia's pytorch docker image, changing that is a separate topic.
Now this PR doesn't work with pytorch:20.09, maybe we can change the code to fix it... perhaps `get_torch_version_tuple() <= (1, 7)`?

Nic-Ma:
Hi @wyli ,

I want to double confirm the root cause is that the docker customized PyTorch 1.6 to make a 1.7a version in it, right?
If we change to `<=1.7`, users with PyTorch 1.7.0 can't this feature anymore:
https://github.com/pytorch/pytorch/blob/v1.7.0/torch/optim/optimizer.py#L167
But maybe not a big problem? 

Thanks.

wyli:
> Hi @wyli ,
> 
> I want to double confirm the root cause is that the docker customized PyTorch 1.6 to make a 1.7a version in it, right?
> If we change to `<=1.7`, users with PyTorch 1.7.0 can't use this feature anymore:
> https://github.com/pytorch/pytorch/blob/v1.7.0/torch/optim/optimizer.py#L167
> But maybe not a big problem?
> 
> Thanks.

yes, 1.7a is actually some 1.6, btw looks like this flag is accurate:
https://github.com/Project-MONAI/MONAI/blob/813368b9f9d498409e881dd4d9340340f2fd0f8d/monai/utils/module.py#L296

if it doesn't work we change to "<=1.7"

wyli:
I think if there's no cyclic import, we can `from monai.utils.module import PT_BEFORE_1_7 `?

PR #2605, opened 2021-07-14
2604 Get raw loss values of hybrid loss

### LINKED ISSUE #2604: Add support to access raw losses of hybrid loss
Feature request from Clara users:

with hybrid loss we need to display each loss separately so that user can calibrate the weights

loses are on different scale to begin with dice range is 0-1  while cross entropy is 0-inf and depends on the size of the volume 
```
need to have out put similar to the one in TF below
Epoch: 1/5, Iter: 1/11 [=                   ]  train_ce_loss: 0.3277  train_dice_loss: 0.9091  train_loss: 4.1865  time: 16.89s        
Epoch: 1/5, Iter: 2/11 [===                 ]  train_ce_loss: 0.2897  train_dice_loss: 0.8875  train_loss: 3.7847  time: 12.38s 
```

Nic-Ma:
Fixes #2604 .

### Description
This PR added support to get the raw loss values of hybrid loss function.

### Status
**Ready**

### Types of changes


monai/losses/dice.py
```diff
@@ -690,9 +696,19 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
             target = torch.squeeze(target, dim=1)
         target = target.long()
         ce_loss = self.cross_entropy(input, target)
+        self.loss_details["ce_loss"] = ce_loss.detach()
         total_loss: torch.Tensor = self.lambda_dice * dice_loss + self.lambda_ce * ce_loss
+        self.loss_details["total_loss"] = total_loss.detach()
+
         return total_loss
 
+    def get_loss_details(self) -> Dict[str, torch.Tensor]:
```
wyli, line 705: I think if we have `print(my_loss.get_loss_details())` it'll cause the gpu tensor to be synced to cpu and have an impact on the speed? not sure if it's a good approach to fetch the intermediate cuda tensor results

Nic-Ma: Hi @wyli ,

Here I just **detach** the loss tensors from graph, not moving it to CPU. Do you mean the `print()` function will potentially move the tensor to CPU? If so, it will impact the speed if you call the `print()` function?
```py
>>> import torch
>>> 
>>> a = torch.tensor(0.5, device="cuda:0")
>>> 
>>> b = a.detach()
>>> 
>>> print(b)
tensor(0.5000, device='cuda:0')
```

Thanks.

wyli: I think it's true according to https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html#avoid-unnecessary-cpu-gpu-synchronization

Nic-Ma: Hi @wyli ,

So in that case, if users want a better perf, they just don't call the `print(my_loss.get_loss_details())`?
Maybe let me enhance the doc-string to make it clear?

Thanks.

Nic-Ma: Users can use this `get_loss_details()` to tune the weights of hybrid loss as the feature request, if want the best perf of training, avoid printing the loss details?

Thanks.

Nic-Ma: updated doc in the latest commit.

Thanks.

wyli: I feel for debugging purposes the user can create a `DiceFocalLoss`, and call `DiceFocalLoss(...).dice(inp, target)`, and `DiceFocalLoss(...).focal(inp, target)` to get the internal values. Perhaps we improve `DiceCELoss` for the same?

Nic-Ma: Hi @wyli ,

Thanks for your suggestion.
For workflow users, I think they can call `get_loss_details()` in a handler to print or debug instead of inheriting trainer and rewrite the `iteration()`?

Thanks.

wyli: I think if we follow this idea, then for many other components, we would have to create temporary variables and add `get_*_details()` everywhere, for example in the networks, optimizers. Not sure if this is our recommended way to debug the components...

Nic-Ma: @ericspod @rijobro , what do you guys think about the solution for this feature request? To tune the weights of hybrid losses. Maybe you have other ideas?

Thanks in advance.

Nic-Ma: @ericspod @rijobro , if you guys have no more comments, I will update the PR to use separate `dice` and `ce` functions according to @wyli 's suggestion.

Thanks.

wyli: perhaps it's a good opportunity to discuss the usability, we can gather more comments on this...

Nic-Ma: Updated in the latest commit to provide the API to compute CE loss in `DiceCELoss`, so users can call `self.dice()` and `self.ce` for debug. Same as `self.focal()` as @wyli suggested.

Thanks.

ericspod (approved):

ericspod:
I think this is a clean solution for this.

PR #2616, opened 2021-07-16
2583 Add DatasetSummary

yiheng-wang-nv:

Fixes #2583 .

### Description
For 3D image segmentation tasks, users may not sure how to set a reasonable voxel spacing value, as well as the normalization parameters, this PR implements a class to address these issues.

### Status
**Ready**

### Types of changes


yiheng-wang-nv:
The initial implementation requires `joblib`, let me choose another way and then reopen this PR.

yiheng-wang-nv:
Hi all, could you please help to review this PR? 
I think this is needed, but I'm not sure if I put the class into the right place. Or should I use a tutorial rather than a MONAI feature here?

wyli:
> Hi all, could you please help to review this PR?
> I think this is needed, but I'm not sure if I put the class into the right place. Or should I use a tutorial rather than a MONAI feature here

> Hi all, could you please help to review this PR?
> I think this is needed, but I'm not sure if I put the class into the right place. Or should I use a tutorial rather than a MONAI feature here?

I think this could be part of monai/data/utils. 
- The implemention could use pytorch or monai dataloader instead of the multiprocessing lib
- it's currently hard coded to use a datalist as in put, but it should be more flexible to take monai dataset as an input
- would be great to reuse this utility to generate the relevant statistics of the dataset, https://github.com/Project-MONAI/MONAI/blob/f943ab4126b37894b3aa5336ceae828731c126d6/monai/handlers/utils.py#L200 instead of hardcoding the type of stats to be median/min/max

monai/data/dataset_calculator.py
```diff
@@ -0,0 +1,128 @@
...
+            num_workers: the maximum number of processes can be used in data loading.
+
+        """
+
+        self.datalist = datalist
+        self.image_key = image_key
+        self.label_key = label_key
+        self.meta_key_postfix = meta_key_postfix
+        self.num_processes = num_processes
+        self.loader = LoadImaged(keys=[image_key, label_key], meta_key_postfix=meta_key_postfix)
```
ericspod, line 62: I'd suggest `loader` should be an argument so that we don't hard code using `LoadImaged`. This would allow us to use input datasets that aren't just dictionaries for one.

monai/data/dataset_calculator.py
```diff
@@ -0,0 +1,128 @@
...
+
+        """
+        intensity = self._run_parallel(self._load_intensity)
+        intensity = np.array(list(itertools.chain.from_iterable(intensity)))
+        min_value, max_value = np.percentile(intensity, [lower, upper])
+        mean_value, std_value = np.mean(intensity), np.std(intensity)
+        output = [min_value, max_value, mean_value, std_value]
+        output = [round(value, 2) for value in output]
+
+        return tuple(output)
```
ericspod, line 128: All these methods begin with underscore, are they meant to be called by another method (otherwise this implies they're meant to be private)?

monai/data/dataset_calculator.py
```diff
@@ -0,0 +1,128 @@
...
+        return image[foreground_idx].tolist()
+
+    def _get_intensity_stats(self, lower: float = 0.5, upper: float = 99.5):
+        """
+        Calculate min, max, mean and std of all intensities. The minimal and maximum
+        values will be processed according to the provided percentiles.
+
+        """
+        intensity = self._run_parallel(self._load_intensity)
+        intensity = np.array(list(itertools.chain.from_iterable(intensity)))
```
ericspod, line 122: This is going to collect all non-zero pixels from all the images correct? This is going to be a very large amount of data for anything other than small datasets. Some way of calculating statistics per-image would avoid this, for min and max values this is easy but for mean and std you would have to additionally record how many pixels these were calculated from and weight the values by this when summing up (ie. weighted average).

monai/data/dataset_calculator.py
```diff
@@ -0,0 +1,128 @@
...
+
+        return tuple(output)
+
+    def _load_intensity(self, path_dict: Dict):
+        """
+        Load intensity from a data's dictionary.
+
+        """
+        data = self.loader(path_dict)
+        image = data[self.image_key]
```
ericspod, line 110: This is hard coding using dictionaries in the loader here, can we support loaders returning arrays directly?

yiheng-wang-nv:
> > Hi all, could you please help to review this PR?
> > I think this is needed, but I'm not sure if I put the class into the right place. Or should I use a tutorial rather than a MONAI feature here
> 
> > Hi all, could you please help to review this PR?
> > I think this is needed, but I'm not sure if I put the class into the right place. Or should I use a tutorial rather than a MONAI feature here?
> 
> I think this could be part of monai/data/utils.
> 
> * The implemention could use pytorch or monai dataloader instead of the multiprocessing lib
> * it's currently hard coded to use a datalist as in put, but it should be more flexible to take monai dataset as an input
> * would be great to reuse this utility to generate the relevant statistics of the dataset, https://github.com/Project-MONAI/MONAI/blob/f943ab4126b37894b3aa5336ceae828731c126d6/monai/handlers/utils.py#L200
>    instead of hardcoding the type of stats to be median/min/max

Thanks @wyli . My question is that when the dataset has images in different spatial sizes, it cannot support `batch_size > 1` directly. Therefore, this PR used multiple processing instead. Do you have any suggestions?

wyli:
> > > Hi all, could you please help to review this PR?
> > > I think this is needed, but I'm not sure if I put the class into the right place. Or should I use a tutorial rather than a MONAI feature here
> > 
> > 
> > > Hi all, could you please help to review this PR?
> > > I think this is needed, but I'm not sure if I put the class into the right place. Or should I use a tutorial rather than a MONAI feature here?
> > 
> > 
> > I think this could be part of monai/data/utils.
> > 
> > * The implemention could use pytorch or monai dataloader instead of the multiprocessing lib
> > * it's currently hard coded to use a datalist as in put, but it should be more flexible to take monai dataset as an input
> > * would be great to reuse this utility to generate the relevant statistics of the dataset, https://github.com/Project-MONAI/MONAI/blob/f943ab4126b37894b3aa5336ceae828731c126d6/monai/handlers/utils.py#L200
> >   
> >   instead of hardcoding the type of stats to be median/min/max
> 
> Thanks @wyli . My question is that when the dataset has images in different spatial sizes, it cannot support `batch_size > 1` directly. Therefore, this PR used multiple processing instead. Do you have any suggestions?

the dataloader is implemented with a proper multiprocessing support, I think you can use it with batch_size=1

monai/data/dataset_calculator.py
yiheng-wang-nv: Thanks @ericspod , now I changed to calculate statistics per-image, and it's not hard to get `mean`, `std`, `min` and `max`. But I'm not sure if it is feasible to get `median` and percentiles if not getting all pixel values. Do you have any suggestions?

yiheng-wang-nv:
> > Hi all, could you please help to review this PR?
> > I think this is needed, but I'm not sure if I put the class into the right place. Or should I use a tutorial rather than a MONAI feature here
> 
> > Hi all, could you please help to review this PR?
> > I think this is needed, but I'm not sure if I put the class into the right place. Or should I use a tutorial rather than a MONAI feature here?
> 
> I think this could be part of monai/data/utils.
> 
> * The implemention could use pytorch or monai dataloader instead of the multiprocessing lib
> * it's currently hard coded to use a datalist as in put, but it should be more flexible to take monai dataset as an input
> * would be great to reuse this utility to generate the relevant statistics of the dataset, https://github.com/Project-MONAI/MONAI/blob/f943ab4126b37894b3aa5336ceae828731c126d6/monai/handlers/utils.py#L200
>    instead of hardcoding the type of stats to be median/min/max

Hi @wyli , thanks for the comments. The code may not be able to be in `monai/data/utils`, since it needs to import `dataset` and `dataloader` and putting into `monai/data/utils` could lead to circular import.

Now I've switched to use monai dataloader.
The input has been changed to use monai dataset.
As for the `write_metrics_reports` handler, it needs to calculate with all accumulated image intensities. I think the way I updated in the latest PR may be faster (see `calculate_statistics` function with can get `mean`, `std`, `max` and `min` and do not need to accumulate all intensities).

Could you please help to re-review the code, thanks!

yiheng-wang-nv:
Hi @ericspod , I've updated the code according to your comments, could you please help to re-review it, thanks!

monai/data/dataset_calculator.py
yiheng-wang-nv: Fixed.

monai/data/dataset_calculator.py
yiheng-wang-nv: Fixed.

monai/data/dataset_calculator.py
yiheng-wang-nv: Fixed.

wyli:
looks like a useful utility, I'd suggest changing its name to `DatasetSummary`, what do you think?

monai/data/dataset_calculator.py
```diff
@@ -0,0 +1,180 @@
...
+            else:
+                image, label = data
+
+            intensities = image[torch.where(label > foreground_threshold)].tolist()
+            if sampling_flag:
+                intensities = intensities[::interval]
+            all_intensities.append(intensities)
+
+        all_intensities = list(chain(*all_intensities))
+        self.data_min_percentile, self.data_max_percentile = np.percentile(
```
ericspod, line 177: We're still accumulating all the foreground pixel values from all the images here, unless we have some way of doing an incremental percentile calculation we are stuck with that.

monai/data/dataset_calculator.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        if len(self.all_meta_data) == 0:
+            self.collect_meta_data()
+        if spacing_key not in self.all_meta_data[0]:
+            raise ValueError("The provided spacing_key is not in self.all_meta_data.")
+
+        all_spacings = torch.vstack([data[spacing_key][:, 1:4] for data in self.all_meta_data]).numpy()
+
+        target_spacing = np.median(all_spacings, axis=0)
+        if max(target_spacing) / min(target_spacing) >= anisotropic_threshold:
```
ericspod, line 97: I think we're missing a test case with anisotropic data.

ericspod (approved):
I think we're good with a few minor changes possibly.

yiheng-wang-nv:
Hi @wyli , I've changed the name, and also add a test case for anisotropic spacings.

wyli:
Perhaps in another iteration we can change the code to use some online histogram estimation such as https://github.com/maki-nage/distogram

yiheng-wang-nv:
Mark: later I will update a tutorial in https://github.com/Project-MONAI/tutorials/tree/master/modules/dynunet_pipeline to show how can we get the spacings and normalization parameters.

PR #2641, opened 2021-07-22
2635 Enhance doc-string of UNet for spatial size

Nic-Ma:
Fixes #2635 .

### Description
As several users asked questions about the input size of UNet, this PR enhanced the doc-string to make it clear.

### Status
**Ready**

### Types of changes


tests/test_unet.py
```diff
@@ -141,7 +141,7 @@ def test_script_without_running_stats(self):
             num_res_units=0,
             norm=("batch", {"track_running_stats": False}),
         )
-        test_data = torch.randn(16, 1, 16, 8)
+        test_data = torch.randn(16, 1, 16, 4)
```
yiheng-wang-nv, line 144: should we add an unittest for wrong input case?

Nic-Ma: I didn't test it because the error will be raised from PyTorch APIs instead of MONAI code.
Anyway, let me add a unit test for it.

Thanks.

Nic-Ma: Hi @yiheng-wang-nv ,

Added unit test in the latest commit.

Thanks.

monai/networks/nets/unet.py
```diff
@@ -57,6 +57,12 @@ def __init__(
             act: activation type and arguments. Defaults to PReLU.
             norm: feature normalization type and arguments. Defaults to instance norm.
             dropout: dropout ratio. Defaults to no dropout.
+
```
wyli, line 60: I think these so depend on the kernel sizes, it's not easy to write down all the viable rules, I'd suggest a tutorial like this https://gist.github.com/fepegar/1fb865494cb44ac043c3189ec415d411

Nic-Ma: Hi @wyli ,

Thanks for sharing the interesting tutorial.
Does our UNet implementation follow the rules as the tutorial? If so, I think we can put the link of this tutorial in the doc-string directly? I saw the tutorial said the implementation of NiftyNet seems easier to write down the rule: "In NiftyNet, the convolutional layers of U-Net use SAME padding. This means that I must just be divisible by 2^N".

Thanks.

Nic-Ma: CC @ericspod for the discussion.
Thanks.

Nic-Ma: @ericspod @rijobro , what do you guys think? Should we put the link of this tutorial in the doc-string directly? Or some pattern to define the input data size?

Thanks.

wyli: The MONAI unet implementation is more flexible, probably the tutorial doesn't fit here. I feel itd be nice to have a tutorial page to show its typical configurations, parameters, model sizes, memory usage etc. Or simply use the basicunet, the setup is straightforward...

Nic-Ma: OK, I submitted a ticket in the tutorial repo: https://github.com/Project-MONAI/tutorials/issues/280 to track this task.

Thanks.

yiheng-wang-nv:
Supplement: `test_unet.py` does not cover the parameters: `kernel_size` and `up_kernel_size`, and `strides` is always `(2, 2)`. I think more test cases are necessary.

Nic-Ma:
> Supplement: `test_unet.py` does not cover the parameters: `kernel_size` and `up_kernel_size`, and `strides` is always `(2, 2)`. I think more test cases are necessary.

Sure, please feel free to submit commits directly.

Thanks.

monai/networks/nets/unet.py
```diff
@@ -49,17 +50,41 @@ def __init__(
...
+            kernel_size: convolution kernel size, the value(s) should be odd. If sequence,
+                its length should equal to dimensions. Defaults to 3.
+            up_kernel_size: upsampling convolution kernel size, the value(s) should be odd. If sequence,
+                its length should equal to dimensions. Defaults to 3.
             num_res_units: number of residual units. Defaults to 0.
             act: activation type and arguments. Defaults to PReLU.
             norm: feature normalization type and arguments. Defaults to instance norm.
             dropout: dropout ratio. Defaults to no dropout.
+
+        Note: Usually, UNet will decrease the spatial size of feature maps in every down block based on
```
wyli, line 64: this can now link to the tutorial

Nic-Ma: Updated in the latest commit.
Thanks.

wyli (approved):

PR #2645, opened 2021-07-22
2644 fixes randbiasfield

### LINKED ISSUE #2644: RandBiasField missing np.exp
**Describe the bug**
The bias field generated by `RandBiasField._generate_random_field` ([link](https://github.com/Project-MONAI/MONAI/blob/0ad9e73639e30f4f1af5a1f4a45da9cb09930179/monai/transforms/intensity/array.py#L434)) should be exponentiated, as done so in [NiftyNet's implementation](https://github.com/NifTK/NiftyNet/blob/935bf4334cd00fa9f9d50f6a95ddcbfdde4031e0/niftynet/layer/rand_bias_field.py#L99).

**To Reproduce**
```
import matplotlib.pyplot as plt
import numpy as np
from monai.transforms import RandBiasField

rand_bias_field = RandBiasField()
arr = np.ones((1,100,100))

bias_field = rand_bias_field(arr)

plt.figure(figsize=(12,6))
plt.subplot(121)
plt.imshow(bias_field[0])

plt.subplot(122)
plt.hist(bias_field.flatten())
plt.show()
```
image

As shown on the histogram on the right, the values generated by the bias field are around 0 and can also be negative. This is not a proper bias field and multiplying an image by this bias field may invert values.

**Expected behavior**
The bias field should have values around 1. This is usually done by exponentiating the bias field since the bias field generated from the polynomial basis function is actually the log-transformed bias field (Refer to [last equation in section 3.1.1](https://www.sciencedirect.com/science/article/pii/S1361841517300257?via%3Dihub) and the NiftyNet implementation of RandBiasField (linked above). 

```
import matplotlib.pyplot as plt
import numpy as np
from monai.transforms import RandBiasField

rand_bias_field = RandBiasField()
arr = np.ones((1,100,100))

bias_field = np.exp(rand_bias_field(arr))

plt.figure(figsize=(12,6))
plt.subplot(121)
plt.imshow(bias_field[0])

plt.subplot(122)
plt.hist(bias_field.flatten())
plt.show()
```

image

**Additional comments**
Are there any plans to make a batched Tensor-compatible implementation of RandBiasField so that this transform can theoretically be done on the GPU after batching? At the moment, the only critical numpy dependencies are the calls to `np.polynomial.legendre.leggrid2d` and `np.polynomial.legendre.leggrid3d`, but these can be easily implemented using PyTorch.

wyli:

Fixes #2644

### Description
adding an exp()

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/intensity/array.py
+++ b/monai/transforms/intensity/array.py
@@ -431,22 +431,19 @@ class RandBiasField(RandomizableTransform):
         self.coeff_range = coeff_range
         self.dtype = dtype
 
-    def _generate_random_field(
-        self,
-        spatial_shape: Tuple[int, ...],
-        rank: int,
-        degree: int,
-        coeff: Tuple[int, ...],
-    ):
+        self._coeff = [1.0]
+
+    def _generate_random_field(self, spatial_shape: Sequence[int], degree: int, coeff: Sequence[float]):
         """
         products of polynomials as bias field estimations
         """
+        rank = len(spatial_shape)
         coeff_mat = np.zeros((degree + 1,) * rank)
         coords = [np.linspace(-1.0, 1.0, dim, dtype=np.float32) for dim in spatial_shape]
         if rank == 2:
             coeff_mat[np.tril_indices(degree + 1)] = coeff
-            field = np.polynomial.legendre.leggrid2d(coords[0], coords[1], coeff_mat)
-        elif rank == 3:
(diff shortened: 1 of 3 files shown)


Nic-Ma (approved):
Looks good to me.
@yiheng-wang-nv could you please help double confirm it?

Thanks.

Issue #2648, opened 2021-07-23
Reimplement 4th place solution of RANZCR kaggle challenge (ETA 9/3)


Nic-Ma:
**Is your feature request related to a problem? Please describe.**
As offline discussed with the code author, plan to reimplement the solution of 4th place in RANZCR kaggle challenge with MONAI components and workflows:
https://www.kaggle.com/c/ranzcr-clip-catheter-line-classification/discussion/229706
- [x] Reproduce the existing pipeline with original code.
- [x] Implement the unsupported transforms and replace all transforms with MONAI transforms.
- [x] Replace network, loss, optimizer, etc. with MONAI components.
- [ ] Rewrite the training flow with MONAI workflows.

Nic-Ma:
For the first step, will implement the `Cutout`, `LongestMaxSize`, `InvertImg` transforms, referring to `Albumentations` package.

Thanks.

Nic-Ma:
Update: @yiheng-wang-nv already reproduced the similar metrics with single fold in both kaggle public LB and private LB, confirmed with the author.

Thanks.

Nic-Ma:
Update for next near steps:
1. Replace the `CSV reading + OpenCV loading` logic with MONAI `CSVDataset + LoadImage transform` @yiheng-wang-nv .
2. Add `IntensityStats` transform to pre-compute the `max`, `mean`, `std` of intensity values of original input image, then other transforms can use it to adjust / clip the `scale` / `shift` parameters, etc. @Nic-Ma 
3. Replace all the albumentation transforms with MONAI transforms @yiheng-wang-nv .

Thanks.

wyli:
Hi @Nic-Ma, not sure about the `IntensityStats` feature request, I thought a lambda function could easily address it. could you create a separate ticket to document it? for the other components, would be great to have a bit more details in separate tickets as well.

Nic-Ma:
Sure, let me open some smaller tickets for items.

Thanks.

Nic-Ma:
Update: @yiheng-wang-nv replaced all the transforms of the pipeline with MONAI transforms and tuned parameters, now it can get slightly better metrics in both the private and public leaderboards when trained on the same dataset:
```
MONAI transforms:
```
![image](https://user-images.githubusercontent.com/27533405/128530285-59c4e5f2-f243-4672-803a-576cccbeb685.png)
```
original pipeline:
```
[...]

PR #2662, opened 2021-07-27
2648 Add LongestRescale transform

Nic-Ma:
part of #2648 .

### Description
This PR added the `LongestRescale` and `LongestRescaled` transforms, inspired by the albumentations API.

### Status
**Ready**

### Types of changes


tests/test_inverse.py
```diff
@@ -249,15 +250,6 @@
     )
 )
 
-TESTS.append(
```
Nic-Ma, line 252: @rijobro I removed this test case here because it's duplicated with previous lines, seems like a copy typo.

Thanks.

wyli:
I think there's a large overlap between this and `Resize`, `Resize` also supports `int` as the `spatial_size`. would be nice to merge this into `Resize`, and provide a flexible `size_mode` for `Resize`. maybe in the future with different `size_mode`, we'll have different interpretation for cases such as `spatial_size=32` and `spatial_size=(32, -1, -1)`

monai/transforms/spatial/array.py
```diff
@@ -378,20 +390,25 @@ def __call__(
...
-        if output_ndim > input_ndim:
-            input_shape = ensure_tuple_size(img.shape, output_ndim + 1, 1)
-            img = img.reshape(input_shape)
-        elif output_ndim < input_ndim:
-            raise ValueError(
-                "len(spatial_size) must be greater or equal to img spatial dimensions, "
-                f"got spatial_size={output_ndim} img={input_ndim}."
-            )
-        spatial_size = fall_back_tuple(self.spatial_size, img.shape[1:])
+        if self.size_mode == "all":
```
wyli, line 393: this could be enhanced with the `look_up_option` api...

wyli (approved):
thanks for the quick updates

monai/transforms/spatial/array.py
Nic-Ma: Sure, updated in the latest commit.
Thanks.

Nic-Ma:
Hi @madil90 ,

May I know why the blossom CI failed? I tried to re-run it, but it still fails.

Thanks in advance.

Issue #2122, opened 2021-04-30
Unable to Train Model in PyTorch Subprocess


aihsani:
**Describe the bug**
Some MONAI Transformations do not seem to allow a developer to use PyTorch multiprocessing (`torch.multiprocessing`) as shown in: https://github.com/pytorch/examples/blob/master/mnist_hogwild/main.py

The relevant lines in the repro code below are:
```python
import torch.multiprocessing as mp
...
p = mp.Process(target=model.train, args=())
p.start()
p.join()
```

The issue seems related to transformations employing `lambda` functions in their parameters (cf. `CropForegroundd`); see stack trace posted.

**To Reproduce**
The code to reproduce the issue is posted below
```python
from monai.transforms import (
    AsDiscrete,
    AddChanneld,
    Compose,
    CropForegroundd,
    LoadImaged,
    Orientationd,
    RandCropByPosNegLabeld,
    ScaleIntensityRanged,
    Spacingd,
    ToTensord,
)
from monai.networks.nets import UNet
from monai.networks.layers import Norm
from monai.losses import DiceLoss
from monai.data import CacheDataset, DataLoader
from monai.apps import download_and_extract
import torch
import tempfile
import os
import glob
import torch.multiprocessing as mp


if __name__ == "__main__":

    directory = os.environ.get("MONAI_DATA_DIRECTORY")
    root_dir = tempfile.mkdtemp() if directory is None else directory
    print(root_dir)


    resource = "https://msd-for-monai.s3-us-west-2.amazonaws.com/Task09_Spleen.tar"
    md5 = "410d4a301da4e5b2f6f86ec3ddba524e"


    compressed_file = os.path.join(root_dir, "Task09_Spleen.tar")
    data_dir = os.path.join(root_dir, "Task09_Spleen")
    if not os.path.exists(data_dir):
        download_and_extract(resource, compressed_file, root_dir, md5)

    train_images = sorted(glob.glob(os.path.join(data_dir, "imagesTr", "*.nii.gz")))
    train_labels = sorted(glob.glob(os.path.join(data_dir, "labelsTr", "*.nii.gz")))
    data_dicts = [
        {"image": image_name, "label": label_name}
        for image_name, label_name in zip(train_images, train_labels)
    ]
    train_files, val_files = data_dicts[:-9], data_dicts[-9:]


    train_transforms = Compose(
        [
            LoadImaged(keys=["image", "label"]),
            AddChanneld(keys=["image", "label"]),
[...]

PR #2667, opened 2021-07-29
2648 Enhance RandLambda to execute deterministic transforms for random part of dataset

Nic-Ma:
part of #2648 .

### Description
This PR enhanced the `RandLambdad` transform and also added array version `RandLambda` transform to execute transform for random part of dataset.

### Status
**Ready**

### Types of changes


rijobro:
Doesn't this imply that we should just implement `RandResize`, `RandCenterSpatialCropd` and `ThresholdIntensityd`?

Nic-Ma:
Hi @rijobro ,

The basic idea of this `RandCompose` is to randomly execute transforms on the data, it's random prob "YES or NO", not like our other random size / random value transforms.
And with this `RandCompose`, no need to implement the "random" version transform for every deterministic transform.

Thanks.

wyli:
my question is why is this implemented as a special compose method, in the example prob=0.2 for Resized, prob=0.3 for CenterSpatialCropd, and prob0.4 for ThresholdIntensityd are all independent events.  Instead, I think each of them could be implemented with a transform wrapper inheriting the `RandomizableTransform` interface https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/transform.py#L236 instead of

Nic-Ma:
Hi @wyli ,

Thanks for your suggestion.
Actually, at the beginning, I planned to enhance the `RandLambdad` transform to support `prob`, but I feel maybe using `RandCompose` can help simplify a list of wrapper of transforms.
Anyway, if you guys prefer to wrapper, let me change back to enhance the `RandLambdad`.

Thanks.

rijobro:
I think enhancing `RandLambda` would be more appropriate. Even though most users will combine transforms in a chain, we shouldn't assume that they will all of the time.

Nic-Ma:
Hi @rijobro  and @wyli ,

I have updated this PR to enhance `RandLambdad` and also added the array version `RandLambda` transform.

Thanks in advance.

wyli:
Thanks, would it be useful to make this invertible?

wyli:
Was partly discussed here https://github.com/Project-MONAI/MONAI/discussions/2004

Nic-Ma:
Hi @wyli @rijobro ,

I added inverse operation to the PR, could you please help review it again?

Thanks in advance.

monai/transforms/utility/dictionary.py
```diff
@@ -852,51 +852,110 @@ class Lambdad(MapTransform):
...
+    Note: The inverse operation doesn't allow to define `extra_info` or access other information, such as the
+        image's original size. If need these complicated information, please write a new InvertibleTransform directly.
+
     """
 
     def __init__(
         self,
         keys: KeysCollection,
         func: Union[Sequence[Callable], Callable],
+        inv_func: Union[Sequence[Callable], Callable] = lambda x: x,
```
wyli, line 870: the lambda might cause the issue https://github.com/Project-MONAI/MONAI/issues/2122

Nic-Ma: Hi @wyli ,

Thanks for pointing this.
I updated in the latest commit.

Thanks.

wyli (approved):

PR #2692, opened 2021-08-03
[2678] Add transform to fill holes and to filter

### LINKED ISSUE #2678: Method to fill holes in segmentations

We have 3D segmentation masks. Every segmented shape is not supposed to have holes within its borders.
Any wholes might be considered potential artifacts.
Especially for our training data we would like to close such holes.

For an example, the following matrix:

```python
[
  [
    [ 1, 1, 1, 2, 2, 2 ],
    [ 1, 1, 1, 2, 2, 2 ],
    [ 1, 1, 1, 2, 2, 2 ],
    [ 0, 3, 0, 0, 4, 0 ],
    [ 3, 3, 3, 4, 0, 4 ],
    [ 0, 3, 0, 0, 4, 0 ],
  ],
  [
    [ 1, 1, 1, 2, 2, 2 ],
    [ 1, 0, 1, 2, 0, 0 ],
    [ 1, 1, 1, 2, 2, 2 ],
    [ 0, 3, 0, 0, 4, 0 ],
    [ 3, 0, 3, 4, 0, 4 ],
    [ 0, 3, 0, 0, 4, 0 ],
  ],
  [
    [ 1, 1, 1, 2, 2, 2 ],
    [ 1, 1, 1, 2, 2, 2 ],
    [ 1, 1, 1, 2, 2, 2 ],
    [ 0, 3, 0, 0, 4, 0 ],
    [ 3, 3, 3, 4, 4, 4 ],
    [ 0, 3, 0, 0, 4, 0 ],
  ],
]
```
Should result in

```python
[
  [
    [ 1, 1, 1, 2, 2, 2 ],
    [ 1, 1, 1, 2, 2, 2 ],
    [ 1, 1, 1, 2, 2, 2 ],
    [ 0, 3, 0, 0, 4, 0 ],
    [ 3, 3, 3, 4, 0, 4 ],
    [ 0, 3, 0, 0, 4, 0 ],
  ],
  [
    [ 1, 1, 1, 2, 2, 2 ],
    [ 1, 1, 1, 2, 0, 0 ],
    [ 1, 1, 1, 2, 2, 2 ],
    [ 0, 3, 0, 0, 4, 0 ],
    [ 3, 3, 3, 4, 0, 4 ],
    [ 0, 3, 0, 0, 4, 0 ],
  ],
  [
    [ 1, 1, 1, 2, 2, 2 ],
    [ 1, 1, 1, 2, 2, 2 ],
    [ 1, 1, 1, 2, 2, 2 ],
    [ 0, 3, 0, 0, 4, 0 ],
    [ 3, 3, 3, 4, 4, 4 ],
    [ 0, 3, 0, 0, 4, 0 ],
  ],
]
```
The only filled holes are the 1 and 3 in the middle slice.
The 2 shape is open to the side and the 4 is open to the back.
The 0 between the classes should stay untouched.

MONAI is missing a "fill_holes" function / transformation.
scipy has an implementation which works for 3D data but only for binary images ([scipy.ndimage.morphology.binary_fill_holes](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.binary_fill_holes.html?highlight=fill_holes#scipy.ndimage.binary_fill_holes)). This requires an iteration over all labels of the image which makes this already slow method unacceptably slow.

I wish that MONAI would implement such a function.

I opened this feature request on the scipy project: https://github.com/scipy/scipy/issues/14504
And this stackoverflow question: https://stackoverflow.com/questions/68608749/performant-way-to-fill-holes-for-categorical-data

I implemented 7 versions using the existing [scipy.ndimage.morphology.binary_fill_holes](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.binary_fill_holes.html?highlight=fill_holes#scipy.ndimage.binary_fill_holes) function (or its implementation) and `numpy`. Here the two best versions so far:

```python
import numpy as np
from scipy.ndimage.morphology import binary_fill_holes

def fill_holes6(img: np.ndarray, applied_labels: np.ndarray) -> np.ndarray:
    output = np.zeros_like(img)
    for i in applied_labels:
        output[binary_fill_holes(img == i)] = i

    return output

def fill_holes7(img: np.ndarray, applied_labels: np.ndarray) -> np.ndarray:
    output = np.zeros(img.shape, dtype=int)
    for i in applied_labels:
        tmp = np.zeros(img.shape, dtype=bool)
        binary_dilation(tmp, structure=None, iterations=-1, mask=img != i, origin=0, border_value=1, output=tmp)
        output[np.logical_not(tmp)] = i
        
    return output
```

In MONAI this could be implemented something like this:

```python
class FillHoles():

    @classmethod
    def _binary_dilation(cls, img: torch.Tensor) -> torch.Tensor:
        img_arr = img.detach().cpu().numpy()
        img_arr = binary_fill_holes(img_arr)
        return torch.as_tensor(img_arr, device=img.device).type(torch.uint8)

    def __init__(self, applied_labels: Union[Sequence[int]]) -> None:
        self.applied_labels = applied_labels

    def __call__(self, img: torch.Tensor,
                 meta_data: Dict[str, Any]) -> Tuple[torch.Tensor, Dict[str, Any]]:
       output = torch.zeros_like(img)
       for i in applied_labels:
          output[self._binary_fill_holes(img == i)] = i

       return output, meta_data
```

I measured the performance the following way (matching my real world data distribution):

```python
import time
import pandas as pd

def measure(funs, t):
    res = []
    for _ in range(t):
        ra = np.random.randint(10, 40)
        sh = np.random.randint(200, 400, 3)
        img = np.random.randint(0, ra, sh)

        applied_labels = np.unique(img)[1:]

        fun_res = []
        for fun in funs:
            start = time.time()
            fun(img, applied_labels)
            end = time.time()
            fun_res.append(end - start)
        res.append(fun_res)
    return np.min(res, axis=0), np.max(res, axis=0), np.mean(res, axis=0), np.std(res, axis=0)

print(measure([fill_holes6, fill_holes7], t=10))
```

For my first implementations I got the following execution times (t=100):

|  | fill_holes1 | fill_holes2 | fill_holes3 |
| -- | ------------- | ------------- | ------------- |
| min  | 6.4s | 6.9s | 6.2s |
| max  | 83.7s | 96.0s | 80.4s  |
| mean  | 32.9s | 37.3s | 31.6s  |
| std  | 17.3s | 20.1s | 16.5 |

This is very slow.
The last implementation fill_holes7 is only 1.27 times faster than fill_holes3.
I really hope there is a more performant way of doing this.

Spenhouet:
Fixes #2678 .

### Description

### Status

### Types of changes


Spenhouet:
@Nic-Ma @wyli I finalized my pull request but the merge checks throw a "import scipy.ndimage (No module named 'scipy')." and I don't get why? I used the same optional import of scipy.ndimage as it is done in `monai.apps.pathology.utils`.
Any idea how to fix that?

Nic-Ma:
Hi @Spenhouet ,

Thanks for your quick PR.
For the CI test errors, please:
1. Add the unit test file to https://github.com/Project-MONAI/MONAI/blob/dev/tests/min_tests.py#L138 to skip min tests.
2. Update the `type-hints` according to error message: monai/transforms/post/array.py:336:13: error: Returning Any from function declared to return "ndarray[Any, Any]"  [no-any-return]

Thanks.

Spenhouet:
@Nic-Ma I don't know how to address the type hint regarding numpy. This does not seem to have been done anywhere in the repository so far. I did use the NdarrayTensor type which is already in use. This one is also not further typeable. So I really don't know how to resolve this?

Nic-Ma:
Hi @Spenhouet ,

Maybe you can change that line to:
```py
return np.asarray(np.where(np.isin(img, self.applied_labels), img, 0))
```

Thanks.

Spenhouet:
@Nic-Ma This resolves it for mypy but then Pylance complains:

> Expression of type "ndarray[Unknown, Unknown]" cannot be assigned to return type "NdarrayTensor@__call__"
  Type "ndarray[Unknown, Unknown]" cannot be assigned to type "NdarrayTensor@__call__"

But since you are not linting with Pylance... this would resolve the issue I guess. Feels just like this actually a mypy issue and should be reported as bug on their side.
I will commit it like this for now.

Spenhouet:
@Nic-Ma @ericspod I feel like the discussion on #2678 is not finished yet. As I see it, the pull request should rather implement the `fill_holes7` logic since it is faster. But @ericspod mentioned that maybe someone might implement a faster solution (I guess directly in C?). Not sure if this should stay open until then or if we should try to merge an implementation of `fill_holes7` first?

wyli (commented):
I think we can merge this initial version first and improve the performance later if needed. What do you think @Nic-Ma @ericspod? Also, 'filter' is a very generic name (and name of a python build-in function), perhaps we should get a more specific name for the transform.

Spenhouet:
@wyli How about `LabelFilter` or `ValueFilter`?

wyli:
Labelfilter sounds good to me

Spenhouet:
I now changed the implementation to the "growing" logic. It is way faster than the connected component one and it becomes much faster if the filling is only done for a subset of labels.

I implemented it so that it directly works on batches and does not need to iterate over batches.

The pull request should be reviewable now.

Spenhouet:
@Nic-Ma Something failed here but I can't see the logs of the pull request checks.

ericspod:
We should merge this version first once we've figured out the issue with the tests. I can't see the logs either so @Spenhouet if you update the branch with changes from dev this will trigger the tests again and hopefully you'll get an error log this time. 

The solution later down the line is to define a convolution operation using max rather than sum as its function or otherwise define binary morphology operators in a Pytorch-compatible way on GPU.

wyli:
there was an issue with the github action today https://www.githubstatus.com/incidents/7p1nnvkgh96y let me rerun this pipeline

Spenhouet:
@wyli Thanks for rerunning the pipeline. The checks passed now.

docs/source/transforms.rst
```diff
@@ -356,6 +356,18 @@ Post-processing
     :members:
     :special-members: __call__
 
+`LabelFilter`
+"""""""""""""""""""""""""""""""
```
Nic-Ma, line 360: Could you please help change the length of  """""""" to be same as the words? Same as other titles.

monai/transforms/post/array.py
```diff
@@ -289,6 +292,144 @@ def __call__(self, img: torch.Tensor) -> torch.Tensor:
...
+
+        Raises:
+            NotImplementedError: The provided image was not a Pytorch Tensor or numpy array.
+
+        Returns:
+            Union[np.ndarray, torch.Tensor]: Pytorch Tensor or numpy array
+            of shape [batch_size, num_channel, spatial_dim1[, spatial_dim2, ...]].
+        """
+        if isinstance(img, np.ndarray):
+            channel_axis = 1
```
Nic-Ma, line 420: Most of MONAI transforms work with `channel-first` data, so it should be `channel_axis = 0`, and you may need to also update the unit tests.

Thanks.

monai/transforms/post/array.py
```diff
@@ -289,6 +292,144 @@ def __call__(self, img: torch.Tensor) -> torch.Tensor:
...
+    """
+
+    def __init__(
+        self, applied_labels: Optional[Union[Iterable[int], int]] = None, connectivity: Optional[int] = None
+    ) -> None:
+        """
+        Initialize the connectivity and limit the labels for which holes are filled.
+
+        Args:
+            applied_labels (Optional[Union[Iterable[int], int]], optional): Labels for which to fill holes. Defaults to None,
```
Nic-Ma, line 391: As you already put type-hints in the code, no need to add the type in doc-strings.
Thanks.

monai/transforms/post/array.py
Spenhouet: This is channel first. Dimension 0 is the batch dimension.

monai/transforms/post/array.py
Spenhouet: I can remove that but it is actually part of the doc standard to do that.

monai/transforms/post/array.py
Nic-Ma: Hi @Spenhouet ,

The transform should work with single channel-first image, without batch dim.
You can refer to the `KeepLargestConnectedComponent`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/post/array.py#L257.

Thanks.

Spenhouet: You are right. I'm a bit confused since I remember having copy pasted the shape definition including the batch part from somewhere but I can't find it in the doc anywhere. Was this in a previous version or did I mix something up? What ever. I will fix it. Thanks for the hint.

Spenhouet: Just sad since I invested at least 30 min. in making the code work with batches.

Nic-Ma: Hi @Spenhouet ,

Thanks for your quick update.
Yes, we used batch-first data for postprocessing transforms before, but changed the behavior since MONAI v0.6.
You can take a look at this doc:
https://github.com/Project-MONAI/MONAI/wiki/v0.5-to-v0.6-migration-guide#decollate-batch-first-tensor-to-list-of-channel-first-tensors
Now we decollate the batch to a list, then execute postprocessing transforms.

Thanks.

Spenhouet: I could leave the logic in to make it compatible with one-hot encoded data. Only the `applied_labels` filter wouldn't work then. It would always fill holes for all labels. It is valid to just throw a `NotImplementedError` in case someone runs this with one-hot and supplies the `applied_labels`? For non one-hot I would perform the squeeze then.

Spenhouet: I might actually have a solution to also make the `applied_labels` work for one-hot encoded data.
But I'm confused again. I can swear I did see a transform before that handled both cases - single channel and one-hot encoded data. Where did that go?

Nic-Ma: Hi @Spenhouet ,

I think you can refer to the `KeepLargestConnectedComponent`, it can handle One-Hot:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/post/array.py#L273

Thanks.

Spenhouet: I believe there is a different one-hot handling needed here. But you will see it when I implemented it :)

Spenhouet:
Interestingly the performance for one-hot encoded images is way better when iterating over the encoding instead of letting the `binary_dilation` do it via the right window / structure.

Implementation with windows / structure:

```python
structure = np.zeros((3, *[3] * spatial_dims))
structure[1, ...] = ndimage.generate_binary_structure(spatial_dims, connectivity or spatial_dims)

img_arr_filtered = img_arr[tuple(applied_labels), ...]
tmp = np.zeros(img_arr_filtered.shape, dtype=bool)
ndimage.binary_dilation(
    tmp,
    structure=structure,
    iterations=-1,
    mask=np.logical_not(img_arr_filtered),
    origin=0,
    border_value=1,
    output=tmp,
)
img_arr[tuple(applied_labels), ...] = np.logical_not(tmp)
```

Implementation with iteration:

```python
structure = ndimage.generate_binary_structure(spatial_dims, connectivity or spatial_dims)

for label in applied_labels:
    tmp = np.zeros(img_arr.shape[1:], dtype=bool)
    ndimage.binary_dilation(
        tmp,
        structure=None,
        iterations=-1,
        mask=np.logical_not(img_arr[label]),
        origin=0,
        border_value=1,
        output=tmp,
    )
    img_arr[label] = np.logical_not(tmp)
```
| | structure  | iteration |
| ---- | ---- | ----- |
| 2 label | 73s | 1.4s |
| all labels | 134s | 31s |

I will therefore implement the iteration.

Spenhouet:
I pushed the change and now let's see if the checks pass.

monai/transforms/utils.py
```diff
@@ -732,6 +734,65 @@ def get_largest_connected_component_mask(img: torch.Tensor, connectivity: Option
...
+            tmp,
+            structure=structure,
+            iterations=-1,
+            mask=np.logical_not(img_arr[label]) if is_one_hot else img_arr[0] != label,
+            origin=0,
+            border_value=1,
+            output=tmp,
+        )
+        if is_one_hot:
+            img_arr[label] = np.logical_not(tmp)
```
Nic-Ma, line 789: Hi @wyli @rijobro , I think for the One-Hot case, we should suppose `applied_labels` as channel indices instead of pure label value, because One-Hot only has values (0, 1), what do you think?

Thanks.

Spenhouet: @Nic-Ma This is how it is implemented, or are you asking if this is acceptable?

wyli: yes, looks good to me

Nic-Ma (approved):
Looks good to me now.
Thanks so much for the great feature and implementation!

PR #2693, opened 2021-08-04
2687 2688 Enhance doc-strings of Affine related transforms

### LINKED ISSUE #2687: The class of `AffineGrid` has wrong docstring
1. The values of parameters like `rotate_params`, `shear_params` and others are fixed, but the descriptions are like : `generate the 1st rotation parameter from uniform ...`

### LINKED ISSUE #2688: Add parameter explanations for `Affine`, `AffineGrid` and others
So far, there are no explanation for the parameter: `shear` and `scale` in `Affine`, `AffineGrid`, `create_shear` and `create_scale`. Users may not know what is `shear`, what is `scale`. 

In addition, I suggest to add value ranges for all parameters belong to `Affine`, thus users can understand what kind of values can be input.

Nic-Ma:
Fixes #2687 
Fixes #2688 .

### Description
This PR enhanced doc-strings of all the `Affine` matrix related transforms according to feedback.

### Status
**Ready**

### Types of changes


monai/transforms/spatial/array.py
```diff
@@ -944,25 +944,26 @@ def __call__(
...
+
+            [
+                [1.0, params[0], params[1], 0.0],
+                [params[2], 1.0, params[3], 0.0],
+                [params[4], params[5], 1.0, 0.0],
+                [0.0, 0.0, 0.0, 1.0],
+            ]
+
+            a tuple of 2 floats for 2D, a tuple of 6 floats for 3D. Defaults to no shearing.
+        translate_params: a tuple of 2 floats for 2D, a tuple of 3 floats for 3D. Translation is in
```
yiheng-wang-nv, line 962: Thanks! `translate` and `scale` are also used for building the affine matrix, should we also add the examples?

yiheng-wang-nv (approved):
Thanks! I suggest we add the link of this [tutorial](https://github.com/Project-MONAI/tutorials/blob/master/modules/transforms_demo_2d.ipynb) in the docstring. It may be helpful for users to decide what kind of values are suitable to choose.

monai/transforms/spatial/array.py
Nic-Ma: Hi @yiheng-wang-nv , I think `translate` and `scale` are easy to understand, every value maps to 1 spatial dim, no need to know how MONAI use it in the affine matrix, so I just added the affine example for `shear` to help users understand what the `shear` arg will be used because it's much more complicated.
Let me try to enhance the doc for the tutorial link.

Thanks.

Nic-Ma: updated in the latest commit.
Thanks.

monai/transforms/spatial/array.py
```diff
@@ -1372,6 +1399,9 @@ def __call__(
 class RandAffine(RandomizableTransform):
     """
     Random affine transform.
+    More details about affine matrix definition: https://www.slicer.org/wiki/Coordinate_systems.
```
wyli, line 1402: There's no need to mention the coordinate system for these data augmentation classes, the affine transforms are not specific to it

wyli (approved):
Thank!

monai/transforms/spatial/array.py
Nic-Ma: updated in the latest commit.
Thanks.

PR #2694, opened 2021-08-04
2689 Add support to use translate param as percent of image size

### LINKED ISSUE #2689: Add Translation as a fraction of the image for `Affine`
So far, `translate_params` in `Affine` only support to translate the image in pixels/voxels.

Reference: https://albumentations.ai/docs/api_reference/augmentations/geometric/transforms/

Nic-Ma:
Fixes #2689 .

### Description
This PR added support to dynamically set translate_param as the percentage of image size for all the `Affine` related transforms.

### Status
**Ready**

### Types of changes


wyli (changes requested):
#2689 aligning the APIs is actually not a trivial task because our affine is defined according to the scipy convention (https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.affine_transform.html), looks like albumentation follows the opencv convention... need to think about it

Nic-Ma:
Hi @wyli ,

Thanks for pointing the difference..
This PR is a non-breaking change, maybe a valuable enhancement to provide better augmentation?
@yiheng-wang-nv @ericspod @rijobro what do you think?

Thanks.

yiheng-wang-nv:
Hi @wyli , I think this feature maybe helpful for some cases. For example, when a training dataset has varies of spatial sizes, specifying a fixed translate range may not be suitable.

ericspod:
This would help generalize a transform sequence to any application and not have to worry about input image size. It's non-breaking so I'm not worried about that, it's additive to the scipy definition.

wyli:
totally agree that this feature is nice, I'm just not sure about the implementation details so I block it -- my concern is that this PR will have "scale * x + translate", which means that the "translate" is defined as shifting the scaled coordinates, but the "translate = translate_percent * original size" is defined in the non-scaled image. This will cause confusion. In my opinion, we should have something like "scale * (x + translate_percent * original_size)" instead of "scale * x + translate_percent * original_size". But maybe I'm wrong here, I haven't got time to go through the details. If you could provide detailed unit tests to show the PR matches argumentation's behaviour, I'm happy to merge it quickly.

Nic-Ma:
Hi @wyli ,

Thanks for your analysis, you made a good point!
I updated the PR to translate based on the percentage of the scaled size, could you please help review it again?

Thanks in advance.

Nic-Ma:
Hi @wyli ,

Albumentation uses the same shape of input image as the output shape, it computes the percentage with input image size instead of scaled size to shift:
https://github.com/albumentations-team/albumentations/blob/master/albumentations/augmentations/geometric/functional.py#L138
In my latest commit, I changed to multiply with our output spatial size to shift, do you think it makes sense?

Thanks.

wyli:
No...as I said earlier, the root cause is the convention difference, and it makes the interface definition inconvenient from our side, or we should think of other easy way out

Nic-Ma:
> totally agree that this feature is nice, I'm just not sure about the implementation details so I block it -- my concern is that this PR will have "scale * x + translate", which means that the "translate" is defined as shifting the scaled coordinates, but the "translate = translate_percent * original size" is defined in the non-scaled image. This will cause confusion. In my opinion, we should have something like "scale * (x + translate_percent * original_size)" instead of "scale * x + translate_percent * original_size". But maybe I'm wrong here, I haven't got time to go through the details. If you could provide detailed unit tests to show the PR matches albumentation behaviour, I'm happy to merge it quickly.

Hi @wyli ,

Thanks for your comments, I am not very sure what you expect to update for this PR now, do you mean to add `albumentation` as optional import and develop a unit test to compare with it?

Thanks.

wyli:
> Thanks for your comments, I am not very sure what you expect to update for this PR now, do you mean to add `albumentation` as optional import and develop a unit test to compare with it?
> 

I think It'd be nice to do an offline comparison of monai vs albumentation RandAffine API, very likely we'll see the gaps and it'll take some effort to align with albumentation. I'd suggest we lower the priority of this PR for now unless it blocks any major features.

ericspod:
Hi @Nic-Ma do we want to revisit this idea? With the current work on lazy evaluation on-going we may want to wait for that to be completed.

Nic-Ma:
Hi @ericspod ,

I think we can close this PR.

Thanks.

PR #2278, opened 2021-05-28
Feature/2231 transforms

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/2231.

### Description
I've sorted transforms into `TorchTransforms` and/or `NumpyTransforms`. 

- `TorchTransform`: convert input to torch, do transform, convert back to original
- `NumpyTransform`: convert input to numpy, do transform, convert back to original
- Both `TorchTransform` and `NumpyTransform`: no need for conversion (e.g., `img[None]`)

I've also added some `enums` to try and unify the input types:

```python
class DataObjects:
    """Common classes used for arrays/tensors and then their usage in dict/mappings."""

    Images = Union[torch.Tensor, np.ndarray]
    Dict = Dict[Hashable, Any]
    Mapping = Mapping[Hashable, Any]
```

We can keep a track of progress with:

```python
from monai import transforms
from inspect import getmembers, isclass
for n,obj in getmembers(transforms):
    if isclass(obj) and issubclass(obj,transforms.Transform) and not issubclass(obj,transforms.MapTransform):
        print(n)
```
And we can add in is not `TorchTransform` and is not `NumpyTransform`, etc. to see which are yet to be considered.

### Status
**Ready/Work in progress/Hold**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick --unittests`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

PR #2740, opened 2021-08-10
RandGaussianNoise and SavitzkyGolaySmooth Torch/NumpyTransforms

rijobro:
Related https://github.com/Project-MONAI/MONAI/issues/2231, https://github.com/Project-MONAI/MONAI/pull/2278.

### Description

This is the first of many PRs to add support for Torch and Numpy input in our transforms, minimising conversion between the two and allowing for GPU data.

In this PR I've added the basic components:
- One example of a transform that is ambivalent to Torch/Numpy input (`RandGaussianSmooth`)
- One example of a transform that potentially requires conversion (`SavitzkyGolaySmooth`)
- Functionality for switching between types
- Functionality for harmonising dtypes
- Tests use numpy, torch and torch.gpu input data
- Convenience enums (`DataObjects`)

### Status

### Types of changes


wyli:
thanks for the PR, I'm looking into the details. @Nic-Ma could you please help review it as well?

rijobro:
@wyli The test seems to be failing on GaussianFilter, but that hasn't been modified. Any ideas? I tried re-running the tests.

Nic-Ma:
Hi @rijobro @wyli ,

Thanks for the quick PR, I planned to review it today, but I am blocked by other internal task now.
Will try to review it ASAP.

Thanks.

monai/transforms/transform.py
```diff
@@ -241,6 +244,21 @@ def __call__(self, data: Any):
         raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
 
 
+class TorchTransform(Transform):
```
wyli, line 247: in the future we may have implementations in `numba`/`cucim`/`cupy`/`jax`, so in my opinion we should use class variable such as `Transform.backends = ["numpy", "torch"]`  to store this information, instead of new classes. `Transform` could have public APIs to query the backend compatibility

wyli:
> @wyli The test seems to be failing on GaussianFilter, but that hasn't been modified. Any ideas? I tried re-running the tests.

looks like it's been addressed by rerunning, probably a GPU testing job scheduling issue, we'll migrate to blossom soon

monai/utils/misc.py
```diff
@@ -322,9 +330,85 @@ def dtype_torch_to_numpy(dtype):
...
+    else:
+        if type(dtype) is not torch.dtype:
+            return dtype
+        return dtype_torch_to_numpy(dtype)
+
+
+def convert_data_type(
+    data: Any,
+    output_type: Optional[type] = None,
+    device: Optional[torch.device] = None,
```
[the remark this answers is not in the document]
rijobro: Not sure I fully understand your first point. If `device` is `None`, the object will remain on the same device.

For the second point, I've allowed `output_type` to be `None` so that this function can be used to change the `dtype` of an image without changing numpy/torch, e.g.: `convert_data_type(data, dtype=torch.float32)`.

monai/transforms/intensity/array.py
```diff
@@ -87,7 +89,7 @@ def randomize(self, im_shape: Sequence[int]) -> None:
         super().randomize(None)
         self._noise = self.R.normal(self.mean, self.R.uniform(0, self.std), size=im_shape)
 
-    def __call__(self, img: Union[torch.Tensor, np.ndarray]) -> Union[torch.Tensor, np.ndarray]:
+    def __call__(self, img: DataObjects.Images) -> DataObjects.Images:
```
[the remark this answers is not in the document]
rijobro: If I use `DataObjects.Images = NdarrayTensor` (which uses `TypeVar`), then I get the following mypy errors:

```
monai/utils/misc.py:359: error: Variable "monai.utils.enums.DataObjects.Images" is not valid as a type  [valid-type]
monai/utils/misc.py:359: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```
where the problematic line of code is:

```python
def convert_data_type(
    data: Any,
    output_type: Optional[type] = None,
    device: Optional[torch.device] = None,
    dtype: Optional[Union[DtypeLike, torch.dtype]] = None,
) -> Tuple[DataObjects.Images, type, Optional[torch.device]]:
```
Any ideas?

monai/utils/misc.py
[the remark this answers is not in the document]
wyli: I see, for the 2nd point the current impl. looks good, we can enhance the docstring later...

for the 1st point, I mainly saw this type of input `device=d[key].device if isinstance(d[key], torch.Tensor) else None` a few times, and think perhaps it could be the default behaviour.

monai/transforms/intensity/array.py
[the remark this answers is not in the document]
wyli: I replace the DataObjects with NdarrayTensor, and it can pass the mypy check:
```diff
diff --git a/monai/utils/type_conversion.py b/monai/utils/type_conversion.py
index c27fe60e..ac03b881 100644
--- a/monai/utils/type_conversion.py
+++ b/monai/utils/type_conversion.py
@@ -3,8 +3,7 @@ from typing import Any, Optional, Sequence, Tuple, Union
 import numpy as np
 import torch
 
-from monai.config.type_definitions import DtypeLike
-from monai.utils.enums import DataObjects
+from monai.config.type_definitions import DtypeLike, NdarrayTensor
 
 __all__ = [
     "dtype_torch_to_numpy",
@@ -78,7 +77,7 @@ def convert_data_type(
     output_type: Optional[type] = None,
     device: Optional[torch.device] = None,
     dtype: Optional[Union[DtypeLike, torch.dtype]] = None,
-) -> Tuple[DataObjects.Images, type, Optional[torch.device]]:
+) -> Tuple[NdarrayTensor, type, Optional[torch.device]]:
     """
     Convert to `torch.Tensor`/`np.ndarray` from `torch.Tensor`/`np.ndarray`/`float`/`int` etc.
```

monai/transforms/intensity/dictionary.py
```diff
@@ -164,7 +168,7 @@ def randomize(self, im_shape: Sequence[int]) -> None:
         for m in self.mean:
             self._noise.append(self.R.normal(m, self.R.uniform(0, self.std), size=im_shape))
 
-    def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.ndarray]:
+    def __call__(self, data: DataObjects.Mapping) -> DataObjects.Dict:
```
Nic-Ma, line 171: Should we simplify to only use `DataObjects.Dict`?
@wyli I don't quite remember why we don't use `Dict` in dict transforms?

Thanks.

monai/utils/misc.py
```diff
@@ -125,6 +123,10 @@ def ensure_tuple_rep(tup: Any, dim: int) -> Tuple[Any, ...]:
         ValueError: Sequence must have length 3, got length 2.
 
     """
+    if isinstance(tup, torch.Tensor):
+        tup = tup.cpu().numpy()
```
Nic-Ma, line 127: Should we `detach` the tensor first?

Thanks.

monai/utils/type_conversion.py
```diff
@@ -0,0 +1,122 @@
...
+    orig_type = type(data)
+    orig_device = data.device if isinstance(data, torch.Tensor) else None
+
+    output_type = output_type or orig_type
+
+    dtype = get_equivalent_dtype(dtype or get_dtype(data), output_type)
+
+    if output_type is torch.Tensor:
+        if orig_type is np.ndarray:
+            if (np.array(data.strides) < 0).any():  # copy if -ve stride
```
Nic-Ma, line 104: Sorry I didn't get why need to copy here?

Thanks in advance.

monai/utils/type_conversion.py
```diff
@@ -0,0 +1,122 @@
...
+                data = data.copy()
+            data = torch.as_tensor(data if data.ndim == 0 else np.ascontiguousarray(data))
```
Nic-Ma, line 106: Maybe we can use the `EnsureType` transform here to simplify the logic?

Thanks.

monai/utils/type_conversion.py
```diff
@@ -0,0 +1,122 @@
...
+        else:
+            data = torch.as_tensor(data)
+        if dtype != data.dtype:
+            data = data.to(dtype)  # type: ignore
+    elif output_type is np.ndarray:
+        if orig_type is torch.Tensor:
+            data = data.detach().cpu().numpy()  # type: ignore
+        else:
+            data = np.array(data)
```
Nic-Ma, line 115: I think `np.asarray()` is better choice here? Because the above `as_tensor()` or `data.detach().cpu().numpy()` are shallow copy.

Thanks.

monai/utils/enums.py
```diff
@@ -233,3 +238,11 @@ class CommonKeys:
     LABEL = "label"
     PRED = "pred"
     LOSS = "loss"
+
+
+class DataObjects:
+    """Common classes used for arrays/tensors and then their usage in dict/mappings."""
+
+    Images = Union[torch.Tensor, np.ndarray]
```
Nic-Ma, line 246: I think we already have such data type?
https://github.com/Project-MONAI/MONAI/blob/dev/monai/config/type_definitions.py#L70

Thanks.

monai/transforms/intensity/array.py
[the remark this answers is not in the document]
Nic-Ma: I also suggest `NdarrayTensor`, thanks.

monai/transforms/transform.py
```diff
@@ -212,6 +213,8 @@ class Transform(ABC):
         :py:class:`monai.transforms.Compose`
     """
 
+    backend: List[str] = []
```
Nic-Ma, line 216: Could you please help add some doc-strings for this `backend` member and how to use it in subclasses?

Thanks.

Nic-Ma (commented):
Sorry for the delay response, very happy to see the great progress of this task.
The overall design looks good to me, put some minor comments inline.
And I want to double confirm the direction: do we want to change transforms to accept both numpy and tensor data and **use PyTorch tensor for the real computation** logic to support differentiable transforms later?

Thanks.

monai/utils/type_conversion.py
rijobro: the following code: `torch.tensor(np.array((1,2,3))[::-1])`
gives the following error:
> **ValueError: At least one stride in the given numpy array is negative, and tensors with negative strides are not currently supported. (You can probably work around this by making a copy of your array  with array.copy().) **

A copy is needed for these. The following is OK: `torch.tensor(np.array((1,2,3))[::-1].copy())`.

rijobro: It seems you're right, further on I have `np.ascontiguousarray` that solves this problem. I've removed it and will add it back in if it becomes necessary for a later PR.

monai/utils/type_conversion.py
rijobro: I don't think it's right to use a class in a function, but I've harmonised them so that they both call `convert_to_numpy` and `convert_to_tensor`.

monai/transforms/intensity/array.py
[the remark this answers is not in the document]
rijobro: Thanks!

rijobro:
@Nic-Ma 

> do we want to change transforms to accept both numpy and tensor data and use PyTorch tensor for the real computation logic to support differentiable transforms later?

Correct, the majority of transforms won't need numpy<->torch conversion, so adding in differentiable transforms should be possible.

rijobro:
@wyli some flake8 errors have appeared since switching to `Ndarraytensor`: https://github.com/Project-MONAI/MONAI/pull/2740/checks?check_run_id=3322770593#step:7:376.

In this snippet: 
```python
   def __call__(self, img: NdarrayTensor) -> NdarrayTensor:
        """
        Apply the transform to `img`.
        """
        self.randomize(img.shape)
        if self._noise is None:
            raise AssertionError
        if not self._do_transform:
            return img
        noise, *_ = convert_data_type(
            self._noise, type(img), dtype=img.dtype, device=img.device if isinstance(img, torch.Tensor) else None
        )
        return img + noise
```
It's complaining about `img.device` for `np.ndarray`. As in, it thinks `img` is only of type `np.ndarray`, even though it's of type `NdarrayTensor`, which should be both `np.ndarray` and `torch.Tensor`.

Nic-Ma (approved):
Thanks for the quick update.
Looks good to me now.

Thanks.

monai/transforms/intensity/dictionary.py
```diff
@@ -164,7 +167,7 @@ def randomize(self, im_shape: Sequence[int]) -> None:
         for m in self.mean:
             self._noise.append(self.R.normal(m, self.R.uniform(0, self.std), size=im_shape))
 
-    def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.ndarray]:
+    def __call__(self, data: Mapping[Hashable, Any]) -> Dict[Hashable, Any]:
```
ericspod, line 170: `Mapping[Hashable, Any]` doesn't really tell us anything since the first type has to be hashable. This and the `Dict` version should be stored as pre-defined types maybe.

monai/transforms/intensity/array.py
```diff
@@ -857,20 +865,23 @@ def __init__(self, window_length: int, order: int, axis: int = 1, mode: str = "z
...
         Returns:
-            np.ndarray containing smoothed result.
+            array containing smoothed result.
 
         """
+        img_t: torch.Tensor
+        img_t, *_ = convert_data_type(img, torch.Tensor)  # type: ignore
+
         # add one to transform axis because a batch axis will be added at dimension 0
         savgol_filter = SavitzkyGolayFilter(self.window_length, self.order, self.axis + 1, self.mode)
```
ericspod, line 881: It would be slightly more efficient to store this as an instance member so it doesn't need to be recreated every time the transform is applied.

monai/transforms/intensity/dictionary.py
rijobro: What do you think those pre-defined types should look like? e.g., `NdarrayTensorDict = Dict[Hashable, NdarrayTensor]`. The problem here is that it implies that all values of the dictionary are of type `NdarrayTensor`, which isn't true. Or is it OK since the only elements we'll deal with in this function are of type `NdarrayTensor`?

rijobro:
@wyli I'm still struggling with the flake8 problem, do you think you could take a look? Cheers

wyli:
Sure I can look into this later today

monai/utils/type_conversion.py
```diff
@@ -0,0 +1,201 @@
...
+        if data is not None and dtype != data.dtype:
+            data = data.astype(dtype)  # type: ignore
+
+    if isinstance(data, torch.Tensor) and device is not None:
+        data = data.to(device)
+
+    return data, orig_type, orig_device
+
+
+def convert_same_type(src: Any, dst: NdarrayTensor) -> Tuple[NdarrayTensor, type, Optional[torch.device]]:
```
wyli, line 191: I added this utility and it doesn't have the typing issue. but I'm not sure about the name of the function, do you have any other idea? @rijobro @Nic-Ma @ericspod

Nic-Ma: Hi @wyli ,

Thanks for the quick update.
It's a little bit confusing, but acceptable. Maybe `convert_to_dst_type()` is more clear?

Thanks.

monai/transforms/intensity/dictionary.py
wyli: for readability, I updated it to `def __call__(self, data: Mapping[Hashable, NdarrayTensor]) -> Dict[Hashable, NdarrayTensor]`

monai/transforms/transform.py
```diff
@@ -213,6 +212,11 @@ class Transform(ABC):
...
+    """Transforms should add data types to this list if they are capable of performing a transform without
```
wyli, line 216: the attribute docstring should be immediately after a simple assignment at the top level of a module. As I update this module I noticed that `Fourier` shouldn't be part of this base API file, so I moved it to transform utils.

PR #2753, opened 2021-08-12
2746 Add round support to AsDiscrete transform

### LINKED ISSUE #2746: Add option for rounding via a transform

Storing NIfTI images is often not done in the right data type (because it is forgotten). I had the issue multiple times that a label / segmentation image which is supposed to only consist of integer values was stored as float and values had slight rounding errors like containing 0.9999999 instead of 1.
When this is loaded with `LoadImage(dtype=np.int16)`, this will get set to 0 instead of 1 since `LoadImage` does only do a `.astype` and this will simply cut of the value.
This is a very hard to find issue if this happens.

- One option could be to add a rounding option to `LoadImage`.
- We could extend `AsDiscrete` with a rounding option.
- It could also be added to `CastToType`
- Another option might be to have a specific `LoadLabels` or `LoadDiscreteImage` transform

The first option is closed to the issue I described but the rounding seems a bit "out of place".
It definitely fits to `AsDiscrete` but for loading label data I would always need to do:

```python
Compose([
    LoadImaged(...),
    AsDiscreted(..., round_to=0),
    CastToTyped(dtype=np.int16),
])
```

I would be in favor of having an explicit  `LoadLabels` or `LoadDiscreteImage` transform. This is what I'm already using in my local implementation.

Nic-Ma:
Fixes #2746 .

### Description
This PR added support to round Tensor in the `AsDiscrete` transform.

### Status
**Ready**

### Types of changes


monai/transforms/post/array.py
```diff
@@ -150,6 +154,7 @@ def __call__(
         n_classes: Optional[int] = None,
         threshold_values: Optional[bool] = None,
         logit_thresh: Optional[float] = None,
+        round_values: Optional[bool] = None,
```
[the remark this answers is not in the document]
Nic-Ma: Hi @wyli ,

Thanks in advance.

monai/transforms/post/array.py
```diff
@@ -179,6 +187,13 @@ def __call__(
         if threshold_values or self.threshold_values:
             img = img >= (self.logit_thresh if logit_thresh is None else logit_thresh)
 
+        rounding = self.rounding if rounding is None else rounding
+        if rounding is not None:
+            if rounding == "torchrounding":
```
wyli, line 192: could use `look_up_options` here

wyli (approved):

monai/transforms/post/array.py
Nic-Ma: Sure, updated in the latest commit.
Thanks.

monai/transforms/post/array.py
```diff
@@ -179,6 +187,11 @@ def __call__(
         if threshold_values or self.threshold_values:
             img = img >= (self.logit_thresh if logit_thresh is None else logit_thresh)
 
+        rounding = self.rounding if rounding is None else rounding
+        if rounding is not None:
+            rounding = look_up_option(rounding, ["torchrounding"])
+            img = torch.round(img)
```
Spenhouet, line 193: @Nic-Ma It might be worth returning this here as integer

```python
return torch.round(img).int()
```
After rounding this can only be integer so returning it as float in the next line does not make sense.

Nic-Ma: Hi @Spenhouet ,

Thanks for your review.
Here we always return `float` because the following logic(metrics, etc.) usually expect float input and we faced several issues about `Long` tensor in computation with float tensor.
I think we can keep the existing behavior to be non-breaking, if you really expect `int`, you can add `CastToType` transform after it.
And of course, if we receive more feedback from users that ask for `int` return type, we can enhance it later.

Thanks.

PR #2756, opened 2021-08-12
NVTX Range Transform

### LINKED ISSUE #2754: NVTX range push/pop transform
NVTX ranges are being defined by push and pop mechanism, currently the nvtx transforms need to be define be separate push and pop transforms (#2713) but it would be useful to have `RangePushPop`  as well, as a more convenient way to annotate a single transform.

drbeh:
Fixes #2754 

### Description
This PR adds Range transform which accept a transform as an input and pushes an NVTX range before that transform, and pops it afterwards. 

### Status
**Ready**

### Types of changes


tests/test_nvtx_transform.py
```diff
@@ -116,33 +119,92 @@ def test_nvtx_transfroms_array(self, input):
...
                 RangePopD(),
-                MarkD("Mark: Transform Ends!"),
+                MarkD("Mark: Transforms (p=1) End!"),
+            ]
+        )
+        output = transforms(input)
+        self.assertIsInstance(output["image"], torch.Tensor)
+        np.testing.assert_array_equal(input["image"], Flip()(output["image"].numpy()))
+
+    # ---------------------
```
Nic-Ma, line 156: Please remove these debug comments.
Thanks.

tests/test_nvtx_transform.py
```diff
@@ -116,33 +119,92 @@ def test_nvtx_transfroms_array(self, input):
...
+            ]
+        )
+        output = transforms(input)
+        self.assertIsInstance(output, torch.Tensor)
+        np.testing.assert_array_equal(input, output)
+        # with prob == 1.0
+        transforms = Compose(
+            [
+                RandMark("Mark: Transforms (p=1) Start!"),
+                Range(RandFlip(prob=1.0)),
```
Nic-Ma, line 177: Could you please also add test cases for `RandRange`?

Thanks.

drbeh: Thanks for the comment. I added `RandRange` and `RandRangeD`.

Nic-Ma (approved):
Looks good to me.
@wyli what do you think?

Thanks.

drbeh:
@wyli, if you don't have any comment on this, I'll merge the PR. Please let me know. Thanks.

wyli:
thanks @drbeh, I have a quick question, what will happen if we have the same `type(transform).__name__` msg for different nvtx transforms?

drbeh:
> thanks @drbeh, I have a quick question, what will happen if we have the same `type(transform).__name__` msg for different nvtx transforms?

If we are profiling the same transform that is used multiple times in the compose, like `ToTensor`, currently, the only way to distinguish them is to assign different name to it. For instance: `Range(ToTensor, "ToTensor_1")` instead of `Range(ToTensor)`.

PR #2757, opened 2021-08-12
Local Patch Shuffle Transform Initial Version Added

finalelement:
Signed-off-by: vnath <vnath@nvidia.com>

### Description
This pull request for adding the Pixel Shuffle Transform. So far I've added the transform for local patch shuffling in the transform/spatial/array

It needs some testing and also simultaneous feedback as how to port it for the dictionary version of it

Also needs test cases.
### Status
**Work in Progress**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -12,7 +12,8 @@
 A collection of "vanilla" transforms for spatial operations
 https://github.com/Project-MONAI/MONAI/wiki/MONAI_Design
 """
-
+import copy
+import random
 import warnings
 from typing import Any, List, Optional, Sequence, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


wyli:
thanks Vish, this looks like a nice starting point. just a reminder that there is still a style error: https://github.com/Project-MONAI/MONAI/pull/2757/checks?check_run_id=3316962486#step:7:116. for the shuffling, looks like the following is efficient as well https://stackoverflow.com/questions/22426609/shuffle-a-numpy-array:
```py
>>> a = np.arange(9).reshape((3,3))
>>> a
array([[0, 1, 2],
       [3, 4, 5],
       [6, 7, 8]])
>>> np.random.shuffle(a.flat)
>>> a
array([[3, 5, 8],
       [7, 6, 2],
       [1, 4, 0]])
```
Also, the implementation currently supports 3d, but I think we could extend to support N-d input in the future.

drbeh:
Does it work only works for 3D? what happen if we use it on 2D?

PR #2773, opened 2021-08-15
2768 Add parameter bias for conv blocks

### LINKED ISSUE #2768: Enhance some networks to set `bias=False` for conv layers
If a conv layer is directly followed by a batch norm layer, the bias in the conv layer is not need, because the `subtract the mean` manipulation will cancel out the effect of bias.

[Reference:](https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html)

In MONAI's `Convolution` class, the default value of `bias` is `True`, and the following networks/blocks cannot configure the `bias` parameter when using `Convolution`:

- `SimpleASPP`
- `HighResNet`
- `VNet`
- `UNet`
- `BasicUNet`
- `AutoEncoder`

yiheng-wang-nv:

Fixes #2768 .

### Description
Except for add parameter bias, this PR also changes `UpCat` to support optional `x_e` in forward function to support encoders that have different number of feature maps .

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/networks/blocks/aspp.py
+++ b/monai/networks/blocks/aspp.py
@@ -39,6 +39,7 @@ class SimpleASPP(nn.Module):
         dilations: Sequence[int] = (1, 2, 4, 6),
         norm_type: Optional[Union[Tuple, str]] = "BATCH",
         acti_type: Optional[Union[Tuple, str]] = "LEAKYRELU",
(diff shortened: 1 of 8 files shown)


wyli (approved):
Thanks!

PR #2793, opened 2021-08-17
2771 HighResBlock convolution

### LINKED ISSUE #2771: `HighResBlock` may have wrong conv block
The `Convolution` block in `HighResBlock` does not utilize acti and norm parameters thus will used default instance norm layer and prelu activation layer. However, it is different from all other `Convolution` blocks in `HighResNet`. Is is a mistake? @wyli

wyli:

Fixes #2771

### Description
this removes the redundant activations after the convolution.
it will break the previously trained models.

### Status
**Ready**

### Types of changes


yiheng-wang-nv (approved):
Thanks!

PR #2800, opened 2021-08-18
get number of data conversions

rijobro:
### Description
Get the number of data conversions required for a `Compose` of transforms

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/compose.py
+++ b/monai/transforms/compose.py
@@ -13,10 +13,9 @@ A collection of generic interfaces for MONAI transforms.
 """
 
 import warnings
-from typing import Any, Callable, Hashable, Mapping, Optional, Sequence, Union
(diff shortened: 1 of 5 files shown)


monai/transforms/compose.py
```diff
@@ -170,6 +171,40 @@ def inverse(self, data):
             data = apply_transform(t.inverse, data, self.map_items, self.unpack_items)
         return data
 
+    def get_number_image_type_conversions(self, test_data, key: Optional[Hashable] = None):
```
wyli, line 174: I feel this is just for profiling/usability purposes, not an essential property of Compose, would be great to make it as a utility, which takes a `Compose` instance as input

rijobro:
@wyli thanks, moved it to its own utility function.

wyli (approved):

PR #2801, opened 2021-08-18
print backends of all MONAI transforms

rijobro:
### Description
Print backends of all MONAI transforms.

### Status
**Ready**

### Types of changes


wyli (approved):
looks nice, I guess we can also create enum object for the `Backend` https://github.com/Project-MONAI/MONAI/blob/e223932ee88509d63407fefbddad8606200e7e69/monai/utils/enums.py#L14

PR #2814, opened 2021-08-20
NVTX Range universal decorator

drbeh:
### Description
This PR implements `Range` which is both a decorator and a context manager for adding NVTX range to most of the MONAI components. It decorates the appropriate method of MONAI components objects such as transfoms, networks, losses, etc. 
It also can be used with `with` runtime context to annotate an arbitrary part of the code.

This PR also removes `Range` transform in favor of `Range` decorator to have only one way of doing this. However, the rest of NVTX transforms (`Mark`, `RangePush`, `RangePop`) still have their own use.

### Status
**Ready**

### Types of changes


monai/utils/nvtx.py
```diff
@@ -0,0 +1,105 @@
...
+        # Define the name to be associated to the range if not provided
+        if self.name is None:
+            name = type(obj).__name__
+            self.name_counter[name] += 1
+            self.name = f"{name}_{self.name_counter[name]}"
+
+        # Define the method to be wrapped if not provided
+        if self.method is None:
+            method_list = [
+                "forward",  # nn.Module
```
wyli, line 59: Looks cool, I think nn.module has both forward and call, will it cause issues here?

drbeh: Thanks, @wyli! Good observation, that's why forward is before call in the list of methods. For `nn.Module`, it annotates `forward`, it's the method that we care about, right?

wyli: I'm not sure, the documentation suggests call instead of using forward https://pytorch.org/docs/stable/generated/torch.nn.Module.html I think it depends on the granularity of measurement as well, maybe we need an option to report call/forward/both

drbeh: You're right, for `nn.Module` annotating call covers more computation than forward but the part that a user has control over  is forward. I checked the source code and the call overhead should be minimal so I'll change the code to annotate the call. 
Anyways, the option is already there to annotation any method and as many as one wants but this is about which method to annotate as default. I'll make them more specific for each type of components that we have in MONAI.

wyli (approved):
thank you, please resolve the conversation then I'll trigger the tests and merge

monai/utils/nvtx.py
drbeh: Now it annotate both forward and call for `nn.Module` and append the method's name to the range's name to distinguish these two ranges.
https://github.com/Project-MONAI/MONAI/blob/69b9aae39dc416144c4663c3f7a31736f254079d/monai/utils/nvtx.py#L117

drbeh:
Wait a moment. I'm doing minor changes.

drbeh:
@wyli, @madil90, do you know why blossom-ci is failing here? Thanks

PR #2832, opened 2021-08-24
Scale intensity

rijobro:
### Description
ScaleIntensity as numpy/torch transform. Requires #2829.

### Status
**Ready**

### Types of changes


wyli (approved):
thanks

Nic-Ma (approved):
Looks good to me, thanks for the quick update!
We may need to address the `# type: ignore` type annotation later.

PR #2843, opened 2021-08-25
CI: pre-commit

Borda:
### Description
Enforcing the already defined styles such as Isort or pep8
Moreover, I would suggest enabling this nice bot that does the fixes for you within open PR
https://pre-commit.ci/

### Status

### Types of changes


wyli:
thanks, this looks great, perhaps we don't need isort/black/flak8 here, those are enforced by `runtests.sh`, monai's codeformater ci, and github actions already. the rest are all very helpful.

Borda:
> monai's codeformater ci, and github actions already. the rest are all very helpful.

there is a certain limitation with using it in GHA, if a bot push to a PR it does not trigger any other actions so the user needs to do e "void" commit to re-start this action... This is the advantage of any external bot, after their edit in PR all GH a are properly triggered :rabbit: 

> runtests.sh

is nice, but persona I would say that have formatting in pre-commit is more transparent also it can be paired wth git so it is used for each commit automatically

wyli:
this codebase has been using the `/black` command for a long time and I'd say the approach works well with our daily workflow and generates quite reliable results efficiently https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#automatic-code-formatting

in my mind this pre-commit.ci tool could be a complementary tool for us to run manually from time to time, fixing minor issues such as `end-of-file-fixer` `check-executables-have-shebangs`

Borda:
> this codebase has been using the `/black`

yes, in such a case, yes, but still as it is implemented in GHA, it would trigger nay consecutive action, right?

> in my mind this pre-commit.ci tool could be a complementary tool

well yes, even though this is rather meant as continuous fixing, you can see how it works for example on TorchMetrics package
https://github.com/PyTorchLightning/metrics/pull/478/commits

wyli:
@monai-bot is a GitHub user and it will trigger consecutive actions. the only caveat is that the PR author should tick `Allow edits from maintainers`

PR #2879, opened 2021-09-01
Split On Grid

drbeh:
### Description
One part of digital pathology pipeline is to split the extracted region/patch  into sub-patches on a grid (for instance 3x3). This PR implement `SplitOnGrid` to enable that for 2D histopathology images.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/apps/pathology/transforms/spatial/array.py
+++ b/monai/apps/pathology/transforms/spatial/array.py
@@ -9,14 +9,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import TYPE_CHECKING, Optional, Tuple, Union
+from typing import Optional, Tuple, Union
 
-from torch.nn import Module
+import torch
 
 from monai.transforms.transform import Transform
-
-if TYPE_CHECKING:
-    import torch
 
 __all__ = ["SplitOnGrid"]
 
--- a/monai/apps/pathology/transforms/spatial/array.py
+++ b/monai/apps/pathology/transforms/spatial/array.py
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Tuple, Union
+from typing import Optional, Tuple, Union
 
 import torch
 
@@ -34,7 +34,9 @@ class SplitOnGrid(Transform):
     """
 
     def __init__(
-        self, grid_size: Union[int, Tuple[int, int]] = (2, 2), patch_size: Union[int, Tuple[int, int]] = (0, 0)
+        self,
+        grid_size: Union[int, Tuple[int, int]] = (2, 2),
+        patch_size: Optional[Union[int, Tuple[int, int]]] = None,
     ):
         # Grid size
         if isinstance(grid_size, int):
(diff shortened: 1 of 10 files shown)


monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -0,0 +1,81 @@
...
+    This transform works only with torch.Tensor inputs.
+
+    Args:
+        grid_shape: a tuple or an integer define the shape of the grid upon which to extract patches.
+            If it's an integer, th evalue will be repeated for each dimension. Default is 2x2
+        patch_size: a tuple or an integer that defines the output patch sizes.
+            If it's an integer, the value will be repeated for each dimension.
+            If None (default), the patch size will be infered from the grid shape.
+
+    Note: the shape of the input image is infered based on the first image used.
```
Nic-Ma, line 33: Should we `prepare_params()` for every image? If so, we don't need `ready` flag?

drbeh: It's only for the this first image to avoid unnecessary computation for each one.

drbeh:
@Nic-Ma, this PR is ready with all the changes that you suggested. Thanks

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -0,0 +1,75 @@
...
+            .transpose(0, 1)
+            .contiguous()
+        )
+        return patches
+
+    def prepare_params(self, image_size):
+        if self.patch_size == (0, 0):
+            self.patch_size = tuple(image_size[i] // self.grid_size[i] for i in range(2))
+
+        self.steps = tuple(
```
Nic-Ma, line 72: Hi @wyli ,

Do you think it's better to make this transform `thread-safe` to support CacheDataset?
Others look good to me now.

Thanks.

wyli: yes I think this should inherit `ThreadUnsafe`, so that the dataset knows to deepcopy it https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/transform.py#L119

Nic-Ma: Or @drbeh could you please optimize this transform to make it thread safe?
If thread unsafe, it needs to deepcopy and always compute the `steps`.

Thanks.

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    Split the image into patches based on the provided grid shape.
+    This transform works only with torch.Tensor inputs.
+
+    Args:
+        grid_shape: a tuple or an integer define the shape of the grid upon which to extract patches.
+            If it's an integer, the value will be repeated for each dimension. Default is 2x2
+        patch_size: a tuple or an integer that defines the output patch sizes.
+            If it's an integer, the value will be repeated for each dimension.
+            The default is (0, 0), where the patch size will be infered from the grid shape.
```
Nic-Ma, line 31: I think `(0, 0)` is not a suitable default value as you actually don't use `(0, 0)`, maybe just default to `None`? And if `None`, compute from the grid shape.

Thanks.

drbeh: You're right and I did the way that you say initially but since it was calculating step based on patch size only once, the type checking was not able to recognize this fact and was producing error. I'll take another look to see if I can solve this in another way or not.

monai/apps/pathology/transforms/spatial/array.py
drbeh: @Nic-Ma could you please let me know which part of the code you think is thread unsafe?
If it is called with multiple threads, the `steps` are being computed at each thread but that shouldn't be a problem, right?

drbeh:
@Nic-Ma I made some changes. Can you check if it's what you had in mind for thread safety. Thanks

Nic-Ma (approved):
Looks good to me now.
@wyli do you have any more comments?

Thanks.

PR #2890, opened 2021-09-03
Update NVTX Range

### LINKED ISSUE #2889: NVTX annotation of multiple use of a transform
NVTX Range decorator currently is able to to patch each transform once. Multiple use of the same transforms with `Range` will result in duplicate annotations.

drbeh:
Fixes #2889 

### Description
This PR updates patching of special methods.

### Status
**Ready**

### Types of changes


monai/utils/nvtx.py
```diff
@@ -109,7 +109,16 @@ def range_wrapper(*args, **kwargs):
             return output
 
         # Replace the method with the wrapped version
-        setattr(owner, method, range_wrapper)
+        if method.startswith("_"):
```
wyli, line 112: The special method name has double underscores, could you please confirm the if condition here?

drbeh: That's right. Previously I wanted to capture `_call_impl` but I changed the logic. Thanks for capturing this. :)

wyli (approved):

PR #2903, opened 2021-09-06
2901 ndim -> spatial_dims

### LINKED ISSUE #2901: consistent parameter `dimensions ` `spatial_dims `
follow up of https://github.com/Project-MONAI/MONAI/pull/2842#issuecomment-905534933
the parameters should be made consistent across different modules. Since this will be a breaking change, it should also use the `deprecated` APIs to provide brief instructions for upgrading.

wyli:

Fixes #2901

### Description
- deprecate ndim
- enhance `deprecated_arg`

### Status
**ready**

### Types of changes


Borda (approved):

monai/losses/image_dissimilarity.py
```diff
@@ -59,9 +59,11 @@ class LocalNormalizedCrossCorrelationLoss(_Loss):
         DeepReg (https://github.com/DeepRegNet/DeepReg)
     """
 
+    @deprecated_arg(name="ndim", since="0.6", msg_suffix="Please use `spatial_dims` instead.")
```
Nic-Ma, line 62: The `since` should be `0.6` or `0.7`?

monai/losses/image_dissimilarity.py
```diff
@@ -81,10 +83,15 @@ def __init__(
...
             smooth_dr: a small constant added to the denominator to avoid nan.
+
+        .. deprecated:: 0.6.0
+            ``ndim`` is deprecated, use ``spatial_dims``.
         """
         super(LocalNormalizedCrossCorrelationLoss, self).__init__(reduction=LossReduction(reduction).value)
 
-        self.ndim = ndim
+        if ndim is not None:
+            spatial_dims = ndim
```
Nic-Ma, line 93: What if user set both `ndim` and `spatial_dims`?

Thanks.

Nic-Ma (commented):
Hi @wyli ,

Thanks for your quick update.
I want to confirm the expected behavior: It will print deprecated warning at 0.7? And will print error at 0.8?

Thanks.

monai/losses/image_dissimilarity.py
Borda: I would say that the new shall have higher priority, so the condition shall be if `ndim` is not default and `spatial_dims` is default

monai/losses/image_dissimilarity.py
wyli: It should be 0.6 otherwise the weekly previews before 0.7 doesn't have this warning message

monai/losses/image_dissimilarity.py
wyli: In the current implementation we couldn't detect if it's a default spatial dim vs a user specified one though (when the value is 3)

wyli:
> Hi @wyli ,
> 
> Thanks for your quick update.
> I want to confirm the expected behavior: It will print deprecated warning at 0.7? And will print error at 0.8?
> 
> Thanks.

It will only print warnings in the current implementation. we'll need another PR to remove the parameters when we decide they are no longer supported. I think we can remove the deprecated APIs before 1.0.

monai/losses/image_dissimilarity.py
Nic-Ma: OK, I see your point.

Thanks.

monai/transforms/intensity/array.py
```diff
@@ -1318,7 +1318,7 @@ def __call__(self, img: Union[np.ndarray, torch.Tensor]) -> Union[torch.Tensor,
         # map back
         img = self.inv_shift_fourier(k, n_dims)
 
-        return img if self.as_tensor_output else img.cpu().detach().numpy()
+        return img if self.as_tensor_output else img.cpu().detach().numpy()  # type: ignore
```
Nic-Ma, line 1321: May I know why adding type-hints ignore here?

Nic-Ma (approved):
Looks good to me.
@yiheng-wang-nv Could you please help double confirm all the network updating tomorrow before merging it?

Thanks in advance.

monai/utils/deprecated.py
```diff
@@ -180,6 +179,9 @@ def _decorator(func):
 
         @wraps(func)
         def _wrapper(*args, **kwargs):
+            if new_name is not None and name in kwargs:
+                # replace the deprecated arg "name" with "new_name"
+                kwargs[new_name] = kwargs[name]
```
wyli, line 184: this change is needed for the required args, for example if we rename the required argument `dim` to `sp_dim`:

deprecated API
```py
def func(dim):
``` 
new API
```py
def func(sp_dim, dim=None):
```
if the user provides `dim` in `func`, it still has the issue of `sp_dim` not being provided in the new version.

So, the deprecated decorator now replace the user-specified deprecated arg with the corresponding latest arg

monai/transforms/intensity/array.py
wyli: `monai.utils.deprecated_arg` as a decorator couldn't preserve the original type annotations of the function to be wrapped...I spend some time on this but couldn't find a good solution yet

PR #2907, opened 2021-09-08
2898 Merge LocalPatchShuffle with RandCoarseDropout

### LINKED ISSUE #2898: merge `LocalPatchShuffling` into `RandCoarseDropout`
If comparing to `RandCoarseDropout`, I think the only difference is that `LocalPatchShuffling` shuffles the content in a hole instead of filling with random values, so maybe we can refactor `RandCoarseDropout` to support all cases, or build a base class and subclasses implement different filling method.

Nic-Ma:
Fixes #2898 .

### Description
This PR merged the `LocalPatchShuffle` with `RandCoarseDropout` transforms to make them more clear.
And rename it to `RandCoarseShuffle`.
The only difference is that the new transform shuffles every channel separately in the same holes, while the original transform creates holes for every channel.

### Status
**Ready**

### Types of changes


monai/transforms/intensity/array.py
```diff
@@ -1697,28 +1683,116 @@ def randomize(self, img_size: Sequence[int]) -> None:
...
+    Kang, Guoliang, et al. "Patchshuffle regularization." arXiv preprint arXiv:1707.07103 (2017).
+    https://arxiv.org/abs/1707.07103
+
+    """
+
+    def _transform_holes(self, img: np.ndarray):
+        for h in self.hole_coords:
+            # shuffle every channel separately
+            for i, c in enumerate(img[h]):
+                patch_channel = c.flatten()
```
wyli, line 1790: I think this is more efficient https://stackoverflow.com/a/22426647 previously my comment was ignored for some reason

wyli: this will give warnings such as "you are shuffling a 'flatiter' object which is not a subclass of 'Sequence'; `shuffle` is not guaranteed to behave correctly. E.g., non-numpy array/tensor objects with view semantics may contain duplicates after shuffling."

wyli (approved):
Thanks

PR #2918, opened 2021-09-09
2701 Add dict version RandCoarseShuffle transform

### LINKED ISSUE #2701: Augmentation Transform | Patch Shuffling (ETA 8/27)
Local pixel/voxel shuffling is a an augmentation that can be used during training, its essentially a way to add noise to the image. In brief the a small sub-block of pixels/voxels are selected from the entire image and then local pixels are shuffled to generate a noisy image

The expectation is to send an image and the transform should be able to shuffle pixels given the hyper-parameters of how many blocks should be shuffled

It should also be noted that pytorch also offers a Pixel Shuffle ([https://pytorch.org/docs/stable/generated/torch.nn.PixelShuffle.html](https://pytorch.org/docs/stable/generated/torch.nn.PixelShuffle.html)). That should not be confused with this one, as it shuffles the dimensions of the image and not sub-blocks/local block etc.

Nic-Ma:
Fixes #2701 .

### Description
This PR added the dict version `RandCoarseShuffled` transform and fixed a random state-related error.

### Status
**Ready**

### Types of changes


wyli (approved):
Looks good to me, please confirm that the self.R call works fine in these dict-based transforms

Nic-Ma:
Hi @wyli ,

Thanks for your review.
Actually, in the dict version transform we don't use the `self.R`, we use the `self.dropper` or `self.shuffle` which is the array based random transform.

Thanks.

PR #2926, opened 2021-09-10
Add device to ToTensor

### LINKED ISSUE #2925: ToTensor Device
Currently when converting from other formats that can reside on gpu like cupy to torch tensor, using `ToTensor`, it will create the tensor on the cpu, and then it should be sent to gpu using `ToDevice`. This will create unnecessary between cpu and gpu and reduce the performance. A solution to that is to have `ToTensor` accept device parameter and directly create the tensor on that machine, and in this case on gpu.

drbeh:
Fixes #2925

### Description
Add device to `ToTensor`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/utils/type_conversion.py
+++ b/monai/utils/type_conversion.py
@@ -83,7 +83,7 @@ def get_dtype(data: Any):
     return type(data)
 
 
-def convert_to_tensor(data, wrap_sequence: bool = False):
+def convert_to_tensor(data, wrap_sequence: bool = False, device: Optional[torch.device] = None):
     """
     Utility to convert the input data to a PyTorch Tensor. If passing a dictionary, list or tuple,
     recursively check every item and convert it to PyTorch Tensor.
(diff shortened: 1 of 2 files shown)


monai/utils/type_conversion.py
```diff
@@ -97,26 +97,29 @@ def convert_to_tensor(data, wrap_sequence: bool = False):
 
     """
     if isinstance(data, torch.Tensor):
-        return data.contiguous()
+        if device is None:
```
Nic-Ma, line 100: I think maybe no need to check `None`, `to()` can support None:
https://github.com/pytorch/ignite/blob/master/ignite/utils.py#L28
@wyli What do you think?

Thanks.

wyli (commented):
thanks, it looks good to me, a minor issue is that it introduces a style warning https://deepsource.io/gh/Project-MONAI/MONAI/run/9269f947-5bda-4b35-80ec-35af55955a22/python/PYL-R1705

monai/utils/type_conversion.py
```diff
@@ -83,7 +83,7 @@ def get_dtype(data: Any):
     return type(data)
 
 
-def convert_to_tensor(data, wrap_sequence: bool = False):
+def convert_to_tensor(data, wrap_sequence: bool = False, device: Optional[torch.device] = None):
```
Nic-Ma, line 86: Hi @wyli ,

I am not sure we only support `device` arg or `kwargs` for `torch.to()` or `torch.as_tensor()` here?
What do you think?

Thanks.

monai/utils/type_conversion.py
drbeh: Thanks, fixed.

monai/utils/type_conversion.py
wyli: Right, 'to' supports some detailed options, 'as_tensor' doesn't support many, perhaps we can enhance it further in another PR if needed.

wyli: Don't understand this integration error, I'll rerun the test https://github.com/Project-MONAI/MONAI/runs/3569135774

wyli (approved):
Integration tests work fine

Issue #2917, opened 2021-09-09
`RandAffine`, `RandFlip`, `RandRotate` with GPU-based `torch.tensor`


wyli:
**Is your feature request related to a problem? Please describe.**
this is as part of #2231 would potentially accelerate the current test-time augmentation pipeline, when combined with https://github.com/Project-MONAI/tutorials/blob/master/acceleration/fast_training_tutorial.ipynb

see also: #2913

cc @diazandr3s

diazandr3s:
This is great! It will help a lot in accelerating TTA used in the MONAI Label Apps

wyli:
addressed by #2922 #2945

Issue #2933, opened 2021-09-13
MONAI transforms can't only support `Tensor` backend


Nic-Ma:
**Is your feature request related to a problem? Please describe.**
Currently, some transforms only support Tensor backend, which will break previous numpy based transform chains.
Need to update them to make sure all the transforms support both `numpy` and `Tensor` backends or only `numpy` backend.

Nic-Ma:
Hi @rijobro ,

I will try to enhance them ASAP before 0.7 release if you have no concerns.

Thanks in advance.

rijobro:
Hi, all of the transforms that I have been updating will perform any necessary numpy<->torch conversions. Do you mean that the updated transforms might now return the data in a different format (e.g., torch instead of numpy) and some of the old transforms can't handle that?

Nic-Ma:
Hi @rijobro ,

Yes, that's the problem, old transforms can't handle the `Tensor` output, and another problem is that: if we input `numpy array` to random transform and output `Tensor`, if the `prob` is < 1.0, then some output will be `numpy array` and some will be `Tensor`, then we can't **collate** them in DataLoader.

Thanks.

rijobro:
I've tried to make it so that for random transforms, I've tried to perform the data conversion that would have been applied if the transform had have been applied. If I've missed that anywhere, please let me know and I'll update. (example of conversion when transform is not applied [here](https://github.com/Project-MONAI/MONAI/pull/2922/files#diff-3a1b9fb96f9cde878a33605fcbd39cf1af51a79c285fe662d80a00f0c43e2d5fL1605)).

Nic-Ma:
Thanks @rijobro ,

I already fixed this kind issue in `RandZoom` transform, I haven't checked other transforms. But I still think we should not only support `Tensor` to be compatible with other old transforms.

Thanks.

wyli:
Just to clarify, my understanding is that the backend is about how the computation is implemented, using numpy or torch API. For now it's fine to have numpy-only or torch-only backend. But we should ensure all the transforms support these input-output types:
- input: numpy, output: numpy
- (optionally) input: torch, output: torch

rijobro:
[...]

PR #2945, opened 2021-09-13
Rotate, Rotated, RandRotate, RandRotated

rijobro:
Partially addresses https://github.com/Project-MONAI/MONAI/issues/2917.

### Description
`Rotate`, `Rotated`, `RandRotate`, `RandRotated`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -441,6 +441,8 @@ class Rotate(Transform, ThreadUnsafe):
             the output data type is always ``np.float32``.
     """
 
+    backend = [TransformBackends.TORCH]
+
     def __init__(
         self,
         angle: Union[Sequence[float], float],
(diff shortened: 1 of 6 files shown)


monai/transforms/spatial/array.py
```diff
@@ -441,14 +441,16 @@ class Rotate(Transform, ThreadUnsafe):
             the output data type is always ``np.float32``.
     """
 
+    backend = [TransformBackends.TORCH]
```
Nic-Ma, line 444: Could you please help add the numpy conversion as we discussed in https://github.com/Project-MONAI/MONAI/issues/2933?

Thanks.

Nic-Ma (approved):

monai/transforms/spatial/array.py
rijobro: Sorry, what do you want me to do? When backend is TORCH, numpy->torch conversion will be done if necessary.

Nic-Ma: Sorry maybe I didn't make it clear, I mean we should support the case `numpy in` -> `numpy out` to be compatible with old transform.

Thanks.

rijobro: Hm, would it be acceptable if I modified all "old" transforms to convert to numpy at the start of the transform? This will ensure backwards compatibility.

wyli: I think @Nic-Ma means we return the original type like what we do here, converting the `zoomed` type to `img` type: https://github.com/Project-MONAI/MONAI/blob/c9b08204b7844b74258348fd74b2b8117912a57e/monai/transforms/spatial/array.py#L644-L645
we can merge all these essential PRs first and review everything for compatibility.

PR #2932, opened 2021-09-12
cuCIM Transform

drbeh:
### Description
This PR implements a wrapper for cuCIM transforms to be used in combination with MONAI transforms.

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [x] New tests added to cover the changes.
- [x] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [x] Quick tests passed locally by running `./runtests.sh --quick --unittests`.
- [x] In-line docstrings updated.
- [x] Documentation updated, tested `make html` command in the `docs/` folder.

PR #2950, opened 2021-09-14
Add dtype to ToCupy

drbeh:
### Description
This PR allow `ToCupy` transform define the expected output `dtype`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/utility/dictionary.py
+++ b/monai/transforms/utility/dictionary.py
@@ -549,17 +549,17 @@ class ToNumpyd(MapTransform):
 class ToCupyd(MapTransform):
     """
     Dictionary-based wrapper of :py:class:`monai.transforms.ToCupy`.
+
+    Args:
+        keys: keys of the corresponding items to be transformed.
+            See also: :py:class:`monai.transforms.compose.MapTransform`
+        allow_missing_keys: don't raise exception if key is missing.
+        dtype: data type specifier. It is inferred from the input by default.
     """
 
     backend = ToCupy.backend
 
     def __init__(self, keys: KeysCollection, allow_missing_keys: bool = False, dtype=None) -> None:
-        """
-        Args:
-            keys: keys of the corresponding items to be transformed.
-                See also: :py:class:`monai.transforms.compose.MapTransform`
-            allow_missing_keys: don't raise exception if key is missing.
-        """
         super().__init__(keys, allow_missing_keys)
         self.converter = ToCupy(dtype=dtype)
 
--- a/monai/transforms/utility/dictionary.py
+++ b/monai/transforms/utility/dictionary.py
@@ -553,7 +553,7 @@ class ToCupyd(MapTransform):
 
     backend = ToCupy.backend
 
-    def __init__(self, keys: KeysCollection, allow_missing_keys: bool = False) -> None:
+    def __init__(self, keys: KeysCollection, allow_missing_keys: bool = False, dtype=None) -> None:
         """
         Args:
             keys: keys of the corresponding items to be transformed.
@@ -561,7 +561,7 @@ class ToCupyd(MapTransform):
             allow_missing_keys: don't raise exception if key is missing.
         """
         super().__init__(keys, allow_missing_keys)
-        self.converter = ToCupy()
+        self.converter = ToCupy(dtype=dtype)
 
     def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> Dict[Hashable, NdarrayOrTensor]:
         d = dict(data)
--- a/monai/utils/type_conversion.py
+++ b/monai/utils/type_conversion.py
@@ -239,10 +239,9 @@ def convert_data_type(
         if data is not None and dtype != data.dtype:
             data = data.astype(dtype)
     elif has_cp and output_type is cp.ndarray:
-        if orig_type is not cp.ndarray:
+        if data is not None:
             data = convert_to_cupy(data, dtype)
-        elif data is not None:
-            data = data.astype(dtype)
+
     else:
         raise ValueError(f"Unsupported output type: {output_type}")
     return data, orig_type, orig_device
--- a/monai/utils/type_conversion.py
+++ b/monai/utils/type_conversion.py
@@ -170,14 +170,7 @@ def convert_to_cupy(data, dtype, wrap_sequence: bool = True):
     """
 
     # direct calls
-    if isinstance(data, cp_ndarray):
-        if dtype is not None:
-            data = data.astype(dtype)
-    elif isinstance(data, np.ndarray):
-        data = cp.asarray(data, dtype)
-    elif isinstance(data, torch.Tensor):
-        data = cp.asarray(data, dtype)
-    elif isinstance(data, (float, int, bool)):
+    if isinstance(data, (cp_ndarray, np.ndarray, torch.Tensor, float, int, bool)):
         data = cp.asarray(data, dtype)
     # recursive calls
     elif isinstance(data, Sequence) and wrap_sequence:
@@ -248,7 +241,7 @@ def convert_data_type(
     elif has_cp and output_type is cp.ndarray:
         if orig_type is not cp.ndarray:
             data = convert_to_cupy(data, dtype)
-        elif data is not None and dtype != data.dtype:
+        elif data is not None:
             data = data.astype(dtype)
     else:
         raise ValueError(f"Unsupported output type: {output_type}")
--- a/monai/utils/type_conversion.py
+++ b/monai/utils/type_conversion.py
@@ -174,7 +174,7 @@ def convert_to_cupy(data, dtype, wrap_sequence: bool = True):
         data = cp.asarray(data, dtype)
     # recursive calls
     elif isinstance(data, Sequence) and wrap_sequence:
-        return cp.asarray(data)
+        return cp.asarray(data, dtype)
     elif isinstance(data, list):
         return [convert_to_cupy(i, dtype) for i in data]
     elif isinstance(data, tuple):
--- a/monai/utils/type_conversion.py
+++ b/monai/utils/type_conversion.py
@@ -16,6 +16,7 @@ __all__ = [
     "get_equivalent_dtype",
     "convert_data_type",
     "get_dtype",
+    "convert_to_cupy",
     "convert_to_numpy",
     "convert_to_tensor",
     "convert_to_dst_type",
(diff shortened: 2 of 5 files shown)


Nic-Ma:
Hi @drbeh ,

Thanks for your enhancement.
May I know your use case for this `ToCupy` transform? I think `cuCIM` already outputs data in cupy array?

Thanks.

drbeh:
> Hi @drbeh ,
> 
> Thanks for your enhancement.
> May I know your use case for this `ToCupy` transform? I think `cuCIM` already outputs data in cupy array?
> 
> Thanks.

Hi @Nic-Ma, you're right! cuCIM output data in the cupy array but it also expect the input data to be in `cupy.ndarray` and in specific data type (if needed). We already had converting to `cupy.ndarray` and this PR add the ability to define `dtype` for that array.

Nic-Ma:
Hi @drbeh ,

I see, could you please help provide the program of `cuCIM` loading part? I am a little bit confused by `it also expects the input data to be in cupy.ndarray`, I thought `cuCIM` reads data from file names.

Thanks.

drbeh:
> Hi @drbeh ,
> 
> I see, could you please help provide the program of `cuCIM` loading part? I am a little bit confused by `it also expects the input data to be in cupy.ndarray`, I thought `cuCIM` reads data from file names.
> 
> Thanks.

That's a good point. I see where the confusion come from. cuCIM team has developed several transforms that we are going to use them in MONAI in conjunction with other transforms (for instance the torch.Tensor based transforms) via a `CuCIM` wrapper (https://github.com/Project-MONAI/MONAI/pull/2932). Thus, although cucim has data loading capabilities, a user can use transforms independently.

Nic-Ma:
OK, I see, thanks for your explanation.
Maybe you can use `CastToType` transform before `ToCupy` transform to achieve expected dtype?
I think it's better to make the `ToCupy` API align with `ToTensor` and `ToNumpy`, etc.
@wyli @rijobro What do you guys think?

Thanks in advance.

drbeh:
> OK, I see, thanks for your explanation.
> Maybe you can use `CastToType` transform before `ToCupy` transform to achieve expected dtype?
> I think it's better to make the `ToCupy` API align with `ToTensor` and `ToNumpy`, etc.
> @wyli @rijobro What do you guys think?
> 
> Thanks in advance.

Thank you @Nic-Ma for your suggestion. What you are saying would work but will create a redundant copy of the image.
For instance if we have a `numpy.ndarray` of `np.uint8` and use `ToCupy`, it creates a `cupy.ndarray` with dtype `np.uint8`. Then if we call `CastToType(np.float32)`, it will create another  `cupy.ndarray` with dtype `np.float32`. 

This is not efficient specially for large images.I believe we should change the `ToTensor` and `ToNumpy` to include dtype too.

drbeh:
@Nic-Ma @wyli , I'd really appreciate your help in reviewing the cucim related PRs since we are depending on them for cucim integration and digital pathology performance optimization, which is supposed to be done by the end of the week,

Nic-Ma:
Hi @drbeh ,

OK, I see your point, as I said in https://github.com/Project-MONAI/MONAI/pull/2926#discussion_r706524927, to avoid to add more and more args, I will submit a PR to support `kwargs` directly soon.

Thanks.

Nic-Ma:
Submitted ticket to track it: https://github.com/Project-MONAI/MONAI/issues/2951
Thanks for raising this topic.

drbeh:
> Submitted ticket to track it: #2951
> Thanks for raising this topic.

@Nic-Ma specifically for `ToCupy` it only supports `dtype` and does not support `device`, so don't you think that can we have this one without kwargs and only with `dtype`?

monai/transforms/utility/dictionary.py
```diff
@@ -553,15 +553,15 @@ class ToCupyd(MapTransform):
 
     backend = ToCupy.backend
 
-    def __init__(self, keys: KeysCollection, allow_missing_keys: bool = False) -> None:
+    def __init__(self, keys: KeysCollection, allow_missing_keys: bool = False, dtype=None) -> None:
         """
         Args:
             keys: keys of the corresponding items to be transformed.
                 See also: :py:class:`monai.transforms.compose.MapTransform`
```
Nic-Ma, line 560: Could you please help add the doc-string for `dtype` arg?

Thanks.

monai/utils/type_conversion.py
```diff
@@ -154,6 +155,49 @@ def convert_to_numpy(data, wrap_sequence: bool = False):
...
+            Tensor, numpy array, cupy array, float, int, bool are converted to cupy arrays
+
+            for dictionary, list or tuple, convert every item to a numpy array if applicable.
+        wrap_sequence: if `False`, then lists will recursively call this function. E.g., `[1, 2]` -> `[array(1), array(2)]`.
+            If `True`, then `[1, 2]` -> `array([1, 2])`.
+    """
+
+    # direct calls
+    if isinstance(data, cp_ndarray):
+        if dtype is not None:
```
Nic-Ma, line 174: I think you don't need to check `None`.

Thanks.

monai/utils/type_conversion.py
```diff
@@ -154,6 +155,49 @@ def convert_to_numpy(data, wrap_sequence: bool = False):
...
+            data = data.astype(dtype)
+    elif isinstance(data, np.ndarray):
+        data = cp.asarray(data, dtype)
+    elif isinstance(data, torch.Tensor):
+        data = cp.asarray(data, dtype)
+    elif isinstance(data, (float, int, bool)):
+        data = cp.asarray(data, dtype)
+    # recursive calls
+    elif isinstance(data, Sequence) and wrap_sequence:
+        return cp.asarray(data)
```
Nic-Ma, line 184: Seems all the above cases can simplified by `cp.asarray(data, dtype)`.

Thanks.

monai/utils/type_conversion.py
```diff
@@ -199,6 +245,11 @@ def convert_data_type(
             data = convert_to_numpy(data)
         if data is not None and dtype != data.dtype:
             data = data.astype(dtype)
+    elif has_cp and output_type is cp.ndarray:
+        if orig_type is not cp.ndarray:
```
Nic-Ma, line 249: Seems the `convert_to_cupy` API can support `cupy arrary` input, so you don't need to check it here again.
Just:
```py
elif has_cp and output_type is cp.ndarray:
    if data is not None:
        data = convert_to_cupy(data, dtype)
```
Right?

tests/test_to_cupy.py
```diff
@@ -22,49 +22,81 @@
 cp, has_cp = optional_import("cupy")
 
 
+@skipUnless(has_cp, "CuPy is required.")
 class TestToCupy(unittest.TestCase):
-    @skipUnless(has_cp, "CuPy is required.")
     def test_cupy_input(self):
-        test_data = cp.array([[1, 2], [3, 4]])
+        test_data = cp.array([[1, 2], [3, 4]], dtype=cp.float32)
```
Nic-Ma, line 28: I am not familiar with Cupy APIs, may I know how to specify the target `device` for the array?

Thanks.

monai/utils/type_conversion.py
drbeh: that's a good point. thanks

monai/utils/type_conversion.py
drbeh: In cupy, `astype(None)` is equivalnet of `astype('float64')`, so I don't want to change the dtype if it is not intended to.

monai/utils/type_conversion.py
drbeh: Good observation but similar to numpy, in cupy, `astype` will make a copy by default so I don't want to `astype` being called if the type and dtype of input and output are the same. https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html#numpy.ndarray.astype

monai/utils/type_conversion.py
Nic-Ma: Hi @drbeh ,

I checked the doc-string of `cupy.asarray()`, seems `None` means "It is inferred from the input by default":
https://docs.cupy.dev/en/stable/reference/generated/cupy.asarray.html

Thanks.

monai/utils/type_conversion.py
Nic-Ma: I think maybe you mixed `asarray()` and `astype()` here? You can see the doc-string of `asarray()`: "If a is already on the device, no copy is performed."
https://docs.cupy.dev/en/stable/reference/generated/cupy.asarray.html

Thanks.

tests/test_to_cupy.py
drbeh: CuPy arrays are  always on GPU! You can change the cuda device with `cp.cuda.Device(1).use()` or with a context manager!
Basically CuPy array on CPU is a NumPy array!

Nic-Ma: `cp.cuda.Device(1).use()` is what I asked, I know Cupy array is on GPU, I just want to know how to specify GPU device in multi-GPUs case.

Thanks.

monai/utils/type_conversion.py
drbeh: Right, I have `astype` in the code but I can change it to `asarray` thanks.

tests/test_to_cupy.py
drbeh: I see. I like your forward thinking! ;)

drbeh:
Hi @Nic-Ma, thank you very much for your constructive comment. I have applied all of them. Please let me know if there is anything else. thanks

monai/utils/type_conversion.py
```diff
@@ -154,6 +155,42 @@ def convert_to_numpy(data, wrap_sequence: bool = False):
...
+        wrap_sequence: if `False`, then lists will recursively call this function. E.g., `[1, 2]` -> `[array(1), array(2)]`.
+            If `True`, then `[1, 2]` -> `array([1, 2])`.
+    """
+
+    # direct calls
+    if isinstance(data, (cp_ndarray, np.ndarray, torch.Tensor, float, int, bool)):
+        data = cp.asarray(data, dtype)
+    # recursive calls
+    elif isinstance(data, Sequence) and wrap_sequence:
+        return cp.asarray(data)
```
Nic-Ma, line 177: Maybe you should also specify `dtype` for a sequence array.

Nic-Ma (approved):

PR #2954, opened 2021-09-15
2231 Fixes tutorial 353

wyli:
Fixes https://github.com/Project-MONAI/tutorials/issues/353
fixes https://github.com/Project-MONAI/tutorials/issues/352
part of #2231

### Description
- enhance resampler to output tensor/nparray
- enhance testing util to support type check
-  convert_to_dst_type with optional `dtype`

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -1290,7 +1290,7 @@ class Resample(Transform):
         grid: Optional[NdarrayOrTensor] = None,
         mode: Optional[Union[GridSampleMode, str]] = None,
         padding_mode: Optional[Union[GridSamplePadMode, str]] = None,
-    ) -> torch.Tensor:
+    ) -> NdarrayOrTensor:
         """
         Args:
             img: shape must be (num_channels, H, W[, D]).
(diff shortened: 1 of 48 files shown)


monai/transforms/spatial/array.py
```diff
@@ -1344,8 +1344,9 @@ def __call__(
                 padding_mode=self.padding_mode.value if padding_mode is None else GridSamplePadMode(padding_mode).value,
                 align_corners=True,
             )[0]
-
-        return out
+        out_val: NdarrayOrTensor
+        out_val, *_ = convert_to_dst_type(out, dst=img, dtype=out.dtype)
```
Nic-Ma, line 1348: 1. Should we add the `numpy` **backend tag** as it can return numpy now?
2. And do you plan to update all the transforms to make sure they can support `numpy in` -> `numpy out`?

Thanks.

wyli: for 2, sure I'll try to capture these as many as possible.
for 1, I think there is confusion between `backend` and input/output type. in my understanding the`backend` means the actual API used to compute, e.g. `torch.flip` or `np.flip`. in/output type means the data type, a transform can take numpy array as input and using torch backend to compute the output, in this case we need two steps of type converting.

Nic-Ma: For 1, I am OK with both definitions, but we need to make sure all the transforms are aligned. Seems currently we have confused cases, please update them.

Thanks.

wyli:
@Nic-Ma blossom ci passed, the GitHub tests were failing because of resources issue

Nic-Ma (approved):
Thanks for the quick enhancement.

PR #2958, opened 2021-09-15
Torch: `map_binary_to_indices`, `map_classes_to_indices`, `correct_crop_centers`, `generate_pos_neg_label_crop_centers`, `generate_label_classes_crop_centers`

rijobro:
### Description
Torch: `map_binary_to_indices`, `map_classes_to_indices`, `correct_crop_centers`, `generate_pos_neg_label_crop_centers`, `generate_label_classes_crop_centers`

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/croppad/array.py
+++ b/monai/transforms/croppad/array.py
@@ -910,16 +910,18 @@ class RandCropByPosNegLabel(Randomizable, Transform):
         image: Optional[np.ndarray] = None,
     ) -> None:
         self.spatial_size = fall_back_tuple(self.spatial_size, default=label.shape[1:])
+        fg_indices_: np.ndarray
+        bg_indices_: np.ndarray
         if fg_indices is None or bg_indices is None:
             if self.fg_indices is not None and self.bg_indices is not None:
                 fg_indices_ = self.fg_indices
                 bg_indices_ = self.bg_indices
             else:
-                fg_indices_, bg_indices_ = map_binary_to_indices(label, image, self.image_threshold)
+                fg_indices_, bg_indices_ = map_binary_to_indices(label, image, self.image_threshold)  # type: ignore
         else:
             fg_indices_ = fg_indices
             bg_indices_ = bg_indices
-        self.centers = generate_pos_neg_label_crop_centers(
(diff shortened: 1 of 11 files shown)


wyli (approved):
looks good to me, @Nic-Ma this might change the behaviour of the spleen seg. notebook...please confirm

monai/transforms/croppad/array.py
```diff
@@ -910,16 +910,18 @@ def randomize(
         image: Optional[np.ndarray] = None,
     ) -> None:
         self.spatial_size = fall_back_tuple(self.spatial_size, default=label.shape[1:])
+        fg_indices_: np.ndarray
```
Nic-Ma, line 913: May I know why the data type is still `numpy array`? I think you added `Tensor` support?

Thanks.

Nic-Ma (approved):
Looks good to me.
Could you please help double confirm by executing this example?
https://github.com/Project-MONAI/tutorials/blob/master/3d_segmentation/torch/unet_training_dict.py
Just also need to add `EnsureTyped()` after `LoadImaged()` to test with Tensor data.

Thanks in advance.

monai/transforms/croppad/array.py
rijobro: I've updated the helper functions, but not actually updated any of the transforms. Hence, the transforms expect numpy input, and therefore expect numpy output from these helper functions. This PR will then enable me to start updating the transforms that use the helper functions.

rijobro: @Nic-Ma thanks for the suggestion, the example script now runs both normally and with `EnsureTyped` after `LoadImaged`.

Nic-Ma: I see your point, thanks!

Nic-Ma:
Hi @rijobro ,

Would you like to add Tensor support for `RandCropByPosNegLabel` and `RandCropByLabelClasses` with a high priority? As you already updated the help utilities.
And I think this PosNegCrop transform is popular in our training tasks, if we can support Tensor in v0.7, that would be great.

Thanks in advance.

PR #2964, opened 2021-09-16
Fix for Jupyter plotting

ericspod:

Fixes # .

### Description
This fixes an issue with the plotting in the Jupyter utilities relating to collated or decollated batches and having batches greater than 1.

### Status
**Work in progress**

### Types of changes


ericspod:
The current test fail looks like it has something to do with Matplotlib changing how it renders things since the content of the stored and generated images is the same, there are not changes to the plotting functions in this PR.

wyli:
change the batch size will also change the loss curves, the expected image needs an update to match the new curve.

ericspod:
There was a bit of variation in how the text was rendered as well and spacing somehow also changed. After accounting for the batch size these differences are still there.

wyli (approved):

PR #2966, opened 2021-09-17
2231 Enhance tensor transforms

Nic-Ma:
Part of #2231 .

### Description
This PR is a followup, enhanced several Tensor operations.

### Status
**Ready**

### Types of changes


monai/transforms/intensity/array.py
```diff
@@ -969,7 +969,9 @@ def __call__(self, img: NdarrayOrTensor) -> torch.Tensor:
         # add one to transform axis because a batch axis will be added at dimension 0
         savgol_filter = SavitzkyGolayFilter(self.window_length, self.order, self.axis + 1, self.mode)
         # convert to Tensor and add Batch axis expected by HilbertTransform
-        out: torch.Tensor = savgol_filter(self.img_t.unsqueeze(0)).squeeze(0)
+        out = savgol_filter(self.img_t.unsqueeze(0)).squeeze(0)
+        out, *_ = convert_to_dst_type(out, dst=img)
```
rijobro, line 973: Do we need this because we were getting errors without it? As I've previously said, I really want to avoid converting back where possible.

Nic-Ma: Yes, that because we have many use cases to use only 1 transform, instead of a chain. We should not break the previous numpy output behavior.

Thanks.

rijobro (approved):

wyli (approved):

PR #2971, opened 2021-09-17
Torch `GaussianSmooth`, `RandGaussianSmooth`, `GaussianSharpen`, `RandGaussianSharpen`

rijobro:
### Description
Torch `GaussianSmooth`, `RandGaussianSmooth`, `GaussianSharpen`, `RandGaussianSharpen`

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @rijobro ,

I am checking the left transforms, seems these 4 intensity transforms already execute with Tensor data, but the code stype doesn't follow our backend design:
    "GibbsNoise",
    "RandGibbsNoise",
    "KSpaceSpikeNoise",
    "RandKSpaceSpikeNoise",
Do you also plan to add backend support for them in v0.7?

Thanks in advance.

rijobro:
Sure, I'll do them on Monday.

monai/networks/layers/simplelayers.py
```diff
@@ -411,7 +411,7 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         Args:
             x: in shape [Batch, chns, H, W, D].
         """
-        _kernel = [gaussian_1d(s, truncated=self.truncated, approx=self.approx) for s in self.sigma]
+        _kernel = [gaussian_1d(s, truncated=self.truncated, approx=self.approx).to(x.device) for s in self.sigma]
```
wyli, line 414: I think it's the user's responsibility to make sure this module `GaussianFilter` and `x` are on the same device, following pytorch nn.modules' behaviour

monai/transforms/intensity/array.py
```diff
@@ -1030,15 +1030,23 @@ class GaussianSmooth(Transform):
...
+    def __call__(self, img: NdarrayOrTensor) -> torch.Tensor:
+        img_t: torch.Tensor
+        img_t, *_ = convert_data_type(img, torch.Tensor, dtype=torch.float)  # type: ignore
+        sigma: Union[Sequence[torch.Tensor], torch.Tensor]
+        if isinstance(self.sigma, Sequence):
+            sigma = [torch.as_tensor(s, device=img_t.device) for s in self.sigma]
+        else:
+            sigma = torch.as_tensor(self.sigma, device=img_t.device)
+        gaussian_filter = GaussianFilter(img_t.ndim - 1, sigma, approx=self.approx)
+        out: torch.Tensor = gaussian_filter(img_t.unsqueeze(0)).squeeze(0)
```
wyli, line 1048: if this is part of v0.7  we  should convert `out` back to `img`'s original type as we discussed

monai/networks/layers/simplelayers.py
rijobro: Unfortunately `ensure_tuple_rep(sigma, int(spatial_dims))` converts to numpy. Therefore, any input that was on the gpu will be copied to the cpu.

wyli: I see, I feel we need a flag in `ensure_tuple_rep` to control the conversion...

rijobro: Yes you're right, I should have just enhanced the underlying functionality. I'll do that instead.

rijobro:
hi @wyli I'm struggling to replicate the problems in this PR locally. I've tried with pytorch 1.6 and 1.7, with cuda toolkit 11.0 and 10.2, to no avail. Any ideas? https://github.com/Project-MONAI/MONAI/pull/2971/checks?check_run_id=3673884529

wyli (approved):

PR #2973, opened 2021-09-17
Add make the name of wsi reader lowercase

drbeh:
### Description
Makes the input name for WSI reader library lowercase to avoid possible confusion.

### Status
**Ready**

### Types of changes


wyli (approved):
thanks, look good to me

PR #3026, opened 2021-09-24
apply pyupgrade

Borda:
### Description

Update Python notation to be better suitable to python 3.6+
in fact, I have just used the `pyupgrade` nook in pre-commit

### Status

### Types of changes

### DIFF
--- a/monai/_version.py
+++ b/monai/_version.py
@@ -1,4 +1,3 @@
-
 # This file helps to compute a version number in source trees obtained from
 # git-archive tarball (such as those provided by githubs download-from-tag
 # feature). Distribution tarballs (built by setup.py sdist) and build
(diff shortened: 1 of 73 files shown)


monai/_version.py
```diff
@@ -81,7 +80,7 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False,
                                  stderr=(subprocess.PIPE if hide_stderr
                                          else None))
             break
-        except EnvironmentError:
+        except OSError:
             e = sys.exc_info()[1]
```
wyli, line 84: This is an automatically generated file, could you revert?

wyli (commented):
I guess the goal is to do this kind of commit automatically in the future, Not sure whether it is compatible with the other code formatting tools. But the current changes look good. We can experiment it with the development branch for a while first

monai/_version.py
Borda: reverted, and added to exluded files

Borda:
> not sure whether it is compatible with the other code formatting tools

it shall be and another case it shall be caught by your CI :]

wyli (approved):
It looks good to me, thanks!

PR #3038, opened 2021-09-28
2975 torch fgbgtoindices

### LINKED ISSUE #2975: Torch `FgBgToIndices`
BTW, to cache the foreground & background indices for pos / neg crop, could you please help also add Tensor backend for `FgBgToIndices` transform?

Thanks in advance.

_Originally posted by @Nic-Ma in https://github.com/Project-MONAI/MONAI/pull/2967#pullrequestreview-757677818_

wyli:

Fixes #2975

### Description
- ~unravel_index is still in numpy, otherwise it supports both backends~ added both backends

### Status
**Ready**

### Types of changes

### DIFF
--- a/docs/source/transforms.rst
+++ b/docs/source/transforms.rst
@@ -1418,3 +1418,6 @@ Utilities
 ---------
 .. automodule:: monai.transforms.utils
     :members:
+
+.. automodule:: monai.transforms.utils_pytorch_numpy_unification
+    :members:
--- a/monai/transforms/utility/array.py
+++ b/monai/transforms/utility/array.py
@@ -35,7 +35,6 @@ from monai.transforms.utils_pytorch_numpy_unification import in1d, moveaxis, unr
 from monai.utils import (
     convert_data_type,
     convert_to_cupy,
-    convert_to_dst_type,
     convert_to_numpy,
     convert_to_tensor,
     ensure_tuple,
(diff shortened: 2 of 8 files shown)


Nic-Ma:
Mark: training time with this PR around `60s`, and without this PR `56-58s`.

Thanks.

wyli:
> Mark: training time with this PR around `60s`, and without this PR `56-58s`.
> 
> Thanks.

I looked into this issue,  the root cause is that
coordinates were (1) converted into list of numbers:
https://github.com/Project-MONAI/MONAI/blob/28fb7d6af1a74aac1576637403603de1c3d8a59a/monai/transforms/utils.py#L427
and (2) the numbers are converted into tensors:
https://github.com/Project-MONAI/MONAI/blob/28fb7d6af1a74aac1576637403603de1c3d8a59a/monai/transforms/croppad/array.py#L405

Previously `fg_indices`, `bg_indices` were numpy arrays, so there's no step (1) conversion.

So I think the issue is not really in this PR, it's from the `generate_pos_neg_label_crop_centers` and `SpatialCrop` of `RandCropByPosNegLabeld`.

Nic-Ma:
Hi @wyli ,

OK, thanks for your investigation!
Let me try to fix the issue in a separate PR ASAP.

Thanks.

wyli:
> Hi @wyli ,
> 
> OK, thanks for your investigation! Let me try to fix the issue in a separate PR ASAP.
> 
> Thanks.

thanks! please help fix the issue and we can merge this PR after that

Nic-Ma:
Sure, sounds good to me. Investigating.

Thanks.

rijobro:
This looks great, thanks!

Nic-Ma:
I think we already found the root cause of the performance issue, no need to put the `fg / bg indices` on GPU as we actually need to move it to CPU and index image.
So I will merge this PR first, then create another PR to optimize the logic.

Thanks.

Nic-Ma (approved):

PR #3039, opened 2021-09-28
Create transform images

rijobro:
Partly addresses https://github.com/Project-MONAI/MONAI/issues/2721.

### Description
Automatic generation of transform example image and add to the docstring.

image

### Status
**Ready**

### Types of changes


rijobro:
Could do but it would mean that we can't automate it (easily) as we can this way. Current images are 250kB, if we have 100 transforms to visualise that's 25MB. Is that too much? We could downsample by factor 2 to reduce to 6MB, and I'm sure we could play with data types to further reduce. What do you think?

wyli:
I think every time we modify the images it'll create another copy in the git history and increase the size of the `.git` folder, this might not scale...

ericspod:
I think it's best we store this data elsewhere, we've used gdrive and other sources before, we can do that for now at least.

rijobro:
Do we have an official MONAI gdrive (or similar)? I wouldn't want to have it on my personal and then accidentally delete it.

wyli:
> Do we have an official MONAI gdrive (or similar)? I wouldn't want to have it on my personal and then accidentally delete it.

Thanks, I was looking for some other options but couldn't get an idea. Creating a new GitHub repo as you did looks good to me. Further in this direction, perhaps we need a way to dynamically inject the URLs with a global variable? https://stackoverflow.com/questions/57417808/standardizing-a-link-using-a-global-variable-in-sphinx-documentation

rijobro:
Hi @ericspod @wyli
I created a [new repository](https://github.com/Project-MONAI/DocImages) and will store the images there. I won't use it as a git submodule as these can get messy. Instead I'll use an environment variable pointing to that repository. Using git is advantageous in that (a) we can all add to it and (b) the URLs that get created are consistent (unlike gdrive that contains some random characters).

ericspod:
Another repository makes sense. We should keep an eye on how large its .git directory gets over time to see if it's really an issue, Github's LFS is a thing as well that's meant to help manage large files and prevent that kind of issue. We keep talking about it and never trying it.

monai/transforms/utils_create_transform_ims.py
```diff
@@ -0,0 +1,220 @@
...
+    url = "https://www.dropbox.com/s/ndz8qtqblkciole/" + fname + "?dl=1"
+    out_path = os.path.join(cache_dir, "MarsAtlas-MNI-Colin27")
+    zip_path = os.path.join(cache_dir, fname)
+
+    download_and_extract(url, zip_path, out_path)
+
+    image, label = sorted(glob(os.path.join(out_path, "*.nii")))
+
+    data = {CommonKeys.IMAGE: image, CommonKeys.LABEL: label}
+
```
ericspod, line 66: You could have `max_size = max(image.shape)` here and then have the `SpatialPadd` in your transform sequence to eliminate lines 76-79 and replace line 75 with `return transforms(data)`.

monai/transforms/utils_create_transform_ims.py
```diff
@@ -0,0 +1,220 @@
...
+    for i, line in enumerate(contents):
+        # find the line containing start of the transform documentation
+        if "`" + transform_name + "`" in line:
+            doc_start = i
+            break
+    if doc_start is None:
+        raise RuntimeError("Couldn't find transform documentation")
+
+    # if image is already in docs, nothing to do
+    image_line = doc_start + 2
```
ericspod, line 99: This relies on very precise formatting I guess Sphinx enforces for us.

ericspod (approved):

monai/transforms/utils_create_transform_ims.py
rijobro: But I won't know what the max size is until I've loaded the image. Which implies two sets of transforms, right?

wyli (approved):

monai/transforms/utils_create_transform_ims.py
ericspod: Yes true, typically we'd choose a fixed image size for SpatialPadd to avoid this problem.

monai/transforms/utils_create_transform_ims.py
```diff
@@ -0,0 +1,220 @@
...
+    ax.set_axis_off()
+    fig.add_axes(ax)
+    ax.imshow(images, cmap="gray")
+    if labels is not None:
+        ax.imshow(labels, cmap="hsv", alpha=0.9)
+    fig.savefig(filename, dpi=images.shape[0])
+    plt.close(fig)
+
+
+def create_transform_im(transform, data, ndim, update_doc=True, out_dir=None, seed=0):
```
wyli, line 168: @SachidanandAlle mentioned that providing this utility to the users so that they can visualise the effects of the transforms for any input images. Looks like a nice feature to have cc @rijobro @Nic-Ma @ericspod

PR #3045, opened 2021-09-29
Torch `Spacing`, `Spacingd`

rijobro:
### Description
Torch `Spacing`, `Spacingd`

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/nifti_writer.py
+++ b/monai/data/nifti_writer.py
@@ -15,17 +15,19 @@ import numpy as np
 import torch
 
 from monai.config import DtypeLike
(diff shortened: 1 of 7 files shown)


wyli (approved):
thanks, for this one and the orientation transform, we should test it with the image writing functions as well https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_nifti_rw.py#L102-L110

rijobro:
I updated `test_nifti_rw.py` to use torch and torch.cuda as input arrays. Is that what you meant?

PR #3052, opened 2021-09-30
3051 Fix dtype issue in Spacing transform

### LINKED ISSUE #3051: Spacing transform outputs float64 Tensor data
**Describe the bug**
The PR https://github.com/Project-MONAI/MONAI/pull/3045 introduced an issue that it converted the output data dytpe to the **computation dtype**, actually it should always output float32:
https://github.com/Project-MONAI/MONAI/blob/0.7.0/monai/transforms/spatial/array.py#L127
And I think no need to deepcopy the image:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/spatial/array.py#L203

Nic-Ma:
Fixes #3051 .

### Description
This PR fixed the output data type issue of Spacing transform.

### Status
**Ready**

### Types of changes


rijobro:
So the purpose of `_dtype` is to convert the image to that type before the interpolation, and then to convert back to float afterwards?

rijobro (approved):

Nic-Ma:
> So the purpose of `_dtype` is to convert the image to that type before the interpolation, and then to convert back to float afterwards?
> 
> Edit: I just checked the linked issue and think this is the case. Thanks for the quick PR.

Yes, it's just used to adjust the interpolation computation.

Thanks for your review.

PR #3056, opened 2021-10-01
Extra transform examples

rijobro:
### Description
Adds the majority of examples for transforms.

new transforms.html can be seen here (I had to add the .txt extension, so remove that):

https://github.com/Project-MONAI/MONAI/files/7266593/transforms.html.txt

Alternatively, you can see the images (out of context) here:

https://github.com/Project-MONAI/DocImages/tree/main/transforms

### Status

**Ready**

### Types of changes


rijobro:
@wyli I tried modifying the docs/conf.py to include this:

```python
# If there is a + in the version, that means that we're not on a tagged version.
# In this case, we'll use the URL to the `main` branch. If we're on a clean version,
# then use that tag.
if "+" in monai.__version__:
    doc_image_url = "https://github.com/Project-MONAI/DocImages/raw/main/"
else:
    doc_image_url = "https://raw.githubusercontent.com/Project-MONAI/DocImages/" + monai.__version__ + "/"


rst_epilog = f"""
.. |DocImageURL| replace:: {doc_image_url}
"""
```

and then `|DocImageURL|` can be accessed inside the .rst files. Unfortunately, it seems that in sphinx, you can't do variable substitution inside of another directive. So `.. image:: |DocImageURL|transforms/BorderPad.png` is taken as a literal, and the variable is not substituted. See [here](https://stackoverflow.com/questions/20513972/substitution-in-a-file-name-with-restructuredtext-sphinx) for same problem, which doesn't look its easy to side-step...

Nic-Ma:
Hi @rijobro ,

@wyli and I are in the national holiday, will try to review ASAP soon.

Thanks.

Nic-Ma (approved):
Thanks very much for the great work here!
I checked the preview, it looks fantastic.
I have some minor comments:
1. The example of `RandCropByPosNegLabel` seems not correct(it's not 8 samples, and why some black regions?):
image
2. Could you please also add example image to `KeepLargestConnectedComponent`? I think it's an important transform and not very straight-forward for users.

Thanks.

Nic-Ma:
BTW, maybe I missed some context discussion, may I know why the `dict` transforms use colorful images and `array` transforms use `black & white` images?

Thanks.

wyli (approved):
thanks, it looks good to me.

rijobro:
@Nic-Ma I used colour for the dictionary transforms because they have the segmentations overlayed. The underlying MR image is black and white in both cases. Let me know if you find it unclear and we can discuss something better.

rijobro:
@Nic-Ma , and you're right. There's something wrong with all the num_samples=8, I'll check that.

Nic-Ma:
> @Nic-Ma I used colour for the dictionary transforms because they have the segmentations overlayed. The underlying MR image is black and white in both cases. Let me know if you find it unclear and we can discuss something better.

Got it, thanks!

Nic-Ma:
> @Nic-Ma , and you're right. There's something wrong with all the num_samples=8, I'll check that.

Sure, please also take a look other `crop-samples` transforms, maybe they are also not correct.

Thanks.

PR #3068, opened 2021-10-04
correct panelled example images

rijobro:
### Description
Fixes the panelling problem for transform examples. Doc already updated, can be seen here: https://docs.monai.io/en/latest/transforms.html#randspatialcropsamplesd.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/utils_create_transform_ims.py
+++ b/monai/transforms/utils_create_transform_ims.py
@@ -302,7 +302,7 @@ def save_image(images, labels, filename, transform_name, transform_args, shapes,
             ax.set_yticks([])
             ax.set_frame_on(False)
             if labels is not None:
-                ax.imshow(labels[row][col], cmap="hsv", alpha=0.9)
+                ax.imshow(labels[row][col], cmap="hsv", alpha=0.9, interpolation="nearest")
     # title is e.g., Flipd(keys=keys, spatial_axis=0)
     title = transform_name + "("
     for k, v in transform_args.items():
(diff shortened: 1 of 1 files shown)


wyli (approved):
Thanks, it looks good to me.

monai/transforms/utils_create_transform_ims.py
```diff
@@ -280,22 +279,24 @@ def get_stacked_before_after(before, after, is_label=False):
...
+    ncol = len(images[0])  # num orthogonal views (either 1 or 3)
+    fig, axes = plt.subplots(nrow, ncol)
+    for row in range(nrow):
         vmin = min(i.min() for i in images[row])
         vmax = max(i.max() for i in images[row])
-        for col in range(len(images[0])):
-            ax = axes[row][col]
+        for col in range(ncol):
+            ax = axes[row][col] if ncol > 1 else axes[row]
             imshow = ax.imshow(images[row][col], cmap="gray", vmin=vmin, vmax=vmax)
```
wyli, line 294: might have some minor issues - the padding transform figure doesn't show the actual size of the images, looks like both before and after are scaled to the same panel size. - for the label vis, maybe the option of no antialiasing should be used to keep the values discrete?

rijobro: I struggled to get the sizes of the figures to reflect the sizes of the image. Instead I displayed the image sizes in the y_label, see [here](https://docs.monai.io/en/latest/transforms.html#borderpadd). I messed with the `gridspec_kw` argument to `imshow`, but didn't get perfect results. Feel free to have a look though. 

Good idea for the anti-aliasing.

PR #3083, opened 2021-10-08
2721 - enhance transform figures

wyli:

part of #2721

### Description
- adds a rough estimation for the height ratios when comparing before/after
- adds a few labels for the right-most col if there's no colorbar

RandSpatialCropSamples
CenterSpatialCropd

### Status
**Ready**

### Types of changes


wyli:
couldn't make the size of subplot panels to be exactly proportional to the actual image sizes @rijobro, so I put some yticks here and I guess they are helpful.

PR #3138, opened 2021-10-15
Adding Torchscript utility functions

ericspod:

### Description
Adding metadata to saved Torchscript objects was dicussed as a useful feature with the Deploy team. The intent is to define saved networks that are self-describing, stating what their inputs and outputs are, what data requirements they, how to interpret their outputs, etc. Other information such as timestamps, versions, names, plain-language descriptions, and other important scientific descriptions can also be added. The mechanism works by saving a mapping of items as a JSON file in the Torchscript object, Pytorch provides the facilities to do this through the standard API calls so there's no need to manipulate the zip files directly.

### Status
**Work in progress**

### Types of changes


dbericat:
@cpbridge related to metadata conversation

ericspod:
Our related [conversation on model data format](https://github.com/Project-MONAI/MONAI/issues/3482).

monai/data/torchscript_utils.py
```diff
@@ -0,0 +1,122 @@
...
+METADATA_FILENAME = "metadata.json"
+
+
+def save_net_with_metadata(
+    jit_obj: torch.nn.Module,
+    filename_prefix_or_stream: Union[str, IO[Any]],
+    include_config_vals: bool = True,
+    append_timestamp: bool = False,
+    meta_values: Optional[Mapping[str, Any]] = None,
+    more_extra_files: Optional[Mapping[str, bytes]] = None,
```
Nic-Ma, line 31: Could you please help also add some unit tests for `more_extra_files`?
Better to have more description in the doc-string.

Thanks.

Nic-Ma:
Hi @ericspod ,

Thanks for your great proposal here, I think this PR overall looks good to me.
I already tried to use in the draft example PR: https://github.com/Project-MONAI/tutorials/pull/487
I see you marked it as `draft`, could you please help complete it then let's enable the feature?
We can enhance it if we have more ideas later.
@wyli @rijobro What do you guys think?

Thanks.

wyli:
yes, these utilities are probably necessary, and we should create a new submodule under 'monai' for https://github.com/Project-MONAI/MONAI/issues/3482, what would be the module name?

Nic-Ma:
> yes, these utilities are probably necessary, and we should create a new submodule under 'monai' for #3482, what would be the module name?

Hi @wyli ,

I totally agree with you, actually I also thought where to put these utils, including:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/networks/utils.py#L432
As they are also utils, how about `monai.utils.jit.py` or `monai.utils.jitscript.py`?

Thanks.

ericspod:
We're  still discussing  what the model package would look like so it might
not use the mechanism I proposed here. I can  finalize this pr anyway if
you'd like though today I'm traveling.

monai/data/torchscript_utils.py
```diff
@@ -0,0 +1,149 @@
...
+        metadict.update(get_config_values())
+        metadict[JITMetadataKeys.TIMESTAMP.value] = now.astimezone().isoformat()
+
+    if meta_values is not None:
+        metadict.update(meta_values)
+
+    json_data = json.dumps(metadict)
+
+    # Pytorch>1.6 can use dictionaries directly, otherwise need to use special map object
+    if pytorch_after(1, 7):
```
Nic-Ma, line 84: I think maybe some typo here, the code is `>1.7` but comment says `>1.6`?

Thanks.

monai/data/torchscript_utils.py
```diff
@@ -0,0 +1,149 @@
...
+        # save the Torchscript bundle with the above dictionary stored as an extra file
+        save_net_with_metadata(m, "test", meta_values=meta)
+
+        # load the network back, `loaded_meta` has same data as `meta` plus version information
+        loaded_net, loaded_meta, _ = load_net_with_metadata("test.pt")
+
+
+    Args:
+        jit_obj: object to save, should be generated by `script` or `trace`.
+        filename_prefix_or_stream: filename or file-like stream object, if filename has no extension it becomes .pt.
```
Nic-Ma, line 64: Suggest to add "``" mark: `.pt`.

Thanks.

monai/data/torchscript_utils.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Args:
+        filename_prefix_or_stream: filename or file-like stream object.
+        map_location: network map location as in `torch.jit.load`.
+        more_extra_files: other extra file data names to load from bundle, see `_extra_files` of `torch.jit.load`.
+    Returns:
+        Triple containing loaded object, metadata dict, and extra files dict containing other file data if present
+    """
+    # Pytorch>1.6 can use dictionaries directly, otherwise need to use special map object
+    if pytorch_after(1, 7):
```
Nic-Ma, line 127: Same comment as `save` function.

Thanks.

monai/data/torchscript_utils.py
```diff
@@ -0,0 +1,149 @@
+# Copyright 2020 - 2021 MONAI Consortium
```
Nic-Ma, line 1: Hi @ericspod ,

@wyli and I think maybe it's better to move this utility to other folder, like: https://github.com/Project-MONAI/MONAI/pull/3138#issuecomment-995556255.

What do you think?

Thanks.

Nic-Ma (approved):
Thanks for the quick update.
It overall looks good to me, put several minor comments inline.

Thanks.

monai/data/torchscript_utils.py
ericspod: For now we don't have enough code to justify that. There's a bit of code in network utils that may make sense to concentrate here if we wanted to have all jit-related routines in one place.

Nic-Ma: ok, thanks for the explanation.

monai/data/torchscript_utils.py
Nic-Ma: Hi @ericspod , could you please help confirm it's 1.6 or 1.7?
Others look to me.

Thanks.

ericspod: The `pytorch_after(1, 7)` is True for any version at 1.7 and up, it's versions below that which require the special code.

Nic-Ma: OK, sure, thanks for the explanation.
I was confused by the comment `Pytorch>1.6`, maybe `Pytorch>=1.7` is more clear with the code..

Thanks.

Issue #2929, opened 2021-09-12
RandZoom on batch


drbeh:
**Is your feature request related to a problem? Please describe.**
`RandZoom` uses `Zoom` which uses [`torch.nn.functional.interpolate`](https://pytorch.org/docs/stable/generated/torch.nn.functional.interpolate.html) under the hood to extract randomly zoomed images. Although `torch.nn.functional.interpolate` is capable of being applied on batch x channels x [optional depth] x [optional height] x width, `Zoom` and RandZoom` don't.

Instead, they treat the channel dimension as spatial dimension. This is causing an issue since there is no error and warning if user apply it on a batched dataset.

This requires to distinguish between 3D multi-channel images vs 2D multi-channel batched images. Although an automatic logic would be hard to achieve, a simple flag can do the work.

wyli:
the concept of minibatch is part of the model optimization using stochastic gradient descent, anything that dealing with minibatches should be in the 'monai.networks' module of monai.

The price for mixing the concepts of batch and nonbatch is not just a simple flag, it requires a full set of dimensionality input validation, and potentially reduces the usability/code readability.

drbeh:
Hi @wyli,  isn't it a too restrictive assumption? 
What about validation and testing (inference in general)? There is no optimization there but mini-batch is required for computational efficiency.

drbeh:
Do we have any decision here? 
Is there any technical issue that prevents us from supporting batch transform for RandZoom? 
Thanks

PR #3141, opened 2021-10-17
cupy support for stain normalization

drbeh:
### Description
This PR implements CuPy support for stain normalization in pathology transforms.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @drbeh ,

I have a question that: as we are updating all the MONAI transforms to support both `numpy array` and `PyTorch Tensor`, is it possible to use `PyTorch Tensor` in the pathology transforms instead of CuPy?
@ericspod @wyli @rijobro , what do you guys think?

Thanks.

drbeh:
@Nic-Ma, I understand that you want to have everything work with `torch.Tensor` and I agree with you that it is a good thing to have. This is the blocker for using torch transforms: https://github.com/Project-MONAI/MONAI/issues/2929

PR #3226, opened 2021-10-31
Fix 3 issues in CSVIterableDataset

Nic-Ma:
### Description
This PR fixed 3 issues in the `CSVIterableDataset`:
1. It only returns 1 item in `iter`, changed `return` to `yield from` to fix it.
2. When `shuffle=False`, the output order is inverse.
3. `length` of buffer is changing during pop, so need to compute in runtime.

Also added more tests to cover the issues.

### Status
**Ready**

### Types of changes


tests/test_csv_iterable_dataset.py
```diff
@@ -158,23 +176,29 @@ def prepare_csv_file(data, filepath):
                 [3.7725, 4.2118, 4.6353, 5.298, 9.5451],
                 [2.0078, 2.2902, 2.0549, 3.0196, 3.8078],
                 [6.4275, 6.2549, 5.9765, 6.2627, 7.7176],
-                [6.8392, 6.4745, 5.8627, 5.1922, 5.2745],
+                [3.3333, 3.2353, 3.4000, 3.1647, 3.0863],
```
Nic-Ma, line 179: This is aligned with the previous test results:
https://github.com/Project-MONAI/MONAI/blob/0.7.0/tests/test_csv_iterable_dataset.py#L157

Thanks.

wyli (approved):
Thanks, it looks good to me.

PR #3237, opened 2021-11-02
MIL component to extract patches

### LINKED ISSUE #3162: Split into Tiles
The `SplitOnGrid` transform in pathology application splits an image into patches give a `grid_shape` [and optional `patch_size`]. However, there are applications where we need to split an image into patches of a specific size for the entire image (tiling) without providing the grid shape.

myron:
Fixes #3162

This is a component for MIL to split the large 2D images into patches/tiles in numpy (ndarray views)

Some open questions, where is the best location for it? under apps/pathology ?
There is a somewhat similar named function (SplitOnGrid) #2879 , but it operates on Tensors, and it's for smaller splits (3x3 ?) and has fewer options. But still, people may get confused


drbeh:
@myron, this components should go here: https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/pathology/transforms/spatial/array.py

`SplitOnGrid` splits the image into patches based on a defined grid for instance 20x20. Although it accepts patch size but the grid is not defined based on that, and allows for cropping and overlapping patches.

I suggest to have something called `SplitIntoTiles` which defines the grid based on patch size, or extend the `SplitOnGrid` to include the option for calculating the grid based on a fixed image size.

myron:
I think I'll need a separate transform, since SplitOnGrid is only for pt tensorts, and mine is only for numpy

Nic-Ma:
Hi @myron @drbeh ,

Would you like to submit a new transform(array level and dict level)? It's better to make sure:
1. Both numpy array and PyTorch Tensor data.
2. Unit tests for all the cases.
3. Add **inverse** operation in the `dict` level transform.
4. [Optional] Add example transform image in the MONAI API doc, like: https://docs.monai.io/en/latest/transforms.html#spatialpad
I can help add the example image if you don't have enough effort.

Thanks in advance.

myron:
> Hi @myron @drbeh ,
> 
> Would you like to submit a new transform(array level and dict level)? It's better to make sure:
> 
>     1. Both numpy array and PyTorch Tensor data.
> 
>     2. Unit tests for all the cases.
> 
>     3. Add **inverse** operation in the `dict` level transform.
> 
>     4. [Optional] Add example transform image in the MONAI API doc, like: https://docs.monai.io/en/latest/transforms.html#spatialpad
>        I can help add the example image if you don't have enough effort.
> 
> 
> Thanks in advance.

Hmm, I can't do that much. 
1) it uses a special convenience function in numpy, which doesn't exist for tensors (so it only works in numpy) 
2) unit tests I can do
3) I'm not sure how to do inverse, it may not be possible (also it is not necessary)
4) hmm, I'm not sure if we have the rights to publically post these WSI images

Nic-Ma:
Hi @myron ,

Sure, please just develop a transform with numpy backend first.
@drbeh or I can work on the following actions.

Thanks.

myron:
I've updated this transform, please take a look

drbeh (changes requested):
Can you add `unittests` and change this PR to a regular PR after that so people can review it? Thanks

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +74,134 @@ def get_params(self, image_size):
...
+            Defaults to ``False``.
+        tile_pad_full: pad image to the size, evenly divisible by tile_size
+            Defaults to ``False``.
+        tile_background_val: the background constant (e.g. 255 for white background)
+            Defaults to ``255``.
+        tile_filter_mode: [min, max, None] filter tiles at the end. If total number of tiles is more then tile_size,
+            then sort by intensity sum, and take smallest (for min), largest (for max) or random (for None)
+            Defaults to ``min``.
+
+        Note: expects input image to be WxHx3 ndarray, and outputs tile_sizex3xWxH
```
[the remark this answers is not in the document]
myron: ok

myron:
I've updated this transform

- added option to return a list of dicts
- added unit tests (all passed)

please review

monai/apps/pathology/transforms/spatial/array.py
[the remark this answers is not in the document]
myron: yes, I changed that

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,133 @@ def get_params(self, image_size):
...
         return patch_size, steps
+
+
+class TileOnGrid(Transform):
+    """
+    Tile the image into patches on a grid and maintain a subset of it.
+    This transform works only with np.ndarray inputs.
+
+    Args:
+        tile_count: number of tiles to extract
```
Nic-Ma, line 86: As this transform name is `TileOnGrid`, I think we don't need the prefix "tile_" in every arg name?

Thanks.

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,133 @@ def get_params(self, image_size):
...
+        self,
+        tile_count: int = None,
+        tile_size: int = 256,
+        tile_step: int = None,
+        tile_all: bool = False,
+        tile_random_offset: bool = False,
+        tile_pad_full: bool = False,
+        tile_background_val: int = 255,
+        tile_filter_mode: str = "min",  # None, min, max
+        random_seed: int = None,
```
Nic-Ma, line 117: If your transform is random transform, please extends from MONAI `Randomizable` and do random logic in the `randomize()` function:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/transform.py#L122
I think you can refer to `RandSpatialCrop`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/croppad/array.py#L479

Thanks.

monai/apps/pathology/transforms/spatial/dictionary.py
```diff
@@ -53,4 +54,80 @@ def __call__(self, data: Mapping[Hashable, torch.Tensor]) -> Dict[Hashable, torc
...
+
+    def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.ndarray]:
+
+        d = dict(data)
+        for key in self.key_iterator(d):
+            d[key] = self.splitter(d[key])
+
+            if self.return_list_of_dicts:
+                image = d.pop(key, None)  # remove this key
+                d = [{**d, **{key: image[i]}} for i in range(len(image))]
```
Nic-Ma, line 127: I feel there are 2 issues here:
1. Should use `deepcopy` to copy other things into every item of the list.
2. If I have multiple keys, this line seems can't generate list of dicts as expected?

I think you can refer to:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/croppad/dictionary.py#L1160

Thanks.

tests/test_tile_on_grid_dict.py
```diff
@@ -0,0 +1,103 @@
...
+            "tile_count": 16,
+            "tile_size": 8,
+            "tile_filter_mode": None,
+            "tile_random_offset": True,
+            "random_seed": 123,
+        }
+    ]
+)
+
+# for tile_count in [16, 64]:
```
Nic-Ma, line 33: Please remove the test code.

Thanks.

tests/test_tile_on_grid.py
```diff
@@ -0,0 +1,96 @@
...
+
+import unittest
+
+import numpy as np
+from parameterized import parameterized
+
+from monai.apps.pathology.transforms import TileOnGrid
+
+TEST_CASES = []
+# TEST_CASES.append([{'tile_count': 16, 'tile_size': 8, 'tile_filter_mode': None, 'tile_random_offset': True, 'random_seed': 123}])
```
Nic-Ma, line 20: Please remove the test code.

Thanks.

monai/apps/pathology/transforms/spatial/dictionary.py
myron: 1. we don't need deep copy - because other keys will be used only from the first Dict later (and all the other dismissed) in collate function.  you sure you want deepcopy?
2. Multiple keys for Tiling this way are not supported, only a single key will be supported.  (that's the plan).  If we try to support multiple keys, each key can potentially have a different number of tiles in nd.array.  So it's not clear how to split them into a single list of dicts

Nic-Ma: Hi @myron ,

1. In your current tutorial use case, it doesn't need to `deepcopy` the other keys. But if other users try to apply this new transform in other tasks, I think the shallow-copy may cause unknown errors.
2. If it truly can only support 1 key here, please add some check for that: if `allow_missing_keys is False` and keys > 1, raise exception to nofity users it only supports 1 key. And no need to execute for-loop for keys anymore.

Thanks.

monai/apps/pathology/transforms/spatial/array.py
myron: I like it more this way.  "tile_size" is more explicit than var "size", and "tile_count" is more descriptive than just "count".   Is there a reason we can't keep it this way?

tests/test_tile_on_grid_dict.py
myron: done

tests/test_tile_on_grid.py
myron: done

monai/apps/pathology/transforms/spatial/dictionary.py
myron: ok, I've added deepcopy, and made it work for any number of keys

monai/apps/pathology/transforms/spatial/array.py
myron: added Randomizable, and changed unit tests

myron:
I've made all the corrections, please check

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,153 @@ def get_params(self, image_size):
...
+            self.random_idxs = self.R.choice(range(tile_total), self.tile_count, replace=False)
+        else:
+            self.random_idxs = None
+
+    def __call__(self, image: np.ndarray) -> np.ndarray:
+
+        # add random offset
+        self.randomize(img_size=image.shape)
+
+        if self.tile_random_offset and self.random_offset is not None:
```
Nic-Ma, line 165:
```py
if self.tile_random_offset is not None and self.random_offset is not None:
```

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,153 @@ def get_params(self, image_size):
...
+            Defaults to None (same as tile_size)
+        tile_all: Extract all non-background tiles, instead of tile_count.
+            Defaults to ``False``.
+        tile_random_offset: Randomize position of tile grid, instead of starting from the top-left corner
+            Defaults to ``False``.
+        tile_pad_full: pad image to the size evenly divisible by tile_size
+            Defaults to ``False``.
+        tile_background_val: the background constant (e.g. 255 for white background)
+            Defaults to ``255``.
+        tile_filter_mode: [min, max, None] filter tiles. If total number of tiles is more then tile_size,
```
Nic-Ma, line 99: May change to: mode must be in `["min", "max", None]`.

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,153 @@ def get_params(self, image_size):
...
+        image = llw.reshape(-1, clen, xsize, ysize)
+
+        # if keep all patches
+        if self.tile_all:
+            # retain only patches with significant foreground content to speed up inference
+            # FYI, this returns a variable number of tiles, so the batch_size much be 1 (per gpu). Used during inference
+            thresh = 0.999 * 3 * self.tile_background_val * self.tile_size * self.tile_size
+            if self.tile_filter_mode == "min":
+                # default, keep non-background tiles (small values)
+                idxs = np.argwhere(image.sum(axis=(1, 2, 3)) < thresh)
```
Nic-Ma, line 199: Can this transform work for both 2D and 3D data?

Thanks.

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,153 @@ def get_params(self, image_size):
...
+
+        if self.tile_random_offset and self.random_offset is not None:
+            image = image[:, self.random_offset[0] :, self.random_offset[1] :]
+
+        # pad to full size, divisible by tile_size
+        if self.tile_pad_full:
+            c, h, w = image.shape
+            pad_h = (self.tile_size - h % self.tile_size) % self.tile_size
+            pad_w = (self.tile_size - w % self.tile_size) % self.tile_size
+            image = np.pad(
```
Nic-Ma, line 173: I would suggest to use MONAI `Pad` transform:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/croppad/array.py#L72

Thanks.

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,153 @@ def get_params(self, image_size):
...
+                elif self.tile_filter_mode == "max":
+                    idxs = np.argsort(image.sum(axis=(1, 2, 3)))[-self.tile_count :]
+                    image = image[idxs]
+                elif len(image) > self.tile_count:
+                    # random subset (more appropriate for WSIs without distinct background)
+                    if self.random_idxs is not None:
+                        image = image[self.random_idxs]
+
+            else:
+                image = np.pad(
```
Nic-Ma, line 221: I would suggest to use MONAI `Pad` transform:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/croppad/array.py#L72

Thanks.

tests/test_tile_on_grid.py
Nic-Ma: I think the debug tests are still here?

Thanks.

monai/apps/pathology/transforms/spatial/array.py
Nic-Ma: Hi @ericspod @wyli ,

What do you guys think about this naming here?

Thanks in advance.

ericspod: I think it's better to have more information in the argument names but if they all have the same prefix it doesn't add any new information about what they are or do. If the prefix was used to differentiate arguments (eg. different size parameters) it would make sense to have a prefix for each, but here if you take away `tile_` nothing is lost because the context of the arguments being related to tiles is still present. I would change `tile_all` to `extract_all` for example, that states the verb and the noun related to what's going on.

monai/apps/pathology/transforms/spatial/array.py
myron: tile_random_offset is Bool, my version is correct

monai/apps/pathology/transforms/spatial/array.py
myron: No, it's currently for 2D only.  We don't have a reason to extend it to 3D, and scope of testing will be large

monai/apps/pathology/transforms/spatial/array.py
myron: I don't see a reason for Pad here.  A simple numpy.pad is concise ,  and easy to read.

monai/apps/pathology/transforms/spatial/array.py
myron: I don't see any reason to use Pad (from monai) here.  A simple np.pad does the job and is concise.

monai/apps/pathology/transforms/spatial/array.py
myron: Thanks for participating in our silly discussion. but I still prefer my names.  Tile is also a verb (to tile or to extract).   For this particular argument I don't mind changing it, but I also don't see a reason for it. For other arguments  "tile_count" is much more descriptive than "count", or "tile_size" is much more informative than "size". It's okay to have redundancy in names, redundancy help to avoid potential bugs.  If I have to search (in the whole Monai project for tile_count  it will immediately find this function,  if I have to search for keyword "count" or "extract_count" to see where else it's used, it may return me a long list of results, that's hard to analyze.

monai/apps/pathology/transforms/spatial/array.py
myron: ok

monai/apps/pathology/transforms/spatial/array.py
Nic-Ma: Could you please help add some description in the doc-string of this transform for the input shape assumption? Because most other MONAI transforms support both 2D and 3D data.

Thanks.

Nic-Ma (approved):
Thanks for the quick updates.
It looks good to me now.
@drbeh Do you have any other comments? I see you marked "request change" in the PR.

Thanks.

Nic-Ma:
And BTW, @myron please fix the mypy code format check issues:
```
monai/apps/pathology/transforms/spatial/array.py:107:27: error: Incompatible default for argument "tile_count" (default has type "None", argument has type "int")  [assignment]
monai/apps/pathology/transforms/spatial/array.py:109:26: error: Incompatible default for argument "tile_step" (default has type "None", argument has type "int")  [assignment]
monai/apps/pathology/transforms/spatial/array.py:131:51: error: Incompatible default for argument "img_size" (default has type "None", argument has type "Sequence[int]")  [assignment]
monai/apps/pathology/transforms/spatial/array.py:139:38: error: Incompatible types in assignment (expression has type "Tuple[int, int]", variable has type "None")  [assignment]
monai/apps/pathology/transforms/spatial/array.py:140:25: error: Value of type "None" is not indexable  [index]
monai/apps/pathology/transforms/spatial/array.py:141:25: error: Value of type "None" is not indexable  [index]
monai/apps/pathology/transforms/spatial/array.py:156:32: error: Incompatible types in assignment (expression has type "ndarray[Any, Any]", variable has type "None")  [assignment]
monai/apps/pathology/transforms/spatial/dictionary.py:88:27: error: Incompatible default for argument "tile_count" (default has type "None", argument has type "int")  [assignment]
monai/apps/pathology/transforms/spatial/dictionary.py:90:26: error: Incompatible default for argument "tile_step" (default has type "None", argument has type "int")  [assignment]
monai/apps/pathology/transforms/spatial/dictionary.py:116:21: error: Incompatible types in assignment (expression has type "int", variable has type "None")  [assignment]
monai/apps/pathology/transforms/spatial/dictionary.py:134:17: error: Incompatible types in assignment (expression has type "List[Dict[Hashable, Any]]", variable has type "Dict[Hashable, ndarray[Any, Any]]")  [assignment]
tests/test_tile_on_grid_dict.py:57:96: error: Incompatible default for argument "tile_filter_mode" (default has type "None", argument has type "str")  [assignment]
tests/test_tile_on_grid_dict.py:80:13: error: Incompatible types in assignment (expression has type "ndarray[Any, Any]", variable has type "List[Any]")  [assignment]
tests/test_tile_on_grid_dict.py:83:34: error: "List[Any]" has no attribute "sum"  [attr-defined]
tests/test_tile_on_grid.py:57:96: error: Incompatible default for argument "tile_filter_mode" (default has type "None", argument has type "str")  [assignment]
tests/test_tile_on_grid.py:80:13: error: Incompatible types in assignment (expression has type "ndarray[Any, Any]", variable has type "List[Any]")  [assignment]
tests/test_tile_on_grid.py:83:34: error: "List[Any]" has no attribute "sum"  [attr-defined]
```

Thanks.

monai/apps/pathology/transforms/spatial/array.py
wyli: Since we are on the topic of multiple instance learning, why don't we use the widely used terms such as bags and instances? Why do we introduce patch/grid/tile? To me these are unnecessarily confusing

myron: this transform is generic, and can be used outside of MIL.  bags and instances will be very confusing terms here.  this transform simply splits the image into square patches.  to split or to tile both sounds like good names, but "split" is already taken

monai/apps/pathology/transforms/spatial/array.py
myron: done

myron:
> And BTW, @myron please fix the mypy code format check issues:
> 
> ```
> monai/apps/pathology/transforms/spatial/array.py:107:27: error: Incompatible default for argument "tile_count" (default has type "None", argument has type "int")  [assignment]
> monai/apps/pathology/transforms/spatial/array.py:109:26: error: Incompatible default for argument "tile_step" (default has type "None", argument has type "int")  [assignment]
> monai/apps/pathology/transforms/spatial/array.py:131:51: error: Incompatible default for argument "img_size" (default has type "None", argument has type "Sequence[int]")  [assignment]
> monai/apps/pathology/transforms/spatial/array.py:139:38: error: Incompatible types in assignment (expression has type "Tuple[int, int]", variable has type "None")  [assignment]
> monai/apps/pathology/transforms/spatial/array.py:140:25: error: Value of type "None" is not indexable  [index]
> monai/apps/pathology/transforms/spatial/array.py:141:25: error: Value of type "None" is not indexable  [index]
> monai/apps/pathology/transforms/spatial/array.py:156:32: error: Incompatible types in assignment (expression has type "ndarray[Any, Any]", variable has type "None")  [assignment]
> monai/apps/pathology/transforms/spatial/dictionary.py:88:27: error: Incompatible default for argument "tile_count" (default has type "None", argument has type "int")  [assignment]
> monai/apps/pathology/transforms/spatial/dictionary.py:90:26: error: Incompatible default for argument "tile_step" (default has type "None", argument has type "int")  [assignment]
> monai/apps/pathology/transforms/spatial/dictionary.py:116:21: error: Incompatible types in assignment (expression has type "int", variable has type "None")  [assignment]
> monai/apps/pathology/transforms/spatial/dictionary.py:134:17: error: Incompatible types in assignment (expression has type "List[Dict[Hashable, Any]]", variable has type "Dict[Hashable, ndarray[Any, Any]]")  [assignment]
> tests/test_tile_on_grid_dict.py:57:96: error: Incompatible default for argument "tile_filter_mode" (default has type "None", argument has type "str")  [assignment]
> tests/test_tile_on_grid_dict.py:80:13: error: Incompatible types in assignment (expression has type "ndarray[Any, Any]", variable has type "List[Any]")  [assignment]
> tests/test_tile_on_grid_dict.py:83:34: error: "List[Any]" has no attribute "sum"  [attr-defined]
> tests/test_tile_on_grid.py:57:96: error: Incompatible default for argument "tile_filter_mode" (default has type "None", argument has type "str")  [assignment]
> tests/test_tile_on_grid.py:80:13: error: Incompatible types in assignment (expression has type "ndarray[Any, Any]", variable has type "List[Any]")  [assignment]
> tests/test_tile_on_grid.py:83:34: error: "List[Any]" has no attribute "sum"  [attr-defined]
> ```
> 
> Thanks.

I've fixed mypy issues

monai/apps/pathology/transforms/spatial/array.py
wyli: Picking the top n instance out of a bag is already implying the mil assumption. 

Also, tile means a completely different thing in numpy, this transform happens to handle only numpy, it is confusing.

myron: so do you want to rename the transform? or rename the arguments? 
please suggest the names

wyli: Perhaps "ImageToBag"? Converting an image to a set of patch instances based on a spatial grid sampling strategy. Then we can have some parameters to define grid properties and the size of a bag.

drbeh:
Thank you @myron. Let me reviewed it again before moving forward and please try not to use force-push. Thanks

Nic-Ma (approved):

monai/apps/pathology/transforms/spatial/array.py
myron: alright, I think we found a comporimise

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,159 @@ def get_params(self, image_size):
...
+            self.random_idxs = self.R.choice(range(tile_total), self.tile_count, replace=False)  # type: ignore
+        else:
+            self.random_idxs = [0]  # type: ignore
+
+    def __call__(self, image: np.ndarray) -> np.ndarray:
+
+        # add random offset
+        self.randomize(img_size=image.shape)
+        # tile_count: int = self.tile_count  # type: ignore
+        tile_step: int = self.step  # type: ignore
```
drbeh, line 169: The same as previous comment:
I don't get the point of this assignment. Why no using self.step instead of tile_step?

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,159 @@ def get_params(self, image_size):
...
+            w = w + pad_w
+
+        h_n = (h - self.tile_size + tile_step) // tile_step
+        w_n = (w - self.tile_size + tile_step) // tile_step
+        tile_total = h_n * w_n
+
+        if self.tile_count is not None and tile_total > self.tile_count:
+            self.random_idxs = self.R.choice(range(tile_total), self.tile_count, replace=False)  # type: ignore
+        else:
+            self.random_idxs = [0]  # type: ignore
```
drbeh, line 162: Also you have too many `# type: ignore`. It should only be used in special cases, which usually caused by type hinting of external libraries. Are you sure you cannot solve this in a more straight way?

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,159 @@ def get_params(self, image_size):
...
+            pad_w = (self.tile_size - w % self.tile_size) % self.tile_size
+            h = h + pad_h
+            w = w + pad_w
+
+        h_n = (h - self.tile_size + tile_step) // tile_step
+        w_n = (w - self.tile_size + tile_step) // tile_step
+        tile_total = h_n * w_n
+
+        if self.tile_count is not None and tile_total > self.tile_count:
+            self.random_idxs = self.R.choice(range(tile_total), self.tile_count, replace=False)  # type: ignore
```
drbeh, line 160: What is the type error that you are getting here?
I can guess that `choice` is returning `ndarray` and you have initialized `self.random_idxs` with list. So they are not of the same type. If that's a problem, instead of ignoring the error, try to use `tolist` method to convert ndarray to list.

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,159 @@ def get_params(self, image_size):
...
+        else:
+            self.random_idxs = [0]  # type: ignore
+
+    def __call__(self, image: np.ndarray) -> np.ndarray:
+
+        # add random offset
+        self.randomize(img_size=image.shape)
+        # tile_count: int = self.tile_count  # type: ignore
```
drbeh, line 168: There are so many commented lines in this code, can you clean them up? thanks

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,159 @@ def get_params(self, image_size):
...
+        tile_step: int = self.step  # type: ignore
+
+        if self.random_offset and self.offset is not None:
```
drbeh, line 171: I don't think that `self.offset` can be None. Isn't `self.offset` always a tuple?

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,159 @@ def get_params(self, image_size):
...
+            shape=((xlen - xsize) // xstep + 1, (ylen - ysize) // ystep + 1, clen, xsize, ysize),
+            strides=(xstride * xstep, ystride * ystep, cstride, xstride, ystride),
+            writeable=False,
+        )
+        image = llw.reshape(-1, clen, xsize, ysize)
+
+        # if keep all patches
+        if self.tile_count is None:
+            # retain only patches with significant foreground content to speed up inference
+            # FYI, this returns a variable number of tiles, so the batch_size much be 1 (per gpu). Used during inference
```
drbeh, line 201: typo: much -> must

drbeh (changes requested):
@myron thanks for working this components.I have reviewed this and I left several comments.

Also can you work on removing all `# type: ignore`. I think that all the situation can be handled without being ignored. Please let me know if you have any question.

*** NOTE *** Please do not force push your next commits and let me know if you are not able to push without it.

monai/apps/pathology/transforms/spatial/array.py
myron: done

monai/apps/pathology/transforms/spatial/array.py
myron: ok

monai/apps/pathology/transforms/spatial/array.py
myron: fixed

monai/apps/pathology/transforms/spatial/array.py
myron: done

monai/apps/pathology/transforms/spatial/array.py
myron: done

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,159 @@ def get_params(self, image_size):
...
+    Args:
+        tile_count: number of tiles to extract, if None Extract all non-background tiles
+            Defaults to ``None``.
+        tile_size: size of the square tile
+            Defaults to ``256``.
+        step: step size
+            Defaults to None (same as tile_size)
+        random_offset: Randomize position of tile grid, instead of starting from the top-left corner
+            Defaults to ``False``.
+        pad_full: pad image to the size evenly divisible by tile_size
```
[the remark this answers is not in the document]
drbeh: @myron, my question is what `full` mean here. Can you help here?

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,152 @@ def get_params(self, image_size):
...
+            Defaults to ``None`` (same as tile_size)
+        random_offset: Randomize position of the grid, instead of starting from the top-left corner
+            Defaults to ``False``.
+        pad_full: pad image to the size evenly divisible by tile_size
+            Defaults to ``False``.
+        background_val: the background constant (e.g. 255 for white background)
+            Defaults to ``255``.
+        filter_mode: mode must be in ["min", "max", "random", None]. If total number of tiles is more than tile_size,
+            then sort by intensity sum, and take the smallest (for min), largest (for max) or random (for random or None) subset
+            Defaults to ``min`` (which assumes background is high value)
```
drbeh, line 100: I don't think that using `min` requires any assumption although I agree that the outcome would be different if background is low or high value. Do you mean that it is better to use `min` when background in high value? and `max` when the background it low value?

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,152 @@ def get_params(self, image_size):
...
+            Defaults to ``256``.
+        step: step size
+            Defaults to ``None`` (same as tile_size)
+        random_offset: Randomize position of the grid, instead of starting from the top-left corner
+            Defaults to ``False``.
+        pad_full: pad image to the size evenly divisible by tile_size
+            Defaults to ``False``.
+        background_val: the background constant (e.g. 255 for white background)
+            Defaults to ``255``.
+        filter_mode: mode must be in ["min", "max", "random", None]. If total number of tiles is more than tile_size,
```
drbeh, line 98: I didn't find anywhere that you are checking for `filter_mode is None`. Is there any reason you have None here?

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,152 @@ def get_params(self, image_size):
...
+            if len(image) >= self.tile_count:
+
+                if self.filter_mode == "min":
+                    # default, keep non-background tiles (smallest values)
+                    idxs = np.argsort(image.sum(axis=(1, 2, 3)))[: self.tile_count]
+                    image = image[idxs]
+                elif self.filter_mode == "max":
+                    idxs = np.argsort(image.sum(axis=(1, 2, 3)))[-self.tile_count :]
+                    image = image[idxs]
+                elif len(image) > self.tile_count:
```
drbeh, line 214: You should check here for `self.filter_mode == "random"`. 
Also raise a ValueError exception if `self.filter_mode` is not one of the expected value.

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,159 @@ def get_params(self, image_size):
...
+        self.background_val = background_val
+        self.filter_mode = filter_mode
+
+        # self.tile_all = (self.tile_count is None)
+
+        # if self.tile_count is None:
+        #     self.tile_count = max((44 * 256 ** 2) // (tile_size ** 2), 1)
+
+        if self.step is None:
+            self.step = self.tile_size  # non-overlapping grid
```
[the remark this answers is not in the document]
drbeh: This one is actually can help you not to do `tile_step = cast(int, self.step)` in the rest of the code.

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,152 @@ def get_params(self, image_size):
...
+
+        self.offset = (0, 0)
+        self.random_idxs = np.array((0,))
+
+    def randomize(self, img_size: Sequence[int]) -> None:
+
+        c, h, w = img_size
+        tile_step = cast(int, self.step)
```
drbeh, line 131: There is no need to cast to `int`. If you correct its assignment in `__init__`, `self.step` has to be int.

tests/test_tile_on_grid_dict.py
```diff
@@ -0,0 +1,123 @@
...
+
+import numpy as np
+from parameterized import parameterized
+
+from monai.apps.pathology.transforms import TileOnGridDict
+
+TEST_CASES = []
+for tile_count in [16, 64]:
+    for tile_size in [8, 32]:
+        for filter_mode in [None, "min", "max"]:
```
drbeh, line 23: Follow the changes in the main code, None -> `"random"`

drbeh (changes requested):

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,152 @@ def get_params(self, image_size):
...
+        else:
+            self.random_idxs = np.array((0,))
+
+    def __call__(self, image: np.ndarray) -> np.ndarray:
+
+        # add random offset
+        self.randomize(img_size=image.shape)
+        tile_step = cast(int, self.step)
+
+        if self.random_offset and self.offset[0] > 0 and self.offset[1] > 0:
```
drbeh, line 164:
```suggestion
        if self.random_offset and (self.offset[0] > 0 or self.offset[1] > 0):
```
Shouldn't be like this? You want to have offset when even one of the coordinate's offset is greater than zero, right?

drbeh (changes requested):

monai/apps/pathology/transforms/spatial/array.py
myron: no, I  want it only if both are positive

monai/apps/pathology/transforms/spatial/array.py
myron: self.step is Optional(int),  if default None is provided it is always set to int in __init__ function. 
Nevertheless, without cast() function, mypy complains on all consecutive operations with tile_stem (such as % /, etc)
I can't pass the mypy test without it

myron: how would you do it without casting?

monai/apps/pathology/transforms/spatial/array.py
myron: I'll do this check in Init once

monai/apps/pathology/transforms/spatial/array.py
myron: fixed

monai/apps/pathology/transforms/spatial/array.py
myron: yes, correct

monai/apps/pathology/transforms/spatial/array.py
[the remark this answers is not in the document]
myron: pad_full means to pad image to the full size that can be evenly divisible by tile_size without any borders

monai/apps/pathology/transforms/spatial/array.py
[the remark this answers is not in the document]
myron: no , I made your change, and remove casting, I get many mypy errors
```
monai/apps/pathology/transforms/spatial/array.py:155:15: error: Unsupported operand types for // ("int" and "None")  [operator]
monai/apps/pathology/transforms/spatial/array.py:155:15: note: Right operand is of type "Optional[int]"
monai/apps/pathology/transforms/spatial/array.py:155:35: error: Unsupported operand types for + ("int" and "None")  [operator]
monai/apps/pathology/transforms/spatial/array.py:156:15: error: Unsupported operand types for // ("int" and "None")  [operator]
monai/apps/pathology/transforms/spatial/array.py:156:15: note: Right operand is of type "Optional[int]"
monai/apps/pathology/transforms/spatial/array.py:156:35: error: Unsupported operand types for + ("int" and "None")  [operator]
```

monai/apps/pathology/transforms/spatial/array.py
myron: without casting I till get, these errors (let's keep casting for mypy, it actually doesn't do anything it just gives mypy a hint)

`monai/apps/pathology/transforms/spatial/array.py:155:15: error: Unsupported operand types for // ("int" and "None") [operator] monai/apps/pathology/transforms/spatial/array.py:155:15: note: Right operand is of type "Optional[int]" monai/apps/pathology/transforms/spatial/array.py:155:35: error: Unsupported operand types for + ("int" and "None") [operator] monai/apps/pathology/transforms/spatial/array.py:156:15: error: Unsupported operand types for // ("int" and "None") [operator] monai/apps/pathology/transforms/spatial/array.py:156:15: note: Right operand is of type "Optional[int]" monai/apps/pathology/transforms/spatial/array.py:156:35: error: Unsupported operand types for + ("int" and "None") [operator]`

monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -73,3 +75,159 @@ def get_params(self, image_size):
...
+            if self.filter_mode == "min":
+                # default, keep non-background tiles (small values)
+                idxs = np.argwhere(image.sum(axis=(1, 2, 3)) < thresh)
+                image = image[idxs.reshape(-1)]
+            elif self.filter_mode == "max":
+                idxs = np.argwhere(image.sum(axis=(1, 2, 3)) >= thresh)
+                image = image[idxs.reshape(-1)]
+
+        else:
+            if len(image) >= self.tile_count:
```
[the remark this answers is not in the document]
myron: ok

monai/apps/pathology/transforms/spatial/array.py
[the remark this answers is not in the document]
drbeh: hmm, that's weird. OK, we move on with this PR and I'll make a followup PR to fix this.

monai/apps/pathology/transforms/spatial/array.py
drbeh: If it has done correctly, we shouldn't need casting but let's keep it as is and I will fix it in a followup PR.

monai/apps/pathology/transforms/spatial/dictionary.py
```diff
@@ -53,4 +55,78 @@ def __call__(self, data: Mapping[Hashable, torch.Tensor]) -> Dict[Hashable, torc
...
+        d = dict(data)
+        for key in self.key_iterator(d):
+            self.splitter.set_random_state(seed=self.seed)  # same random seed for all keys
+            d[key] = self.splitter(d[key])
+
+        if self.return_list_of_dicts:
+            d_list = []
+            for i in range(len(d[self.keys[0]])):
+                d_list.append({k: d[k][i] if k in self.keys else copy.deepcopy(d[k]) for k in d.keys()})
+            d = d_list  # type: ignore
```
drbeh, line 126: What's the purpose of this line?

monai/apps/pathology/transforms/spatial/dictionary.py
```diff
@@ -53,4 +55,78 @@ def __call__(self, data: Mapping[Hashable, torch.Tensor]) -> Dict[Hashable, torc
...
+    def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Union[Dict[Hashable, np.ndarray], List[Dict]]:
+
+        self.randomize()
+
+        d = dict(data)
+        for key in self.key_iterator(d):
+            self.splitter.set_random_state(seed=self.seed)  # same random seed for all keys
+            d[key] = self.splitter(d[key])
+
+        if self.return_list_of_dicts:
```
drbeh, line 122: This functionality `return_list_of_dicts = True` has not covered in your test cases.

tests/test_tile_on_grid.py
```diff
@@ -0,0 +1,112 @@
...
+    # @parameterized.expand(TEST_CASES2)
+    # def test_tile_pathce_random_call(self, input_parameters):
+    #
+    #     img, tiles = make_image(**input_parameters, seed=123)
+    #
+    #     tiler = TileOnGrid(**input_parameters)
+    #     tiler.set_random_state(seed=123)
+    #
+    #     output = tiler(img)
+    #     np.testing.assert_equal(output, tiles)
```
drbeh, line 108: Why this is commented out?

tests/test_tile_on_grid_dict.py
```diff
@@ -0,0 +1,124 @@
...
+
+        splitter = TileOnGridDict(**input_parameters)
+
+        output = splitter({key: img})
+        output = output[key]
+
+        np.testing.assert_equal(tiles, output)
+
+    @parameterized.expand(TEST_CASES2)
+    def test_tile_pathce_random_call(self, input_parameters):
```
drbeh, line 105:
```suggestion
    def test_tile_patch_random_call(self, input_parameters):
```

drbeh (changes requested):

monai/apps/pathology/transforms/spatial/dictionary.py
myron: the type of the return changes from dict to list, so mypy complains

tests/test_tile_on_grid.py
myron: by mistake, I'll uncomment it

tests/test_tile_on_grid_dict.py
myron: ok

monai/apps/pathology/transforms/spatial/dictionary.py
myron: added

drbeh (approved):

PR #3235, opened 2021-11-01
Add `dtype` to ScaleIntensityRange transforms

Nic-Ma:
Fixes #3234 .

### Description
This PR added `dtype` arg to `ScaleIntensityRange` transforms, same as other `intensity scale` transforms which may change data type during computation, especially for `int` or `long` input.

### Status
**Ready**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick --unittests`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

PR #3244, opened 2021-11-02
fixes rescale_array

wyli:

follow-up of #3235

### Description
```
File "/home/jenkins/agent/workspace/MONAI-premerge/monai/monai/transforms/utils.py", line 167, in rescale_array

    norm = (arr - mina) / (maxa - mina)  # normalize the array first

RuntimeError: Integer division of tensors using div or / is no longer supported, and in a future release div will perform true division as in Python 3. Use true_divide or floor_divide (// in Python) instead.
```

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/utils.py
+++ b/monai/transforms/utils.py
@@ -157,7 +157,8 @@ def rescale_array(
     """
     Rescale the values of numpy array `arr` to be from `minv` to `maxv`.
     """
-    dtype_ = dtype or arr.dtype
+    if dtype is not None:
+        arr, *_ = convert_data_type(arr, dtype=dtype)
     mina = arr.min()
     maxa = arr.max()
 
@@ -165,10 +166,7 @@ def rescale_array(
         return arr * minv
 
     norm = (arr - mina) / (maxa - mina)  # normalize the array first
-    arr = (norm * (maxv - minv)) + minv  # rescale by minv and maxv, which is the normalized array by default
-
-    ret, *_ = convert_data_type(arr, dtype=dtype_)
-    return ret
+    return (norm * (maxv - minv)) + minv  # rescale by minv and maxv, which is the normalized array by default
 
 
 def rescale_instance_array(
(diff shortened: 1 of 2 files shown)


ericspod (approved):

monai/transforms/utils.py
```diff
@@ -157,18 +157,16 @@ def rescale_array(
...
 
     if mina == maxa:
         return arr * minv
 
     norm = (arr - mina) / (maxa - mina)  # normalize the array first
-    arr = (norm * (maxv - minv)) + minv  # rescale by minv and maxv, which is the normalized array by default
-
-    ret, *_ = convert_data_type(arr, dtype=dtype_)
-    return ret
+    return (norm * (maxv - minv)) + minv  # rescale by minv and maxv, which is the normalized array by default
```
Nic-Ma, line 169: Hi @ericspod @wyli ,

I think this change is not enough, it will cause below behavior:
```py
>>> from monai.transforms import rescale_array
>>> 
>>> 
>>> from monai.transforms import rescale_array
>>> import numpy as np
>>> 
>>> a = np.array([1, 2, 3])
>>> 
>>> b = rescale_array(a, minv=80, maxv=100, dtype=None)
>>> 
>>> print(a.dtype, b.dtype)
int64 float64
```

Let me try to enhance it ASAP.

Thanks.

wyli: I think the current one is fine for this utility, when dtype is none, it fallbacks to  the numpy-like behaviour.

Nic-Ma: Hi @wyli @ericspod ,

Current one doesn't work as expected for some cases not only `dtype=None`, for example:
```py
>>> from monai.transforms import rescale_array
>>> import numpy as np
>>> 
>>> a = np.array([1, 2, 3])
>>> b = rescale_array(a, minv=80, maxv=100, dtype=np.uint8)
>>> 
>>> print(a.dtype, b.dtype)
int64 float64
```

I submitted an enhance: https://github.com/Project-MONAI/MONAI/pull/3246, could you please help take a look?

Thanks in advance.

PR #3246, opened 2021-11-02
Enhance rescale_array

Nic-Ma:
Follow up of #3235 

### Description
This PR added conversion for the output in rescale_array.

### Status
**Ready**

### Types of changes


monai/transforms/utils.py
```diff
@@ -156,17 +156,26 @@ def rescale_array(
...
+        arr: input array to rescale.
+        minv: minimum value of target rescaled array.
+        maxv: maxmum value of target rescaled array.
+        dtype: data type of output, if None, same as input array.
+
     """
-    if dtype is not None:
-        arr, *_ = convert_data_type(arr, dtype=dtype)
+    dtype_ = dtype or arr.dtype
+    arr, *_ = convert_data_type(arr, dtype=dtype_)
```
ericspod, line 168: Is this doing something that the original didn't do? If `dtype` is `None` then doing no conversion seems fine, converting when `type_` is `arr.dtype` seems like a no-op?

Nic-Ma: Hi @ericspod ,

OK, sure, I changed the rescale_array back with more doc-strings for args, and updated `ScaleIntensity` to adjust the dtype in transform.

Thanks in advance.

Issue #3243, opened 2021-11-02
LoadImage Transform returns transposed image in MONAI 0.7.0


aarpon:
**Describe the bug**
The `monai.transforms.LoadImage` Transformer (independent of the reader used) erroneously returns a transpose of the image in MONAI 0.7.0, whereas it returned the correct image in previous versions.

**To Reproduce**
It is most obvious when the image is not square. I saw it when using images with shape (700, 1100). 

With MONAI == 0.6.0:

```python
from monai.transforms import LoadImage
loadImage = LoadImage()
out = loadImage("filename")   // shape = (700, 1100)
```

With MONAI == 0.7.0:

```python
from monai.transforms import LoadImage
loadImage = LoadImage()
out = loadImage("filename")   // shape = (1100, 700)
```
Changing the `reader` does not change the behavior.

**Expected behavior**
The behavior of `LoadImage` in 0.7.0 should be the same as in previous versions and consistent both with other readers and with the Numpy logic of shape = (rows, columns).

**Environment**

MONAI 0.6.0:

```bash
MONAI version: 0.6.0
Numpy version: 1.21.2
Pytorch version: 1.9.1+cu102
MONAI flags: HAS_EXT = False, USE_COMPILED = False
MONAI rev id: 0ad9e73639e30f4f1af5a1f4a45da9cb09930179

Optional dependencies:
Pytorch Ignite version: 0.4.5
Nibabel version: 3.2.1
scikit-image version: 0.18.3
Pillow version: 8.4.0
Tensorboard version: 2.7.0
gdown version: 4.2.0
TorchVision version: 0.10.1+cu102
ITK version: 5.1.2
tqdm version: 4.62.3
lmdb version: 1.2.1
psutil version: 5.8.0
pandas version: 1.3.4
einops version: 0.3.2
```

MONAI 0.7.0:

```bash
================================
Printing MONAI config...
================================
MONAI version: 0.7.0
Numpy version: 1.21.2
Pytorch version: 1.9.1+cu102
MONAI flags: HAS_EXT = False, USE_COMPILED = False
MONAI rev id: bfa054b9c3064628a21f4c35bbe3132964e91f43

Optional dependencies:
Pytorch Ignite version: 0.4.5
Nibabel version: 3.2.1
scikit-image version: 0.18.3
Pillow version: 8.4.0
Tensorboard version: 2.7.0
gdown version: 4.2.0
TorchVision version: 0.10.1+cu102
[...]

PR #3247, opened 2021-11-03
3212 3243 Add utility to check missing files in datalist

### LINKED ISSUE #3212: Add utility to check missing images in Decathlon datalist
Currently, we have the utility `decathlon_datalist` to load Decathlon stype images and labels:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/decathlon_datalist.py
It would be nice to have the utility to check whether some image or label files are missing before a heavy training.
NVIDIA Clara Train already has this feature, will try to reimplement in MONAI stype.

Nic-Ma:
Fixes #3212 #3243 .

### Description
This PR added the utility function to check missing files of Decathlon datalist before training.

### Status
**Ready**

### Types of changes


monai/data/decathlon_datalist.py
```diff
@@ -148,3 +149,35 @@ def load_decathlon_properties(data_property_file_path: str, property_keys: Union
             raise KeyError(f"key {key} is not in the data property file.")
         properties[key] = json_data[key]
     return properties
+
+
+def check_missing_files(datalist: List[Dict], keys: KeysCollection, allow_missing_keys: bool = False):
+    """Checks whether some files in the Decathlon datalist are missing.
+    It would be helpful to check missing files before a heavy training.
```
ericspod, line 156: "heavy training run"?

monai/data/decathlon_datalist.py
```diff
@@ -148,3 +149,35 @@ def load_decathlon_properties(data_property_file_path: str, property_keys: Union
...
+        for k in ensure_tuple(keys):
+            if k not in item:
+                if not allow_missing_keys:
+                    raise ValueError(f"key `{k}` is missing in the datalist item: {item}")
+                continue
+
+            for f in ensure_tuple(item[k]):
+                if not isinstance(f, str):
+                    raise ValueError(f"filepath of key `{k}` must be a string or a list of strings, but got: {f}.")
+                if not os.path.exists(f):
```
ericspod, line 180: Is it a concern if `f` isn't an absolute path but a relative path? This would be checking that `f` exists relative to the current directory, maybe want to use `abspath` before checking it exists. The paths put into `missing_files` perhaps should be absolute as well.

ericspod (approved):
I think it's fine with just a minor concern about absolute paths.

monai/data/decathlon_datalist.py
```diff
@@ -148,3 +149,35 @@ def load_decathlon_properties(data_property_file_path: str, property_keys: Union
...
+
+    Args:
+        datalist: a list of data items, every item is a dictionary.
+            ususally generated by `load_decathlon_datalist` API.
+        keys: expected keys to check in the datalist.
+        allow_missing_keys: whether allow missing keys in the datalist items.
+            if False, raise exception if missing. default to False.
+
+    Returns:
+        A list of missing file name.
```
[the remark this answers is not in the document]
Nic-Ma: updated in the latest commit.

Thanks.

monai/data/decathlon_datalist.py
Nic-Ma: Sure, thanks for your great suggestion!
That's a good point, I added support in the latest commit.

Thanks.

monai/data/decathlon_datalist.py
```diff
@@ -148,3 +149,40 @@ def load_decathlon_properties(data_property_file_path: str, property_keys: Union
...
+                    raise ValueError(f"key `{k}` is missing in the datalist item: {item}")
+                continue
+
+            for f in ensure_tuple(item[k]):
+                if not isinstance(f, str):
+                    raise ValueError(f"filepath of key `{k}` must be a string or a list of strings, but got: {f}.")
+                if isinstance(root_dir, str):
+                    f = os.path.join(root_dir, f)
+                if not os.path.exists(f):
+                    missing_files.append(f)
```
wyli, line 186: `os.path.exists` does not distinguish folder vs file, is that expected?

Nic-Ma: Hi @wyli ,

Thanks for your review.
I think `os.path.exists` is OK for this feature, because `f` can also be a folder to load, for example, the DICOM series for 1 3D image.

Thanks.

wyli: For example, if there's a folder named "test.img" but the datalist requires "test.img" file, the checking will be incorrect

Nic-Ma: I see, I think that's very special case, it's not easy to know the expected item is a file or folder. Maybe we don't need to check that so far.

Thanks.

PR #3252, opened 2021-11-03
Smooth field

ericspod:
### Description
This adds a few transforms that use a smooth field to adjust contrast and intensity over regions of an input image rather than for the whole image. This is useful for simulating different acquisition qualities in MR images. This isn't finished yet with more and better tests needed and @rijobro to look over to see what needs to be changed to be more compatible with the push for Pytorch+Numpy hybrid transforms. 

This adds a new transform directory for these, perhaps others should be added for MR-specific transforms and other groupings to reduce the size of the `array.py`/`dictionary.py` files that currently exist. 

### Status
**Work in progress**

### Types of changes


monai/transforms/smooth_field/dictionary.py
```diff
@@ -0,0 +1,146 @@
...
+    ):
+        RandomizableTransform.__init__(self, prob)
+        MapTransform.__init__(self, keys)
+
+        self.trans = RandSmoothFieldAdjustContrast(spatial_size, rand_size, padder, mode, align_corners, 1.0, gamma)
+        self.apply_same_field = apply_same_field
+
+    def set_random_state(
+        self, seed: Optional[int] = None, state: Optional[np.random.RandomState] = None
+    ) -> "Randomizable":
```
Nic-Ma, line 62: The return type should be "RandSmoothFieldAdjustContrastd"

Nic-Ma:
Please note to add an example image in the doc: https://github.com/Project-MONAI/DocImages
CC @rijobro as Richard is an expert on it.

Thanks.

ericspod:
Thanks, I was going to ask about the image docs.

monai/transforms/smooth_field/dictionary.py
```diff
@@ -0,0 +1,148 @@
...
+        mode: Union[InterpolateMode, str] = InterpolateMode.AREA,
+        align_corners: Optional[bool] = None,
+        prob: float = 0.1,
+        gamma: Union[Sequence[float], float] = (0.5, 4.5),
+        apply_same_field: bool = True,
+    ):
+        RandomizableTransform.__init__(self, prob)
+        MapTransform.__init__(self, keys)
+
+        self.trans = RandSmoothFieldAdjustContrast(spatial_size, rand_size, padder, mode, align_corners, 1.0, gamma)
```
rijobro, line 57: i would use keyword arguments here. I'm guessing the 1.0 refers to the probability, since we want this to be handled by the parent class, but it would help readability for other users if this were explicit.

monai/transforms/smooth_field/dictionary.py
```diff
@@ -0,0 +1,148 @@
...
+    def set_random_state(
+        self, seed: Optional[int] = None, state: Optional[np.random.RandomState] = None
+    ) -> "RandSmoothFieldAdjustContrastd":
+        super().set_random_state(seed, state)
+        self.trans.set_random_state(seed, state)
+        return self
+
+    def randomize(self, data: Optional[Any] = None) -> None:
+        super().randomize(None)
+        self.trans.randomize()
```
rijobro, line 69:
```suggestion
        if self._do_transform:
            self.trans.randomize()
```

monai/transforms/smooth_field/dictionary.py
```diff
@@ -0,0 +1,148 @@
...
+    def set_random_state(
+        self, seed: Optional[int] = None, state: Optional[np.random.RandomState] = None
+    ) -> "RandSmoothFieldAdjustIntensityd":
+        super().set_random_state(seed, state)
+        self.trans.set_random_state(seed, state)
+        return self
+
+    def randomize(self, data: Optional[Any] = None) -> None:
+        super().randomize(None)
+        self.trans.randomize()
```
rijobro, line 132: same as above.

tests/test_smooth_field.py
```diff
@@ -0,0 +1,72 @@
...
+
+from monai.transforms import RandSmoothFieldAdjustContrastd, RandSmoothFieldAdjustIntensityd
+from tests.utils import assert_allclose, is_tf32_env
+
+_rtol = 5e-3 if is_tf32_env() else 1e-4
+
+INPUT_SHAPE1 = (1, 8, 8)
+INPUT_SHAPE2 = (2, 8, 8)
+
+TESTS_CONTRAST = [
```
rijobro, line 26: Could you make sure that you test np, torch and torch.cuda input (eg https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_crop_foreground.py#L23)?

monai/transforms/smooth_field/dictionary.py
```diff
@@ -0,0 +1,148 @@
...
+        for key in self.key_iterator(d):
+            if not self.apply_same_field:
+                self.randomize()  # new field for every key
+
+            d[key] = self.trans(d[key], False)
+
+        return d
+
+
+class RandSmoothFieldAdjustIntensityd(RandomizableTransform, MapTransform):
```
rijobro, line 88: Could you set the transform backends (eg https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/croppad/array.py#L89)? I suspect it will be trivial to make them compatible with both np and torch inputs.

rijobro (approved):
Looks good once some of the suggested changes are incorporated. As you say, more tests needed, too.

rijobro:
> Thanks, I was going to ask about the image docs.

For the image docs, you need to:

1. clone https://github.com/Project-MONAI/DocImages
2. set the environment variable `MONAI_DOC_IMAGES` to point to this folder
3. for each of your new transforms to be visualised, add to the bottom of this file: https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/utils_create_transform_ims.py, and then run the file
4. files will be saved to `MONAI_DOC_IMAGES` folder -- commit and PR from there.

wyli (approved):
Thanks, it looks good to me.

PR #3262, opened 2021-11-05
3233 Add CumulativeAverage metric

### LINKED ISSUE #3233: Add AverageMeter metric handler
It would be useful to have some AverageMeter metric or handler to continuously computing the average value during training, draft idea code:
```py
class AverageMeter(object):
    """Computes and stores the average and current value"""
    def __init__(self):
        self.reset()

    def reset(self):
        self.val = 0
        self.avg = 0
        self.sum = 0
        self.count = 0

    def update(self, val, n=1):
        if n > 0:
            self.val = val
            self.sum += val * n

            self.count += n
            self.avg = self.sum / self.count if self.count > 0 else self.sum
```

Nic-Ma:
Fixes #3233 .

### Description
This PR added the `CumulativeAverage` metric to continuously compute average on some input data, it can support distributed data-parallel and multi-classes.

### Status
**Ready**

### Types of changes


monai/metrics/cumulative_average.py
```diff
@@ -0,0 +1,77 @@
...
+
+import torch
+
+from monai.metrics.utils import do_metric_reduction
+from monai.utils import MetricReduction
+
+from .metric import CumulativeIterationMetric
+
+
+class CumulativeAverage(CumulativeIterationMetric):
```
wyli, line 22: I think it's easier to follow if it inherits `Cumulative`, it doesn't have to be metrics...

Nic-Ma: Thanks for your suggestion.
That's good idea, I updated in the latest commit.

Thanks.

myron (approved):
looks good

monai/metrics/cumulative_average.py
```diff
@@ -0,0 +1,70 @@
...
+        self.sum = None
+        self.not_nans = None
+
+    def aggregate(self):  # type: ignore
+        """
+        Sync data from all the ranks and compute the average value with previous sum value.
+
+        """
+        data = self.get_buffer()
+        if not isinstance(data, torch.Tensor):
```
wyli, line 60: can this class be generic, not limited to tensor operations?

monai/metrics/cumulative_average.py
```diff
@@ -0,0 +1,70 @@
...
+            raise ValueError("the data to aggregate must be PyTorch Tensor.")
+
+        # do metric reduction
+        f, not_nans = do_metric_reduction(data, reduction=MetricReduction.SUM_BATCH)
```
wyli, line 64: this is the only line that the class is related to "metric", do we really need this in this class?

monai/metrics/cumulative_average.py
```diff
@@ -0,0 +1,70 @@
...
+        # clear the buffer for next update
+        super().reset()
```
wyli, line 66: why reset the buffer here? in my opinion `self.sum` and `self.non_nans` could be local variables and they could always be computed from `self._buffers`?

wyli (commented):
thanks, please see comments inline

monai/metrics/cumulative_average.py
Nic-Ma: Hi @wyli ,

Actually, @myron and I also discussed about this logic today, it looks more clear to always compute from `self.buffers`, but it's better to clear the buffer after every aggregation and only keep the `sum` & `not_nans` for the next computation to **save memory** and **performance**, otherwise, we need to repeatedly compute the average value from all the buffer when aggregating.
What do you think?

Thanks.

monai/metrics/cumulative_average.py
Nic-Ma: Hi @wyli ,

Here we leveraged the reduction utility to compute SUM on not_nans, do you mean to compute here directly?
```py
not_nans = (~(torch.isnan(data))).float().sum(dim=0)
f = data.sum(dim=0)  # the batch sum
```

Thanks.

monai/metrics/cumulative_average.py
Nic-Ma: OK, I think maybe this line is not necessary, to support multi-GPUs sync, the data in the base class `Cumulative` is already must be `torch.Tensor`.
I added this line because the following `SUM` logic is based on torch.Tensor APIs.

Thanks.

monai/metrics/cumulative_average.py
wyli: To compute the running average perhaps this class should override the append and extend?

Nic-Ma: Hi @wyli ,

OK, that's an interesting idea, let me think about it.
The only obvious difference is that we need to sync across ranks before computation.

Thanks.

myron: @wyli I agree with Nic about clearing memory. We need to do aggregation for logging frequently (every iteration), if don't clear, then the GPU memory array will grow constantly and require sync across gpu on a larger size every time.
@Nic-Ma if it's easy to chage append (you can try), but I need it for MIL the sooner the better

monai/metrics/cumulative_average.py
Nic-Ma: removed in the latest commit.

Thanks.

monai/metrics/cumulative_average.py
Nic-Ma: Updated in the latest commit and added test for numpy data.

Thanks.

monai/metrics/cumulative_average.py
Nic-Ma: Hi @wyli ,

I updated the PR based on the online discussion, could you please help review it again?

Thanks in advance.

wyli (approved):
Thanks, it looks good to me.

PR #3277, opened 2021-11-06
3276 - adds an option to skip renorm

### LINKED ISSUE #3276: ScaleIntesityRange performance
Related to https://github.com/Project-MONAI/MONAI/issues/3234

I need to normalize Uint8 image as (255 - x)/255, 
for which, the closest transform I found is
ScaleIntensityRange(a_min=255., a_max=0., b_min=0., b_max=1., clip=False, dtype=np.float32)

But with this, the overall training speed is 12% slower. And this transform is 5-6 times slower the my Manual simple function. 
First, the casting to float32 happens at the very end of ScaleIntensityRange, so all calculations are in double and only at the end cast to float32 (my example runtime 0.19s). If you pre-cast to float32, then speed improves ~2x (0.09s), but still not as fast as direct calculation (0.04s)
 
any recommendations on that?  is there a different transform I can use?
any should we change casting to float32 to be the first line in ScaleIntensityRange?
thanks

```
import numpy as np
from monai.transforms import Compose, ScaleIntensityRange, CastToType
import time

a=np.random.randint(255, size=(64, 3, 256, 256), dtype=np.uint8)

scaler = ScaleIntensityRange(a_min=255., a_max=0., b_min=0., b_max=1., clip=False, dtype=np.float32)
castscaler = Compose([CastToType(dtype=np.float32), scaler])

start = time.time()
b = (255 - a).astype(np.float32) / float(255)
print('Directly:', time.time()-start)

start = time.time()
b=scaler(a)
print('ScaleIntensityRange on uint8, time:', time.time()-start)

start = time.time()
b=castscaler(a)
print('ScaleIntensityRange pre-casted time:', time.time()-start)
```

Output:
```
Directly: 0.036348819732666016
ScaleIntensityRange on uint8, time: 0.19264698028564453
ScaleIntensityRange pre-casted time: 0.09259486198425293
```

### LINKED ISSUE #3283: test_cachedataset_persistent_workers freeze randomly for pytorch:21.09
**Describe the bug**
```
Starting test: test_duplicate_transforms (tests.test_cachedataset_persistent_workers.TestTransformsWCacheDatasetAndPersistentWorkers)...

Loading dataset:   0%|          | 0/1 [00:00<?, ?it/s]
Loading dataset: 100%|██████████| 1/1 [00:00<00:00, 600.47it/s]
Sending interrupt signal to process
Killing processes
kill finished with exit code 0
Sending interrupt signal to process
Killing processes
kill finished with exit code 2
```

this happens randomly and is difficult to reproduce.

wyli:

Fixes #3276
Fixes #3283

### Description
if the target range is none, skipping the renorm

### Status

### Types of changes


tests/test_cachedataset_persistent_workers.py
```diff
@@ -19,15 +19,14 @@
...
-        data = [{"img": im} for _ in range(2)]
+        data = [{"img": create_test_image_2d(128, 128, num_seg_classes=1, channel_dim=0)[0]} for _ in range(2)]
 
         # at least 1 deterministic followed by at least 1 random
         transform = Compose([Spacingd("img", pixdim=(1, 1)), RandAffined("img", prob=1.0)])
 
         # cachedataset and data loader w persistent_workers
         train_ds = CacheDataset(data, transform, cache_num=1)
-        train_loader = DataLoader(train_ds, num_workers=2, persistent_workers=True)
+        train_loader = DataLoader(train_ds, num_workers=1, persistent_workers=True)
```
Nic-Ma, line 29: May I know why changing the test workers from 2 to 1?
What't the root cause of test issue?

Thanks.

monai/transforms/utils.py
```diff
@@ -150,12 +150,13 @@ def zero_margins(img: np.ndarray, margin: int) -> bool:
 
 def rescale_array(
     arr: NdarrayOrTensor,
-    minv: float = 0.0,
-    maxv: float = 1.0,
+    minv: Optional[float] = 0.0,
```
Nic-Ma, line 153: As you added support for `None`, could you please also change below line:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/intensity/array.py#L436
To:
`if self.minv is not None or self.maxv is not None:`

Thanks.

tests/test_cachedataset_persistent_workers.py
Nic-Ma: Hi @wyli ,

Have you found how to reproduce it locally?

Thanks.

wyli: No I haven't

monai/transforms/utils.py
wyli: sure, I updated the if condition.

tests/test_cachedataset_persistent_workers.py
wyli: thanks, I added inline comments for future investigations.

Nic-Ma (approved):
Looks good to me.

Thanks.

PR #3278, opened 2021-11-06
Tifffile support in WSIReader

### LINKED ISSUE #3251: `tifffile` support in WSIReader
WSIReader currently supports `cuCIM` and `OpenSlide`, which can read patches from very large whole slide images without loading the entire image into memory. However, `tifffile` is another popular library to load TIFF files (which the most common format for whole slide imaging). Although it does not support extraction without loading the whole image, there are use cases that loading the whole slide image is desirable and `tifffile` can do it faster than `cuCIM` and `OpenSlide`.Thus supporting `tifffile` in `WSIReader` would be helpful.

Related issue: https://github.com/Project-MONAI/MONAI/issues/3161

drbeh:
Fixes #3251 

### Description
This PR adds support for `tifffile` in WSIReader to read TIFF files (which is common in histopathology whole slide imaging). While `cucim` and `openslide` both can load patches from large whole slide images without loading the entire image into memory, `tifffile`  loads the entire image into before extracting any patch. Thus memory consideration is needed when using `tifffile` for patch extraction.

### Status
**Ready**

### Types of changes


drbeh:
@Nic-Ma could you please take a look at this and run ``/build`` here?

monai/data/image_reader.py
```diff
@@ -689,8 +694,10 @@ def __init__(self, backend: str = "OpenSlide", level: int = 0):
             self.wsi_reader, *_ = optional_import("openslide", name="OpenSlide")
         elif self.backend == "cucim":
             self.wsi_reader, *_ = optional_import("cucim", name="CuImage")
+        elif self.backend == "tifffile":
```
Nic-Ma, line 697: I would suggest to change this `optional_import` to the beginning of this file and add the **dependency check** like:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/image_reader.py#L135
Maybe it's not very clear to you, I can do that for `WSIReader` later when this PR merged.

Thanks.

monai/data/image_reader.py
```diff
@@ -780,13 +787,22 @@ def _extract_region(
         level: int = 0,
         dtype: DtypeLike = np.uint8,
     ):
-        # reverse the order of dimensions for size and location to be compatible with image shape
-        location = location[::-1]
-        if size is None:
-            region = img_obj.read_region(location=location, level=level)
+        if self.backend == "tifffile":
+            region = img_obj.asarray(level=level)
+            img_obj.close()
```
Nic-Ma, line 792: Maybe it's better to close the `img_obj` at the same place it opened, instead of this `extract` function?

Thanks.

drbeh: I agree with you @Nic-Ma but the logic of image reader does not allow this since the object is being created at `.read` method and but used in `.get_data`. So here I am closing it as soon as possible. Thanks

monai/data/image_reader.py
drbeh: Thanks for your comment @Nic-Ma. I checked the implementation for for `@require_pkg` and it seems that it cannot handle the situation here, where the required package depends on the requested `backend`. So it either depends on `cuCIM` or `OpenSlide` or `Tifffile` but not all of them.

monai/data/image_reader.py
Nic-Ma: Is there a way to check whether the image is opened or not?
So if users call `get_data()` multiple times, they should be notified to call `read()` first.

Thanks.

myron: @drbeh I use tiffile to read image simply as
` image = tifffile.imread(image_path, key=self.level)`

do we need to call close() at all?

drbeh: @myron you're right but the way classes inherited from `ImageReader` work is to use `read()` to read image object and then use `get_data()` to retrieve the image, so instead of `imread`, we should use `TiffFile` which requires to be closed.

drbeh: @Nic-Ma I checked on this issue. The TiffFile object is being opened where `.asarray()` is being called so the implementation is actually fine. However, to avoid confusion I used context manager to when calling `asarray`.
I also updated a test to include multiple calls to `get_data` to make sure that there is not any problem with closing this file.

myron: looks good to me

Nic-Ma (approved):
Looks good to me.

Thanks for the update.

monai/data/image_reader.py
Nic-Ma: Hi @drbeh ,

I submitted PR https://github.com/Project-MONAI/MONAI/pull/3312 for this enhancement.

Thanks.

PR #3280, opened 2021-11-06
revise metrics and cumulative docstrings/APIs

wyli:

### Description
this is mostly non-breaking, updated the base cumulative API to have `extend` and `append`. fixed multiple docstrings to be clear about the shape assumptions.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/metrics/metric.py
+++ b/monai/metrics/metric.py
@@ -68,9 +68,7 @@ class IterationMetric(Metric):
             return self._compute_list(y_pred, y)
         # handling a single batch-first data
         if isinstance(y_pred, torch.Tensor):
-            y_ = None
-            if y is not None and isinstance(y, torch.Tensor):
-                y_ = y.detach()
+            y_ = y.detach() if isinstance(y, torch.Tensor) else None
             return self._compute_tensor(y_pred.detach(), y_)
         raise ValueError("y_pred or y must be a list/tuple of `channel-first` Tensors or a `batch-first` Tensor.")
 
@@ -79,9 +77,14 @@ class IterationMetric(Metric):
         Execute the metric computation for `y_pred` and `y` in a list of "channel-first" tensors.
 
         The return value is a "batch-first" tensor, or a list of "batch-first" tensors.
-        When it's a list of tensors, each item in the list can represent (batched) class-wise metrics.
+        When it's a list of tensors, each item in the list can represent a specific type of metric values.
 
-        Note: subclass may enhance the operation with multi-threads to accelerate.
+        For example, `self._compute_tensor` may be implemented as returning a list of `batch_size` items,
+        where each item is a tuple of three values `tp`, `fp`, `fn` for true positives, false positives,
+        and false negatives respectively. This function will return a list of three items,
+        (`tp_batched`, `fp_batched`, `fn_batched`), where each item is a `batch_size`-length tensor.
+
+        Note: subclass may enhance the operation to have multi-thread support.
         """
         if y is not None:
             ret = [self._compute_tensor(p.detach().unsqueeze(0), y_.detach().unsqueeze(0)) for p, y_ in zip(y_pred, y)]
@@ -188,6 +191,7 @@ class Cumulative:
         """
         Extend the local buffers with new ("batch-first") data.
         A buffer will be allocated for each `data` item.
+        Compared with `self.append`, this method adds a "batch" of data to the local buffers.
 
         Args:
             data: each item can be a "batch-first" tensor or a list of "channel-first" tensors.
@@ -212,6 +216,8 @@ class Cumulative:
         """
         Add samples to the local cumulative buffers.
         A buffer will be allocated for each `data` item.
+        Compared with `self.extend`, this method adds a single sample (instead
+        of a "batch") to the local buffers.
 
         Args:
             data: each item will be converted into a torch tensor.
--- a/monai/metrics/metric.py
+++ b/monai/metrics/metric.py
@@ -59,7 +59,7 @@ class IterationMetric(Metric):
         Returns:
             The computed metric values at the iteration level.
             The output shape could be a `batch-first` tensor or a list of `batch-first` tensors.
-            When it's a list of tensors, each item in the list can represent (batched) class-wise metrics.
+            When it's a list of tensors, each item in the list can represent a specific type of metric.
 
         """
         ret: TensorOrList
--- a/monai/metrics/metric.py
+++ b/monai/metrics/metric.py
@@ -15,121 +15,165 @@ from typing import Any, List, Optional
 import torch
 
 from monai.config import TensorOrList
-from monai.utils import evenly_divisible_all_gather
+from monai.utils import convert_data_type, evenly_divisible_all_gather
+
+__all__ = ["Metric", "IterationMetric", "Cumulative", "CumulativeIterationMetric"]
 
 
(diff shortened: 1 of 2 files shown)


monai/metrics/metric.py
```diff
@@ -15,121 +15,165 @@
...
+
+    The following is an example of extending with variable length data:
+
+    .. code-block:: python
+
+        import torch
+        from monai.metrics import Cumulative
+
+        c = Cumulative()
+        c.extend(torch.zeros((8, 2)), torch.zeros((6, 2)))  # adds batches
```
Nic-Ma, line 160: Could you please help explain more about the usage of `extend`?
I didn't get why adding this new API against the `append` or the previous `add`.

Thanks.

monai/metrics/metric.py
```diff
@@ -15,121 +15,165 @@
...
     def _compute_list(self, y_pred: TensorOrList, y: Optional[TensorOrList] = None):
         """
-        Excute the computation for the y_pred and y items of a iteration, the data is in the list shape.
-        Will concat the results to guarantee the output shape of ret is BCHW[D], otherwise it's list of batch-first,
-        which is against our principle that data in metrics should be BCHW[D] or list of channel-first.
-        Note: subclass may enhance the operation with multi-threads to accelerate.
+        Execute the metric computation for `y_pred` and `y` in a list of "channel-first" tensors.
 
+        The return value is a "batch-first" tensor, or a list of "batch-first" tensors.
+        When it's a list of tensors, each item in the list can represent (batched) class-wise metrics.
```
Nic-Ma, line 82: I think usually the multi-class metrics is a multi-channel Tensor, not items of the list, here if every item of `ret` is a list, means the metric returns multiple items for every iteration, for example:
```py
class MyMetric:
    ...
    def _compute_tensor(self):
        # return some median computation result for every iteration, will aggregate for the final value later
        return TP, FP, WD
```

What do you think?

Thanks.

monai/metrics/metric.py
wyli: yes, these APIs will provide flexibility in handling the batch dimension.

monai/metrics/metric.py
```diff
@@ -140,64 +184,103 @@ def reset(self):
...
+                they will be concatenated at the 0-th dimension when `get_buffer()` is called.
+        """
+        if self._buffers is None:
+            self._buffers = [[] for _ in data]
+        for b, d in zip(self._buffers, data):
+            # converting to pytorch tensors so that we can use the distributed API
+            d_t: torch.Tensor
+            d_t, *_ = convert_data_type(d, output_type=torch.Tensor, wrap_sequence=True)  # type: ignore
+            try:
+                b.extend([x[0] for x in torch.split(d_t, 1, dim=0)])
```
Nic-Ma, line 203: I doult there may be a little performance drop to split the batch into list and stack them in `sync()`, the previous code store `batch-first` Tensor into buffer directly and concat several `batch-first` Tensors in `sync()`.
Maybe you can do some quick performance test if batch size is bigger?

Thanks in advance.

wyli: Hi Nic, this is to have consistent buffer elements in the implementation, and the base class is to show the concepts. If the performance is an issue, the subclass can provide other implementations.

monai/metrics/metric.py
```diff
@@ -15,121 +15,168 @@
...
-            y_pred: the model prediction data to compute, must be a list of `channel-first` Tensor
+            y_pred: the raw model prediction data at one iteration, must be a list of `channel-first` Tensor
                 or a `batch-first` Tensor.
             y: the ground truth to compute, must be a list of `channel-first` Tensor
                 or a `batch-first` Tensor.
 
+        Returns:
+            The computed metric values at the iteration level.
+            The output shape could be a `batch-first` tensor or a list of `batch-first` tensors.
+            When it's a list of tensors, each item in the list can represent (batched) class-wise metrics.
```
Nic-Ma, line 62: Could you please help also update this line to `a specific type of metric values.`?

Thanks.

monai/metrics/metric.py
```diff
@@ -140,64 +187,106 @@ def reset(self):
...
+                they will be concatenated at the 0-th dimension when `get_buffer()` is called.
+        """
+        if self._buffers is None:
+            self._buffers = [[] for _ in data]
+        for b, d in zip(self._buffers, data):
+            # converting to pytorch tensors so that we can use the distributed API
+            d_t: torch.Tensor
+            d_t, *_ = convert_data_type(d, output_type=torch.Tensor, wrap_sequence=True)  # type: ignore
+            try:
+                b.extend([x[0] for x in torch.split(d_t, 1, dim=0)])
```
Nic-Ma, line 207: I think maybe `torch.unbind()` is more suitable than `torch.split` here?

Thanks.

Nic-Ma (approved):
Thanks for the quick update.
Looks good to me, just some minor comments.

Thanks.

monai/metrics/metric.py
```diff
@@ -79,9 +77,14 @@ def _compute_list(self, y_pred: TensorOrList, y: Optional[TensorOrList] = None):
         Execute the metric computation for `y_pred` and `y` in a list of "channel-first" tensors.
 
         The return value is a "batch-first" tensor, or a list of "batch-first" tensors.
-        When it's a list of tensors, each item in the list can represent (batched) class-wise metrics.
+        When it's a list of tensors, each item in the list can represent a specific type of metric values.
 
-        Note: subclass may enhance the operation with multi-threads to accelerate.
+        For example, `self._compute_tensor` may be implemented as returning a list of `batch_size` items,
```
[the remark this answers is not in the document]
wyli: Hi Nic, this is to show the number of items in the list is `batch_size`, and in the later docstring each item has the length of `batch_size`.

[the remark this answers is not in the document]
Nic-Ma: OK, I see.
Maybe you can try to make it more clear to avoid confusion, some thing like:
`returning a list of `batch-first` items, length of each item is same as `batch_size`?

Thanks in advance.

monai/metrics/metric.py
wyli: Not sure why we make the change, could you please clarify the advantage?

monai/metrics/metric.py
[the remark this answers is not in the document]
wyli: I think it's clear that I use "`" to denote the size, and the double-quote for the word"batch_first"

monai/metrics/metric.py
Nic-Ma: I think `torch.split()` is mainly used for specified length of data chunks, here we fallback to use `length=1`, seems `torch.unbind` is simpler for this use case?

Thanks.

monai/metrics/metric.py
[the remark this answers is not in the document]
Nic-Ma: OK, I missed line 79.

Thanks.

monai/metrics/metric.py
wyli: It will change the outcome for one of the test case:
`[x[0] for x in torch.split(torch.tensor([]), 1, dim=0)]` vs `torch.unbind(torch.tensor([]), 0)`, I think the split is better here.

Nic-Ma: OK.

Thanks.

PR #3285, opened 2021-11-07
3284 torch version check

### LINKED ISSUE #3284: pytorch version utility
**Describe the bug**
the current utilities for identifying pytorch versions are not robust to those from nvidia pytorch docker images (such as `PT_BEFORE_1_7`, `version_leq(torch.__version__, '1.10')`). Given that these are important for compatibility with different pytorch versions, the utility should be unified and tested.

wyli:
Fixes #3284

### Description
Adds a utility to identify PyTorch versions

### Status
**Ready**

### Types of changes


monai/utils/module.py
```diff
@@ -450,7 +453,47 @@ def _try_cast(val: str):
...
+        while len(parts) < 3:
+            parts += ["0"]
+        c_major, c_minor, c_patch = parts[:3]
+    except (AttributeError, ValueError, TypeError):
+        c_major, c_minor = get_torch_version_tuple()
+        c_patch = "0"
+    c_mn = int(c_major), int(c_minor)
+    mn = int(major), int(minor)
+    if c_mn != mn:
+        return c_mn > mn
```
ericspod, line 485: Would it make sense to use [version_leq](https://github.com/Project-MONAI/MONAI/blob/f54b7af0647dc5a1ea259765a19d3d3bcb3c42c5/monai/utils/module.py#L410) in this context? Thinking about this and `version_leq` in general maybe we should be using [packaging.version](https://packaging.pypa.io/en/latest/version.html) if we could assume `packaging` is always going to be installed, it's third party but it would be an odd environment that didn't have it.

wyli: the third party one is not included in some docker images...I added your suggestion as the preferred solution.

wyli:
could you review this? @Nic-Ma @ericspod

ericspod (approved):

PR #3289, opened 2021-11-08
2213 allows for non-invertible transforms in TTA

### LINKED ISSUE #2213: Random Intensity transforms for test-time augmentation

wyli:

Fixes #2213

### Description
this is a non-breaking change, removes the strong assumption of all invertibles:
https://github.com/Project-MONAI/MONAI/blob/8c0ff376587797aabd2abbdb569fb691880fc094/monai/data/test_time_augmentation.py#L136-L137

### Status
**Ready**

### Types of changes


monai/data/test_time_augmentation.py
```diff
@@ -180,8 +185,10 @@ def __call__(
                 batch_output = batch_output.detach().cpu()
             if isinstance(batch_output, np.ndarray):
                 batch_output = torch.Tensor(batch_output)
-
-            transform_info = batch_data[transform_key]
+            transform_info = batch_data.get(transform_key, None)
+            if transform_info is None:
+                # no invertible transforms, adding dummy info for identity invertible
+                transform_info = [[0.0] for _ in range(self.batch_size)]
```
Nic-Ma, line 191: I am not sure whether we should add dummy data here, maybe just leave it to `None` if no invertible transforms?

Thanks.

Nic-Ma (approved):

monai/data/test_time_augmentation.py
wyli: `None` doesn't work with one of the collate method

Nic-Ma: Ah, yes, that's the issue.
Or maybe we can use `"None"` to be consistent with other similar places, like:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/spatial/dictionary.py#L259

Thanks.

PR #3311, opened 2021-11-11
3306 Update document of DiceLoss

### LINKED ISSUE #3306: Documentation for DiceLoss is misleading
The [documentation for `DiceLoss`](https://docs.monai.io/en/latest/losses.html?highlight=dice%20loss#diceloss) says (my bold):

> Compute average Dice loss between two tensors. It can support both multi-classes and multi-labels tasks. Input **logits** _input_ (BNHW[D] where N is number of classes) is compared with ground truth _target_ (BNHW[D]). Axis N of _input_ is expected to have **logit** predictions for each class rather than being image channels, while the same axis of target can be 1 or N (one-hot format)...

The term "logit" here doesn't seem to match how `DiceLoss` [actually works](https://github.com/Project-MONAI/MONAI/blob/dev/monai/losses/dice.py#L162-L176); it expects class probabilities and not logit values. That can change for one of the inputs if the parameters `sigmoid` or `softmax` are set to `True`, but those parameters are `False` by default.

Nic-Ma:
Fixes #3306 .

### Description
This PR updated the document of `DiceLoss` to make the logits or probabilities input clear.

### Status
**Ready**

### Types of changes


monai/losses/dice.py
```diff
@@ -27,16 +27,16 @@
...
-    values added to the intersection and union components of the inter-over-union calculation to smooth results
-    respectively, these values should be small. The `include_background` class attribute can be set to False for
-    an instance of DiceLoss to exclude the first category (channel index 0) which is by convention assumed to be
-    background. If the non-background segmentations are small compared to the total image size they can get
-    overwhelmed by the signal from the background so excluding it in such cases helps convergence.
+    The data `input` (BNHW[D] where N is number of classes) is compared with ground truth `target` (BNHW[D]).
 
-    Milletari, F. et. al. (2016) V-Net: Fully Convolutional Neural Networks forVolumetric Medical Image Segmentation, 3DV, 2016.
+    Note that axis N of `input` is expected to be logits or probabilities for each class, must set `sigmoid=True`
+    or `softmax=True` if passing logits as input. And the same axis of `target` can be 1 or N (one-hot format).
```
yiheng-wang-nv, line 33: or specifying `other_act`

Nic-Ma: Sure, updated in the latest commit.

Thanks.

yiheng-wang-nv (approved):

PR #3312, opened 2021-11-11
3251 Add dependency check in WSIReader

### LINKED ISSUE #3251: `tifffile` support in WSIReader

Nic-Ma:
Fixes #3251 .

### Description
This PR is follow up of #3251 to add the dependency check in WSIReader.

### Status
**Ready**

### Types of changes


yiheng-wang-nv (approved):

monai/data/image_reader.py
```diff
@@ -690,14 +694,18 @@ class WSIReader(ImageReader):
...
-        if self.backend == "openslide":
-            self.wsi_reader, *_ = optional_import("openslide", name="OpenSlide")
-        elif self.backend == "cucim":
-            self.wsi_reader, *_ = optional_import("cucim", name="CuImage")
-        elif self.backend == "tifffile":
-            self.wsi_reader, *_ = optional_import("tifffile", name="TiffFile")
-        else:
-            raise ValueError('`backend` should be "cuCIM", "OpenSlide", or "TiffFile')
+
+        @require_pkg(pkg_name=self.backend)
```
wyli, line 698: Why hiding this API, we lose the flexibility of inheriting it easily

Nic-Ma: Hi @wyli ,

I think it's not supported in python to set `pkg_name=self.backend` directly if not hiding the function in `init()`:
```py
def __init__(self, backend: str = "OpenSlide", level: int = 0):
    super().__init__()
    self.backend = backend.lower()
    self.wsi_reader = self._set_reader(self.backend)
    self.level = level

@require_pkg(pkg_name=self.backend)
def _set_reader(self, backend: str):
    if backend == "openslide":
        return OpenSlide
    if backend == "cucim":
        return CuImage
    if backend == "tifffile":
        return TiffFile
    raise ValueError("`backend` should be 'cuCIM', 'OpenSlide' or 'TiffFile'.")
```

Maybe I missed some python feature?

Thanks.

wyli: I think you can use the decorator in the constructor as a normal function "require_pkg(backend)(self._set_reader)"

Nic-Ma: updated in the latest commit.

Thanks.

wyli (approved):
Thanks, it looks good to me.

PR #3325, opened 2021-11-12
3292 Add support for no keys in all the transforms when allow_missing_keys is True

### LINKED ISSUE #3292: RandAffined fails with KeyError when allow_missing_keys and key at index 0 is missing
**Describe the bug**
We have a pipeline in which not all samples have the same keys.
The `RandAffined` transform accesses fallback data via index 0.
If index 0 is a missing key, then this transform will fail with a `KeyError`.

```python
805   sp_size = fall_back_tuple(self.rand_affine.spatial_size, data[self.keys[0]].shape[1:])
806   # change image size or do random transform
807   do_resampling = self._do_transform or (sp_size != ensure_tuple(data[self.keys[0]].shape[1:]))
```

**Expected behavior**
If the option `allow_missing_keys` is set, I would expect the `RandAffined` transform to take the first element of the existing keys as fallback.

Nic-Ma:
Fixes #3292 .

### Description
This PR added support for no matching keys in all the MONAI dict transforms when `allow_missing_keys` is True.
I checked all the dict level transforms for the case that `allow_missing_keys` is True and no matching key in the input data.

### Status
**Ready**

### Types of changes


monai/apps/deepgrow/transforms.py
```diff
@@ -652,8 +652,12 @@ def bounding_box(self, points, img_shape):
 
     def __call__(self, data):
         d: Dict = dict(data)
+        image_key = first(self.key_iterator(d))
+        if image_key is None:
```
wyli, line 656: technically `None` is a valid key:
```py
>>> import numpy as np
>>> from monai.transforms import ScaleIntensityd
>>> ScaleIntensityd(None)({None: np.zeros((1, 2,3))})
{None: array([[[0., 0., 0.],
        [0., 0., 0.]]], dtype=float32)}
>>> 

```

Nic-Ma: Yes, I know that.
I think we discussed it before.
Anyway, maybe it's better to check the length of available keys, let me try later.

Thanks.

Nic-Ma: Updated in the latest commit.

Thanks for the comments.

monai/apps/deepgrow/transforms.py
```diff
@@ -652,8 +644,12 @@ def bounding_box(self, points, img_shape):
 
     def __call__(self, data):
         d: Dict = dict(data)
+        exist_keys = list(self.key_iterator(d))
```
wyli, line 647: in this way we lose the nice iterator style... but at the same time I don't have a better idea to check whether the iterator has no item cc @rijobro @ericspod

Nic-Ma: Hi @wyli @ericspod @rijobro ,

Thanks for your review here.
I think it's OK to get the length by listing the iterator as the `self.keys` are very small data, do you have any other ideas?

Thanks in advance.

Nic-Ma:
CI test error depends on: https://github.com/Project-MONAI/MONAI/pull/3337.

Thanks.

monai/apps/deepgrow/transforms.py
Nic-Ma: Hi @rijobro @ericspod ,

Could you please help share some comments here? Or any other better idea？

Thanks in advance.

wyli: in this case perhaps we can have
```py
first_key = next(iterator, [])
if first_key == []:
    print("iterator empty")
```

Nic-Ma: Sure, I updated the PR according to your suggestion.

Thanks in advance.

monai/transforms/croppad/dictionary.py
```diff
@@ -669,7 +677,11 @@ def __init__(
         self.max_roi_scale = max_roi_scale
 
     def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> Dict[Hashable, NdarrayOrTensor]:
-        img_size = data[first(self.key_iterator(data))].shape[1:]  # type: ignore
+        first_key: Union[Hashable, List] = next(self.key_iterator(data), [])  # type: ignore
+        if first_key == []:
+            return data  # type: ignore
```
wyli, line 682: This one has the type ignore, any idea why it's different from the other modifications?

monai/apps/deepgrow/transforms.py
wyli: Using monai.utils.first(iterator, []) will also improve the readability. Should we consider 'self.first_key(data)' a base class API?

Nic-Ma: Actually, I planned to add to the base class API, but we usually don't change base class frequently, so I didn't do it.
Lst me add it to base class directly.

Thanks.

monai/transforms/croppad/dictionary.py
Nic-Ma: Hi @wyli ,

That because this transform is subclass of another transform and don't have the d = dict(data) logic, d has different type from data.

Thanks.

wyli (approved):
Thanks, it looks good to me.

PR #3318, opened 2021-11-11
Add affine param to Affine transform (#3313)

Spenhouet:
Signed-off-by: Sebastian Penhouet <s.penhouet@outlook.de>

Fixes #3313 .

### Description
Add the affine parameter to the Affine transform.

### Status
**Ready**

### Types of changes
- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [x] In-line docstrings updated.

PR #3326, opened 2021-11-12
3313 Add `affine` arg to the dict transform

### LINKED ISSUE #3313: The Affine transform currently does not have an affine matrix parameter
The `Affine` transform forwards its parameters to the `AffineGrid` class.
But it currently does not provide the "affine" parameter which the `AffineGrid` class also has.

This allows to directly define a affine transformation matrix instead of individual parameters for transformations like rotation.
If computations are directly done with affine transformation matrices, this parameter is really helpful (necessary).

The only change necessary, is to add the `affine` parameter and to forward it to the `AffineGrid` class.

```python
class Affine(Transform):
    ...
    def __init__(
        ...
        affine: Optional[NdarrayOrTensor] = None,
        ...
    ) -> None:
        ...
        self.affine_grid = AffineGrid(
            ...
            affine=affine,
            ...
        )
```

We patched the affine transform locally.

Nic-Ma:
Fixes #3313 .

### Description
This PR is follow up of #3318 , added the `affine` arg to the dict level transform.

### Status
**Ready**

### Types of changes


monai/transforms/spatial/dictionary.py
```diff
@@ -662,6 +666,7 @@ def __init__(
             shear_params=shear_params,
             translate_params=translate_params,
             scale_params=scale_params,
+            affine=affine,
```
yiheng-wang-nv, line 669: Thanks @Nic-Ma . Could you also add unittest for it, and also for `class Affine`?

yiheng-wang-nv: It seems `class AffineGrid` also misses the corresponding unit test for `affine` parameter, could you please add it as well?

Nic-Ma: Sure, I will add it soon.

Thanks for your finding.

yiheng-wang-nv (approved):

monai/transforms/spatial/dictionary.py
Nic-Ma: Added unit tests in the latest commit.

Thanks.

PR #3332, opened 2021-11-14
3329 compatibility with pathlike obj

### LINKED ISSUE #3329: download_and_extract to handle Path objects and downloaded file auto naming
I rather work with pathlib and Path objects than with raw string paths. Today I stumbled upon two problems trying to use `download_and_extract()` from the apps module. 

1. `filepath` and `output_dir` parameters don't get Path objects and it generates:

> AttributeError: 'PosixPath' object has no attribute 'rstrip'

Wrapping them by str() helps to avoid the error.

As it is shown in the docs to use string, I think it would be good to have wider possibilities with that, especially with such broadly used lib as pathlib. An alternative to that would be a more precise error note.

2. `filepath` parameter cause a RuntimeError when just a directory path with no file name and extension is given. What is confusing the error if pointing towards `hash_val` parameter and not the one it should.

> RuntimeError: md5 check of existing file failed: filepath=/home/user/MEDnist/mednist_practice_1/notebooks/../mnist_data, expected md5=0bc7306e7427e00ad1c5526a6677552d.

It would be comfortable to use just a directory path and get the file name and its extension from the file downloaded. I see a problem here when having no knowledge of what an extension of the downloading file is. An alternative, same as before, is to have a more clear error description pointing the problem itself.

wyli:

Fixes #3329

### Description

### Status

### Types of changes


monai/apps/datasets.py
```diff
@@ -105,19 +107,17 @@ def get_num_classes(self) -> int:
...
-        class_names = sorted(x for x in os.listdir(dataset_dir) if os.path.isdir(os.path.join(dataset_dir, x)))
+        dataset_dir = Path(dataset_dir)
+        class_names = sorted(f"{x}" for x in dataset_dir.iterdir() if (dataset_dir / x).is_dir())
         self.num_class = len(class_names)
         image_files = [
-            [
-                os.path.join(dataset_dir, class_names[i], x)
-                for x in os.listdir(os.path.join(dataset_dir, class_names[i]))
-            ]
+            [f"{dataset_dir.joinpath(class_names[i], x)}" for x in (dataset_dir / class_names[i]).iterdir()]
```
Nic-Ma, line 120: Could you please help add a comment here to remind us that Path is not supported JSON?

Thanks.

Nic-Ma (approved):

monai/apps/datasets.py
wyli: this is not about json, the path obj is not compatible with the default collate.

PR #3341, opened 2021-11-15
Enhance workflow to avoid modification metrics

Nic-Ma:
### Description
This PR enhanced the init logic of workflow engines to avoid modification on the input `k_metric` dict, it caused potential issue in NVIDIA Clara project as the metric dict is changed in this function. Follow the similar pattern of our dict transforms.

### Status
**Ready**

### Types of changes


wyli (approved):
Thanks, it looks good to me.

Issue #3106, opened 2021-10-11
Creating tutorials for both network architecture search and operation Search


dongyang0122:
**Is your feature request related to a problem? Please describe.**
Creating tutorials (notebook or markdown) for both network architecture search and operation Search.

PR #3344, opened 2021-11-15
add dints model

dongyang0122:
Signed-off-by: dongy <dongy@nvidia.com>

fixes https://github.com/Project-MONAI/MONAI/issues/3106
core component of https://openaccess.thecvf.com/content/CVPR2021/papers/He_DiNTS_Differentiable_Neural_Network_Topology_Search_for_3D_Medical_Image_CVPR_2021_paper.pdf 

### Description
this is the primary PR that's needed for https://github.com/Project-MONAI/MONAI/milestone/28, the rest of the milestone could be done via the tutorial repo.

### Status
**Ready**

### Types of changes


monai/networks/blocks/dints_block.py
```diff
@@ -0,0 +1,107 @@
...
+    def __init__(self, in_channel: int, out_channel: int):
+        super().__init__()
+        self._in_channel = in_channel
+        self._out_channel = out_channel
+
+        assert self._out_channel % 2 == 0
+        self.relu = nn.ReLU()
+        self.conv_1 = nn.Conv3d(self._in_channel, self._out_channel // 2, 1, stride=2, padding=0, bias=False)
+        self.conv_2 = nn.Conv3d(self._in_channel, self._out_channel // 2, 1, stride=2, padding=0, bias=False)
+        self.bn = nn.InstanceNorm3d(self._out_channel)
```
wyli, line 52: the activation function and feature norm function are currently hard-coded, is this a design choice or should this be flexible, with options such as https://github.com/Project-MONAI/MONAI/blob/33b6d61fdd8a15247e63bf7db16ac32f98f83566/monai/networks/blocks/dynunet_block.py#L48-L49

monai/networks/blocks/dints_block.py
```diff
@@ -0,0 +1,107 @@
...
+            padding1 = (0, padding, 0)
+        elif self._p3dmode == 2:  # 1 x 3 x 3
+            kernel_size0 = (1, kernel_size, kernel_size)
+            kernel_size1 = (kernel_size, 1, 1)
+            padding0 = (0, padding, padding)
+            padding1 = (padding, 0, 0)
+
+        self.op = nn.Sequential(
+            nn.ReLU(),
+            nn.Conv3d(self._in_channel, self._in_channel, kernel_size0, padding=padding0, bias=False),
```
wyli, line 86: I think the kernel_size and padding value should be compatible to ensure the output shape is the same as the input.

monai/networks/nets/dints.py
```diff
@@ -0,0 +1,556 @@
...
+                    self.memory[blk_idx, res_idx] = np.array(
+                        [
+                            _.memory + self.cell_tree[str((blk_idx, res_idx))].preprocess.memory if _ is not None else 0
+                            for _ in self.cell_tree[str((blk_idx, res_idx))].op._ops[:cell_ops]
+                        ]
+                    )
+
+        # define cell and macro arhitecture probabilities
+        self.log_alpha_c = torch.nn.Parameter(
+            torch.zeros(num_blocks, len(code2out), cell_ops).normal_(1, 0.01).cuda().requires_grad_()
```
wyli, line 269: the tensors are hard-coded to cuda device, I think this should be changed to be optional.

monai/networks/nets/dints.py
```diff
@@ -0,0 +1,556 @@
...
+        cell=Cell,
+        cell_ops: int = 5,
+        channel_mul: float = 1.0,
+        num_blocks: int = 6,
+        num_depths: int = 3,
+        use_stem: bool = False,
+        code: list = None,
+    ):
+        """
+        Initialize NAS network search space
```
wyli, line 142: would be great to have a short paragraph to describe the network including the paper references.

monai/networks/nets/dints.py
```diff
@@ -0,0 +1,556 @@
...
+            filter_nums: default init 64. Double channel number after downsample
+            topology related varaibles from gen_mtx():
+                trans_mtx: feasible path activation given node activation key
+                code2in: path activation to its incoming node index
+                code2ops: path activation to operations of upsample 1, keep 0, downsample -1
+                code2out: path activation to its output node index
+                node_act_list: all node activation codes [2^num_depths-1, res_num]
+                node_act_dict: node activation code to its index
+                tidx: index used to convert path activation matrix (depth,depth) in trans_mtx to path activation
+                    code (1,3*depth-2)
```
wyli, line 165: might be useful to include some examples like https://github.com/Project-MONAI/MONAI/blob/33b6d61fdd8a15247e63bf7db16ac32f98f83566/monai/networks/nets/unetr.py#L65-L66

monai/networks/nets/dints.py
```diff
@@ -0,0 +1,556 @@
...
+                )
+                self.stem_up[str(res_idx)] = nn.Sequential(
+                    nn.ReLU(),
+                    nn.Conv3d(filter_nums[res_idx], filter_nums[res_idx], 3, stride=1, padding=1, bias=False),
+                    nn.InstanceNorm3d(filter_nums[res_idx]),
+                    nn.Conv3d(filter_nums[res_idx], num_classes, 1),
+                    nn.Upsample(scale_factor=2 ** res_idx, mode="trilinear", align_corners=True),
+                )
+
+        # define NAS search space
```
wyli, line 242: `class DiNTS` becomes a complex python object (in terms of functionality). is it possible to decouple it into smaller pieces, for example, making this search space a standalone, small and compact pytorch nn.Module?

wyli (commented):
thanks, please see the comments in-line, and also there are some style checks https://deepsource.io/gh/Project-MONAI/MONAI/run/553698d2-c29b-4f8c-998a-4433e3c90521/python/ for your consideration

monai/networks/blocks/dints_block.py
```diff
@@ -0,0 +1,107 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+import torch.nn as nn
+
+__all__ = ["FactorizedIncreaseBlock", "FactorizedReduceBlock", "P3DReLUConvBNBlock", "ReLUConvBNBlock"]
+
+
+class FactorizedIncreaseBlock(nn.Module):
```
Nic-Ma, line 18: Maybe we can change the class name to be more general?  Does `scale_factor` must be `2`? If not, can we make the `scale_factor`,  inerpolation `mode` and `align_corners` as configurable parameters?

Thanks.

monai/networks/blocks/dints_block.py
```diff
@@ -0,0 +1,107 @@
...
+        self.bn = nn.InstanceNorm3d(self._out_channel)
+
+    def forward(self, x):
+        x = self.relu(x)
+        out = torch.cat([self.conv_1(x), self.conv_2(x[:, :, 1:, 1:, 1:])], dim=1)
+        out = self.bn(out)
+        return out
+
+
+class P3DReLUConvBNBlock(nn.Module):
```
Nic-Ma, line 61: Hopefully you can add more doc-string descriptions for these classes.

Thanks.

monai/networks/nets/dints.py
```diff
@@ -0,0 +1,556 @@
...
+                        path_activation * _code_prob_a[blk_idx] + (1 - path_activation) * (1 - _code_prob_a[blk_idx])
+                    ).prod(-1)
+                    / norm[blk_idx]
+                )
+            probs_a = torch.stack(probs_a)
+            return probs_a, code_prob_a
+        else:
+            return None, code_prob_a
+
+    def _get_memory_usage(self, in_size, cell_memory: bool = False, code: bool = None, full: bool = False):
```
Nic-Ma, line 305: I feel maybe it's better to use `buffer` or other word instead of `memory`?

Thanks.

monai/networks/blocks/dints_block.py
dongyang0122: Currently the `scale_factor` is fixed as 2 in the algorithm. other values have not been used before.

monai/networks/blocks/dints_block.py
```diff
@@ -0,0 +1,211 @@
...
+
+from monai.networks.layers.factories import Conv
+from monai.networks.layers.utils import get_act_layer, get_norm_layer
+
+__all__ = ["FactorizedIncreaseBlock", "FactorizedReduceBlock", "P3DReLUConvNormBlock", "ReLUConvNormBlock"]
+
+
+class FactorizedIncreaseBlock(nn.Module):
+    """
+    Up-sampling the feature by 2 using stride.
```
wyli, line 26: don't think this is using stride to do upsampling. I'm updating the docstring

monai/networks/blocks/dints_block.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+
+    def __init__(
+        self,
+        in_channel: int,
+        out_channel: int,
+        spatial_dims: int = 3,
+        act_name: Union[Tuple, str] = "RELU",
+        norm_name: Union[Tuple, str] = "INSTANCE",
```
wyli, line 35: it's necessary to create docstrings to explain each input argument.
for example 
https://github.com/Project-MONAI/MONAI/blob/a4e7ba0325ed30cfbb91d8b21ca698de18ef9829/monai/networks/blocks/downsample.py#L37-L41

monai/networks/blocks/dints_block.py
wyli: > don't think this is using stride to do upsampling. I'm updating the docstring

I include it in https://github.com/dongyang0122/MONAI/pull/1

wyli:
could you please help fix the flake8 issues https://github.com/Project-MONAI/MONAI/runs/4263587551?check_suite_focus=true#step:7:118, the local commands to check them are here https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#checking-the-coding-style

monai/networks/blocks/dints_block.py
```diff
@@ -0,0 +1,211 @@
...
+                dilation=1,
+            ),
+            get_norm_layer(name=norm_name, spatial_dims=3, channels=self._out_channel),
+        )
+
+    def forward(self, x):
+        return self.op(x)
+
+
+class ReLUConvNormBlock(nn.Module):
```
wyli, line 177: if the `act_name` are flexible now, the class name shouldn't have `ReLU`

tests/test_p3d.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2020 - 2021 MONAI Consortium
```
wyli, line 1: could you please rename this file to be more informative

monai/networks/nets/dints.py
```diff
@@ -0,0 +1,690 @@
...
+                        bias=False,
+                        dilation=1,
+                    ),
+                    get_norm_layer(name=norm_name, spatial_dims=spatial_dims, channels=filter_nums[res_idx - 1]),
+                    nn.Upsample(scale_factor=2 ** (res_idx != 0), mode="trilinear", align_corners=True),
+                )
+
+    def forward():
+        pass
+
```
wyli, line 276: will this work?

monai/networks/blocks/dints_block.py
Nic-Ma: I think maybe can refer to these args:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/networks/nets/unet.py#L121-L122

Thanks.

tests/test_factorized_reduce.py
```diff
@@ -0,0 +1,35 @@
...
+
+import torch
+from parameterized import parameterized
+
+from monai.networks.blocks.dints_block import FactorizedReduceBlock
+
+TEST_CASES_3D = [
+    [{"c_in": 32, "c_out": 16}, (7, 32, 24, 16, 8), (7, 16, 12, 8, 4)],
+    [{"c_in": 16, "c_out": 32}, (7, 16, 23, 15, 7), (7, 32, 11, 7, 3)],
+    # [{'c_in':1, 'c_out':2}, (1, 1, 1, 1, 1), (1, 2, 0, 0, 0)]
```
Nic-Ma, line 22: Please remove this comment line.

Thanks.

Nic-Ma:
Hi @wyli @dongyang0122 ,

I am trying to add some TorchScript unit tests for DiNTS network in this PR, should I submit a PR to @dongyang0122 's repo or just commit to this PR directly?

Thanks.

Nic-Ma:
There are TorchScript test errors:
1. torch.jit.frontend.UnsupportedNodeError: annotated assignments without assigned value aren't supported:
  File "/workspace/data/medical/MONAI/monai/networks/nets/dints.py", line 366
        blk_idx = self.num_blocks - 1
        start = False
        _temp: torch.Tensor
        ~ <--- HERE
        for res_idx in range(self.num_depths - 1, -1, -1):
            if start:

2. torch.jit.frontend.UnsupportedNodeError: GeneratorExp aren't supported:
  File "/workspace/data/medical/MONAI/monai/networks/nets/dints.py", line 112
            weight: architecture weights for cell operations.
        """
        return sum(self.ops[idx.item()](x) * ops[idx.item()] * weight[idx.item()] for idx in (ops == 1).nonzero())
                  ~ <--- HERE

I already fixed the first issue in the commit, tried to fix the second one but didn't solve it.. @dongyang0122 @wyli Do you guys have any idea to change `sum(self.ops[idx.item()](x) * ops[idx.item()] * weight[idx.item()] for idx in (ops == 1).nonzero())`?

Thanks.

Nic-Ma:
Add @ericspod as Eric is an expert on TorchScript logic.

Thanks in advance.

wyli:
Could we consider torchscript later? I'm still asking Dong to refactor the details, I don't think torchscript should block this PR.

Nic-Ma:
> Could we consider torchscript later? I'm still asking Dong to refactor the details, I don't think torchscript should block this PR.

OK, sure, we can consider TorchScript later.
But I think we should support TorchScript in 0.8 as we declared all the networks support TorchScript, especially when we try to use it in NVIDIA Clara later, it must support TorchScript export.

Thanks.

ericspod:
For the Torchscript:
1. It's saying what the issue is, something of the form `var: Type` isn't allowed without assigning something to `var` because this needs to be translated into an equivalent C++ statement. You can safely assign `x` to it in this case since it will be set to something else immediately in the following loop.
2. Generated expressions like this aren't allowed so you'll have to define this with a explicit loop. 
`sum(self.ops[idx.item()](x) * ops[idx.item()] * weight[idx.item()] for idx in (ops == 1).nonzero())` just becomes:
```python
    ops_sum:float =0
    for idx in range(len(ops)):
        if ops[idx] == 1:
            ops_sum += self.ops[idx](x) * ops[idx.item()] * weight[idx.item()]
```
This is likely faster anyhow.

Nic-Ma:
> For the Torchscript:
> 
> 1. It's saying what the issue is, something of the form `var: Type` isn't allowed without assigning something to `var` because this needs to be translated into an equivalent C++ statement. You can safely assign `x` to it in this case since it will be set to something else immediately in the following loop.
> 2. Generated expressions like this aren't allowed so you'll have to define this with a explicit loop.
>    `sum(self.ops[idx.item()](x) * ops[idx.item()] * weight[idx.item()] for idx in (ops == 1).nonzero())` just becomes:
> 
> ```python
>     ops_sum:float =0
>     for idx in range(len(ops)):
>         if ops[idx] == 1:
>             ops_sum += self.ops[idx](x) * ops[idx.item()] * weight[idx.item()]
> ```
> 
> This is likely faster anyhow.

Hi @ericspod ,

Thanks very much for your help here.
I updated the code to:
```py
ops_sum = 0
for i, o in enumerate(ops):
    if o == 1:
        ops_sum += self.ops[i](x) * o * weight[i]
return ops_sum
```
But still got the TorchScript error about index:
```
RuntimeError: 
Expected integer literal for index. ModuleList/Sequential indexing is only supported with integer literals. Enumeration is supported, e.g. 'for index, v in enumerate(self): ...':
  File "/workspace/data/medical/MONAI/monai/networks/nets/dints.py", line 132
        for i, o in enumerate(ops):
            if o == 1:
                ops_sum += self.ops[i](x) * o * weight[i]
                           ~~~~~~~~~~~ <--- HERE
        return ops_sum
```
I tried to change the index value but didn't solve it.. do you have any idea?

Thanks in advance.

wyli:
I'm getting some compatible modules for torchscript, still testing them

ericspod:
@Nic-Ma,
Torchscript is finicky at times about what it will accept, it has to be compatible with C++ so it excludes many Python idioms like `enumerate`.  You will have to loop through indices old school keeping:

```python
ops_sum = 0
for i in range(len(ops)):
    if ops[i] == 1:
        ops_sum += self.ops[i](x) * ops[i] * weight[i]
return ops_sum
```

Torchscript can understand what to do with `for i in range(len(ops))` in terms of converting it to a construct like `for(int i=0;i<ops.size();i++)`.

wyli (approved):
We added the basic APIs for DiNTS, with both 2d and 3D support.
after the architecture search, the generated network structure is compatible with torchscript JIT. We may have follow-up PRs to revise the details, but the general functionality is tested locally with the tutorials.

PR #3352, opened 2021-11-17
3346 Simplify AsDiscrete transform

Nic-Ma:
Fixes #3346 .

### Description
This PR simplified `AsDiscrete` according to the feedback.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @wyli @ericspod ,

As we changed many args of the transform, I feel it's very confusing to maintain the deprecated args in it, so I deleted the deprecated args directly. Can we accept this slightly breaking change in v0.8?

Thanks.

monai/transforms/post/array.py
```diff
@@ -122,94 +122,65 @@ class AsDiscrete(Transform):
     Args:
         argmax: whether to execute argmax function on input data before transform.
             Defaults to ``False``.
-        to_onehot: whether to convert input data into the one-hot format.
-            Defaults to ``False``.
-        num_classes: the number of classes to convert to One-Hot format.
+        to_onehot: if not None, convert input data into the one-hot format with specified number of classes.
+            Defaults to ``None``.
+        threshold_values: if not None, threshold the float values to int number 0 or 1 with specified theashold.
```
wyli, line 127: threshold_values -> threshold

monai/transforms/post/array.py
```diff
@@ -122,94 +122,65 @@ class AsDiscrete(Transform):
...
 
     backend = [TransformBackends.TORCH]
 
-    @deprecated_arg("n_classes", since="0.6")
     def __init__(
         self,
         argmax: bool = False,
-        to_onehot: bool = False,
-        num_classes: Optional[int] = None,
-        threshold_values: bool = False,
```
wyli, line 149: this will break many tutorials, we may provide better error messages
```py
@deprecated_arg(name="threshold_values", new_name="threshold", since="0.8")
def __init__(threshold = None, ...):
   if threshold_values is True:
       raise ValueError("threshold_values=True is deprecated, please use threashold=value instead.")
```

Nic-Ma: Sure, let me add the error message.

Thanks.

wyli: had a typo `since="0.8"` should be 0.7, otherwise it won't work with the release candidates such as v0.8rc0

monai/transforms/post/array.py
```diff
@@ -122,94 +122,68 @@ class AsDiscrete(Transform):
     Args:
         argmax: whether to execute argmax function on input data before transform.
             Defaults to ``False``.
-        to_onehot: whether to convert input data into the one-hot format.
-            Defaults to ``False``.
-        num_classes: the number of classes to convert to One-Hot format.
+        to_onehot: if not None, convert input data into the one-hot format with specified number of classes.
+            Defaults to ``None``.
+        threshold: if not None, threshold the float values to int number 0 or 1 with specified theashold.
```
wyli, line 127: forgot the `threshold=False` case, should it be allowed (convert to None) or raise value error?

Nic-Ma: Hi @wyli ,

In that case, should we also add error message for `to_onehot=True/False`? That's why I said there are too many deprecated args here and making the code confusing.

Thanks.

Nic-Ma: Anyway, let me add the logic to improve the experience for users.

Thanks.

ericspod:
We should use the deprecated decorators to mark removed arguments and document it in the docstring, that should be enough for breaking changes though I know my code a lot of the tutorials will need changing. Raising exceptions for wrong value types would help as well. We're going to be changing other transforms a lot as well before 1.0 so I don't think this will be the only thing.

Nic-Ma:
Hi @wyli @ericspod ,

I updated the PR to add the deprecated information and raise error, could you please help review it again?

Thanks in advance.

monai/transforms/post/array.py
```diff
@@ -122,94 +122,108 @@ class AsDiscrete(Transform):
     Args:
         argmax: whether to execute argmax function on input data before transform.
             Defaults to ``False``.
-        to_onehot: whether to convert input data into the one-hot format.
-            Defaults to ``False``.
-        num_classes: the number of classes to convert to One-Hot format.
+        to_onehot: if not None, convert input data into the one-hot format with specified number of classes.
```
wyli, line 125: thanks, could you add a few lines of examples for different use cases?

Nic-Ma: Sure, added examples in the latest commit.

Thanks in advance.

wyli (approved):

PR #3355, opened 2021-11-17
Improve `TileOnGrid`

drbeh:
### Description
This PR works around having optional tile step size and removes unnecessary casting.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/apps/pathology/transforms/spatial/array.py
+++ b/monai/apps/pathology/transforms/spatial/array.py
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional, Sequence, Tuple, Union, cast
+from typing import Optional, Sequence, Tuple, Union
 
 import numpy as np
 import torch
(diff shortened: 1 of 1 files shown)


monai/apps/pathology/transforms/spatial/array.py
```diff
@@ -113,14 +113,16 @@ def __init__(
...
         self.tile_size = tile_size
-        self.step = step
         self.random_offset = random_offset
         self.pad_full = pad_full
         self.background_val = background_val
         self.filter_mode = filter_mode
 
-        if self.step is None:
-            self.step = self.tile_size  # non-overlapping grid
+        if step is None:
```
Nic-Ma, line 121: Can be: `self.step = step or self.tile_size`?
Or `self.step = self.tile_size if step is None step`.

Thanks.

Nic-Ma (commented):
May also add some unit tests to cover the cases?

Thanks.

monai/apps/pathology/transforms/spatial/array.py
drbeh: Thank you @Nic-Ma. The first one would fail if step is zero. For the second one, I like what you are suggesting (because it's one line) but I think it is more readable this way and also we are able to add comments to clarify the action.

drbeh:
> May also add some unit tests to cover the cases?
> 
> Thanks.

It is not a new case but I can see that this is what is missing from @myron's unittests.

drbeh:
Hi @Nic-Ma, I asked @myron to add the missing test case https://github.com/Project-MONAI/MONAI/issues/3359
so I think we can merge this. Thanks

Nic-Ma (approved):

PR #3358, opened 2021-11-18
3357 Add wrap_sequence to several transforms

### LINKED ISSUE #3357: Missing `wrap_sequence` in `ToNumpy` and `EnsureType`
These two transforms (also for their dict versions) miss the parameter `wrap_sequence`.

Nic-Ma:
Fixes #3357 .

### Description
This PR added the `wrap_sequence` parameter to several transforms.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @rijobro @wyli @yiheng-wang-nv ,

Please note that in this PR I set different default values of `wrap_sequence` for `utils(False)` and `transforms(True)`, following our current pattern in: https://github.com/Project-MONAI/MONAI/blob/dev/monai/utils/type_conversion.py and https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/utility/array.py#L356
I think it matches the common usage, we always convert a list of data(like classification labels) to a numpy array or tensor in the transforms.

Thanks.

monai/transforms/utility/array.py
```diff
@@ -377,41 +392,52 @@ def __init__(
...
+        dtype: target data type when converting to numpy array.
+        wrap_sequence: if `False`, then lists will recursively call this function, default to `True`.
+            E.g., if `False`, `[1, 2]` -> `[array(1), array(2)]`, if `True`, then `[1, 2]` -> `array([1, 2])`.
+
     """
 
     backend = [TransformBackends.TORCH, TransformBackends.NUMPY]
 
-    def __init__(self, dtype: DtypeLike = None) -> None:
+    def __init__(self, dtype: DtypeLike = None, wrap_sequence: bool = True) -> None:
```
yiheng-wang-nv, line 431: Hi @Nic-Ma , here the default value of `wrap_sequence` has been changed from `False` (used directly in `convert_to_numpy`) to `True`, but I do not see new test case about this class, thus is it not covered?

Nic-Ma: Hi @yiheng-wang-nv ,

The unit tests use `type_test=False` to compare results, so it didn't cover this case, let update to add more tests.

Thanks.

Nic-Ma: Added tests in the latest commit.

Thanks.

Nic-Ma:
Hi @wyli @yiheng-wang-nv ,

Do you have any other concerns about this PR?

Thanks in advance.

yiheng-wang-nv (approved):
Thanks, looks good on me.

PR #3362, opened 2021-11-19
Update setup.cfg with tifffile and imagecodes

drbeh:
### Description
Add  `tifffile` and `imagecodes` dependencies as build options dependencies in setup.cfg.

### Status
**Ready**

### Types of changes

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -23,3 +23,5 @@ einops
 transformers
 mlflow
 tensorboardX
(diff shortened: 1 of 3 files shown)


Nic-Ma:
Could you please help also add to other optional installation related parts?
You can search "pandas" as reference and see where we need to add.

Thanks.

wyli (approved):

PR #3373, opened 2021-11-19
3370 option of zip_longest to decollate

### LINKED ISSUE #3370: `decollate_batch` return an empty list
`decollate_batch` will return an empty list, if input is a dict, and one of its value is an empty list.
For example:
```
data = {
    "class": ['ToTensord', 'ToTensord'],
    "id": torch.tensor([1, 2]),
    "orig_size": [],
}
```
`decollate_batch(data)` will return `[]`

This case comes from a classification task, where I used `SupervisedTrainer`, and labels are achieved from `ToNumpyd`. The `data` in the above example belongs to `label_transforms` which is in `engine.state.batch`

wyli:

Fixes #3370

### Description
defaults to pad the to batch data to be decollated when the items suggest different batch sizes.

### Status
**Ready**

### Types of changes


monai/transforms/inverse_batch_transform.py
```diff
@@ -73,17 +75,23 @@ def __init__(
                 if set to `None`, use the `num_workers` of the transform data loader.
             detach: whether to detach the tensors. Scalars tensors will be detached into number types
                 instead of torch tensors.
+            pad_batch: when the items in a batch indicate different batch size,
+                whether to pad all the sequences to the longest.
+                If False, the batch size will be the length of the shortest sequence.
+            fill_value: the value to fill the padded sequences when `pad_batch=True`.
```
Nic-Ma, line 81: Could you please help add the default behavior when `fill_value=None`?

Thanks.

monai/data/utils.py
```diff
@@ -297,7 +296,24 @@ def list_data_collate(batch: Sequence):
...
+    returns batch_size and the list of non-iterable items.
+    """
+    batch_size, non_iterable = 0, []
+    for k, v in data.items() if isinstance(data, Mapping) else enumerate(data):
+        if not isinstance(v, Iterable) or isinstance(v, (str, bytes)) or (isinstance(v, torch.Tensor) and v.ndim == 0):
+            # Not running the usual list decollate here:
+            # don't decollate ['test', 'test'] into [['t', 't'], ['e', 'e'], ['s', 's'], ['t', 't']]
+            # torch.tensor(0) is iterable but iter(torch.tensor(0)) raises TypeError: iteration over a 0-d tensor
+            non_iterable.append(k)
+        elif hasattr(v, "__len__"):
```
Nic-Ma, line 311: I think maybe we should only check the length of `Tensor` or `Numpy array`?
For example, below test code will output batch_size=`3` instead of `2`:
```py
data = {"image": torch.Tensor([1, 2]), "label": torch.Tensor([3, 4]), "meta_dict": {"filename": ["test1", "test2"], "affine": [0, 0], "original_shape": [1, 1]}}
print(_non_zipping_check(data)[0])
```

Thanks.

wyli: that'll be wrong for `_non_zipping_check({"image": [1, 2, 3], "label": [1, 2]})`. this function is only used after `[decollate_batch(b, detach, pad=pad, fill_value=fill_value) for b in batch]`, I'll try to revise to improve the readability.

Nic-Ma: Hi @wyli ,

Thanks for your quick update, I think the functionality is correct now.
But I feel it's still very confusing because `_non_zipping_check` calls `decollate_batch` and `decollate_batch` also calls `_non_zipping_check` in the recursive way.
I understand you like to make every APIs available to use directly, but In this case, I think `_non_zipping_check` should not be used independently, so maybe we don't need to set 2 separate functions, `_non_zipping_check` can be just an inline function of `decollate_batch` or we define smaller inline functions to make the logic easier to read?
What do you think?

Thanks.

wyli: I'd like to keep it a separate function, and starting with underscore means a private function.

Nic-Ma: It's OK to keep in 2 functions, is it possible to avoid `A calls B and B calls A`? Maybe put the `decollate` logic out of `_non_zipping_check`?

Thanks.

wyli: That's my previous version, I think it causes confusions. This one also shows the fact that the batch size calculation is always for dict/list and after the decollate. Let me know if you have a better idea, I'm happy to change.

Nic-Ma: OK, let's merge this PR to unblock v0.8 release first, I will also think about it, maybe we can improve it later.

Thanks.

Nic-Ma (approved):

Nic-Ma:
Hi @wyli ,

Seems `/integration-test` doesn't work properly?

Thanks.

monai/data/utils.py
wyli: sure, this function is currently not very useful outside of this decollate process, perhaps we can improve it if we find some other useful cases.

PR #3385, opened 2021-11-23
3368 add `frame_dim` to TensorBoard plot utility

### LINKED ISSUE #3368: Add `frame_dim` to the matshow3d utility
Currently, for higher dimensional data (like `BCHWD`), `matshow3d` will expand to a list of frame images on the flattened `BCH` dim. Would be useful to add a `frame_dim` arg to support any dim.

Nic-Ma:
Fixes #3368 .

### Description
This PR mainly contains 4 parts:
1. Added support to specify `frame_dim` to plot the GIF image or RGB video in TensorBoard, similar to #3369 .
2. Removed `add_animated_gif_no_channels` as the input data shape actually doesn't work with `make_animated_gif_summary` and no place really calls this API. I think it was ported from other codebase by mistake.
3. Removed the `animation_axes` and `image_axes` args from the private function. They are actually not used(already declared in the doc-string before).
4. Added some missing doc-strings.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @ericspod @wyli ,

I also tested the `frame_dim` parameter in the tutorial:
https://github.com/Project-MONAI/tutorials/blob/master/modules/transform_visualization.ipynb
The output GIF image is as expected.

Thanks.

monai/visualize/img2tensorboard.py
```diff
@@ -67,18 +73,18 @@ def make_animated_gif_summary(
     image: Union[np.ndarray, torch.Tensor],
     writer=None,
     max_out: int = 3,
-    animation_axes: Sequence[int] = (3,),
-    image_axes: Sequence[int] = (1, 2),
-    other_indices: Optional[Dict] = None,
+    frame_dim: int = -3,
```
wyli, line 76: I feel the previous `image_axes` is more powerful, with the flexibility of doing transpose operation, why removing it? also the docstring still has the option `image_axes`

Nic-Ma: Hi @wyli ,

Thanks for your review and feedback.
I think for most cases, we only need to specify the dim as frames, so I think a `frame_dim` is easier to understand and use, and also align with our `matshow3d` API.
Do you have any use cases that need to specify frame dim and also transpose the left 2 spatial dims? Maybe we don't need to consider this case so far? I suspect very few users used this API directly.
Sorry for the mistake in doc-string, I will update it ASAP.

Thanks.

wyli: yes, a potential use case: https://github.com/Project-MONAI/tutorials/issues/351

Nic-Ma: OK, thanks for your sharing.
Let me try to update ASAP.

Nic-Ma: Hi @wyli @ericspod ,

I checked again the previous code logic in `make_animated_gif_summary`, if I understand correctly, actually it doesn't have the transpose logic(same as the doc-string said it's not used):
https://github.com/Project-MONAI/MONAI/blob/dev/monai/visualize/img2tensorboard.py#L92
`image = image[tuple(slicing)]` just do redundant logic as all the image axes are in `axis_order`.
Do you mean to implement `transpose` logic to all these visualization functions and `matshow3d`?

Thanks.

Nic-Ma: Hi @ericspod @rijobro ,

What do you guys think about this feature? Should we just use `frame_dim` to specify the dimension as frames, or use `image_axes` and `frame_axis` to powerfully support `transpose` all the axes?

Thanks in advance.

wyli (approved):

PR #3389, opened 2021-11-23
Adding repeats option to ThreadDataLoader

ericspod:

### Description

### Status
**Ready**

### Types of changes


tests/test_thread_buffer.py
```diff
@@ -78,6 +78,22 @@ def test_time(self):
                 f"Buffered time {buffered_time} should be less than unbuffered time {unbuffered_time}",
             )
 
+    def test_dataloader_repeats(self):
+        dataset = Dataset(data=self.datalist, transform=self.transform)
```
wyli, line 82: I suspect the repeat will generate incorrect output if `self.transform` makes in-place changes to the `data`

ericspod: The transforms are run only once per batch, the result from the transform sequence is yielded multiple times.

PR #3390, opened 2021-11-23
stain to color

drbeh:
### Description
This PR change the digital pathology stain transforms to color transforms to have consistency with cuCIM transforms and to make it more generalizable for future developments.

### Status
**Ready**

### Types of changes


wyli (changes requested):
But I don't think we need to change this for now, we can update when we have conflicting concepts (when introducing colour transforms that are not part of the stain processing). Also, to be consistent with `monai.transforms` it should be `intensity`.

drbeh:
That's a possibility but `color` and `intensity`, although very related, are a bit different. In radiology, we don't have this color concept and color profile since we are dealing with grayscale images but in pathology, the images are in RGB color space (or other color spaces).

For instance, cuCIM uses color transforms in addition to intensity transforms: https://github.com/rapidsai/cucim/tree/branch-22.02/python/cucim/src/cucim/core/operations

drbeh:
> But I don't think we need to change this for now, we can update when we have conflicting concepts (when introducing colour transforms that are not part of the stain processing). Also, to be consistent with `monai.transforms` it should be `intensity`.

I believe it's better to put components where they are supposed to be rather than fixing it later, which will cause backward incompatibility.

wyli:
> > But I don't think we need to change this for now, we can update when we have conflicting concepts (when introducing colour transforms that are not part of the stain processing). Also, to be consistent with `monai.transforms` it should be `intensity`.
> 
> 
> 
> I believe it's better to put components where they are supposed to be rather than fixing it later, which will cause backward incompatibility.

Is there a use case for this? Also, would be great to review the pathology apps naming and module structures, looks like we had restructurings multiple times, maybe it's time to finalise the structure design for v1.0.

drbeh:
I agree! We should finalize it for v1.0.

PR #3393, opened 2021-11-24
3387 Fix 0.0 or None value in AsDiscrete

### LINKED ISSUE #3387: `AsDiscreted` bug when `threshold=0.0`
Reproduce the bug:

```
from monai.transforms import AsDiscreted
a = AsDiscreted(keys="pred", threshold=0.0)
```

Screen Shot 2021-11-23 at 10 56 39 PM

Nic-Ma:
Fixes #3387 .

### Description
This PR fixed the None or 0.0 value of to_onehot and threshold at runtime.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @yiheng-wang-nv ,

Please note that, we don't support the case that: set the `to_onehot` or `threshold` not to None at init(), and set the `to_onehot` or `threshold` to None at call() to **disable** the ToOneHot or Threshold actions.
It can only support to **enable** the action at runtime.
I think it's same as all the `align_corners` arg of call() in the spatial transforms:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/spatial/array.py#L552

Thanks.

yiheng-wang-nv (approved):
Thanks @Nic-Ma

tests/test_as_discrete.py
```diff
@@ -45,6 +45,16 @@
         ]
     )
 
+    # test threshold = 0.0
+    TEST_CASES.append(
```
wyli, line 49: these new cases still don't test the runtime args?

wyli: ok I saw your comments (but this test case becomes unnecessary)

PR #3421, opened 2021-11-30
only push lambdad transform if inverse is given

rijobro:
Only add the lambdad transform to the list of invertible transforms if an inverse function has been supplied.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/utility/dictionary.py
+++ b/monai/transforms/utility/dictionary.py
@@ -1003,7 +1003,8 @@ class Lambdad(MapTransform, InvertibleTransform):
             ret = self._transform(data=d[key], func=func)
             if overwrite:
                 d[key] = ret
-            self.push_transform(d, key)
+            if self.inv_func is not None:
+                self.push_transform(d, key)
         return d
 
     def _inverse_transform(self, transform_info: Dict, data: Any, func: Callable):


monai/transforms/utility/dictionary.py
```diff
@@ -1003,7 +1003,8 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> Dict[Hashable, N
             ret = self._transform(data=d[key], func=func)
             if overwrite:
                 d[key] = ret
-            self.push_transform(d, key)
+            if self.inv_func is not None:
```
Nic-Ma, line 1006: Hi @rijobro ,

As the type-hints stated, the input data type of `inv_func` is `Callable` or `Sequence[Callable]`, so maybe we don't need to check the `None` case here? I see we usually don't check unexpected python data types so far..

And BTW, even checking it, `self.inv_func = ensure_tuple_rep(inv_func, len(self.keys))` makes `self.inv_func` to be a tuple of `None`, so `if self.inv_func is not None` will be False, should check every item of the tuple.
What do you think?

Thanks.

rijobro:
Hi @Nic-Ma thanks for the review. I've realised I can do this in an alternative way, so closing the PR.

PR #3422, opened 2021-11-30
check if orig_size is list of tensors

rijobro:
Sometimes after the dataloader, `transform[TraceKeys.ORIG_SIZE]` is given as a list of `torch.Tensor`. Calling `np.asarray` on this will return a numpy array containing a list (`dtype=object`), which causes problems. This fix will convert to a single `torch.Tensor` if necessary and then convert that to a numpy array.

I'm not quite sure the best way to add a test for this. I suspect all our inverses should also be run through the DataLoader, maybe with and `EnsureType` at the end to simulate the numpy->torch conversion that we normally experience.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/croppad/dictionary.py
+++ b/monai/transforms/croppad/dictionary.py
@@ -23,6 +23,7 @@ from math import ceil, floor
 from typing import Any, Callable, Dict, Hashable, List, Mapping, Optional, Sequence, Tuple, Union
 
 import numpy as np
+import torch
 
 from monai.config import IndexSelection, KeysCollection
 from monai.config.type_definitions import NdarrayOrTensor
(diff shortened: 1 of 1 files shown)


monai/transforms/croppad/dictionary.py
```diff
@@ -315,7 +316,10 @@ def inverse(self, data: Mapping[Hashable, NdarrayOrTensor]) -> Dict[Hashable, Nd
         for key in self.key_iterator(d):
             transform = self.get_most_recent_transform(d, key)
             # Create inverse transform
-            orig_size = np.array(transform[TraceKeys.ORIG_SIZE])
+            orig_size = transform[TraceKeys.ORIG_SIZE]
+            if isinstance(orig_size[0], torch.Tensor):
+                orig_size = torch.as_tensor(orig_size)
+            orig_size = np.asarray(orig_size)
```
Nic-Ma, line 322: Hi @rijobro ,

I think maybe you can use the `convert_data_type` utility here to convert to numpy array with `wrap_sequence` arg?

Thanks.

PR #3427, opened 2021-12-01
3293 Remove extra deep supervision modules of DynUNet

### LINKED ISSUE #3293: Further enhance DynUNet via removing unused parts
In current implementation, no matter `deep_supervision` is used, the corresponding heads will be used. In addition, extra heads may be maintained, when `deep_supr_num` is less than `the number of up sample layer - 1`. It may also be related with https://github.com/Project-MONAI/MONAILabel/issues/492 

I planned to change it within #3265 , but it brought breaking changes (https://github.com/Project-MONAI/MONAI/pull/3265/commits/574fe60f721b0918dd740dd19ac93dbff1967d72) and may make users' old weights cannot be used.

Therefore, here I open another ticker, and I will do a further enhancement.

yiheng-wang-nv:

Fixes #3293 

### Description
Our current implementation of DynUNet may have extra modules:

1. When `deep_supervision=False`, `self.deep_supervision_heads` is created but useless.
2. When `deep_supervision=True` and `len(self.upsamples) - 1 > deep_supr_num`, extra heads may be created but they will not be used in `forward`.

With these extra modules, when using `DistributedDataParallel`, `find_unused_parameters=True` must be specified ( see: [here](https://github.com/Project-MONAI/MONAILabel/issues/492)).

This PR enhances the network structure and remove the potential extra modules, and also solve the issue mentioned in https://github.com/Project-MONAI/MONAILabel/issues/492

Whereas,  one thing needs to be considered is that when users want to reuse weights achieved from the old version, they may need to remove the extra key and values in the old state dict ( like what I changed in [test_network_consistency.py](https://github.com/Project-MONAI/MONAI/pull/3427/files#diff-ed0900ba2f36662a1420528a81103add6142c97bc8e963792874cbec40e3522eR67)). Therefore, do you have any suggestions for this breaking changes? should I add a function like `def load_old_version_weights()`? @ericspod @wyli @Nic-Ma 

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/networks/nets/dynunet.py
+++ b/monai/networks/nets/dynunet.py
@@ -79,6 +79,8 @@ class DynUNet(nn.Module):
     For example, if `strides=((1, 2, 4), 2, 1, 1)`, the minimal spatial size of the input is `(8, 16, 32)`, and
     the spatial size of the output is `(8, 8, 8)`.
 
+    For backwards compatibility with old weights, please set `strict=False` when calling `load_state_dict`.
+
     Usage example with medical segmentation decathlon dataset is available at:
     https://github.com/Project-MONAI/tutorials/tree/master/modules/dynunet_pipeline.
 
--- a/monai/networks/nets/dynunet.py
+++ b/monai/networks/nets/dynunet.py
@@ -102,17 +102,16 @@ class DynUNet(nn.Module):
         norm_name: feature normalization type and arguments. Defaults to ``INSTANCE``.
         act_name: activation layer type and arguments. Defaults to ``leakyrelu``.
         deep_supervision: whether to add deep supervision head before output. Defaults to ``False``.
-            If ``True``, in training mode, the forward function will output not only the last feature
-            map, but also the previous feature maps that come from the intermediate up sample layers.
+            If ``True``, in training mode, the forward function will output not only the final feature map
+            (from `output_block`), but also the feature maps that come from the intermediate up sample layers.
             In order to unify the return type (the restriction of TorchScript), all intermediate
-            feature maps are interpolated into the same size as the last feature map and stacked together
+            feature maps are interpolated into the same size as the final feature map and stacked together
             (with a new dimension in the first axis)into one single tensor.
-            For instance, if there are three feature maps with shapes: (1, 2, 32, 24), (1, 2, 16, 12) and
-            (1, 2, 8, 6). The last two will be interpolated into (1, 2, 32, 24), and the stacked tensor
-            will has the shape (1, 3, 2, 8, 6).
+            For instance, if there are two intermediate feature maps with shapes: (1, 2, 16, 12) and
+            (1, 2, 8, 6), and the final feature map has the shape (1, 2, 32, 24), then all intermediate feature maps
+            will be interpolated into (1, 2, 32, 24), and the stacked tensor will has the shape (1, 3, 2, 32, 24).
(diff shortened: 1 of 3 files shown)


Nic-Ma:
Hi @ericspod @wyli ,

Could you please help share some feedback as @yiheng-wang-nv asked?

Thanks in advance.

monai/networks/nets/dynunet.py
```diff
@@ -180,30 +180,52 @@ def create_skips(index, downsamples, upsamples, superheads, bottleneck):
...
+                        next_layer=next_layer,
+                        heads=self.heads,
+                        super_head=superheads[0],
+                    )
+                else:
+                    return DynUNetSkipLayer(
+                        index, downsample=downsamples[0], upsample=upsamples[0], next_layer=next_layer
+                    )
+
+        if not self.deep_supervision:
```
ericspod, line 217: Is this `if-else` block reachable? The above `if-else` block has returns in both parts, the latter in a `if-else` based on the boolean flag `super_head_flag`. I think in any possibility this will hit a return unless I missed something.

tests/test_network_consistency.py
```diff
@@ -61,7 +61,13 @@ def test_network_consistency(self, net_name, data_path, json_path):
 
         # Create model
         model = nets.__dict__[net_name](**model_params)
-        model.load_state_dict(loaded_data["model"])
+        state_dict = loaded_data["model"]
```
ericspod, line 64: I would suggest `model = getattr(nets, net_name)(**model_params)` for the previous line.

tests/test_network_consistency.py
```diff
@@ -61,7 +61,13 @@ def test_network_consistency(self, net_name, data_path, json_path):
...
+        model_dict = model.state_dict()
+        state_dict = {
+            k: v for k, v in state_dict.items() if (k in model_dict) and (model_dict[k].shape == state_dict[k].shape)
+        }
+        model_dict.update(state_dict)
+        model.load_state_dict(model_dict)
```
ericspod, line 70: Can you load weights directly from `state_dict` without the filtering if you set the `strict` argument to `False`?

ericspod:
In regards to the state dict loading, if `strict=False` will work then we should comment in the class' docstring that this is necessary for backwards compatibility with old weights. 

If not then having a well-named function for doing the filtering as you do in the test would be the way to go. The generalized idea is that you're loading only the weights which match in name and shape as those in the target, this is a generalized concept that would go into a utility function you can use here, again noting its need in the docstring.

monai/networks/nets/dynunet.py
yiheng-wang-nv: Hi @ericspod , thanks for the review, the above `if-else` block is actually in `def create_skips`

tests/test_network_consistency.py
yiheng-wang-nv: Thanks, this method works

wyli:
Thanks, could you please look into the style warnings? https://deepsource.io/gh/Project-MONAI/MONAI/run/3fe27cd5-0727-4e8a-80fe-af5e9146090d/python/

yiheng-wang-nv:
Hi @wyli , thanks, I updated the code format.

wyli (approved):
Thanks, it looks good to me. In the network consistency test, the `strict` flag is set to false as a workaround for https://github.com/Project-MONAI/MONAI-extra-test-data/issues/9

PR #3428, opened 2021-12-01
[WIP] Enhance optional return data of ViT network

Nic-Ma:
Fixes https://github.com/Project-MONAI/tutorials/issues/464

### Description
This PR added support to optional return feature maps of ViT network, and also added unit tests for TorchScript.

### Status
**Work in progress**

### Types of changes


yiheng-wang-nv:
HI @Nic-Ma , I think the issue comes from: https://github.com/Project-MONAI/MONAI/blob/f94e76824cf310d908e407394bda3b0507b3d3c0/monai/networks/blocks/patchembedding.py#L97

It is not related this PR's changes, and the current implementation of ViT does not support torchscript support. We may need to fix that issue first.

Nic-Ma:
Hi @yiheng-wang-nv ,

Thanks very much for your investigation here.
As the issue is not related to this PR, would you like to submit a separate PR to fix the TorchScript issue first?

Thanks in advance.

monai/networks/nets/vit.py
```diff
@@ -109,4 +111,4 @@ def forward(self, x):
         x = self.norm(x)
         if self.classification:
             x = self.classification_head(x[:, 0])
-        return x, hidden_states_out
+        return (x, hidden_states_out) if self.get_hidden_states else x
```
wyli, line 114: I feel this kind of variable return types for the network design should be discussed cc @ericspod @atbenmurray

Nic-Ma: OK, so @ericspod @rijobro what's your opinion about the return data type?
And please note that we can't define a separate function to get the feature maps out of `forward()` due to distributed data-parallel requirements.

Thanks.

ericspod: I suspect Torchscript doesn't like the fact that the return type differs based on your condition. If `self.get_hidden_states` is True the return type is a pair containing a tensor and a list of tensors, otherwise it's a tensor. You'll have to return the same thing regardless of your state, so if `self.get_hidden_states` is False you return a pair containing the tensor and an empty list:

```python
    def forward(self, x):
            x = self.patch_embedding(x)
            if self.classification:
                cls_token = self.cls_token.expand(x.shape[0], -1, -1)
                x = torch.cat((cls_token, x), dim=1)
            hidden_states_out = []

            if self.get_hidden_states
                for blk in self.blocks:
                    x = blk(x)
                    hidden_states_out.append(x)

            x = self.norm(x)
            if self.classification:
                x = self.classification_head(x[:, 0])
            return x, hidden_states_out
```

Nic-Ma: Hi @ericspod ,

Thanks very much for your help and example here.
I think the initial feature request is to make the ViT can only return `x` in `forward()` to be compatible with most other networks. Return feature maps only if the option is `True`.
But we can't get feature maps in a separate function due to distributed data parallel issue which requires gradients computation must be in `forward()`.
Do you have any idea?

Thanks in advance.

ericspod: Can you store the feature maps in a object attribute which is typed correctly but initially `None`, then return only `x`? That may work and be Torchscript compatible, allowing these objects to be retrieved later after `forward` is called. You may have to initalize this to something empty if Torchscript can't figure out what to do with it. If not then you're at an impass the architecture of Pytorch/Torchscript creates. We may want to consider changing our inferer classes to accommodate tuple returns such that they will return a selected member from the tuple instead of the whole thing to maintain compatibility. With that change you can return a tuple always. This still isn't the standard expectation of a Pytorch network but would work with your original intent.

Nic-Ma: Hi @ericspod ,

Thanks for your suggestion.
It may solve the TorchScript issue, but as I described in the previous comment, I remember if we compute loss with the values not returned from `forward()`, it may cause PyTorch distributed data-parallel issue, related to: https://discuss.pytorch.org/t/how-to-use-torch-nn-parallel-distributeddataparallel-and-torch-utils-checkpoint-together/96338.
What do you think?

Thanks.

ahatamiz:
Hi @Nic-Ma

Thanks for the efforts here. But I recommend to consider the usage of ViT for UNETR when modifying it:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/networks/nets/unetr.py#L191

The original issue which led to this PR could have been easily avoided upon looking at the type of the final outputs. I can do a tutorial on this ViT classification to demonstrate its usage. Maybe that's a better solution than changing it ?

monai/networks/nets/vit.py
ericspod: If we want to have a `forward` returning only one value for the network itself but need it to return two when training with DDP, we may need a wrapper class whose `forward` collects the two values from a delegated instance of the network. Something like:

```python
class DDPWrapper(nn.Module):
    def __init__(self, net):
        self.net = net
    def forward(self,x):
        pred=self.net(x)
        return pred, self.net.other_things()
```

We otherwise might be stuck with an incompatibility between what DDP needs and adhering to the protocol of one return value. I think this kind of wrapper should work to train `net`.

Nic-Ma:
Hi @ahatamiz ,

Thanks very much for your help here, sure, looking forward to your tutorial on this ViT classification. Let's have more discussion based on your tutorial later.

Thanks.

Issue #2620, opened 2021-07-16
Image writer with itk backend to support more output image formats


wyli:
**Is your feature request related to a problem? Please describe.**
The `SaveImage` transform should be extended to support writing output in different formats using ITK.
https://github.com/Project-MONAI/MONAI/blob/830f8bb34532f63e188b43738f804fff6f059d5b/monai/transforms/io/array.py#L260-L261

the current nifti output doesn't work well with Slicer https://github.com/Project-MONAI/MONAILabel/issues/211

see also https://github.com/InsightSoftwareConsortium/ITK/issues/2663

Nic-Ma:
Hi @wyli ,

I have a question about the feature request of this ticket:
Currently, we have `nifti_saver` & `nifti_writer` and `png_saver` & `png_writer`, maybe there are 2 options:
1. If we just want to add support to save in DICOM format, we can add `dicom_saver` & `dicom_writer` based on ITK APIs.
2. If want to support many other formats, I think we need to totally refactor the saving part and make it just like the inverse of the `ImageReader` and `LoadImage`:
(1). Remove `nifti_saver` & `nifti_writer` and `png_saver` & `png_writer`.
(2). Add `nibabel_writer` for NIfTI, `pil_writer` for PNG, JPG, JPEG, BMP, `itk_writer` for any other formats.
(3). Use `SaveImage` transform as high-level API, mapping to `LoadImage`.

What do you think or you guys have other ideas?

Thanks.

wyli:
I think as an initial step we could add a minimal itk_writer and make it an option in SaveImage in v0.7. The goal is to encourage the use of SaveImage and we'll refactor the underlying implementation later (beyond 0.7) -- For refactoring the writer, option 2 looks good, we still need to review the quality of these writer APIs and probably get comments/proposals from the working groups.

Nic-Ma:
Hi @wyli ,

Plan makes sense, we can develop a `itk_writer` and call it in `SaveImage` directly instead of a `itk_saver` layer. When it's done in v0.7, we can refactor `nifti_writer` & `nifti_saver` to `nibabel_writer`, and refactor `png_writer` & `png_saver` to `pil_writer` based on the same pattern of `itk_writer`.
I can develop a PR for `ITKWriter` with the similar design of `ITKReader`, may not support many formats in the first step.

Thanks.

PR #3443, opened 2021-12-03
[poc] 2620 - image writer

wyli:

part of #2620 #3595

### Description
ongoing discussions about the image writer module

### Status
**on hold**

### Types of changes


monai/data/image_writer.py
```diff
@@ -0,0 +1,725 @@
...
+    "register_writer",
+    "resolve_writer",
+]
+
+SUPPORTED_WRITERS: Dict = {}
+
+
+def register_writer(data_format, *im_writer):
+    """
+    Register ``ImageWriter``, so that writing a file of ``data_format``
```
ericspod, line 69: Format is mentioned in a number of places but we don't specify what we mean by format. Is this meant to be the file extension or some identifier of a format? For common files like "png" it would be both the extension and the name of the format, but for JPEG we have "jpg" or "jpeg" as valid extensions so which would be use as the format name needs to be resolved. Same with NIfTI, the extension could be "nii" or "nii.gz".

This raises the question if we want to support any file compressed with gzip adding ".gz" to the extension, a file with "png" or "png.gz" should resolve to the same writers.

monai/data/image_writer.py
```diff
@@ -0,0 +1,725 @@
...
+            affine = np.eye(4, dtype=np.float64)
+        affine = to_affine_nd(sr, affine)  # type: ignore
+        target_affine = to_affine_nd(sr, target_affine) if target_affine is not None else affine
+
+        if np.allclose(affine, target_affine, atol=AFFINE_TOL):
+            # no affine changes, return (data, affine)
+            return data, ensure_mat44(target_affine)
+
+        # resolve orientation
+        start_ornt = nib.orientations.io_orientation(affine)
```
ericspod, line 344: It might make sense to put this code into a separate utility function if the same sort of operation is needed elsewhere:

```python
def convert_nifti_transform(init_affine, target_affine, data_shape):
    """Return an NIfTI-resolved affine representing the transform from `init_affine` space to `target_affine` space."""
    start_ornt = nib.orientations.io_orientation(init_affine)
    target_ornt = nib.orientations.io_orientation(target_affine)
    ornt_transform = nib.orientations.ornt_transform(start_ornt, target_ornt)
    _affine = affine @ nib.orientations.inv_ornt_aff(ornt_transform, data_shape)
    return _affine
```

(I think the docstring is right but do check)

monai/data/image_writer.py
```diff
@@ -0,0 +1,725 @@
...
+        target_ornt = nib.orientations.io_orientation(target_affine)
+        ornt_transform = nib.orientations.ornt_transform(start_ornt, target_ornt)
+        data_shape = data.shape
+        data = nib.orientations.apply_orientation(data, ornt_transform)
```
ericspod, line 348: This line could be in the `if` clause below since it isn't needed unless we can get away without resampling?

monai/data/image_writer.py
```diff
@@ -0,0 +1,725 @@
...
+            data = data.astype(dtype, copy=False)
+        if reverse_indexing:
+            data = np.moveaxis(data, 0, 1)
+
+        return PILImage.fromarray(data, mode=kwargs.pop("image_mode", None))
+
+
+def init():
+    """Initialize the image writer modules according to the file extension name."""
+    register_writer(".png", PILWriter)
```
ericspod, line 721: We can add ".jpg", ".bmp", ".tif", and others to use PILWriter. TIF files are sometimes used for 16-data other formats don't store well.

monai/transforms/io/array.py
```diff
@@ -289,55 +288,53 @@ class SaveImage(Transform):
...
+            output_dir=output_dir,
+            postfix=output_postfix,
+            extension=output_ext,
+            parent=separate_folder,
+            makedirs=True,
+            data_root_dir=data_root_dir,
+        )
+
+        self.output_ext = output_ext
+        self.writers = image_writer.resolve_writer(output_format or self.output_ext) if writer is None else (writer,)
```
ericspod, line 325: Should use `self.output_ext.lower()` just to be safe?

monai/transforms/io/array.py
```diff
@@ -346,6 +343,29 @@ def __call__(self, img: Union[torch.Tensor, np.ndarray], meta_data: Optional[Dic
...
+                (
+                    writer_cls(output_dtype=self.output_dtype, scale=self.scale)
+                    .set_data_array(img, channel_dim=0, squeeze_end_dims=self.squeeze_end_dims)
+                    .set_metadata(meta_dict=meta_data, resample=self.resample, **self.resample_dict)
+                    .write(filename, verbose=self.print_log)
+                )
+            except Exception as e:
+                logging.getLogger(self.__class__.__name__).exception(e, exc_info=True)
+                logging.getLogger(self.__class__.__name__).info(
+                    f"{writer_cls.__class__.__name__}: unable to load {filename}."
```
ericspod, line 361: "...Unable to write..."

monai/data/image_writer.py
wyli: thanks, addressed by https://github.com/Project-MONAI/MONAI/pull/3674/commits/e189fbfc155670a71a4a198e64a1e0fca05d7a5e.

monai/data/image_writer.py
wyli: added a `reorient_spatial_axes` util https://github.com/Project-MONAI/MONAI/pull/3674/commits/e189fbfc155670a71a4a198e64a1e0fca05d7a5e

monai/data/image_writer.py
wyli: thanks, for now I updated to have the mapping "filename extension" <> imagewriter,  we can have an additional layer such as: "filename extension" <> fileformat <> imagewriter the latest is at https://github.com/Project-MONAI/MONAI/pull/3773.

wyli:
comments are addressed by multiple PRs, this becomes out-of-date, please see https://github.com/Project-MONAI/MONAI/issues/3595

PR #3457, opened 2021-12-08
improve error message if reader not available

rijobro:
When a valid reader is not present, I get the following error:

```
RuntimeError: can not find a suitable reader for file: ('rawdata/sub-verse752/sub-verse752_dir-ax_ct.nii.gz',).
    Please install the reader libraries, see also the installation instructions:
    https://docs.monai.io/en/latest/installation.html#installing-the-recommended-dependencies.
   The current registered: [<monai.data.image_reader.NumpyReader object at 0x7f703531f160>, <monai.data.image_reader.PILReader object at 0x7f703531f220>].
```

I found this confusing since I gave the filename as a string but the error was being printed as a tuple. This PR improves the error message by converting back to string if instance is a tuple and length is 1.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/io/array.py
+++ b/monai/transforms/io/array.py
@@ -204,6 +204,8 @@ class LoadImage(Transform):
                         break
 
         if img is None or reader is None:
+            if isinstance(filename, tuple) and len(filename) == 0:
+                filename = filename[0]
             raise RuntimeError(
                 f"can not find a suitable reader for file: {filename}.\n"
                 "    Please install the reader libraries, see also the installation instructions:\n"
--- a/monai/transforms/io/array.py
+++ b/monai/transforms/io/array.py
@@ -204,10 +204,10 @@ class LoadImage(Transform):
                         break
 
         if img is None or reader is None:
-            if isinstance(filename, tuple) and len(filename) == 0:
+            if isinstance(filename, tuple) and len(filename) == 1:
                 filename = filename[0]
             raise RuntimeError(
-                f"can not find a suitable reader for file: {filename}.\n"
+                f"cannot find a suitable reader for file: {filename}.\n"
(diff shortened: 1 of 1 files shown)


monai/transforms/io/array.py
```diff
@@ -204,6 +204,8 @@ def __call__(self, filename: Union[Sequence[PathLike], PathLike], reader: Option
                         break
 
         if img is None or reader is None:
+            if isinstance(filename, tuple) and len(filename) == 0:
```
Nic-Ma, line 207: May I know why the length equals to `0`?

Thanks.

rijobro: Apologies, should be 1!

rijobro:
Tested with:

```bash
conda create -n quick_test -y
conda activate quick_test
conda install pytorch cpuonly -c pytorch
conda install numpy -y
touch test.nii.gz
python -c "import monai; monai.transforms.LoadImage()('test.nii.gz')"
```
which gives:
```
RuntimeError: cannot find a suitable reader for file: test.nii.gz.
    Please install the reader libraries, see also the installation instructions:
    https://docs.monai.io/en/latest/installation.html#installing-the-recommended-dependencies.
   The current registered: [<monai.data.image_reader.NumpyReader object at 0x7f064cb10d00>].
```

instead of the previous error:
```
RuntimeError: cannot find a suitable reader for file: ('test.nii.gz',).
    Please install the reader libraries, see also the installation instructions:
    https://docs.monai.io/en/latest/installation.html#installing-the-recommended-dependencies.
   The current registered: [<monai.data.image_reader.NumpyReader object at 0x7f33156e3d00>].
```

Nic-Ma (approved):

PR #3478, opened 2021-12-13
3465 Support string as dtype

### LINKED ISSUE #3465: Support string dtype in transforms when processing Tensor data
As numpy array can accept the `dtype` arg in string, so many previous programs use `dtype="float32"` in the MONAI transforms. If we put `ToTensor` before the transform to make it work on GPU Tensor, it will fail because PyTorch Tensor doesn't support string for dtype.
It broken many NVIDIA Clara MMARs when switching to GPU transforms.

Would be nice to enhance the `get_equivalent_dtype` utility to properly convert the string dtype of numpy array for PyTorch Tensor.

Nic-Ma:
Fixes #3465 .

### Description
This PR added support to use `string` as dtype, especially for transforms to be compatible with numpy array and tensor.
Similar to ticket: https://github.com/pytorch/pytorch/issues/40568.

### Status
**Ready**

### Types of changes


wyli (approved):
Thanks, it looks good to me.

PR #3484, opened 2021-12-14
Improve NVTX Range Naming

drbeh:
### Description
This PR make some improvement in NVTX Range. 
- To not have redundant numbering when a transform is used once, it does not append number to the first call of a transform but will start the numbering for the next ones from two (e.g., Transform, Transform_2, Transform_3, etc) 
- When wrapper transforms are being used (i.e. CuCIM and TorchVision), it extract the underlying transfrom name that are being called and append to the name of the transform for profiling.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/utils/nvtx.py
+++ b/monai/utils/nvtx.py
@@ -62,6 +62,26 @@ class Range:
         # Define the name to be associated to the range if not provided
         if self.name is None:
             name = type(obj).__name__
+            # If CuCIM or TorchVision transform wrappers are being used,
+            # append the underlying transform to the name for more clarity
+            if name in [
+                "CuCIM",
+                "CuCIMd",
+                "CuCIMD",
+                "CuCIMDict",
+                "RandCuCIM",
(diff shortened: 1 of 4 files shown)


dongyang0122 (commented):
The changes look good to me.

monai/utils/nvtx.py
```diff
@@ -62,8 +62,31 @@ def __call__(self, obj: Any):
         # Define the name to be associated to the range if not provided
         if self.name is None:
             name = type(obj).__name__
+            # If CuCIM or TorchVision transform wrappers are being used,
+            # append the underlying transform to the name for more clarity
+            if name in [
```
wyli, line 67: I suspect its more intuitive with
```py
if "CuCIM" in name or "TorchVision" in name:
```

drbeh: Yeah, I like it! Thanks!

wyli (approved):
Thanks, it looks good to me.

PR #3493, opened 2021-12-15
Make bending energy loss invariant to resolution

### LINKED ISSUE #3485: Bending energy loss is not scale invariant
Bending energy loss is intended to be applied to a vector field to tell us something about the regularity of the image deformation that the vector field defines when interpreted as a displacement map. I think such a measure should not depend on the resolution of the images that are involved, and so it should also not depend on the spatial resolution used to describe the vector field.

**The problem:** Currently, `BendingEnergyLoss` does vary with image resolution, scaling roughly as the inverse square of the image resolution.

**Why it's a problem:** This makes it tricky to use `BendingEnergyLoss` in practice. Often it is the regularization objective for training a registration model. The weight to put on a regularization objective like this is typically a very sensitive hyperparameter. Having to retune it every time one uses a different image resolution (e.g. in a fully convolutional network where different resolutions can be used) is prohibitively inconvenient. One can always correct for the way in which it scales, but this makes it overly technical to use properly.

**Demo:** I made a [demo notebook here](https://github.com/ebrahimebrahim/deep-atlas/blob/main/bending_loss_scale_test.ipynb) while trying to work out the issue for myself. There's an estimate of the scaling law, and a plot that shows that the estimate agrees with the empirical result. There's no need to look at this notebook to understand the issue or make comments below-- I'm just linking it here as a supplement.

**Example:** To show an example from the notebook, the following two vector fields are the "same" vector field, but at different resolutions:

vector fields

The bending loss should not be so different for these two resolutions.

**Clarification:** I do believe the current implementation is correctly following the bending energy formula. What I'm suggesting is that we should divide out the image size and work with a "unitless" version of that formula instead. This would be more useful.

**Clarification 2:** Perhaps a better term is "resolution-invariant." When I say "scale-invariant" I'm talking about image scale, not the scale of the displacement vectors alone. A good way to think about it is in terms of units. The components of a displacement vector field might be described in meters, and the underlying spatial locations at which those vectors live are also described in meters. If one converts _both_ to centimeters, then that shouldn't change the bending energy loss. And it wouldn't, if we understand bending energy loss to have units. But normally when dealing with loss functions we do not think about units. It should be made unitless.

**Previous discussion:** I see that something similar to this issue was [previously discussed](https://github.com/Project-MONAI/MONAI/issues/1405), but I don't understand how it was resolved. Tagging people involved in that discssion: @kate-sann5100 @YipengHu @wyli 

**Potential fix:** I can contribute a fix-- I suggest one at the bottom of the demo notebook that I linked. However I'd first like to hear if there's agreement about this... _Should bending energy loss be scale invariant?_

ebrahimebrahim:
Fixes #3485

### Description
See #3485 for a complete description.

- Introduce `normalize` parameter to `BendingEnergyLoss` constructor, which defaults to `True`. This makes loss computation robust to different image resolutions. Setting `normalize` to `False` explicitly provides the old behavior, which follows the "bending energy" formula as it appears in publications.
- Raise error when number of vector components does not match the number of spatial dimensions.
- Add test cases for the new error.
- Adjust old test cases to expect the new bending loss behavior, and add test cases to test both behaviors
- Adjust tensor shapes in old test cases so that the number of vector components always matches the number of spatial dimensions.

**Feedback wanted:** Should `normalize=True` be the default? I preferred this because I felt that the default behavior should be resolution-invariant... but this is also the reason it's a breaking change. If `BendingEnergyLoss` is already used in some project, then a revision would be needed in that project. If I set `normalize=False` by default, it's not a breaking change.

### Status
**Ready**

- [x] Run some less trivial tests
- [x] Get feedback to my question above

### Types of changes

### DIFF
--- a/monai/losses/deform.py
+++ b/monai/losses/deform.py
@@ -52,12 +52,12 @@ class BendingEnergyLoss(_Loss):
         DeepReg (https://github.com/DeepRegNet/DeepReg)
     """
 
-    def __init__(self, normalize: bool = True, reduction: Union[LossReduction, str] = LossReduction.MEAN) -> None:
+    def __init__(self, normalize: bool = False, reduction: Union[LossReduction, str] = LossReduction.MEAN) -> None:
         """
         Args:
             normalize:
                 Whether to divide out spatial sizes in order to make the computation roughly
-                invariant to image scale (i.e. vector field sampling resolution). Defaults to True.
+                invariant to image scale (i.e. vector field sampling resolution). Defaults to False.
             reduction: {``"none"``, ``"mean"``, ``"sum"``}
                 Specifies the reduction to apply to the output. Defaults to ``"mean"``.
 
--- a/monai/losses/deform.py
+++ b/monai/losses/deform.py
@@ -52,9 +52,12 @@ class BendingEnergyLoss(_Loss):
         DeepReg (https://github.com/DeepRegNet/DeepReg)
     """
 
-    def __init__(self, reduction: Union[LossReduction, str] = LossReduction.MEAN) -> None:
+    def __init__(self, normalize: bool = True, reduction: Union[LossReduction, str] = LossReduction.MEAN) -> None:
         """
         Args:
+            normalize:
+                Whether to divide out spatial sizes in order to make the computation roughly
+                invariant to image scale (i.e. vector field sampling resolution). Defaults to True.
             reduction: {``"none"``, ``"mean"``, ``"sum"``}
                 Specifies the reduction to apply to the output. Defaults to ``"mean"``.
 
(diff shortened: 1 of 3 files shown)


wyli (approved):
It looks good to me. Thanks for the additional input validation I was thinking about the same.

I'm inclined to default to False as that's the basic formulation described in the literature.

kate-sann5100 (approved):
Nice work! Looks good to me as well. You can merge after setting the default option to False if you wish.

ebrahimebrahim:
I think I agree with setting default to `False`. Maybe it's more important that the default behavior match usage of the term elsewhere, rather than being the most convenient behavior. Also it's nice to have it be non-breaking. I will make that change now.

wyli:
thank you, could you merge the latest `dev` branch into this PR?

ebrahimebrahim:
@wyli should be ready now!

PR #3529, opened 2021-12-21
Update CSVDataset in dataset.py

nvahmadi:
Fixes lack of control over `pd.read_csv()` parameters, e.g. it is now possible to denote the file delimiter.

### Description
`kwargs` argument currently addresses merge operation. There is no way to change `kwargs` for `read_csv ()` call (e.g. to change delimiter from standard `,` to `\t`. 
Suggestion: Have `kwargs_read_csv` and `kwargs_merge` arguments to allow separate and finer user-control over both loading & merging mechanism.

### Status
**Ready for review**

### Types of changes


monai/data/dataset.py
```diff
@@ -1314,22 +1314,23 @@ def __init__(
         col_types: Optional[Dict[str, Optional[Dict[str, Any]]]] = None,
         col_groups: Optional[Dict[str, Sequence[str]]] = None,
         transform: Optional[Callable] = None,
-        **kwargs,
+        **kwargs_read_csv: Optional[Dict] = {},
```
wyli, line 1317: thanks, these would be without the `**` and accepting dictionaries, could you please also update the unit tests in `tests/test_csv_dataset.py`? (to run the tests: `python -m tests.test_csv_dataset`)

Nic-Ma: Hi @nvahmadi ,

Could you please help also update CSVIterableDataset?
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/iterable_dataset.py#L122

Thanks.

PR #3530, opened 2021-12-21
3525 Fix invertible issue in OneOf compose

### LINKED ISSUE #3525: `OneOf` Compose can't support non-invertible transforms
Issue reported from internal SA team, `OneOf` compose method can't support non-invertible transforms.
Program to reproduce the issue:
```py
transform = Compose([
    Resized(keys="img", spatial_size=[100, 100, 100]),
    OneOf([
        RandScaleIntensityd(keys="img", factors=0.5, prob=1.0),
        RandShiftIntensityd(keys="img", offsets=0.5, prob=1.0),
    ]),
])
transform.set_random_state(seed=0)
result = transform({"img": np.ones((1, 101, 102, 103))})

result = transform.inverse(result)
print(result["img"].shape)
```
Got the error:
```
  File "/workspace/data/medical/MONAI/monai/transforms/transform.py", line 82, in apply_transform
    return _apply_transform(transform, data, unpack_items)
  File "/workspace/data/medical/MONAI/monai/transforms/transform.py", line 53, in _apply_transform
    return transform(parameters)
  File "/workspace/data/medical/MONAI/monai/transforms/compose.py", line 264, in inverse
    raise RuntimeError(f"Applied OneOf transform is not invertible (applied index: {index}).")
RuntimeError: Applied OneOf transform is not invertible (applied index: 1).

...
```

Nic-Ma:
Fixes #3525 .

### Description
This PR enhanced `OneOf` compose to support non-invertible transforms.

### Status
**Ready**

### Types of changes


monai/transforms/compose.py
```diff
@@ -256,12 +256,9 @@ def inverse(self, data):
                 # and then remove the OneOf transform
                 self.pop_transform(data, key)
         if index is None:
-            raise RuntimeError("No invertible transforms have been applied")
+            # no invertible transforms have been applied
```
wyli, line 259: I think this should still raise the error, `index` should be known after line 241

Nic-Ma: Hi @wyli ,

The previous logic is a little bit misleading: it saved duplicated `index` info for **every applied key** of **invertible transform**, so it may be multiple items in the data or not existing (if not invertible transform).
I updated to only save 1 index info in the data for this `OneOf` compose and added `deepcopy` logic as all the other inverse functions in invertible transforms.

Thanks in advance.

wyli: What is the issue of tracking every applied key? I think it's good. The modification here might cause confusions about the concepts of transform class key and data key. Class name is already tracked in the dict https://github.com/Project-MONAI/MONAI/blob/58ded727f1d3eeb21cf7b492afb53f07f13ddfe2/monai/transforms/inverse.py#L60

Nic-Ma: Hi @wyli ,

Actually, I am OK with either solution, maybe I didn't describe the previous issue clearly:
Here is the previous code logic:
```py
  _transform = self.transforms[index]
  data = apply_transform(_transform, data, self.map_items, self.unpack_items)
  # if the data is a mapping (dictionary), append the OneOf transform to the end
  if isinstance(data, Mapping):
      for key in data.keys():
          if self.trace_key(key) in data:  ---------> HERE
              self.push_transform(data, key, extra_info={"index": index})
```
It checks all the keys in the data, save the `index` info only when the **transform is invertible transform** and it handles this key(`keys=[...]`). So if the transform is not invertible, `index` will be None in the `inverse()` function, I changed it to not raise error for this case. Your previous comment asked to still raise error for None index, so I changed to always save the `index` info in this version. If you prefer to the previous change, I can reset it.
What do yo think?

Thanks.

wyli: I'm sorry that I misunderstood your previous solution, actually that fixes #3525 already.

Nic-Ma: Sure, no problem, I reverted the commit.

Thanks.

wyli (approved):
Thanks, it looks good to me.

PR #3551, opened 2021-12-29
Smooth Deform

ericspod:
### Description
This adds a non-rigid deformation transform class based on smooth fields which can be faster than the existing 2D/3D transforms. This includes changes to the `SmoothField` to improve speed and operate entirely with tensors. A few constructor variables and some behaviour has been changed as well as a few changes to the existing smooth field classes.

### Status
**Ready**

### Types of changes


monai/transforms/smooth_field/array.py
```diff
@@ -14,90 +14,163 @@
...
+        pad_slice = slice(None) if self.pad == 0 else slice(self.pad, -self.pad)
+        self.rand_slices = (0, slice(None)) + (pad_slice,) * len(self.rand_size)
+
+        self.set_spatial_size(spatial_size)
 
     def randomize(self, data: Optional[Any] = None) -> None:
-        self.field = self.R.uniform(self.low, self.high, (self.channels,) + self.rand_size)  # type: ignore
-        if self.padder is not None:
-            self.field = self.padder(self.field)
+        self.field[self.rand_slices] = torch.from_numpy(self.R.uniform(self.low, self.high, self.crand_size))
```
Nic-Ma, line 94: Maybe use `convert_data_type` to convert to `self.device` directly?
https://github.com/Project-MONAI/MONAI/blob/dev/monai/utils/type_conversion.py#L213

Thanks.

monai/transforms/smooth_field/array.py
```diff
@@ -14,90 +14,163 @@
...
+                recompute_scale_factor=False,
+            )
 
-        return rescale_array(resized_field, self.field.min(), self.field.max())
+            mina = resized_field.min()
+            maxa = resized_field.max()
+            minv = self.field.min()
+            maxv = self.field.max()
+
+            # faster than rescale_array (?)
```
Nic-Ma, line 134: May I know why it's faster than `rescale_array`?
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/utils.py#L151

Thanks.

monai/transforms/smooth_field/array.py
```diff
@@ -217,7 +325,137 @@ def __call__(self, img: np.ndarray, randomize: bool = True):
...
 
+        # everything below here is to be computed using the destination type (numpy, tensor, etc.)
+
         out = img * rfield
-        out, *_ = convert_to_dst_type(out, img, img.dtype)
 
         return out
+
+
+class RandSmoothDeform(RandomizableTransform):
```
Nic-Ma, line 335: Could you please help add the transform image for these 2 new transforms?
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/utils_create_transform_ims.py#L705
If you don't want to investigate it, I can help add it later when this PR is merged.

Thanks in advance.

monai/transforms/smooth_field/array.py
```diff
@@ -217,7 +325,137 @@ def __call__(self, img: np.ndarray, randomize: bool = True):
...
+        prob: probability transform is applied
+        def_range: value of the deformation range in image size fractions, single min/max value  or min/max pair
+        grid_dtype: type for the deformation grid calculated from the field
+        grid_mode: interpolation mode used for sampling input using deformation grid
+        grid_padding_mode: padding mode used for sampling input using deformation grid
+        grid_align_corners: if True align the corners when sampling the deformation grid
+        device: Pytorch device to define field on
+    """
+
+    backend = [TransformBackends.TORCH, TransformBackends.NUMPY]
```
Nic-Ma, line 358: As your computation is only on `Tensor`, the `backend` should be: `[TransformBackends.TORCH]`.
All the transforms support `numpy in numpy out` and `tensor in tensor out` so far.

Thanks.

monai/transforms/smooth_field/array.py
ericspod: We should be sticking to our defined conversion routines as a matter of uniformity but I know that `uniform` will give me a numpy array that's typically very small so this conversion is going to be much faster than `convert_data_type`. One if the issues I'm trying to address with the smooth field types is improving performance and some of these compromises may be necessary.

monai/transforms/smooth_field/array.py
ericspod: `rescale_array` rescales the input to the unit range then again to the requested range and has various checks, this avoids a lot of unneeded computation, and uses the inline Pytorch operators to avoid needless copying. Like the previous mentioned issue we're deviating from standard here for performance reasons so the question is if it's worth it.

monai/transforms/smooth_field/array.py
ericspod: I've added what I think should work for this, I've reduced the size of the random fields for all the smooth field transforms to (10,10,10) which should give better looking results.

monai/transforms/smooth_field/array.py
ericspod: Changed.

monai/transforms/smooth_field/array.py
Nic-Ma: Thanks for your update, could you please help also upload the added / updated images to:
https://github.com/Project-MONAI/DocImages

Thanks in advance.

monai/transforms/smooth_field/array.py
Nic-Ma: OK, I see your point.
Is it possible to also change the `self.field.to(self.device).clone()` part to this line?

Thanks.

monai/transforms/smooth_field/array.py
Nic-Ma: OK, BTW, do you think it's possible to enhance the `rescale_array` for better performance as you mentioned?
I think that would be also useful for other components.

Thanks.

Nic-Ma (approved):
Looks good to me.
@wyli @rijobro Do you have any comments about the `convert_data_type` and `rescale_array` related performance parts?

Thanks in advance.

monai/transforms/smooth_field/array.py
ericspod: `rescale_array` is first rescaling to unit range then to the desired range so it's doing something slightly different. It also has to check that the input's range is not zero but that's checked here in the constructor, so the time savings is somewhat particular to this application.

monai/transforms/smooth_field/array.py
ericspod: I changed it to remove `to(self.device)`, that part isn't needed but I do want to clone in case the unresized field is wanted for something and this should be fresh.

monai/transforms/smooth_field/array.py
ericspod: I can update the images but maybe better if you could do it after merge as I haven't gotten into it yet. We should update the readme of the DocImages repo to describe what to do or automate it somehow.

Nic-Ma: Sure, I can update that repo later.

Thanks.

monai/transforms/smooth_field/array.py
Nic-Ma: ok, I see.

Thanks.

monai/transforms/smooth_field/dictionary.py
```diff
@@ -81,67 +99,76 @@ def randomize(self, data: Optional[Any] = None) -> None:
...
 
         if not self._do_transform:
             return data
 
         d = dict(data)
 
-        for key in self.key_iterator(d):
+        for idx, key in enumerate(self.key_iterator(d)):
             if not self.apply_same_field:
                 self.randomize()  # new field for every key
```
wyli, line 112: for this `apply_same_field=False`, is it an equivalent of creating multiple instances
```py
rand_smooth = RandSmoothFieldAdjustContrastd("img", ...)
rand_smooth1 = RandSmoothFieldAdjustContrastd("img1", ...)
```
I think we should drop this option to simplify the API...

wyli: this comment is not resolved before merging... could you help clarify the use case of `apply_same_field`? @ericspod

ericspod: It was an edge case to consider if there was some reason to distort two images in different ways, eg. for registration training, and it would be more efficient than having two transforms. I can remove it in a separate PR if you think it's worth removing.

ericspod: https://github.com/Project-MONAI/MONAI/pull/3556

PR #3560, opened 2021-12-31
3559 Enhance `DatasetSummary` for several points

### LINKED ISSUE #3559: Enhance `DatasetSummary` for several points
Thanks for @dongyang0122 's great work to develop `DatasetSummary`, would be nice to enhance the below items:
1. It only has `meta_key_postfix` arg, missed `meta_key` to be consistent with other MONAI modules when using flexible meta key: https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/io/dictionary.py#L68
2. The `spacing` data or `image`, `label` may not be `Tensor`, it depends on the `Dataset` and `collate_fn` of DataLoader, need to convert data type to `Tensor` first.
2. It supports `kwargs` for DataLoader but didn't cover it in the unit tests, would be nice to add it.

Nic-Ma:
Fixes #3559 .

### Description
This PR enhanced the `DatasetSummary` for below items:
1. Added `meta_key` arg to be consistent with other MONAI modules.
2. Added support for non-Tensor data.
3. Added more unit tests for non-Tensor data and `**kwargs` for `DataLoader`.

### Status
**Ready**

### Types of changes


monai/data/dataset_summary.py
```diff
@@ -92,8 +100,8 @@ def get_target_spacing(self, spacing_key: str = "pixdim", anisotropic_threshold:
             self.collect_meta_data()
         if spacing_key not in self.all_meta_data[0]:
             raise ValueError("The provided spacing_key is not in self.all_meta_data.")
-
-        all_spacings = torch.cat([data[spacing_key][:, 1:4] for data in self.all_meta_data], dim=0).numpy()
+        all_spacings = concatenate(to_cat=[data[spacing_key][:, 1:4] for data in self.all_meta_data], axis=0)
```
wyli, line 103: Thanks, I think this class could be enhanced with the recent 'Cumulative' API, it will ensure torch tensors.

Nic-Ma: Ah, yes, let me try to update it with the `append` API of `Cumulative`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/metrics/metric.py#L215

Thanks for your reminder.

Nic-Ma: Hi @wyli ,

I spent much time today to try to use `Cumulative`, but seems it's not very suitable for this DatasetSummary.
Maybe I misunderstood your suggestion, I tried to use `Cumulative` for the `all_spacings`, but actually, here we:
1. Get all the data from saved meta data dictionaries in a for loop directly, not like the iterations append / extend of Cumulative.
2. It `concat` the result instead of `stack` the result in the `sync` of Cumulative:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/metrics/metric.py#L255

So I feel maybe we don't need to use `Cumulative` here? Or could you please help share more details about your suggestion?

Thanks in advance.

Nic-Ma:
new flake8 errors are fixed in https://github.com/Project-MONAI/MONAI/pull/3562.

Thanks.

monai/data/dataset_summary.py
wyli: Sure, I can look into this later, it's not a blocking issue.

wyli (approved):

monai/data/dataset_summary.py
Nic-Ma: Thanks, after some expepiments, I have some ideas but need to change the API or args, maybe we can have an online discussion about this topic when you looking into it later.

Thanks in advance.

PR #3570, opened 2022-01-04
498 Add logger_handler to LrScheduleHandler

Nic-Ma:
Fixes https://github.com/Project-MONAI/tutorials/discussions/498.

### Description
This PR added `logger_handler` to the `LrScheduleHandler`, same as other MONAI components, like:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/handlers/stats_handler.py#L57
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/utility/array.py#L561

### Status
**Ready**

### Types of changes


monai/handlers/lr_schedule_handler.py
```diff
@@ -47,6 +48,9 @@ def __init__(
                 `True` is epoch level, `False` is iteration level.
             step_transform: a callable that is used to transform the information from `engine`
                 to expected input data of lr_scheduler.step() function if necessary.
+            logger_handler: if `print_lr` is True, add additional handler to log the learning rate: save to file, etc.
+                all the existing python logging handlers: https://docs.python.org/3/library/logging.handlers.html.
+                the handler should have a logging level of at least `INFO`.
```
wyli, line 53: is this line 53 copied here by mistake?

Nic-Ma: Hi @wyli ,

It's not a mistake, the `logger_handler` should take level > INFO as we use `logger.info` to save:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/handlers/lr_schedule_handler.py#L85

Thanks.

tests/test_handler_stats.py
```diff
@@ -136,18 +136,19 @@ def _train_func(engine, batch):
             # set up testing handler
             stats_handler = StatsHandler(name=key_to_handler, tag_name=key_to_print, logger_handler=handler)
             stats_handler.attach(engine)
+            stats_handler.logger.setLevel(logging.INFO)
 
             engine.run(range(3), max_epochs=2)
             handler.close()
             stats_handler.logger.removeHandler(handler)
             with open(filename) as f:
                 output_str = f.read()
```
wyli, line 145: What was the issue with this test? I remember previously it worked

Nic-Ma: Hi @wyli ,

Sorry I forgot to describe the issue here.
The previous test code actually didn't run into the `assert`:
```py
for idx, line in enumerate(output_str.split("\n")):
    if grep.match(line):  # ---------------> here is FALSE, so following logic didn't run at all and passed
        if idx in [1, 2, 3, 6, 7, 8]:
            self.assertTrue(has_key_word.match(line))
```

Thanks.

wyli: Right, looks like we have the same issue for the other test cases https://app.codecov.io/gh/Project-MONAI/MONAI/blob/dev/tests/test_handler_stats.py

Nic-Ma: Hi @wyli ,

Nice catch by the tool! Let me update them all.

Thanks.

wyli (approved):
Thanks, it looks good to me.

tests/test_handler_stats.py
Nic-Ma: Updated all the wrong test cases in the latest commit.

Thanks.

Issue #3603, opened 2022-01-07
Remove `"image"`, `"label"`, `"image_meta_dict"` and `"label_meta_dict"` from codebase


rijobro:
**Is your feature request related to a problem? Please describe.**
As a first step towards https://github.com/Project-MONAI/MONAI/issues/2424, we should remove all explicit calls to `"image_meta_dict"` and `"label_meta_dict"`.

wyli:
primary data types are now migrated to metatenor

PR #3605, opened 2022-01-07
replace `"image"` and `"label"` with `CommonKeys.IMAGE` and `CommonKeys.LABEL`

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/3603.

This prepares the codebase for tracking meta data.

### Status
**Ready**

### Types of changes


wyli (commented):
I feel `DictPostFixes.META` is quite useful for linking data and metadata. not sure about `CommonKeys.IMAGE` and `CommonKeys.LABEL`, the "image" is more intuitive, and we shouldn't rely on these for data typing, as it will introduce framework-like contraints... cc @ericspod @Nic-Ma

rijobro:
Externally, users can still use `"image"`, I just thought that internally we should avoid using "magic" variables where possible which means converting `"image"` to `CommonKeys.IMAGE` etc. 

I can revert this component of the PR if you would like.

ericspod:
In general we shouldn't have magic strings to avoid issues with typos or using inconsistent names in places, our code should exhibit what we believe to be good practice. Users can indeed still use string literals if they want so they aren't constrained to using our constants. It is more verbose this way but I don't feel it's unintuitive, the names of our constants should be as descriptive as the string literals they replace.

wyli:
> the names of our constants should be as descriptive as the string literals they replace.

So how about renaming `CommonKeys.Image` to `DefaultKeyStr.Image`? or how to make sure we don't give an impression that the keys must be some object/class defined by MONAI? The keys can be any Hashable...

also, would be great to have a separate PR to introduce `DictPostFixes`, looks like we all happy with it.

rijobro:
Ok, this PR has been updated to only include `CommonKeys` changes.

tests/test_deepgrow_transforms.py
```diff
@@ -73,42 +79,42 @@
...
+    CommonKeys.IMAGE: np.arange(1000).reshape((1, 5, 10, 20)),
+    f"{CommonKeys.IMAGE}_meta_dict": {"foreground_cropped_shape": (1, 10, 20, 40)},
     "guidance": [[[6, 10, 14], [8, 10, 14]], [[8, 10, 16]]],
 }
 
 DATA_10 = {
-    "image": np.arange(9).reshape((1, 1, 3, 3)),
-    "image_meta_dict": {
+    CommonKeys.IMAGE: np.arange(9).reshape((1, 1, 3, 3)),
+    f"{CommonKeys.IMAGE}_meta_dict": {
```
ericspod, line 117: The "_meta_dict" part is now like a magic string, should we have that stored in `CommonKeys` and have `f"{CommonKeys.Image}_{CommonKeys.MetaDict}"`?

ericspod:
I tried going through everything but there's a lot of changes but it looks good what I saw. The only thing I mentioned was using "_meta_dict" as a literal in places. If we instead had `CommonKeys.MetaDict` then we could make keys like `f"{CommonKeys.Image}_{CommonKeys.MetaDict}"` or access dictionaries like `d[CommonKeys.Image, CommonKeys.MetaDict] = x` which will associate `x` with a pair containing the two key values. If we went with this approach there would be a lot more changes to make to use that syntax so not in this PR.

ericspod (approved):

tests/test_deepgrow_transforms.py
rijobro: see https://github.com/Project-MONAI/MONAI/pull/3671.

ericspod: Yeah that looks good though we should commit this PR first.

PR #3647, opened 2022-01-12
3578 Support single channel with OneHot format in KeepLargestConnectedComponents

Nic-Ma:
Fixes #3578 .

### Description
Thanks for @function2-llx feedback in #3578, this PR added support for the case that all the data are in OneHot format but some of them may only has 1 channel in `KeepLargestConnectedComponents` transform.

### Status
**Ready**

### Types of changes


monai/transforms/post/array.py
```diff
@@ -305,7 +305,11 @@ class KeepLargestConnectedComponent(Transform):
     backend = [TransformBackends.NUMPY]
 
     def __init__(
-        self, applied_labels: Union[Sequence[int], int], independent: bool = True, connectivity: Optional[int] = None
+        self,
+        applied_labels: Union[Sequence[int], int],
+        independent: bool = True,
+        connectivity: Optional[int] = None,
+        single_channel_onehot: bool = False,
```
wyli, line 312: thanks, I think we can have an `is_one_hot` flag indicating whether the input will be interpreted as one-hot. `img.shape[0]` is never a reliable condition for this check...

Nic-Ma: Hi @wyli ,

Thanks for your review and suggestion, adding this `is_one_hot` arg will make it breaking change, is it acceptable?

wyli: I see, sorry I misunderstood this, I guess the actual feature request is the question of `applied_label` indexing from 0 or 1?

single-channel onehot is a bit confusing, it's only possible when num_classes is 1 and the data are all zeros, which is not very useful in practice...

Nic-Ma: Hi @wyli ,

Or it can be the case as @holgerroth asked before, the data is in OneHot format and sliced with only 1 channel?
Anyway, I think you are right about "img.shape[0] is never a reliable condition for this check...", so I already totally updated in the latest commit to use `is_onehot` arg for it, I feel it's acceptable to change the API arg.
What do you think?

Thanks.

wyli (approved):
thanks, it looks good to me, we can add `is_onehot=None` to resume to the previous behaviour?

Nic-Ma:
> thanks, it looks good to me, we can add `is_onehot=None` to resume to the previous behaviour?

Thanks.

PR #3649, opened 2022-01-12
`EnsureChannelFirst`: avoid re-creation of `AddChannel`

rijobro:
### Description
Each time `EnsureChannelFirst.__call__` is called, `AddChannel` or `AsChannelFirst` is created and then immediately destroyed. Better to create both during the `__init__` and then reuse.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/inferers/inferer.py
+++ b/monai/inferers/inferer.py
@@ -30,7 +30,8 @@ class Inferer(ABC):
     Example code::
 
         device = torch.device("cuda:0")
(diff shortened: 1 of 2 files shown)


wyli (approved):
Thanks, it looks good to me.

rijobro:
Sorry, realise that since the channel dimension is determined during `__call__`, the AsChannelFirst object can't be created during the `__init__`. The `AddChannel` can, though.

Issue #3595, opened 2022-01-06
refactoring the image writer modules


wyli:
This is a follow-up on the image writer of [the previous I/O module proposals](https://github.com/Project-MONAI/MONAI/issues/856).

### Goal and anticipated benefits
The current image writing module has been primarily built around the NIfTI specifications using the nibabel package as the backend.
The goal is to redesign the module in order to decouple the universal and the backend-specific implementations.

This will make it easier to bring in various writer backends such as ITK-Python, therefore supporting different image formats and more powerful writer module customisations.

related tickets: https://github.com/Project-MONAI/MONAI/issues/2620 https://github.com/Project-MONAI/MONAI/issues/2613 https://github.com/Project-MONAI/MONAILabel/issues/211
proof of concept https://github.com/Project-MONAI/MONAI/pull/3443

### Details
The writer module should have (1) universal logic (2) backend-specific logic (3) mechanism for defaulting/selecting backends. Specifically:

(1) In the context of image writing, the universal logic handles the image-related outputs from the deep learning workflows, such as:
-  fetching data array and metadata
-  arrangements/resampling of spatial/channel dimensions
-  preparing target data types
- preparing output folder structures, potentially compatible with BIDS (brain imaging data structure) and NDWB (neurodata without boarders) 

(2) The backend-specific logic includes
- creating data representations according to the backend APIs
- calling the backend APIs to finish the data writing

(3) The backend selection logic operates based on the user-specified parameters, backend system availability, and current system default configurations.

wyli:
https://github.com/Project-MONAI/MONAI/pull/3443 is now a working POC, it provides:

- An `ImageWriter` base class with backend-agnostic utilities such as `resample_if_needed`
- Subclasses (`PILWriter`, `ITKWriter`, `NibabelWriter`) with backend-specific logic, implementing
  - `set_data_array`
  - `set_metadata`
  - `write`
- `register_writer`, `resolve_writer` so that it selects a suitable writer based on the file extension name
- `FolderLayout` for generating organised filenames
- Revised existing `SaveImage` transform to use the new writer module.

they are non-breaking changes, the primary usage will be:
https://github.com/Project-MONAI/MONAI/blob/3206fa64809790a1faf6b20ca7e97d8047d4c551/monai/transforms/io/array.py#L353-L356

and has the basic support for `.nrrd` and `.dcm` writing:
[...]

PR #3674, opened 2022-01-18
3595 3766 adds a base writer and an itk writer

wyli:

part of #3595

### Description
this PR adds a writer base class and an itkwriter subclass,
internally the affines are assumed in the `RAS` space to be aligned with nifti.

more consistency tests will be added when the nibabelwrite is in place.
the tests are ready in https://github.com/Project-MONAI/MONAI/pull/3443

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/image_writer.py
+++ b/monai/data/image_writer.py
@@ -1,0 +1,399 @@
+# Copyright (c) MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import TYPE_CHECKING, Mapping, Optional, Sequence, Union
+
+import numpy as np
+import torch
+
+from monai.apps.utils import get_logger
+from monai.config import DtypeLike, NdarrayOrTensor, PathLike
+from monai.data.utils import compute_shape_offset, ensure_mat44, to_affine_nd
+from monai.networks.layers import AffineTransform
+from monai.transforms.utils_pytorch_numpy_unification import ascontiguousarray, moveaxis
+from monai.utils import GridSampleMode, GridSamplePadMode, convert_data_type, optional_import, require_pkg
+
+AFFINE_TOL = 1e-3
+DEFAULT_FMT = "%(asctime)s %(levelname)s %(filename)s:%(lineno)d - %(message)s"
+logger = get_logger(module_name=__name__, fmt=DEFAULT_FMT)
+
+if TYPE_CHECKING:
+    import itk  # type: ignore
+    import nibabel as nib
+
+    has_itk = has_nib = has_pil = True
+else:
+    itk, has_itk = optional_import("itk", allow_namespace_pkg=True)
+    nib, has_nib = optional_import("nibabel")
+
+__all__ = ["ImageWriter", "ITKWriter"]
+
+
+class ImageWriter:
+    """
+    The class is a collection of utilities to write images to disk.
+
+    Main aspects to be considered are:
+
+        - dimensionality of the data array, arrangements of spatial dimensions and channel/time dimensions
+            - ``convert_to_channel_last()``
+        - metadata of current affine and output affine, the data array should be converted accordingly
+            - ``get_meta_info()``
+            - ``resample_if_needed()``
+        - data type of the output image
+            - as part of ``resample_if_needed()``
+
+    Subclasses of this class should implement the backend-specific functions:
+
+        - ``set_data_array()`` to set the data array (numpy array or torch tensor)
+            - this calls the base class utilities for array handling
+        - ``set_metadata()`` to set the metadata and output affine
+            - this calls the base class utilities for metadata (affine, resmaple) handling
+        - backend-specific data object
+            - ``create_backend_obj()``
+        - backend-specific writing function
+            - ``write()``
+
+    The primary usage is:
+
+    .. code-block:: python
+
+        writer = MyWriter()  # subclass of ImageWriter
+        writer.set_data_array(data_array)
+        writer.set_metadata(meta_dict).write(filename_or_obj)
+        writer.write(filename)
+
+    Create an image writer object based on ``data_array`` and ``metadata``.
+    Spatially it supports up to three dimensions (with the resampling step
+    supports both 2D and 3D).
+
+    When saving multiple time steps or multiple channels `data_array`, time
+    and/or modality axes should be the at the `channel_dim`. For example,
+    the shape of a 2D eight-class and channel_dim=0, the segmentation
+    probabilities to be saved could be `(8, 64, 64)`; in this case
+    ``data_array`` will be converted to `(64, 64, 1, 8)` (the third
+    dimension is reserved as a spatial dimension).
+
+    The ``metadata`` could optionally have the following keys:
+
+        - ``'original_affine'``: for data original affine, it will be the
+            affine of the output object, defaulting to an identity matrix.
+        - ``'affine'``: it should specify the current data affine, defaulting to an identity matrix.
+        - ``'spatial_shape'``: for data output spatial shape.
+
+    When ``metadata`` is specified and ``resample=True``, the saver will
+    try to resample data from the space defined by `"affine"` to the space
+    defined by `"original_affine"`, for more details, please refer to the
+    ``convert_to_target_affine`` method.
+    """
+
+    def __init__(self, **kwargs):
+        self.data_obj = None
+        for k, v in kwargs.items():
+            setattr(self, k, v)
+
+    def set_data_array(self, data_array, **kwargs):
+        raise NotImplementedError(f"Subclasses of {self.__class__.__name__} must implement this method.")
+
+    def set_metadata(self, meta_dict: Optional[Mapping], **options):
+        raise NotImplementedError(f"Subclasses of {self.__class__.__name__} must implement this method.")
+
+    def write(self, filename_or_obj: PathLike, verbose: bool = True, **kwargs):
+        """subclass should implement this method to call the backend-specific writing APIs."""
+        if verbose:
+            logger.info(f"writing: {filename_or_obj}")
+
+    @classmethod
+    def create_backend_obj(cls, data_array: NdarrayOrTensor, **kwargs) -> np.ndarray:
+        """
+        Subclass should implement this method to return a backend-specific data representation object.
+        This method is used by ``cls.create_data_obj`` and the input ``data_array`` is 'channel-last'.
+        """
+        return convert_data_type(data_array, np.ndarray)[0]  # type: ignore
+
+    @classmethod
+    def resample_if_needed(
+        cls,
+        data_array: NdarrayOrTensor,
+        affine: Optional[NdarrayOrTensor] = None,
+        target_affine: Optional[np.ndarray] = None,
+        output_spatial_shape: Union[Sequence[int], np.ndarray, None] = None,
+        mode: Union[GridSampleMode, str] = GridSampleMode.BILINEAR,
+        padding_mode: Union[GridSamplePadMode, str] = GridSamplePadMode.BORDER,
+        align_corners: bool = False,
+        dtype: DtypeLike = np.float64,
+    ):
+        """
+        Convert the ``data_array`` into the coordinate system specified by
+        ``target_affine``, from the current coordinate definition of ``affine``.
+
+        If the transform between ``affine`` and ``target_affine`` could be
+        achieved by simply transposing and flipping ``data_array``, no resampling
+        will happen.  Otherwise, this function resamples ``data_array`` using the
+        transformation computed from ``affine`` and ``target_affine``.
+
+        This function assumes the NIfTI dimension notations. Spatially it
+        supports up to three dimensions, that is, H, HW, HWD for 1D, 2D, 3D
+        respectively. When saving multiple time steps or multiple channels,
+        time and/or modality axes should be appended after the first three
+        dimensions. For example, shape of 2D eight-class segmentation
+        probabilities to be saved could be `(64, 64, 1, 8)`. Also, data in
+        shape `(64, 64, 8)` or `(64, 64, 8, 1)` will be considered as a
+        single-channel 3D image. The ``convert_to_channel_last`` method can be
+        used to convert the data to the format described here.
+
+        Note that the shape of the resampled ``data_array`` may subject to some
+        rounding errors. For example, resampling a 20x20 pixel image from pixel
+        size (1.5, 1.5)-mm to (3.0, 3.0)-mm space will return a 10x10-pixel
+        image. However, resampling a 20x20-pixel image from pixel size (2.0,
+        2.0)-mm to (3.0, 3.0)-mm space will output a 14x14-pixel image, where
+        the image shape is rounded from 13.333x13.333 pixels. In this case
+        ``output_spatial_shape`` could be specified so that this function
+        writes image data to a designated shape.
+
+        Args:
+            data_array: input data array to be converted.
+            affine: the current affine of ``data_array``. Defaults to identity
+            target_affine: the designated affine of ``data_array``.
+                The actual output affine might be different from this value due to precision changes.
+            output_spatial_shape: spatial shape of the output image.
+                This option is used when resampling is needed.
+            mode: available options are {``"bilinear"``, ``"nearest"``, ``"bicubic"``}.
+                This option is used when resampling is needed.
+                Interpolation mode to calculate output values. Defaults to ``"bilinear"``.
+                See also: https://pytorch.org/docs/stable/nn.functional.html#grid-sample
+            padding_mode: available options are {``"zeros"``, ``"border"``, ``"reflection"``}.
+                This option is used when resampling is needed.
+                Padding mode for outside grid values. Defaults to ``"border"``.
+                See also: https://pytorch.org/docs/stable/nn.functional.html#grid-sample
+            align_corners: boolean option of ``grid_sample`` to handle the corner convention.
+                See also: https://pytorch.org/docs/stable/nn.functional.html#grid-sample
+            dtype: data type for resampling computation. Defaults to
+                ``np.float64`` for best precision. If ``None``, use the data type of input data.
+        """
+        data: np.ndarray
+        data, *_ = convert_data_type(data_array, np.ndarray)  # type: ignore
+
+        sr = min(data.ndim, 3)
+        if affine is None:
+            affine = np.eye(4, dtype=np.float64)
+        affine = to_affine_nd(sr, affine)  # type: ignore
+        target_affine = to_affine_nd(sr, target_affine) if target_affine is not None else affine
+
+        if np.allclose(affine, target_affine, atol=AFFINE_TOL):
+            # no affine changes, return (data, affine)
+            return data, ensure_mat44(target_affine)
+
+        # resolve orientation
+        if has_nib:  # this is to avoid dependency on nibabel
+            start_ornt = nib.orientations.io_orientation(affine)
+            target_ornt = nib.orientations.io_orientation(target_affine)
+            ornt_transform = nib.orientations.ornt_transform(start_ornt, target_ornt)
+            data_shape = data.shape
+            data = nib.orientations.apply_orientation(data, ornt_transform)
+            _affine = affine @ nib.orientations.inv_ornt_aff(ornt_transform, data_shape)
+            if np.allclose(_affine, target_affine, atol=AFFINE_TOL):
+                return data, ensure_mat44(_affine)
+
+        # need resampling
+        dtype = dtype or data.dtype  # type: ignore
+        if output_spatial_shape is None:
+            output_spatial_shape, _ = compute_shape_offset(data.shape, _affine, target_affine)
+        output_spatial_shape_ = list(output_spatial_shape) if output_spatial_shape is not None else []
+        sp_dims = min(data.ndim, 3)
+        output_spatial_shape_ += [1] * (sp_dims - len(output_spatial_shape_))
+        output_spatial_shape_ = output_spatial_shape_[:sp_dims]
+        original_channels = data.shape[3:]
+        if original_channels:  # multi channel, resampling each channel
+            data_np: np.ndarray = data.reshape(list(data.shape[:3]) + [-1])  # type: ignore
+            data_np = np.moveaxis(data_np, -1, 0)  # channel first for pytorch
+        else:  # single channel image, need to expand to have a channel
+            data_np = data[None]
+        affine_xform = AffineTransform(
+            normalized=False, mode=mode, padding_mode=padding_mode, align_corners=align_corners, reverse_indexing=True
+        )
+        data_torch = affine_xform(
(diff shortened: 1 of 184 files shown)


monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+                has been moved to the end). So if input is `(H,W,D,C)` and C==1, then it will be saved as `(H,W,D)`.
+                If D is also 1, it will be saved as `(H,W)`. If ``False``, image will always be saved as `(H,W,D,C)`.
+            spatial_ndim: modifying the spatial dims if needed, so that output to have at least
+                this number of spatial dims. If ``None``, the output will have the same number of
+                spatial dimensions as the input.
+            contiguous: if ``True``, the output will be contiguous.
+        """
+        # change data to "channel last" format
+        if channel_dim is not None:
+            # _chns = ensure_tuple_rep(channel_dim)
```
Nic-Ma, line 264: please remove the comment code.

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+        ``channel_dim`` is ``None``, a new axis will be appended to the last
+        dimension.
+
+        When ``squeeze_end_dims`` is ``True``, a postprocessing step will be
+        applied to remove any trailing singleton dimensions.
+
+        Args:
+            data: input data to be converted to "channel-last" format.
+            channel_dim: specifies the axis of the data array that is the channel dimension.
+                ``None`` indicates no channel dimension.
```
Nic-Ma, line 253: Could you please help describe the supported values? I remember it can be multiple channels dims now?
I checked below code, seems it only supports 1 channel dim: `while len(data.shape) < spatial_ndim + 1`.

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+            contiguous: if ``True``, the output will be contiguous.
+        """
+        # change data to "channel last" format
+        if channel_dim is not None:
+            # _chns = ensure_tuple_rep(channel_dim)
+            # data = moveaxis(data, _chns, tuple(range(-len(_chns), 0)))
+            data = moveaxis(data, channel_dim, -1)
+        else:  # adds a channel dimension
+            data = data[..., None]
+        # To ensure at least three spatial dims
```
Nic-Ma, line 269: Seems the code can support not 3 dims?

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+        data_np = data_torch[0].detach().cpu().numpy()
+        if original_channels:
+            data_np = np.moveaxis(data_np, 0, -1)  # channel last
+            data_np = data_np.reshape(list(data_np.shape[:3]) + list(original_channels))
+        else:
+            data_np = data_np[0]
+        return data_np, ensure_mat44(target_affine)
+
+    @classmethod
+    def convert_to_channel_last(
```
Nic-Ma, line 233: I feel it's better to change this logic to the transform `AsChannelLast`, then users can apply below 3 transforms in postprocessing:
```py
AffineResample()  # the resample logic in this PR
AsChannelLast()
SaveImage()
```
What do you think?
@ericspod @rijobro for more discussions.

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+
+        This method returns the following fields (the default value is ``None``):
+
+            - ``'original_affine'``: for data original affine (before any image processing),
+            - ``'affine'``: for the current data affine (representing the current coordinate information),
+            - ``'spatial_shape'``: for data original spatial shape.
+        """
+        if not metadata:
+            default_dict = {"original_affine": None, "affine": None, "spatial_shape": None}
+            metadata = default_dict
```
Nic-Ma, line 295: Can be simplified: `metadata = {"original_affine": None, "affine": None, "spatial_shape": None}`
And seems no need to set default with `None`, just `metadata = {}` is same?

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+    try to resample data from the space defined by `"affine"` to the space
+    defined by `"original_affine"`, for more details, please refer to the
+    ``resample_if_needed`` method.
+    """
+
+    def __init__(self, **kwargs):
+        """the constructor supports adding new instance members."""
+        self.data_obj = None
+        for k, v in kwargs.items():
+            setattr(self, k, v)
```
Nic-Ma, line 102: I didn't find the usage of these attributes.
And could you please help add doc-string to describe what's the expected values of this `kwargs` arg?

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+            default_dict = {"original_affine": None, "affine": None, "spatial_shape": None}
+            metadata = default_dict
+        original_affine = metadata.get("original_affine")
+        affine = metadata.get("affine")
+        spatial_shape = metadata.get("spatial_shape")
+        return original_affine, affine, spatial_shape
+
+
+@require_pkg(pkg_name="itk")
+class ITKWriter(ImageWriter):
```
Nic-Ma, line 303: Could you please help add doc-strings of **args** for all the functions in this class?

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+
+    def set_metadata(self, meta_dict: Optional[Mapping] = None, resample: bool = True, **options):
+        """Resample ``self.dataobj`` if needed.  This method assumes ``self.data_obj`` is a 'channel-last' ndarray."""
+        original_affine, affine, spatial_shape = self.get_meta_info(meta_dict)
+        self.data_obj, self.affine = self.resample_if_needed(
+            data_array=self.data_obj,
+            affine=affine,
+            target_affine=original_affine if resample else None,
+            output_spatial_shape=spatial_shape,
+            mode=options.pop("mode", GridSampleMode.BILINEAR),
```
Nic-Ma, line 330: I feel it's necessary to describe all the supported args in `options` in the doc-string.

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+    def __init__(self, output_dtype: DtypeLike = np.float32, **kwargs):
+        super().__init__(output_dtype=output_dtype, affine=None, channel_dim=0, **kwargs)
+
+    def set_data_array(self, data_array, channel_dim: Optional[int] = 0, squeeze_end_dims: bool = True, **kwargs):
+        """Convert ``data_array`` into 'channel-last' numpy ndarray."""
+        self.data_obj = self.convert_to_channel_last(
+            data=data_array,
+            channel_dim=channel_dim,
+            squeeze_end_dims=squeeze_end_dims,
+            spatial_ndim=kwargs.pop("spatial_ndim", 3),
```
Nic-Ma, line 317: I feel it's necessary to describe all the supported args in `kwargs` in the doc-string.

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+        channel_dim: Optional[int] = 0,
+        affine: Optional[NdarrayOrTensor] = None,
+        dtype: DtypeLike = np.float32,
+        **kwargs,
+    ):
+        """Create an ITK object from ``data_array``.  This method assumes a 'channel-last' ``data_array``."""
+        data_array = super().create_backend_obj(data_array)
+        _is_vec = channel_dim is not None
+        if _is_vec:
+            data_array = np.moveaxis(data_array, -1, 0)  # from channel last to channel first
```
Nic-Ma, line 362: I think here you only revert the order of spatial-dims for ITK, it's not consistent with our `ITKReader` logic for **single channel data**, which reverts all the dims:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/image_reader.py#L346
Could you please help make them consistent in this PR?

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+        data_array = data_array.T.astype(dtype, copy=True, order="C")
+        itk_obj = itk.GetImageFromArray(data_array, is_vector=_is_vec, ttype=kwargs.pop("ttype", None))
```
Nic-Ma, line 364: May I know the usage case of this `is_vector` arg here?
I don't quite understand the relationship between `is_vec` and ITK dim order vs numpy dim order.

Thanks in advance.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+
+
+@require_pkg(pkg_name="itk")
+class ITKWriter(ImageWriter):
+    """
+    Write data and metadata into files on disk using ITK-python.
+    """
+
+    def __init__(self, output_dtype: DtypeLike = np.float32, **kwargs):
+        super().__init__(output_dtype=output_dtype, affine=None, channel_dim=0, **kwargs)
```
Nic-Ma, line 309: Can you also put `affine` and `channel_dim`  in the `__init__` of ITKWriter?

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+
+    def set_metadata(self, meta_dict: Optional[Mapping], **options):
+        raise NotImplementedError(f"Subclasses of {self.__class__.__name__} must implement this method.")
+
+    def write(self, filename_or_obj: PathLike, verbose: bool = True, **kwargs):
+        """subclass should implement this method to call the backend-specific writing APIs."""
+        if verbose:
+            logger.info(f"writing: {filename_or_obj}")
+
+    @classmethod
```
Nic-Ma, line 115: May I know why using `classmethod` here? Maybe making it into a utility function directly?

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+        spacing = np.sqrt(np.sum(np.square(_affine[:d, :d]), 0))
+        spacing[spacing == 0] = 1.0
+        _direction: np.ndarray = np.diag(1 / spacing)
+        _direction = _affine[:d, :d] @ _direction
+        itk_obj.SetSpacing(spacing.tolist())
+        itk_obj.SetOrigin(_affine[:d, -1].tolist())
+        itk_obj.SetDirection(itk.GetMatrixFromArray(_direction))
+        return itk_obj
+
+    @staticmethod
```
Nic-Ma, line 381: May I know why using `classmethod` here? Maybe making it into a utility function directly?

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+            while len(data.shape) > spatial_ndim + 1:
+                data = data[..., 0, :]
+        # if desired, remove trailing singleton dimensions
+        while squeeze_end_dims and data.shape[-1] == 1:
+            data = np.squeeze(data, -1)
+        if contiguous:
+            data = ascontiguousarray(data)
+        return data
+
+    @classmethod
```
Nic-Ma, line 282: May I know why using `classmethod` here? Maybe making it into a utility function directly?

Thanks.

Nic-Ma (commented):
Thanks for the quick implementation.
The overall logic looks good to me, except for the `resample`, `as_channel_last` logic.
Put some minor comments inline.

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+
+        # need resampling
+        dtype = dtype or data.dtype  # type: ignore
+        if output_spatial_shape is None:
+            output_spatial_shape, _ = compute_shape_offset(data.shape, _affine, target_affine)
+        output_spatial_shape_ = list(output_spatial_shape) if output_spatial_shape is not None else []
+        sp_dims = min(data.ndim, 3)
+        output_spatial_shape_ += [1] * (sp_dims - len(output_spatial_shape_))
+        output_spatial_shape_ = output_spatial_shape_[:sp_dims]
+        original_channels = data.shape[3:]
```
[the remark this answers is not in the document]
wyli: for spatial dimensions, it supports up to 3D, and it could have multiple time/modality dimensions.

monai/data/image_writer.py
wyli: Yes, I forgot to reenable the multiple channel_dim here, will update in the latest

monai/data/image_writer.py
wyli: they are consistent here, it is in https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/image_reader.py#L348, there will be unit tests to check  `SaveImage`/`LoadImage` in the follow-up PRs.

monai/data/image_writer.py
wyli: this flag and the dtype will determine the underlying itk object type, for example
```py
>>> import numpy as np
>>> array = np.zeros((1,2,3))
>>> import itk
>>> itk.GetImageFromArray(array, is_vector=False)
<itk.itkImagePython.itkImageD3; proxy of <Swig Object of type 'itkImageD3 *' at 0x7f8b5aedb600> >
>>> itk.GetImageFromArray(array, is_vector=True)
<itk.itkVectorImagePython.itkVectorImageD2; proxy of <Swig Object of type 'itkVectorImageD2 *' at 0x7f8b68fad1e0> >
```
the same array could be interpreted as a 3d spatial image or a 2d vector image.

monai/data/image_writer.py
wyli: the PIL writer would override this class as affine is not needed in that case

monai/data/image_writer.py
wyli: subclasses can extend this method for different backend, and they share some common logic such as casting torch tensor to ndarray.

monai/data/image_writer.py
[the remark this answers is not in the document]
Nic-Ma: I mean `[3:]` can't get the channel dim of 2D data, maybe use `sp_dims` instead?

Thanks.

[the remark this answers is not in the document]
wyli: maybe the docstring answers your question, it follows the nifti format as described in the docstring https://github.com/Project-MONAI/MONAI/blob/20d0bf93b348ef25f8dad066869233302d8e293d/monai/data/image_writer.py#L151-L159 

I'll try to make it a generic function/transform, it'll be useful for the metadata handling.

monai/data/image_writer.py
wyli: The current design is the other way around: `SaveImage` as a transform should work well with the channel-first format, so that we hide the backend implementation details. For example, we can `SaveImage` to 'dump' intermediate transform outcomes conveniently. `SaveImage` internally calls resample/channel_last steps based on the requirements of the writer backend, not all backend consume the same channel last format.

Nic-Ma: Hi @wyli ,

Thanks for your explanation, I see your design point.
But I am not very sure which way is better to go, as one of our MONAI design principles is "easy to integrate", maybe we should keep every component simple and independent? Daguang also raised the idea to extract `resample` from `SaveImage` during Clara 4.0 development.
@ericspod @rijobro Could you guys please help share your opinions?

Thanks in advance.

monai/data/image_writer.py
Nic-Ma: I mean for **single channel** data, would it be `img.GetNumberOfComponentsPerPixel() == 1`? So the logic will be different from here.

Thanks.

monai/data/image_writer.py
Nic-Ma: OK, I see. Seems you don't want to implement it as `static` function and it also works with other class functions.

Thanks.

monai/data/image_writer.py
wyli: I'm trying to create a `SpatialResample` transform, and `SaveImage` will internally call this transform if the user runs `SaveImage(resample=True)`， the user would also have the `SpatialResample` transform. I don't see why `SaveImage` shouldn't integrate the resampling functionality.

monai/data/image_writer.py
Nic-Ma: I still don't understand this line: `_is_vec = channel_dim is not None`, why if `channel_dim` is not None, then the data is **2d vector image**?

Thanks.

wyli: in the example `np.zeros((1,2,3))` could be saved as a 2d image with channels (e.g. `channel_dim=-1` for channel-last) or a 3d image without any channel (`channel_dim=None`).

monai/data/image_writer.py
Nic-Ma: Hi @wyli ,

That makes sense to me, if also providing `SpatialResample` transform, it's OK to call it in `SaveImage`.
And so this `resample` logic would be in the `SpatialResample` transform instead of this `ImageWriter` module, right?

Thanks.

monai/data/image_writer.py
Nic-Ma: But if the data is 4d like `[128, 128, 128, 3]` and `channel_dim` is `-1`, it will be treated as `_is_vec = True`?

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+    @classmethod
+    def create_backend_obj(cls, data_array: NdarrayOrTensor, **kwargs) -> np.ndarray:
+        """
+        Subclass should implement this method to return a backend-specific data representation object.
+        This method is used by ``cls.write`` and the input ``data_array`` is assumed 'channel-last'.
+        """
+        return convert_data_type(data_array, np.ndarray)[0]  # type: ignore
+
+    @classmethod
+    def resample_if_needed(
```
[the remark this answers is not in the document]
ericspod: It's probably better to separate functionality so that definitions do only one thing at a time unless there's good reason to do multiples. I guess that's what the feedback was saying?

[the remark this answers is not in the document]
wyli: thanks this is addressed in https://github.com/Project-MONAI/MONAI/pull/3701.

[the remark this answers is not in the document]
Nic-Ma: Cool! thanks for the quick update!

monai/data/utils.py
```diff
@@ -547,6 +548,30 @@ def set_rnd(obj, seed: int) -> int:
     return seed
 
 
+def affine_to_spacing(affine: NdarrayTensor, dims: int = 3, dtype=float, suppress_zeros: bool = True) -> NdarrayTensor:
```
Nic-Ma, line 551: Hi @wyli ,

I think this is the function used in: https://github.com/Project-MONAI/MONAI/issues/3766, right?

Thanks.

wyli: Yes, this commit  is to address it,  commit message links to it https://github.com/wyli/MONAI/commit/b32336e596009127c46c813149296fa835697471

monai/data/utils.py
```diff
@@ -547,6 +548,30 @@ def set_rnd(obj, seed: int) -> int:
...
+    """
+    Computing the current spacing from the affine matrix.
+
+    Args:
+        affine: a d x d affine matrix.
+        dims: indexing of the spatial dimensions `affine[:dims, :dims]`.
```
Nic-Ma, line 557: Maybe we should use arg name `ndim` to be consistent with others?

Thanks.

monai/data/utils.py
```diff
@@ -60,6 +60,7 @@
     "list_data_collate",
     "worker_init_fn",
     "set_rnd",
+    "affine_to_spacing",
```
Nic-Ma, line 63: Please also add new function `orientation_ras_lps` here.

Thanks.

Nic-Ma (approved):
Thanks for the quick update.
Looks good to me now, except for several minor comments inline.
@ericspod @rijobro Do you have any other comments?

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,362 @@
...
+            channel_dim: specifies the channel axes of the data array to move to the last.
+                ``None`` indicates no channel dimension, a new axis will be appended as the channel dimension.
+                a sequence of integers indicates multiple non-spatial dimensions.
+            squeeze_end_dims: if ``True``, any trailing singleton dimensions will be removed (after the channel
+                has been moved to the end). So if input is `(H,W,D,C)` and C==1, then it will be saved as `(H,W,D)`.
+                If D is also 1, it will be saved as `(H,W)`. If ``False``, image will always be saved as `(H,W,D,C)`.
+            spatial_ndim: modifying the spatial dims if needed, so that output to have at least
+                this number of spatial dims. If ``None``, the output will have the same number of
+                spatial dimensions as the input.
+            contiguous: if ``True``, the output will be contiguous.
```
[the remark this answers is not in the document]
wyli: it'll cause a copying operation, it's mostly not in use.

monai/data/image_writer.py
```diff
@@ -0,0 +1,362 @@
...
+            data_array: input data array.
+            channel_dim: channel dimension of the data array. This is used to create a Vector Image if it is not ``None``.
+            affine: affine matrix of the data array. This is used to compute `spacing`, `direction` and `origin`.
+            dtype: output data type.
+            kwargs: keyword arguments. Current `itk.GetImageFromArray` will read ``ttype`` from this dictionary.
+        """
+        data_array = super().create_backend_obj(data_array)
+        _is_vec = channel_dim is not None
+        if _is_vec:
+            data_array = np.moveaxis(data_array, -1, 0)  # from channel last to channel first
```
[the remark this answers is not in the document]
wyli: not really, after `set_data_array` the `data_array` here is channel-last.

monai/data/utils.py
wyli: it's not actually ndim, it could be `-1` here for the indexing.

monai/data/image_writer.py
wyli: After your initial comment, I've updated the docstring:
```
The constructor supports adding new instance members.
The current member in the base class is ``self.data_obj``, the subclasses can add more members,
so that necessary meta information can be stored in the object and shared among the class methods.
```

wyli:
> Thanks for the quick update. Looks good to me now, except for several minor comments inline.

thanks Nic, I updated the PR for this use case of writing with a different spatial size:
```py
# write as 2d image, channel-first
np_data = np.arange(48).reshape(3, 4, 4)
writer = ITKWriter(output_dtype=np.uint8)
writer.set_data_array(np_data, channel_dim=0)
writer.set_metadata({"spatial_shape": (5, 5)})
writer.write("test1.png")
```

could you review again to confirm?
also when writing the examples I found that
```py
writer = ITKWriter(channel_dim=0)
writer.set_data_array(..., channel_dim=0)
```
will cause confusion, so I removed the constructor's `channel_dim`.

monai/data/utils.py
Nic-Ma: If so, I would suggest to change another arg name, `dims` looks like a list of something, confusing.

Thanks.

monai/data/image_writer.py
```diff
@@ -0,0 +1,395 @@
...
+    def set_data_array(self, data_array, **kwargs):
+        raise NotImplementedError(f"Subclasses of {self.__class__.__name__} must implement this method.")
+
+    def set_metadata(self, meta_dict: Optional[Mapping], **options):
+        raise NotImplementedError(f"Subclasses of {self.__class__.__name__} must implement this method.")
+
+    def write(self, filename: PathLike, verbose: bool = True, **kwargs):
+        """subclass should implement this method to call the backend-specific writing APIs."""
+        if verbose:
+            logger.info(f"writing: {filename}")
```
Nic-Ma, line 106: Seems the `logger` is not defined? Could you please help confirm it?

Thanks.

Nic-Ma (approved):
Looks good to me except for 2 minor comments inline.

Thanks.

monai/data/utils.py
wyli: sure updated to `r`.

monai/data/image_writer.py
wyli: it's a module-level variable, now it's added to `__all__`

PR #3676, opened 2022-01-18
3672 Add more messages for AUC warning

### LINKED ISSUE #3672: Change exception to warning in AUC metric
Currently, when the input data is all 0 or all 1, AUC will raise an error:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/metrics/rocauc.py#L69

@SachidanandAlle , @madil90 , and our internal QAs all encountered this error in classification tasks before, especially when **verifying the pipeline with very few samples**. For example, it's easy to trigger the exception in chest X-ray classification task as the task has 15 classes, easy to get all 0 or all 1 samples for a class.

I would like to change the exception to warning and skip the AUC computation for this case, or add a flag to control whether a warning or exception?
@ericspod @wyli @rijobro Do you guys have any other concerns?

Thanks in advance.

Nic-Ma:
Fixes #3672 .

### Description
This PR enhanced the warning messages according to @ericspod 's comment https://github.com/Project-MONAI/MONAI/issues/3672#issuecomment-1015532297.

### Status
**Ready**

### Types of changes


ericspod (approved):

monai/metrics/rocauc.py
```diff
@@ -66,9 +66,17 @@ def aggregate(self):  # type: ignore
...
-        warnings.warn("y values must be 0 or 1, can not be all 0 or all 1, skip AUC computation and return `Nan`.")
+    y_unique = y.unique()
+    if y_unique.equal(torch.tensor([0], dtype=y.dtype, device=y.device)):
+        warnings.warn("y values can not be all 0, skip AUC computation and return `Nan`.")
         return float("nan")
+    if y_unique.equal(torch.tensor([1], dtype=y.dtype, device=y.device)):
+        warnings.warn("y values can not be all 1, skip AUC computation and return `Nan`.")
+        return float("nan")
+    if not y_unique.equal(torch.tensor([0, 1], dtype=y.dtype, device=y.device)):
+        warnings.warn("y values must be 0 or 1, skip AUC computation and return `Nan`.")
```
wyli, line 77: perhaps just print `y_unique` in the message instead of 3 comparisons?

Nic-Ma: Hi @wyli ,

Thanks for your suggestion, the problem is that: if users forgot to convert labels to [0, 1], it may have many many unique values.
How about change it to 2 prints: (1) all 0 or all 1, (2) not only [0, 1] ?

BTW, I think the sklearn solution you shared also have 3 separated errors for this check:
https://github.com/scikit-learn/scikit-learn/blob/7e1e6d09bcc2eaeba98f7e737aac2ac782f0e5f1/sklearn/metrics/_ranking.py#L991
https://github.com/scikit-learn/scikit-learn/blob/7e1e6d09bcc2eaeba98f7e737aac2ac782f0e5f1/sklearn/metrics/_ranking.py#L1000
https://github.com/scikit-learn/scikit-learn/blob/7e1e6d09bcc2eaeba98f7e737aac2ac782f0e5f1/sklearn/metrics/_ranking.py#L549

Thanks.

wyli: I think if `y_unique` has many elements we can only print first a few?

Nic-Ma: Hi @wyli ,

I simplified the logic in the latest commit, could you please help review it again?
Some test result:
```
y values can not be all 1.0, skip AUC computation and return `Nan`.
y values can not be all 0.0, skip AUC computation and return `Nan`.
..y values must be 0 or 1, but in [0.0, 1.0, 2.0, 3.0], skip AUC computation and return `Nan`.
```

Thanks in advance.

PR #3687, opened 2022-01-20
remove tensor-array conversion in `Orientation`

wyli:

### Description
non-breaking update to avoid tensor <-> ndarray conversion for the input image in `Orientation` transform

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -330,7 +330,7 @@ class Orientation(Transform):
             data_array.permute(full_transpose.tolist()) if _is_tensor else data_array.transpose(full_transpose)  # type: ignore
         )
         new_affine = to_affine_nd(affine_np, new_affine)
-        out = data_array.contiguous() if _is_tensor else np.ascontiguousarray(data_array)
+        out = data_array.contiguous() if _is_tensor else np.ascontiguousarray(data_array)  # type: ignore
         new_affine, *_ = convert_to_dst_type(src=new_affine, dst=affine, dtype=torch.float32)
 
         if self.image_only:
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -322,10 +322,12 @@ class Orientation(Transform):
         spatial_ornt[:, 0] += 1  # skip channel dim
         spatial_ornt = np.concatenate([np.array([[0, 1]]), spatial_ornt])
         axes = [ax for ax, flip in enumerate(spatial_ornt[:, 1]) if flip == -1]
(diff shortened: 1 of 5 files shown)


tests/test_orientation.py
```diff
@@ -110,14 +169,15 @@
...
+    def test_ornt(self, in_type, init_param, img, data_param, expected_data, expected_code):
+        img = in_type(img)
         ornt = Orientation(**init_param)
         res = ornt(img, **data_param)
         if not isinstance(res, tuple):
-            np.testing.assert_allclose(res, expected_data)
+            assert_allclose(res, expected_data, type_test=False)
             return
-        np.testing.assert_allclose(res[0], expected_data)
+        assert_allclose(res[0], expected_data, type_test=False)
```
rijobro, line 180: if the tests are created such with `p(expected_data)`, then we can remove the `type_test`, right?

rijobro (commented):
looks good to me, thanks!

monai/transforms/spatial/array.py
```diff
@@ -310,21 +310,27 @@ def __call__(
...
+        if axes:
+            data_array = torch.flip(data_array, dims=axes) if _is_tensor else np.flip(data_array, axis=axes)  # type: ignore
+        full_transpose = np.arange(len(data_array.shape))
+        full_transpose[: len(spatial_ornt)] = np.argsort(spatial_ornt[:, 0])
+        data_array = (
+            data_array.permute(full_transpose.tolist()) if _is_tensor else data_array.transpose(full_transpose)  # type: ignore
+        )
         new_affine = to_affine_nd(affine_np, new_affine)
-        out, *_ = convert_to_dst_type(src=data_array_np, dst=data_array)
+        out = data_array.contiguous() if _is_tensor else np.ascontiguousarray(data_array)  # type: ignore
```
Nic-Ma, line 333: I think you can simplify it to:
```py
out, *_ = convert_to_dst_type(src=data_array, dst=data_array)
```
The `ToTensor` and `ToNumpy` logic already have the `contiguous` conversion:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/utils/type_conversion.py#L173
https://github.com/Project-MONAI/MONAI/blob/dev/monai/utils/type_conversion.py#L120

Thanks.

Nic-Ma (approved):
Looks good to me except for minor comment.

Thanks.

monai/transforms/spatial/array.py
wyli: thanks, I updated the PR and added logic to skip the transpose if not needed

Nic-Ma:
Thanks for the quick update.
I think it looks good to me.
I am not very familiar with medical orientation computation, @ericspod @rijobro could you please help double confirm?

Thanks in advance.

PR #3690, opened 2022-01-20
3686 Skip workflow run if data is empty or the specified epoch_length is 0

### LINKED ISSUE #3686: Skip workflow run if no data provided
Feature request from @SachidanandAlle :
Earlier we were not able to support zero input for validation, when there is no items for validation, we like to skip the evaluator.
can we add that fix if it's not there? which is needed to run train/finetune over few images, because some-times user submits only few samples for (re)training, and won't provide samples for evaluation/validation step.

Nic-Ma:
Fixes #3686 .

### Description
This PR enhanced the workflow to skip run if data is empty or the specified epoch_length is 0.

### Status
**Ready**

### Types of changes


ericspod (approved):

SachidanandAlle (approved):
LGTM.. should be good if we have the test case for multi-gpu as well

Nic-Ma:
Hi @ericspod @SachidanandAlle ,

Thanks for the review.
I spent much time today to add support for multi-gpu training case, I tried many methods, but no ideal method can guarantee our distributed communication logic works fine with less ranks that the setting, it's PyTorch or NCCL logic, we can't dynamically add or reduce ranks. For example, some handlers or metrics need all ranks to run the same logic to `all-gather` the result, otherwise, it will hang.
I already added unit tests for multi-gpu training and added more message in the warning for the hanging case.

Thanks.

PR #3701, opened 2022-01-23
3697 add spatial resample

### LINKED ISSUE #3697: decouple the resampling step from image writer
We got some feedback from NVIDIA Clara users about the `resample` operation in our saver, would it be better to put the `resample` logic in a separate module? Maybe a single `transform`, then users can apply it in postprocessing transforms or preprocessing transforms

_Originally posted by @Nic-Ma in https://github.com/Project-MONAI/MONAI/pull/3674#discussion_r788306760_

wyli:
Fixes #3697

### Description
adds a resample transform based on the affine matrices
this is provided with an option using monai's csrc backend

### Status
**ready**

### Types of changes


wyli:
this is non-breaking. I'll add more tests after addressing any major comments @Project-MONAI/core-reviewers

Nic-Ma:
Hi @wyli ,

Thanks for the quick implementation.
I think the overall design is aligned with our discussion in the ImageWriter PR.
@ericspod Do you have any other concerns to go this way?

Thanks in advance.

monai/transforms/spatial/array.py
```diff
@@ -82,12 +85,167 @@
...
+            img_ = img_.reshape(xform_shape)
+        if align_corners:
+            _t_r = torch.diag(torch.ones(len(xform), dtype=xform.dtype, device=xform.device))  # type: ignore
+            for idx, d_dst in enumerate(spatial_size[:spatial_rank]):
+                _t_r[idx, -1] = (max(d_dst, 2) - 1.0) / 2.0
+            xform = xform @ _t_r
+            if not USE_COMPILED:
+                _t_l = normalize_transform(in_spatial_size, xform.device, xform.dtype, align_corners=True)  # type: ignore
+                xform = _t_l @ xform
+            affine_xform = Affine(
```
ericspod, line 222: Object creation is going to be a bit expensive, though this object doesn't create large data segments, so trying to use this transform quickly for a large number of smallish images is going to suffer for a constant runtime cost.

ericspod:
One issue is that the purpose of the various transforms is getting less clear, if I want to do an affine transform on an image do I use `Affine`, `SpatialResample`, `AffineGrid` (the docstring for this isn't clear about it's purpose though the arguments for `__call__` show it won't work on images), or `Resample`. `SpatialResample` uses `Affine` and `AffineTransform` so it's clearer looking at the code to use this, but then why have `Affine`? 



A more general issue that I encountered with `Rand*DElastic` is that the coordinate space for `grid_pull` (ours) and `grid_sample` (Pytorch) are different and so conversion has to be done in `Resample` [here](https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/spatial/array.py#L1571) and below. I think in my profiling this was a significant cause of slowness with `Rand*DElastic` that was avoided in the way `RandSmoothDeform` created it's coord field with meshgrid and used the smooth field. I realize there are functional differences between `RandSmoothDeform` and `Rand*DElastic` but eliminating the need to manipulate coordinate fields, and keeping as much object creation in the constructor as possible, improved performance significantly with visually similar results. `SpatialResample` will have the same issue of performance since it ultimately uses `Resample`. It's a bit out of scope here but I had to mention it somewhere!

monai/transforms/spatial/array.py
wyli: agree with the idea in general, but in this case the `affine` has to be provided at the constructor of `Affine`, because the grid will be created within `__call__`.

wyli:
Sure for the first concern I'm revising the docstring to clarify the usage.

For the `Resampling` coordinates issue, 
- pytorch uses [-1, 1] and axes (k, j, i), channel first
- ours csrc backend uses [0, size-1] and axes [i, j, k], channel last
- for the grid computations in the data augmentations it seems natural to use  [-(size-1)/2, (size-1)/2] (for example we often rotate an image around its centre instead of `[0,0]`). 

These describe the same thing with different conventions, should be converted properly somewhere before calling the backend resampling API. I added a `norm_coords: bool`  in `Resample` to make the normalisation optional, it should speedup things a little bit.
in the future we may define a default convention and further optimise it...

monai/transforms/spatial/array.py
```diff
@@ -82,12 +86,176 @@
...
+        self.mode = mode
+        self.padding_mode = padding_mode
+        self.align_corners = align_corners
+        self.dtype = dtype
+
+    def __call__(
+        self,
+        img: NdarrayOrTensor,
+        src: Optional[NdarrayOrTensor] = None,
+        dst: Optional[NdarrayOrTensor] = None,
```
Nic-Ma, line 133: Maybe `src_affine` and `dst_affine` are more clear?

Thanks.

monai/transforms/spatial/array.py
```diff
@@ -82,12 +86,176 @@
...
+                See also: https://pytorch.org/docs/stable/generated/torch.nn.functional.grid_sample.html
+                When `USE_COMPILED` is `True`, this argument uses
+                ``"nearest"``, ``"bilinear"``, ``"bicubic"`` to indicate 0, 1, 3 order interpolations.
+                See also: https://docs.monai.io/en/stable/networks.html#grid-pull
+            padding_mode: {``"zeros"``, ``"border"``, ``"reflection"``}
+                Padding mode for outside grid values. Defaults to ``"border"``.
+                See also: https://pytorch.org/docs/stable/generated/torch.nn.functional.grid_sample.html
+            align_corners: Geometrically, we consider the pixels of the input as squares rather than points.
+                See also: https://pytorch.org/docs/stable/generated/torch.nn.functional.grid_sample.html
+            dtype: data type for resampling computation. Defaults to ``np.float64`` for best precision.
```
Nic-Ma, line 165: It's: default to `self.dtype` or `np.float64`.

Thanks.

Nic-Ma (commented):
Thanks for the update, I reviewed `SpatialResample` transform, will continue to review tonight.

monai/transforms/spatial/array.py
```diff
@@ -1449,8 +1644,10 @@ def __init__(
         spatial_size: Optional[Union[Sequence[int], int]] = None,
         mode: Union[GridSampleMode, str] = GridSampleMode.BILINEAR,
         padding_mode: Union[GridSamplePadMode, str] = GridSamplePadMode.REFLECTION,
+        norm_coords: bool = True,
         as_tensor_output: bool = True,
         device: Optional[torch.device] = None,
+        dtype: DtypeLike = np.float32,
```
Nic-Ma, line 1650: Should it be `np.float64` or `np.float32` here?

Thanks.

monai/transforms/spatial/dictionary.py
```diff
@@ -607,6 +765,7 @@ def __init__(
         padding_mode: GridSamplePadModeSequence = GridSamplePadMode.REFLECTION,
         as_tensor_output: bool = True,
         device: Optional[torch.device] = None,
+        dtype: Union[DtypeLike, torch.dtype] = np.float32,
```
Nic-Ma, line 768: Seems the default value doesn't match doc-string.

Thanks.

Nic-Ma (commented):
Hi @wyli ,

Thanks for your quick update.
Could you please help confirm and unify all the default value of `dtype` arg?

Thanks.

monai/transforms/spatial/dictionary.py
wyli: thanks that's a good point, I reviewed the dtype options, they should be good now. float32 is good as a data augmentation default...

monai/transforms/spatial/array.py
```diff
@@ -1485,10 +1680,21 @@ def __init__(
...
             padding_mode: {``"zeros"``, ``"border"``, ``"reflection"``}
                 Padding mode for outside grid values. Defaults to ``"reflection"``.
                 See also: https://pytorch.org/docs/stable/generated/torch.nn.functional.grid_sample.html
+            norm_coords: whether to normalize the coordinates from `[-(size-1)/2, (size-1)/2]` to
+                `[0, size - 1]` or `[-1, 1]` to be compatible with the underlying resampling API.
+                If the coordinates are generated by ``monai.transforms.utils.create_grid``
+                and the ``affine`` doesn't include the normalization, this argument should be set to ``True``.
+                If the output `self.affine_grid` is already normalized, this argument should be set to ``False``.
             device: device on which the tensor will be allocated.
+            dtype: data type for resampling computation. Defaults to ``np.float32``.
```
Nic-Ma, line 1695: May I know why the `dtype` of resampling computation is `float32` and not consistent with `SpatialResample`?

Thanks.

monai/transforms/spatial/array.py
```diff
@@ -1085,6 +1255,9 @@ class AffineGrid(Transform):
             pixel/voxel relative to the center of the input image. Defaults to no translation.
         scale_params: scale factor for every spatial dims. a tuple of 2 floats for 2D,
             a tuple of 3 floats for 3D. Defaults to `1.0`.
+        dtype: data type for the grid computation. Defaults to ``np.float32``.
```
Nic-Ma, line 1258: I want to double confirm here the `grid computation` is with `float32`, other `resampling computation` is `float64`?

Thanks.

monai/transforms/spatial/dictionary.py
```diff
@@ -646,6 +805,9 @@ def __init__(
                 See also: https://pytorch.org/docs/stable/generated/torch.nn.functional.grid_sample.html
                 It also can be a sequence of string, each element corresponds to a key in ``keys``.
             device: device on which the tensor will be allocated.
+            dtype: data type for resampling computation. Defaults to ``np.float32``.
```
Nic-Ma, line 808: May I know why the `dtype` of resampling computation is `float32` and not consistent with `SpatialResampled`?

Thanks.

Nic-Ma:
Thanks for the quick update, this PR overall looks good to me now, put minor comments inline.
I didn't check carefully about the `affine` computation and C++ resampling logic as I am not very familiar with that.
@ericspod @rijobro Could you please help double confirm this PR as you put some comments before?
If you guys don't have any other comments or concerns, I will approve this big PR soon.

Thanks in advance.

monai/transforms/spatial/dictionary.py
wyli: the primary use case is random data augmentation, float32 is good enough with the speed and memory and compatibility

monai/transforms/spatial/array.py
wyli: yes I think it's good, these are also non-breaking changes

tests/test_integration_fast_train.py
```diff
@@ -123,6 +123,7 @@ def test_train_timing(self):
                     range_x=np.pi / 4,
                     mode=("bilinear", "nearest"),
                     align_corners=True,
+                    dtype=np.float64,
```
wyli, line 126: @Nic-Ma  if we use np.float32 here might speed up it a bit...

Nic-Ma: OK, let's mark it here for future performance tuning.

Thanks.

monai/transforms/spatial/dictionary.py
```diff
@@ -131,6 +135,160 @@
...
+        self,
+        keys: KeysCollection,
+        mode: GridSampleModeSequence = GridSampleMode.BILINEAR,
+        padding_mode: GridSamplePadModeSequence = GridSamplePadMode.BORDER,
+        align_corners: Union[Sequence[bool], bool] = False,
+        dtype: Optional[Union[Sequence[DtypeLike], DtypeLike]] = np.float64,
+        meta_keys: Optional[KeysCollection] = None,
+        meta_key_postfix: str = DEFAULT_POST_FIX,
+        meta_src_keys: Optional[KeysCollection] = "src_affine",
+        meta_dst_keys: Optional[KeysCollection] = "dst_affine",
```
Nic-Ma, line 165: Should we default the keys to `affine` and `original_affine` and describe the common use case in doc-string?
I think for most of the cases, we use this transform to resample image back to the original affine.

Thanks.

Nic-Ma (approved):
Looks good to me now.
@ericspod @rijobro do you guys have any other concerns or comments?

Thanks.

wyli:
Integration test is ok https://github.com/Project-MONAI/MONAI/runs/5044690171?check_suite_focus=true

monai/transforms/spatial/dictionary.py
wyli: that is one use case, but I think it is a generic spatial transform and the concepts of 'affine' 'original_affine' are specifically part of the IO in our modules

Nic-Ma: I am OK with both solutions, @ericspod @rijobro what do you guys think?

Thanks.

PR #3711, opened 2022-01-25
3710 Add support to set delimiter for CSV files and change default to comma

### LINKED ISSUE #3710: Optional split mark for CSV files
Feature request from @holgerroth :  the CSV output of in the metrics summary reports are split by `tap` instead of comma.
Would be nice to change the default to comma to be consistent with `CSVSaver` and better to add support to set the split mark for CSV writing logic.

Nic-Ma:
Fixes #3710 .

### Description
According to @holgerroth 's feedback, this PR added support to set `delimiter` for `CSVSaver`, `ClassificationSaver`, `SaveClassificationd`.
And changed the default value to comma in `MetricsWriter` handler.

### Status
**Ready**

### Types of changes


Nic-Ma:
Added unit tests to cover the `delimiter` arg in all the components.

Thanks.

holgerroth (approved):
LGTM

monai/data/csv_saver.py
```diff
@@ -48,6 +49,8 @@ def __init__(
                 otherwise, will append new content to the CSV file.
             flush: whether to write the cache data to CSV file immediately when `save_batch` and clear the cache.
                 default to False.
+            delimiter: the delimiter character in the saved file, default to ",", for more details:
+                https://docs.python.org/3/library/csv.html#csv.Dialect.delimiter.
```
wyli, line 53: any reason not using the build-in python csv lib here? the url is from python csv may be it'll cause confusion because the implementation doesn't use it...

wyli (approved):

monai/data/csv_saver.py
Nic-Ma: Hi @wyli , I don't quite remember the initial reason, seems we suppose to support not only `.csv` format?
Because actually users can set `filename` with other ext here.
Let me try to explain more in the doc-string?

Thanks.

Nic-Ma: Hi @wyli ,

I updated the PR in the latest commit.

Thanks in advance.

wyli: Thanks, it looks good to me.

Issue #3716, opened 2022-01-26
3D Densenet does not support deterministic training in GPU mode


yiheng-wang-nv:
@holgerroth and I found that when we use GPU devices, using 3D DenseNet cannot do deterministic training.

I prepared a simple code block that can help to reproduce this issue:

```
import torch
import monai
from monai.utils import set_determinism
import os

# os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':4096:8'
# torch.use_deterministic_algorithms(True)
set_determinism(seed=0)

loss_function = torch.nn.CrossEntropyLoss()
device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
# device = "cpu"

spatial_dims = 3
model = monai.networks.nets.DenseNet121(
    spatial_dims=spatial_dims, in_channels=1, out_channels=2).to(device)
optimizer = torch.optim.Adam(model.parameters(), 1e-5)
model.train()

lb = torch.tensor([1, 0]).to(device)
all_loss = 0
shape = [2, 1, 64, 64, 64][:2+spatial_dims]
for i in range(10):

    inp = torch.randn(shape).to(device)
    optimizer.zero_grad()
    out = model(inp)
    loss = loss_function(out, lb)
    loss.backward()
    optimizer.step()
    
    all_loss += loss.item()
print(all_loss)
```
When device is gpu and `spatial_dims = 3`, running the above code twice will achieve different results.
In addition, when uncomment two lines before `set_determinism(seed=0)`, you will see the related error, which shows that the 3D adaptive average pooling layer is non-deterministic when using GPU.

wyli:
clarified in the docstring #3717

csheaff:
@wyli @yiheng-wang-nv 

Resnet also seems to have this issue:

> max_pool3d_with_indices_backward_cuda does not have a deterministic implementation, but you set 'torch.
> _algorithms(True, warn_only=True)'. You can file an issue at https://github.com/pytorch/pytorch/issues 
> tize adding deterministic support for this operation. (Triggered internally at /opt/pytorch/pytorch/ate
> t.cpp:71.)

I am trying identify a good classification model that is deterministic with Monai but cannot. EfficientNet is but doesn't learn very well. Any suggestions?

PR #3717, opened 2022-01-26
3716 enhance doc-strings of DenseNet

yiheng-wang-nv:

Related to #3716  .

### Description
In this PR, I enhanced the doc-strings of DenseNet to explain its non-determinism.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/networks/nets/densenet.py
+++ b/monai/networks/nets/densenet.py
@@ -148,7 +148,7 @@ class DenseNet(nn.Module):
     """
     Densenet based on: `Densely Connected Convolutional Networks <https://arxiv.org/pdf/1608.06993.pdf>`_.
     Adapted from PyTorch Hub 2D version: https://pytorch.org/vision/stable/models.html#id16.
-    This network is non-determistic When `in_channels` is 3 and CUDA is enabled. Please check the link below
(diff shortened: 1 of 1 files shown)


monai/networks/nets/densenet.py
```diff
@@ -148,6 +148,9 @@ class DenseNet(nn.Module):
     """
     Densenet based on: `Densely Connected Convolutional Networks <https://arxiv.org/pdf/1608.06993.pdf>`_.
     Adapted from PyTorch Hub 2D version: https://pytorch.org/vision/stable/models.html#id16.
+    This network is non-determistic When `spatial_dims` is 3 and CUDA is enabled. Please check the link below
```
wyli, line 151: I think it'll only affect the training (with backprop), the forward part is deterministic could you help double check? Thanks!

wyli (approved):

monai/networks/nets/densenet.py
yiheng-wang-nv: Thanks @wyli , yes, if only do forward part, it is not affected. Should it be mentioned in the doc-strings?

wyli: thank you, I think the docstring is fine...

PR #3593, opened 2022-01-06
[WIP] 3482 Develop components for the model package MVP

Nic-Ma:
Fixes #3482 .

### Description
Develop the complete prototype of model package in this branch as @wyli suggested: https://github.com/Project-MONAI/MONAI/pull/3518#issuecomment-998600114.
Then split into smaller PRs for review.

### Status
**Work in progress**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick --unittests  --disttests`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

Issue #3482, opened 2021-12-13
Develop MVP of model package


Nic-Ma:
**Is your feature request related to a problem? Please describe.**
Thanks for the interesting technical discussion with @wyli @atbenmurray @rijobro , we plan to develop the model package feature step by step. For the initial step, I will try to implement a MVP, referring to some existing solutions, like NVIDIA Clara MMAR, ignite online package, etc. Draft task steps:

- [ ] Search specified python packages and build python instances from dictionary config with `name` / `path` & `args`.
- [ ] Recursively parse configs in a dictionary with dependencies and executable code, for example:
       `{"dataset": {"args": {"data": "#load_decathlon_datalist()"}}, "dataloader": "args": {"data": "@dataset"}}`.
- [ ] Add support to load config from a JSON file, run program from shell commands.
- [ ] Add support to save the raw config dictionaries into TorchScript model.

dbericat:
@cpbridge this is the current PR to expand model metadata.

Nic-Ma:
Hi @wyli @ericspod @rijobro @atbenmurray ,

I updated the ticket description to make it more clear.

Thanks.

wyli:
We've discussed that the MVP should focus on the minimal structure, including model weights, basic environment/system info, maintaining changelog/versioning, how to ensure accessibility for both human and machine.

Nic-Ma:
Hi @wyli ,

Sure, I added these items to the MVP task list.

Thanks.

dbericat:
@MMelQin @gigony @GreySeaWolf @cpbridge @mocsharp @joshliberty @vikashg thoughts on proposed MVP?

ericspod:
We should discuss a clear definition of requirements and objectives. We want to define a format of a single file or multiples files which contains the model weights at least with secondary information describing how to use it for various use cases. This will allow a human or a program to determine what sort of model it is, how to use the model, and what tasks to use it for. For our MVP we want to consider the starting position of what the model weight storage and metadata storage would look like and if it would achieve that objective to some degree.

The base level set of requirements I would suggest are:
* Ability to save model weights
* Ability to save runnable model in Torchscript and/or ONNX
* Include information on model name, version, description, authorship, intended use
* Include package and technical information needed to use model, eg. MONAI version, Pytorch and other package versions
* Description of input and output, eg. how to interpret predictions returned
* Description in some form of how to transform input data into the form expected by the model (ie. preprocess transform sequence) and transform for output 

[...]

PR #3720, opened 2022-01-26
3482 Add `ConfigComponent` for config parsing

Nic-Ma:
Task step 2 of #3482 .

### Description
This PR implemented the `ConfigComponent` feature for config parsing, it's for task 2 of #3482 .
The whole proposal is in draft PR: https://github.com/Project-MONAI/MONAI/pull/3593.

### Status
**Ready**

### Types of changes


Nic-Ma:
I tried to cover all the possible config cases in the unit tests.

Thanks.

wyli (dismissed):
(Still under discussion)

wyli:
I'm still looking into the existing implementations outside of monai...

for this pr,
- at the component config level looks like it is decoupled from the format-specific elements of json and yaml. could we have a yaml demo? looks like multiple research repositories are using yaml for workflow configs.
- the concept of unique id is unclear to me, why and where to ensure the uniqueness? also `class_table[name] = modname` doesn't allow the same class name under different packages, this could be an issue when importing from different packages
- the class scanner could be mostly replaced by `pydoc.locate`?  see https://github.com/facebookresearch/hydra/blob/5f605a863dd4aed8f432884fcd1e1b388c8fb559/hydra/_internal/utils.py#L554-L558
https://github.com/python/cpython/blob/main/Lib/pydoc.py#L1707
- this repo is achieving a similar goal https://github.com/catalyst-team/hydra-slayer 
- the hydra implementation for class instantiation also looks powerful https://github.com/facebookresearch/hydra/blob/5f605a863dd4aed8f432884fcd1e1b388c8fb559/hydra/_internal/instantiate/_instantiate2.py#L121 (test cases: https://github.com/facebookresearch/hydra/blob/5f605a863dd4aed8f432884fcd1e1b388c8fb559/tests/instantiate/test_instantiate.py#L64), do we really want to implement from scratch here... 

cc @ericspod @rijobro

Nic-Ma:
Thanks for @wyli 's review and suggestions.

Hi @ericspod @rijobro ,

Could you please help also take a look and share some opinions?
I will come back to work on this ticket tomorrow.

Thanks in advance.

ericspod:
Hydra does do some of the things we want to include here, do we want to add it as a dependency?  I'm not sure it could be an optional dependency since the model archiving functionality would need it to be there and it's a pretty important that it's present in MONAI.

wyli:
this one also looks interesting https://github.com/jmespath/jmespath.py

ericspod:
> this one also looks interesting https://github.com/jmespath/jmespath.py

Something structured like this looks like a good idea, I do like the "foo.baz" syntax. It's rather more complex that we need for this application perhaps, again though the concern is adding another dependency. 

I have a bit of code in a tool for generating random data that instantiates objects from a YAML schema. This [function](https://github.com/KCL-BMEIS/RandomDataset/blob/33422b92254bf0c9c9a3e0e7aed096f037a9b41c/randomdataset/schemaparser.py#L43) to creates objects from the dictionary derived from the schema recursively so that new objects can be used as constructor arguments when instantiating others. The [find_type_def](https://github.com/KCL-BMEIS/RandomDataset/blob/33422b92254bf0c9c9a3e0e7aed096f037a9b41c/randomdataset/utils.py#L7) is pretty well the same as `get_class` and much of the other code here is doing a very similar thing. 

I'm good with this going forward if we want to keep the momentum going on the model archive.

ericspod (approved):

Nic-Ma:
Hi @ericspod @wyli ,

Thanks very much for your review and comments, suggestions.
I was busy with NVIDIA Clara issues when came back from the Chinese New Year holiday this week, now I am checking other repos' implementations of config parsing and trying to resolve your comments.

Thanks.

Nic-Ma:
Hi @wyli @ericspod ,

Thanks for your great suggestions, I have totally refined this PR for:
1. Added `**kwargs` to override or add args when building config.
2. Added support to config `function` with default args based on `partial`.
3. Replaced `get_class()` with enhanced `locate()`, referring to `Hydra`.
4. Added support for same class name in different packages, will give a warning and suggest user to specify the `path` directly.

About @wyli 's questions:
1. The unique ID is used to mark every component in the nested config, then other configs can easily use it as argument, or get the config item in python script and modify for lazy instantiation. For example:
```
{
    "preprocessing": {
        "<name>": "Compose",
        "<args>": {
            "transforms": [
                {
                    "<name>": "LoadImaged",
                    ...
                },
                {
                    "<name>": "RandRotated",
                    ...
                }
            ]
        }
    },
    "debug": {
        "<name>": "SomeDebugTool",
        "<args>": {
            "first_random_transform": "@preprocessing#args#transforms#1"       -----> example here
        }
    }
}
```
2. The `module scanner` is different from the Hydra locate, it's designed to preload the class and function names in the specified packages and modules, so **users only need to set class or function name** in the config, instead of the full path like Hydra example:
https://github.com/facebookresearch/hydra/blob/main/tests/instantiate/test_instantiate.py#L93
And actually, if you don't want to set class name, this PR also supports full path, you just need to use the `<path>` instead pf `<name>` in config.
I think for users of this model package, **90% config content will be MONAI components**, they can preload MONAI modules without error.
And thanks for the sharing, I updated the instantiation part referring to Hydra.

3. I see Hydra support recursively building nested targets (B defined in A config), but it doesn't support dependent targets (A, B at the same level, A is one arg of B, which is addressed by "@" in this PR), so I think we can't use `Hydra` directly.
And I agree with @ericspod , we should try to avoid introducing new 3rd dependency.

4. I prepared both JSON and YAML config example as below:
JSON:
```
{
  "preprocessing": [
    {
      "<name>": "LoadImaged",
      "<args>": {
        "keys": "image"
      }
    },
    {
      "<name>": "Orientationd",
      "<args>": {
        "keys": "image",
        "axcodes": "RAS"
      }
    },
    {
      "<name>": "ToTensord",
      "<args>": {
        "keys": "image"
      }
    }
  ],
  "datalist": {
    "<name>": "DatasetFunc",
    "<args>": {
      "data": "./dataset.json",
      "func": "$monai.data.load_decathlon_datalist",
      "is_segmentation": true,
      "data_list_key": "test",
      "base_dir": "@dataset_dir"
    }
  },
  "dataset": {
    "<name>": "Dataset",
    "<args>": {
      "data": "@datalist",
      "transform": "@preprocessing"
    }
  },
  "dataloader": {
    "<name>": "DataLoader",
    "<args>": {
      "dataset": "@dataset",
      "batch_size": 1,
      "shuffle": false,
      "num_workers": 4
    }
  }
}
```
YAML:
```
preprocessing:
- "<name>": LoadImaged
  "<args>":
    keys: image
- "<name>": Orientationd
  "<args>":
    keys: image
    axcodes: RAS
- "<name>": ToTensord
  "<args>":
    keys: image
datalist:
  "<name>": DatasetFunc
  "<args>":
    data: "./dataset.json"
    func: "$monai.data.load_decathlon_datalist"
    is_segmentation: true
    data_list_key: test
    base_dir: "@dataset_dir"
dataset:
  "<name>": Dataset
  "<args>":
    data: "@datalist"
    transform: "@preprocessing"
dataloader:
  "<name>": DataLoader
  "<args>":
    dataset: "@dataset"
    batch_size: 1
    shuffle: false
    num_workers: 4
```

Could you guys please help review it again?

Thanks in advance.

ericspod:
We had briefly discussed JSON vs. YAML in the past, in addition to your reasons we've used JSON elsewhere in MONAI so we should be consistent about using it elsewhere. I prefer the look of YAML but it's a matter of taste, some prefer whitespace delineation vs. bracket delineation.

Nic-Ma:
Thanks @ericspod for your review and double confirmation.
@wyli Could you please help also take a look? If you have any more modification requests or questions, I will try to address them ASAP.

Thanks in advance.

monai/utils/module.py
```diff
@@ -193,7 +211,127 @@ def load_submodules(basemod, load_all: bool = True, exclude_pattern: str = "(.*[
...
+    """
+    Locate an object by name or dotted path, importing as necessary.
+    Refer to Hydra: https://github.com/facebookresearch/hydra/blob/v1.1.1/hydra/_internal/utils.py#L554.
+
+    Args:
+        path: full path of the expected component to locate and import.
+
+    """
+    parts = [part for part in path.split(".") if part]
+    for n in reversed(range(1, len(parts) + 1)):
```
ericspod, line 281: This might be a little better:

```python
parts = [p for p in path.split(".") if p]
for n in range(len(parts), 0, -1)):
        try:
            obj = import_module(".".join(parts[:n]))
            break
        except Exception as exc_import:
            if n == 1:
                raise ImportError(f"Cannot import module '{path}'.") from exc_import
```

ericspod:
I had a minor suggestion but otherwise still approved!

monai/utils/module.py
```diff
@@ -193,7 +211,127 @@ def load_submodules(basemod, load_all: bool = True, exclude_pattern: str = "(.*[
...
+    Args:
+        pkgs: the expected packages to scan modules and parse component names in the config.
+        modules: the expected modules in the packages to scan for all the components.
+            for example, to parser "LoadImage" in config, `pkgs` can be ["monai"], `modules` can be ["transforms"].
+        excludes: if any string of the `excludes` exists in the full module name, don't import this module.
+
+    """
+
+    def __init__(
+        self, pkgs: Sequence[str], modules: Optional[Sequence[str]] = None, excludes: Optional[Sequence[str]] = None
```
wyli, line 228: had a quick test of scanning 'itk' and 'torchvision' but it doesn't work... I still feel this is importing is heavy and do we need this if the config is lazily instantiated?
```py
>>> from monai.utils.module import ComponentScanner
ComponentScanner>>> ComponentScanner(["itk"])
  File "/Users/Documents/MONAI/monai/utils/module.py", line 233, in __init__
    self._components_table = self._create_table()
  File "/Users/Documents/MONAI/monai/utils/module.py", line 246, in _create_table
    module = import_module(modname)
  File "/usr/local/anaconda3/envs/py37/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/usr/local/anaconda3/envs/py37/lib/python3.7/site-packages/itk/ITKAnisotropicSmoothingPython.py", line 13, in <module>
    from . import _ITKAnisotropicSmoothingPython
ImportError: PyCapsule_Import could not import module "_ITKCommonPython"
>>> ComponentScanner(["torchvision"])
  File "/Users//Documents/MONAI/monai/utils/module.py", line 233, in __init__
    self._components_table = self._create_table()
  File "/Users//Documents/MONAI/monai/utils/module.py", line 246, in _create_table
    module = import_module(modname)
  File "/usr/local/anaconda3/envs/py37/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 670, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 583, in module_from_spec
  File "<frozen importlib._bootstrap_external>", line 1043, in create_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
ImportError: dynamic module does not define module export function (PyInit__C)
>>> 
```

monai/utils/module.py
Nic-Ma: Sure, updated in the latest commit.

Thanks.

monai/utils/module.py
Nic-Ma: Hi @wyli ,

Thanks for your review and test.
In the overall example, I plan to let users define the `packages` and `modules` in the `metadata.json`.
For 90% MMARs, only need to import `monai` and `torch` with specific `modules`(transforms, networks, etc.)
If you want to use some `ITK` APIs for spetial logic, you can do it in 2 ways:
(1) use "<path>": "itk.XXX" directly like the Hydra example.
(2) Only define the expected APIs of ITK in the `packages` and `modules` instead of the whole ITK.

Maybe you can help provide a simple python logic with ITK, then I try to convert it to a config and parse?

Thanks.

monai/utils/module.py
```diff
@@ -193,7 +211,127 @@ def load_submodules(basemod, load_all: bool = True, exclude_pattern: str = "(.*[
...
+            name: name of the expected class or function.
+
+        """
+        mods = self._components_table.get(name, None)
+        if isinstance(mods, list) and len(mods) == 1:
+            mods = mods[0]
+        return mods
+
+
+def locate(path: str):
```
wyli, line 271: any idea why we don't use `pydoc.locate` directly?

monai/utils/module.py
wyli: Hi @Nic-Ma my question is not about itk. It's about the API we provide here. We mainly offer Python APIs in this package, if they only work for certain use cases of config files we should make it a private class and validate the inputs. This class recursively imports all attributes in the given packages but the actual config might only use a tiny portion for a given workflow, mostly unnecessarily.

Nic-Ma: Hi @wyli ,

That's a good point, I agree it can be a private class. I didn't expect users to use it directly.
In the whole picture, `ConfigParser` will use it internally:
https://github.com/Project-MONAI/MONAI/pull/3593/files#diff-57ffc6dd5a285d7b6e1fa51795ee9e39e296abe50c04006ae0edf0e979af3fbfR52

Let me move the class out of `utils` and be `private`.

Thanks.

monai/utils/module.py
Nic-Ma: Hi @wyli ,

`pydoc.locate` can't work for several cases, that's why `Hydra` reimplemented it. Especially to import in **reverse order**:
https://github.com/facebookresearch/hydra/blob/main/hydra/_internal/utils.py#L558
For example, you can try `pydoc.locate` with this path "torch.optim._multi_tensor.adam.Adam":
https://github.com/pytorch/pytorch/blob/master/torch/optim/_multi_tensor/adam.py#L5
```
>>> import torch
>>> import pydoc
>>> 
>>> print(pydoc.locate("torch.optim._multi_tensor.adam.Adam"))
None
```

Thanks.

wyli: I see, `torch.optim._multi_tensor` looks like a private module, perhaps we shouldn't support configuring with the packages private modules?

monai/utils/module.py
Nic-Ma: Hi @ericspod @wyli ,

Thanks for the discussion.
I have a tradeoff idea: for the first MVP, let's only support to load the `classes` or `functions` from MONAI codebase as qualified `<name>`. For APIs from any other packages, just use `<path>` with the full path.
I believe more than 90% APIs in the JSON config should come from MONAI itself.
We can enhance this part later when we get enough feedback later.
What do you think?

Thanks.

Nic-Ma: Hi @ericspod @wyli ,

I updated this part and the unit tests according to both you guys' comments.
Could you please help take a look?

Thanks in advance.

monai/utils/module.py
wyli: I thought this is expected because
```py
>>> import torch
>>> torch.optim.Adam
<class 'torch.optim.adam.Adam'>
>>> torch.optim.adam.Adam
AttributeError: module 'torch.optim' has no attribute 'adam'
>>> 
```

Nic-Ma: Hi @wyli ,

But both below imports work:
```py
>>> from torch.optim import Adam
>>> 
>>> a = Adam(...)
>>> 
>>> from torch.optim.adam import Adam
>>> 
>>> b = Adam(...)
```

`torch.optim.Adam` should be supported to config and parse.
What do you think?

Thanks.

monai/utils/module.py
Nic-Ma: Hi @ericspod , does the `load_submodules` already load all the `classes` and `functions` of MONAI code? Then we can call something like `pydoc.locate("LoadImaged")` directly to get the `LoadImaged` class? Seems it should not work for same name APIs.
If not, I think we still need `ComponentScanner`.

Thanks in advance.

ericspod: It was meant to, if it's still being used in the same way then it should continue to do that. Running `python -c 'import monai,sys;print([m for m in sys.modules.keys() if "monai" in m])'` seems to indicate it does.

Nic-Ma: Hi @ericspod @wyli ,

Thanks for the tips, seems `load_submodules` already loaded all the `modules`. I totally simplified the `ComponentScanner` to leverage `load_submodules` results to scan all the `classes` and `functions` directly with a `excludes` filter. And removed the args for `pkgs` and `modules`.
I think the API is simple and clear now, it only works for MONAI components.
Do you have any other comments or concerns?

Thanks in advance.

monai/utils/module.py
wyli: agree that `torch.optim.Adam` should be supported,
this seems to work:
```py
>>> import pydoc
>>> pydoc.locate('torch.optim.Adam')
<class 'torch.optim.adam.Adam'>
>>> 
```
and the `torch.optim.Adam` usage is consistent with the documentation https://pytorch.org/docs/stable/generated/torch.optim.Adam.html#torch.optim.Adam

it's mentioned in this ticket: https://github.com/pytorch/pytorch/issues/271

so I feel it's good to use `pydoc` here.

Nic-Ma: Hi @wyli ,

Thanks for your testing.
OK, as we aligned only need to support `<name>` for MONAI APIs, I want to double confirm that we don't have such "del XXX" case in any MONAI `__init__.py` source code, right?
If yes, let me remove this `locate` function and use `pydoc` directly.

Thanks.

Nic-Ma: Hi @wyli @ericspod ,

BTW, I added `ComponentScanner` unit tests to ensure all the scanned MONAI components can support `locate`.
Do you have any other comments or concerns?

Thanks in advance.

wyli (commented):
thanks Nic, I put some detailed in-line comments. there's no need to reply to the comments if you address them by updating the PR. please just reply if I misunderstand anything in the comments...

wyli:
in general I also feel the definition of `ConfigComponent` is unclear, it has an internal `self.config`, but quite a few of its methods also takes `config: Optional[Dict] = None`, the class should have a clearer pattern. Also I couldn't understand `self.id` vs `dependent_ids`, we need a clear definition of `id` in this feature...

monai/apps/mmars/config_resolver.py
```diff
@@ -0,0 +1,227 @@
...
+    Map the all the names and the module names in a table.
+
+    Args:
+        excludes: if any string of the `excludes` exists in the full module name, don't import this module.
+
+    """
+
+    def __init__(self, excludes: Optional[Sequence[str]] = None):
+        self.excludes = [] if excludes is None else excludes
+        self._components_table = self._create_table()
```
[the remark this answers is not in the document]
Nic-Ma: I usually prefer to put init logic in the constructor of a class to make the following logic `read-only`, in case we may want to support multi-thread parsing logic.
But anyway, we don't have such feature so far, updated your suggestion in the latest commit, it may help to optimize the performance especially for `lazy instantiation`.

Thanks.

monai/apps/mmars/config_resolver.py
```diff
@@ -0,0 +1,227 @@
...
+from importlib import import_module
+from typing import Any, Dict, List, Optional, Sequence
+
+from monai.apps.mmars.utils import search_configs_with_deps, update_configs_with_deps
+from monai.utils.module import instantiate
+
+__all__ = ["ComponentScanner", "ConfigComponent"]
+
+
+class ComponentScanner:
```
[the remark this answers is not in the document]
Nic-Ma: `ComponentLocator` looks better to me, updated in the latest commit.

Thanks.

monai/apps/mmars/config_resolver.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    Scan all the available classes and functions in the MONAI package.
+    Map the all the names and the module names in a table.
+
+    Args:
+        excludes: if any string of the `excludes` exists in the full module name, don't import this module.
+
+    """
+
+    def __init__(self, excludes: Optional[Sequence[str]] = None):
```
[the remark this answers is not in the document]
Nic-Ma: I added `MOD_START = "monai"` in the latest commit, I think `START` may be more accurate in our case here.

Thanks.

monai/apps/mmars/config_resolver.py
```diff
@@ -0,0 +1,227 @@
...
+        self.excludes = [] if excludes is None else excludes
+        self._components_table = self._create_table()
+
+    def _create_table(self):
```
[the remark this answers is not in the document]
Nic-Ma: I changed it to `_find_module_names` and `_find_classes_or_functions` in the latest commit.

Thanks.

monai/apps/mmars/config_resolver.py
```diff
@@ -0,0 +1,227 @@
...
+                for name, obj in inspect.getmembers(module):
+                    if (inspect.isclass(obj) or inspect.isfunction(obj)) and obj.__module__ == modname:
+                        if name not in table:
+                            table[name] = []
+                        table[name].append(modname)
+            except ModuleNotFoundError:
+                pass
+        return table
+
+    def get_component_module_name(self, name):
```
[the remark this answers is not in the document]
Nic-Ma: Yes, added `typehint` in the latest commit.

Thanks.

monai/apps/mmars/config_resolver.py
```diff
@@ -0,0 +1,227 @@
...
+            the id can be useful to quickly get the expected item in a complicated and nested config content.
+        config: config content of current component, can be a `dict`, `list`, `string`, `float`, `int`, etc.
+        scanner: ComponentScanner to help get the `<name>` or `<path>` in the config and build instance.
+        globals: to support executable string in the config, sometimes we need to provide the global variables
+            which are referred in the executable string. for example: `globals={"monai": monai} will be useful
+            for config `"collate_fn": "$monai.data.list_data_collate"`.
+
+    """
+
+    def __init__(self, id: str, config: Any, scanner: ComponentScanner, globals: Optional[Dict] = None) -> None:
```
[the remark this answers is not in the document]
Nic-Ma: Added default `locator` and default `id` to `None`. ID actually can be `None` if no other components depend on it.

Thanks.

monai/apps/mmars/config_resolver.py
```diff
@@ -0,0 +1,227 @@
...
+        ):
+            # if marked as `disabled`, skip parsing
+            return config
+
+        args = config.get("<args>", {})
+        args.update(kwargs)
+        path = self._get_path(config)
+        return instantiate(path, **args)
+
+    def _get_path(self, config):
```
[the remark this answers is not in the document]
Nic-Ma: Removed this function directly, not necessary.

Thanks.

monai/apps/mmars/config_resolver.py
```diff
@@ -0,0 +1,227 @@
...
+        if isinstance(config, dict):
+            for v in config.values():
+                if self._check_dependency(v):
+                    return True
+        if isinstance(config, str):
+            if config.startswith("&") or "@" in config:
+                return True
+        return False
+
+    def build(self, config: Optional[Dict] = None, **kwargs) -> object:
```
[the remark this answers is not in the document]
Nic-Ma: That's a good point, actually it's better to be a static function, updated in the latest commit.

Thanks.

monai/apps/mmars/config_resolver.py
```diff
@@ -0,0 +1,227 @@
...
+    - `{"preprocessing": [{"<name>": "LoadImage", "<args>": {"keys": "image"}}]}`
+    - `{"<name>": "LoadImage", "<args>": {"keys": "image"}}`
+    - `"<name>": "LoadImage"`
+    - `"keys": "image"`
+
+    It can search the config content and find out all the dependencies, then build the config to instance
+    when all the dependencies are resolved.
+
+    Here we predefined 4 kinds special marks (`<>`, `#`, `@`, `$`) to parse the config content:
+    - "<XXX>": like "<name>" is the name of a target component, to distinguish it with regular key "name"
```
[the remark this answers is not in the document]
Nic-Ma: Updated in the latest commit.
`get_id` will be useful when resolving dependencies with other config components in the whole picture, `get_config` can be useful for lazy instantiation to modify the config content even before resolving dependencies.

Thanks.

monai/apps/mmars/config_resolver.py
```diff
@@ -0,0 +1,227 @@
...
+            not isinstance(config, dict)
+            or ("<name>" not in config and "<path>" not in config)
+            or config.get("<disabled>") is True
+        ):
+            # if marked as `disabled`, skip parsing
+            return config
+
+        args = config.get("<args>", {})
+        args.update(kwargs)
+        path = self._get_path(config)
```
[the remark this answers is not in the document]
Nic-Ma: Sure, updated in the latest commit.
As I also added the `**kwargs` support at runtime, I didn't add the `unpack` arg, maybe we can consider to add it in future if really necessary.

Thanks.

monai/apps/mmars/config_resolver.py
```diff
@@ -0,0 +1,227 @@
...
+        Recursively search all the content of current config compoent to get the ids of dependencies.
+        It's used to build all the dependencies before build current config component.
+        For `dict` and `list`, treat every item as a dependency.
+        For example, for `{"<name>": "DataLoader", "<args>": {"dataset": "@dataset"}}`, the dependency ids:
+        `["<name>", "<args>", "<args>#dataset", "dataset"]`.
+
+        """
+        return search_configs_with_deps(config=self.config, id=self.id)
+
+    def get_updated_config(self, deps: dict):
```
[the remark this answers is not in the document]
Nic-Ma: Updated the function name to `resolve_config` as it not only resolved the dependencies, but also called `eval()` for the code, so not only "resolve deps".
About your question: Sorry the previous code is not very clear about this part. It should not replace the `self.config` directly, the `self.config` is still useful for lazy instantiation. For example you can `get_config()` and modify something and `set_config()` and `resolve_config` again. You can find more clear code and doc-string in the latest commit.

Thanks.

monai/apps/mmars/config_resolver.py
```diff
@@ -0,0 +1,227 @@
...
+        """
+        If all the dependencies are ready in `deps`, update the config content with them and return new config.
+        It can be used for lazy instantiation, the returned config has no dependencies, can be built immediately.
```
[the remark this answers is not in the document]
Nic-Ma: added `no_deps` arg when setting the config content.

Thanks.

monai/apps/mmars/utils.py
```diff
@@ -0,0 +1,88 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+from typing import Dict, List, Optional, Union
+
+
+def search_configs_with_deps(config: Union[Dict, List, str], id: str, deps: Optional[List[str]] = None) -> List[str]:
```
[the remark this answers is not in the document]
Nic-Ma: Below code can work in the latest commit:
```py
c = ConfigComponent({'name': '@other'}, id='test')
c.resolve_config(deps={'other': 1})
print(c.get_resolved_config())
```
Output is:
```
{'name': 1}
```

Thanks.

Nic-Ma:
Hi @wyli ,

Thanks very much for your review and detailed suggestions.
I totally updated the PR based on your comments and my new ideas, also added more doc-strings and unit tests.
Could you please take a look again?

Thanks in advance.

monai/apps/mmars/utils.py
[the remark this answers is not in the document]
wyli: thanks, for the same `c`, I think now `c.build()` should return `1`?

monai/utils/module.py
```diff
@@ -210,12 +210,14 @@ def instantiate(path: str, **kwargs):
     """
 
     component = locate(path)
+    if component is None:
+        raise ModuleNotFoundError(f"Cannot locate '{path}'.")
```
wyli, line 214: is it ok to raise an error here? @Nic-Ma

Nic-Ma: Looks good to me as we already covered all the MONAI classes and functions in the locate unit test.

Thanks.

Nic-Ma:
I have merged this PR into the whole draft PR: https://github.com/Project-MONAI/MONAI/pull/3593 and verified the overall features.
Will merge this PR and create following PRs for other parts.

Thanks.

PR #3739, opened 2022-01-29
3734 Enhance `CacheDataset` to avoid duplicated cache

### LINKED ISSUE #3734: Avoid duplicated cache content in CacheDataset
Feature request from @dongyang0122 during NVIDIA Clara development:
If I use some method (I think it's `resample_dataset` API of MONAI) to augment the original dataset and use `CacheDataset`, will it cache duplicated content for the same samples? Or it only caches the unique ones?

Actually, our `PersistentDataset` can only cache unique items because it uses Hash key as ID instead of indices, I think we can enhance `CacheDataset` to optionally also support caching in a dict with Hash key.
@ericspod @wyli @rijobro Do you have any concerns?

Nic-Ma:
Fixes #3734 .

### Description
This PR enhanced`CacheDataset` to avoid caching duplicated content for the augmented dataset.

### Status
**Ready**

### Types of changes


monai/data/dataset.py
```diff
@@ -1348,3 +1348,37 @@ def __init__(
...
+        kwargs: other arguments of `CacheDataset` except for `data`.
+
+    """
+
+    def __init__(self, data: Sequence, hash_func: Callable[..., bytes] = pickle_hashing, **kwargs) -> None:
+        self.hash_func = hash_func
+        mapping = {self.hash_func(v): v for v in data}
+        # only compute cache for the unique items of dataset
+        super().__init__(data=list(mapping.values()), **kwargs)
+        self._cache_keys = list(mapping)[: self.cache_num]
```
wyli, line 1373: Thanks, this looks straightforward, why not merge this as an option to the cache dataset and by default enable it? With this, perhaps the original cachedatset is less useful.

Nic-Ma: Hi @wyli ,

The reason I didn't add it in the CacheDataset:
1. It may be faster to use indices directly(no hash computation and no need to search the key name in the list)? Performance is very important for CacheDataset.
2. I am not sure how it affect the subclasses, like `SmartCacheDataset` replacement logic with duplicated items..

Anyway, if you think it's better to update the `CacheDataset` directly, I can think about it more.
What do you think?

Thanks.

wyli: Sure if the solution is simple and efficient I think we should merge and make it default. Otherwise the current is ok

Nic-Ma: Hi @wyli ,

Thanks for your review.
I tried to leverage this Hash logic for the subclass `SmartCacheDataset`, but it's really not clear to develop the "replacement" behavior, for example: if we still use all the data for training, everytime we do replacement, the `unique cache` size may be different, so the training may out of memory at some epoch. And in this case, the definition of `cache_num` will be different from `CacheDataset` which is only the unique cache size.
I would suggest current independent `HashCacheDataset` for this use case, if we get more feature requests for it, we can consider to revisit it and discuss how to make it work for all the subclasses of `CacheDataset`.
What do you think?

Thanks in advance.

wyli: I agree we can consider the samrtcachedataset later. I think the idea here is to have an option in the new cachedataset to switch between the original cachedataset vs hash-based cachedataset. it should be non-breaking and doesn't affect the subclass of cachedataset. or do I misunderstand?

Nic-Ma: Hi @wyli ,

Thanks for your suggestion.
And I also found a bug in the previous `CacheDataset`, when calling `set_data()` to update data and cache, the `self.cache_num` is not updated...
Let me try to integrate HashCacheDataset into CacheDataset directly.

Thanks.

Nic-Ma: Updated in the latest commit.

Thanks.

Nic-Ma:
Mark: `autoformat` is updated, so `/black` introduced some extra change in this PR.

Thanks.

wyli:
Sure, I'll separate the style changes into another style PR...

Nic-Ma:
> Sure, I'll separate the style changes into another style PR...

Oh, OK, let me submit it.

Thanks.

monai/data/dataset.py
```diff
@@ -754,6 +775,12 @@ def _load_cache_item(self, idx: int):
         return item
 
     def _transform(self, index: int):
+        if self.hash_as_key:
+            key = self.hash_func(self.data[index])
+            if key in self._cache_keys[: self.cache_num]:
+                # if existing in cache, get the index
+                index = self._cache_keys.index(key)
```
wyli, line 782: this will have O(n) searching twice, instead of O(1) hash look up... perhaps we can have (after line 742)
```py
self._cache = dict(zip(self._cache_keys[: self.cache_num], self._fill_cache())
```
then here the dict cache will have fast access:
```py
key = self.hash_func(self.data[index])
if key in self._cache:
    index = key  # this should work with `data = self._cache[index]`
```

monai/data/dataset.py
```diff
@@ -695,19 +697,30 @@ def __init__(
...
+                if key exists, avoid saving duplicated content. it can help save memory when
+                the dataset has duplicated items or augmented dataset.
+            hash_func: if `hash_as_key`, a callable to compute hash from data items to be cached.
+                defaults to `monai.data.utils.pickle_hashing`.
 
         """
         if not isinstance(transform, Compose):
             transform = Compose(transform)
         super().__init__(data=data, transform=transform)
+        self.set_num = cache_num
```
wyli, line 710:
```suggestion
        self.set_num = cache_num  # tracking the user-provided option
```

monai/data/dataset.py
```diff
@@ -695,19 +697,30 @@ def __init__(
...
+        self.set_rate = cache_rate
```
wyli, line 711:
```suggestion
        self.set_rate = cache_rate  # tracking the user-provided option
```

monai/data/dataset.py
Nic-Ma: Hi @wyli ,

Thanks for your sugestion.
So in this way, the `self._cache` will be `dict` in Hash mode and be `list` in the previous regular mode?
Or maybe let me change it to `dict` directly even for previous regular mode?

Thanks.

wyli: yes, to me it makes sense to have the hash-based using dict and index-based using list

monai/data/dataset.py
Nic-Ma: updated in the latest commit.

Thanks.

monai/data/dataset.py
Nic-Ma: updated in the latest commit.

Thanks.

monai/data/dataset.py
Nic-Ma: OK, updated in the latest commit.

Thanks.

wyli (approved):

PR #3770, opened 2022-02-05
3763 Enhance the doc of `ThreadDataLoader` for `num_workers`

### LINKED ISSUE #3763: Clarify the `num_workers` in `ThreadDataLoader`
When I was introducing GPU transforms and the associated `ThreadDataLoader` to users, got several times feedback about the `num_workers` arg, which is confusing that users think it means the multi-threads in `ThreadDataLoader`, but actually it's the multi-processing workers of PyTorch DataLoader.
Would be nice to clarify this arg and the use cases.

Nic-Ma:
Fixes #3763 .

### Description
This PR enhanced the doc of `ThreadDataLoader` to clarify `num_workers` arg and the typical usage.

### Status
**Ready**

### Types of changes


monai/data/thread_buffer.py
```diff
@@ -100,7 +106,11 @@ class ThreadDataLoader(DataLoader):
         buffer_size: number of items to buffer from the data source.
         buffer_timeout: time to wait for an item from the buffer, or to wait while the buffer is full when adding items.
         num_workers: number of the multi-processing workers in PyTorch DataLoader.
+            note that it's not the number of threads for the `ThreadBuffer` in this class, which always uses 2 threads:
```
wyli, line 109: I don't see any good use case with num_workers>0, perhaps we just remove it?

Nic-Ma: Makes sense to me.
I removed it in the latest commit and added some description in `kwargs`.

Thanks.

wyli: Thanks, I think after this, we don't need to mention num_workers in the docstring to avoid confusion, correct?

Nic-Ma: OK, updated in the latest commit.

Thanks.

wyli (approved):
Thanks, it looks good to me.

PR #3774, opened 2022-02-07
3769 Enhance logger logic of StatsHandler and DataStats

Nic-Ma:
Fixes #3769 .

### Description
Thanks for the feedback in #3769 , this PR mainly:
1. Added support to print log if the base log level is higher than INFO.
2. Fixed the wrong logic of `StatsHandler` if `name=None` and `addtional_handler`, in the current code it will attach the additional handler to the system root logger.
3. Fixed the **duplicated print** in `DataStats` if base log level is already INFO.
4.  Added warning message if using `engine.logger` and may not be able to print.
5. Enhanced doc-string to explain the expected behavior.
6. Removed unnecessary global logging setting in the unit tests.

### Status
**Ready**

### Types of changes


monai/handlers/stats_handler.py
```diff
@@ -91,13 +99,19 @@ def __init__(
...
-        self.logger = logging.getLogger(name)
-        self._name = name
-
         self.tag_name = tag_name
         self.key_var_format = key_var_format
-        if logger_handler is not None:
-            self.logger.addHandler(logger_handler)
+        self.logger = logging.getLogger(name)  # if `name` is None, will default to `engine.logger` when attached
+        if name is not None:
+            if logging.root.getEffectiveLevel() > logging.INFO:
```
wyli, line 106: normally the application/user of this `StatsHandler` should prepare appropriate loggers/handlers from the outside and pass them (by either the logger's name or logger handler) to this class. This is to decouple the logic of logging configuration and the actual code of generating logging content. So I think this addition should be in a logger config in https://github.com/Project-MONAI/tutorials/blob/master/3d_classification/ignite/densenet_training_dict.py instead of hard-coded here. cc @ericspod

Nic-Ma: Hi @wyli ,

We already have the similar logic in `DataStats` transform:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/utility/array.py#L594
Because for 90% users of `DataStats` or `StatsHandler`, they expect to see the log on screen, and I already received this feedback several times from @holgerroth and our SAs when they missed to call `logging.basicConfig(stream=sys.stdout, level=logging.INFO)`.
So maybe it's better to add `stdout` handler for user experience if they use this API to print stats?

Thanks.

Nic-Ma: Hi @wyli ,

Updated in the latest commit.

Thanks in advance.

monai/handlers/stats_handler.py
```diff
@@ -34,12 +34,26 @@ class StatsHandler:
...
+
     Default behaviors:
         - When EPOCH_COMPLETED, logs ``engine.state.metrics`` using ``self.logger``.
         - When ITERATION_COMPLETED, logs
           ``self.output_transform(engine.state.output)`` using ``self.logger``.
 
-    Usage example is available in the tutorial:
+    Usage example::
+
+        logging.basicConfig(level=logging.INFO)
```
ericspod, line 49: For `basicConfig` it might be worth adding `stream=sys.stdout` to demonstrate output to screen which I think is a commonly desired behaviour.

monai/transforms/utility/array.py
```diff
@@ -590,10 +595,12 @@ def __init__(
...
         self._logger_name = "DataStats"
         _logger = logging.getLogger(self._logger_name)
-        _logger.setLevel(logging.INFO)
-        console = logging.StreamHandler(sys.stdout)  # always stdout
-        console.setLevel(logging.INFO)
-        _logger.addHandler(console)
+        if logging.root.getEffectiveLevel() > logging.INFO:
+            # if the root log level is higher than INFO, set a separate stream handler to record
+            _logger.setLevel(logging.INFO)
+            console = logging.StreamHandler(sys.stdout)
```
ericspod, line 601: This is the same code as the original but I wonder if hard-coding changing the log level of `_logger` to INFO and writing to stdout is desirable.

monai/handlers/stats_handler.py
Nic-Ma: Sure, added in the latest commit.

Thanks.

monai/handlers/stats_handler.py
```diff
@@ -77,12 +91,13 @@ def __init__(
...
             state_attributes: expected attributes from `engine.state`, if provided, will extract them
                 when epoch completed.
-            name: identifier of logging.logger to use, defaulting to ``engine.logger``.
+            name: identifier of logging.logger to use, if None, defaulting to ``engine.logger``.
             tag_name: when iteration output is a scalar, tag_name is used to print
                 tag_name: scalar_value to logger. Defaults to ``'Loss'``.
             key_var_format: a formatting string to control the output string format of key: value.
-            logger_handler: add additional handler to handle the stats data: save to file, etc.
-                all the existing python logging handlers: https://docs.python.org/3/library/logging.handlers.html.
+            logger_handler: only work when `name` is not None, add additional handler to the new logger to handle
```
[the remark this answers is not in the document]
Nic-Ma: Hi @ericspod ,

I removed this arg according to @wyli 's suggestion, let's expect to setup the logger at outside with same `name` directly, then get the logger by `name` in StatsHandler.

Thanks.

monai/transforms/utility/array.py
Nic-Ma: Changed to always `setLevel(logging.INFO)` in the latest commit.
Checking the `root.level` is to avoid adding duplicated stdout handler when root level is already INFO.

Thanks.

ericspod (approved):

Nic-Ma:
Hi @wyli @ericspod ,

Thanks for your review.
I tested the PR with workflow integration test and some tutorial examples.
Also submitted a PR in tutorial to show the usage: https://github.com/Project-MONAI/tutorials/pull/540.

Thanks.

PR #3776, opened 2022-02-07
KeepLargest... by default use all labels

rijobro:
- We also required that the user define the applied labels at construction time. However, for many label segmentation, not all labels will necessarily always be present, and this causes an error. I think the default should be to apply to all present labels (calculated on the fly), but allow the user to limit to a subset of labels (as is the current implementation).
- Improved testing
- Harmonise implementations between `FillHoles` and `KeepLargest...`

### Status
**Ready**

### Types of changes


rijobro:
@wyli @Nic-Ma I've reverted the modifications to one-hot in this PR.

wyli:
thanks, this provides a good default for `applied_labels`, I'll merge once the unit tests are fixed. (`torch.moveaxis` is for torch 1.7+, our `moveaxis` is robust https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/utils_pytorch_numpy_unification.py#L55)

wyli:
Still there is a test issue, looks line the windows API requires a long into for onehot...

wyli (approved):

PR #3782, opened 2022-02-09
3432 make vit support torchscript

### LINKED ISSUE #3432: ViT does not support torchscript

yiheng-wang-nv:

Fixes #3432 .

### Description

This PR is used to fix #3422 , different from the way used in #3433 which requires to remove `einops`, this PR keeps this library.

During this PR, the direct calling of `einops.arrange` in forward functions of `class SABlock` has been replaced by using `Rearrange` layers in order to support torchscript.

### Status

### Types of changes


wyli (approved):
thanks this looks good me, if @ahatamiz has no further comments we merge it today.

ericspod:
There may be some deep-seated incompatibilities between Torchscript and Einops: https://github.com/arogozhnikov/einops/issues/115 The issue with `build-gpu/GPU-quick-py3 (PT18+CUDA102)` failing may be related, later versions of Pytorch after 1.8 may work, I ran tests with Pytorch 1.10 just now and that version at least works. We may have to state only a certain Pytorch version is scriptable and put the condition on the script test for ViT.

PR #3802, opened 2022-02-14
fixes spacing/orientation xform ordering

wyli:

part of https://github.com/Project-MONAI/tutorials/issues/545

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/apps/deepgrow/dataset.py
+++ b/monai/apps/deepgrow/dataset.py
@@ -126,8 +126,8 @@ def _default_transforms(image_key, label_key, pixdim):
         [
             LoadImaged(keys=keys),
             AsChannelFirstd(keys=keys),
-            Spacingd(keys=keys, pixdim=pixdim, mode=mode),
             Orientationd(keys=keys, axcodes="RAS"),
+            Spacingd(keys=keys, pixdim=pixdim, mode=mode),
         ]
     )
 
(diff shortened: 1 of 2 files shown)


monai/apps/deepgrow/dataset.py
```diff
@@ -126,8 +126,8 @@ def _default_transforms(image_key, label_key, pixdim):
         [
             LoadImaged(keys=keys),
             AsChannelFirstd(keys=keys),
-            Spacingd(keys=keys, pixdim=pixdim, mode=mode),
```
wyli, line 129: the orientation should happen first, could you please help evaluate the impact of this change? most likely we don't need to retrain any existing models... @SachidanandAlle @finalelement @diazandr3s.

SachidanandAlle: yes.. this should be ok.. we change the orientation while training the model..  for existing pre-trained models.. this should not impact..

wyli: ok, that's right thanks for confirming it.

diazandr3s: Agree with Sachi. We are already using DeepEdit without the Spacing transform :)

holgerroth (approved):
LGTM

PR #3818, opened 2022-02-18
3482 Add ReferenceResolver to manage config items

Nic-Ma:
Task step 3-1 of #3482 .

### Description
This PR implemented the `ReferenceResolver` to manage all the `ConfigItem` and resolve the references.

### Status
**Ready**

### Types of changes


wyli (approved):
thanks, I made some more non-breaking changes. `find_refs_in_config` and `update_config_with_refs` share many similarities, I suspect we can somehow merge these two with an additional 'update: bool` flag. But I think it's also good in its current shape...

Nic-Ma:
> thanks, I made some more non-breaking changes. `find_refs_in_config` and `update_config_with_refs` share many similarities, I suspect we can somehow merge these two with an additional 'update: bool` flag. But I think it's also good in its current shape...

Thanks for your review and update.
I would suggest to keep current shape of the 2 functions, we may revisit and update the logic during development, let's keep it simple and easy to ready.

Thanks.

PR #3822, opened 2022-02-19
3482 3829 Add `ConfigParser` to recursively parse config content

Nic-Ma:
Task step 3-2 of https://github.com/Project-MONAI/MONAI/issues/3482 .
Fixes https://github.com/Project-MONAI/MONAI/issues/3829

### Description
This PR added the `ConfigParser` to recursively parse config content.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @wyli @ericspod ,

I have simplified the `parse` logic and removed `self.parsed` status flag.
And added 3 levels operations for the config in the doc-string of `ConfigParser`:
`get_config`, `get_config_item(resolve=True/False)`, `get_resolved_content`

Thanks in advance.

wyli:
Could you please help improve the APIs according to cpython/configparser? maybe we could use similar terminology such as `items`, `set`, `get` `read`. Note that the constructor only handles options https://github.com/python/cpython/blob/d327517b54c148eba499c777b99760356102cbe0/Lib/configparser.py#L17-L21 please have a look at the overall design. 

Our goal here is to have pythonic APIs decoupling Python scripts and workflow configurations. Please help ensure the APIs are user-friendly. I feel the examples here are very intuitive https://docs.python.org/3/library/configparser.html#module-configparser could we achieve something similar?

Nic-Ma:
Hi @wyli ,

Thanks for your sharing.
I see your point now, I think you prefer the pythonic API style, something like:
```py
parser = ConfigParser()
parser['transforms'][3]['<args>']['interp_order'] = "bilinear"
```
I can try to optimize current APIs in this way.

Thanks.

monai/apps/manifest/config_parser.py
```diff
@@ -77,7 +82,7 @@ def __init__(
     ):
         self.config = None
         self.globals: Dict[str, Any] = {}
-        globals = {"monai": "monai", "torch": "torch", "np": "numpy"} if globals is None else globals
+        globals = {"monai": "monai", "torch": "torch", "np": "numpy", "numpy": "numpy"} if globals is None else globals
```
wyli, line 85: I added "numpy" itself here

monai/apps/manifest/config_parser.py
```diff
@@ -176,28 +185,35 @@ def _do_parse(self, config, id: str = ""):
         else:
             self.reference_resolver.add_item(ConfigItem(config=item_conf, id=id))
 
-    def parse(self):
+    def parse(self, reset: bool = False):
```
wyli, line 188: added an option to do forced reset in `parse`

monai/apps/manifest/config_parser.py
```diff
@@ -176,28 +185,35 @@ def _do_parse(self, config, id: str = ""):
...
+        Args:
+            reset: whether to reset the ``reference_resolver`` before parsing. Defaults to False.
 
         """
+        if reset:
+            self.reference_resolver.reset()
         self._do_parse(config=self.config)
 
-    def get_parsed_content(self, id: str):
+    def get_parsed_content(self, id: str = "", **kwargs):
```
wyli, line 200: added support of `instantiate: bool` and `eval_expr: bool`

monai/apps/manifest/reference_resolver.py
```diff
@@ -65,11 +68,11 @@ def add_item(self, item: ConfigItem):
         """
         id = item.get_id()
         if id in self.items:
-            warnings.warn(f"id '{id}' is already added.")
+            # warnings.warn(f"id '{id}' is already added.")
```
wyli, line 71: mute this warning because we may repeatedly parse configs

monai/apps/manifest/reference_resolver.py
```diff
@@ -105,42 +111,46 @@ def _resolve_one_item(self, id: str, waiting_list: Optional[Set[str]] = None):
...
-                if d not in self.items:
-                    raise ValueError(f"the referring item `{d}` is not defined in the config content.")
+                try:
+                    look_up_option(d, self.items, print_all_options=False)
+                except ValueError as err:
+                    raise ValueError(f"the referring item `@{d}` is not defined in the config content.") from err
                 # recursively resolve the reference first
-                self._resolve_one_item(id=d, waiting_list=waiting_list)
+                self._resolve_one_item(id=d, waiting_list=waiting_list, **kwargs)
+                waiting_list.remove(d)
```
wyli, line 127: added this `waiting_list.remove(d)` otherwise there is a bug when `_resolve_one_item(id="")`

wyli (approved):
thanks, I revised the APIs and fixed a few issues. Some changes are highlighted please confirm and feel free to update. in general it looks good... @Nic-Ma

monai/apps/manifest/config_parser.py
Nic-Ma: I changed the default value of this arg to `True`, but actually I really didn't find a typical use case for `reset = False`, could you please help share some more info?

Thanks in advance.

monai/apps/manifest/config_parser.py
Nic-Ma: OK,  I see you also added this `kwargs` for `get_item()` in resolver, I didn't see any `instanatiate=False` use case in `get_item()` because no matter what the `kwargs`, `get_item()` always return the resolved `ConfigItem` instead of the instance or expression result in `resolved_content`.
Anyway, I didn't remove it in this PR, maybe you have some use cases I didn't see?

Thanks.

monai/apps/manifest/reference_resolver.py
Nic-Ma: OK, I removed this comment directly.

Thanks.

Nic-Ma:
Hi @wyli ,

Thanks for your review.
I slightly adjusted and enhanced the doc-string for the case `get_parsed_content(id, instantiate=False)`.
If you have more comments, we can adjust in PR https://github.com/Project-MONAI/MONAI/pull/3832 later, I will merge this PR for now.

Thanks.

Nic-Ma:
Hi @wyli @ericspod ,

I have updated the module name from `monai.apps.manifest` to `monai.bundle`, could you please help double confirm it again?

Thanks in advance.

monai/bundle/config_parser.py
```diff
@@ -0,0 +1,224 @@
...
+            config = config[indexing]
+        return config
+
+    def __setitem__(self, id: Union[str, int], config: Any):
+        """
+        Set config by ``id``.  Note that this method should be used before ``parse()`` or ``get_parsed_content()``
+        to ensure the updates are included in the parsed content.
+
+        Args:
+            id: id of the ``ConfigItem``, ``"#"`` in id are interpreted as special characters to
```
Nic-Ma, line 124: Hi @wyli ,

I saw you made the `sep` configurable in `ReferenceResolver`, do you want to change all the doc-string for the "#" description?

Thanks.

wyli: I think for now `#` is more readable as part of the docstring

PR #3832, opened 2022-02-21
3482 add `run` API for common training, evaluation and inference

Nic-Ma:
Task step 7 of https://github.com/Project-MONAI/MONAI/issues/3482 .

### Description
This PR added the `run` API to execute most common cases of supervised training, evaluation or inference from the config directly.

### Status
**Ready**

### Types of changes


Nic-Ma:
This PR depends on #3822 .

Thanks.

Nic-Ma:
Hi @ericspod @wyli , Here I use Use the `distutils.util.strtobool()` to convert `string` to `bool`, maybe you guys have other better options?

Thanks in advance.

Nic-Ma:
Hi @ericspod @wyli @rijobro ,

Due to the Blossom CI issue, we can't merge any PRs recently, this PR depends on #3822 but it's almost ready for review. Could you please help review this draft PR first?

Thanks in advance.

monai/bundle/utils.py
```diff
@@ -0,0 +1,141 @@
...
+            https://pyyaml.org/wiki/PyYAMLDocumentation.
+
+    """
+    with open(filepath) as f:
+        if filepath.lower().endswith(".json"):
+            return json.load(f, **kwargs)
+        if filepath.lower().endswith((".yml", ".yaml")):
+            if "Loader" not in kwargs:
+                kwargs["Loader"] = yaml.FullLoader
+            return yaml.load(f, **kwargs)
```
ericspod, line 42: Should use `yaml.safe_load` if we can, we're not using YAML's ability to construct Python objects with `!!python/object` syntax so it's safer to disallow it unless that creates issues.

monai/bundle/utils.py
```diff
@@ -0,0 +1,141 @@
...
+    if len(result) != 1:
+        raise ValueError(f"path should only contain 1 file, but got: {path}.")
+
+    # split the path into filepath and target id of the content
+    paths = path.split(result[0])
+    parser = ConfigParser(config=load_config_file(paths[0] + result[0], **kwargs))
+    return parser[paths[1][1:] if paths[1] != "" else ""]
+
+
+def parse_id_value(pair: str) -> Tuple[str, Any]:
```
ericspod, line 71: Maybe add another function `parse_id_dict` to use where a dictionary is being created:

```python
def parse_id_dict(pairs: Optional[Sequence[str]]) -> Dict[str, Any]:
    if pairs is None:
        return {}
    return dict(map(parse_id_value, pairs))
```

tests/test_bundle_run.py
```diff
@@ -0,0 +1,136 @@
...
+            overridefile1 = os.path.join(tempdir, "override1.JSON")
+            with open(overridefile1, "w") as f:
+                # test override with part of the overriding file
+                json.dump({"move_net": "$@network_def.to(@device)"}, f)
+            overridefile2 = os.path.join(tempdir, "override2.JSON")
+            with open(overridefile2, "w") as f:
+                # test override with the whole overriding file
+                json.dump("Dataset", f)
+
+            ret = os.system(
```
ericspod, line 124: `os.system` isn't a reliable function to use because it depends on creating a new shell and the environment that has isn't well controlled. This will likely fail in certain Anaconda environments for example. It will also be sensitive to how the command is parsed which will vary by platform. I would recommend `subprocess.check_call` with `sys.executable` as the path to the interpreter itself so that you don't have to worry about what `python` is aliasing.

monai/bundle/run.py
```diff
@@ -0,0 +1,55 @@
...
+        help=("ID name of the target workflow, it must have the `run` method, follow MONAI `BaseWorkflow`."),
+        required=True,
+    )
+
+    args = parser.parse_args()
+    logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
+    override = {}
+    if args.override is not None:
+        for pair in args.override:
```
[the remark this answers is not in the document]
Nic-Ma: Sure, updated in the latest commit.

Thanks.

monai/bundle/utils.py
Nic-Ma: Sure, updated in the latest commit.

Thanks.

monai/bundle/utils.py
Nic-Ma: Good point, added in the latest commit.

Thanks.

tests/test_bundle_run.py
Nic-Ma: Good point, updated in the latest commit.

Thanks.

Nic-Ma:
Hi @ericspod @wyli ,

I updated the PR for:
1. Removed the CLI entry, I think we don't need it so far, users can leverage many out-of-the-box CLI tool.
2. Added unit tests to leverage `google fire` CLI tool. It can't parse complicated args like "AA#BB#<args>" directly, so I still kept the `id_value_to_dict` utility function.
3. Added support to define the `run` args in a JSON or YAML file, can help simplify the command, refer to MLFlow: https://www.mlflow.org/docs/latest/projects.html#mlproject-file.
4. Changed file name to `scripts.py`, will also add the `schema verification` logic here in PR https://github.com/Project-MONAI/MONAI/pull/3865.

Do you think there is any missing features for the regular train / evaluate / inference `run`?

Thanks in advance.

wyli:
thanks, it looks good, one minor point for the command line, is it possible to have `python -m monai.bundle run` as the primary interface, and make "fire" an internal and optional API?

setup.cfg
```diff
@@ -50,6 +50,8 @@ all =
     mlflow
     matplotlib
     tensorboardX
+    pyyaml
+    fire
```
Nic-Ma, line 54: Hi @wyli ,

I only use `fire` in the unit tests, is it necessary to add `fire` in this `setup.cfg` to make CI pass?

Thanks.

wyli: this is for `pip install monai[fire]`, if this pip command is required then this .cfg is correct

Nic-Ma:
> thanks, it looks good, one minor point for the command line, is it possible to have `python -m monai.bundle run` as the primary interface, and make "fire" an internal and optional API?

Thanks for your review.
Sure, added support to use `monai.bundle.scripts` as entry with `fire` directly, and added the unit tests to cover it.
It still also support to call the `run()` method in other CLI tool.

Thanks.

tests/test_bundle_run.py
Nic-Ma: Hi @ericspod ,

When I changed to `subprocess`, the windows CI test can't pass:
https://github.com/Project-MONAI/MONAI/runs/5394963470?check_suite_focus=true#step:10:579
Maybe we need another command for windows platform?
```
ERROR: test_shape_0 (tests.test_bundle_run.TestBundleRun)
  File "C:\hostedtoolcache\windows\Python\3.8.10\x64\lib\site-packages\parameterized\parameterized.py", line 533, in standalone_func
  File "D:\a\MONAI\MONAI\tests\test_bundle_run.py", line 134, in test_shape
    ret = subprocess.check_call(
  File "C:\hostedtoolcache\windows\Python\3.8.10\x64\lib\subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command 'C:\hostedtoolcache\windows\Python\3.8.10\x64\python.exe -m monai.bundle.scripts run --meta_file=C:\Users\RUNNER~1\AppData\Local\Temp\tmpnlqqp0ze\meta.yaml --config_file=C:\Users\RUNNER~1\AppData\Local\Temp\tmpnlqqp0ze\config.json --override='{postprocessing#<args>#transforms#2#<args>#output_postfix: seg, network: <file>C:\Users\RUNNER~1\AppData\Local\Temp\tmpnlqqp0ze\override1.JSON#move_net, dataset#<name>: <file>C:\Users\RUNNER~1\AppData\Local\Temp\tmpnlqqp0ze\override2.JSON}' --target=evaluator' returned non-zero exit status 1.
```

Thanks.

Nic-Ma: I skipped the `subprocess` test for windows.

Thanks.

Nic-Ma:
Hi @ericspod @wyli ,

I tried to define the abstract class for scripts, but seems we don't have enough usages so far, it's not clear and easy to extract the common features now. So I restored to put the essential features in functions directly, refer to MLFlow:
https://www.mlflow.org/docs/latest/_modules/mlflow/projects.html#run
Could you please help review the PR again?

Thanks in advance.

monai/bundle/utils.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    items = pair.split(":")
+    # remove blanks around id
+    id = items[0].strip()
+    value: Union[str, int, float, bool] = ""
+    if len(items) > 1:
+        # rejoin the rest, and remove blanks around value
+        value = ":".join(items[1:]).strip()
+
+    # try to convert the correct data type
```
wyli, line 106: This if tree could be replaced by ast.literal_eval()?

tests/test_bundle_run.py
```diff
@@ -0,0 +1,181 @@
...
+import unittest
+
+import nibabel as nib
+import numpy as np
+import yaml
+from parameterized import parameterized
+
+from monai.transforms import LoadImage
+
+TEST_CASE_1 = [
```
wyli, line 26: Could you make this test case a JSON file in https://github.com/Project-MONAI/MONAI/tree/dev/tests/testing_data, perhaps adds a yaml version if needed...

monai/bundle/scripts.py
```diff
@@ -0,0 +1,88 @@
...
+
+def run(
+    meta_file: Optional[Union[str, Sequence[str]]] = None,
+    config_file: Optional[Union[str, Sequence[str]]] = None,
+    override: Optional[Union[Dict, str]] = None,
+    target: Optional[str] = None,
+    args_file: Optional[str] = None,
+):
+    """
+    Specify metadata file and config file to run a regular training or evaluation program.
```
wyli, line 28: I think we should move the examples to a readme file in the folder monai/bundle, with more examples and an overview of the syntax @, #

docs/source/bundle.rst
```diff
@@ -32,3 +32,21 @@ Model Bundle
...
+
+`Scripts`
+---------
+.. autofunction:: run
+
+`Utilities`
+-----------
+.. autofunction:: load_config_file
+
+.. autofunction:: load_config_file_content
```
wyli, line 44: Form a use's perspective I don't know the difference of load_config_file and load_config_file_content, parse_config_file. Could you revise all the function names in utilities if they are frequently used?

wyli (commented):
Thanks, I put some minor comments, I can help update the docstrings in this PR after revising and let you review it.

monai/bundle/utils.py
Nic-Ma: Totally removed these logic in the latest commit.
Let's rely on `fire` or other CLI tools directly.

Thanks.

tests/test_bundle_run.py
Nic-Ma: Added both JSON and YAML test files.

Thanks.

monai/bundle/scripts.py
Nic-Ma: Yeah, we definitely need to prepare a detailed document for all the features in `monai.bundle`.
Maybe start from @ericspod 's PR: https://github.com/Project-MONAI/MONAI/pull/3834

Thanks.

docs/source/bundle.rst
Nic-Ma: Hi @wyli ,

Thanks for your suggestion, that's a good point.
I totally refactored the config reading logic in the latest commit.

Thanks.

Nic-Ma:
Hi @wyli @ericspod ,

Thanks for your review and feedback, I totally updated the PR based on comments and other reference projects.
I agree we should write detailed user manual documentation later.

Thanks.

wyli:
`python -m monai.bundle run` works fine with the `__main__.py` file, there's a minor issue with `python -m monai.bundle run --help` described here https://github.com/google/python-fire/issues/76

tests/test_bundle_run.py
```diff
@@ -0,0 +1,82 @@
...
+            cmd += " --target evaluator"
+            ret = subprocess.check_call(cmd.split(" "))
+            self.assertEqual(ret, 0)
+            self.assertTupleEqual(saver(os.path.join(tempdir, "image", "image_seg.nii.gz")).shape, expected_shape)
+
+            # here test the script with `google fire` tool as CLI
+            cmd = f"{sys.executable} -m fire monai.bundle.scripts run --meta_file {meta_file}"
+            cmd += f" --config_file {config_file} --override {{'evaluator#<args>#amp':False,{override}}}"
+            cmd += " --target evaluator"
+            ret = subprocess.check_call(cmd.split(" "))
```
ericspod, line 76: I suspect using `split` is dividing the arguments incorrectly if there are spaces in any of the values you're passing in that should be kept as one item together. The `cmd` value should be built up as a list rather than as a string then broken up, so `cmd = [sys.executable, '"-m", "fire", "monai.bundle.scripts", "run", "--meta-file", meta_file, ...]`. This is one of those things that can sometimes vary by platform, paths in Windows have a tendency to have spaces in them more often than other OSs (ie. "Program Files") so splitting by space isn't reliable.

monai/bundle/scripts.py
```diff
@@ -0,0 +1,95 @@
...
+    .. code-block:: bash
+
+        # Execute this module as CLI entry:
+        python -m monai.bundle run --meta_file=<meta path> --config_file=<config path> --target=trainer
+
+        # Override some config values at runtime, set `override` as a dict:
+        python -m monai.bundle run --override='{"net#<args>#ndims": 2}' ...
+
+        # Override some config values at runtime:
+        python -m monai.bundle run --"net#<args>#input_chns" 1 ...
```
wyli, line 42: Hi @Nic-Ma, I just added `**kwargs` to support this use case of `python -m monai.bundle run --"net#<args>#input_chns" 1`, do you have any concerns? also I'm not sure about `override: Optional[Union[Dict, str]]`, when is `str` input useful?

Nic-Ma: `override` should not be `str`, that's outdated code and I forgot to update the typehints.

Thanks.

monai/bundle/config_parser.py
```diff
@@ -170,38 +170,27 @@ def set(self, config: Any, id: str = ""):
...
-                data = deepcopy(parser[id])
-            else:
-                # id is in another config file
-                parser = ConfigParser(config=ConfigReader.load_config_file(paths[0]))
-                data = parser[paths[1][len(self.ref_resolver.sep) :] if paths[1] != "" else ""]
-            # recursively check the resolved content
-            return self._do_resolve(data)
+            path, ids = ConfigReader.split_path_id(config[len(self.macro) :])
+            parser = ConfigParser(config=self.get() if not path else ConfigReader.load_config_file(path))
+            return self._do_resolve(config=deepcopy(parser[ids]))
```
Nic-Ma, line 186: Hi @wyli ,

I wrote `data=...` separately in the previous code because we don't need `deepcopy` if the config is from file.
But anyway, I think it's not a problem to copy a small config content to simplify the code.

Thanks.

monai/bundle/config_reader.py
```diff
@@ -24,123 +25,118 @@
...
-            return None
-        items = src.split(result[0])
-        # return file path and the rest
-        return items[0] + result[0], items[1]
+        path, *ids = f"{src}".rsplit(ReferenceResolver.sep, 1)
+        ids = ids[0] if ids else ""
+        path_name = cls.path_match.findall(path)
+        if not path_name:
+            return "", src  # the src is a pure id
+        if len(path_name) < 1 and len(path_name[0]) < 1:
```
Nic-Ma, line 116: I don't think this `if` will hapen, maybe we can simplify the logic to?
```py
        if not path_name:
            return "", src  # the src is a pure id
        return path_name[0][0], ids
```

Thanks.

monai/bundle/scripts.py
Nic-Ma: Hi @wyli ,

I understand your proposal. But I feel it's a little bit confusing to have both `override` and `kwargs` args, maybe we can simply keep only one of them?
If only keeps `kwargs`, it can't support the case that the overriding keys happen to be "target", "config_file", "meta_file" or "args_file". And we may have more similar issues in other future scripts.
And may also consider supporting `kwargs` in the `args_file` file?
What do you think?

Thanks in advance.

monai/bundle/config_reader.py
```diff
@@ -0,0 +1,157 @@
...
+    def split_path_id(cls, src: str) -> Tuple[str, str]:
+        """
+        Split `src` string into two parts: a config file path and component id.
+        The file path should end with `(json|yaml|yml)`. The component id should be separated by `#` if it exists.
+
+        Args:
+            src: source string to split.
+
+        """
+        path, *ids = f"{src}".rsplit(ReferenceResolver.sep, 1)
```
Nic-Ma, line 113: Hi @wyli ,

I feel this logic may not be correct, the `src` can be something like:
`"~/data/nic#work/config.json#net#<args>#channels"`
Maybe we should use `cls.path_match.findall` directly on `src`?

Thanks.

monai/bundle/config_reader.py
wyli: It'll be here for some corner cases for 'src'

monai/bundle/config_reader.py
wyli: Sure, I feel we need a refactoring for this to unify the ref/macro/I'd syntax, could you work on this?

monai/bundle/scripts.py
wyli: I reviewed the usage and found that providing a dict input at CLI is not that straightforward in terms of syntax '{"net#<args>#ndims": 2}'. The kwargs would make it simpler, I think we should keep it. Perhaps we remove 'override'? Please help look into this,(sometimes it's not override the existing options but adding new options?)

Nic-Ma: OK, so I think we need to change `target` arg to other names like `runner_name`, `target` is slightly easier to conflict with some keys in the config content.

Thanks.

monai/bundle/config_reader.py
Nic-Ma: Sure, updated in the latest commit.

Thanks.

monai/bundle/config_reader.py
Nic-Ma: I updated this function logic in the latest commit.

Thanks.

wyli (approved):
thanks @Nic-Ma, this looks nice now and I'll revise and merge it soon.

ericspod (approved):

Nic-Ma:
Hi @ericspod @wyli ,

Thanks for your review and updates.
I am trying to unify the `_name_` and `_path_` key words to `_target_`, similar to Hydra:
https://github.com/facebookresearch/hydra/blob/main/examples/instantiate/object/conf/db/mysql.yaml#L1

Thanks.

monai/bundle/config_item.py
```diff
@@ -308,7 +310,7 @@ class ConfigExpression(ConfigItem):
 
     """
 
-    prefix = "$"
+    prefix = EXPR_KEY
```
Nic-Ma, line 313: Hi @wyli ,

If already defined the key in `utils`, do we still need this `prefix` member variable?
Same in all the other places.

Thanks.

monai/bundle/scripts.py
```diff
@@ -0,0 +1,112 @@
...
+    k_v = zip(["meta_file", "config_file", "target_id"], [meta_file, config_file, target_id])
+    for k, v in k_v:
+        if v is not None:
+            override[k] = v
+
+    full_kv = zip(
+        ("meta_file", "config_file", "target_id", "args_file", "override"),
+        (meta_file, config_file, target_id, args_file, override),
+    )
+    print("\n--- input summary of monai.bundle.scripts.run ---")
```
Nic-Ma, line 90: Hi @wyli ,

Thanks for adding more printing information.
I feel it's better to leverage `logging` instead of `print` because users may trigger these scripts in complicated environments like FL, AutoML, etc.
Could you please help change it similar to:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/image_writer.py#L34?

Thanks.

monai/bundle/scripts.py
```diff
@@ -0,0 +1,112 @@
...
+    for name, val in full_kv:
+        print(f"> {name}: {pprint.pformat(val)}")
+    print("---\n\n")
+
+    _args = _update_default_args(args=args_file, **override)
```
Nic-Ma, line 95: Hi @wyli ,

To print the run information, I think it's better to print the updated `_args` here instead of the `(meta_file, config_file, target_id, args_file, override)`.
What do you think?

Thanks.

Nic-Ma (commented):
Hi @wyli ,

Thanks for your update, I put some minor comments inline, others look good to me.

PR #3834, opened 2022-02-21
MONAI Archive Specification

ericspod:
### Description
This is the document setting out the specification of the MONAI Archive format for portable self-descriptive models. This is currently very minimal compared to what has been prototyped already by @Nic-Ma and discussed in #3482 and elsewhere. This PR is for discussion of what the format specification should be, what should be added or changed, how this does or does not meet the needs of the other subprojects, how code will use and interact with archive models, and how code can be generated based off the metadata included with the models. 

### Status
**Work in progress**

### Types of changes


ericspod:
This metadata specification varies slightly from the definition given in the schema given by @Nic-Ma [here](https://github.com/Project-MONAI/MONAI/blob/3049e280f2424962bb2a69261389fcc0b98e0036/monai/apps/mmars/schema/metadata.json). The example JSON metadata file has "channel_def" values mapping channel indices to descriptions of what the channels mean, use "is_patch_data" to state the data is not patch-wise but the whole image, and a number of the provided tags are considered optional or user-defined.

Nic-Ma:
The metadata change looks good to me.
My next PR  by chance will be `metadata` verification and `network` input / output verification.

Thanks.

docs/source/mar_specification.rst
```diff
@@ -0,0 +1,137 @@
...
+  ┃  ┣━ model.pt
+  ┃  ┗━ model.ts
+  ┗━ docs
+     ┣━ README.md
+     ┗━ license.txt
+
+
+These files mostly are required to be present with the given names for the directory to define a valid MAR:
+
+* **metadata.json**: netadata information in JSON format relating to the type of model, definition of input and output tensors, versions of the model and used software, and other information described below.
```
CPBridge, line 31:
```suggestion
* **metadata.json**: metadata information in JSON format relating to the type of model, definition of input and output tensors, versions of the model and used software, and other information described below.
```

docs/source/mar_specification.rst
```diff
@@ -0,0 +1,137 @@
...
+* **authorship**: state author(s) of the model.
+* **copyright**: state model copyright.
+* **network_data_format**: defines the format, shape, and meaning of inputs and outputs to the model, contains keys "inputs" and "outputs" relating named inputs/outputs to their format specifiers (defined below).
+
+Tensor format specifiers are used to define input and output tensors and their meanings, and must be a dictionary containing at least these keys:
+
+* **type**: what sort of data the tensor represents: "image", "label", etc.
+* **format**: what format of information is stored: "magnitude", "hounsfield", "kspace", "segmentation", "multiclass", etc.
+* **num_channels**: number of channels the tensor has, assumed channel dimension first.
+* **spatial_shape**: shape of the spatial dimensions of the form "[H]", "[H, W]", or "[H, W, D]"
```
CPBridge, line 65: I am concerned that in some cases this will be too restrictive. I have a few models that do not have a fixed input shape, but may have restrictions on the input shape, e.g. `[32n, 32n, 2n]` for something that must be as multiple 32 down the first two axes and a multiple of 2 down the last.

docs/source/mar_specification.rst
```diff
@@ -0,0 +1,137 @@
...
+* **task**: plain-language description of what the model is meant to do.
+* **description**: longer form plain-language description of what the model is, what it does, etc.
+* **authorship**: state author(s) of the model.
+* **copyright**: state model copyright.
+* **network_data_format**: defines the format, shape, and meaning of inputs and outputs to the model, contains keys "inputs" and "outputs" relating named inputs/outputs to their format specifiers (defined below).
+
+Tensor format specifiers are used to define input and output tensors and their meanings, and must be a dictionary containing at least these keys:
+
+* **type**: what sort of data the tensor represents: "image", "label", etc.
+* **format**: what format of information is stored: "magnitude", "hounsfield", "kspace", "segmentation", "multiclass", etc.
```
CPBridge, line 63: What is `mulitclass`? I'm guessing this is a array of integers where each integer represents a class index. In this case, I would personally refer to this as a "label map"

docs/source/mar_specification.rst
```diff
@@ -0,0 +1,137 @@
...
+metadata.json File
+==================
+
+This file contains the metadata information relating to the model, including what the shape and format of inputs and outputs are, what the meaning of the outputs are, what type of model is present, and other information. The JSON structure is a dictionary containing a defined set of keys with additional user-specified keys. The mandatory keys are as follows:
+
+* **version**: version of the stored model.
+* **monai_version**: version of MONAI the MAR was generated on, later versions expected to work.
+* **pytorch_version**: version of Pytorch the MAR was generated on, later versions expected to work.
+* **numpy_version**: version of Numpy the MAR was generated on, later versions expected to work.
+* **optional_packages_version**: dictionary relating optional package names to their versions, these packages are not needed but are recommended to be isntalled with this stated minimum version.
```
CPBridge, line 53:
```suggestion
* **optional_packages_version**: dictionary relating optional package names to their versions, these packages are not needed but are recommended to be installed with this stated minimum version.
```

CPBridge: Also what about required dependencies (apart from torch, monai, numpy)?

CPBridge (commented):
Overall looks good but I have some minor suggestions

docs/source/mar_specification.rst
ericspod: The idea with that was to allow multiple categories for each voxel, so multi-labeling is assigning a label to each voxel as a number, to have multi-class requires keeping N channels for N classes to handle voxels that would be referenced in multiple channels.

docs/source/mar_specification.rst
ericspod: Those would go into `optional_packages_version` so maybe call this `other_package_versions` instead.

docs/source/mar_specification.rst
GreySeaWolf: How about allowing "any size".  That might be the case while dealing with Digital Pathology Slides...

Nic-Ma: Hi @GreySeaWolf , may I know what's the `any size` you mean? Can we use "*" to represent it as @ericspod described?

Thanks.

docs/source/mar_specification.rst
Nic-Ma: Hi @ericspod , I would suggest `optional_packages_version` or `optional_dependencies_version`, to be consistent with our installation guide doc:

Thanks.

Nic-Ma:
Hi @ericspod @wyli ,

I found another problem from NVIDIA Clara users: Actually, developer of MMAR may be not very clear about what optional packages are used in this MMAR, especially when using MONAI docker.
I think we may need to provide some utily API to easily get all the optional packages used in a MMAR?
What do you think?

Thanks in advance.

vikashg:
@Nic-Ma and @ericspod. If we are thinking about backwards compatibility with CLARA and tensorrtserver (specially TensorRTServer), we need to include the names of the input and output nodes of the network . The names are important when specifying the config.pbtxt file used in TensorRTServer; Please find below an example config.pbtxt 
```
name: "segmentation_mri_brain_tumors_br16_t1c2tc_v1"
platform: "tensorflow_graphdef"
max_batch_size: 32
input [
  {
    name: "NV_MODEL_INPUT"
    data_type: TYPE_FP32
    dims: [ 1, 224, 224, 128]
  }
]
output [
  {
    name: "NV_MODEL_OUTPUT"
    data_type: TYPE_FP32
    dims: [ 2, 224, 224, 128]
  }
]
instance_group [
  {
    kind: KIND_GPU
    count: 2
  }
]
```
Please let me know if you have questions.

Nic-Ma:
Hi @vikashg ,

Thanks for your sharing.
I think we don't need to consider the backwards compatibility with CLARA and tensorrtserver so far.
Maybe we can raise this discussion in future versions.
@ericspod Please feel free to correct me if I misunderstood anything.

Thanks.

Nic-Ma:
Hi @ericspod ,

I developed the full schema for the metadata example in this PR, and put the schema on our repo：
https://github.com/Project-MONAI/MONAI-extra-test-data/releases/tag/0.8.1
This PR shows the draft API and usage: https://github.com/Project-MONAI/MONAI/pull/3865.

Thanks.

vikashg:
@Nic-Ma I am unaware if there is something that has changed in TRT server. But they do need the node names. Yes lets discuss it in the meeting. I think it should be included so we can also deploy the models.

docs/source/mar_specification.rst
```diff
@@ -0,0 +1,137 @@
...
+
+This specification defines the directory structure a MAR must have and the necessary files it must contain. Additional files may be included and the directory packaged into a zip file or included as extra files directly in a Torchscript file.
+
+Directory Structure
+===================
+
+A MAR package is defined primarily as a directory with a set of specifically named subdirectories containing the model and metadata files. The root directory should be named for the model, given as "ModelName", and should contain the following structure:
+::
+  ModelName
```
MMelQin, line 18: Should we call out the version of the model? I have seen use cases requiring multiple versions of the same model, and the Triton inference supports multiple versions of the same named model.

Nic-Ma: Hi @MMelQin ,

I think we already included the `version` and `changelog` of the model package:
https://github.com/Project-MONAI/MONAI/pull/3834/files#diff-139c21b7be482dc057c2b0b9d131f86500621c01b7132cb15060c54b2fb22775R84
@ericspod Maybe you can update the description to make it more clear?

Thanks in advance.

docs/source/mar_specification.rst
Nic-Ma: Hi @ericspod ,

I got a question here: What's the expected data type of `spatial_shape` in the metadata? Seems you are using `str` in this example. We need to finalize the data type then check it with the schema. Otherwise, it's hard to check network input / output shape with scripts.
What do you think?

Thanks.

Nic-Ma:
> 

Hi @ericspod @wyli ,

What do you guys think about this problem?

Thanks in advance.

Nic-Ma:
> Hi @ericspod @wyli ,
> 
> I found another problem from NVIDIA Clara users: Actually, developer of MMAR may be not very clear about what optional packages are used in this MMAR, especially when using MONAI docker. I think we may need to provide some utily API to easily get all the optional packages used in the related MONAI python code of a MMAR? What do you think?
> 
> Thanks in advance.

Hi @ericspod @wyli ,

What do you guys think about this problem?

Thanks in advance.

docs/source/mar_specification.rst
ericspod: Shapes are going to have to be defined as lists of either strings to accommodate "2n" or "*" as specifiers or integers for exact know sizes. For example, a 2D image that's 256x256 in the spatial dimension with any number of channels would be `["*", 256, 256]` by if the spatial dimensions had to be multiples `n` of a power `p` of 2 we would have `["*", "2**p*n", "2**p*n"]`. The expression syntax we would allow here may get a bit complicated and requires validation in separate steps.

ericspod:
> Hi @ericspod @wyli ,
> 
> I found another problem from NVIDIA Clara users: Actually, developer of MMAR may be not very clear about what optional packages are used in this MMAR, especially when using MONAI docker. I think we may need to provide some utily API to easily get all the optional packages used in the related MONAI python code of a MMAR? What do you think?
> 
> Thanks in advance.

Something like [pipreqs](https://github.com/bndr/pipreqs) but with a project or network being bundled being considered the project.

ericspod:
There's also [pipdeptree](https://github.com/naiquevin/pipdeptree) that will show a user what dependencies there are for a given root directory.

wyli (approved):
thanks, looks good to me as a starting point for the monai bundle concepts

docs/source/mb_specification.rst
```diff
@@ -0,0 +1,143 @@
...
+              }
+          },
+          "outputs":{
+              "pred": {
+                  "type": "image",
+                  "format": "segmentation",
+                  "num_channels": 2,
+                  "spatial_shape": [160, 160, 160],
+                  "dtype": "float32",
+                  "value_range": [0, 1],
```
Nic-Ma, line 137: Hi @ericspod ,

Some internal researchers raised a question about this "outputs" sections, is it exactly the output of network or output after postprocessing? If just output of network, the value range is not [0, 1].

Thanks.

Nic-Ma: I see, so do you want any change in this metadata config?

Thanks.

ericspod: We should discuss what this should be first if we need to go deeper into specifying things like I say, then we'll have an idea of what to put here. I'm afraid we're encountering more details we need to address as we go.

PR #3833, opened 2022-02-21
[3823] ResampleToMatch

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/3823.

### Description
Resample an image (or images) to match a given meta data by aligning the affine matrix and ensuring that the output image is of the same dimensions.

### Status
**Ready**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [x] New tests added to cover the changes.
- [x] Documentation updated, tested `make html` command in the `docs/` folder.

PR #3847, opened 2022-02-23
enhance to use spatial_shape ResampleToMatch

wyli:

follow up of https://github.com/Project-MONAI/MONAI/pull/3833

### Description
spatial_shape is a more reliable attribute from the readers for resampletomatch

https://github.com/Project-MONAI/MONAI/blob/127e823dc52224e3b201679b16c5d76e9c57e2ab/monai/data/image_reader.py#L266
### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -293,13 +293,11 @@ class ResampleToMatch(SpatialResample):
         dtype = dtype or self.dtype
         src_affine = src_meta.get("affine")
         dst_affine = dst_meta.get("affine")
-        ndim = len(img.shape[1:])
-        spatial_size = dst_meta.get("dim", [])[1 : ndim + 2]
         img, updated_affine = super().__call__(
             img=img,
             src_affine=src_affine,
             dst_affine=dst_affine,
-            spatial_size=spatial_size,
+            spatial_size=dst_meta.get("spatial_shape"),
             mode=mode,
             padding_mode=padding_mode,
             align_corners=align_corners,
@@ -307,6 +305,8 @@ class ResampleToMatch(SpatialResample):
         )
         dst_meta = deepcopy(dst_meta)
         dst_meta["affine"] = updated_affine
+        if "spatial_shape" in dst_meta:
+            dst_meta["spatial_shape"] = src_meta.get("spatial_shape")
         if "dim" in dst_meta:
             dst_meta["dim"] = src_meta.get("dim", [])
         if "pixdim" in dst_meta:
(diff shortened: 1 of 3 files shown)


monai/transforms/spatial/array.py
```diff
@@ -293,24 +293,18 @@ def __call__(  # type: ignore
...
             padding_mode=padding_mode,
             align_corners=align_corners,
             dtype=dtype,
         )
         dst_meta = deepcopy(dst_meta)
         dst_meta["affine"] = updated_affine
-        if "dim" in dst_meta:
-            dst_meta["dim"] = src_meta.get("dim", [])
-        if "pixdim" in dst_meta:
-            dst_meta["pixdim"] = src_meta.get("pixdim", [])
```
rijobro, line 313: If these aren't updated, then they won't be consistent with the new image surely?

rijobro: The unit test first saves the image and then loads it back in before checking the meta data is correct. This works because the savers will update the metadata, but really this should be updated by the transform.

wyli: that's a good point, the saver might have restored the resampling. I updated the tests to have the comparison with the resampled meta... https://github.com/Project-MONAI/MONAI/pull/3847/commits/12bb4a28672be301cce4d8017e6b5c13c278325a

so in my understanding
```py
ResampleToMatch()(data["im2"], 
                  src_meta=data["im2_meta_dict"], 
                  dst_meta=data["im1_meta_dict"])
```
the im2 will be processed and saved in `im1_meta_dict` for after this transform, perhaps you are thinking about resuming the im2_meta_dict at some point (when saving it)?

rijobro: The way I envisaged this transform is that `im2` would be resampled such that it is aligned with `im1_meta_dict`. The problem is that although we'll update the image, we don't update the meta data, which is what I was attempting to do with the 4 lines that have been removed. I don't think we should be doing anything like `im2_meta_dict = deepcopy(im1_meta_dict)` because it's only the spatial aspects that have been updated.

rijobro: Just added some tests to your PR and the meta data seems to be being updated so all good! Sorry about that.

wyli: thanks, no worries, actually it's a good point for discussion,
monai reader/writer specifically handles `original_affine` and that's the 'raw' affine
https://github.com/Project-MONAI/MONAI/blob/d6327a765cf23dafd680b2a4b472edf9639e1087/monai/data/image_reader.py#L264

https://github.com/Project-MONAI/MONAI/blob/d6327a765cf23dafd680b2a4b472edf9639e1087/monai/data/image_writer.py#L324

now after `ResampleToMatch` not sure what to put at `original_affine`. this request seems to want to avoid the original affines ones https://github.com/Project-MONAI/MONAI/discussions/3842

rijobro (approved):

PR #3860, opened 2022-02-25
3854 Add `detach` API to all the MONAI transforms

### LINKED ISSUE #3854: Add `detach` logic to handlers
Currently, our handlers have the `attach` method but don't have the logic to detach from the engine.
Refer to the ignite `detach` API:
https://github.com/pytorch/ignite/blob/master/ignite/metrics/metric.py#L411

Would be nice to support it.

Nic-Ma:
Fixes #3854 .

### Description
This PR added `detach` API to all the MONAI handlers to support runtime update, be consistent with ignite:
https://github.com/pytorch/ignite/blob/master/ignite/metrics/metric.py#L411
And added unit tests to cover them.

### Status
**Ready**

### Types of changes


wyli (changes requested):
This is a major revision that requires a proposal and discussion s as it change the way we use engine and handles.

ericspod:
The `Engine.add_event_handler` method returns a `RemovableEventHandle` object used to detach the handler, perhaps this should be used?

Nic-Ma:
Hi @vfdev-5 ,

I checked the source code of `RemovableEventHandle` in ignite, but I didn't find a typical example exactly based on `attach / detach / RemovableEventHandle` logic, something similar to:
https://github.com/pytorch/ignite/blob/master/ignite/contrib/handlers/base_logger.py#L190
Could you please help share some example or comments on this topic? In our case, we may attach several methods to several events in the `attach` API, what's the best way to implement `detach`?

Thanks in advance.

wyli:
Could you please create a proposal for adding detach for all the handlers? I don't see the use cases here. Providing such functionality can be confusing the engines lifecycle. This is not our high priority milestones for v1.0. I'd suggest the core dev team should focus on the key tasks.

vfdev-5:
Hi @Nic-Ma 

Here is an example of how to use `RemovableEventHandle` as a context manager:
```python
engine = Engine(...)

def print_epoch(engine):
    print(f"Epoch: {engine.state.epoch}")

with engine.add_event_handler(Events.EPOCH_COMPLETED, print_epoch):
    # print_epoch handler registered for a single run
    engine.run(data)

# print_epoch handler is now unregistered
```
- source: https://github.com/pytorch/ignite/blob/5bccc154a776a4af6226b20447b1962886df6393/ignite/engine/events.py#L429-L484
- docs : https://pytorch.org/ignite/v0.4.8/generated/ignite.engine.events.RemovableEventHandle.html#removableeventhandle

Essentially, `RemovableEventHandle.remove()` is responsible for removing the handler from engine.

Another example can be found [here](https://pytorch.org/ignite/v0.4.8/concepts.html#events-and-handlers), see "Event handlers can be detached via remove_event_handler or via the RemovableEventHandle reference."
```python
def log_metrics(engine, title):
    print(f"Epoch: {trainer.state.epoch} - {title} accuracy: {engine.state.metrics['acc']:.2f}")

@trainer.on(Events.EPOCH_COMPLETED)
def evaluate(trainer):
    with evaluator.add_event_handler(Events.COMPLETED, log_metrics, "train"):
        evaluator.run(train_loader)

    with evaluator.add_event_handler(Events.COMPLETED, log_metrics, "validation"):
        evaluator.run(validation_loader)

    with evaluator.add_event_handler(Events.COMPLETED, log_metrics, "test"):
        evaluator.run(test_loader)
```

Hope this helps.

> In our case, we may attach several methods to several events in the attach API, what's the best way to implement detach?

Maybe, a solution could be to store all returned instances of `RemovableEventHandle` when adding removable handlers and when they should be removed, call `remove()` method on all stored objects...

Nic-Ma:
Thanks for @vfdev-5 's detailed examples and suggestions.

@wyli @ericspod , how about removing the `detach` API in this PR, just try to save the `RemovableEventHandle` in every handler when attaching it? Then if some outside logic wants to detach the handler, it can call the `remove()` method of every `RemovableEventHandle` as @vfdev-5 suggested.
I updated `SmartCacheHandler` in the latest commit to show the idea, what do you think?
```py
class SmartCacheHandler:
    def __init__(self, smartcacher: SmartCacheDataset) -> None:
        if not isinstance(smartcacher, SmartCacheDataset):
            raise TypeError("smartcacher must be a monai.data.SmartCacheDataset.")
        self.smartcacher = smartcacher
        self.removables: List = []

    def attach(self, engine: Engine) -> None:
        self.removables = []
        self.removables.append(engine.add_event_handler(Events.STARTED, self.started))
        self.removables.append(engine.add_event_handler(Events.EPOCH_COMPLETED, self.epoch_completed))
        self.removables.append(engine.add_event_handler(Events.COMPLETED, self.completed))

    def started(self, engine: Engine) -> None:
        self.smartcacher.start()

    def epoch_completed(self, engine: Engine) -> None:
        self.smartcacher.update_cache()

    def completed(self, engine: Engine) -> None:
        self.smartcacher.shutdown()
```

Thanks in advance.

wyli:
Before we discuss how to implement it, we should first discuss whether it is needed to have this feature for all the handlers. That's why we need a proposal. Let's not skip the requirement analysis.

ericspod:
We should discuss the need, but would do it this way with a `removables` list and a method `detach` that calls `remove` on each. The `removables` attribute and `detach` method should be in a root `Handler` class all the others inherit so that `detach` need be defined only once.

yhwen:
We have a use case in FL cross-validation, which we don't want to use the TensorBoardStatsHandler to write the TB events any more. We ended up with creating a separate engine with its own different handlers. If the "detach" functionality is available, we could use this "detach" to manipulate which handlers to use and reuse the same engine. 

Also to have both "attach" and "detach" functions in pair seems to logical. And it will be useful.

wyli:
It might be useful for a few cases for the tensorboard or for some heavy lifting metrics, I don't understand why we create the API for all the handlers and to detach all the events in the handlers. To me a simple utility `def detach(engine, handler, event_type)` could be sufficient. but this PR changes 19 different event handlers. @yhwen

vfdev-5:
My 2 cents about removing attached handlers, there is [`Engine.remove_event_handler` method](https://pytorch.org/ignite/generated/ignite.engine.engine.Engine.html#ignite.engine.engine.Engine.remove_event_handler)

Nic-Ma:
Hi @yhwen ,

Thanks for your sharing, as you already ended up creating a separate engine, I will close this PR for now.
May consider to reopen it if any other related requirement is raised.

Thanks for the great discussion guys @ericspod @vfdev-5 @wyli .

PR #3871, opened 2022-03-01
3867 fix non intensity error

### LINKED ISSUE #3867: `RandKSpaceSpikeNoise` got an unexpected keyword argument 'dim'
**Describe the bug**
`RandKSpaceSpikeNoise` does not work with NumPy array if `intensity_range` is not set manually.

In case a NumPy array is provided, the linked part will call `np.mean(.., dim=...)` but NumPy does not support dim arguments (should be `axis`). The implementation works fine for `torch.Tensors`.

https://github.com/Project-MONAI/MONAI/blob/d6327a765cf23dafd680b2a4b472edf9639e1087/monai/transforms/intensity/array.py#L1816-L1818

**To Reproduce**
Call transform on NumPy array with default values (specifically `intensity_range =None`).

**Expected behavior**
No crash.

**Screenshots**

**Environment**

```
```

**Additional context**

yiheng-wang-nv:

Fixes #3867 .

### Description
This PR fixes the `_set_default_range` issue in class `RandKSpaceSpikeNoise`.

There are two errors in this function:

1. For numpy based img, `np.mean` does not have the arg `dim` (this bug is reported in #3867 , thanks to @mibaumgartner !)
2. the generated intensity range has the same type as `img`, if `img` is saved in GPU, error will be raised in the following `self.R.uniform` manipulation step.

In addition, this PR adds the test case (none intensity_range) for `RandKSpaceSpikeNoise`, and the test case (none k_intensity) for `KSpaceSpikeNoise`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/intensity/array.py
+++ b/monai/transforms/intensity/array.py
@@ -1813,9 +1813,11 @@ class RandKSpaceSpikeNoise(RandomizableTransform, Fourier):
         n_dims = len(img.shape[1:])
 
         k = self.shift_fourier(img, n_dims)
-        k, *_ = convert_data_type(k, output_type=np.ndarray)
-        log_abs = np.log(np.absolute(k) + 1e-10)
-        shifted_means = np.mean(log_abs, tuple(range(-n_dims, 0))) * 2.5
+        mod = torch if isinstance(k, torch.Tensor) else np
+        log_abs = mod.log(mod.absolute(k) + 1e-10)
+        shifted_means = mod.mean(log_abs, tuple(range(-n_dims, 0))) * 2.5
+        if isinstance(k, torch.Tensor):
+            shifted_means = shifted_means.to("cpu")
         return tuple((i * 0.95, i * 1.1) for i in shifted_means)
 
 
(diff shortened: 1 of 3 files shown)


monai/transforms/intensity/array.py
```diff
@@ -1813,9 +1813,10 @@ def _set_default_range(self, img: NdarrayOrTensor) -> Sequence[Sequence[float]]:
         n_dims = len(img.shape[1:])
 
         k = self.shift_fourier(img, n_dims)
+        k, *_ = convert_data_type(k, output_type=np.ndarray)
```
Nic-Ma, line 1816: May I know why convert `k` to numpy array here?

Thanks.

yiheng-wang-nv: If k is torch.tensor in gpu device, this line: https://github.com/Project-MONAI/MONAI/pull/3871/files#diff-83ae933a39467d71ab978c0bba9debaa2db45f24a74c127a664f1044fc0017dfR1784 will raise an error.

In addition, I think the generated intensity range should has the same type as the argument `intensity_range`:
Sequence[Union[Sequence[float], float]]

Nic-Ma (approved):
Hi @rijobro @wyli ,

Looks good to me, but I am not very familiar with the algorithm here, could you please help double confirm it?

Thanks in advance.

monai/transforms/intensity/array.py
yiheng-wang-nv: Thanks @Nic-Ma , I submitted a new commit, and put the convert step after calculation.

PR #3882, opened 2022-03-02
3881 using sform/qform code 1 for the writers

### LINKED ISSUE #3881: inconsistent output s/qform code for SaveImage
**Describe the bug**
```py
import numpy as np
import nibabel as nib
from monai.transforms import SaveImaged, LoadImaged
from monai.data import ITKWriter, NibabelWriter

# create a testing volume
img = nib.Nifti1Image(np.ones((3, 4, 5, 1, 1)), np.eye(4))
img.header.set_qform(np.eye(4), code=2)
nib.save(img, "test_qform.nii.gz")

# load/save test
data_in = LoadImaged(keys="img", reader="nibabelreader")({"img": "test_qform.nii.gz"})
SaveImaged(keys="img", output_postfix="nib", writer=NibabelWriter)(data_in)
SaveImaged(keys="img", output_postfix="itk", writer=ITKWriter)(data_in)

# print the saved
ori_hdr = nib.load("test_qform.nii.gz").header
print(ori_hdr['sform_code'], ori_hdr['qform_code'])
itk_hdr = nib.load("test_qform/test_qform_itk.nii.gz").header
print(itk_hdr['sform_code'], itk_hdr['qform_code'])
nib_hdr = nib.load("test_qform/test_qform_nib.nii.gz").header
print(nib_hdr['sform_code'], nib_hdr['qform_code'])
```

output:
```
2022-03-02 11:31:22,658 INFO image_writer.py:190 - writing: test_qform/test_qform_nib.nii.gz
2022-03-02 11:31:22,662 INFO image_writer.py:190 - writing: test_qform/test_qform_itk.nii.gz
2 3
1 1
2 0
```

**Expected behavior**
would be great to have a consistent output for

- `print(itk_hdr['sform_code'], itk_hdr['qform_code'])`
- `print(nib_hdr['sform_code'], nib_hdr['qform_code'])`

the itk writer is hard-coded:
https://github.com/InsightSoftwareConsortium/ITK/blob/15aa8d58a01a1279e1fafa2214c82edd80b17e92/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L2174-L2176

**Environment**

```
MONAI version: 0.8.1+147.g274769c8
Nibabel version: 3.2.1
```

wyli:
Fixes #3881 
Follow-up of https://github.com/Project-MONAI/tutorials/discussions/481

### Description
following https://github.com/InsightSoftwareConsortium/ITK/blob/15aa8d58a01a1279e1fafa2214c82edd80b17e92/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L2174-L2176

### Status
**Ready**

### Types of changes


Nic-Ma (approved):
Hi @thewtex @ericspod ,

I am not familiar with the ITK settings, could you please help double confirm it?

Thanks in advance.

wyli:
(also CC'ing @hjmjohnson in case there are more comments)

ericspod:
I'm not sure hard-coding these values is the right process, either we should be setting them to whatever is appropriate or inheriting whatever the original values were if present in metadata. ITK hard codes it but that's not necessarily the best thing to do, nor is inconsistent values in the header a problem so long as they are set correctly. Setting both these values to 1 selects transform method 3 (https://www.nitrc.org/docman/view.php/26/204/TheNIfTI, pages 12 and 13). If an input nifti uses another method then the header of the output image will be different but both ITK and our code will do this consistent with one another. If we did hard code the values as such we have to fill in the `srow_*` values correctly as well as `qoffset_*` and `quatern_*` values.

wyli:
> If we did hard code the values as such we have to fill in the `srow_*` values correctly as well as `qoffset_*` and `quatern_*` values.

this part is done by the `set_qform` and `set_sform` API, should be fine:
- https://github.com/nipy/nibabel/blob/75c2b85cadb3ef427167ac90a5c0860937cc9e75/nibabel/nifti1.py#L1068
- https://github.com/nipy/nibabel/blob/1eb21d8d05f599baf6a2d396f2404604124ef027/nibabel/nifti1.py#L935

hjmjohnson:
I will need some time to make a more detailed review.  I have found that all options are a catch-22 with nifti files.

The qform is lossy with respect to the dicom standard, and often results in numerical precision problems when dealing with physical space.  The sform has enough precision to be lossless with respect to the dicom standard, but does not necessarily impose that the representation has orthogonal axis.  

To further complicate matters, many external tools have developed different protocols and interpretations of when/how to use each of the storage format.

Even ITK's behavior has evolved over time trying to adapt to the environment. The most recent changes opted to try to be less lossy during data conversion so that physical space can be reliably represented with respect to the original dicom data.  This change occurred when nifti binary masks were not reported in the same physical space as the dicom originating space.  The problem turned out to be small errors in translation to/from qform representations.

dzenanz:
@hjmjohnson did you take a second look here?

hjmjohnson:
> @hjmjohnson did you take a second look here?

Yes.  I don't think there is a clear solution from ITK's standpoint.  ITK defaults to writing the sform content because it most closely matches the internal ITK storage representation, and it provides the least lossy path from DICOM representations of physical2index space mapping.

dzenanz:
So, approve PR?

hjmjohnson (approved):
I think setting the sform and qform codes to 1  is the safest way.

PR #3884, opened 2022-03-02
SplitDim

rijobro:
### Description
Split dimensions of an image (useful for 3d->2d conversion). updates the affine matrix, too.

Deprecates SplitChannel as SplitDim is the same but more general.

### Status
**Ready**

### Types of changes


ericspod (approved):

monai/transforms/utility/dictionary.py
```diff
@@ -415,9 +417,44 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> Dict[Hashable, N
...
+                if self.update_meta:
+                    orig_meta = d.get(PostFix.meta(key), None)
+                    if orig_meta is not None:
+                        split_meta_key = PostFix.meta(split_key)
+                        d[split_meta_key] = deepcopy(orig_meta)
+                        dim = self.splitter.dim
+                        if dim > 0:  # don't update affine if channel dim
+                            shift = np.zeros_like(d[split_meta_key]["affine"])  # type: ignore
+                            shift[dim - 1, -1] = i * d[split_meta_key]["pixdim"][dim]  # type: ignore
+                            d[split_meta_key]["affine"] += shift @ d[split_meta_key]["affine"]  # type: ignore
```
wyli, line 430: I think the shift should happen in the image space (untested):
```suggestion
                            shift = np.eye(len(d[split_meta_key]["affine"]))
                            shift[dim - 1, -1] = i
                            d[split_meta_key]["affine"] = d[split_meta_key]["affine"] @ shift
```

with the current PR's implementation when the axis is flipped in the physical<>image space transform, the shift should be -shift

wyli (approved):

monai/transforms/utility/array.py
```diff
@@ -281,33 +283,56 @@ def __call__(self, img: NdarrayOrTensor) -> NdarrayOrTensor:
...
+        dim: dimension on which to split
+        keepdim: if `True`, output will have singleton in the split dimension. If `False`, this
+            dimension will be squeezed.
     """
 
     backend = [TransformBackends.TORCH, TransformBackends.NUMPY]
 
-    def __init__(self, channel_dim: int = 0) -> None:
-        self.channel_dim = channel_dim
+    def __init__(self, dim: int = -1, keepdim: bool = True) -> None:
```
Nic-Ma, line 301: May I know why change the default value from `0` to `-1`?

Thanks.

rijobro: No particularly strong reason, as the default could be anything. I went for `-1` as I thought the most common use-case would be converting 3d to 2d, so splitting along the depth dimension.

Nic-Ma (approved):

monai/transforms/utility/array.py
Nic-Ma: OK, I see.
I feel `split channels` is also a very useful example, could you please help add some description for it in the "for example" of doc-string?

Thanks.

wyli:
thanks, looks all good, let me update the doc and merge this one.

PR #3885, opened 2022-03-02
support of writer name as option in SaveImage

wyli:
### Description
add support of
```py
SaveImaged(keys="img", output_postfix="itk", writer="ITKWriter")(data_in)
```
as a shortcut of
```py
from monai.data import ITKWriter
SaveImaged(keys="img", output_postfix="itk", writer=ITKWriter)(data_in)
```

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/io/array.py
+++ b/monai/transforms/io/array.py
@@ -19,6 +19,7 @@ import sys
 import traceback
 import warnings
 from pathlib import Path
+from pydoc import locate
 from typing import Dict, List, Optional, Sequence, Union
 
 import numpy as np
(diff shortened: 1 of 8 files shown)


monai/transforms/io/array.py
```diff
@@ -335,6 +337,8 @@ def __init__(
         )
 
         self.output_ext = output_ext.lower() or output_format.lower()
+        if isinstance(writer, str):
+            writer, _ = optional_import("monai.data", name=f"{writer}")
```
Nic-Ma, line 341: Maybe we can use `locate` or `importlib` to user extended writer of `ImageWriter`?
For example `writer="custom.writer.MyWriter"`.

Thanks.

wyli: thanks, i've updated to use two searches (is there a simpler approach?) https://github.com/Project-MONAI/MONAI/blob/9a474dabe47b0ebf1a2c06874d9aa832cf603fb4/monai/transforms/io/array.py#L341-L347

monai/transforms/io/array.py
```diff
@@ -335,6 +338,13 @@ def __init__(
         )
 
         self.output_ext = output_ext.lower() or output_format.lower()
+        if isinstance(writer, str):
+            writer_, has_built_in = optional_import("monai.data", name=f"{writer}")  # search built-in
+            if not has_built_in:
+                writer_ = locate(f"{writer}")  # search dotted path
```
Nic-Ma, line 344: Could you please help add a simple unit test to cover this feature?
I think it would be pretty useful for BYOC writers.

Thanks in advance.

Nic-Ma (approved):
Thanks for the quick update.
Looks good to me, put a minor comment inline.

Thanks.

monai/transforms/io/array.py
wyli: yes it's tested with the `"monai.data.NibabelWriter"` test case, this string works fine because of `locate(writer)`

PR #3897, opened 2022-03-07
update Transchex model

### LINKED ISSUE #3880: class MultiModal - updating in the for loop
**Describe the bug**
It looks like that in the 'class MultiModal(BertPreTrainedModel):' inside the 'forward' function in lines 240, 242, and 244, the variables are not getting updated. As the input argument, they receive the very initial input and not the output of the previous layer which is supposed to be the reason for using a 'for loop'.
Is that a bug, or is there something that I'm missing out on?

ahatamiz:
Signed-off-by: ahatamizadeh <ahatamizadeh@nvidia.com>

Fixes #3880  .

### Description
This PR patches the issue with updating the parameters in TransCheX model.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/networks/nets/transchex.py
+++ b/monai/networks/nets/transchex.py
@@ -185,12 +185,15 @@ class BertMixedLayer(nn.Module):
 
     def __init__(self, config) -> None:
         super().__init__()
-        self.att = BertAttention(config)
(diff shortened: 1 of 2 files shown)


wyli (approved):
thanks for the quick fix!

PR #3927, opened 2022-03-14
3482 Verify network input output shape

Nic-Ma:
Task step 6 of https://github.com/Project-MONAI/MONAI/issues/3482 .

### Description
This PR added the script to verify network input output data shape.

### Status
**Ready**

### Types of changes


monai/bundle/scripts.py
```diff
@@ -32,14 +38,49 @@ def _update_default_args(args: Optional[Union[str, Dict]] = None, **kwargs) -> D
...
+        p: multiply factor to generate fake data shape if dim of expected shape is "x*n", default to 1.
+        any: specified size to generate fake data shape if dim of expected shape is "*", default to 1.
+
+    """
+    ret = []
+    for i in shape:
+        if isinstance(i, int):
+            ret.append(i)
+        elif isinstance(i, str):
+            ret.append(any if i == "*" else eval(i, {"p": p, "n": n}))
```
ericspod, line 74: I was thinking that we would want to allow any arbitrary variables and not hard-code `m` and `p` as the only multiples and powers that we could have. It's gets very complicated what sort of expression we'd allow and then NP-hard to find values that satisfy a set of spatial size expressions, but I still think it's tractable to do it for our small amount of data. One thing we could use is the `ast` module to parse the expression and discover what variables are present, such as:

```python
import ast
def get_names(expr):
    tree=ast.parse(expr)
    names=[m.id for m in ast.walk(tree) if isinstance(m, ast.Name)]
    return names

get_names("2**p*m+1+2*v")  # returns ['v', 'm', 'p']
```

Nic-Ma: Added the check in the latest commit.

Thanks for the suggestion.

monai/bundle/scripts.py
```diff
@@ -52,12 +55,61 @@ def _update_args(args: Optional[Union[str, Dict]] = None, ignore_none: bool = Tr
 
 
 def _log_input_summary(tag: str, args: Dict):
+    """
```
wyli, line 58: The private functions might change

monai/bundle/scripts.py
```diff
@@ -52,12 +55,61 @@ def _update_args(args: Optional[Union[str, Dict]] = None, ignore_none: bool = Tr
...
+    Log the arguments of bundle scripts.
+
+    Args:
+        tag: tag to identify the script in the log.
+        args: arguments of the script to log.
+
+    """
```
wyli, line 65: Could you remove the first \n in line 66? The format of the first line doesn't look right after using logger info

monai/bundle/scripts.py
```diff
@@ -52,12 +55,61 @@ def _update_args(args: Optional[Union[str, Dict]] = None, ignore_none: bool = Tr
...
+    for i in shape:
+        if isinstance(i, int):
+            ret.append(i)
+        elif isinstance(i, str):
+            if i == "*":
+                ret.append(any)
+            else:
+                for c in _get_var_names(i):
+                    if c not in ["p", "n"]:
+                        raise ValueError(f"only support variables 'm' and 'p' so far, but got: {c}.")
```
wyli, line 106: The error msg has a typo m-> n?

monai/bundle/scripts.py
```diff
@@ -172,3 +224,88 @@ def verify_metadata(
...
+    p: Optional[int] = None,
+    n: Optional[int] = None,
+    any: Optional[int] = None,
+    args_file: Optional[str] = None,
+    **override,
+):
+    """
+    Verify the input and output data shape and data type of network defined in the metadata.
+    Will test with fake Tensor data according to the network args with id:
+    input data:
```
wyli, line 243: Could you remove the examples in line 234-249, please put one command line example instead

monai/bundle/scripts.py
```diff
@@ -172,3 +224,88 @@ def verify_metadata(
...
+    p = _args.pop("p", 1)
+    n = _args.pop("n", 1)
+    any = _args.pop("any", 1)
+
+    # the rest key-values in the _args are to override config content
+    for k, v in _args.items():
+        parser[k] = v
+
+    net = parser.get_parsed_content(id).to(device_)
+    input_channels = parser["_meta_#network_data_format#inputs#image#num_channels"]
```
wyli, line 295: If one of the field is not provided, it could be a warning instead of key error?

tests/testing_data/inference.json
```diff
@@ -1,5 +1,5 @@
 {
-    "device": "$torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')",
+    "device": "$torch.device('cuda' if torch.cuda.is_available() else 'cpu')",
```
wyli, line 2: The previous version will avoid unnecessary allocation of multi gpu memories, could we keep it?

wyli (commented):
Thanks, please see some comments inline

monai/bundle/scripts.py
Nic-Ma: Sure, removed in the latest commit.

Thanks.

monai/bundle/scripts.py
Nic-Ma: Sure, removed in the latest commit.

Thanks.

monai/bundle/scripts.py
Nic-Ma: Thanks for your finding, updated in the latest commit.

tests/testing_data/inference.json
Nic-Ma: updated in the latest commit.

Thanks.

monai/bundle/scripts.py

Thanks.

monai/bundle/scripts.py
Nic-Ma: Thanks for your suggestion.
I think we should raise an exception to stop the verification if missing the keys, I enhanced the error log to make it more clear in the latest commit.

Thanks.

wyli (approved):

PR #3929, opened 2022-03-14
3914 optional lps to ras

### LINKED ISSUE #3914: ITKReader silently changes the affine matrix to RAS without informing the users
**Describe the bug**
The function _get_affine(self, img) of ITKReader() in monai/data/image_reader.py
silently flips the affine matrix to match with the format of nibabel.
See line 307: flip_diag = [[-1, 1], [-1, -1, 1], [-1, -1, 1, 1]][sr - 1]  # itk to nibabel affine

However, the user is not aware of this operation unless they check the source code of ITKReader().
If they thought the affine was the raw affine, it would cause mistakes.

**Additional context**
Three possible solutions:
1.  let the operation of flipping affine matrix be a separate transform
2. set a bool input for ITKReader() to indicate whether to flip the affine.
3. add a key in image_mata_dict to indicate that the image affine has been flipped.

There might be other solutions, but users need to know this operation has been done.

wyli:

Fixes #3914

### Status
**Ready**

### Types of changes

### DIFF
--- a/.github/workflows/pythonapp-gpu.yml
+++ b/.github/workflows/pythonapp-gpu.yml
@@ -19,10 +19,10 @@ jobs:
     strategy:
       matrix:
         environment:
+          - "PT19+CUDA114"
           - "PT17+CUDA102"
           - "PT18+CUDA102"
           - "PT18+CUDA112"
-          - "PT19+CUDA114"
           - "PT110+CUDA116"
           - "PT110+CUDA102"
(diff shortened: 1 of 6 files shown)


Can-Zhao (approved):

Nic-Ma (approved):
Looks good to me.
Do we also need to change image writers for this problem?

Thanks.

PR #3939, opened 2022-03-15
3938 enhance milmodel for torchscript

### LINKED ISSUE #3938: MILModel does not fully support torchscript
the `test_script` function in `test_milmodel.py` only covers the case where `mil_mode` is `mean`, and testing more modes will see errors.

yiheng-wang-nv:

Fixes #3938  .

### Description
With the changes, `MILModel` will support torchscript for all `mil_mode`.

### Status
**Work in progress**

### Types of changes

### DIFF
--- a/monai/networks/nets/milmodel.py
+++ b/monai/networks/nets/milmodel.py
@@ -195,7 +195,7 @@ class TransPyramidModule(nn.Module):
 
         super().__init__()
 
-        self.backbone = backbone
+        self.backbone: models.ResNet = backbone  # type: ignore
         transformer_list = nn.ModuleList(
             [
                 nn.TransformerEncoder(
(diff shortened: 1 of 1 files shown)


myron:
this looks good, but shouldn't we support all modes?

yiheng-wang-nv:
> this looks good, but shouldn't we support all modes?

Hi @myron , sure. To support `att_trans_pyramid ` mode, the hook based method may need to be replaced. Let me update this PR for you to take a look.

monai/networks/nets/milmodel.py
```diff
@@ -203,42 +166,105 @@ def calc_head(self, x: torch.Tensor) -> torch.Tensor:
...
+        )
+        self.transformer = transformer_list
+        self.attention = nn.Sequential(nn.Linear(nfc, 2048), nn.Tanh(), nn.Linear(2048, 1))
+        self.myfc = nn.Linear(nfc, num_classes)
 
-        sh = x.shape
-        x = x.reshape(sh[0] * sh[1], sh[2], sh[3], sh[4])
+    def forward(self, x: torch.Tensor, batch: int, channel: int, no_head: bool = False):
+
+        x = self.backbone.conv1(x)  # type: ignore
```
yiheng-wang-nv, line 235: Hi @wyli , here I manually add `# type: ignore` for multiple lines, otherwise `mypy` will raise:
` error: "Tensor" not callable  [operator]`. Do you have any other suggestions to avoid doing like this? Thanks!

wyli: have you tried to define the type when setting `self.backbone`? instead of here:https://github.com/Project-MONAI/MONAI/blob/94713c4ca7e1aa2a13db53b0c9fb19dc1f0a938b/monai/networks/nets/milmodel.py#L203

could try something like `self.backbone: model.ResNet = backbone  # type: ignore`...

essentially this type is not specific enough https://github.com/Project-MONAI/MONAI/blob/94713c4ca7e1aa2a13db53b0c9fb19dc1f0a938b/monai/networks/nets/milmodel.py#L195

yiheng-wang-nv:
Hi @myron , to support `att_trans_pyramid` mode, the way of using `def calc_head` seems not feasible. Therefore I defined a new class and put everything of the `att_trans_pyramid` mode into it. The code may look quite different from your original version, could you please help to review it? Feel free to ask if you have any questions, thanks!

PR #3945, opened 2022-03-15
3942 3966 adds _requires_ and import support 

### LINKED ISSUE #3942: adds a `_requires_` key for `ConfigComponent`
`ConfigComponent` may require another config section to run beforehand, this could be denoted by 

```yaml
comp:
    _target_: monai.module
    _requires_: @another_section_name
```

### LINKED ISSUE #3966: support of 'import' statement in bundle config
would be great to have the capability of importing modules
```json
{
  "import_glob": "$import glob",
  ...
}
```
instead of local workarounds such as 
```json
{
  "files": "$__import__('glob').glob()",
  ...
}
```

part of https://github.com/Project-MONAI/MONAI/issues/3482

wyli:

Fixes #3942
Fixes #3966

### Status
**Ready**

### Types of changes


monai/bundle/config_item.py
```diff
@@ -164,17 +165,22 @@ class ConfigComponent(ConfigItem, Instantiable):
...
-    Currently, two special keys (strings surrounded by ``_``) are defined and interpreted beyond the regular literals:
+    Currently, three special keys (strings surrounded by ``_``) are defined and interpreted beyond the regular literals:
 
-        - class or function identifier of the python module, specified by one of the two keys.
-            - ``"_target_"``: indicates build-in python classes or functions such as "LoadImageDict",
-                or full module name, such as "monai.transforms.LoadImageDict".
+        - class or function identifier of the python module, specified by ``"_target_"``,
+          indicating a build-in python class or function such as ``"LoadImageDict"``,
+          or a full module name, such as ``"monai.transforms.LoadImageDict"``.
+        - ``"_requires_"``: specifies reference IDs (string starts with ``"@"``) or ``ConfigExpression``
```
Nic-Ma, line 173: 1. I think this doc-string is slightly misleading, if a component use `@` to refer to another component, it already depends on another component to evaluate or initialize first.
For example: `"dataloader": {"dataset": "@dataset"}`.
2. This `_requires_` is only useful for the case that component A and component B are **independent**, but A must run after B is evaluated or initialized.
3. And I think this `_requires_` key is not only useful in `ConfigComponent`, for example, maybe expression A must run after expression B?
4. And maybe it can be `_require_` instead of `_requires_`?

Thanks.

tests/testing_data/inference.json
```diff
@@ -1,6 +1,10 @@
 {
     "dataset_dir": "/workspace/data/Task09_Spleen",
+    "import_glob": "$import glob",
     "device": "$torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')",
+    "set_seed": "$monai.utils.set_determinism(0)",
+    "print_test_name": "$print('json_test')",
+    "print_glob_file": "$print(@import_glob.__file__)",
```
Nic-Ma, line 7: I feel these 2 lines:
```
"import_glob": "$import glob",
"print_glob_file": "$print(@import_glob.__file__)",
```
is not simpler than:
```
"print_glob_file": "$print(__import__('glob').__file__)",
```

If we can set the imports outside this config file (for example in the `run` script), then the expression can be:
```
"print_glob_file": "$print(glob.__file__)",
```

What do you think?

Thanks.

wyli: The local workaround wouldn't support the possibility of setting package variables.

monai/bundle/config_item.py
wyli: thanks, these are good points. after the discussions we want to make sure the 'import' happens early to omit 'requires' for it, and we don't support '_requries_' for configexpression for now (because that could be addressed by other config patterns such as a list of expression notations)

monai/bundle/reference_resolver.py
```diff
@@ -121,8 +123,14 @@ def _resolve_one_item(self, id: str, waiting_list: Optional[Set[str]] = None, **
             waiting_list = set()
         waiting_list.add(id)
 
-        ref_ids = self.find_refs_in_config(config=item_config, id=id)
-        for d in ref_ids:
+        for t, v in self.items.items():
```
Nic-Ma, line 126: Is this a workaround to ensure `imports` expression are before real code?
Thanks for the update, let me think about it and get back to you soon.

Thanks.

monai/bundle/config_item.py
```diff
@@ -337,3 +367,19 @@ def is_expression(cls, config: Union[Dict, List, str]) -> bool:
...
+
+        Args:
+            config: input config content to check.
+        """
+        if not cls.is_expression(config):
+            return False
+        if "import" not in config:
+            return False
+        return isinstance(
+            first(ast.iter_child_nodes(ast.parse(f"{config[len(cls.prefix) :]}"))), (ast.Import, ast.ImportFrom)
```
Nic-Ma, line 384: This line is quite complicated, could you please help add some unit tests to cover all the cases?

Thanks in advance.

Nic-Ma (commented):
Thanks for the quick update.

monai/bundle/config_item.py
```diff
@@ -164,17 +165,24 @@ class ConfigComponent(ConfigItem, Instantiable):
...
+    Currently, three special keys (strings surrounded by ``_``) are defined and interpreted beyond the regular literals:
 
-        - class or function identifier of the python module, specified by one of the two keys.
-            - ``"_target_"``: indicates build-in python classes or functions such as "LoadImageDict",
-                or full module name, such as "monai.transforms.LoadImageDict".
-        - ``"_disabled_"``: a flag to indicate whether to skip the instantiation.
+        - class or function identifier of the python module, specified by ``"_target_"``,
+          indicating a build-in python class or function such as ``"LoadImageDict"``,
+          or a full module name, such as ``"monai.transforms.LoadImageDict"``.
+        - ``"_requires_"`` (optional): specifies reference IDs (string starts with ``"@"``) or ``ConfigExpression``
```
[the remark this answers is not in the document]
wyli: > 3. May I know what the "or `ConfigExpression`" case here is?
> 
one example is for logging: https://github.com/Project-MONAI/MONAI/blob/9fc4d40f6556b495796411e9995c45522864944e/tests/testing_data/inference.yaml#L72-L73 it's also flexible, could be used to insert break points before instantiate

Nic-Ma (approved):

PR #3956, opened 2022-03-16
add MetaTensor

rijobro:
Sub-class `torch.Tensor` to add in `affine` and `meta`.

This will enable us to drop `image_meta_dict` etc by incorporating the relevant meta data inside of the class.

I propose taking a tutorial notebook, and implementing the required transforms so that they work with this new paradigm. Once that has been demonstrated, we should be able to incorporate it into the main code.

I would then propose that we:

1. phase out `_meta_dict` and `_transforms` throughout our code.
2. use torch throughout or transforms rather than supporting both numpy and torch.

I think these two changes should make our code substantially simpler and more correct (meta data is inherently attached to the correct object, rather than relying on similar names).

We should also have the option to disable meta data, I have created the getters and setters `get_track_meta` and `set_track_meta`.

```python
# before
img = torch.zeros((1,10,10))
image_meta_dict = {"affine": torch.eye(4), "fname": "/path/to/file"}
data = {"img": img, "image_meta_dict": image_meta_dict}

# proposed
img = torch.zeros((1,10,10))
affine = torch.eye(4)
meta = {"fname": "/path/to/file"}
img = MetaTensor(img, affine, meta)
data = {"img": img}
```

### Status
**Work in progress**

### Types of changes

### DIFF
--- a/docs/source/data.rst
+++ b/docs/source/data.rst
@@ -107,15 +107,22 @@ Patch-based dataset
 .. autoclass:: GridPatchDataset
   :members:
 
-`PatchIter`
-~~~~~~~~~~~
-.. autoclass:: PatchIter
-  :members:
-
 `PatchDataset`
 ~~~~~~~~~~~~~~
 .. autoclass:: PatchDataset
   :members:
+
+`PatchIter`
+"""""""""""
+.. autoclass:: PatchIter
+    :members:
+    :special-members: __call__
+
+`PatchIterd`
+""""""""""""
+.. autoclass:: PatchIterd
+    :members:
+    :special-members: __call__
 
 Image reader
 ------------
--- a/monai/data/__init__.py
+++ b/monai/data/__init__.py
@@ -32,7 +32,7 @@ from .decathlon_datalist import (
     load_decathlon_properties,
 )
 from .folder_layout import FolderLayout
-from .grid_dataset import GridPatchDataset, PatchDataset, PatchIter
+from .grid_dataset import GridPatchDataset, PatchDataset, PatchIter, PatchIterd
 from .image_dataset import ImageDataset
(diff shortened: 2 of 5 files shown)


monai/data/meta_tensor.py
```diff
@@ -0,0 +1,257 @@
...
+        transforms: TransformsType | None = None,
+    ) -> None:
+        """If `affine` is given, use it. Else, if `affine` exists in the input tensor, use it. Else, use
+        the default value. The same is true for `meta` and `transforms`."""
+        self.set_initial_val("affine", affine, x, self.get_default_affine)
+        self.set_initial_val("meta", meta, x, self.get_default_meta)
+        self.set_initial_val("transforms", transforms, x, self.get_default_transforms)
+
+    @staticmethod
+    def get_meta_tensors(args):
```
Nic-Ma, line 134: May need to support nested meta info.
Similar to bundle access: https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/config_parser.py#L118.

Thanks.

monai/data/utils.py
```diff
@@ -346,9 +347,21 @@ def list_data_collate(batch: Sequence):
             ret = {}
             for k in elem:
                 key = k
-                ret[key] = default_collate([d[key] for d in data])
+                data_for_batch = [d[key] for d in data]
+                ret[key] = default_collate(data_for_batch)
+                if isinstance(ret[key], MetaTensor) and all(isinstance(d, MetaTensor) for d in data_for_batch):
```
Nic-Ma, line 352: May I know why need to collate `affine`, `meta`, `transforms` explicitly? We don't have such logic for `image_meta_dict` now.

Thanks.

PR #3963, opened 2022-03-17
3962 Update CacheDataset num workers

### LINKED ISSUE #3962: `CacheDataset` gets stuck when initializing cache content
Feedback from Andriy when using NVIDIA Clara MMARs:
one peculiar issue if we mount dataset from NGC, then training gets stuck in 90% of times (both train.sh and train_multi_gpu.sh)
I believe it has something to do with SmartCache (since it’s the only MMAR that uses it) and ngc-dataset interoperability. 
It says : “Loading Data: 0%” and just gets stuck, sometimes it goes to “Loading Data: 6%” and then gets stuck.
Is it something we should be worried about?
 
PS: If I copy data to local DGX folder first, then training runs fine.

yiheng-wang-nv:

Fixes #3962 .

### Description
In this PR, the default value of `num_workers` is set to 1 for `CacheDataset` and all its subclasses. Some of the subclasses use the value 0, although it will be replaced by 1 automatically, zero thread is meaningless and it is actually an illegal input for `multiprocessing.pool.ThreadPool` (used within the `CacheDataset`, the source code of it is in [here](https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/multiprocessing/pool.py#L205)) thus I also modified them.
In addition, this PR also fixes a few typos and unify the docstrings for same arguments in different classes.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/apps/pathology/data/datasets.py
+++ b/monai/apps/pathology/data/datasets.py
@@ -119,8 +119,10 @@ class SmartCachePatchWSIDataset(SmartCacheDataset):
             will take the minimum of (cache_num, data_length x cache_rate, data_length).
         num_init_workers: the number of worker threads to initialize the cache for first epoch.
             If num_init_workers is None then the number returned by os.cpu_count() is used.
+            If a value less than 1 is speficied, 1 will be used instead.
         num_replace_workers: the number of worker threads to prepare the replacement cache for every epoch.
             If num_replace_workers is None then the number returned by os.cpu_count() is used.
+            If a value less than 1 is speficied, 1 will be used instead.
         progress: whether to display a progress bar when caching for the first epoch.
         copy_cache: whether to `deepcopy` the cache content before applying the random transforms,
             default to `True`. if the random transforms don't modify the cache content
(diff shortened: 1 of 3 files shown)


monai/apps/pathology/data/datasets.py
```diff
@@ -119,8 +119,10 @@ class SmartCachePatchWSIDataset(SmartCacheDataset):
             will take the minimum of (cache_num, data_length x cache_rate, data_length).
         num_init_workers: the number of worker threads to initialize the cache for first epoch.
             If num_init_workers is None then the number returned by os.cpu_count() is used.
+            If a value less than 1 is speficied, 1 will be used instead.
```
Nic-Ma, line 122: So `0` is actually not work for the logic?

Thanks.

yiheng-wang-nv: Hi @Nic-Ma , yes.
```
        if self.num_workers is not None:
            self.num_workers = max(int(self.num_workers), 1)
```
The above code will convert it into 1, and without the convert, error will be raised when input 0 into `ThreadPool `.

wyli (approved):

PR #3965, opened 2022-03-17
3482 add export script for TorchScript file

Nic-Ma:
Task step 4 and 13 of https://github.com/Project-MONAI/MONAI/issues/3482 .

### Description
This PR added `export` script to export the model, config, metadata to a TorchScript file.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @ericspod @wyli ,

This PR is ready for review, it adds `export` script and support `customized` ConfigItem and ReferenceResolver in ConfigParser.
Could you please help review it when you are available?

Thanks in advance.

wyli:
Thanks, I had a look and the custom resolver still needs more discussions. could you have a separate PR for `export` and perhaps @ericspod could review and confirm the API usage.

Nic-Ma:
Hi @wyli @ericspod ,

Sure, I removed the customized config item and resolver logic, could you please help review it again?

Thanks in advance.

tests/test_bundle_export.py
```diff
@@ -0,0 +1,56 @@
...
+            parser.export_config_file(config=def_args, filepath=def_args_file)
+            parser.read_config(config_file)
+            net = parser.get_parsed_content("network_def")
+            save_state(src=net if key_in_ckpt == "" else {key_in_ckpt: net}, path=ckpt_file)
+
+            cmd = [sys.executable, "-m", "monai.bundle", "export", "network_def", "--filepath", ts_file]
+            cmd += ["--meta_file", meta_file, "--config_file", config_file, "--ckpt_file", ckpt_file]
+            cmd += ["--key_in_ckpt", key_in_ckpt, "--args_file", def_args_file]
+            ret = subprocess.check_call(cmd)
+            self.assertEqual(ret, 0)
```
ericspod, line 51: `check_call` will raise `CalledProcessError` if the return code isn't 0 so either you don't need this line or use `call` instead.

monai/bundle/scripts.py
```diff
@@ -321,3 +325,85 @@ def verify_net_in_out(
         if output.dtype != output_dtype:
             raise ValueError(f"dtype of output data `{output.dtype}` doesn't match: {output_dtype}.")
     logger.info("data shape of network is verified with no error.")
+
+
+def export(
```
ericspod, line 330: Could we call this `checkpoint_export` instead to differentiate from other places functions/methods of this name are used? I would suggest adding another function `net_export` to take a network instance and do the exporting with that so that people have something they can call in their own code with the network they just finished training. We could also have a third function which takes a Torchscript file instead of a checkpoint file which could copy it into the bundle and extract the checkpoint file from it to save separately. These latter extensions we could leave for later

ericspod (approved):
Looks good to me, I suggested adding functionality to do the same thing from a Torchscript object as well as checkpoint file but this isn't necessary.

tests/test_bundle_export.py
Nic-Ma: Thanks for your tips, updated in the latest commit.

monai/bundle/scripts.py
Nic-Ma: Hi @ericspod ,

I changed it to `ckpt_export` in the latest commit.
About the TorchScript input file idea, how can we automatically determine it's a checkpoint or torchscript file?

Thanks.

monai/bundle/scripts.py
```diff
@@ -321,3 +325,85 @@ def verify_net_in_out(
...
+    parser = ConfigParser()
+    config_file_ = _args.pop("config_file")
+    parser.read_config(f=config_file_)
+    meta_file = _args.pop("meta_file")
+    if meta_file is not None:
+        parser.read_meta(f=meta_file)
+    id = _args.pop("net_id", "")
+    path = _args.pop("filepath")
+    ckpt = torch.load(_args.pop("ckpt_file"))
+    key = _args.pop("key_in_ckpt", "")
```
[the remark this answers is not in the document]
Nic-Ma: Sure, updated in the latest commit with unified API.

Thanks.

monai/bundle/scripts.py
Nic-Ma: I feel it's a little bit confusing to load `TorchScript` model as "checkpoint" in this function, maybe we can add a separate function for "convert TorchScript model" later?

Thanks.

wyli:
thanks, I just saw that `subprocess.check_call` in the unit tests wouldn't count towards the total code coverage, could you help have a fix? is it possible to use `coverage` as the `sys.executable`?

Nic-Ma:
> thanks, I just saw that `subprocess.check_call` in the unit tests wouldn't count towards the total code coverage, could you help have a fix? is it possible to use `coverage` as the `sys.executable`?

Hi @wyli ,

Sure, I updated to use `coverage run` instead.

Thanks in advance.

PR #3974, opened 2022-03-22
3482 Add support for relative IDs in the config content

Nic-Ma:
Task step 12 of https://github.com/Project-MONAI/MONAI/issues/3482 .

### Description
According to the task plan, this PR added support for relative reference ID, like: "test": "@###data#1".

### Status
**Ready**

### Types of changes


wyli:
I think this makes it difficult to distinguish the empty id `""` and the parent reference?

Nic-Ma:
> I think this makes it difficult to distinguish the empty id `""` and the parent reference?

Hi @wyli ,

The regex in this PR is `re.compile(rf"{ref}{sep}+")`, which only match the relative IDs at beginning of the string.

Thanks.

monai/bundle/reference_resolver.py
```diff
@@ -172,6 +174,39 @@ def get_resolved_content(self, id: str, **kwargs):
...
+            id: id name for current config item to compute relative id.
+            value: input value to resolve relative ids.
+
+        """
+        # get the prefixes like: "@####", "@###", "@#"
+        prefixes = sorted(set().union(cls.relative_id_prefix.findall(value)), reverse=True)
+        current_id = id.split(cls.sep)
+
+        for p in prefixes:
+            length = len(p[len(cls.ref) :]) // len(cls.sep)
```
wyli, line 200: this could be substring counting: `p[len(cls.ref) :].count(cls.sep)`

monai/bundle/reference_resolver.py
```diff
@@ -172,6 +174,39 @@ def get_resolved_content(self, id: str, **kwargs):
...
+                "C": {"key": "@##A", "value1": 2, "value2": "@#value1", "value3": [3, 4, "@#1"]},
+            }
+
+        Args:
+            id: id name for current config item to compute relative id.
+            value: input value to resolve relative ids.
+
+        """
+        # get the prefixes like: "@####", "@###", "@#"
+        prefixes = sorted(set().union(cls.relative_id_prefix.findall(value)), reverse=True)
```
wyli, line 196: any idea why `sorted`?

Nic-Ma: Hi @wyli ,

Because we should replace the longer string first, for example `$@####AAA#BB + @##CCC#DD`.

Thanks.

monai/bundle/reference_resolver.py
Nic-Ma: Updated in the latest commit.

Thanks.

monai/bundle/reference_resolver.py
```diff
@@ -172,6 +174,39 @@ def get_resolved_content(self, id: str, **kwargs):
         """
         return self._resolve_one_item(id=id, **kwargs)
 
+    @classmethod
+    def resolve_relative_ids(cls, id: str, value: str) -> str:
```
ericspod, line 178: It's an internal function but I think we should have more description of what this is actually doing here. The meaning of the relative IDs "@#A" vs "@##A" or something else isn't explained, it would help to motivate what this does here. The example doesn't demonstrate calling the function with inputs and expected outputs.

Nic-Ma: Hi @ericspod ,

Sure, added more description in the latest commit.

Thanks for your suggestion.

Nic-Ma:
Hi @wyli @ericspod ,

I revisited the feature, actually, the relative id conversion is kind of "macro replacement" and we need to support it for both the `macro token %XXX` case and `reference @XXX` case. So I changed the logic to `ConfigParser` and no need to call it twice now.

Thanks in advance.

wyli (approved):

PR #3978, opened 2022-03-22
3961 fixes testing memory bug and revert the workaround

### LINKED ISSUE #3961: Revert the workaround of base image downgrading (PR 3949)
As described and implemented in https://github.com/Project-MONAI/MONAI/pull/3949, pytorch docker images after v21.10 seems to have a much larger memory footprint when launching the dataloader with num_workers>1 for the integration test:  
https://github.com/Project-MONAI/MONAI/blob/6ff85350d8cd698507e4476997e32e09b3be5eed/tests/test_integration_classification_2d.py#L59
the base image was downgraded as a temp workaround to avoid this issue in thg 16GB v100 CI.

wyli:
Fixes #3961

### Description
Multiprocessing in the 'spawn' mode would trigger the `has_cupy` evaluation for each subprocess. the evaluations will cause multple cuda context loading which takes a good amount of memory. The issue becomes obvious when 'num_workers' is large.

The solution here is to skip cupy check if it's in subprocess. (the main testing process is named [by default "MainProcess"]( https://github.com/python/cpython/blob/4aea656d62860e78cd8384f2de375f0d4f1db579/Lib/multiprocessing/process.py#L395)). A stress test is added to avoid this bug in the future.

### Status
**Ready**

### Types of changes

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -62,7 +62,7 @@ jobs:
     if: github.repository == 'Project-MONAI/MONAI'
     strategy:
       matrix:
-        container: ["pytorch:21.02", "pytorch:21.10"]  # 21.02 for backward comp.
+        container: ["pytorch:21.02", "pytorch:22.02"]  # 21.02 for backward comp.
     container:
       image: nvcr.io/nvidia/${{ matrix.container }}-py3  # testing with the latest pytorch base image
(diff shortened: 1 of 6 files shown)


Nic-Ma (approved):
Thanks for the deep investigation for the issue!
Looks good to me.

PR #3980, opened 2022-03-22
3482 Add support for customized ConfigItem and resolver

Nic-Ma:
Task step 13 of https://github.com/Project-MONAI/MONAI/issues/3482 .

### Description
This PR added support for customized ConfigItem and resolver in `ConfigParser`.

### Status
**Ready**

### Types of changes


wyli:
thanks, I think we should consider this optional feature later when the module is more stable...

Nic-Ma:
> thanks, I think we should consider this optional feature later when the module is more stable...Perhaps we remove this feature from #3482? probably it's not part of an MVP.

Hi @wyli ,

We added this task to #3482 when we were analyzing @dongyang0122 's YAML config for DiNTS, it can be supported by a customized `ConfigComponent`.

Thanks.

Issue #4003, opened 2022-03-25
pylint update broken


wyli:
**Describe the bug**
see also https://github.com/PyCQA/pylint/issues/5969

PR #4004, opened 2022-03-25
Update cuCIM min dependency to support random transforms

drbeh:
### Description
The new release of cuCIM (22.2.1) includes important updates to the random transforms, where the randomness is apply per-image instead of per-batch which was the case in the previous versions. Our updated pathology tutorials depend on this version to work so it should be set as the minimum version in MONAI.

### Status
**Ready**

### Types of changes


wyli (approved):

wyli:
(there's an irrelevant issue affecting https://github.com/Project-MONAI/MONAI/issues/4003 I'll fix it and merge this PR)

wyli:
looks like the cucim pypi update breaks a test here https://github.com/Project-MONAI/MONAI/runs/5695808906?check_suite_focus=true#step:7:22030

drbeh:
That's weird. I'll check on the test.

Nic-Ma (approved):

PR #4029, opened 2022-03-31
3482 Add detection for circular macro replacement

Nic-Ma:
### Description
Thanks for Gigon's suggestion, this PR added support to detect circular macro replacement in a config file.
And it also fixed a bug that missing the `id` arg in recursive operation, and now we don't do recursive logic for the case that macro text is in another config file.
Added many unit tests to cover the cases.

### Status
**Ready**

### Types of changes


monai/bundle/config_parser.py
```diff
@@ -266,18 +266,32 @@ def _do_resolve(self, config: Any, id: str = ""):
                 go one level further into the nested structures.
                 Use digits indexing from "0" for list or other strings for dict.
                 For example: ``"xform#5"``, ``"net#channels"``. ``""`` indicates the entire ``self.config``.
+            waiting_list: set of macro replacement ids pending to be resolved.
+                It's used to detect circular references such as:
+                `{"A": {"dep": "%B"}, "B": {"dep": "%A"}}`.
 
         """
+        if waiting_list is None:
+            waiting_list = set()
```
ericspod, line 275: If `waiting_list` is actually a list we could look at the previous item in the exception to help the user trace what's happened.

PR #4036, opened 2022-03-31
Add support for `level` in `PatchWSIDataset`

### LINKED ISSUE #4034: Support for chosen level in `PatchWSIDataset`
Following up on #3995, user likes to be able to define the whole slide image `level` when using `PatchWSIDataset` in order to load patches at that `level`.

For the solution, we should consider that the `location` is always refer to the coordinates at `level=0`.

drbeh:
Fixes #4034 

### Description
Add supports in `PatchWSIDataset` to accept `level` as an argument and extract patches from the whole slide images at that `level`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/apps/pathology/data/datasets.py
+++ b/monai/apps/pathology/data/datasets.py
@@ -32,10 +32,10 @@ class PatchWSIDataset(Dataset):
         region_size: the size of regions to be extracted from the whole slide image.
         grid_shape: the grid shape on which the patches should be extracted.
         patch_size: the size of patches extracted from the region on the grid.
-        level: the level at which the patches to be extracted (default to 0).
         transform: transforms to be executed on input data.
         image_reader_name: the name of library to be used for loading whole slide imaging, either CuCIM or OpenSlide.
             Defaults to CuCIM.
+        kwargs: additional parameters for ``WSIReader``
 
     Note:
         The input data has the following form as an example:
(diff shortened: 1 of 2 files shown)


Nic-Ma:
Hi @drbeh ,

Thanks for your quick update according to user's feedback, I have 2 comments:
1. Should we also add the `level` arg to other datasets in the file?
Like `SmartCachePatchWSIDataset`, `MaskedInferenceWSIDataset`.
3. As `WSIReader` also supports `**kwargs`, could you please help add it to the datasets?
Then I think it can help support more similar requirements in future.

What do you think?

Thanks.

drbeh:
Thank you @Nic-Ma,  it was very helpful comment.

monai/apps/pathology/data/datasets.py
```diff
@@ -32,6 +32,7 @@ class PatchWSIDataset(Dataset):
         region_size: the size of regions to be extracted from the whole slide image.
         grid_shape: the grid shape on which the patches should be extracted.
         patch_size: the size of patches extracted from the region on the grid.
+        level: the level at which the patches to be extracted (default to 0).
```
Nic-Ma, line 35: As you changed the arg to `kwargs`, could you please help update the doc-string to `kwargs` here?
And also add doc-string to other datasets?

Thanks.

Nic-Ma (approved):

PR #4050, opened 2022-04-01
added 2D (normalized) surface dice metric

SilviaSeidlitz:
Signed-off-by: Silvia Seidlitz <s.seidlitz@dkfz-heidelberg.de>

Adds implementation of the normalized surface dice (NSD) metric for 2D images.

### Description
The NSD implementation is based on the adaptation for 2D images described in [our paper](https://doi.org/10.48550/arXiv.2111.05408). This metric determines which fraction of a segmentation boundary is correctly predicted. A boundary element is considered correctly predicted if the closest distance to the reference boundary is smaller than or equal to the specified threshold, which is related to the (clinically) acceptable amount of deviation in pixels. The NSD is bounded between 0 and 1.
The interface is similar to the interface for the average surface distance computation.

### Status
**Ready**

### Types of changes


monai/metrics/surface_dice.py
```diff
@@ -0,0 +1,233 @@
...
+        :math:`b` and class :math:`c`.
+    """
+
+    if not include_background:
+        y_pred, y = ignore_background(y_pred=y_pred, y=y)
+
+    if not isinstance(y_pred, torch.Tensor) or not isinstance(y, torch.Tensor):
+        raise ValueError("y_pred and y must be PyTorch Tensor.")
+
+    if y_pred.ndimension() != 4 or y.ndimension() != 4:
```
wyli, line 181: do you think the function could be extended to handle 3D and this value error could be removed? I don't see it's invoking any 2d only APIs in this function...

SilviaSeidlitz: Of course it could be extended, but the way 3D surfaces & surface distances are handled in deep mind is different from MONAI: Instead of defining surfaces at surface voxel boundaries, surface points are located on a grid that is shifted by half of the grid spacing along each axis (cf. Fig. 10 in their [paper](https://arxiv.org/abs/1809.04430)):

grafik

I think it should be discussed more broadly whether to keep the surface computation currently used in MONAI, or to switch to the deep mind computation, or to establish a switch such that the user can decide on the computation. This choice should be made consistently for all boundary-overlap metrics. Has this topic already been discussed?

wyli: I don't think this has been discussed cc@AReinke @csudre. Agree that having a flag to use different approximations would be nice. we can have a new ticket afterwards.

For this PR for the same reason, in 2D it's also a different approximation compared with the deepmind's? because the local lengths are assumed all 1s, but the deepmind's code has the possibility of 1 and 0.5*sqrt(2)...would be great to have the paper reference in the docstring https://arxiv.org/abs/2111.05408.

SilviaSeidlitz: Yes, in 2D as well this implementation is different from Deep Mind's implementation. I added a clarification to the documentation. The advantage of this implementation is that it is simpler to interpret (proportion of boundary pixels that are correct) and more comparable with how other metrics in MONAI are implemented. E.g. when assessing the dice score, segmentations are also considered at the discrete pixel level with the boundaries of the segmentation being the pixel edges instead of constructing non-discrete boundaries and thus having boundary pixels contributing only partially. Also, I find it inconsistent that the non-discrete boundaries are not used for distance calculations in Deep Mind's implementation.
I would definetely welcome expanding the implementation to 3D, but since I'm not working with 3D data, others should decide whether e.g. the interpretation of the NSD as proportion of boundary pixels that are correct is desirable or consideration of anisotropic voxel spacing is required.

wyli: Thank you, I'm merging this one soon!

wyli (approved):

PR #4061, opened 2022-04-04
4060 Add PatchIter and PatchIterd transform

### LINKED ISSUE #4060: Add a `dict` version `PatchIterd`
Thanks for user's feedback in https://github.com/Project-MONAI/MONAI/discussions/4058, it would be nice to add a dict version `PatchIterd` with expected `keys` for the `GridPatchDataset` dataset.

Nic-Ma:
Fixes #4060 .

### Description
This PR added `PatchIter` and `PatchIterd` transform according to user's feedback.

### Status
**Ready**

### Types of changes


monai/transforms/croppad/array.py
```diff
@@ -1212,3 +1213,58 @@ def __call__(self, img: NdarrayOrTensor) -> np.ndarray:
...
+        self.kwargs = kwargs
+
+    def __call__(self, img: NdarrayOrTensor):
+        """
+        Args:
+            img: the image to generate patches from.
+
+        """
+        img_np, *_ = convert_data_type(img, np.ndarray)
+        yield from iter_patch(
```
wyli, line 1263: this is very different from the other transforms, can we keep it in the grid_dataset module?

monai/transforms/croppad/dictionary.py
```diff
@@ -1525,6 +1529,53 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> Dict[Hashable, N
...
+
+    def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]):
+        d = dict(data)
+        for t in zip(*[self.patch_iter(d[key]) for key in self.key_iterator(d)]):
+            coords = t[0][1]  # use the coordinate of the first item
+            ret = {k: v[0] for k, v in zip(self.key_iterator(d), t)}
+            # fill in the extra keys with unmodified data
+            for key in set(d.keys()).difference(set(self.keys)):
+                ret[key] = deepcopy(d[key])
+            yield ret, coords
```
wyli, line 1576: would be great to also store the coords as part of the data dict.

Nic-Ma: OK, I added the `coords` in the dict in the latest commit.

Thanks.

monai/transforms/croppad/array.py
Nic-Ma: I think we also have other similar transforms that don't return transformed image, like `BoundingRect`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/croppad/array.py#L1175
And the `Transform` and `MapTransform` base classes don't define the return data of `__call__()` function:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/transform.py#L242
What do you think?

Thanks.

wyli: this is a class to convert images for patch-level analysis:
```py
image_dataset = Dataset(list_of_images, transforms=...)
patch_iter = PatchIter(...)
patch_dataset = GridPatchDataset(image_dataset, patch_iter, transforms=...)
shuffled_patches = ShuffleBuffer(patch_dataset)
```
so the image/patch dataset has its own preprocessing steps, and the patchiter describes the logic of generating patches from images.

But I don't think this pipeline is very important given that we have various cropping transforms and the sliding window methods.

Nic-Ma: OK, let me change it back to `grid` logic to be compatible with the existing logic.

Thanks.

Nic-Ma: Hi @wyli ,

I have updated the `PatchIter` and `PatchIterd` according to your comments, could you please help review it again?

Thanks in advance.

monai/data/grid_dataset.py
```diff
@@ -33,56 +55,83 @@ def __init__(
...
+        mode: Union[NumpyPadMode, str] = NumpyPadMode.WRAP,
+        **pad_opts,
+    ):
+        self.keys = ensure_tuple(keys)
+        self.patch_iter = PatchIter(patch_size=patch_size, start_pos=start_pos, mode=mode, **pad_opts)
+
+    def __call__(self, data: Mapping[Hashable, np.ndarray]):
+        d = dict(data)
+        for patch in zip(*[self.patch_iter(d[key]) for key in self.keys]):
+            coords = patch[0][1]  # use the coordinate of the first item
```
ericspod, line 115: Are we assuming that every image `d[key]` is the same shape? If so there isn't really a need to consider patch iteration for every image in `[self.patch_iter(d[key]) for key in self.keys]` but only one, so `self.patch_iter(d[first(self.keys)])` assuming `self.keys` is non-empty.

wyli (approved):
thanks this looks ok although I think 1. `PatchIter`'s docstring change and `docs/source/data.rst` changes are unnecessary, and 2. it might be useful to include the original image size and `start_pos` in each returned patch dictionary, so that it's possible to (partially) reconstruct the original image from the patches.

monai/data/grid_dataset.py
Nic-Ma: Hi @ericspod ,

Thanks for your review.
I don't quite understand your suggestion, could you please help explain more?
I think we need to iterate every image to get the corresponding patches at the same position.

Thanks in advance.

Nic-Ma: Hi @ericspod ,

I think in `{k: v[0] for k, v in zip(self.keys, patch)}`, the `patch` is a list of patches corresponding to every key, like "image", "label", etc. Generated from the `zip(*[XXX])`:
`zip(*[self.patch_iter(d[key]) for key in self.keys])`
So `v` is every item in the list, why you think `v[0]` is always the same?
Or maybe I misunderstood something?

Thanks in advance.

Nic-Ma: Hi @ericspod ,

Thanks for your explanation, I see your point.
I think `self.patch_iter` returns the patch data instead of the slices, so we need to run for every key.
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/utils.py#L245

Thanks.

ericspod: `self.patch_iter` is giving you both the patch data and the slice indices, if all the images are the same size the patch data will vary but the indices will not. I get what's going on here better and I'm happy for this to go through. We could change some things to maybe make it a little more efficient know the indices are always the same but let's leave that for some other time. Thanks!

Nic-Ma: Thanks for the discussion, I agree with your points.

PR #4075, opened 2022-04-05
Fix a logical error in level verification

drbeh:
### Description
This PR fixes a logical error in level verification. It did not have impact on the functionality of the reader but was verifying the default level.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/image_reader.py
+++ b/monai/data/image_reader.py
@@ -848,7 +848,8 @@ class WSIReader(ImageReader):
         return patches, metadata
 
     def _get_max_level(self, img_obj):
-        """ Return the maximum number of levels in the whole slide image
+        """ 
+        Return the maximum number of levels in the whole slide image
         Args:
             img: the whole slide image object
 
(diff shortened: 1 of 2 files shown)


monai/data/image_reader.py
```diff
@@ -845,8 +846,6 @@ def get_data(
         return patches, metadata
 
     def _check_level(self, img, level):
```
Nic-Ma, line 848: I feel maybe it's better to change this function to `_get_level_count()`, then compare `level` and `level_count` at the outside.
What do you think?

Thanks.

Nic-Ma (approved):

drbeh:
PyPI is dow! We should run the test again when it's back.

monai/data/image_reader.py
```diff
@@ -844,21 +847,21 @@ def get_data(
...
-    def _check_level(self, img, level):
-        level = self.level
+    def _get_max_level(self, img_obj):
+        """
+        Return the maximum number of levels in the whole slide image
+        Args:
+            img: the whole slide image object
 
-        level_count = 0
+        Returns:
```
Nic-Ma, line 856: I feel no need this `returns` part as it's exactly same as the doc-string.

Thanks.

Nic-Ma (approved):

PR #4077, opened 2022-04-05
meta tensor

rijobro:
### Description
First PR to MetaTensor branch.

As promised, we'll merge to `MetaTensor` branch via PRs and code reviews. This is the first PR and just adds the object and some tests. `MetaObj` allows for subclassing `np.ndarray` in the future should we want to.

### Status
**Ready**

### Types of changes


monai/data/meta_tensor.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def get_default_affine(self) -> torch.Tensor:
+        return torch.eye(4, device=self.device)
+
+    def as_tensor(self) -> torch.Tensor:
+        """
+        Return the `MetaTensor` as a `torch.Tensor`.
+        It is OS dependent as to whether this will be a deep copy or not.
+        """
+        with warnings.catch_warnings():
```
ericspod, line 72: What warning is being suppressed here?

wyli:
This is non-breaking, I think we should try to merge this into the `dev` branch directly so that we can potentially get more early feedback... after this PR we can collaborate on the `MetaTensor` branch at full speed for all the necessary updates.

for the naming of key classes and modules: `meta_obj`, `MetaObj`, `meta_tensor`, `MetaTensor`, do you agree with the current PR? @ericspod @Nic-Ma @atbenmurray @yiheng-wang-nv (can discuss on Friday as well).

rijobro:
Should `MetaTensor` be `MetaTensorImage` to add in another layer of inheritance and allow for `MetaTensorPoints` etc?

monai/data/meta_tensor.py
rijobro: Warning is `To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).`

I don't know how else to cast from `MetaTensor` to `torch.Tensor`. Using `torch.as_tensor` returns the same type. And their suggestion in the warning won't work as `clone` will return the same type, too.

rijobro:
@wyli Is there much point merging this into `dev`? It won't be used until we have some functionality that uses it.

wyli:
> @wyli Is there much point merging this into `dev`? It won't be used until we have some functionality that uses it.

yes it'll be convenient for creating demos or experimenting with metatensor using `pip install monai-weekly` or pip install from git+dev branch. I think the `MetaTensor` branch will be largely broken as we introduce breaking changes in the following weeks/months.

monai/data/meta_obj.py
```diff
@@ -0,0 +1,172 @@
...
+    Copying metadata:
+        * For `c = a + b`, then the meta data will be copied from the first
+        instance of `MetaImage`.
+    """
+
+    _meta: dict
+    _affine: torch.Tensor
+
+    def set_initial_val(self, attribute: str, input_arg: Any, input_tensor: MetaObj, default_fn: Callable) -> None:
+        """
```
Nic-Ma, line 79: Could you please help add doc-string for the `args` in this function?
I don't understand what `input_arg`, `input_tensor`, `default_fn` mean before I look into the code carefully.

Thanks.

monai/data/meta_obj.py
```diff
@@ -0,0 +1,172 @@
...
+        use a default value.
+        """
+        if input_arg is None:
+            input_arg = getattr(input_tensor, attribute, None)
+        if input_arg is None:
+            input_arg = default_fn(self)
+        setattr(self, attribute, input_arg)
+
+    @staticmethod
+    def get_tensors_or_arrays(args: Sequence[Any]) -> list[MetaObj]:
```
Nic-Ma, line 91: I feel this function is more like "flatten data"? And no need to mention `tensor` or `array` in this base class.

Thanks.

monai/data/meta_tensor.py
ericspod: The code and the warning for this buried deep in the [C++ code](https://github.com/pytorch/pytorch/blob/master/torch/csrc/utils/tensor_new.cpp#L860). It might be best for now to change the filter to actually catch this warning in case it changes in the future to say something about deprecation when we'd have to worry about it explicitly. The only other solution would be to extract and detach the data with `self.data.detach()` and extract other information to construct a tensor from scratch. `torch.Tensor` produces a view with non-copied data so I'm not sure it works.

Nic-Ma:
Hi @vfdev-5 ,

We are trying to extend PyTorch Tensor with medical metadata properties.
Could you please help also take a look at this first PR from the PyTorch team angle?

Thanks in advance.

vfdev-5:
@Nic-Ma thanks for pinging !

I'll check the PR in details a bit later, right now few comments/info (in case if helpful):
- In torchvision there is a work on inheriting from pytorch tensors => tensor images:
https://github.com/pytorch/vision/blob/08cc9a7fb5baa9655fa5d2dd44a6fcfcbe48a1c7/torchvision/prototype/features/_image.py#L37 , you can check out how it is done
- "meta" name may be confusing as pytorch has "meta" device as well (~ lazy tensors)

monai/data/meta_obj.py
```diff
@@ -0,0 +1,172 @@
...
+import numpy as np
+import torch
+
+_TRACK_META = True
+_TRACK_TRANSFORMS = True
+
+__all__ = ["get_track_meta", "get_track_transforms", "set_track_meta", "set_track_transforms", "MetaObj"]
+
+
+def set_track_meta(val: bool) -> None:
```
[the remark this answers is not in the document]
ericspod: Richard says it's similar to how `set_determinism` works in that `_TRACK_META` sets whether to propagate metadata or not, in the latter case operations produce regular tensors. For `_TRACK_TRANSFORMS` it's not used but will be for tracking metadata through transforms specifically.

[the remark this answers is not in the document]
Nic-Ma: OK, I see. Hope we don't have any multi-thread issues with global vars.

Thanks.

Nic-Ma:
And as @vfdev-5 mentioned in another conversation, this `MetaTensor` may have issue with `TorchScript`?

Thanks.

monai/data/meta_obj.py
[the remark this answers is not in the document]
wyli: sure, very unlikely we'll change this via multiple threads. (btw there's no need to use `global` if it's just reading the variable `return _TRACK_META` https://stackoverflow.com/a/423596)

monai/data/meta_obj.py
```diff
@@ -0,0 +1,172 @@
...
+        If there has been a change in `id` (e.g., `a+b`), then deepcopy. Else (e.g., `a+=1`), don't.
+        """
+        id_in = id(input_meta_objs[0]) if len(input_meta_objs) > 0 else None
+        deepcopy_required = id(self) != id_in
+        attributes = ("affine", "meta")
+        default_fns: tuple[Callable, ...] = (self.get_default_affine, self.get_default_meta)
+        for attribute, default_fn in zip(attributes, default_fns):
+            self._copy_attr(attribute, input_meta_objs, default_fn, deepcopy_required)
+
+    def get_default_meta(self) -> dict:
```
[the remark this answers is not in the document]
rijobro: It seemed cleaner to have specific functions for each of the default attributes. Otherwise there might be `{}` around the code but it wouldn't be obvious why.

tests/utils.py
```diff
@@ -715,5 +715,10 @@ def query_memory(n=2):
     TEST_NDARRAYS = TEST_NDARRAYS + (gpu_tensor,)  # type: ignore
 
 
+TEST_DEVICES = [[torch.device("cpu")]]
+if torch.cuda.is_available():
+    TEST_DEVICES.append([torch.device("cuda")])
```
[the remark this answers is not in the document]
wyli: yes in our multigpu testing environment cuda:0 is mostly more precise. we can make it configurable (by id) and robust, such as:
```py
def get_test_devices(id=["0"]):
   ...
   if torch.cuda.device_count() < len(id)
   ...
```
but this could be done in the future PRs if we encounter actual memory issues.

monai/data/meta_obj.py
```diff
@@ -0,0 +1,172 @@
...
+        out = []
+        for a in args:
+            if isinstance(a, (list, tuple)):
+                out += MetaObj.get_tensors_or_arrays(a)
+            elif isinstance(a, MetaObj):
+                out.append(a)
+        return out
+
+    def _copy_attr(
+        self, attribute: str, input_objs: list[MetaObj], default_fn: Callable, deepcopy_required: bool
```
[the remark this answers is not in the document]
rijobro: That would shadow `deepcopy` imported from `copy`.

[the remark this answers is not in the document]
Nic-Ma: OK, I just feel maybe `deepcopy_required` is a little bit longer, I am fine with it or `deep_copy`.

Thanks.

monai/data/meta_tensor.py
```diff
@@ -0,0 +1,74 @@
...
+    Behavior should be the same as `torch.Tensor` aside from the extended
+    functionality.
+
+    Copying metadata:
+        * For `c = a + b`, then the meta data will be copied from the first
+        instance of `MetaTensor`.
+    """
+
+    @staticmethod
+    def __new__(cls, x, affine: torch.Tensor | None = None, meta: dict | None = None, *args, **kwargs) -> MetaTensor:
```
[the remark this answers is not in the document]
rijobro: My black version (`22.3`, MacOS, installed with pip) does this automatically...

monai/data/meta_tensor.py
rijobro: A bit more experimenting and I think I've found that `MetaTensor().as_subclass(torch.Tensor)` does what we want.

rijobro:
Hi, think I've tried to resolve any outstanding issues. I also added testing for pickling, torchscript and amp.

rijobro:
@wyli @Nic-Ma I can't reproduce [this error](https://github.com/Project-MONAI/MONAI/runs/5873453182?check_suite_focus=true#step:9:8456) anywhere and I've tried on quite a few different machines (MacOS, Windows, 3xUbuntu). Can any of you reproduce?

wyli:
> @wyli @Nic-Ma I can't reproduce [this error](https://github.com/Project-MONAI/MONAI/runs/5873453182?check_suite_focus=true#step:9:8456) anywhere and I've tried on quite a few different machines (MacOS, Windows, 3xUbuntu). Can any of you reproduce?

I can replicate with MacOS py3.8 and with `pip install -U -r requirements-dev.txt`, let me have a look...

tests/test_meta_tensor.py
```diff
@@ -0,0 +1,232 @@
...
+        self.assertEqual(get_track_transforms(), True)
+
+    @parameterized.expand(TEST_DEVICES)
+    def test_torchscript(self, device):
+        shape = (1, 3, 10, 8)
+        im, _ = self.get_im(shape, device=device)
+        conv = torch.nn.Conv2d(im.shape[1], 5, 3)
+        conv.to(device)
+        im_conv = conv(im)
+        traced_fn = torch.jit.trace(conv, im)
```
wyli, line 189: perhaps we only support this?
```suggestion
        traced_fn = torch.jit.trace(conv, im.as_tensor())
```

monai/data/meta_tensor.py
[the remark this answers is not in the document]
wyli: this is triggered by pyupgrade when `from __future__ import annotations` is used https://github.com/asottile/pyupgrade#pep-604-typing-rewrites I think we can keep this as long as the precommit hook consistently ensures this behaviour, what do you think?

monai/data/meta_obj.py
```diff
@@ -0,0 +1,262 @@
...
+    Behavior should be the same as extended class (e.g., `torch.Tensor` or `np.ndarray`)
+    aside from the extended meta functionality.
+
+    Copying of information:
+        * For `c = a + b`, then auxiliary data (e.g., meta data) will be copied from the
+        first instance of `MetaObj`.
+    """
+
+    _meta: dict
+    _affine: Any
```
wyli, line 117: could you make this `_affine` an item in `_meta`? I think with the `affine` property and getter, it's already very easy to use. this underlying structure should be simple. or is there a reason to make `_affine` special?

monai/data/meta_obj.py
```diff
@@ -0,0 +1,262 @@
...
+
+class MetaObj:
+    """
+    Abstract base class that stores data as well as any extra meta data and an affine
+    transformation matrix.
+
+    This allows for subclassing `torch.Tensor` and `np.ndarray` through multiple
+    inheritance.
+
+    Meta data is stored in the form of of a dictionary. Affine matrices are stored in
```
wyli, line 101:
```suggestion
    Meta data is stored in the form of a dictionary. Affine matrices are stored in
```

monai/data/meta_tensor.py
```diff
@@ -0,0 +1,97 @@
...
+
+__all__ = ["MetaTensor"]
+
+
+class MetaTensor(MetaObj, torch.Tensor):
+    """
+    Class that inherits from both `torch.Tensor` and `MetaObj`, adding support for meta
+    data.
+
+    Meta data is stored in the form of of a dictionary. Affine matrices are stored in
```
wyli, line 28:
```suggestion
    Meta data is stored in the form of a dictionary. Affine matrices are stored in
```

wyli (commented):
the spell checker says "meta data" in the docstring should be "metadata" or "meta-data", please consider unifying the terms for the concepts...

monai/data/meta_obj.py
Nic-Ma: I agree with @wyli , don't see any obvious reason to split `affine` and `meta dict`.

Thanks.

wyli: I tried these APIs with more toy examples, it seems they don't work with the default collate...

rijobro: No they don't currently work with collate, but if you check my [full PR](https://github.com/rijobro/MONAI/blob/MetaTensorDev/monai/data/utils.py#L358-L361), I updated our default collation and decollation to support this. 

It's a shame that there's not a more general solution, but by the time the `default_collate` arrives in `__torch_function__`, it's `torch.stack` that's being called. We have no way of differentiating between `torch.stack` that's being called by itself versus when it's being called because of collation, so can't put that sort of logic inside the `MetaTensor` class itself.

rijobro:
I will update "meta data" -> "metadata" but "meta data" is used a lot throughout our codebase (189 times).

monai/data/meta_obj.py
rijobro: If `affine` lived inside of `meta`, then we would access it with `im.meta["affine"]`. We're back to using strings again. And if a lot of our transforms are going to update the affine, then `im.affine` is a lot more succinct and less prone to errors than `im.meta["affine"]`. I'd rather leave it as it is, but I am happy to change it if I'm outvoted.

rijobro: Also with `affine` as its own object, it's easy to ensure that the affine and the image are on the same device, which will be useful when transforms update it.

wyli: perhaps misunderstanding here... I completely agree with the `im.affine` via the property interface, the comment is about making the internal data structure more consistent (merging `_affine` into `_meta`, not `affine` into `meta`). Essentially we can have
```py
@property
def affine(self) -> Any:
    return self._meta['affine']
```

wyli: > No they don't currently work with collate, but if you check my [full PR](https://github.com/rijobro/MONAI/blob/MetaTensorDev/monai/data/utils.py#L358-L361), I updated our default collation and decollation to support this.
> 
> It's a shame that there's not a more general solution, but by the time the `default_collate` arrives in `__torch_function__`, it's `torch.stack` that's being called. We have no way of differentiating between `torch.stack` that's being called by itself versus when it's being called because of collation, so can't put that sort of logic inside the `MetaTensor` class itself.

interesting, do you think implementing an internal flag such as 'self._meta["is_batch"]' can bring more flexibilities in `__torch__function__`?

rijobro: So would the constructor be `MetaTensor(x, affine=foo, meta=bar)` or `MetaTensor(x, meta=bar)` (where the affine already lives inside of `meta`)?

If the former, what do we do if the key `affine` already exists inside of `meta`? If the latter, do we have to check inside of `meta` and if `affine` is missing then we set it to be the default? I'm struggling to see how either is cleaner.

wyli: I think we use the former, when both `meta` has an affine and `affine=foo` is provided, we raise a warning? we may want to validate the `meta=bar` against the actual `x` anyway? such as spatial shape, data type...

rijobro: Ok. By the way, I propose that we remove any redundant information, such as spatial shape and data type, from meta. Both of these can be extracted from the image directly and only serve to cause confusion when the two inevitably drift out of sync.

wyli: I see, that's great!

rijobro:
@wyli bumping min version of pytorch to 1.7 as I don't think 1.6 has the same implementation of `__torch_function__` in 1.6. That OK?

image

Any idea why `min-dep-pytorch (1.6.0)` is showing up in the list of checks even though 1.6.0 has been removed from the matrix?
image

wyli:
Yes, that's defined in the branch policy: Screenshot 2022-04-11 at 12 56 20

I think dropping pytorch 1.6 support should be done with a dedicated PR, Nic and I can work on it, please don't worry about it in this PR.

rijobro:
Ok, so you want me to revert the previous commit? But then I won't be able to get the checks to pass. Also, do you understand this error? It's related, but I don't see why MetaTensor would be created here: https://github.com/Project-MONAI/MONAI/runs/5970080871?check_suite_focus=true#step:7:22293.

wyli:
> Ok, so you want me to revert the previous commit? But then I won't be able to get the checks to pass. Also, do you understand this error? It's related, but I don't see why MetaTensor would be created here: https://github.com/Project-MONAI/MONAI/runs/5970080871?check_suite_focus=true#step:7:22293.

yes, please revert the dropping 1.6 related, Nic and I can take care of this. 

for the pytorch 1.8 issue, any idea @charliebudd? https://github.com/Project-MONAI/MONAI/runs/5970080871?check_suite_focus=true

rijobro:
@charliebudd ah ok, thanks

rijobro:
@wyli let me know when you've bumped the min pytorch version, hopefully this PR will be good to go at that point.

rijobro:
Hi @vfdev-5 have you had the chance to look this PR over? Would be great to get your insight.

Nic-Ma:
> Hi @vfdev-5 have you had the chance to look this PR over? Would be great to get your insight.

Hi @rijobro ,

I synced with @vfdev-5 offline, he tried to review this PR but is still not clear about the context and initial intent of this PR, could you please help provide some design doc or usage examples to help him better understand it?

Thanks in advance.

rijobro:
Sure @Nic-Ma. Hi @vfdev-5, we would like to attach our meta data to our images. Looking at the [pytorch docs](https://pytorch.org/docs/stable/notes/extending.html#subclassing-torch-tensor) for extending `torch.Tensor`, ours is similar in concept to the `MetadataTensor`, but in implementation it more similar to the `LoggingTensor` example. This is because we subclass `torch.Tensor` rather than wrap it so that `isinstance` still works.

Previously, we used dictionaries of data containing images, their metadata and the transforms that have been applied to them, e.g.,  `{"img": torch.Tensor, "img_meta_dict": dict, "img_transforms": list}`. Throughout our code, we sometimes hardcode the suffix `_meta_dict`, and in other places we allow the user to change the default. This gets messy, especially when we have multiple images (e.g., `img` and `seg`). 

Looking to the future, we would like our transforms to be able to update metadata when they are applied, so we need to first tidy things up, as there isn't scope for adding extra functionality as things stand.

Hence, what would have previously been:

```python
transform = LoadImaged("img", "seg")
data = {"img": some_filename, "seg": some_other_filename}
data = transform(data)
print(data)  #  {"img": ...,  "img_meta_dict": ..., "seg": ..., "seg_meta_dict": ...}
```
is now
```python
print(data)  # {"img": ..., "seg": ...}
```

and now "img" has the meta data and affine transformation matrix that can be extracted with:

```python
im = data["img"]
print(im.meta)  # dictionary with meta data
print(im.affine)  # 4x4 or 3x3  torch.Tensor
```

Now that every image has its attached affine transformation matrix, it should be much easier to make transforms that correctly update it.

We allow users to disable this functionality in case they aren't interested in it, and it is causing them bother for whatever reason.

monai/data/meta_tensor.py
```diff
@@ -0,0 +1,149 @@
...
+            import torch
+            from monai.data import MetaTensor
+
+            t = torch.tensor([1,2,3])
+            affine = torch.eye(4)
+            meta = {"some": "info"}
+            m = MetaTensor(t, affine=affine, meta=meta)
+            m2 = m+m
+            assert isinstance(m2, MetaTensor)
+            assert m2.meta == meta
```
vfdev-5, line 52: Strictly speaking m2.meta is not equal to meta as m2.meta will have "affine" key in addition.

vfdev-5:
@rijobro thanks a lot for details!

In terms of implementation, I'll check and comment out in the code.

Another point I was wondering as far as I understand tensor subclassing feature but not sure if this point is related to your idea. When subclassing torch tensor with `__torch_function__` we have a possibility to route depending on input func whether ops are changing output type (back to torch.Tensor) or keep it as a subclass.
So, in your case looks like any op keep MetaTensor as MetaTensor. For example, if we sum up it we still in the domain:
```python
t = torch.tensor([1,2,3])
affine = torch.eye(4)
meta = {"some": "info"}
m = MetaTensor(t, affine=affine, meta=meta)
res = m.sum()
print(res)
> tensor(6)
MetaData
	some: info
        [0., 1., 0., 0.],
        [0., 0., 1., 0.],
        [0., 0., 0., 1.]])
```
If this is expected and intended behaviour, please ignore this point.

monai/data/meta_tensor.py
```diff
@@ -0,0 +1,149 @@
...
+            may return a `torch.Tensor` instead of MetaTensor`.
+        - A warning will be raised if in the constructor `affine` is not `None` and
+            `meta` already contains the key `affine`.
+    """
+
+    @staticmethod
+    def __new__(cls, x, affine: torch.Tensor | None = None, meta: dict | None = None, *args, **kwargs) -> MetaTensor:
+        return torch.as_tensor(x, *args, **kwargs).as_subclass(cls)  # type: ignore
+
+    def __init__(self, x, affine: torch.Tensor | None = None, meta: dict | None = None) -> None:
```
vfdev-5, line 68: I may be wrong but I think we could implement that using only `__new__` ?

wyli:
>  When subclassing torch tensor with `__torch_function__` we have a possibility to route depending on input func whether ops are changing output type (back to torch.Tensor) or keep it as a subclass. 

that's a good point, perhaps we need this to differentiate w/o the batch dim for the collate...

monai/data/meta_tensor.py
rijobro: good point, thanks!

monai/data/meta_tensor.py
rijobro: yup! 👍

wyli (approved):
thanks, it looks good to me, I put some minor docstring updates here https://github.com/rijobro/MONAI/pull/3

monai/data/meta_obj.py
```diff
@@ -0,0 +1,207 @@
...
+            out += "".join(f"\t{k}: {v}\n" for k, v in self.meta.items())
+        else:
+            out += "None"
+
+        return out
+
+    @property
+    def meta(self) -> dict:
+        """Get the meta."""
+        return self._meta
```
Nic-Ma, line 202: Hi @rijobro ,

Thanks for the good idea to construct `meta` property.
If `self._meta` is not initialized yet, I suggest to give a clear error message or return `get_default_meta()`?
What do you think?
Current error is hard to understand:
```py
class Test:
    meta: dict

    @property
    def meta(self) -> dict:
        """Get the meta."""
        return self._meta

    @meta.setter
    def meta(self, d: dict) -> None:
        """Set the meta."""
        self._meta = d


a = Test()
print(a.meta)
```
```
  File "/workspace/PythonTests/main.py", line 16, in <module>
    print(a.meta)
  File "/workspace/PythonTests/main.py", line 7, in meta
    return self._meta
AttributeError: 'Test' object has no attribute '_meta'
```

Thanks.

monai/data/meta_tensor.py
```diff
@@ -0,0 +1,148 @@
...
+                determined using `PostFix.meta`.
+
+        Return:
+            A dictionary consisting of two keys, the main data (stored under `key`) and
+                the metadata.
+        """
+        return {key: self.as_tensor(), PostFix.meta(key): self.meta}
+
+    @property
+    def affine(self) -> torch.Tensor:
```
Nic-Ma, line 141: Hi @rijobro ,

May I know why change `affine` property from `MetaObj` to `MetaTensor`? Seems it's not `Tensor` specific logic?
I didn't read all the comments and conversations, maybe I missed some points.

Thanks in advance.

PR #4107, opened 2022-04-12
Redesign whole slide image reading

### LINKED ISSUE #4071: Redesign of WSIReader
Right now `WSIReader` is based on `ImageReader` class with `read` and `get_data` methods. However, reading whole slide images regardless of the library to use have its own additional complexities that are not captured in the base class. 
Also, currently, we have implemented different backends `cuCIM`, `OpenSlide`, and TiffFile` through if statements!

My suggestion would be to have a `WSIReader` base class that implements those additional functionalities such as evaluating the level based on ppm/resolution, validating the pyramidal levels, etc. along with abstract methods that each of the libraries has to implement. In this way, we can implement each backend in different classes like `CuCIMReader` (for cuCIM) inherited from `WSIReader`.

This not only makes the code cleaner and more readable but also decouples MONAI integration of third-party WSI reader libraries and makes the future contribution easier and safer.

The overall structure would look like this: (will update #4005)
wsi

### LINKED ISSUE #4098: `WSIReader` to return patch information
Currently, `WSIReader` does not return the information about the patch and the whole slide image that is being loaded in metadata when calling `get_data` method. This information can be helpful in the following preprocessing and post-processing transforms, for instance to generate attention maps as requested here: #4090

drbeh:
Fixes #4071 
Fixes #4098

### Description
This PR is a re-design and re-implementation of `WSIReader` while keeping the API the same as before and provide full backward compatibility (support for all the backends will come in the next PRs). This PR implements `BaseWSIReader`, `WSIReader`, and `CuCIMWSIReader`.

### Types of changes


monai/data/wsi_reader.py
```diff
@@ -0,0 +1,310 @@
...
+        Args:
+            data: file name or a list of file names to read.
+            kwargs: additional args for backend reading API in `read()`, will override `self.kwargs` for existing keys.
+                more details in `cuCIM`: https://github.com/rapidsai/cucim/blob/v21.12.00/cpp/include/cucim/cuimage.h#L100.
+
+        Returns:
+            image object or list of image objects
+
+        """
+        wsi_list: List = []
```
Nic-Ma, line 107: Is there any use case for a list of WSI files?
I think the `get_data()` API doesn't support that?

Thanks.

monai/data/wsi_reader.py
```diff
@@ -0,0 +1,310 @@
...
+        if patch.ndim != 3:
+            raise ValueError(
+                f"The image dimension should be 3 but has {patch.ndim}. "
+                "`WSIReader` is designed to work only with 2D images with color channel."
+            )
+
+        return patch
+
+
+class WSIReader(BaseWSIReader):
```
Nic-Ma, line 201: I feel we can remove or deprecate this `WSIReader` class directly, no need to maintain it anymore if we define `CuCIMWSIReader` and other subclass readers.
@wyli @ericspod what do you think?

Thanks in advance.

monai/data/wsi_reader.py
drbeh: That's a good point! I will simplify it.

monai/data/wsi_reader.py
drbeh: I believe it's good to keep it here. It provides not only some backward compatibilities but also a simpler interface for people who doesn't know about underlying libraries like cucim, openslide, etc.

monai/data/wsi_reader.py
drbeh: Actually we shouldn't check it. This class is inherited from `ImageReader`, and it expect filename or list of filenames, so we should not make this argument more specific. It is violation of Liskov substitution principle.

https://github.com/Project-MONAI/MONAI/blob/443fc0cf11e644d1b85ae2a6a43e7820ecdc604c/monai/data/image_reader.py#L79

monai/data/wsi_reader.py
```diff
@@ -0,0 +1,288 @@
...
+            self.backend_lib = AnyBackendWSIReader(level=level, **kwargs)
+
+    """
+
+    def __init__(self, backend="cucim", level: int = 0, **kwargs):
+        super().__init__(level, **kwargs)
+        self.backend = backend.lower()
+        # Any new backend can be added below
+        if self.backend == "cucim":
+            self.backend_lib = CuCIMWSIReader(level=level, **kwargs)
```
Nic-Ma, line 182: I would suggest to change the arg name `backend_lib` to `reader` or `backend_reader`.
What do you think?

Thanks.

monai/data/wsi_reader.py
```diff
@@ -0,0 +1,288 @@
...
+    The support for any backend can be achieved by
+
+    .. code-block:: python
+
+        if self.backend == "any_backend":
+            self.backend_lib = AnyBackendWSIReader(level=level, **kwargs)
+
+    """
+
+    def __init__(self, backend="cucim", level: int = 0, **kwargs):
```
Nic-Ma, line 177: Although this class is kind of `wrapper`, I still suggest to add doc-strings for every function, because user will only check the doc-strings of this class when using it.

Thanks.

monai/data/wsi_reader.py
Nic-Ma: OK, I see your point. Let's keep it for now.

Thanks.

monai/data/wsi_reader.py
Nic-Ma: Hi @drbeh ,

I understand the principle point. The problem is that: the other readers can `read()` a list of images and `get_data()` also supports a list of image objects:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/image_reader.py#L99
but here your `get_data()` function doesn't(or no need) work with a list of image objects:
https://github.com/Project-MONAI/MONAI/pull/4107/files#diff-e0abb01781c5bed5c90cf46759d8e2aeac0793b86bd48ab49b40b6b7f7f5c1c2R94
How about this:
1. Keep your current API and args.
2. Add a check in the `get_data()`: if passing `wsi` object, ok, if passing a list of `wsi` objects, should only contains one item in it and use it. For example:
```py
def get_data(self, wsi, ...):
    if isinstance(wsi, (tuple, list)):
        if len(wsi) != 1:
            raise ValueError("only support 1 `wsi` image object so far.")
        wsi = wsi[0]
```

What do you think?

Thanks.

drbeh: Hi @Nic-Ma, thanks. That's a good idea but I can make them to load the same patch from multiple wsi objects. I think this way would be more compatible with the rest of readers.

drbeh: Ah, the output of `get_data` is also too specific! I'll go with input checking.

drbeh:
Hi @Nic-Ma, do you have any other comments here? Thanks

tests/test_wsireader.py
```diff
@@ -155,17 +143,22 @@ def test_read_region(self, file_path, patch_info, expected_img):
...
+            kwargs = {"name": None, "offset": None} if self.backend == "tifffile" else {}
+            reader = WSIReader(self.backend, **kwargs)
+            img_obj = reader.read(file_path, **kwargs)
+            if self.backend == "tifffile":
+                with self.assertRaises(ValueError):
+                    reader.get_data(img_obj, **patch_info)[0]
+            else:
+                # Read twice to check multiple calls
+                img = reader.get_data(img_obj, **patch_info)[0]
+                img2 = reader.get_data(img_obj, **patch_info)[0]
```
Nic-Ma, line 157: Good test case!

Thanks.

monai/data/wsi_reader.py
```diff
@@ -0,0 +1,428 @@
...
+        size: Optional[Tuple[int, int]] = None,
+        level: Optional[int] = None,
+        dtype: DtypeLike = np.uint8,
+        mode: str = "RGB",
+    ):
+        """
+        Verifies inputs, extracts patches from WSI image and generates metadata, and return them.
+
+        Args:
+            wsi: a whole slide image object loaded from a file or a lis of such objects
```
Nic-Ma, line 130: typo: `lis` -> `list`

monai/data/wsi_reader.py
```diff
@@ -0,0 +1,428 @@
...
+            mode: the output image mode, 'RGB' or 'RGBA'
+
+        Returns:
+            a tuples, where the first element is an image patch [CxHxW] or stack of patches,
+                and second element is a dictionary of metadata
+        """
+        patch_list: List = []
+        metadata = {}
+        # CuImage object is iterable, so ensure_tuple won't work on single object
+        if not isinstance(wsi, List):
```
Nic-Ma, line 145: I think no need these 2 lines, `ensure_tuple()` can automatically support a single object, just like your code:
https://github.com/Project-MONAI/MONAI/pull/4107/files#diff-e0abb01781c5bed5c90cf46759d8e2aeac0793b86bd48ab49b40b6b7f7f5c1c2R382

Thanks.

monai/data/wsi_reader.py
```diff
@@ -0,0 +1,428 @@
...
+
+        Args:
+            wsi: a whole slide image object loaded from a file
+            level: the level number where the size is calculated
+
+        """
+        raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
+
+    @abstractmethod
+    def get_level_count(self, wsi):
```
Nic-Ma, line 75: Could you please help add the `typehints` or `doc-string` to describe the return data/type of these interfaces or also in the subclass?
`get_level_count`, `get_patch`, `get_metadata`, `get_data`?

Thanks.

monai/data/wsi_reader.py
```diff
@@ -0,0 +1,428 @@
...
+
+        The list of supported suffixes are read from `self.supported_formats`.
+
+        Args:
+            filename: filename or a list of filenames to read.
+
+        """
+        return is_supported_format(filename, self.supported_formats)
+
+    def verify_output(self, patch: np.ndarray, mode: str):
```
Nic-Ma, line 198: I feel we don't need this function because it's simple and only used once, maybe just put this check in `get_data()` directly?

Thanks.

monai/data/wsi_reader.py
Nic-Ma: Could you please help also add the doc-strings of `args` for this init function? `backend` and `level`, `kwargs`.

Thanks.

monai/data/wsi_reader.py
```diff
@@ -0,0 +1,428 @@
...
+    - `read` reads a whole slide image object from a given file
+    - `get_size` returns the size of the whole slide image of a given wsi object at a given level.
+    - `get_level_count` returns the number of levels in the whole slide image
+    - `get_patch` extracts and returns a patch image form the whole slide image
+    - `get_metadata` extracts and returns metadata for a whole slide image and a specific patch.
+
+
+    """
+
+    supported_formats: List[str] = []
```
Nic-Ma, line 54: I would suggest to use `supported_ suffixes` to be consistent with other readers and the function name:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/image_reader.py#L387
And I think you are making a good change to put the `supported_suffixes` as a class member, I may also change all the other readers soon.

Thanks.

Nic-Ma (commented):
Hi @drbeh ,

Thanks for your quick update!
I put some minor comments inline, others look good to me.

Thanks.

monai/data/wsi_reader.py
drbeh: These lines are essential since `ensure_tuple` gets confused when the object itself is iterable.

Nic-Ma: Oh, I see. I didn't know the `wsi` is an iterable object.

Thanks for the explanation.

drbeh:
Hi @Nic-Ma, thanks for your comments. I have addressed them all and will merge this PR but please let me know if you have any comment and I will include it in the following PRs.

Nic-Ma:
Cool, thanks for your quick update.
Looks good to me.

PR #4109, opened 2022-04-12
Make `pixelshuffle` scriptable

### LINKED ISSUE #4108: Make `pixelshuffle` scriptable.

Modify the `pixelshuffle` method in such a way that when using this upsampling method on any network, it can be scripted using `torch.jit.script`

This should not raise an error:
```python
import torch
from monai.networks.utils import pixelshuffle

torch.jit.script(pixelshuffle)
```

ramonemiliani93:
Signed-off-by: Ramon Emiliani <ramon.emiliani@umontreal.ca>

Fixes #4108.

### Description
Modifies the existing `pixelshuffle` function in such a way that it is scriptable using `torch.jit.script`.

### Status
**Work in progress**

### Types of changes

### DIFF
--- a/monai/networks/utils.py
+++ b/monai/networks/utils.py
@@ -281,14 +281,16 @@ def pixelshuffle(
             f"divisible by scale_factor ** dimensions ({factor}**{dim}={scale_divisor})."
         )
 
-    org_channels = channels // scale_divisor
+    org_channels = int(channels // scale_divisor)
     output_size = [batch_size, org_channels] + [d * factor for d in input_size[2:]]
 
-    indices = tuple(range(2, 2 + 2 * dim))
-    indices_factor, indices_dim = indices[:dim], indices[dim:]
-    permute_indices = (0, 1) + sum(zip(indices_dim, indices_factor), ())
+    indices = list(range(2, 2 + 2 * dim))
+    indices = indices[dim:] + indices[:dim]
+    permute_indices = [0, 1]
+    for idx in range(dim):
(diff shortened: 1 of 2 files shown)


monai/networks/utils.py
```diff
@@ -281,14 +281,16 @@ def pixelshuffle(
             f"divisible by scale_factor ** dimensions ({factor}**{dim}={scale_divisor})."
         )
 
-    org_channels = channels // scale_divisor
+    org_channels = int(channels // scale_divisor)
```
wyli, line 284: thanks, could you help add a test case like this one https://github.com/Project-MONAI/MONAI/blob/1516ca758090dd8ed2890b64554a77f681a0e224/tests/test_se_blocks.py#L68-L72 for the subpixel upsampling https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_subpixel_upsample.py

ramonemiliani93: Done @wyli 👌

ramonemiliani93:
@wyli Seems `range` is not scriptable on PyTorch 1.7 and the test fails, what do you think we should do in this case?

wyli:
> @wyli Seems `range` is not scriptable on PyTorch 1.7 and the test fails, what do you think we should do in this case?

was going to mention this, I added the condition to skip https://github.com/Project-MONAI/MONAI/pull/4109/commits/026f840a531be6749216b0fc62c87045826c7331

wyli (approved):
Thanks, it looks good to me.

Issue #4094, opened 2022-04-08
Checkpoint Export Functionality Should Require Bundle Directory Path Only


ericspod:
**Is your feature request related to a problem? Please describe.**
Currently to export a bundle as a Torchscript file one has to specify the files to include individually, eg:

```bash
python -m monai.bundle ckpt_export network_def \
    --filepath foo.ts \
    --ckpt_file foo/models/model.pt \
    --meta_file foo/configs/metadata.json \
    --config_file foo/configs/inference.json
```

**Describe the solution you'd like**
I think the optimal behaviour is to specify export of a whole bundle directory which will include all the config files in the `configs` directory and use `models/model.pt` as the model weight file implicitly. The network definition `network_def` is in config files in the tutorials example  but perhaps it should be moved to the `metadata.json` file since there will really only be the one primary network definition per bundle. The actual command line to export would be :

```bash
python -m monai.bundle ckpt_export --bundle_path /path/to/bundle/dir --out_path foo.ts 
```

 (I would also suggest changing `filepath` to `out_path` to be clearer about what it is.)

**Describe alternatives you've considered**
The current command works just fine so this is a QoL improvement.

Nic-Ma:
Thanks for your feedback, maybe let's try to collect feedback from first users and adjust APIs of bundle soon.

ericspod:
Also we should allow any number of json files to be stored, so anything that's in the configs directory gets kept in the exported Torchscript file without their names being changed, currently the config file gets stored as "config". I think we want to support using downloaded models that have multiple configs in them with other software which takes as a parameter a config name to use to do something.

Nic-Ma:
Hi @ericspod ,

Currently, the `ckpt_export` API can support multiple config files but will merge them together and save with "config":
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/scripts.py#L374
So you mean we should save the configs separately instead of merging them before exporting?

Thanks.

ericspod:
I would suggest we should have them as separate names so that the user can choose one or the other depending on the task. The files should also retain their extensions since the parser relies on these to determine if they're json or yaml. 

A separate enhance that would be help is being able to load metadata and configuration from the zip file directly, I'm currently having to save files out to a temporary directory and read them from there but we can have a cleaner process I think.

Nic-Ma:
Hi @ericspod ,

[...]

PR #4114, opened 2022-04-12
[WIP] 4095 Add bundle download

### LINKED ISSUE #4095: Checkpoint Download Automation for Bundles
The example bundle uses `download_ckpt` to specify how to download the checkpoint for the bundle from Huggingface. This happens when the config file it's specified is used but doesn't when doing other operations like `verify_net_in_out`. 

The specification for the URL of a checkpoint file should be made a formal concept in the metadata specification so that it's downloaded as an automated step whenever a bundle is used for anything. This would keep the tutorials repo clean of large objects and make it easier for users to get the whole bundle without having to download the dataset it's paired with. If users don't want to train or do inference but use the bundle for something else this would be easier. Since the specification makes the file `models/model.pt` an explicitly-defined file which is always present we can define a `ckpt_url` value in the metadata referring to a file to be downloaded as stored in that location if the file is missing.

I downloaded the checkpoint manually by copy-pasting the code snippet to do the download. This wasn't a mountanous workload but it would make automated operations on bundles easier to have explicit mechanisms to do this. 

**Additional context**
The idea is to automate preparing bundles for packaging into other contexts such as a MONAI Deploy operator, so the weights have to be present for this to be done and if they need downloading that is a manual step at present.

yiheng-wang-nv:
Fixes #4095 .

### Description
unit test need to be added.

### Status
**Work in progress**

### Types of changes

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -11,6 +11,7 @@
 
 import ast
 import json
+import os
 import pprint
 import re
 from logging.config import fileConfig
@@ -19,7 +20,7 @@ from typing import Dict, Optional, Sequence, Tuple, Union
 import torch
 from torch.cuda import is_available
 
-from monai.apps.utils import download_url, get_logger
+from monai.apps.utils import download_url, extractall, get_logger
 from monai.bundle.config_parser import ConfigParser
 from monai.config import IgniteInfo, PathLike
 from monai.data import save_net_with_metadata
@@ -29,6 +30,7 @@ from monai.utils import check_parent_dir, get_equivalent_dtype, min_version, opt
 validate, _ = optional_import("jsonschema", name="validate")
 ValidationError, _ = optional_import("jsonschema.exceptions", name="ValidationError")
 Checkpoint, has_ignite = optional_import("ignite.handlers", IgniteInfo.OPT_IMPORT_VERSION, min_version, "Checkpoint")
+requests_get, has_requests = optional_import("requests", name="get")
 
 logger = get_logger(module_name=__name__)
 
@@ -114,6 +116,75 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
         else:
             raise ValueError(f"spatial shape items must be int or string, but got: {type(i)} {i}.")
     return tuple(ret)
+
+
+def _get_git_release_url(
+    repo_owner: str,
+    repo_name: str,
+    tag_name: str,
+    filename: Optional[str] = None,
+):
+    if filename is not None:
+        return f"https://github.com/{repo_owner}/{repo_name}/releases/download/{tag_name}/{filename}"
+    else:
+        raise NotImplementedError("download the whole package is not implemented so far.")
+
+
+def download(
+    repo: str,
+    package: str,
+    filename: Optional[str] = None,
+    source: str = "github",
+    download_path: str = "download",
+    hash_val: Optional[str] = None,
+    hash_type: str = "md5",
+    extract: bool = False,
+    has_base: bool = True,
+    version: int = 1,
+    # args_file
+):
+    """
+    download the bundle package or a file that belongs to the package from the specified source.
+    This function refers to:
+    https://pytorch.org/docs/stable/_modules/torch/hub.html
+
+    Args:
+        repo: the bundle package. The format depends on the source. If the source is `github`,
+            it should be in the form of `repo_owner/repo_name`. If the source is `ngc`, it should
+            be in the form of `org/team`.
+        package: the bundle package name. If the source is `github`, it should be the same as the
+            release tag.
+        filename: the filename of the bundle package that needs to be downloaded. It is an optional
+            argument and if not specified, the whole bundle package will be downloaded.
+        source: the place that saved the bundle package. So far, only `github` and `ngc` are supported.
+            For the `github` source, the bundle package should be within the releases.
+        download_path: target filepath to save the downloaded file (including the filename).
+            If undefined, `os.path.basename(url)` will be used.
+        hash_val: expected hash value to validate the downloaded file.
+            if None, skip hash validation.
+        hash_type: 'md5' or 'sha1'.
+        extract: whether to extract the downloaded file.
+        output_dir: target directory to save extracted files.
+        has_base: whether the extracted files have a base folder. This flag is used when checking if the existing
+            folder is a result of `extractall`, if it is, the extraction is skipped. For example, if A.zip is unzipped
+            to folder structure `A/*.png`, this flag should be True; if B.zip is unzipped to `*.png`, this flag should
+            be False.
+        version: this argument only works on ngc souce, and it represents the version of the model.
+
+    """
+    if source == "github":
+        if len(repo.split("/")) != 2:
+            raise ValueError("if source is `github`, repo should be in the form of `repo_owner/repo_name`.")
+        repo_owner, repo_name = repo.split("/")
+        url = _get_git_release_url(repo_owner, repo_name, tag_name=package, filename=filename)
+        filepath = os.path.join(download_path, filename)
+    elif source == "ngc":
+        # to be modified
+        url = f"https://api.ngc.nvidia.com/v2/models/{repo}/{package}/versions/{version}/zip"
+        filepath = os.path.join(download_path, f"{package}.zip")
+    download_url(url=url, filepath=filepath, hash_val=hash_val, hash_type=hash_type)
+    if extract is True:
+        extractall(filepath=filepath, output_dir=download_path, has_base=has_base)
 
 
 def run(
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -11,10 +11,10 @@
 
 import ast
 import json
-import os
 import pprint
 import re
 from logging.config import fileConfig
+from pathlib import Path
 from typing import Dict, Optional, Sequence, Tuple, Union
 
 import torch
@@ -122,26 +122,63 @@ def _get_git_release_url(
     repo_owner: str,
     repo_name: str,
     tag_name: str,
-    filename: Optional[str] = None,
+    filename: str,
 ):
-    if filename is not None:
-        return f"https://github.com/{repo_owner}/{repo_name}/releases/download/{tag_name}/{filename}"
+    return f"https://github.com/{repo_owner}/{repo_name}/releases/download/{tag_name}/{filename}"
+
+
+def _get_git_release_assets(
+    repo_owner: str,
+    repo_name: str,
+    tag_name: str,
+):
+    url = f"https://api.github.com/repos/{repo_owner}/{repo_name}/releases/tags/{tag_name}"
+    resp = requests_get(url)
+    resp.raise_for_status()
+    assets_list = json.loads(resp.text)["assets"]
+    assets_info = {}
+    for asset in assets_list:
+        assets_info[asset["name"]] = asset["browser_download_url"]
+    return assets_info
+
+
+def _download_from_github(
+    repo: str,
+    tag_name: str,
+    bundle_dir: Path,
+    filename: Optional[str] = None,
+    progress: bool = True,
+    extract: bool = False,
+):
+    if len(repo.split("/")) != 2:
+        raise ValueError("if source is `github`, repo should be in the form of `repo_owner/repo_name`.")
+    repo_owner, repo_name = repo.split("/")
+    if filename is None:
+        # download the whole bundle package if filename is not provided
+        assets_info = _get_git_release_assets(repo_owner, repo_name, tag_name=tag_name)
+        for name, url in assets_info.items():
+            download_url(url=url, filepath=bundle_dir / f"{name}", hash_val=None, progress=progress)
+        if extract is True:
+            logger.info("When download a whole bundle package, extract is not supported, skip extracting.")
+        logger.info(f"All files within the bundle package {tag_name} are downloaded.")
     else:
-        raise NotImplementedError("download the whole package is not implemented so far.")
+        # download a single file
+        url = _get_git_release_url(repo_owner, repo_name, tag_name=tag_name, filename=filename)
+        filepath = bundle_dir / f"{filename}"
+        download_url(url=url, filepath=filepath, hash_val=None, progress=progress)
+        if extract is True:
+            extractall(filepath=filepath, output_dir=bundle_dir, has_base=False)
 
 
 def download(
-    repo: str,
-    package: str,
+    repo: Optional[str] = None,
+    package: Optional[str] = None,
+    bundle_dir: Optional[PathLike] = None,
     filename: Optional[str] = None,
     source: str = "github",
-    download_path: str = "download",
-    hash_val: Optional[str] = None,
-    hash_type: str = "md5",
+    progress: bool = True,
     extract: bool = False,
-    has_base: bool = True,
-    version: int = 1,
-    # args_file
+    args_file: Optional[str] = None,
 ):
     """
     download the bundle package or a file that belongs to the package from the specified source.
@@ -149,42 +186,62 @@ def download(
     https://pytorch.org/docs/stable/_modules/torch/hub.html
 
     Args:
-        repo: the bundle package. The format depends on the source. If the source is `github`,
-            it should be in the form of `repo_owner/repo_name`. If the source is `ngc`, it should
-            be in the form of `org/team`.
-        package: the bundle package name. If the source is `github`, it should be the same as the
-            release tag.
+        repo: the repo name. The format depends on the source. If `None`, must be provided in `args_file`.
+            If the source is `github`, it should be in the form of `repo_owner/repo_name`.
+            For example: `Project-MONAI/MONAI`.
+        package: the bundle package name. If `None`, must be provided in `args_file`.
+            If the source is `github`, it should be the same as the release tag.
+        bundle_dir: target directory to store the download data.
+            Default is `bundle` subfolder under `torch.hub get_dir()`.
+            If undefined, `os.path.basename(url)` will be used.
         filename: the filename of the bundle package that needs to be downloaded. It is an optional
             argument and if not specified, the whole bundle package will be downloaded.
         source: the place that saved the bundle package. So far, only `github` and `ngc` are supported.
             For the `github` source, the bundle package should be within the releases.
-        download_path: target filepath to save the downloaded file (including the filename).
-            If undefined, `os.path.basename(url)` will be used.
-        hash_val: expected hash value to validate the downloaded file.
-            if None, skip hash validation.
-        hash_type: 'md5' or 'sha1'.
-        extract: whether to extract the downloaded file.
-        output_dir: target directory to save extracted files.
-        has_base: whether the extracted files have a base folder. This flag is used when checking if the existing
-            folder is a result of `extractall`, if it is, the extraction is skipped. For example, if A.zip is unzipped
-            to folder structure `A/*.png`, this flag should be True; if B.zip is unzipped to `*.png`, this flag should
-            be False.
-        version: this argument only works on ngc souce, and it represents the version of the model.
+        progress: whether to display a progress bar.
+        extract: whether to extract the downloaded file. This argument only works when download a single file.
+        args_file: a JSON or YAML file to provide default values for all the args in this function.
+            so that the command line inputs can be simplified.
 
     """
-    if source == "github":
-        if len(repo.split("/")) != 2:
-            raise ValueError("if source is `github`, repo should be in the form of `repo_owner/repo_name`.")
-        repo_owner, repo_name = repo.split("/")
-        url = _get_git_release_url(repo_owner, repo_name, tag_name=package, filename=filename)
-        filepath = os.path.join(download_path, filename)
-    elif source == "ngc":
-        # to be modified
-        url = f"https://api.ngc.nvidia.com/v2/models/{repo}/{package}/versions/{version}/zip"
-        filepath = os.path.join(download_path, f"{package}.zip")
-    download_url(url=url, filepath=filepath, hash_val=hash_val, hash_type=hash_type)
-    if extract is True:
-        extractall(filepath=filepath, output_dir=download_path, has_base=has_base)
+    _args = _update_args(
+        args=args_file,
+        repo=repo,
+        package=package,
+        bundle_dir=bundle_dir,
+        filename=filename,
+        source=source,
+        progress=progress,
+        extract=extract,
+    )
+    if "repo" not in _args:
+        raise ValueError(f"`repo` is required for 'monai.bundle download'.\n{download.__doc__}")
+    if "package" not in _args:
+        raise ValueError(f"`package` is required for 'monai.bundle download'.\n{download.__doc__}")
+    _log_input_summary(tag="download", args=_args)
+    repo_, package_, bundle_dir_, filename_, source_, progress_, extract_ = _pop_args(
+        _args, "repo", "package", bundle_dir=None, filename=None, source="github", progress=True, extract=False
+    )
+
+    if not bundle_dir_:
+        get_dir, has_home = optional_import("torch.hub", name="get_dir")
+        if has_home:
+            bundle_dir_ = Path(get_dir()) / "bundle"
+        else:
+            raise ValueError("bundle_dir=None, but no suitable default directory computed. Upgrade Pytorch to 1.6+ ?")
+    bundle_dir_ = Path(bundle_dir_)
+
+    if source_ == "github":
+        _download_from_github(
+            repo=repo_,
+            tag_name=package_,
+            bundle_dir=bundle_dir_,
+            filename=filename_,
+            progress=progress_,
+            extract=extract_,
+        )
+    else:
+        raise NotImplementedError("So far, only `github` source is supported.")
 
 
 def run(
--- a/docs/source/bundle.rst
+++ b/docs/source/bundle.rst
@@ -36,6 +36,7 @@ Model Bundle
 `Scripts`
 ---------
 .. autofunction:: ckpt_export
+.. autofunction:: download
 .. autofunction:: run
 .. autofunction:: verify_metadata
 .. autofunction:: verify_net_in_out
--- a/monai/bundle/__init__.py
+++ b/monai/bundle/__init__.py
@@ -12,5 +12,5 @@
 from .config_item import ComponentLocator, ConfigComponent, ConfigExpression, ConfigItem, Instantiable
 from .config_parser import ConfigParser
 from .reference_resolver import ReferenceResolver
-from .scripts import ckpt_export, run, verify_metadata, verify_net_in_out
+from .scripts import ckpt_export, download, run, verify_metadata, verify_net_in_out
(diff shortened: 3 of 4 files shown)


monai/bundle/scripts.py
```diff
@@ -116,6 +121,212 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+        else:
+            raise ValueError("bundle_dir=None, but no suitable default directory computed. Upgrade Pytorch to 1.6+ ?")
+    bundle_dir_ = Path(bundle_dir_)
+
+    if url_ is not None:
+        if filename_ is not None:
+            filepath = bundle_dir_ / f"{filename_}"
+            download_url(url=url_, filepath=filepath, hash_val=None, progress=progress_)
+            if extract_ is True:
+                extractall(filepath=filepath, output_dir=bundle_dir_, has_base=True)
```
Nic-Ma, line 258: May I know in which case we should extract a downloaded file from a bundle package?

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,212 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+        # download a single file
+        url = _get_git_release_url(repo_owner, repo_name, tag_name=tag_name, filename=filename)
+        filepath = download_path / f"{filename}"
+        download_url(url=url, filepath=filepath, hash_val=None, progress=progress)
+        if extract is True:
+            extractall(filepath=filepath, output_dir=download_path, has_base=True)
+
+
+def download(
+    package: Optional[str] = None,
```
Nic-Ma, line 177: I would suggest to use `name` as the arg name instead of `package`.

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,212 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+        raise ValueError(f"To download from source: {source}, `package` must be provided.")
+    if "repo" not in _args and url is None:
+        raise ValueError(f"To download from source: {source}, `repo` must be provided.")
+    _log_input_summary(tag="download", args=_args)
+    package_, bundle_dir_, filename_, source_, repo_, url_, progress_, extract_ = _pop_args(
+        _args, package=None, bundle_dir=None, filename=None, source="github", repo=None, url=None, progress=True, extract=False
+    )
+
+    if not bundle_dir_:
+        get_dir, has_home = optional_import("torch.hub", name="get_dir")
```
Nic-Ma, line 246: I think no need to consider the case that no `bundle_dir` provided so far.

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,212 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+        url = _get_git_release_url(repo_owner, repo_name, tag_name=tag_name, filename=filename)
+        filepath = download_path / f"{filename}"
+        download_url(url=url, filepath=filepath, hash_val=None, progress=progress)
+        if extract is True:
+            extractall(filepath=filepath, output_dir=download_path, has_base=True)
+
+
+def download(
+    package: Optional[str] = None,
+    bundle_dir: Optional[PathLike] = None,
```
Nic-Ma, line 178: I would suggest to use `filepath` as the arg directly, same as:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/utils.py#L150

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,212 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+    filename: Optional[str] = None,
```
Nic-Ma, line 179: I think no need to specifiy this arg, the `name` arg can support the whole bundle or a single file in the bundle, like: `spleen_segmentation#train.json`.
@wyli What do you think?

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,212 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+            download_path=bundle_dir_,
+            filename=filename_,
+            progress=progress_,
+            extract=extract_,
+        )
+    else:
+        raise NotImplementedError("So far, only support to download from url or `github` source.")
+
+
+def load(
```
Nic-Ma, line 276: Same comment as `download` API, maybe combine `weights_name` and `package` to `name`?

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,212 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+            url=url,
+            progress=progress,
+            extract=extract,
+        )
+    # loading with `torch.jit.load`
+    if is_ts_model is True:
+        return torch.jit.load(model_file_path, map_location=map_location)
+    # loading with `torch.load`
+    model_dict = torch.load(model_file_path, map_location=map_location)
+    return model_dict
```
Nic-Ma, line 327: I think we can also support to return a instantiated model object with additional `net_name` and `kwargs`, similar to the huggingface torchhub API:
https://github.com/huggingface/transformers/blob/main/hubconf.py#L105

Thanks.

monai/bundle/scripts.py
yiheng-wang-nv: Thanks Nic, I removed the arg and add the extract logic within the code by default.

monai/bundle/scripts.py
yiheng-wang-nv: Thanks, I think it is helpful in `def load`, but no need to put in `def download`

yiheng-wang-nv:
Hi @Nic-Ma @ericspod @wyli , to do the unit test, I added a new release in: https://github.com/Project-MONAI/MONAI-extra-test-data/releases/tag/test_bundle
I'm not sure if it is a good way, and do you have any suggestions? Thanks!

monai/bundle/scripts.py
```diff
@@ -116,6 +121,216 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+            download_url(url=url, filepath=filepath, hash_val=None, progress=progress)
+            if filepath.name.endswith(("zip", "tar", "tar.gz")):
+                extractall(filepath=filepath, output_dir=download_path, has_base=True)
+        logger.info(f"All files within the bundle {tag_name} are downloaded.")
+    else:
+        # download a single file
+        url = _get_git_release_url(repo_owner, repo_name, tag_name=tag_name, filename=filename)
+        filepath = download_path / f"{filename}"
+        download_url(url=url, filepath=filepath, hash_val=None, progress=progress)
+        if filepath.name.endswith(("zip", "tar", "tar.gz")):
```
Nic-Ma, line 159: Same as above.

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,216 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+
+    .. code-block:: bash
+
+        # Execute this module as a CLI entry, and download the whole bundle:
+        python -m monai.bundle download --name "bundle_name" --source "github" --repo "repo_owner/repo_name"
+
+        # Execute this module as a CLI entry, and download a single file:
+        python -m monai.bundle download --name "bundle_name#filename" --repo "repo_owner/repo_name"
+
+        # Execute this module as a CLI entry, and download a single file via URL:
```
Nic-Ma, line 188: Is it possible to support downloading the whole bundle from a URL?

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,216 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+        # Other args still can override the default args at runtime:
+        python -m monai.bundle download --args_file "/workspace/data/args.json" --repo "repo_owner/repo_name"
+
+    Args:
+        name: the bundle name. If `None` and `url` is `None`, it must be provided in `args_file`.
+            If `source` is `github`, it should be the same as the release tag.
+            If only a single file is expected to be downloaded, `name` should be a string that consisted with
+            the bundle name, a separator `#` and the weights name. For example: if the bundle name is "spleen" and
+            the weights name is "model.pt", then `name` is "spleen#model.pt".
+        bundle_dir: target directory to store the download data.
```
Nic-Ma, line 201: downloaded data.

ericspod:
The release for the test data has the right files but the bundle specification requires things to be in specific subdirectories, the alternative is to put the files into the right directory and then zip it all together into an archive file which would contain the README.md and other files and subdirectories in its root.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,216 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+        else:
+            raise ValueError("bundle_dir=None, but no suitable default directory computed. Upgrade Pytorch to 1.6+ ?")
+    bundle_dir = Path(bundle_dir)
+
+    weights_name: str = "model.pt"
+    if name is not None:
+        if len(name.split(ID_SEP_KEY)) == 2:
+            weights_name = name.split(ID_SEP_KEY)[1]
+        else:
+            raise ValueError(f"The format of `name` is wrong.\n{run.__doc__}")
```
Nic-Ma, line 308: This logic is not correct and your unit tests don't cover it.
If the `name` is the whole bundle name without `#`, `len(name.split(ID_SEP_KEY))` will be 1.

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,216 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+    model_file_path = os.path.join(bundle_dir, weights_name)
+
+    if not os.path.exists(model_file_path):
+        if name is None and url is None:
+            raise ValueError(f"To download and load model from source: {source}, `name` must be provided.")
+        if repo is None and url is None:
+            raise ValueError(f"To download and load model from source: {source}, `repo` must be provided.")
```
Nic-Ma, line 315: Could you please help combine these 2 `ValueError`?

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,216 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+    # loading with `torch.jit.load`
+    if is_ts_model is True:
+        return torch.jit.load(model_file_path, map_location=map_location)
+    # loading with `torch.load`
+    model_dict = torch.load(model_file_path, map_location=map_location)
+
+    if net_name is not None:
+        net_config = {"_target_": net_name}
+        for k, v in net_kwargs.items():
+            net_config[k] = v
```
Nic-Ma, line 326: I think here you can use `net_kwargs` as the config dict directly:
```py
net_kwargs["_target_"] = net_name
configer = ConfigComponent(config=net_kwargs)
```
What do you think?

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,216 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+        download_url(url=url_, filepath=filepath, hash_val=None, progress=progress_)
+        filepath = Path(filepath)
+        if filepath.name.endswith(("zip", "tar", "tar.gz")):
+            extractall(filepath=filepath, output_dir=bundle_dir_, has_base=True)
+    elif source_ == "github":
+        _download_from_github(
+            repo=repo_, tag_name=name_, download_path=bundle_dir_, filename=filename, progress=progress_
+        )
+    else:
+        raise NotImplementedError("So far, only support to download from url or `github` source.")
```
ericspod, line 256: "Currently only download from provided URL in (`url` argument) or Github (`source` == "github") is implemented" might be a more informative message.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,216 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+    model_file_path = os.path.join(bundle_dir, weights_name)
+
+    if not os.path.exists(model_file_path):
+        if name is None and url is None:
+            raise ValueError(f"To download and load model from source: {source}, `name` must be provided.")
+        if repo is None and url is None:
+            raise ValueError(f"To download and load model from source: {source}, `repo` must be provided.")
+        download(name=name, bundle_dir=bundle_dir, source=source, repo=repo, url=url, progress=progress)
+    # loading with `torch.jit.load`
+    if is_ts_model is True:
```
ericspod, line 318: `is_ts_model` is always a bool so wouldn't `if is_ts_model:` be correct here?

Nic-Ma: Hi @ericspod , I think we should not rely on filename to check it's torchscript or weights, I remember in your previous `TorchScript` utils, the torchscript model name can be `XXX.pt` or `XXX.pth`?

Thanks.

Nic-Ma:
> The release for the test data has the right files but the bundle specification requires things to be in specific subdirectories, the alternative is to put the files into the right directory and then zip it all together into an archive file which would contain the README.md and other files and subdirectories in its root.

Hi @ericspod ,

That's a good point, seems we can't support subdirectories directly in Github storage. `zip the files` can support subdirectories, but then users can't view files in the web page directly and can't download single file(like weights, config).
So if we really can't find other solutions, maybe let's upload the files in "flatten" way so far as @yiheng-wang-nv did in test?
What do you think?

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,216 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+
+        # Set default args of `run` in a JSON / YAML file, help to record and simplify the command line.
+        # Other args still can override the default args at runtime:
+        python -m monai.bundle download --args_file "/workspace/data/args.json" --repo "repo_owner/repo_name"
+
+    Args:
+        name: the bundle name. If `None` and `url` is `None`, it must be provided in `args_file`.
+            If `source` is `github`, it should be the same as the release tag.
+            If only a single file is expected to be downloaded, `name` should be a string that consisted with
+            the bundle name, a separator `#` and the weights name. For example: if the bundle name is "spleen" and
```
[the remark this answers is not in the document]
yiheng-wang-nv: Thanks Nic, here is a mistake, it should be filename rather than weights name.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,216 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
     return tuple(ret)
 
 
+def _get_git_release_url(repo_owner: str, repo_name: str, tag_name: str, filename: str):
+    return f"https://github.com/{repo_owner}/{repo_name}/releases/download/{tag_name}/{filename}"
+
+
+def _get_git_release_assets(repo_owner: str, repo_name: str, tag_name: str):
+    url = f"https://api.github.com/repos/{repo_owner}/{repo_name}/releases/tags/{tag_name}"
+    resp = requests_get(url)
```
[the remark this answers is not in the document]
yiheng-wang-nv: Hi @Nic-Ma , the following line: `resp.raise_for_status()` will raise the corresponding error.

monai/bundle/scripts.py
```diff
@@ -116,6 +121,216 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+        if name is None and url is None:
+            raise ValueError(f"To download and load model from source: {source}, `name` must be provided.")
+        if repo is None and url is None:
+            raise ValueError(f"To download and load model from source: {source}, `repo` must be provided.")
+        download(name=name, bundle_dir=bundle_dir, source=source, repo=repo, url=url, progress=progress)
+    # loading with `torch.jit.load`
+    if is_ts_model is True:
+        return torch.jit.load(model_file_path, map_location=map_location)
+    # loading with `torch.load`
+    model_dict = torch.load(model_file_path, map_location=map_location)
```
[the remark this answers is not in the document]
yiheng-wang-nv: Hi Nic, thanks very much of posting this question. Let me correct it.
In addition, it seems `map_location` of `monai.apps.load_from_mmar` also has this issue, let me post a bug ticket for it.

monai/bundle/scripts.py
ericspod: We do actually have bundle names stated in the specification as specifically "model.pt" and "model.ts" with the expectation that a bundle contains a single network. If we want to bundle together multiple networks we have to rework this definition because I had wanted names to be fixed so that other code can be written based on those assumptions.

Nic-Ma:
Hi @ericspod ,

@yiheng-wang-nv and I tried to revisit this "subdirectories" issue today, how about let's simply the first version API:
1. Put the bundle in a `zip` file with well folder structures, then upload it to the Github release of some repo. For the CI test bundle, we can just put it to: https://github.com/Project-MONAI/MONAI-extra-test-data/releases/tag/0.8.1.
2. The `monai.bundle download` API only supports downloading the whole bundle zip directly, similar to our existing MMAR API: https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/mmars/mmars.py#L97.
3. I imagine we will show a list of bundles and the corresponding description in https://monai.io/, so we don't need to consider the files browser of a bundle in Github release directly.
4. The `load` API still supports to automatically download a bundle and load weights or torchscript model from it.

What do you think?

Thanks in advance.

monai/bundle/scripts.py
Nic-Ma: Hi @ericspod , I feel the `model.pt` and `model.ts` names in the documentation are as examples or defaults, not force to use that?  But I agree that's a good point we should discuss it. We already received feature request from internal team for GAN models, etc.
@rijobro @wyli What do you guys think about the model names?

Thanks in advance.

ericspod: I had meant them to be specifically fixed file names but I guess that wasn't clear. We'll discuss changes then and what to do about filenames but it's helpful that there's a pattern that permits automation. The directory names are definitely fixed and that helps knowing what's what in a bundle.

ericspod:
> Hi @ericspod ,
> 
> @yiheng-wang-nv and I tried to revisit this "subdirectories" issue today, how about let's simply the first version API:
> 
>     1. Put the bundle in a `zip` file with well folder structures, then upload it to the Github release of some repo. For the CI test bundle, we can just put it to: https://github.com/Project-MONAI/MONAI-extra-test-data/releases/tag/0.8.1.
> 
>     2. The `monai.bundle download` API only supports downloading the whole bundle zip directly, similar to our existing MMAR API: https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/mmars/mmars.py#L97.
> 
>     3. I imagine we will show a list of bundles and the corresponding description in https://monai.io/, so we don't need to consider the files browser of a bundle in Github release directly.
> 
>     4. The `load` API still supports to automatically download a bundle and load weights or torchscript model from it.
> 
> 
> What do you think?
> 
> Thanks in advance.

OK sounds good let's go with this and address any issues we might come up with in subsequent revisions. This looks good for the first version.

monai/bundle/scripts.py
```diff
@@ -116,6 +120,166 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+            raise ValueError("bundle_dir=None, but no suitable default directory computed. Upgrade Pytorch to 1.6+ ?")
+    bundle_dir = Path(bundle_dir)
+
+    weights_name: str = "model.ts" if load_ts_module is True else "model.pt"
+    model_file_path = os.path.join(bundle_dir, name, weights_name)
+    if not os.path.exists(model_file_path):
+        download(name=name, bundle_dir=bundle_dir, source=source, repo=repo, progress=progress)
+
+    # loading with `torch.jit.load`
+    if load_ts_module is True:
```
Nic-Ma, line 268: For `TorchScript` model, I would suggest to leverage `load_net_with_metadata ` to return `model`, `metadata`, `configs`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/torchscript_utils.py#L100
It's consistent with our `ckpt_export` API:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/scripts.py#L418
Seems you just need to add an optional arg `config_files` mapping to the `more_extra_files` arg.

What do you think?

monai/bundle/scripts.py
```diff
@@ -249,10 +413,13 @@ def verify_metadata(
     try:
         # the rest key-values in the _args are for `validate` API
         validate(instance=metadata, schema=schema, **_args)
-    except ValidationError as e:
+    except exceptions.ValidationError as e:
         # as the error message is very long, only extract the key information
         logger.info(re.compile(r".*Failed validating", re.S).findall(str(e))[0] + f" against schema `{url}`.")
         return
+    except exceptions.SchemaError as e:
```
Nic-Ma, line 420: Does this change failed the CI tests?

Thanks.

Nic-Ma (approved):
Thanks for the quick update.
Looks good to me except for some minor suggestions.

Thanks.

monai/bundle/scripts.py
Nic-Ma: I will also enhance the `TorchScript` utils according to the discussion: https://github.com/Project-MONAI/MONAI/issues/4094#issuecomment-1105363600

Thanks.

monai/bundle/scripts.py
yiheng-wang-nv: Hi Nic, without this change, `pylint` will raise error. https://github.com/Project-MONAI/MONAI/runs/6075933692?check_suite_focus=true

Nic-Ma: I restored the code and tested with CI.

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +120,166 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+    if bundle_dir is None:
+        get_dir, has_home = optional_import("torch.hub", name="get_dir")
+        if has_home:
+            bundle_dir = Path(get_dir()) / "bundle"
+        else:
+            raise ValueError("bundle_dir=None, but no suitable default directory computed. Upgrade Pytorch to 1.6+ ?")
+    bundle_dir = Path(bundle_dir)
+
+    weights_name: str = "model.ts" if load_ts_module is True else "model.pt"
+    model_file_path = os.path.join(bundle_dir, name, weights_name)
```
Nic-Ma, line 263: I would suggest to add `model_file` arg referring to:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/mmars/mmars.py#L212
If None, the default logic should be:
```py
if model_file is None:
    model_file = os.path.join(bundle_dir, name, "models", "model.ts" if load_ts_module is True else "model.pt")
```

monai/bundle/scripts.py
```diff
@@ -116,6 +120,175 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+
+def load(
+    name: str,
+    load_ts_module: bool = False,
+    bundle_dir: Optional[PathLike] = None,
+    source: str = "github",
+    repo: Optional[str] = None,
+    progress: bool = True,
+    device: Optional[str] = None,
+    model_file=None,
```
Nic-Ma, line 233: I would move `model_file` arg to be after `name` arg, and add typehint for it.

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +120,175 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+    config_files: Sequence[str] = (),
+    net_name: Optional[str] = None,
+    **net_kwargs,
+):
+    """
+    Load model weights or TorchScript module of a bundle.
+    If loading a TorchScript module, the corresponding metadata dict, and extra files dict will be returned (please
+    check `monai.data.load_net_with_metadata` for more details).
+    If the weights file does not exist locally, it will be downloaded first.
+    The function can return weights, an instantiated network that loaded the weights, or a TorchScript module.
```
Nic-Ma, line 243: I feel this line is confusing, duplicated with the first 2 lines of the doc-string.
Could you please reorganize the description? Maybe clearly **list the return cases**: 1, 2, 3?

Thanks.

tests/test_bundle_download.py
```diff
@@ -0,0 +1,167 @@
...
+                    net_name=model_name,
+                    **net_args,
+                )
+                model_2.eval()
+                output_2 = model_2.forward(input_tensor)
+                torch.testing.assert_allclose(output_2, expected_output)
+
+    @parameterized.expand([TEST_CASE_4])
+    @skip_if_quick
+    def test_load_ts_module(self, bundle_files, bundle_name, repo, device, model_file):
```
Nic-Ma, line 144: Could you please help add unit tests for the `config_files` arg? May also verify the loaded config content.

Thanks.

Nic-Ma (approved):
Thanks for the quick update.
Looks good to me except for some minor comments inline.

Thanks.

monai/bundle/scripts.py
```diff
@@ -116,6 +120,179 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+
+        # Execute this module as a CLI entry, and download bundle:
+        python -m monai.bundle download --name "bundle_name" --source "github" --repo "repo_owner/repo_name/release_tag"
+
+        # Execute this module as a CLI entry, and download bundle via URL:
+        python -m monai.bundle download --name "bundle_name" --url <url>
+
+        # Set default args of `run` in a JSON / YAML file, help to record and simplify the command line.
+        # Other args still can override the default args at runtime:
+        python -m monai.bundle download --args_file "/workspace/data/args.json" --source "github"
```
ericspod, line 176: A simple example of what such an args file would look like in this context would help explain its usage I think.

monai/bundle/scripts.py
```diff
@@ -116,6 +120,179 @@ def _get_fake_spatial_shape(shape: Sequence[Union[str, int]], p: int = 1, n: int
...
+    model_dict = torch.load(full_path, map_location=torch.device(device))
+
+    if net_name is None:
+        return model_dict
+    net_kwargs["_target_"] = net_name
+    configer = ConfigComponent(config=net_kwargs)
+    model = configer.instantiate()
+    model.to(device)  # type: ignore
+    model.load_state_dict(model_dict)  # type: ignore
+    return model
```
ericspod, line 293: When loading the weights only we could go about loading the extra files to replicate the behaviour of providing `more_extra_files=config_files` in `load_net_with_metadata`, but we can leave that for future iteration.

ericspod (approved):

monai/bundle/scripts.py
Nic-Ma: Thanks for your suggestion. I agree we can revisit it in the future if need to get both the model weights and configs.

monai/bundle/scripts.py
yiheng-wang-nv: Thanks Eric, I enhanced it.

PR #4115, opened 2022-04-12
`ToMetaTensor` and `FromMetaTensor` transforms

rijobro:
Requires [this PR](https://github.com/Project-MONAI/MONAI/pull/4077).

### Description
Convert between dictionary data and `MetaTensor`. Will be useful for maintaining backwards compatibility. Not a breaking change, so can go into `dev`.

### Status

### Types of changes

### DIFF
--- a/monai/data/meta_tensor.py
+++ b/monai/data/meta_tensor.py
@@ -94,6 +94,7 @@ class MetaTensor(MetaObj, torch.Tensor):
         if isinstance(x, torch.Tensor) and not isinstance(x, MetaTensor):
             self.meta = deepcopy(self.meta)
         self.affine = self.affine.to(self.device)
+
     def _copy_attr(self, attribute: str, input_objs: list[MetaObj], default_fn: Callable, deep_copy: bool) -> None:
         super()._copy_attr(attribute, input_objs, default_fn, deep_copy)
         val = getattr(self, attribute)
--- a/monai/data/meta_tensor.py
+++ b/monai/data/meta_tensor.py
@@ -12,6 +12,7 @@
 from __future__ import annotations
 
 import warnings
+from copy import deepcopy
 from typing import Callable
 
 import torch
@@ -87,6 +88,11 @@ class MetaTensor(MetaObj, torch.Tensor):
             out.affine = x.affine
         else:
             out.affine = out.get_default_affine()
+        out.affine = out.affine.to(out.device)
+
+        # if we are creating a new MetaTensor, then deep copy attributes
+        if isinstance(x, torch.Tensor) and not isinstance(x, MetaTensor):
+            out.meta = deepcopy(out.meta)
         out.affine = out.affine.to(out.device)
 
         return out
--- a/monai/transforms/meta_utility/dictionary.py
+++ b/monai/transforms/meta_utility/dictionary.py
@@ -1,0 +1,102 @@
+# Copyright (c) MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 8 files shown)


wyli (commented):
thanks, I think these transforms are useful during the migration of meta_dict -> metatensor, but eventually we don't need these, so I'm fine with or without this PR in `dev`, any ideas @Nic-Ma @ericspod ?

rijobro:
In our codebase you're right that these would be temporary, but if a user has custom transforms, this would be useful.
```python
Compose([
    LoadImaged(),  # will use MetaTensor
    CustomTransformd(),  # expects dict
    OtherMonaiTransformd(),  # will use MetaTensor
])
```
would become:
```python
Compose([
    LoadImaged(),  # will use MetaTensor
    FromMetaTensord(),
    CustomTransformd(),  # expects dict
    ToMetaTensord(),
    OtherMonaiTransformd(),  # will use MetaTensor
])
```

ericspod:
These will be needed in the long run like @rijobro suggests.

wyli (approved):
Thanks, it looks good to me.

monai/data/meta_tensor.py
```diff
@@ -89,6 +90,11 @@ def __new__(cls, x, affine: torch.Tensor | None = None, meta: dict | None = None
             out.affine = out.get_default_affine()
         out.affine = out.affine.to(out.device)
 
+        # if we are creating a new MetaTensor, then deep copy attributes
+        if isinstance(x, torch.Tensor) and not isinstance(x, MetaTensor):
+            out.meta = deepcopy(out.meta)
+        out.affine = out.affine.to(out.device)
```
Nic-Ma, line 96: Hi @wyli ,

Could you please help confirm whether this part logic works well with your PR design https://github.com/Project-MONAI/MONAI/pull/4131?

Thanks in advance.

monai/transforms/meta_utility/dictionary.py
```diff
@@ -0,0 +1,102 @@
...
+    "FromMetaTensord",
+    "FromMetaTensorD",
+    "FromMetaTensorDict",
+    "ToMetaTensord",
+    "ToMetaTensorD",
+    "ToMetaTensorDict",
+]
+
+
+class FromMetaTensord(MapTransform, InvertibleTransform):
```
Nic-Ma, line 37: I am wondering as these 2 transforms are temporary, should we make it as `deprecated` since `0.9.0` or `1.0.0`?

Thanks in advance.

Nic-Ma (approved):

monai/data/meta_tensor.py
wyli: yes, this makes sense, perhaps we need another method such as `as_meta_tensor` to provide an option of avoiding heavy deepcopy...

rijobro: I don't think the deepcopy will be heavy. It's only the meta data (not the image) and would only happen when constructing a `MetaTensor` from `torch.Tensor`, which we wouldn't be doing often.

monai/transforms/meta_utility/dictionary.py
rijobro: I don't think they will be temporary, see https://github.com/Project-MONAI/MONAI/pull/4115#issuecomment-1098255474.

Nic-Ma: ok, I see, thanks.

monai/data/meta_tensor.py
wyli: not sure if we should have these two assumptions... 1. metadata is always lightweight 2. we don't often convert torch tensor to meta tensor. It's ok for the current use cases, but we should also think about future extensions.

rijobro:
@wyli , can you run blossom on this please?

wyli:
> @wyli , can you run blossom on this please?

looks good, I'll merge this one soon.

PR #4130, opened 2022-04-13
Meta tensor load image

rijobro:
This shows how I would propose to modify `LoadImage` to deal with `MetaTensor`. 

This caused a lot of errors throughout our codebase, which I solved with `FromMetaTensord` wherever necessary.

This is a temporary fix and the usages of `FromMetaTensord` will be deleted as other parts of the codebase are updated to be compatible with `MetaTensor`.

This is a breaking change, so the PR is to `MONAI/MetaTensor` instead of `dev`.

### Status
**Ready**

### Types of changes


rijobro:
@Nic-Ma @ericspod @wyli 
This branch seems to be working. Also all of the monai tutorials correctly run on this modified branch: https://github.com/Project-MONAI/tutorials/tree/MetaTensor. Could you please review this PR when you have the chance?

monai/transforms/io/array.py
```diff
@@ -241,14 +245,46 @@ def __call__(self, filename: Union[Sequence[PathLike], PathLike], reader: Option
...
+        if not get_track_meta() or meta is None:
+            return img
+
+        if "affine" in meta:
+            meta["affine"] = torch.as_tensor(meta["affine"])
+
+        # TODO: delete extra metadata
+        for i in range(8):
+            for k in ("dim", "pixdim"):
+                if f"{k}[{i}]" in meta:
```
ericspod, line 269: Somewhere in utils we could put a helper function for cleaning keys:

```python
def remove_keys(dictobj,*keys):
    for k in keys:
        dictobj.pop(k, None)
```

monai/transforms/io/array.py
```diff
@@ -241,14 +245,46 @@ def __call__(self, filename: Union[Sequence[PathLike], PathLike], reader: Option
...
+                    del meta[f"{k}[{i}]"]
+        for k in (
```
ericspod, line 271: Maybe move these keys to a global list somewhere (`TRANSFORM_KEYS`)? What we're doing is removing the transform values because they don't make sense here, this might be done elsewhere and it would be good to know what keys these are and in case add more for other formats since these are for Nifti. Then you can have:

```
remove_keys(meta, *TRANSFORM_KEYS)
```

monai/transforms/utility/dictionary.py
```diff
@@ -425,7 +425,13 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> Dict[Hashable, N
                         d[split_meta_key] = deepcopy(orig_meta)
                         dim = self.splitter.dim
                         if dim > 0:  # don't update affine if channel dim
-                            shift = np.eye(len(d[split_meta_key]["affine"]))  # type: ignore
+                            affine = d[split_meta_key]["affine"]  # type: ignore
+                            ndim = len(affine)
+                            shift = (
```
ericspod, line 430: Maybe a little clearer as statements:

```suggestion
if isinstance(affine, torch.Tensor):
    shift = torch.eye(ndim, device=affine.device, dtype=affine.dtype)
else:
    shift = np.eye(ndim)
```

monai/transforms/io/array.py
```diff
@@ -241,14 +245,46 @@ def __call__(self, filename: Union[Sequence[PathLike], PathLike], reader: Option
...
         meta_data[Key.FILENAME_OR_OBJ] = f"{ensure_tuple(filename)[0]}"  # Path obj should be strings for data loader
-
-        return img_array, meta_data
+        img = self.join_im_and_meta(img_array, meta_data)
+        if self.ensure_channel_first:
+            img = EnsureChannelFirst()(img)
+        return img
+
+    @staticmethod
+    def join_im_and_meta(im, meta: dict):
```
ericspod, line 256: If this is general purpose maybe it makes more sense in metadata utilities.

ericspod (approved):
Looks good, I have a few optional suggestions.

wyli:
Hi @rijobro, branches with the name pattern `'feature/*'` will trigger more integration tests https://github.com/Project-MONAI/MONAI/blob/641a079fe13fbacb536dfb09a0a56e3b9d8e56b5/.github/workflows/setupapp.yml#L9
after this PR, I'll rename this `MetaTensor` branch to `feature/MetaTensor` to benefit from this...

rijobro:
@wyli Any idea what's going wrong with the checks? Some jobs are failing before starting with: 

```
Reading package lists...
W: GPG error: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64  InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A4B469963BF863CC
E: The repository 'https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64  InRelease' is not signed.
Error: Process completed with exit code 100.
```

wyli:
> @wyli Any idea what's going wrong with the checks? Some jobs are failing before starting with:
> 
> ```
> Reading package lists...
> W: GPG error: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64  InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A4B469963BF863CC
> E: The repository 'https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64  InRelease' is not signed.
> Error: Process completed with exit code 100.
> ```

it's a known NGC issue and the team is working on it, (https://forums.developer.nvidia.com/t/invalid-public-key-for-cuda-apt-repository/212901)

wyli:
for the record for more tests, after this PR I'll 
- rename `metatensor` to `feature/MetaTensor`
- update the cron task to fetch the tutorial branch https://github.com/Project-MONAI/tutorials/tree/MetaTensor

monai/data/meta_obj.py
```diff
@@ -189,9 +189,7 @@ def get_default_meta(self) -> dict:
 
     def __repr__(self) -> str:
         """String representation of class."""
-        out: str = super().__repr__()
```
rijobro, line 192: I think this line should be reverted, I think the output should be the same as `torch.Tensor`, but with the extra metadata (now we have metadata _instead_ of voxel data). @wyli what do you think?

wyli: sure, please feel free to change, the corresponding test case is this line: https://github.com/Project-MONAI/MONAI/blob/18fca8811a3cc6fdbd494cf5ebe5a0d8b6392acb/tests/test_meta_tensor.py#L201

PR #4137, opened 2022-04-14
`MetaTensor`: collate; decollate; dataset; dataloader; out=; indexing and iterating across batches

rijobro:
Add collation and decollation for `MetaTensor`. Also add support for `out=` used as a kwarg, e.g., `torch.add(a,b,out=c)`.

Non-breaking, so going into `dev`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -473,11 +473,11 @@ def decollate_batch(batch, detach: bool = True, pad=True, fill_value=None):
         if batch.ndim == 0:
             return batch.item() if detach else batch
         out_list = torch.unbind(batch, dim=0)
-        # if of type MetaTensor, decollate the metadata and affines
+        # if of type MetaTensor, decollate the metadata
         if isinstance(batch, MetaTensor):
             metas = decollate_batch(batch.meta)
             for i in range(len(out_list)):
-                out_list[i].meta = metas[i]
+                out_list[i].meta = metas[i]  # type: ignore
         if out_list[0].ndim == 0 and detach:
             return [t.item() for t in out_list]
         return list(out_list)
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -28,6 +28,7 @@ import torch
 from torch.utils.data._utils.collate import default_collate
 
 from monai.config.type_definitions import NdarrayOrTensor, NdarrayTensor, PathLike
+from monai.data.meta_tensor import MetaTensor
 from monai.networks.layers.simplelayers import GaussianFilter
 from monai.utils import (
     MAX_SEED,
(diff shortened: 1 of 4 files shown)


monai/data/utils.py
```diff
@@ -346,9 +347,15 @@ def list_data_collate(batch: Sequence):
             ret = {}
             for k in elem:
                 key = k
-                ret[key] = default_collate([d[key] for d in data])
-            return ret
-        return default_collate(data)
+                data_for_batch = [d[key] for d in data]
+                ret[key] = default_collate(data_for_batch)
+                if isinstance(ret[key], MetaTensor) and all(isinstance(d, MetaTensor) for d in data_for_batch):
```
Nic-Ma, line 352: May I know what's the `meta` in the collated `ret[key]` if not all the items are `MetaTensor`?

Thanks in advance.

Nic-Ma (approved):
Hi @rijobro ,

Thanks for the quick update.
I am wondering whether it's possible to give a **warning** if user doesn't use our `list_data_collate`? Otherwise, I feel it is easily ignored and lead to unknown errors..

Thanks.

monai/data/utils.py
rijobro: I envisage it being the case where all of them or none of them will be `MetaTensor`.

rijobro:
@ericspod @wyli @Nic-Ma I upgraded the logic so that a `MetaTensor` knows if it is a batch of data or not. This is so that indexing and iterating with batches of data return a subset of the metadata, as expected:

```python
dl = DataLoader(ds)
batch = next(iter(dl))
batch[0]  # should only return the 0th image and 0th metadata
batch[:, -1]  # should return all metadata
batch[..., -1]  # should return all metadata
next(iter(batch))  # should return 0th image and 0th metadata.

```

rijobro:
@Nic-Ma could you run `/build`?

Nic-Ma:
> 

Hi @rijobro ,

Thanks for your update.
I am wondering whether it's possible to give a warning if user doesn't use our list_data_collate?

Thanks.

Nic-Ma:
> Hi @rijobro ,
> 
> Thanks for the quick update. I am wondering whether it's possible to give a **warning** if user doesn't use our `list_data_collate`? Otherwise, I feel it is easily ignored and lead to unknown errors..
> 
> Thanks.

Hi @rijobro ,

Thanks for your update.
I am wondering whether it's possible to give a warning if user doesn't use our list_data_collate?

Thanks.

rijobro:
From the following code snippet, the first is good (uses our collation fn). The middle case can have a warning. But how would we do the last case? Or are you just hoping for a warning in the middle case?

```python
import torch
from monai.data.dataset import Dataset
from monai.data.dataloader import DataLoader
from torch.utils.data import DataLoader as TorchDataLoader

data = [MetaTensor(torch.rand((1, 20, 20))) for _ in range(5)]
ds = Dataset(data)

# good
dl = DataLoader(ds)
next(iter(dl))

# can report warning
dl_no_collate = DataLoader(ds, collate_fn=lambda x: x)
next(iter(dl_no_collate))

# can't report warning
torch_dl = TorchDataLoader(ds)
next(iter(torch_dl))
```

Nic-Ma:
Hi @rijobro ,

Yeah, I was asking about the last case, I understand it's hard to give a warning there. Just worried about user experience.
If you don't have any ideas, I am OK to merge this PR first.

Thanks.

rijobro:
I don't have any better ideas at the moment, but will certainly reflect on it. The `MetaTensor` docstring contains this info:

> When creating a batch with this class, use `monai.data.DataLoader` as opposed to `torch.utils.data.DataLoader`, as this will take care of collating the metadata properly.

PR #4152, opened 2022-04-20
Redesign PatchWSIDataset

### LINKED ISSUE #3930: Refactor PatchWSIDataset
Currently, `PatchWSIDataset` lives in `monai/app` and works only with `WSIReader` for specific pathology use-cases. In our effort to blend this component in MONAI core, it can be extended to use any derivative of `ImageReader` class; however, some considerations are required to avoid unnecessary caching for backend libraries with intrinsic cache (like OpenSlide). Moreover, reading a subregion needed to be handled separately for non-WSI reader classes.

P.S. will be discussed with pathology WG to explore more pathology use-cases.

drbeh:
Fixes #3930 

### Description
This PR refactor PatchWSIDataset and uses new WSIReader as the backend.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/wsi_datasets.py
+++ b/monai/data/wsi_datasets.py
@@ -9,12 +9,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import inspect
 from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 
 from monai.data import Dataset
-from monai.data.wsi_reader import WSIReader
+from monai.data.wsi_reader import BaseWSIReader, WSIReader
 from monai.transforms import apply_transform
 from monai.utils import ensure_tuple_rep
 
@@ -31,8 +32,10 @@ class PatchWSIDataset(Dataset):
         size: the size of patch to be extracted from the whole slide image.
         level: the level at which the patches to be extracted (default to 0).
         transform: transforms to be executed on input data.
-        reader_name: the name of library to be used for loading whole slide imaging, as the backend of `monai.data.WSIReader`
-            Defaults to CuCIM.
+        reader: the module to be used for loading whole slide imaging,
+            - if `reader` is a string, it defines the backend of `monai.data.WSIReader`. Defaults to cuCIM.
+            - if `reader` is a class (inherited from `BaseWSIReader`), it is initialized and set as wsi_reader.
+            - if `reader` is an instance of a a class inherited from `BaseWSIReader`, it is set as the wsi_reader.
 
     Note:
         The input data has the following form as an example:
@@ -52,7 +55,8 @@ class PatchWSIDataset(Dataset):
         size: Optional[Union[int, Tuple[int, int]]] = None,
         level: Optional[int] = None,
         transform: Optional[Callable] = None,
-        reader_name: str = "cuCIM",
+        reader="cuCIM",
+        **kwargs,
     ):
         super().__init__(data, transform)
 
@@ -62,12 +66,23 @@ class PatchWSIDataset(Dataset):
         else:
             self.size = ensure_tuple_rep(size, 2)
 
-        # Create a default level
+        # Create a default level that override all levels if it is not None
         self.level = level
+        # Set the default WSIReader's level to 0 if level is not provided
+        if level is None:
+            level = 0
 
-        # Setup the WSI reader backend
-        self.reader_name = reader_name.lower()
-        self.image_reader = WSIReader(backend=self.reader_name)
+        # Setup the WSI reader
+        self.wsi_reader: Union[WSIReader, BaseWSIReader]
+        if isinstance(reader, str):
+            self.reader_name = reader.lower()
+            self.wsi_reader = WSIReader(backend=self.reader_name, level=level, **kwargs)
+        elif inspect.isclass(reader) and issubclass(reader, BaseWSIReader):
+            self.wsi_reader = reader(level=level, **kwargs)
+        elif isinstance(reader, BaseWSIReader):
+            self.wsi_reader = reader
+        else:
+            raise ValueError(f"unsupported reader type: {reader}.")
 
         # Initialized an empty whole slide image object dict
         self.wsi_object_dict: Dict = {}
@@ -75,7 +90,7 @@ class PatchWSIDataset(Dataset):
     def _get_wsi_object(self, sample: Dict):
         image_path = sample["image"]
         if image_path not in self.wsi_object_dict:
-            self.wsi_object_dict[image_path] = self.image_reader.read(image_path)
+            self.wsi_object_dict[image_path] = self.wsi_reader.read(image_path)
         return self.wsi_object_dict[image_path]
 
     def _get_label(self, sample: Dict):
@@ -87,7 +102,7 @@ class PatchWSIDataset(Dataset):
 
     def _get_level(self, sample: Dict):
         if self.level is None:
-            return sample.get("level")
+            return sample.get("level", 0)
         return self.level
 
     def _get_size(self, sample: Dict):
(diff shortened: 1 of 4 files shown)


monai/data/wsi_datasets.py
```diff
@@ -0,0 +1,116 @@
...
+            return sample.get("level")
+        return self.level
+
+    def _get_size(self, sample):
+        if self.size is None:
+            return ensure_tuple_rep(sample.get("size"), 2)
+        return self.size
+
+    def _get_data(self, sample):
+        if self.reader_name == "openslide":
```
Nic-Ma, line 101: I don't quite understand why only empty the objects dictionary when reader is `openslide`?

Thanks.

monai/data/wsi_datasets.py
```diff
@@ -0,0 +1,116 @@
...
+        self.level = level
+
+        # Setup the WSI reader backend
+        self.reader_name = reader_name.lower()
+        self.image_reader = WSIReader(backend=self.reader_name)
+
+        # Initialized an empty whole slide image object dict
+        self.wsi_object_dict: Dict = {}
+
+    def _get_wsi_object(self, sample):
```
Nic-Ma, line 74: Could you please help add typehints for the args of all the functions?
I got confused by the data type at first look.

Thanks.

monai/data/wsi_datasets.py
```diff
@@ -0,0 +1,116 @@
...
+        image_path = sample["image"]
```
Nic-Ma, line 75: Seems here we suppose all the `sample` must have these hard-code keys: "image", "label", "level", "size", "location". Should we consider to make it more general or configurable?

Thanks.

monai/data/wsi_datasets.py
```diff
@@ -0,0 +1,116 @@
...
+        if image_path not in self.wsi_object_dict:
+            self.wsi_object_dict[image_path] = self.image_reader.read(image_path)
+        return self.wsi_object_dict[image_path]
+
+    def _empty_wsi_object_dict(self):
```
Nic-Ma, line 80: Maybe change the function name to `_reset_wsi_obj_dict`, or no need this function, just reset the dict in the `_get_data()`?

Thanks.

Nic-Ma:
Hi @drbeh ,

BTW, about the `SmartCachePatchWSIDataset`, I think your previous suggestion was correct, we should enhance the `CacheDataset` and `SmartCacheDataset`, etc. Then we can use them as `decorators` on your own dataset object directly, to replace:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/pathology/data/datasets.py#L166
Let me think about this problem soon.

Thanks.

monai/data/wsi_datasets.py
drbeh: `OpenSlide` has a caching mechanism that we want to avoid here since we don't want to cache the tiles that related to the already extracted patch. It can create problems when extracting thousands of patches from hundreds of whole slide images.

monai/data/wsi_datasets.py
```diff
@@ -0,0 +1,116 @@
...
+
+    """
+
+    def __init__(
+        self,
+        data: List,
+        size: Optional[Union[int, Tuple[int, int]]] = None,
+        level: Optional[int] = None,
+        transform: Optional[Callable] = None,
+        reader_name: str = "cuCIM",
```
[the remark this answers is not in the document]
drbeh: Hi @Nic-Ma, we can do that but I don't see the benefit of this generalization since this works only with a very specific readers, namely `OpenSlide` and `CuCIM`.

monai/data/wsi_datasets.py
drbeh: Sure, will do.

monai/data/wsi_datasets.py
drbeh: Make sense. I'll just reset the dict.

drbeh:
> Hi @drbeh ,
> 
> BTW, about the `SmartCachePatchWSIDataset`, I think your previous suggestion was correct, we can slightly optimize the `CacheDataset` and `SmartCacheDataset`, etc. Then we can use them as `decorators` on another dataset object in future.
> 
> For now, you can use `PatchWSIDataset` as the `data` arg of `SmartCacheDataset` in your program to replace: https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/pathology/data/datasets.py#L166
> 
> We must implement `SmartCachePatchWSIDataset` before because Clara 4.0 only supports 1 dataset in the config.
> 
> Thanks.

Sure, will add SmartCache in a seperate PR as the test cases need some attention.

monai/data/wsi_datasets.py
drbeh: I agree! But this dataset is based on `WSIReader` which requires very specific inputs  "level", "size", "location", which are very standard terms in digital pathology. I think adding flexibility may cause confusion for the user her.

monai/data/wsi_datasets.py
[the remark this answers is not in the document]
Nic-Ma: Hi @drbeh ,

I think the main benefit is to support customized readers, for example, users may extend your `CuCIMReader` with some their own logic, then use it in the `PatchWSIDataset` dataset.
What do you think?

Thanks.

monai/data/wsi_datasets.py
Nic-Ma: OK, I see. Thanks for your explanation.
Could you please help add some comments or doc-string to describe it, to help others understand the code?

Thanks in advance.

monai/data/wsi_datasets.py
Nic-Ma: OK, I am OK to keep it for now, we may revisit it if really facing the feature request someday.

Thanks.

monai/data/wsi_datasets.py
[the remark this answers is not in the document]
drbeh: Sure, I can add  (1) backend name, (3) reader class, and (4) reader instance if you think that someone would use it.
However, due to the way `WSIReader` is designed as an interface for all the concrete implementation of `BaseWSIReader` (2) reader class module path will never be used.

monai/data/wsi_datasets.py
drbeh: Sure

monai/data/wsi_datasets.py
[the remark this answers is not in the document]
Nic-Ma: Sure, sounds good to me.
If we have the request to config the expected class as `string` type in JSON config, we may need to support that later.

Thanks.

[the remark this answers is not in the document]
drbeh: Sure, make sense.

drbeh:
Hi @Nic-Ma, any other comment here? thanks :)

monai/data/wsi_datasets.py
```diff
@@ -0,0 +1,134 @@
...
+    Args:
+        data: the list of input samples including image, location, and label (see the note below for more details).
+        size: the size of patch to be extracted from the whole slide image.
+        level: the level at which the patches to be extracted (default to 0).
+        transform: transforms to be executed on input data.
+        reader: the module to be used for loading whole slide imaging,
+            - if `reader` is a string, it defines the backend of `monai.data.WSIReader`. Defaults to cuCIM.
+            - if `reader` is a class (inherited from `BaseWSIReader`), it is initialized and set as wsi_reader.
+            - if `reader` is an instance of a a class inherited from `BaseWSIReader`, it is set as the wsi_reader.
+
```
Nic-Ma, line 39: Could you please help add the doc-string for arg `kwargs`?

Thanks.

Nic-Ma (approved):
Thanks for the quick update.
Looks good to me.

PR #4159, opened 2022-04-22
4094 Enhance `ckpt_export` to save config files

### LINKED ISSUE #4094: Checkpoint Export Functionality Should Require Bundle Directory Path Only

Nic-Ma:
Fixes #4094 .

### Description
This PR enhanced the `ckpt_export` script to save config files separately.
And also added unit tests to load the TorchScript model with meta, configs.

### Status
**Ready**

### Types of changes


monai/bundle/scripts.py
```diff
@@ -415,12 +419,22 @@ def ckpt_export(
     # convert to TorchScript model and save with meta data, config content
     net = convert_to_torchscript(model=net)
 
+    extra_files: Dict = {}
+    for i in ensure_tuple(config_file_):
+        # split the filename and directory
+        _, filename = os.path.split(i)
```
ericspod, line 425:
```suggestion
        filename = os.path.basename(i)
```

monai/bundle/scripts.py
```diff
@@ -415,12 +419,22 @@ def ckpt_export(
...
+        # remove extension
+        filename, _ = os.path.splitext(filename)
```
ericspod, line 427: Why not just remove this line? `filename` will contain the base name of the file which can be put directly into `extra_files`.

monai/bundle/scripts.py
```diff
@@ -415,12 +419,22 @@ def ckpt_export(
...
+        if filename in extra_files:
+            raise ValueError(f"already have config in the extra files with same file name: {filename}.")
```
ericspod, line 429:
```suggestion
            raise ValueError(f"Base filename '{filename}' is given multiple times in config file list")
```

The alternative is make sure the filenames are unique but that seems too complicated.

ericspod (changes requested):
I think we can preserve the extensions of files we're saving with the minor change I'd suggest unless I'm missing something about why this was there originally.

monai/bundle/scripts.py
Nic-Ma: Thanks for your suggestion, so you mean if the config is YAML file, we should `yaml.dumps()` to serialize it?
Now this PR always uses `json.dumps()` to serialize:
https://github.com/Project-MONAI/MONAI/pull/4159/files#diff-a715c9207f1e727d7a6ed7c6b25dfb57045d4429fc0b4eed3f5396e26ab34f80R430

Thanks.

monai/bundle/scripts.py

Thanks.

monai/bundle/scripts.py

Thanks.

monai/bundle/scripts.py
Nic-Ma: As discussed offline, let's save the configs without extension.

Thanks.

Nic-Ma:
Hi @ericspod ,

Thanks for your review, I updated the PR according to your comments.

Thanks in advance.

ericspod (approved):

PR #4163, opened 2022-04-24
4160 Enhance mean dice

### LINKED ISSUE #4160: Enhance `DiceMetric` to support counting negative samples
I think the `DiceMetric` may has an error when there is no foreground for `y` but has foreground for `y_pred`. According to the formula, the result should be 0, but so far the class will return `nan`:

```
torch.where(y_o > 0, (2.0 * intersection) / denominator, torch.tensor(float("nan"), device=y_o.device))
```

The earliest commit I can found is already use this way for calculation: see https://github.com/Project-MONAI/MONAI/pull/285/files#diff-df0f76defe29c2c91286e52334ead7a0f1f54e392d1a3e8280e2e714800dc4cbL96

I think we may need to use:
```
torch.where(denominator > 0, (2.0 * intersection) / denominator, torch.tensor(float("nan"), device=y_o.device))
```
In addition, we use a fixed value: `nan` to fill the cases that the ground truth (as I mentioned, may need to be placed by denominator) has no foreground. However, in some practical situations, people may consider to use other values such as 1 instead. For example, in 
the latest Kaggle competition: https://www.kaggle.com/competitions/uw-madison-gi-tract-image-segmentation/overview/evaluation , it says:
```
where X is the predicted set of pixels and Y is the ground truth. The Dice coefficient is defined to be 1 when both X and Y are empty.
```

Therefore, I think we may need to add an argument here thus users can specify which value to use.

Hi @ristoh @Nic-Ma @wyli @ericspod , could you please help to double check it? Thanks!

yiheng-wang-nv:

Fixes #4160 .

### Description
This PR enhances the `DiceMetric` via adding an option to handle the empty ground truth cases.

### Status
**Ready**

### Types of changes


Nic-Ma:
Would be necessary to also this option in other metrics? I am not sure about that.
Need @dongyang0122 @ericspod @wyli @holgerroth to help evaluate from the research use cases angle.

Thanks in advance.

yiheng-wang-nv:
Hi @Nic-Ma , thanks for the suggestion.
As for the `HausdorffDistanceMetric`, I think we can also enhance it. 
So far, for cases that do not have the foreground, the distance will be `nan` (if prediction also has no foreground) or `inf` (if prediction has foreground). This conforms the original definition. However, some researchers may consider to use `distance to the boundary or background`. For example, in the following paper, which is referred by [scipy.spatial.distance.directed_hausdorff](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.directed_hausdorff.html#id1)
```
A. A. Taha and A. Hanbury, “An efficient algorithm for calculating the exact Hausdorff distance.” IEEE Transactions On Pattern Analysis And Machine Intelligence, vol. 37 pp. 2153-63, 2015.
```

ericspod:
I stated in the original issue before noticing this, but with the dice loss we have [two distinct values for smoothing the numerator and denominator](https://github.com/Project-MONAI/MONAI/blob/d63f4906688bc740067807a7d4e2112f1b006706/monai/losses/dice.py#L179) which can be set to allow the user to select which way to handle empty ground truths. I think we should have the same solution here for consistency rather than a special case for when the ground truth is empty.

monai/metrics/meandice.py
```diff
@@ -104,6 +113,9 @@ def compute_meandice(y_pred: torch.Tensor, y: torch.Tensor, include_background:
             The values should be binarized.
         include_background: whether to skip Dice computation on the first channel of
             the predicted output. Defaults to True.
+        ignore_empty: whether to ignore empty ground truth cases during calculation.
+            If `True`, NaN value will be set for empty ground truth cases.
+            If `False`, 1 will be set if the predictions of empty ground truth cases are also empty.
```
wyli, line 118: Hi @yiheng-wang-nv, I think we can provide a more generic option of `empty_default`: the Dice score to use when the ground truth class is missing, default to `"nan"`.

then in the code:

```py
default_val = torch.tensor(float(empty_default), device=y_o.device)
return torch.where(denominator > 0, (2.0 * intersection) / denominator, default_val)
```

wyli: Sorry I misunderstood the where condition perhaps we don't need this default value option for now.

wyli: Sorry I misunderstood the where condition perhaps we don't need this default value option for now.

wyli (approved):
I think this change improves the current dev codebase.

PR #4171, opened 2022-04-25
Update WSIReader

drbeh:
### Description
Updates `WSIReader` to move RGB and RBGA check to the base class.

### Status
**Ready**

### Types of changes


drbeh:
I moved the checks to the base class but I kept the `read` method separate for each concrete class. Since `cuCIM` and `OpenSlide` shares lots of similarities, it's tempting to move read to the base class (as I did initially) but let's keep it separate for know since other backends might have different way of calling whole slide objects. Later on, we can easily change it as it does not expose any API to the user.

Nic-Ma (approved):

PR #4175, opened 2022-04-26
[DO NOT REVIEW] Dummy test: new temp dgx runner

pxLi:
Signed-off-by: Peixin Li <pxli@nyu.edu>

Fixes # .

### Description

### Status

### Types of changes


wyli:
the integration tests are getting `EERROR: Unexpected bus error encountered in worker. This might be caused by insufficient shared memory (shm).` https://github.com/Project-MONAI/MONAI/runs/6172474180?check_suite_focus=true  I'm tuning this arg.

pxLi:
> /integration-test the integration tests are getting `EERROR: Unexpected bus error encountered in worker. This might be caused by insufficient shared memory (shm).` https://github.com/Project-MONAI/MONAI/runs/6172474180?check_suite_focus=true I'm tuning this arg.

The error seems to asking more shm space, I think we may need to increase it to some larger value (4g or even 8g) for integrations tests. The docker daemon only provide 64MB as default, that's why I put up this PR to increase that to 2Gi.

pxLi:
Let's see if integration could pass w/ default 64m shm-size. We could continually adjust the value here

wyli (approved):

PR #4190, opened 2022-04-27
Make transforms optional

drbeh:
### Description
This PR makes the transform argument in the existing datasets optional (`CacheDataset`, `SmartDataset`, etc.). I believe the user should be able to use these datasets without providing a transform.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -200,8 +200,8 @@ class PersistentDataset(Dataset):
     def __init__(
         self,
         data: Sequence,
-        transform: Union[Sequence[Callable], Callable],
         cache_dir: Optional[Union[Path, str]],
+        transform: Optional[Union[Sequence[Callable], Callable]] = None,
         hash_func: Callable[..., bytes] = pickle_hashing,
         pickle_module: str = "pickle",
         pickle_protocol: int = DEFAULT_PROTOCOL,
@@ -374,9 +374,9 @@ class CacheNTransDataset(PersistentDataset):
     def __init__(
         self,
         data: Sequence,
-        transform: Union[Sequence[Callable], Callable],
         cache_n_trans: int,
         cache_dir: Optional[Union[Path, str]],
+        transform: Optional[Union[Sequence[Callable], Callable]] = None,
         hash_func: Callable[..., bytes] = pickle_hashing,
         pickle_module: str = "pickle",
         pickle_protocol: int = DEFAULT_PROTOCOL,
@@ -476,7 +476,7 @@ class LMDBDataset(PersistentDataset):
     def __init__(
         self,
         data: Sequence,
-        transform: Union[Sequence[Callable], Callable],
+        transform: Optional[Union[Sequence[Callable], Callable]] = None,
         cache_dir: Union[Path, str] = "cache",
         hash_func: Callable[..., bytes] = pickle_hashing,
         db_name: str = "monai_cache",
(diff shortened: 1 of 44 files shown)


Nic-Ma:
Hi @drbeh ,

Thanks for raising this discussion.
I think we didn't default `transform` to `None` in the cache datasets because if you don't have transform, then no need to cache anymore, so no need to use these datasets.
What do you think?
CC @wyli @ericspod @rijobro for more discussion if necessary.

Thanks.

drbeh:
> Hi @drbeh ,
> 
> Thanks for raising this discussion. I think we didn't default `transform` to `None` in the cache datasets because if you don't have transform, then no need to cache anymore, so no need to use these datasets. What do you think? CC @wyli @ericspod @rijobro for more discussion if necessary.
> 
> Thanks.

Hi @Nic-Ma, thanks for your reply. I see your point but what if the user want to cache the images (without any transform) to save the IO time?

Nic-Ma:
Hi @drbeh ,

Thanks for your explanation.
If you want to cache the loaded images, usually at least you need a `LoadImaged` transform  instead of `None` transform.
Could you please help provide some example or draft code to help us understand your use case?

Thanks in advance.

Nic-Ma:
Hi @drbeh ,

And I think maybe it's better to add `None` support for the transform arg but doesn't set the default value to None, because for most cases we expect transforms, refer to:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/dataset.py#L204

What do you think?

Thanks.

monai/data/dataset.py
```diff
@@ -883,8 +883,8 @@ class SmartCacheDataset(Randomizable, CacheDataset):
     def __init__(
         self,
         data: Sequence,
-        transform: Union[Sequence[Callable], Callable],
         replace_rate: float,
+        transform: Optional[Union[Sequence[Callable], Callable]] = None,
```
wyli, line 887: this changes the argument list and potentially break external codes, perhaps we can give `replace_rate` a default value and remain the previous arg list? cc @Nic-Ma

drbeh: I added set the default value for `replace_rate` as 0.5 for now. @Nic-Ma let me know if any other value is more appropriate.
We can also remove the default value for transform but I think it would be nice to have a default value for `replace_rate` from new user's perspective so that they can just start using it without thinking about it.

Nic-Ma: Hi @drbeh ,

For regular training cases, according to my experience, I never used a `replace_rate` greater than `0.2`. `0.5` may cause the training hard to converge?
Maybe default to `0.1` to make the training more stable, just like the random `prob` in all the transforms:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/intensity/array.py#L95
I am not sure, what's your `replace_rate` in the pathology task?
@dongyang0122 Do you have any best practice for the `replace_rate` in SmartCache?

Thanks in advance.

dongyang0122: @Nic-Ma @drbeh We used to use replace_rate 0.2 and it worked fine. In general, a small replace_rate will give you better model convergence since the model sees similar data at each training epoch. A larger replace_rate will lead to larger variance in the training process, and the convergence will be slightly worse.

drbeh: @dongyang0122 @Nic-Ma thanks for your input. I agree in the pathology tumor detection use case, we also used small values, so I set the default value to `0.1`.

drbeh:
@wyli @Nic-Ma any other comments here?

wyli (approved):

PR #4194, opened 2022-04-28
Add InstanceNorm3dNVFuser support

### LINKED ISSUE #4085: Add `InstanceNorm3dNVFuser` based on NVIDIA Apex
`Fast InstanceNorm` was added via NVFuser in APEX and can be used as a drop-in replacement for `nn.InstanceNorm3d`.
Would be great to have a way to pass additional parameters to the norm layer, since it might be useful when fusions are added.

Will implement InstanceNorm3dNVFuser, starting from DynUNet.

yiheng-wang-nv:

Fixes #4085 .

### Description
This PR enhances the layer factory via adding the `InstanceNorm3dNVFuser`, a faster version of `nn.InstanceNorm3d` implemented by `apex`.

The consistency between `InstanceNorm3dNVFuser` and `nn.InstanceNorm3d` has been validated via using `DynUNet`.

### Status
**Ready**

### Types of changes


monai/networks/layers/factories.py
```diff
@@ -242,6 +245,25 @@ def sync_batch_factory(_dim) -> Type[nn.SyncBatchNorm]:
     return nn.SyncBatchNorm
 
 
+@Norm.factory_function("instance_nvfuser")
+def instance_nvfuser_factory(dim):
```
Nic-Ma, line 249: I feel this function name is not very clear? and please add the GitHub link of `apex` to the doc-string. Because we can't install `apex` by `pip install` directly as an optional package, need to let users know how to install it if not using NVIDIA docker image.

Thanks.

monai/networks/layers/factories.py
```diff
@@ -242,6 +245,25 @@ def sync_batch_factory(_dim) -> Type[nn.SyncBatchNorm]:
...
+    """
+    `InstanceNorm3dNVFuser` is a faster verison of InstanceNorm layer and implemented in `apex`.
+    It only supports 3d tensors as the input. It also requires to use with CUDA and non-Windows OS.
+    In this function, if the required library `apex.normalization.InstanceNorm3dNVFuser` does not exist,
+    `nn.InstanceNorm3d` will be returned instead.
+
+    """
+    types = (nn.InstanceNorm1d, nn.InstanceNorm2d, nn.InstanceNorm3d)
+    if dim != 3:
```
Nic-Ma, line 258: I think this logic is not correct, if I pass `dim=2`, why use `dim-1` instead?

Thanks.

monai/networks/layers/factories.py
```diff
@@ -242,6 +245,25 @@ def sync_batch_factory(_dim) -> Type[nn.SyncBatchNorm]:
...
+        warnings.warn(f"`InstanceNorm3dNVFuser` only supports 3d cases, use {types[dim - 1]} instead.")
+        return types[dim - 1]
+    if not has_nvfuser:
+        warnings.warn("`apex.normalization.InstanceNorm3dNVFuser` is not found, use nn.InstanceNorm3d instead.")
```
Nic-Ma, line 262: @wyli Should we fallback to `nn.InstanceNorm3d` or raise error directly?
I am not sure about the researcher's expected behavior here.

Thanks.

tests/test_dynunet.py
```diff
@@ -118,6 +121,32 @@ def test_script(self):
...
+                for affine in [True, False]:
+                    norm_param = {"eps": eps, "momentum": momentum, "affine": affine}
+                    input_param["norm_name"] = ("instance", norm_param)
+                    input_param_fuser = input_param.copy()
+                    input_param_fuser["norm_name"] = ("instance_nvfuser", norm_param)
+                    net = DynUNet(**input_param).to("cuda")
+                    net_fuser = DynUNet(**input_param_fuser).to("cuda")
+                    net_fuser.load_state_dict(net.state_dict())
+
+                    input_tensor = torch.randn(input_shape).to("cuda")
```
Nic-Ma, line 141: Should we use `channel_last` shape for the input data and net?
And I think maybe we should use `cuda:0` instead of `cuda` as @wyli mentioned before.

Thanks.

monai/networks/layers/factories.py
yiheng-wang-nv: Thanks Nic!

monai/networks/layers/factories.py
yiheng-wang-nv: Hi Nic, if pass `dim=2`, will use types[2-1] which corresponds to `nn.InstanceNorm2d`.

tests/test_dynunet.py
yiheng-wang-nv: Hi @Nic-Ma and @wyli , for the `cuda:0` thing, it's my first time to see this requirement, but I found that it's common to use:
```
"cuda" if torch.cuda.is_available() else "cpu"
```
in MONAI code base. Could you help to explain more about it?  Thanks!

wyli: Yes the tests are running in a multigpu environment and the current device might not be cuda:0, explicitly using cuda:0 can make sure it's using the first visible gpu.

monai/networks/layers/factories.py
Nic-Ma: I am not sure actually..`apex` seems should not be in the function name..
Hi @wyli , do you have any suggestion? Maybe just use the original `instance_nvfuser_factory`?

Thanks.

monai/networks/layers/factories.py
wyli: I think missing the apex module is not a critical issue, so a warning is good.

monai/networks/layers/factories.py
wyli: thanks, I think `instance` and `instance_nvfuser` make sense here.

wyli (approved):
Thanks, it looks good to me.

Nic-Ma (approved):
Hi @yiheng-wang-nv ,

Thanks for the quick update, looks good to me now.
Could you please help add some unit tests to ensure the `InstanceNorm3dNVFuser` is compatible with `TorchScript` exporting?

Thanks in advance.

monai/networks/layers/factories.py
mmarcinkiewicz: I'm pretty sure you've guys talked it through, but my two cents:

I would imagine that `instance_nvfuser_factory` would be merged with `instance_factory`, and the logic would be the following:
`if has_nvfuser and dim == 3: return InstanceNorm3dNVFuser`
`else: return types[dim - 1]`.

This way a user will simply request `InstanceNorm` and it transparently will try to load the NVFuser implementation, and will fallback to native if unsuccessful. What do you think?

yiheng-wang-nv:
> Hi @yiheng-wang-nv ,
> 
> Thanks for the quick update, looks good to me now. Could you please help add some unit tests to ensure the `InstanceNorm3dNVFuser` is compatible with `TorchScript` exporting?
> 
> Thanks in advance.

Hi @Nic-Ma and @wyli , unfortunately, this layer uses the customized autograd function, which is not supported in TorchScript.
The error is like:
```
Python builtin <built-in method apply of FunctionMeta object at 0x558ca33eca90> is currently not supported in Torchscript
```
 (please click this link for more details: https://github.com/pytorch/pytorch/issues/22329)
I updated the doc-strings about it, what do you think? Thanks!

Nic-Ma (approved):

monai/networks/layers/factories.py
yiheng-wang-nv: Hi @mmarcinkiewicz , thanks for the suggestion. So far, `InstanceNorm3dNVFuser` does not support TorchScript because of using a customized autograd function (Please check my comments below for more details). In addition, it also requires to use cuda in a non-Windows env. I'm not sure if combing it into `instance_factory` with these checks is a good way. Hi @Nic-Ma @wyli , what do you think? Thanks!

PR #4205, opened 2022-04-29
Implementation of boundary-aware ShapeLoss

arnauddhaene:
Fixes # .

### Description
An implementation of a slight variation of the boundary-aware `ShapeLoss` described in [Huang et al., 2021](https://ieeexplore.ieee.org/document/9433775).

### Status
**Work in progress**: the feature has been implemented, but unit testing needs to be added.

### Types of changes


monai/losses/shape.py
```diff
@@ -0,0 +1,172 @@
...
+import numpy as np
+import torch
+from scipy import ndimage
+from torch.nn.modules.loss import _Loss
+
+from monai.networks import one_hot
+from monai.utils import LossReduction
+
+
+class ShapeLoss(_Loss):
```
wyli, line 25: this class name should be more specific? perhaps `ShapeDistLoss`?

monai/losses/shape.py
```diff
@@ -0,0 +1,172 @@
...
+        """Creates a distance map from a 2D binary mask.
+
+        Args:
+            mask (torch.Tensor): Binary mask of shape [WD].
+
+        Returns:
+            torch.Tensor: Boundary distance map of shape [WD]
+        """
+        # Convert to NumPy to use with SciPy
+        roi: np.ndarray = mask.numpy()
```
wyli, line 87: please use `monai.utils.convert_data_type`: `roi = convert_data_type(mask, np.ndarray)`, it can handle the device properly...

monai/losses/shape.py
```diff
@@ -0,0 +1,172 @@
...
+            >>> from monai.losses.dice import *  # NOQA
+            >>> import torch
+            >>> from monai.losses.dice import ShapeLoss
+            >>> B, C, H, W = 7, 5, 3, 2
+            >>> input = torch.rand(B, C, H, W)
+            >>> target_idx = torch.randint(low=0, high=C - 1, size=(B, H, W)).long()
+            >>> target = one_hot(target_idx[:, None, ...], num_classes=C)
+            >>> self = ShapeLoss(reduction='none')
+            >>> loss = self(input, target)
+            >>> assert np.broadcast_shapes(loss.shape, input.shape) == input.shape
```
wyli, line 121: would be great to create a test case for this class, you can follow this one https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_local_normalized_cross_correlation_loss.py

you can run a single test case within the codebase:
```bash
pip install -U -r requirements-dev.txt  # install testing tools
python -m tests.test_local_normalized_cross_correlation_loss
```

monai/losses/shape.py
```diff
@@ -0,0 +1,172 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import warnings
+from itertools import product
+from typing import Callable, Optional, Union
+
+import numpy as np
+import torch
+from scipy import ndimage
```
wyli, line 18: please use `distance_transform_edt, _ = optional_import("scipy.ndimage.morphology", name="distance_transform_edt")` (example https://github.com/Project-MONAI/MONAI/blob/5c15138f484ba027b90fd12f7038a519f740d213/monai/metrics/utils.py#L22) because the project doesn't assume the user environment has scipy installed.

wyli (commented):
thanks for the PR, I put some initial comments inline, please ignore them if you are already working on these aspects...

Bala93:
Hi @arnauddhaene, thanks for this PR. It will be very useful for the community, as distance transform losses do generally provide improve HD and Dice. I have a doubt though, I see in code there is distance tranform being calculated in numpy. If that is the case, how will the backpropagation work in the training?

arnauddhaene:
Hi @Bala93, thank you!

We do not need to backprop on the distance transform. We are in essence transforming the ground truth to a non-binary mask that will penalize the network for boundary misclassification.

The result of the transform is compared to the logits produced by the network, and the difference is used as loss to update the network's weights via backprop.

In fact, the `distance_maps` for each example could be computed once before training as auxiliary ground truth mask to prevent duplicate computations and speed up training. However, doing this doesn't fit as nicely in the MONAI framework and also is not robust to train-time augmentations that can modify the ground truth mask (elastic transform, rotation, etc.).

I hope this answers your question.

KumoLiu:
Hi @arnauddhaene, is there any more progress on this PR so far, do you plan to finish this one?
Thanks!

dimka11:
Why no use distance_transform_edt from monai.transforms that use cupy/cucim with GPU instead slow CPU scipy?

TheHuntsman4:
I noticed that this PR has been inactive for a while. I’ve implemented a version of the loss function that more closely follows the one described in the research paper. Would it be fine to make a new PR for this?

ericspod:
Hi @arnauddhaene this PR looks like it would be good to integrate into MONAI still, if you would like to resume work that would be great, otherwise @TheHuntsman4 can propose his version instead. Thanks both for your contributions!

PR #4216, opened 2022-05-02
Bug fix and improvement in WSI

### LINKED ISSUE #4215: SmartCache for `PatchWSIDataset`
Add SmartCache functionality for the new `PatchWSIDataset`. https://github.com/Project-MONAI/MONAI/pull/4152#issuecomment-1104873486

drbeh:
### Description
Fixes a bug and makes some updates in `WSIReader` and `PatchWSIDataset`

### Status
**Ready**

### Types of changes


drbeh:
@Nic-Ma, I removed `SmartCacheWSIDataset` from this PR and kept only the minor improvement and bug fix. Can you please review it?

Nic-Ma:
Hi @drbeh ,

I think we don't have missing action items for moving the existing `WSI` components to core, could you please help also unify `test_wsireader_new.py` and `test_wsireader.py` in this PR?
Please feel free to correct me if I misunderstood anything.

Thanks in advance.

monai/data/wsi_reader.py
```diff
@@ -358,6 +367,7 @@ def get_metadata(self, patch: np.ndarray, location: Tuple[int, int], size: Tuple
         """
         metadata: Dict = {
             "backend": "cucim",
+            "wsi_path": wsi.path,
```
Nic-Ma, line 370: May I know why add the `wsi_path` item in the cuCIM metadata?
1. Maybe we should also add it in the `openSlide` reader?
2. Should we define the `metadata` keys in the base class to ensure different readers output the same format?

Thanks.

drbeh:
> Hi @drbeh ,
> 
> I think we don't have missing action items for moving the existing `WSI` components to core, could you please help also unify `test_wsireader_new.py` and `test_wsireader.py` in this PR? Please feel free to correct me if I misunderstood anything.
> 
> Thanks in advance.

Hi @Nic-Ma, I don't think that we should combine them in a single file since they are two different components with similar names. What I am planing to do is following:
1. First, update the pipelines and tutorials with the new components and make sure that they are working as expected.
2. Next, deprecate the old components.
3. Finally, rename `test_xxx.py` -> `test_xxx_depricated.py` and `test_xxx_new.py` -> `test_xxx.py`.

Please let me know if you think differently.

monai/data/wsi_reader.py
drbeh: I agree with both. For 1, I have already add it, and will do 2.

drbeh: The reason for adding WSI path is to know which WSI the patches is extracted from so that we can combine the patches in the related WSI.

Nic-Ma:
Hi @drbeh ,

I feel we may not need to keep the deprecated APIs, delete the APIs and tests directly?
@wyli What do you think?

Thanks.

monai/data/wsi_reader.py
drbeh: All done!

drbeh:
> Hi @drbeh ,
> 
> I feel we may not need to keep the deprecated APIs, delete the APIs and tests directly? @wyli What do you think?
> 
> Thanks.

Deleting a component directly is not a good practice. The way that is usually done is to deprecate the component first to notify users (via the deprecation message) and give them time to update their code. Then after a period of time delete them all.

Nic-Ma:
Hi @drbeh ,

I understand the deprecate process, we already have several deprecated modules, for example:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/png_writer.py#L22
`deprecate` may cause the code hard to read, so sometimes we just make breaking changes directly, depends how hard users switch to the new APIs.
Anyway, it's not a blocker issue, if @wyli @ericspod you guys also like the deprecate way, I am OK with it.

Thanks in advance.

Nic-Ma (approved):
Looks good to me.

Thanks.

wyli:
the PR description says it fixes a bug, was wondering if the relevant test cases should be updated to cover the change?

PR #4220, opened 2022-05-03
Add a warning for input file shape mismatches

MathijsdeBoer:
### Description
When images loaded by `LoadImaged` do not have the same shape, there is no way of knowing untill this causes issues down the line, which may not happen untill the end of training.
With the new default settings of the `__init__` function, the warning is shown once for each input path when not all loaded images in a dict are of equal size.
This can also be disabled if the user is aware of such a possibility, by setting the argument `log_warning` to `False`.

#### Rationale
I once had to debug a shape mismatch between my `image` and `label` inputs where one of my ~60 patients had a faulty label image.
This didn't cause issues untill the `RandCropByPosNegLabeld` transform that I was using threw a shape mismatch error.
That on it's own is not a bad hint to the user that something's wrong with the input data, but I initially suspected some of my other augmentation transforms.

Furthermore, this error would not always happen at the same time (due to the shuffling of the training set), not at the same stage (train/val split) or sometimes not until after training when the faulty patient ended up in the test-set.
That would be another clue to the fact that one of my entries is wrong, but it would have saved me a few days of debugging if the `LoadImaged` transform would have just told me about this potential issue from the word go instead.

### Status
**Work in Progress**

### Types of changes

Not a breaking change, no new tests available for a `print` statement, `make html` did not work on my system. I might be missing some bit of software for that.

### DIFF
--- a/monai/transforms/io/dictionary.py
+++ b/monai/transforms/io/dictionary.py
@@ -127,22 +127,44 @@ class LoadImaged(MapTransform):
 
         """
         d = dict(data)
+        image_shapes = []
+        # From python 3.6 and up, dictionaries are ordered, so we can assume that the dict orders we get
+        # are all in the same order, unless the user goes out of their way to shuffle them.
+        first_path = list(d.values())[0]
+
         for key, meta_key, meta_key_postfix in self.key_iterator(d, self.meta_keys, self.meta_key_postfix):
-            data = self._loader(d[key], reader)
+            image_data = self._loader(d[key], reader)
+            # _loader returns a tuple of (ndarray, dict), we want the ndarray.shape
+            image_shapes.append(image_data[0].shape)
+
             if self._loader.image_only:
-                if not isinstance(data, np.ndarray):
+                if not isinstance(image_data, np.ndarray):
                     raise ValueError("loader must return a numpy array (because image_only=True was used).")
-                d[key] = data
+                d[key] = image_data
             else:
-                if not isinstance(data, (tuple, list)):
+                if not isinstance(image_data, (tuple, list)):
                     raise ValueError("loader must return a tuple or list (because image_only=False was used).")
-                d[key] = data[0]
-                if not isinstance(data[1], dict):
+                d[key] = image_data[0]
+                if not isinstance(image_data[1], dict):
                     raise ValueError("metadata must be a dict.")
                 meta_key = meta_key or f"{key}_{meta_key_postfix}"
                 if meta_key in d and not self.overwriting:
                     raise KeyError(f"Meta data with key {meta_key} already exists and overwriting=False.")
-                d[meta_key] = data[1]
+                d[meta_key] = image_data[1]
+
+        # Only warn the first time per sample to prevent spamming the logs
+        if self.log_warning and (first_path not in self.has_warned_about):
+            # all() is a lazy function, i.e. it will stop at the first instance of shapes mismatching
+            if not all(x == image_shapes[0] for x in image_shapes):
+                print("WARNING: Loaded image shapes do not match for input dictionary:")
+                print(dict(data))
+                print("\nResulting in image shapes:")
+                [print(f"\t{x}: {y.shape}") for x, y in zip(d.keys(), d.values())]
+                print("\nIf you do not wish to see this warning, pass `log_warning = False` to `LoadImaged()`")
+
+                # We save the path of the first key for later comparisons.
+                self.has_warned_about.append(first_path)
+
         return d
 
 


MathijsdeBoer:
I tried to perform the commands exactly as given by your DCO thing, but now I seem to have accidentally signed off on too many commits.

wyli:
I think you can locally revert to an earlier state with `git reflog`  good luck...

MathijsdeBoer:
There we go, I guess this is a good a crash course for Git than any.

I ended up following [this gist](https://gist.github.com/YakDriver/95f7d165ff5030c000a813d974a170dc), and that worked out.

monai/transforms/io/dictionary.py
```diff
@@ -127,22 +132,44 @@ def __call__(self, data, reader: Optional[ImageReader] = None):
...
                 if meta_key in d and not self.overwriting:
                     raise KeyError(f"Meta data with key {meta_key} already exists and overwriting=False.")
-                d[meta_key] = data[1]
+                d[meta_key] = image_data[1]
+
+        # Only warn the first time per sample to prevent spamming the logs
+        if self.log_warning and (first_path not in self.has_warned_about):
+            # all() is a lazy function, i.e. it will stop at the first instance of shapes mismatching
+            if not all(x == image_shapes[0] for x in image_shapes):
+                print("WARNING: Loaded image shapes do not match for input dictionary:")
```
ericspod, line 164: Instead of printing we should be using the [warnings module ](https://docs.python.org/3/library/warnings.html) and/or the [logging module](https://docs.python.org/3/library/logging.html). This allows us to control when and how things are printed or logged. Instead of multiple print statements you can construct a multi-line warning string then warn/log it.

monai/transforms/io/dictionary.py
```diff
@@ -127,22 +132,44 @@ def __call__(self, data, reader: Optional[ImageReader] = None):
...
+                print(dict(data))
+                print("\nResulting in image shapes:")
+                [print(f"\t{x}: {y.shape}") for x, y in zip(d.keys(), d.values())]
```
ericspod, line 167: I don't recommend using `for` expressions as shorthand for `for` loops, or any other expression in place of a statement. This adheres to Python style guides like [this](https://docs.python-guide.org/writing/style/) which I think is generally for us to follow.

rijobro:
I can certainly see why this would be useful in your use case. But I think the name should be something more specific than `log_warning`; it's a specific kind of warning, so maybe more like `check_image_sizes_match`. 

From the code (and apologies if I'm wrong on this), I think you check that all dimensions match each other, which would raise warnings if the number of channels between the image and label do not match.

I'm not sure if the default should be `True`. It's true that if it's set to `False` then users may not realise they have a problem. But if we set it to `True`, users with images and labels that are of different sizes intentionally will start seeing warnings.

Nic-Ma:
Hi @rijobro @ericspod ,

I agree with your comment, maybe we can call the arg `align_spatial_shape`?
And actually, I feel maybe it's more suitable to put this check in `EnsureChannelFirstd` transform instead of `LoadImaged` because images may have unknown dims in `LoadImaged` transform.
What do you think?

Thanks.

rijobro:
Honestly, I'm not entirely sure it belongs in any of our transforms. Perhaps we could have a checklist suggesting debugging steps for users, and this could be one of the points?

rijobro:
Yes I certainly see your point that if we put it in a checklist, less people benefit from it than if we place it in our code. You're right, maybe we just need to put in the warning `...if you're expecting this, disable this warning with sanity_check=False`.

We probably also want to check for `any(np.isnan(x))`, as NaN's can often create a problem for users and the cause isn't always immediately obvious.

monai/transforms/io/dictionary.py
MathijsdeBoer: Would replacing this with
`'\n'.join([f"{key}: {value.shape}" for key, value in zip(d.keys(), d.values())])`
be more acceptable?

Or would you rather use a full `for` loop and build the string before printing?

ericspod: What you can do is build up a list of strings, say `msg`,  so here you would use a for loop at appends strings to `msg`, then the final line would be `msg = "\n".join(msg)`. You could then log or warn using `msg` as the full string.

MathijsdeBoer: In moving to the `warnings` module, I have cleaned up the string building.
```python
warnings.warn(
    message="Loaded image shapes do not match for input dictionary:" +
                    f"\n{dict(data)}" +
                    "\nResulting in image shapes:" +
                    '\n'.join(f"\t{key}: {value.shape}" for key, value in d.items()) +
                    "If you do not wish to see this warning, pass `warn_on_shape_mismatch = False` to " +
                    "LoadImaged()",
    category=UserWarning
)
```

Is that more in line with your expectations?

ericspod: It's closer to what I would have done, we're now getting into personal style but it's much better. I would have done this which is maybe too busy looking:

```python
msg = [
  "Loaded image shapes do not match for input dictionary:",
  f"{dict(data)}",
  "Resulting in image shapes:" 
]
msg += [f"\t{key}: {value.shape}" for key, value in d.items() if hasattr(value, "shape")]
msg.append("If you do not wish to see this warning, pass `warn_on_shape_mismatch = False` to LoadImaged()")

warnings.warn(message="\n".join(msg), category=UserWarning)
```

monai/transforms/io/dictionary.py
```diff
@@ -127,22 +133,48 @@ def __call__(self, data, reader: Optional[ImageReader] = None):
...
         """
         d = dict(data)
+        image_shapes = []
+        # From python 3.6 and up, dictionaries are ordered, so we can assume that the dict orders we get
+        # are all in the same order, unless the user goes out of their way to shuffle them.
+        first_path = list(d.values())[0]
+
         for key, meta_key, meta_key_postfix in self.key_iterator(d, self.meta_keys, self.meta_key_postfix):
-            data = self._loader(d[key], reader)
+            image_data = self._loader(d[key], reader)
```
rijobro, line 142: Could you revert the naming of `image_data` back to `data`? Would simplify the `diff`.

MathijsdeBoer: Done, I wanted to keep the input `data` intact in case I needed to log it in the warning, I've just gone and stored a copy in `input_dict` instead.

monai/transforms/io/dictionary.py
```diff
@@ -127,8 +133,18 @@ def __call__(self, data, reader: Optional[ImageReader] = None):
...
+        # From python 3.6 and up, dictionaries are ordered, so we can assume that the dict orders we get
+        # are all in the same order, unless the user goes out of their way to shuffle them.
+        first_path = list(d.values())[0]
+        # Keep a copy of the original input dict around in case we need to log it
+        input_dict = dict(data)
+
         for key, meta_key, meta_key_postfix in self.key_iterator(d, self.meta_keys, self.meta_key_postfix):
             data = self._loader(d[key], reader)
+            # _loader returns a tuple of (ndarray, dict), we want the ndarray.shape
+            image_shapes.append(data[0].shape)
```
rijobro, line 146: if `image_only` is `True`, then data will be just the ndarray, not a tuple. So you probably want something like:
```suggestion
            image_shapes.append(data[0].shape if isinstance(data, tuple) else data.shape)
```

MathijsdeBoer:
Well, I broke the DCO again, but I'd really rather not go back to this mess of accidental branches:
image

monai/transforms/io/dictionary.py
```diff
@@ -116,6 +120,8 @@ def __init__(
             raise ValueError("meta_keys should have the same length as keys.")
         self.meta_key_postfix = ensure_tuple_rep(meta_key_postfix, len(self.keys))
         self.overwriting = overwriting
+        self.warn_on_shape_mismatch = warn_on_shape_mismatch
+        self.has_warned_about = []
```
wyli, line 124:
```suggestion
        self.has_warned_about: list = []
```

wyli (dismissed):
Thanks, it looks good to me.

monai/transforms/io/dictionary.py
```diff
@@ -143,6 +159,24 @@ def __call__(self, data, reader: Optional[ImageReader] = None):
...
+
+        # Only warn the first time per sample to prevent spamming the logs
+        if self.warn_on_shape_mismatch and (first_path not in self.has_warned_about):
+            # all() is a lazy function, i.e. it will stop at the first instance of shapes mismatching
+            if not all(x == image_shapes[0] for x in image_shapes):
+                warnings.warn(
+                    message="Loaded image shapes do not match for input dictionary:"
+                    + f"\n{input_dict}"
+                    + "\nResulting in image shapes:"
+                    + "\n".join(f"\t{key}: {value.shape}" for key, value in d.items())
```
wyli, line 171:
```suggestion
                    + "\n".join(f"\t{key}: {value.shape}" for key, value in d.items() if hasattr(value, "shape"))
```
this needs additional checks for the attribute, something like:
```suggestion
                    + "\n".join(f"\t{key}: {value.shape}" for key, value in d.items() if hasattr(value, "shape"))
```

MathijsdeBoer:
I want to add some code that checks shapes when the amount of dimensions (image with channel, label without channel) do not match, before I feel comfortable finishing up my pull request.

monai/transforms/io/dictionary.py
MathijsdeBoer: I've opted for a slightly different solution that will print `"Not an Image"` whenever there is no `shape` attribute. On reflection, maybe `"No shape available"` might've been more appropriate/general.

wyli:
> I want to add some code that checks shapes when the amount of dimensions (image with channel, label without channel) do not match, before I feel comfortable finishing up my pull request.

sorry I thought this PR is ready... I'll remove the approval for now and wait for your further updates.

Issue #3805, opened 2022-02-14
Non-breaking refactoring of the array transforms to dispatch by input data type (and by additional info.)


wyli:
Multiple feature requests have described common use cases of metadata handling and computational backends for pre-/post-processing transformations.

### The proposed solution
We agreed that these features are essential MONAI core capabilities, and as a first step, they should be implemented at the primary level of the array-based ‘vanilla’ transforms.

The current array-level transforms were initially (in MONAI v0.1) designed to facilitate NumPy ndarray processing using the NumPy APIs. These were further extended to support both torch tensor and numpy ndarray inputs and outputs (in MONAI v0.7). We have agreed to further this extension with clearer APIs at a finer granularity.

### Goals and benefits
The goal is to decouple data handling logic depending on input data type, so that dictionary-level transform wrappers can leverage the APIs to simultaneously process images, coordinates, metadata, transform randomisable info, transform traceable and invertible info. And this potentially enables optimised transform chains based on the backend dispatched.

From the developer's perspective, extension/subclasses of the array-level transforms can choose to implement one or multiple backends, or simply overwrite the `__call__` to bypassing the dispatch logic. 

### Details
A potential implementation in pseudo-code would be:

```py
def dispatch_by_type(xform, data, **kwargs):
    xform_func = xform.dispatch_type_dict[str(type(data))]
    return xform_func(data, **kwargs)


class Transform:
    dispatch_type_dict={
        "np.ndarray": self.call_impl, 
        "torch.Tensor": self.call_torch,
        "dict": self.call_meta_dict,
    }
    
    def __init__(self):
        ...

    def call_impl(self, img_np, **kwargs):
        ...

    def call_torch(self, img_torch, **kwargs):
        ...

    def call_meta_dict(self, meta, **kwargs):
        ...

    def __call__(self, data, **kwargs):
        return dispatch_by_type(self, data, **kwargs)
```

PR #4222, opened 2022-05-04
Meta tensor channel

rijobro:
### Description
Channel-based transforms to use `MetaTensor`.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @rijobro ,

Thanks for your quick update for MetaTensor support.
I feel maybe we don't need to remove the `NdarrayOrTensor` support for the current stage? It doesn't block the development of `MetaTensor` in transforms so far. And we already completed `NdarrayOrTensor` support for all transforms.
If really necessary to deprecate `NdarrayOrTensor`, we can remove it later?

Thanks.

rijobro:
It's true that `MetaTensor` is a subclass of `torch.Tensor`, which is a sub-part of `NdarrayOrTensor`, so we could leave it. But the proposed changes mean that the input to our transforms will become `torch.Tensor`, and so I don't think it's bad that these PRs start implementing that. It enables us to remove some `# type: ignore`, and should improve readability. 

I don't see the advantage of leaving them as `NdarrayOrTensor`, let me know your opinions on that one. Thanks

monai/data/image_reader.py
```diff
@@ -833,8 +836,12 @@ def get_data(
         metadata["spatial_shape"] = np.asarray(region.shape[:-1])
         metadata["original_channel_dim"] = -1
 
+        # combine image and metadata
+        if not isinstance(region, MetaTensor):
+            region = MetaTensor(region, meta=deepcopy(metadata))
```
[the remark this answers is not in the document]
rijobro: Thanks. Deepcopy removed in https://github.com/Project-MONAI/MONAI/pull/4222/commits/4383f24c2cf1d36d3df4f17db0b4423e257ebb58.

wyli:
I think I agree with @Nic-Ma, perhaps removing `NdarrayOrTensor` (and removing the numpy support) could be done in a separate dedicated PR? our previous discussion here https://github.com/Project-MONAI/MONAI/issues/3805 should be revisited...)

rijobro:
Ok, how about this:
1. If a transform needs metadata (e.g., `EnsureChannelFirstd`), I will switch this to `torch.Tensor`, since it now can't be used with a `np.ndarray` as the metadata is no longer present,
2. If no metadata is required (e.g.,`AddChanneld`), I will leave this as `NdarrayOrTensor`.

Does that sounds reasonable?

wyli:
> 

sounds good to me, what do you think @Nic-Ma ?

Nic-Ma:
Hi @rijobro ,

Plan sounds good to me!
Thanks for the discussion.

monai/data/image_reader.py
```diff
@@ -833,8 +834,13 @@ def get_data(
         metadata["spatial_shape"] = np.asarray(region.shape[:-1])
         metadata["original_channel_dim"] = -1
 
+        # combine image and metadata
+        if not isinstance(region, MetaTensor):
```
Nic-Ma, line 838: I think this wsi_reader is deprecated, maybe no need to update it?
We should update this part instead:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/wsi_reader.py#L234
@drbeh Could you please help double confirm it?

Thanks in advance.

monai/data/image_reader.py
```diff
@@ -833,8 +834,13 @@ def get_data(
...
+            region = MetaTensor(region, meta=metadata)
+
         # Make it channel first
-        region = EnsureChannelFirst()(region, metadata)
+        region = EnsureChannelFirst()(region).numpy()
```
Nic-Ma, line 842: And may I know why cocnvert it to numpy array?

Thanks.

Nic-Ma (approved):
Overall looks good to me except for some minor comments inline.

Thanks.

monai/data/image_reader.py
rijobro: Prior to these changes, it was a `np.ndarray` that gets passed through. Passing a `torch.Tensor` causes knock-on problems that I didn't think were worth spending the time to fix.

Nic-Ma: OK, I see your point, I think it's OK for the current stage, could you please help add a `FIXME` comment?

Thanks!

monai/data/image_reader.py
drbeh: That's correct! `WSIReader` in `image_reader.py` is going to be deprecated in favor of the one in `wsi_reader.py`. Thanks

monai/data/image_reader.py
rijobro: Done.

Nic-Ma:
Hi @rijobro ,

As this PR is going to merge, could you please help add the doc-string for args of `update_meta`:
https://github.com/Project-MONAI/MONAI/blob/feature/MetaTensor/monai/data/meta_tensor.py#L117
I got confused before checking the code details.

Thanks in advance.

rijobro:
> Hi @rijobro ,
As this PR is going to merge, could you please help add the doc-string for args of update_meta:
https://github.com/Project-MONAI/MONAI/blob/feature/MetaTensor/monai/data/meta_tensor.py#L117
I got confused before checking the code details.
Thanks in advance.

Sure @Nic-Ma, will do.

Nic-Ma (approved):

rijobro:
@drbeh I needed to change [this line](https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/post/array.py#L736) from `self.filter.to(prob_map)` to self.filter.to(prob_map.device)`. Does this seem correct to you? Was the original code just to make sure the filter was on the same device as `prob_map`?

drbeh:
> @drbeh I needed to change [this line](https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/post/array.py#L736) from `self.filter.to(prob_map)` to `self.filter.to(prob_map.device)`. Does this seem correct to you? Was the original code just to make sure the filter was on the same device as `prob_map`?

@rijobro you are right but `self.filter.to(prob_map)` is equivalent of `self.filter.to(prob_map.device).to(prob_map.dtype)` and I'm not sure if we need the `dtype` here. I can double check if you really want to change it.

wyli (approved):
mostly tested, merging it to trigger full integration tests.

PR #4236, opened 2022-05-06
gradient based saliency

rijobro:
Adds `VanillaGrad`, `SmoothGrad` and guided backprop versions of both of these, inspired by https://github.com/pikahhh/pytorch-smoothgrad.

### Status
**Ready**

### Types of changes

image


ericspod:
We should have docstrings for these definitions. The module replacer decorator looks like a useful general purpose idea that could be generalised and put somewhere as a utility.

monai/visualize/gradient_based.py
```diff
@@ -0,0 +1,163 @@
...
+    def backward(self, x: torch.Tensor) -> torch.Tensor:
+        out: torch.Tensor = _AutoGradReLU().backward(x)
+        return out
+
+
+class VanillaGrad:
+    def __init__(self, model: torch.nn.Module) -> None:
+        self.model = model
+
+    def get_grad(self, x: torch.Tensor, index: torch.Tensor | int | None) -> torch.Tensor:
```
wyli, line 70: I think this class is doing a similar thing, do you think we can reuse or improve the existing one? https://github.com/Project-MONAI/MONAI/blob/facf5cc658e645c6cba57529a97902ebbddee2b7/monai/visualize/class_activation_maps.py#L137-L145

rijobro: There's certainly overlap, I'm not sure how to harmonise the classes together though, I struggle slightly with `ModelWithHooks`. Do you think you'd be able to have a look?

wyli: Hi @rijobro this modification https://github.com/rijobro/MONAI/pull/4 passes the unit test, I havn't tested it with the updated notebook, I'll try to do it soon.

rijobro: thanks! Do you know `autograd.Function` well? It seems clunky that I've created both a `torch.autograd.Function` and a `torch.nn.Module`, I wondered if it was possible to do it with just one.

rijobro: Great, thanks.

rijobro: @wyli The results are fine, but I'm seeing this warning when using your changes on the cats and dogs notebook: 

> Backward hook for no_layer is not triggered; `requires_grad` of no_layer should be `True`.

wyli: > thanks! Do you know `autograd.Function` well? It seems clunky that I've created both a `torch.autograd.Function` and a `torch.nn.Module`, I wondered if it was possible to do it with just one.

I think the current implementation works fine, the only minor issue is `torch.nn.Module` doesn't have `backward`, so the module backward is not really needed.

wyli (approved):
verified with the unit tests and the notebook examples

PR #4239, opened 2022-05-06
Sliding Patch WSI Dataset

### LINKED ISSUE #4179: Sliding Patch WSIReader for Inference
Sliding PatchWSIDataset, that extract patches in a sliding window manner from a whole slide image considering only the tissue at a lower resolution.

drbeh:
Fixes #4179 

### Description
This PR implements `SlidingPatchWSIDataset` that extract patches from one or multiple whole slide images in a sliding-window manner. This is useful for the inference of the pathology pipelines.

### Status
**Ready**

### Types of changes


monai/data/wsi_datasets.py
```diff
@@ -135,3 +136,87 @@ def _transform(self, index: int):
...
+        **kwargs,
+    ):
+        super().__init__(data=data, size=size, level=level, transform=transform, reader=reader, **kwargs)
+        self.overlap = overlap
+
+        # Create single sample for each patch (in a sliding window manner)
+        self.data = []
+        for sample in data:
+            sliding_samples = self._make_patches(sample)
+            self.data.extend(sliding_samples)
```
wyli, line 188: could you please consider the existing implementations for patch analysis? https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/grid_dataset.py

drbeh: Thanks for your suggestion, @wyli! Taking a quick look at `iter_patch`, I think it is not generalizable to WSI patches. For its functionalities and some sub-routines, it requires the entire image as an input to work, right?
https://github.com/Project-MONAI/MONAI/blob/209db9e08129855df878634639d4c2700d9acd83/monai/data/utils.py#L192-L199

For WSI we don't have the image array in the memory although we have some information like the dimension and size. Do you think without the image array, we can leverage the existing patch analysis implementations?

drbeh:
Hi @Nic-Ma can you too help to review this PR? Thanks

monai/data/wsi_reader.py
```diff
@@ -174,7 +174,7 @@ def get_data(
             # Verify location
             if location is None:
                 location = (0, 0)
-            wsi_size = self.get_size(each_wsi, level)
+            wsi_size = self.get_size(each_wsi, 0)
```
Nic-Ma, line 177: May I know why change `level` to `0` here?

Thanks.

monai/data/wsi_datasets.py
```diff
@@ -135,3 +136,92 @@ def _transform(self, index: int):
...
+
+class SlidingPatchWSIDataset(PatchWSIDataset):
+    """
+    This dataset extracts patches from whole slide images (without loading the whole image)
+    It also reads labels for each patch and provides each patch with its associated class labels.
+
+    Args:
+        data: the list of input samples including image, location, and label (see the note below for more details).
+        size: the size of patch to be extracted from the whole slide image.
+        level: the level at which the patches to be extracted (default to 0).
```
Nic-Ma, line 149: missed the doc-string for arg `overlap`.

Thanks.

monai/data/wsi_datasets.py
wyli: sure, there are also https://github.com/Project-MONAI/MONAI/blob/209db9e08129855df878634639d4c2700d9acd83/monai/data/utils.py#L122
and https://github.com/Project-MONAI/MONAI/blob/209db9e08129855df878634639d4c2700d9acd83/monai/data/utils.py#L152
and the iterator avoids saving all the coordinates in memory, looks quite efficient.

the patch-based analysis is important for monai. if the existing utilities don't look good, could you please help improve and adapt?

monai/data/wsi_reader.py
drbeh: It was a bug that is fixed. For `wsi_size`, I need the size of the full resolution image, which is level `0`.

monai/data/wsi_datasets.py
drbeh: @wyli, I checked these functions and non of them are re-usable for whole slide patches.
There are two fundamental differences: 
|  | Array Patch | WSI Patch
| -- | -- | --
| patch definition | array slices | starting location and size 
| padding | any np.pad mode | handled by 3rd party WSI reader libraries and generally limited to constant values |
| resolution | there is no notion of resolution | patches can be defined on any resolution level but always need to be calculated in level=0

Also here in this dataset we don't want to have iterators, since we want the size of the dataset to be defined based on number of patches and usually there is no problem with having even millions of patch locations in memory.

drbeh: I'll make a similarly named function `iter_wsi_patch_location` in `monai.data.utils` and will use it in this dataset but it will be different than the other patch function.

monai/data/utils.py
```diff
@@ -149,6 +149,48 @@ def iter_patch_slices(
...
+
+    Yields:
+        Tuple of patch location
+    """
+    ndim = len(image_size)
+    patch_size = get_valid_patch_size(image_size, patch_size)
+    start_pos = ensure_tuple_size(start_pos, ndim)
+
+    # Get the patch size at level=0
+    patch_size_0 = [p * downsample for p in patch_size]
```
wyli, line 184: I think the user can directly input `patch_size_0` instead of `downsample` and `patch_size`?

monai/data/utils.py
```diff
@@ -149,6 +149,48 @@ def iter_patch_slices(
...
+    # Calculate steps, which depends on the amount of overlap
+    steps = [round(p * (1.0 - overlap)) for p in patch_size_0]
```
wyli, line 186: great to have the `overlap` param, could this be part of `iter_patch` as well?

tests/test_sliding_patch_wsi_dataset.py
```diff
@@ -226,11 +222,11 @@ def setUpClass(cls):
         cls.backend = "cucim"
 
 
-@skipUnless(has_osl, "Requires openslide")
-class TestSlidingPatchWSIDatasetOpenSlide(SlidingPatchWSIDatasetTests.Tests):
-    @classmethod
-    def setUpClass(cls):
-        cls.backend = "openslide"
+# @skipUnless(has_osl, "Requires openslide")
```
wyli, line 225: this change seems to be unrelated to #4179?

drbeh: That's commented out by mistake! Thanks for pointing it out.

tests/test_sliding_patch_wsi_dataset.py
```diff
@@ -137,12 +137,12 @@
 TEST_CASE_LARGE_0 = [
     {"data": [{"image": FILE_PATH, "level": 8, "size": (64, 50)}]},
     [
-        {"step_loc": (0, 0), "size": (64, 50), "level": 8, "ratios": (32914 / 128, 46000 / 179)},
-        {"step_loc": (0, 1), "size": (64, 50), "level": 8, "ratios": (32914 / 128, 46000 / 179)},
-        {"step_loc": (0, 2), "size": (64, 50), "level": 8, "ratios": (32914 / 128, 46000 / 179)},
-        {"step_loc": (1, 0), "size": (64, 50), "level": 8, "ratios": (32914 / 128, 46000 / 179)},
-        {"step_loc": (1, 1), "size": (64, 50), "level": 8, "ratios": (32914 / 128, 46000 / 179)},
-        {"step_loc": (1, 2), "size": (64, 50), "level": 8, "ratios": (32914 / 128, 46000 / 179)},
+        {"step_loc": (0, 0), "size": (64, 50), "level": 8, "ratio": 257.06195068359375},
```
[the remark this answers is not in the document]
drbeh: Actually, they are important changes. Now the downsampling is being read from the resolution metadata of whole slide images rather than using the image shape to infer that, so there will be a slight difference between them.

monai/data/utils.py
drbeh: I see you point but the problem is that the user cares about the `patch_size` (at the resolution level she is working with) and does not necessarily know `patch_size_0`, which is the patch size at `level=0`.

drbeh: I see you point but the problem is that the user cares about the patch_size (at the resolution level she is working with) and does not necessarily know patch_size_0, which is the patch size at level=0.

monai/data/utils.py
drbeh: @wyli I added the overlap, can you please review the changes?

drbeh:
@wyli @Nic-Ma could you please help to merge this PR? thanks

monai/data/utils.py
wyli: I think this parameter can optionally be a vector so that it supports anisotropic overlaps along the dimensions

monai/data/wsi_datasets.py
wyli: thanks for the summary, in the latest version `downsample` is not varying within `iter_wsi_patch_location`, so can we precompute `patch_size_` before line 203 here: https://github.com/Project-MONAI/MONAI/blob/1700fdab7972221efcc87baa963db68d24ff4b3a/monai/data/wsi_datasets.py#L203-L204
so that we avoid `iter_wsi_patch_location` (because duplicating the code adds maintenance efforts). Looks like `padded: bool` could be part of `iter_patch` as well?

monai/data/utils.py
drbeh: Sure, I'll make the change.

drbeh: Done!

monai/data/wsi_reader.py
```diff
@@ -85,6 +85,18 @@ def get_level_count(self, wsi) -> int:
         """
         raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
 
+    @abstractmethod
+    def get_downsample(self, wsi, level: int) -> float:
```
Nic-Ma, line 89: Maybe update the function name to make it more clear: `get_downsample_ratio()`?

Thanks.

Nic-Ma (commented):
Hi @wyli @ericspod ,

Do you have any other comments?

Thanks.

monai/data/utils.py
```diff
@@ -254,6 +266,52 @@ def iter_patch(
...
+    patch_size_ = tuple(p * downsample for p in patch_size)
+
+    # calculate steps, which depends on the amount of overlap
+    steps = tuple(round(p * (1.0 - o)) for p, o in zip(patch_size_, overlap))
+
+    # calculate the last starting location (depending on the padding)
+    end_pos = image_size if padded else tuple(s - round(p) + 1 for s, p in zip(image_size, patch_size_))
+
+    # evaluate the starting locations for patches
+    return product(*tuple(starmap(range, zip(start_pos, end_pos, steps))))
```
ericspod, line 312:
```suggestion
    return product(*starmap(range, zip(start_pos, end_pos, steps)))
```

monai/data/wsi_datasets.py
drbeh: That's a good idea! However, since `iter_patch_slices` return `slice` while here we just need the starting position, I will modify it to have one function `iter_patch_position`.

monai/data/utils.py
```diff
@@ -243,9 +258,12 @@ def iter_patch(
...
     iter_size = tuple(s + p for s, p in zip(arr.shape, patch_size_))
 
-    for slices in iter_patch_slices(iter_size, patch_size_, start_pos_padded):
+    for position in iter_patch_position(
+        image_size=iter_size, patch_size=patch_size_, start_pos=start_pos_padded, overlap=overlap, padded=True
+    ):
+        slices = tuple(slice(s, s + p) for s, p in zip(position, patch_size_))
         # compensate original image padding
-        coords_no_pad = tuple((coord.start - p, coord.stop - p) for coord, p in zip(slices, patch_size_))
+        coords_no_pad = tuple((s - p, s) for s, p in zip(position, patch_size_))
```
ericspod, line 266: This calculation used to be in `iter_patch_slices`, things have been changed around but I'd rather this slice calculation be done in `iter_patch_position` for use cases where access to the slices themselves is needed. Otherwise places where a slice or range is needed again will have to duplicate what this line is doing which is a bit error-prone.

drbeh: I brought back `iter_patch_slices` and the calculation is being done there now. However, since this calculation it the only difference between `iter_patch_slices` and `iter_patch_position`, now `iter_patch_slices` calls `iter_patch_position` and just do this calculation.

monai/data/wsi_datasets.py
drbeh: I made some modification to remove any redundancy. Now we have `iter_patch_position` that returns the upper left corner of each patch (this is what we need for wsi patches), and `iter_patch_slices` that calls `iter_patch_position` and creates slices using the positions.

drbeh: @wyli let me know if you have any other comments. Thanks

wyli: thanks, the new impl. are more generic than the current `dev`'s, that looks good to me.

monai/data/utils.py
```diff
@@ -192,10 +163,88 @@ def dense_patch_slices(
...
+
+def iter_patch_position(
+    image_size: Sequence[int],
+    patch_size: Union[Sequence[int], int],
+    start_pos: Sequence[int] = (),
+    overlap: Union[Sequence[float], float] = 0.0,
+    padded: bool = False,
+):
+    """
+    Yield successive tuples of slices defining patches of size `patch_size` from an array of dimensions `dims`. The
```
wyli, line 208: please review this docstring

wyli:
there's a test error...
```
[2022-05-25T13:08:02.322Z] ======================================================================
[2022-05-25T13:08:02.322Z] FAIL: test_read_patches_9 (tests.test_sliding_patch_wsi_dataset.TestSlidingPatchWSIDatasetCuCIM)
[2022-05-25T13:08:02.322Z] ----------------------------------------------------------------------
[2022-05-25T13:08:02.322Z] Traceback (most recent call last):
[2022-05-25T13:08:02.322Z]   File "/opt/conda/lib/python3.8/site-packages/parameterized/parameterized.py", line 533, in standalone_func
[2022-05-25T13:08:02.322Z]     return func(*(a + p.args), **p.kwargs)
[2022-05-25T13:08:02.322Z]   File "/home/jenkins/agent/workspace/MONAI-premerge/monai/tests/test_sliding_patch_wsi_dataset.py", line 227, in test_read_patches
[2022-05-25T13:08:02.322Z]     self.assertEqual(len(dataset), len(expected))
[2022-05-25T13:08:02.322Z] AssertionError: 1 != 2
[2022-05-25T13:08:02.322Z] 
[2022-05-25T13:08:02.322Z] ----------------------------------------------------------------------

```
with `nvcr.io/nvidia/pytorch:22.04-py3` V100

drbeh:
> test_sliding_patch_wsi_dataset.py

You're right! I know, I don't know why I forgot to do that.

drbeh:
> there's a test error...
> 
> ```
> [2022-05-25T13:08:02.322Z] ======================================================================
> [2022-05-25T13:08:02.322Z] FAIL: test_read_patches_9 (tests.test_sliding_patch_wsi_dataset.TestSlidingPatchWSIDatasetCuCIM)
> [2022-05-25T13:08:02.322Z] ----------------------------------------------------------------------
> [2022-05-25T13:08:02.322Z] Traceback (most recent call last):
> [2022-05-25T13:08:02.322Z]   File "/opt/conda/lib/python3.8/site-packages/parameterized/parameterized.py", line 533, in standalone_func
> [2022-05-25T13:08:02.322Z]     return func(*(a + p.args), **p.kwargs)
> [2022-05-25T13:08:02.322Z]   File "/home/jenkins/agent/workspace/MONAI-premerge/monai/tests/test_sliding_patch_wsi_dataset.py", line 227, in test_read_patches
> [2022-05-25T13:08:02.322Z]     self.assertEqual(len(dataset), len(expected))
> [2022-05-25T13:08:02.322Z] AssertionError: 1 != 2
> [2022-05-25T13:08:02.322Z] 
> [2022-05-25T13:08:02.322Z] ----------------------------------------------------------------------
> ```
> 
> with `nvcr.io/nvidia/pytorch:22.04-py3` V100, please let me know if you need more info to debug
> 
> I think the class could inherit `Randomizable` to ensure the deterministic behaviour e.g.
> 
> https://github.com/Project-MONAI/MONAI/blob/eb60b9caf0a103af53ab5f3d620ceb2c116a4851/monai/apps/datasets.py#L33

@wyli 
I'll add the `Randomizable` but with `set_determinisim` (and using numpy random), I'm wondering why it should produce different results! Do you have any idea?

wyli:
I guess because there are multiple `skipUnless` in the tests, depending on the skipping conditions the `np.random.randint` is called various times: https://github.com/Project-MONAI/MONAI/blob/b520d6820441476eb7efc99865badc90c0be2219/monai/data/wsi_datasets.py#L232 
with the randomizable API, it's possible to use `self.R.randint(low, high)`, so that it controls the local random state.  pls see the `seed` usage: (https://github.com/Project-MONAI/MONAI/blob/db61a08a904aaa9751b1ce4e5501fa0f102fce47/tests/test_mednistdataset.py#L59)

wyli:
I think the current errors are irrelevant to the PR, I'm working on it

PR #4240, opened 2022-05-07
[DO NOT REVIEW] dummy test: verify blackduck status [skip ci]

pxLi:
test dummy only


pxLi:
confirmed failing `Could not communicate with Black Duck` in multiple repos. Reported issue to internal infra owners

PR #4250, opened 2022-05-10
Update to Bundle Specifiation

ericspod:

### Description
From previous conversations it seemed like updates to the specification to clarify some points was needed. I have added language to make it clear some files are required with their specific names, clarified zipped bundles, and added a mention for ONNX saved models which we would like to support. There was other points discussed that perhaps should be clarified here so please propose anything further to update.

### Status
**Ready**

### Types of changes

### DIFF
--- a/docs/source/mb_specification.rst
+++ b/docs/source/mb_specification.rst
@@ -6,7 +6,7 @@ MONAI Bundle Specification
 Overview
 ========
 
-This is the specification for the MONAI Bundle (MB) format of portable described deep learning models. The objective of a MB is to define a packaged network or model which includes the critical information necessary to allow users and  programs to understand how the model is used and for what purpose. A bundle includes the stored weights of a model as a pickled state dictionary and/or a Torchscript object. Additional JSON files are included to store metadata about the model, information for constructing training, inference, and post-processing transform sequences, plain-text description, legal information, and other data the model creator wishes to include.
+This is the specification for the MONAI Bundle (MB) format of portable described deep learning models. The objective of a MB is to define a packaged network or model which includes the critical information necessary to allow users and programs to understand how the model is used and for what purpose. A bundle includes the stored weights of a single network as a pickled state dictionary plus optionally a Torchscript object and/or an ONNX object. Additional JSON files are included to store metadata about the model, information for constructing training, inference, and post-processing transform sequences, plain-text description, legal information, and other data the model creator wishes to include.
 
 This specification defines the directory structure a bundle must have and the necessary files it must contain. Additional files may be included and the directory packaged into a zip file or included as extra files directly in a Torchscript file.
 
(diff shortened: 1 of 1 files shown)


ericspod:
@wyli @Nic-Ma @daguangxu @yiheng-wang-nv  We had discussed some details in the bundle to clarify, anything other changes?

docs/source/mb_specification.rst
```diff
@@ -6,7 +6,7 @@ MONAI Bundle Specification
...
+This is the specification for the MONAI Bundle (MB) format of portable described deep learning models. The objective of a MB is to define a packaged network or model which includes the critical information necessary to allow users and programs to understand how the model is used and for what purpose. A bundle includes the stored weights of a single network as a pickled state dictionary plus optionally a Torchscript object and/or an ONNX object. Additional JSON files are included to store metadata about the model, information for constructing training, inference, and post-processing transform sequences, plain-text description, legal information, and other data the model creator wishes to include.
```
Nic-Ma, line 9: Is it necesssary to explicitly say `ONNX` in this doc?
@wyli What do you think?

Thanks.

docs/source/mb_specification.rst
```diff
@@ -22,26 +22,33 @@ A MONAI Bundle is defined primarily as a directory with a set of specifically na
   ┃  ┗━ metadata.json
   ┣━ models
   ┃  ┣━ model.pt
-  ┃  ┗━ model.ts
+  ┃  ┣━ model.ts
+  ┃  ┗━ model.onnx
```
Nic-Ma, line 26: Similar question as above, maybe put "*" for `model.ts` and `model.onnx` as they are optional?

Thanks.

docs/source/mb_specification.rst
```diff
@@ -22,26 +22,33 @@ A MONAI Bundle is defined primarily as a directory with a set of specifically na
...
      ┗━ license.txt
 
 
-These files mostly are required to be present with the given names for the directory to define a valid bundle:
+The following files are **required** to be present with the given filenames for the directory to define a valid bundle:
 
 * **metadata.json**: metadata information in JSON format relating to the type of model, definition of input and output tensors, versions of the model and used software, and other information described below.
 * **model.pt**: the state dictionary of a saved model, the information to instantiate the model must be found in the metadata file.
+
+The following files are optional but must have these names in the directory given above:
```
Nic-Ma, line 37: I think this is the discussion in Yiheng's previous PR, @wyli do you also think we should force the file names here?

Thanks.

wyli: Which pr are you referring to?

docs/source/mb_specification.rst
wyli: I think the texts described them as optional and the current PR improves the specification. If we identified new format that could be allowed as optional file, we can update this list. What do you think?

docs/source/mb_specification.rst
ericspod: It's not necessary but I thought to add it here to future proof a bit.

docs/source/mb_specification.rst
ericspod: I think that's fine, I added the stars anyway.

monai/bundle/scripts.py
```diff
@@ -600,10 +600,15 @@ def ckpt_export(
...
         filename, _ = os.path.splitext(filename)
+        # because all files are stored as JSON their name parts without extension must be unique
         if filename in extra_files:
-            raise ValueError(f"filename '{filename}' is given multiple times in config file list.")
+            raise ValueError(f"Filename part '{filename}' is given multiple times in config file list.")
+        # the file may be JSON or YAML but will get loaded and dumped out again as JSON
         extra_files[filename] = json.dumps(ConfigParser.load_config_file(i)).encode()
 
+    # add .json extension to all extra files which are always encoded as JSON
+    extra_files = {k + ".json": v for k, v in extra_files.items()}
```
Nic-Ma, line 610: I think you need to update the corresponding unit tests of `ckpt_export`.

Thanks.

docs/source/mb_specification.rst
Nic-Ma: Hi @wyli ,

I mean this question for you: https://github.com/Project-MONAI/MONAI/pull/4114#discussion_r854211411.

Thanks.

monai/bundle/scripts.py
ericspod: Done.

docs/source/mb_specification.rst
wyli: Ok I think we could have a primary 'model.pt', and optionally have some additional 'model_*.pt' files here

ericspod: The important thing from my perspective is that having fixed filenames means we can rely on that assumption in other code which consumes bundles, ie. the deploy operator. It's easier if we know the files `model.pt` and `metadata.json` are always there. Bundles also store only one model because it's so much easier to provide the metadata for one network and also because that's what Torchscript supports in its format, but having other `.pt` files for alternative weights or other data is perfectly fine with any other filename the user wants. I can discuss other optional files in more detail if that would help in the specification.

wyli (approved):
Thanks, it looks good to me.

Nic-Ma (approved):

PR #4251, opened 2022-05-10
Split patch functionality for `PatchWSIReader`

### LINKED ISSUE #4247: Split WSI regions into sub-patches without new dimension
`GridSplit` stack up all the split patches in a new dimension, which makes sense; however, for some use cases like the classification-based pathology tumor detection, we want to treat each of these patches as a separate image, not add new dimension and batch them together via data loader.

It seems that the best way to solve this situation is to add an option for split patches in the `PatchWSIDataset`, so it can split the region into sub-patches and treat them like separate images.

drbeh:
Fixes #4247 

### Description
This PR add split patch functionality to `PatchWSIReader`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -2493,8 +2493,12 @@ class GridSplit(Transform):
         # Patch size
         self.size = None if size is None else ensure_tuple_rep(size, len(self.grid))
 
-    def __call__(self, image: NdarrayOrTensor) -> List[NdarrayOrTensor]:
-        if self.grid == (1, 1) and self.size is None:
+    def __call__(
+        self, image: NdarrayOrTensor, size: Optional[Union[int, Tuple[int, int], np.ndarray]] = None
+    ) -> List[NdarrayOrTensor]:
+        input_size = self.size if size is None else ensure_tuple_rep(size, len(self.grid))
+
+        if self.grid == (1, 1) and input_size is None:
             if isinstance(image, torch.Tensor):
                 return [image]
             elif isinstance(image, np.ndarray):
@@ -2502,11 +2506,14 @@ class GridSplit(Transform):
             else:
                 raise ValueError(f"Input type [{type(image)}] is not supported.")
 
-        size, steps = self._get_params(image.shape[1:])
+        split_size, steps = self._get_params(image.shape[1:], input_size)
         patches: List[NdarrayOrTensor]
         if isinstance(image, torch.Tensor):
             unfolded_image = (
-                image.unfold(1, size[0], steps[0]).unfold(2, size[1], steps[1]).flatten(1, 2).transpose(0, 1)
+                image.unfold(1, split_size[0], steps[0])
+                .unfold(2, split_size[1], steps[1])
+                .flatten(1, 2)
+                .transpose(0, 1)
             )
             # Make a list of contiguous patches
             patches = [p.contiguous() for p in unfolded_image]
@@ -2516,7 +2523,7 @@ class GridSplit(Transform):
             n_channels = image.shape[0]
             strided_image = as_strided(
                 image,
-                shape=(*self.grid, n_channels, size[0], size[1]),
+                shape=(*self.grid, n_channels, split_size[0], split_size[1]),
                 strides=(x_stride * x_step, y_stride * y_step, c_stride, x_stride, y_stride),
                 writeable=False,
             )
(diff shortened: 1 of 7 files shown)


Nic-Ma:
Thanks for the implementation of new feature request.
I am wondering if it's possible to support the `split patch functionality` in a transform, which handles the `output` of `PatchWSIDataset` with other transforms?

Thanks.

wyli:
I lose the overall picture of the WSI-related reader/dataset modules, would be great to have some tutorials in https://github.com/Project-MONAI/tutorials/tree/master/modules to demonstrate the use cases, otherwise it's not easy to help review these PRs.

wyli:
(For an example of the tutorials: this PR proposes code changes https://github.com/Project-MONAI/MONAI/pull/4236, and the corresponding usage is shown in a PR to the tutorials https://github.com/Project-MONAI/tutorials/pull/690. These greatly help the reviewing/discussion process...)

drbeh:
@Nic-Ma @ericspod, this split does not have any randomness (like what we have discussed before) and I had in mind to use the transform only but the reason to have the split here is to managing the following transforms. What I mean is that the split transform generate several patches (from few patches to few hundreds or thousands of patches), and we want these patches treated independently like images in a batch. It was not clear to me how its output will be handle by its following transforms. Assume the following:

- input patch: (3, 16, 16)
- (3, 20, 24) -> GridSplit (2x2) -> (4, 3, 5, 6) [mind the additional dimension]
- (4, 3, 5, 6) -> OTHER TRANSFORMS -> (4, 3, 5, 6) [does all other transform acknowledge the batch dimension?]
- (4, 3, 5, 6) -> BATCHING (10) -> (40, 3, 5, 6) [not (10, 4, 3, 5, 6), can we achieve this without a new collate function?]

I would really like to make this as a separate transform so I'd appreciate it if you have any solution that can create the batches as we like (without implementing a new collate function).

Nic-Ma:
Hi @drbeh ,

I think your problem is similar to our `RandCropByPosNegLabeld` transform which generates a list of patches:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/croppad/dictionary.py#L1042
The following transforms will apply for each of the item in the patch list:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/transform.py#L88

Thanks.

drbeh:
> https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/transform.py#L88

@Nic-Ma, this is what the current implementation is already leveraging and that's why I put `GridSplit` in the dataset so the transform chain can be applied on each of them.

However, this is not the case if a transform at the middle of chain creates a list of images, right?

Nic-Ma:
> RandCropByPosNegLabeld

`RandCropByPosNegLabeld` can be in the middle of a transform chain, and we also have several other transforms generate a list of samples.
`monai.data.DataLoader` can batch for this case with the built-in collate function.

Thanks.

drbeh:
@Nic-Ma @ericspod, I have updated the `GridSplit` and now the transform and dataset are completely separated. Thanks for your inputs.

@wyli, I have updated the tumor detection pipeline that uses these transforms and dataset: https://github.com/Project-MONAI/tutorials/pull/697

monai/transforms/spatial/array.py
```diff
@@ -2535,62 +2535,67 @@ def __init__(self, grid: Tuple[int, int] = (2, 2), size: Optional[Union[int, Tup
...
         self.size = None if size is None else ensure_tuple_rep(size, len(self.grid))
 
-    def __call__(self, image: NdarrayOrTensor) -> NdarrayOrTensor:
-        if self.grid == (1, 1) and self.size is None:
+    def __call__(
+        self, image: NdarrayOrTensor, size: Optional[Union[int, Tuple[int, int], np.ndarray]] = None
+    ) -> List[NdarrayOrTensor]:
+        input_size = self.size if size is None else ensure_tuple_rep(size, len(self.grid))
+
+        if self.grid == (1, 1) and input_size is None:
```
Nic-Ma, line 2543: Here I think you can simplify the code to `return [image]` as we already suppose the input image is `tensor` or `numpy array`.

Thanks.

Nic-Ma (dismissed):
Thanks for the update.
Looks good to me except for some minor comments inline.
And I would suggest to add description for the return data shape in the doc-string.
@wyli @ericspod do you have any other comments?

Thanks.

drbeh:
> > @wyli, I have updated the tumor detection pipeline that uses these transforms and dataset: [Project-MONAI/tutorials#697](https://github.com/Project-MONAI/tutorials/pull/697)
> 
> thanks, @Nic-Ma could you please help run and verify the tutorial (perhaps with a subset of CAMELYON16 for a quick setup)?

That's a good idea. I'll add a subset dataset to the PR of the tutorial to facilitate the testing but we can remove it before merge.

drbeh:
@Nic-Ma @wyli, I have added a subset dataset ('training.txt') into this tutorial PR: https://github.com/Project-MONAI/tutorials/pull/697

- You can use this dataset for both training and validation with a flag of `--train-file ../training.txt --valid-file ../training.txt` since we don't want to check any performance here.

- You also need to download one image `tumor_091.tif` and store it in `IMAGE_ROOOT/training/images/`, where `IMAGE_ROOT` is arbitrary and can be set using `--root IMAGE_ROOT`: https://drive.google.com/file/d/0BzsdkU4jWx9BVWx1eERwYnlVWVE/view?usp=sharing&resourcekey=0-Zd3y-ucNOjPq3nqsx-P3sQ

Please let me know if you have any question.

Nic-Ma:
Hi @drbeh @wyli ,

I tried to run the tutorial, seems it can run normally:
image

Thanks.

monai/transforms/spatial/dictionary.py
```diff
@@ -2174,17 +2174,23 @@ def __init__(
         self,
         keys: KeysCollection,
         grid: Tuple[int, int] = (2, 2),
-        size: Optional[Union[int, Tuple[int, int]]] = None,
+        size: Optional[Union[int, Tuple[int, int], Dict[Hashable, Union[int, Tuple[int, int], None]]]] = None,
```
Nic-Ma, line 2177: May I know why you change the `size` to be a Dict?
I didn't see the use case to define different size for keys.
And the corresponding doc-string of this arg says it's a tuple.

Thanks.

drbeh: You can see the use case in the tutorial: https://github.com/Project-MONAI/tutorials/blob/27717a16a143a035f889c75e8fb6e7bad3cbffcc/pathology/tumor_detection/ignite/camelyon_train_evaluate.py#L84

I'll update the docstring.

Nic-Ma (approved):

PR #4270, opened 2022-05-13
Recursive NVTX Range

drbeh:
### Description
This PR adds recursive functionality to MONAI's NVTX Range. This enables annotating all individual transforms that are chained by using only one `Range(recursive=True).

### Status
**Ready**

### Types of changes


wyli (approved):

monai/utils/nvtx.py
```diff
@@ -53,12 +57,28 @@ def __init__(
...
         self.methods = methods
         self.append_method_name = append_method_name
+        self.recursive = recursive
 
     def __call__(self, obj: Any):
+        if self.recursive is True:
+            from monai.transforms.compose import Compose
+
+            if isinstance(obj, Compose):
+                annotated_transforms = [Range(recursive=False)(t) for t in obj.transforms]
```
wyli, line 72: will this support the transforms with recursive compose of compose or `OneOf`?

monai/utils/nvtx.py
```diff
@@ -21,7 +22,8 @@
 from torch.optim import Optimizer
 from torch.utils.data import Dataset
 
-from monai.utils import ensure_tuple, optional_import
+from monai.utils.misc import ensure_tuple
+from monai.utils.module import optional_import
```
wyli, line 26: could you please revert this one because it's not necessary change

drbeh:
Hi @wyli, I added support for Compose/OneOf of Compose/OneOf. do you have any other comment? thanks.

monai/utils/nvtx.py
```diff
@@ -53,12 +56,46 @@ def __init__(
...
 
     def __call__(self, obj: Any):
+        if self.recursive is True:
+            if isinstance(obj, Iterable):
+                annotated_transforms = [Range(recursive=True)(t) for t in obj]
+                return Range(self.name, recursive=False)(annotated_transforms)
+
+            from monai.transforms.compose import Compose, OneOf
+
+            if isinstance(obj, OneOf):
```
wyli, line 74: I think line 74 and 86 could be merged using `type(obj)`:
```py
if isinstance(obj, Compose):
    ...
    Range(name)(type(obj)(...))
```

wyli, line 74: I think line 74 and 86 could be merged using `type(obj)`:
```py
if isinstance(obj, Compose):
    ...
    Range(name)(type(obj)(...))
```

drbeh: ditto

monai/utils/nvtx.py
```diff
@@ -53,12 +56,46 @@ def __init__(
...
+                return Range(self.name, recursive=False)(
+                    Compose(
+                        annotated_transforms,
+                        map_items=obj.map_items,
+                        unpack_items=obj.unpack_items,
+                        log_stats=obj.log_stats,
+                    )
+                )
+
+            self.recursive = False
```
wyli, line 97: could the overall recursion be simplified to this or similar ideas of in-place change, could you please help test out?

```suggestion
            if isinstance(obj, Iterable):
                return [Range(recursive=True)(t) for t in obj]

            from monai.transforms.compose import Compose, OneOf

            if isinstance(obj, Compose):
                obj.transforms = Range(recursive=True)(obj.transforms)
            self.recursive = False
```

drbeh: Yeah, that's cool. Except for a caveat that is missed in the unittests for Iterables. I'll update it.

monai/utils/nvtx.py
```diff
@@ -53,12 +55,26 @@ def __init__(
...
     def __call__(self, obj: Any):
+        if self.recursive is True:
+            if isinstance(obj, (list, tuple)):
+                return type(obj)(Range(recursive=True)(t) for t in obj)
+
+            from monai.transforms.compose import Compose
+
+            if isinstance(obj, Compose):
+                obj.transforms = Range(recursive=True)(obj.transforms)
+                return Range(self.name, recursive=False)(obj)
```
wyli, line 74: thanks, looks like this line created an additional `Range` object which is not really needed? 
```suggestion
```

wyli, line 74: thanks, looks like this line created an additional `Range` object which is not really needed? 
```suggestion
```

drbeh: It annotates the whole scope of transforms that a compose defines. For instance, we want to annotation "Preprocessing" part that consists many transforms, each of which we also want to be annotated.
```python
    train_preprocess = Compose(
        [
            ToNumpyd(keys="image"),
            RandFlipd(keys="image", prob=0.5),
            RandRotate90d(keys="image", prob=0.5, max_k=3, spatial_axes=(-2, -1)),
            CastToTyped(keys="image", dtype=np.float32),
            RandZoomd(keys="image", prob=0.5, min_zoom=0.9, max_zoom=1.1),
            ScaleIntensityRanged(keys="image", a_min=0.0, a_max=255.0, b_min=-1.0, b_max=1.0),
            ToTensord(keys=("image", "label")),
        ]
    )
    train_preprocess = Range("Preprocessing", recursive=True)(train_preprocess)

```

wyli: yes, when it reaches line 73, it's already within a call of `Range(recursive=True)(Compose)`, the program can continue the annotating instead of starting a new object and call `Range(recursive=False)(Compose)` again (line 74).

Issue #4303, opened 2022-05-20
RandAffine cannot return same transform when randomize param set to False.


fluentgcc:
**Describe the bug**

RandAffine cannot return same transform when randomize param set to False.

**To Reproduce**
```python
from monai.transforms.spatial.array import RandAffine

rand_affine = RandAffine(
            prob=1,
            rotate_range=(np.pi/6, 0, 0),
            translate_range=((-2, 2),(-2, 2),(-2, 2)),
            scale_range=((-0.1,0.1),(-0.1, 0.1),(-0.1,0.1)),
            spatial_size=(16,16,16),
            cache_grid=True,
            padding_mode = "zeros"
    
        )

arr = torch.randn((1, 16,16,16))*100

rand_affine.randomize(None)

arr1 = rand_affine(arr ,randomize=False)
m1 = rand_affine.rand_affine_grid.get_transformation_matrix()
print(m1)
arr2 = rand_affine(arr, randomize=False)
m1 = rand_affine.rand_affine_grid.get_transformation_matrix()
print(m1)

print(torch.sum(arr1-arr2))
```

**Expected behavior**
when randomize=False, return the same result.

**Screenshots**
<img width="943" alt="截屏2022-05-20 16 01 47" src="https://user-images.githubusercontent.com/8706549/169482693-0d10a917-9403-4afe-8855-eeecece7a7b7.png">
<img width="743" alt="截屏2022-05-20 16 01 59" src="https://user-images.githubusercontent.com/8706549/169482747-86393efb-a93e-49b8-9f13-1e2a5124a14a.png">

**Environment**

Ensuring you use the relevant python executable, please paste the output of:

```
python -c 'import monai; monai.config.print_debug_info()'
```
```
================================
Printing MONAI config...
================================
MONAI version: 0.8.1+257.g6eaa3ab1.dirty
Numpy version: 1.21.6
Pytorch version: 1.11.0
MONAI flags: HAS_EXT = True, USE_COMPILED = False
MONAI rev id: 6eaa3ab1b0f5ba568d69785cefbdb1cde74d259a
MONAI __file__: /ssd/libs/MONAI/MONAI/monai/__init__.py

Optional dependencies:
Pytorch Ignite version: NOT INSTALLED or UNKNOWN VERSION.
Nibabel version: NOT INSTALLED or UNKNOWN VERSION.
scikit-image version: 0.19.2
Pillow version: 9.1.1
[...]

PR #4304, opened 2022-05-20
fix RandAffine with randomize==False

rijobro:
Fixes https://github.com/Project-MONAI/MONAI/issues/4303.

### Description
Pass the `randomize` flag on to `RandAffineGrid` so that it randomises at the correct moment. Unit test added to ensure this.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -1529,6 +1529,8 @@ class RandAffineGrid(Randomizable, Transform):
         Args:
             spatial_size: output grid size.
             grid: grid to be transformed. Shape must be (3, H, W) for 2D or (4, H, W, D) for 3D.
+            randomize: boolean as to whether the grid parameters governing the grid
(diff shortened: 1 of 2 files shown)


rijobro:
@fluentgcc Thanks for spotting the bug, this should solve the issue.

monai/transforms/spatial/array.py
```diff
@@ -1530,7 +1533,8 @@ def __call__(
         Returns:
```
Nic-Ma, line 1533: Could you please help add the new arg to doc-string?

Thanks.

Nic-Ma (approved):
Thanks for the quick fix.
Looks good to me, put minor comments inline.

Thanks.

monai/transforms/spatial/array.py
rijobro: Good point, done!

PR #4308, opened 2022-05-20
multiprocessing test metatensor

wyli:

- Fixes https://github.com/Project-MONAI/MONAI/runs/6521801558?check_suite_focus=true

### Description
may need to register new logic to the pickler
according to https://github.com/pytorch/pytorch/blob/53e0d7a3bae621ef0d07cfd44957736e9149dd8e/torch/_tensor.py#L77-L79

cc @rijobro 

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/meta_tensor.py
+++ b/monai/data/meta_tensor.py
@@ -60,6 +60,7 @@ class MetaTensor(MetaObj, torch.Tensor):
         - Older versions of pytorch (<=1.8), `torch.jit.trace(net, im)` may
           not work if `im` is of type `MetaTensor`. This can be resolved with
           `torch.jit.trace(net, im.as_tensor())`.
+        - For pytorch < 1.8, sharing `MetaTensor` instances across processes may not be supported.
         - A warning will be raised if in the constructor `affine` is not `None` and
           `meta` already contains the key `affine`.
         - You can query whether the `MetaTensor` is a batch with the `is_batch` attribute.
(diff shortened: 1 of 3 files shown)


rijobro:
Should these not go into `dev` rather than `feature/MetaTensor`?

tests/test_meta_tensor.py
```diff
@@ -488,6 +491,15 @@ def test_construct_with_pre_applied_transforms(self):
         m = MetaTensor(im, applied_operations=data[PostFix.transforms(key)])
         self.assertEqual(len(m.applied_operations), len(tr.transforms))
 
+    @SkipIfBeforePyTorchVersion((1, 8, 1))
+    def test_multiprocessing(self):
```
rijobro, line 495: Is this test alone enough to satisfy ourselves that multi-processing is supported?

wyli: this is mainly to fix the current integration error, I think there might be more issues when we refactor more components, so I think maybe it's fine to merge this to `feature/MetaTensor` first...

wyli:
> Should these not go into `dev` rather than `feature/MetaTensor`?

thanks @rijobro, I think this one is now good for `dev`, it also fixes a constructor issue identified during testing:

```py
>>> from monai.data import MetaTensor
>>> MetaTensor([1.0], device="cpu")
TypeError: __init__() got an unexpected keyword argument 'device'

```

monai/data/meta_tensor.py
```diff
@@ -82,10 +83,17 @@ def __new__(
         *args,
         **kwargs,
     ) -> MetaTensor:
-        return torch.as_tensor(x, *args, **kwargs).as_subclass(cls)  # type: ignore
+        _kwargs = {"device": kwargs.pop("device", None), "dtype": kwargs.pop("dtype", None)} if kwargs else {}
```
rijobro, line 86: If doing `device`, should we also do `dtype`?

wyli: yes, the current pr supports both `device` and `dtype`

rijobro: Oh sorry I missed that!

rijobro (approved):

PR #4319, opened 2022-05-23
Spacing MetaTensor

rijobro:
Spacing compatible with `MetaTensor`. Array transform now also has `inverse` method.

### Status
**Ready**

### Types of changes


monai/transforms/spatial/array.py
```diff
@@ -425,16 +432,28 @@ def __call__(
             data_array (resampled into `self.pixdim`), original affine, current affine.
 
         """
-        sr = int(data_array.ndim - 1)
+        # if the input isn't MetaTensor and output isn't desired to be one either,
+        # nothing to do.
+        if not isinstance(data_array, MetaTensor) and not get_track_meta():
+            warnings.warn(
+                "`Orientation` applied to non-MetaTensor object and metadata tracking is off, transform does nothing."
+            )
```
wyli, line 440: I think this checking should be a utility function so it'd be easier to maintain

monai/transforms/spatial/array.py
```diff
@@ -425,16 +432,28 @@ def __call__(
...
+        affine_np: np.ndarray
+        if isinstance(data_array, MetaTensor):
+            affine_np, *_ = convert_data_type(data_array.affine, np.ndarray)
+            affine_ = to_affine_nd(sr, affine_np)
+        else:
+            warnings.warn("`data_array` is not of type `MetaTensor, assuming affine to be identity.")
             # default to identity
-            affine_np = affine = np.eye(sr + 1, dtype=np.float64)
+            affine_np = np.eye(sr + 1, dtype=np.float64)
             affine_ = np.eye(sr + 1, dtype=np.float64)
```
wyli, line 456: (might be better to reuse `MetaTensor.get_default_affine` in the future)

wyli (approved):
thanks the transform becomes easier to use with MetaTensor

monai/transforms/spatial/array.py
rijobro: Hm, default MetaTensor affine is always torch.eye(4), whereas here we allow for 3x3 etc matrices.

rijobro:
@wyli Do you think you can help me with [this error](https://github.com/Project-MONAI/MONAI/runs/6556706016?check_suite_focus=true#step:9:13718) in the github action? I can't reproduce it locally. I think it's complaining that `img_.device` doesn't work when `img_` is a `np.ndarray`. But the code before looks like this:

```python
if get_track_meta():
    img_ = img if isinstance(img, MetaTensor) else MetaTensor(img)
    img_, *_ = convert_data_type(img, dtype=_dtype)
else:
    img_, *_ = convert_data_type(img, torch.Tensor, dtype=_dtype)
```
It seems to me that in all cases, `img_` should be of type `torch.Tensor` and so `device`` should be available.

wyli:
> `img_` should be of type `torch.Tensor` and so `device`` should be available.

looks like you addressed the issue... I'll merge this PR and https://github.com/Project-MONAI/MONAI/pull/4308 and have the full integration tests soon.

PR #4321, opened 2022-05-23
`GridPatch` to Extract Tiles/Patches

### LINKED ISSUE #4012: Generalized Tile transform
Refactor and possibly generalize `TileOnGrid` transform in pathology, for blending it into core MONAI as laid out in #4005. 

https://github.com/Project-MONAI/MONAI/blob/a676e3876903e5799eb3d24245872ea4f3a5b152/monai/apps/pathology/transforms/spatial/array.py#L110-L262

drbeh:
Fixes #4012 

### Description
This PR implement `GridPath` and `GridPatchd` that extracts patches sweeping the entire image. It will return a list of patches (and not a generator) to have a priorly defined number of patches, which is useful for the subsequent transforms and handlers.

### Status

### Types of changes


drbeh:
@wyli @Nic-Ma This PR requires PR #4239 to be merged first. Thanks

monai/transforms/spatial/array.py
```diff
@@ -2609,3 +2619,168 @@ def _get_params(
...
+            if sort_key == "min":
+                self.sort_key = self._get_patch_sum
+            if sort_key == "max":
+                self.sort_key = self._get_negative_patch_sum
+            else:
+                ValueError(f'sort_key should be either "min", "max", or "random", "{sort_key}" was given.')
+        else:
+            self.sort_key = sort_key
+
+    @staticmethod
```
Nic-Ma, line 2680: It's a staticmethod but starts with "_", I feel confused it's for public use or not?

Thanks.

monai/transforms/spatial/array.py
```diff
@@ -2609,3 +2619,168 @@ def _get_params(
...
+            output = output[: self.fix_num_patches]
+            if len(output) < self.fix_num_patches:
+                if self.pad_opts.get("constant_values"):
+                    patch = np.ones((array.shape[0], *self.patch_size)) * self.pad_opts["constant_values"]
+                else:
+                    patch = np.zeros((array.shape[0], *self.patch_size))
+                location = np.zeros((3, len(self.patch_size)))
+                output += [(patch, location)] * (self.fix_num_patches - len(output))
+
+        if not isinstance(array, np.ndarray):
```
Nic-Ma, line 2714: No need to check whether it's `np.ndarray`, `convert_to_dst_type` will not convert if already the same type.

Thanks.

monai/transforms/spatial/array.py
```diff
@@ -2609,3 +2619,168 @@ def _get_params(
...
+            output = [convert_to_dst_type(src=patch, dst=array)[0] for patch in output]
+
+        self.num_patches = len(output)
```
Nic-Ma, line 2717: This line will make the transform thread-unsafe for `CacheDataset` and multi-thread modules, is it possible to return the num of patches directly?

Thanks.

monai/transforms/spatial/array.py
```diff
@@ -2609,3 +2619,168 @@ def _get_params(
...
+        return output
+
+
+class RandGridPatch(GridPatch, RandomizableTransform):
+    """
+    Return all (or a subset of) the patches sweeping the entire image with a random starting position.
+
+    Args:
+        patch_size: size of patches to generate slices for, 0 or None selects whole dimension
+        start_pos: starting position in the array, default is 0 for each dimension.
```
Nic-Ma, line 2727: doc-string doesn't match the real args.

Thanks.

monai/transforms/spatial/array.py
```diff
@@ -2609,3 +2619,168 @@ def _get_params(
...
+        self,
+        patch_size: Sequence[int],
+        min_start_pos: Optional[Union[Sequence[int], int]] = None,
+        max_start_pos: Optional[Union[Sequence[int], int]] = None,
+        fix_num_patches: Optional[int] = None,
+        overlap: Union[Sequence[float], float] = 0.0,
+        sort_key: Optional[Union[Callable, str]] = None,
+        pad_mode: Union[NumpyPadMode, str] = NumpyPadMode.CONSTANT,
+        pad_opts: Optional[Dict] = None,
+        seed: int = 0,
```
Nic-Ma, line 2758: I think we should not set the random seed within the transform.
CC @wyli .

Thanks.

monai/transforms/spatial/array.py
```diff
@@ -2609,3 +2619,168 @@ def _get_params(
...
+            min_start_pos = ensure_tuple_rep(self.min_start_pos, len(self.patch_size))
+        if self.max_start_pos is None:
+            max_start_pos = tuple(s % p for s, p in zip(array.shape[1:], self.patch_size))
+        else:
+            max_start_pos = ensure_tuple_rep(self.max_start_pos, len(self.patch_size))
+
+        self.start_pos = tuple(
+            self.R.randint(low=low, high=high + 1) for low, high in zip(min_start_pos, max_start_pos)
+        )
+        return super().__call__(array)
```
Nic-Ma, line 2786: I feel maybe it's better to decouple the GridPatch and RandGridPatch transforms.
Refer to: https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/croppad/array.py#L540.
What do you think?

Thanks.

monai/transforms/spatial/dictionary.py
```diff
@@ -2194,6 +2203,159 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> List[Dict[Hashab
...
+
+
+class RandGridPatchd(RandomizableTransform, MapTransform):
+    """
+    Return all (or a subset of) the patches sweeping the entire image with a random starting position.
+
+    Args:
+        keys: keys of the corresponding items to be transformed.
+        patch_size: size of patches to generate slices for, 0 or None selects whole dimension
+        start_pos: starting position in the array, default is 0 for each dimension.
```
Nic-Ma, line 2282: Wrong doc-string for args.

Thanks.

monai/transforms/spatial/array.py
drbeh: I followed this: https://github.com/Project-MONAI/MONAI/blob/eb60b9caf0a103af53ab5f3d620ceb2c116a4851/monai/apps/datasets.py#L81

monai/transforms/spatial/array.py
drbeh: Thanks, it's an important point. Let me check on that later.

monai/transforms/spatial/array.py
drbeh: May I ask what is the benefit of that? Anyways, `RandGridPatch` will depends on `GridPatch` but in this way at least we don't need to create an `GridPatch` object for each call of `RandGridPatch`, right?

monai/transforms/spatial/array.py
```diff
@@ -2609,3 +2619,168 @@ def _get_params(
...
+    backend = [TransformBackends.TORCH, TransformBackends.NUMPY]
+
+    def __init__(
+        self,
+        patch_size: Sequence[int],
+        start_pos: Sequence[int] = (),
+        fix_num_patches: Optional[int] = None,
+        overlap: Union[Sequence[float], float] = 0.0,
+        sort_key: Optional[Union[Callable, str]] = None,
+        pad_mode: Union[NumpyPadMode, str] = NumpyPadMode.CONSTANT,
```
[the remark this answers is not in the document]
drbeh: The issue is with `iter_patch` that uses np,pad. I'll update all of them with `Pad`.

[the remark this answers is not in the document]
drbeh: I added support for both numpy and torch pad mod.

monai/transforms/spatial/array.py
drbeh: I removed "_".

monai/transforms/spatial/array.py
Nic-Ma: Hi @drbeh ,

Here is the reference design for transforms:
https://github.com/Project-MONAI/MONAI/wiki/Developer-Guide-Transforms#randomized-transforms
We don't need to set `seed` within a transform, users can call `set_random_state(seed=XXX)` outside a random transform if they want deterministic behavior. Otherwise, it's non-deterministic behavior for every call.

Datasets don't have an internal random state, so we explicitly added the `seed` option to control the logic.

CC @ericspod @wyli for more discussion if necessary.

Thanks.

monai/transforms/spatial/array.py
Nic-Ma: I think it's not a strict requirement so far.
@ericspod @rijobro @wyli May I know you guys' opinions?
If you guys prefer to the `inherit` way for this transform, I am OK.

Thanks.

monai/transforms/spatial/array.py
drbeh: That makes sense! I removed seed.

monai/transforms/spatial/array.py
drbeh: I made it thread safe. thanks

monai/transforms/spatial/dictionary.py
```diff
@@ -2194,6 +2203,164 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> List[Dict[Hashab
...
+            sort_fn=sort_fn,
+            pad_mode=pad_mode,
+            pad_opts=pad_opts,
+        )
+
+    def set_random_state(
+        self, seed: Optional[int] = None, state: Optional[np.random.RandomState] = None
+    ) -> "RandGridPatchd":
+        self.patcher.set_random_state(seed, state)
+        super().set_random_state(seed, state)
```
Nic-Ma, line 2340: Could you please help adjust the code sequence to:
```py
super().set_random_state(seed, state)
self.patcher.set_random_state(seed, state)
```
No special reason, just to make the random behavior consistent with all the other transforms, like:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/intensity/dictionary.py#L188

Thanks.

monai/transforms/spatial/dictionary.py
```diff
@@ -2194,6 +2203,164 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> List[Dict[Hashab
...
+        return self
+
+    def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> List[Dict]:
+        d = dict(data)
+        original_spatial_shape = d[first(self.keys)].shape[1:]
+        output = []
+        results = [self.patcher(d[key]) for key in self.keys]
```
Nic-Ma, line 2347: Here the `self.patcher` will execute different random behavior for different keys.
If you want to ensure the consistent random behavior for keys, please refer to:
(1) Enable random operation for the dict transform:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/intensity/dictionary.py#L194
(2) Then disable random operation for the array transform with `randomize=False`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/intensity/dictionary.py#L205

Thanks.

Nic-Ma (commented):
Thanks for the quick update.
It overall looks good to me, just some minor comments inline and several questions to @ericspod @rijobro .

Thanks.

monai/transforms/spatial/array.py
```diff
@@ -2582,7 +2593,7 @@ def __call__(
             # Flatten the first two dimensions
             strided_image = strided_image.reshape(-1, *strided_image.shape[2:])
             # Make a list of contiguous patches
-            patches = [np.ascontiguousarray(p) for p in strided_image]
+            patches = [np.copy(p) for p in strided_image]
```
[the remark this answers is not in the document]
ericspod: If the array is already contiguous `ascontiguousarray` should do nothing and be more efficient, otherwise it'll make a copy. Copying always makes sure the patches are detached from `strided_image`, but since this is getting reshaped into a new array anyhow this isn't an issue so I'd revert this change.

monai/transforms/spatial/array.py
```diff
@@ -2609,3 +2620,167 @@ def _get_params(
...
+    ):
+        self.patch_size = ensure_tuple(patch_size)
+        self.start_pos = ensure_tuple(start_pos)
+        self.pad_mode: NumpyPadMode = convert_pad_mode(dst=np.zeros(1), mode=pad_mode)
+        self.pad_opts = {} if pad_opts is None else pad_opts
+        self.overlap = overlap
+        self.fix_num_patches = fix_num_patches
+        self.sort_fn: Optional[Callable]
+        if isinstance(sort_fn, str):
+            if sort_fn == "random":
```
ericspod, line 2670: With magic constants like "random" I'd suggest putting these into a Enum class somewhere so that it's encoded what the valid values are.

monai/transforms/spatial/array.py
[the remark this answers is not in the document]
drbeh: Thanks @Nic-Ma and @ericspod for your comments. Actually there is more context that need to taken into account for this change. Here we are reshaping the array using `as_strided` and with `writeable=False` (which is recommended to avoid accidental write operations), this will cause the numpy array even when converted to torch.Tensor not be writable, so subsequent transforms on this Tensor may have undefined behavior. `ascontiguousarray` won't change writability if the array is contiguous. So, I should either make sure that we always get a writable copy of the array, or I can set `writeable=True`. The latter is more efficient but can be problematic if there is an undiscovered bug. What do you think?

monai/transforms/spatial/array.py
```diff
@@ -2609,3 +2620,167 @@ def _get_params(
...
+        )
+        if self.sort_fn is not None:
+            output = sorted(patch_iterator, key=self.sort_fn)
+        else:
+            output = list(patch_iterator)
+        if self.fix_num_patches:
+            output = output[: self.fix_num_patches]
+            if len(output) < self.fix_num_patches:
+                if self.pad_opts.get("constant_values"):
+                    patch = np.ones((array.shape[0], *self.patch_size)) * self.pad_opts["constant_values"]
```
ericspod, line 2709: `np.full`?

monai/transforms/spatial/array.py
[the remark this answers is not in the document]
ericspod: Before this line you have `strided_image = strided_image.reshape(-1, *strided_image.shape[2:])` which I think will make `strided_image` a fresh array which is writable even if the shape doesn't change. Is that correct?

monai/transforms/spatial/array.py
```diff
@@ -2609,3 +2620,167 @@ def _get_params(
...
+
+        output = [convert_to_dst_type(src=patch, dst=array)[0] for patch in output]
+
+        return output
+
+
+class RandGridPatch(GridPatch, RandomizableTransform):
+    """
+    Extract all the patches sweeping the entire image in a row-major sliding-window manner with possible overlaps,
+    and with random offset for the starting position of upper left corner of the image.
```
ericspod, line 2723: By "upper left" corner you mean the minimal corner? To cover 2D and 3D volumes I suggest we refer to the minimal corner at index (0,0) or (0,0,0).

monai/transforms/spatial/array.py
```diff
@@ -2609,3 +2620,167 @@ def _get_params(
...
+            Defaults to image size modulo patch size.
+        fix_num_patches: number of patches to return. Defaults to None, which returns all the available patches.
+        overlap: the amount of overlap of neighboring patches in each dimension (a value between 0.0 and 1.0).
+            If only one float number is given, it will be applied to all dimensions. Defaults to 0.0.
+        sort_fn: a callable or string that defines the order of the patches to be returned. If it is a callable, it
+            will be passed directly to the `key` argument of `sorted` function. The string can be "min" or "max",
+            which are, respectively, the minimum and maximum of the sum of intensities of a patch across all dimensions
+            and channels. Also "random" creates a random order of patches.
+            By default no sorting is being done and patches are returned in a row-major order.
+        pad_mode: {``"constant"``, ``"edge"``, ``"linear_ramp"``, ``"maximum"``, ``"mean"``,
```
ericspod, line 2739: You could refer to NumpyPadMode and PytorchPadMode instead of listing the values here to save a little space if you wanted.

monai/transforms/spatial/array.py
```diff
@@ -2609,3 +2620,167 @@ def _get_params(
...
+            fix_num_patches=fix_num_patches,
+            overlap=overlap,
+            sort_fn=sort_fn,
+            pad_mode=pad_mode,
+            pad_opts=pad_opts,
+        )
+        self.min_start_pos = min_start_pos
+        self.max_start_pos = max_start_pos
+
+    def __call__(self, array: NdarrayOrTensor):
```
ericspod, line 2773: We should be calling `randomize` here I think like [here](https://github.com/Project-MONAI/MONAI/blob/61a0dc35b8a5edf78487b30c02c7aa0e51fc3409/monai/transforms/croppad/array.py#L535).

monai/transforms/spatial/array.py
ericspod: This inheritance pattern is the better way of doing things for now, the issue is that `RandGridPatch` is not substitutable with `GridPatch` which is perhaps a theoretical problem more than a practical one. I would rather avoid creating objects just to throw them out again (poltergeists).

In discussion with @rijobro and @atbenmurray we should really be going in the direction Pytorch does with it's `functional` submodules where the actual meat of computation is defined in functions. Classes then call these functions in their `__call__` or `forward` methods and so are often thin wrappers around the functions. In this instance the body of `GridPatch.__call__` would be put into a function with a large number of arguments in place of class members, and then it and `RandGridPatch.__call__` would call this function. The two classes wouldn't need an inheritance relationship because the reuse mechanism is moved to using functions. This is beyond the scope of this PR so I would suggest using inheritance here.

monai/transforms/spatial/array.py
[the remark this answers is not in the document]
ericspod: Ok so yes you're right the `copy` is needed.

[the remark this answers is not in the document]
Nic-Ma: Thanks for the discussion, much clearer now.

monai/transforms/spatial/array.py
Nic-Ma: OK, we can use `inheritance` for now, may revisit it in the future.

Thanks.

monai/transforms/spatial/dictionary.py
drbeh: Thanks, that's an important point. I solved it differently. I believe randomness should be controlled only with random state. There is no point for this transform with `randomize=False` but if we reset the random state in call for each of the keys, then we end up with the same randomization for all of them. Please take a look at the code and let me know what you think,=.

monai/transforms/spatial/dictionary.py
Nic-Ma: Sure, I see the change.

Thanks.

monai/transforms/spatial/dictionary.py
Nic-Ma: Hi @drbeh ,

Sorry I pasted a wrong link for your reference, this is a typical example for the **consistent random behavior of keys**:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/intensity/dictionary.py#L203-L205
@rijobro @ericspod What do you think?

Thanks.

drbeh:
Hi @Nic-Ma @ericspo, thank you very much for your comments. I have addressed them all. Please let me know if anything is missing.

monai/transforms/spatial/dictionary.py
drbeh: Thanks @Nic-Ma! I see but I believe in this case that randomization does not depend on the value of images, it's more straightforward to manage the randomization with setting `set_random_state` instead of having an additional flag to skip it. Does it make sense?

Nic-Ma: Thanks for the quick update, looks good to me.
Please feel free to raise a discussion with @ericspod @rijobro @wyli about your previous random implementation if necessary.

Thanks.

ericspod:
I'm good here!

PR #4331, opened 2022-05-24
4255 Update bundle download

### LINKED ISSUE #4255: Update Bundle download function
The bundle hosting place has been defined, thus the download function should be updated.

yiheng-wang-nv:

Fixes #4255 .

### Description
This PR is used to update the default repo of bundle download and load functions. In addition, it also enhances the load function.

### Status
**Ready**

### Types of changes


monai/bundle/scripts.py
```diff
@@ -293,7 +300,7 @@ def load(
     configer = ConfigComponent(config=net_kwargs)
     model = configer.instantiate()
     model.to(device)  # type: ignore
-    model.load_state_dict(model_dict)  # type: ignore
+    copy_model_state(dst=model, src=model_dict if key_in_ckpt is None else model_dict[key_in_ckpt])  # type: ignore
```
wyli, line 303: do you think this util should be used somewhere in this function? https://github.com/Project-MONAI/MONAI/blob/2bd56dc7b35422a60fbd39bed38e094c423a986b/monai/networks/utils.py#L383-L394

monai/bundle/scripts.py
```diff
@@ -247,15 +253,16 @@ def load(
         model_file: the relative path of the model weights or TorchScript module within bundle.
             If `None`, "models/model.pt" or "models/model.ts" will be used.
         load_ts_module: a flag to specify if loading the TorchScript module.
-        bundle_dir: the directory the weights/TorchScript module will be loaded from.
+        bundle_dir: directory the weights/TorchScript module will be loaded from.
             Default is `bundle` subfolder under`torch.hub get_dir()`.
-        source: the place that saved the bundle.
+        source: place that saved the bundle.
```
ericspod, line 258: It's not clear what's meant by this, may be "storage location name" and state that "github" is currently the only available value. This should be updated where this argument is used as well.

yiheng-wang-nv: Thanks Eric, I updated the docstrings.

monai/bundle/scripts.py
Nic-Ma: Hi @wyli ,

I think this PR refers to: https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/scripts.py#L592.
The `copy_model_state()` API leverages the `get_state_dict` as you mentioned.

Thanks.

monai/bundle/scripts.py
```diff
@@ -202,7 +200,13 @@ def download(
 
     _log_input_summary(tag="download", args=_args)
     name_, bundle_dir_, source_, repo_, url_, progress_ = _pop_args(
-        _args, name=None, bundle_dir=None, source="github", repo=None, url=None, progress=True
+        _args,
+        name=None,
+        bundle_dir=None,
+        source="github",
+        repo="Project-MONAI/model-zoo/hosting_storage_v1",
```
Nic-Ma, line 207: I think maybe no need to set default here as you already set in the `__init__()`?

Thanks.

monai/bundle/scripts.py
```diff
@@ -151,7 +151,7 @@ def download(
     name: Optional[str] = None,
     bundle_dir: Optional[PathLike] = None,
     source: str = "github",
-    repo: Optional[str] = None,
+    repo: str = "Project-MONAI/model-zoo/hosting_storage_v1",
```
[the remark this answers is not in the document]
yiheng-wang-nv: Thanks Nic, I changed the descriptions.

monai/bundle/scripts.py
wyli: sure, in that case calling `model_dict[key_in_ckpt]` here might be unstable, because it's assuming a dictionary?

yiheng-wang-nv: Thanks @wyli . As defined in https://docs.monai.io/en/latest/mb_specification.html `model.pt` is " the state dictionary of a saved model". Could we just add some more descriptions about it?

wyli: I think this is confusing, I'd suggest:

```py
model_dict = torch.load...
if not isinstance(model_dict, Mapping):
    warnings.warn(f"the state dictionary from {full_path} should be a dictionary but go {type(model_dict}.")
    model_dict = get_state_dict(model_dict)

```

wyli (approved):

monai/bundle/scripts.py
```diff
@@ -281,19 +281,22 @@ def load(
 
     if device is None:
         device = "cuda:0" if is_available() else "cpu"
-    # loading with `torch.jit.load`
+    # loading with `torch.jit. load`
```
Nic-Ma, line 284: May I know why added a blank here? Is it a typo?

Thanks.

yiheng-wang-nv: Thanks Nic, it seems the code format tool mis-modify this place, I just pushed a new commit.

Nic-Ma (approved):

PR #4332, opened 2022-05-24
Meta tensor spatial resample

rijobro:
### Description
`SpatialResample` for `MetaTensor`. Array transform includes `inverse`.

### Status
**Ready**

### Types of changes


rijobro:
@wyli thanks, merged.

tests/test_nifti_rw.py
```diff
@@ -30,14 +30,14 @@
...
         )
-        TESTS.append(
-            [
-                TEST_IMAGE,
-                TEST_AFFINE,
-                dict(reader="NibabelReader", image_only=False, as_closest_canonical=True),
-                np.arange(24).reshape((2, 4, 3)),
-            ]
-        )
+        # TESTS.append(
```
wyli, line 33: @rijobro I removed this test case, because when the affine suggests only a simple reorientation has been applied, and at the same time the user requests a new resampling output spatial shape, the inputs are ambiguous. perhaps we skip this test for now.

wyli (approved):
thanks, I went through the implementation and the unit tests work fine.

tests/test_nifti_rw.py
rijobro: Sounds good, thanks!

PR #4357, opened 2022-05-26
improve dtype conversion

rijobro:
### Description
Remove need for lookup table for dtype conversion. Create a zero-size array and convert to and from numpy/torch, as necessary. Also add functionality to create dtype from string, e.g., `"float32"` -> `np.float32`.

### Status
**Ready**

### Types of changes


monai/utils/type_conversion.py
```diff
@@ -31,36 +32,29 @@
...
-def dtype_numpy_to_torch(dtype):
+def dtype_numpy_to_torch(dtype: np.dtype) -> torch.dtype:
     """Convert a numpy dtype to its torch equivalent."""
-    # np dtypes can be given as np.float32 and np.dtype(np.float32) so unify them
-    dtype = np.dtype(dtype) if isinstance(dtype, (type, str)) else dtype
-    return look_up_option(dtype, _np_to_torch_dtype)
+    return torch.from_numpy(np.zeros([], dtype=dtype)).dtype
 
 
 def get_equivalent_dtype(dtype, data_type):
```
wyli, line 60: I was looking at this, shall we make the `dtype` of `get_equivalent_dtype` support string input such as `"torch.float32"`

rijobro: sure.

wyli: thanks! fyi... the string dtype is used here to store the type in applied operations (the np.dtype/torch.dtype object is not compatible with the collate function) and will need a revision after this PR https://github.com/Project-MONAI/MONAI/blob/fc5c34065326342ad91e08dea475614e16c42154/monai/transforms/spatial/array.py#L496

wyli (approved):
I think this type conversion is more robust than the current impl. cc@Nic-Ma

monai/utils/type_conversion.py
```diff
@@ -35,37 +36,32 @@
...
+    return torch.zeros([], dtype=dtype).numpy().dtype  # type: ignore
 
 
-def dtype_numpy_to_torch(dtype):
+def dtype_numpy_to_torch(dtype: np.dtype) -> torch.dtype:
     """Convert a numpy dtype to its torch equivalent."""
-    # np dtypes can be given as np.float32 and np.dtype(np.float32) so unify them
-    dtype = np.dtype(dtype) if isinstance(dtype, (type, str)) else dtype
-    return look_up_option(dtype, _np_to_torch_dtype)
+    return torch.from_numpy(np.zeros([], dtype=dtype)).dtype
```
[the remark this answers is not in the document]
wyli: might be useful to cache the results because often there are just a few combinations... https://docs.python.org/3/library/functools.html#functools.cache

PR #4377, opened 2022-05-28
Enhance expression logic of bundle config

Nic-Ma:
Fixes https://github.com/Project-MONAI/model-zoo/issues/25.

### Description
This PR enhanced the expression logic of bundle config to support reference in lambda or other callables.
And added more unit tests and integration test to cover it.

### Status
**Ready**

### Types of changes


Nic-Ma:
I will update the test for the pickle issue of lambda in windows.

Thanks.

tests/test_config_parser.py
```diff
@@ -211,6 +211,16 @@ def test_contains(self):
             self.assertTrue("entry" in parser)
             self.assertFalse("entr" in parser)
 
+    def test_lambda_reference(self):
+        configs = {
+            "patch_size": [8, 8],
+            "transform": {"_target_": "Lambda", "func": "$lambda x: x.reshape((1, *@patch_size))"},
```
wyli, line 217: we can avoid global variables by:
```suggestion
            "transform": {"_target_": "Lambda", "func": "$lambda x, s=@patch_size: x.reshape((1, *s))"},
```

perhaps this PR is not strictly required at the moment? cc @ericspod

Nic-Ma: Hi @wyli ,

Thanks for your suggestion. I tested it locally, it can work for not only python built-in types but also class instances.
It could workaround the issue, but I am not sure whether users will easily encounter the issue and repeatedly ask it?

@ericspod @rijobro @drbeh @atbenmurray , let me narrow down the issue and solutions:
Actually, this is not a bundle specific issue, the main question is how to make the below python program work:
```py
def test_func(fn):
    a = "hello"
    print(fn(a))


def main():
    var = "world"
    test_func(fn=eval("lambda x: x + var"))


if __name__ == "__main__":
    main()
```
```
  File "/workspace/PythonTests/main.py", line 12, in <module>
    main()
  File "/workspace/PythonTests/main.py", line 8, in main
    test_func(fn=eval("lambda x: x + var"))
  File "/workspace/PythonTests/main.py", line 3, in test_func
    print(fn(a))
  File "<string>", line 1, in <lambda>
NameError: name 'var' is not defined.
```
Currently, we have 2 solutions:
(1) In this PR's solution, it adds `var` to the `globals` arg of python `eval`, so change the `main()` to:
```py
def main():
    var = "world"
    test_func(fn=eval("lambda x: x + var", {"var": var}))
```
(2) Another solution is to change all this kind of lambda functions to be as below:
```py
def main():
    var = "world"
    test_func(fn=eval("lambda x, s=var: x + s"))
```
Which solution do you prefer to? Or do you have another other better solutions?

Thanks in advance.

ericspod: I think the problem with the initial code is that the meaning of `locals` isn't what you expect. In `eval` the `locals` dictionary provides local values for evaluating the expression, these are used to reduce the expression to a result. The result in the expression `lambda x: x + var` is not a reduction in any way so putting `var` into `locals` doesn't cause it to be used. What is done is the construction of a callable (the lambda) which has it's own local environment containing only `x` and not `var` because it does not include local variables from the environment it was created in, from its perspective `var` is a global variable that is to say it's an unbound variable. When the lambda is called and `x+var` is evaluated `x` is a local argument variable but `var` is expected here to be a global value and placed in the `globals` dictionary which the lambda will bind to on creation. Typically for using `eval` I think the expected practice is to provide the expression itself and wrap that in a lambda if needed:

```python
...
def main():
    var = "world"
    expression = "x + var"
    test_func(fn=lambda x: eval(expression,{},{"x":x, "var":"world"}))
...
```

This doesn't necessarily suit the way you're going about solve this issue however so I wouldn't suggest changing everything. Note also we can't do `lambda x: eval(expression,{},locals())` because `locals` in the scope of the lambda contains only `x` again so we're back at the original problem. If I called `locals()` before creating the lambda and saved it to a variable `_locals` and passed that in it would fail because this wouldn't contain `x`. Scoping in Python gets weird sometimes.

The workaround I've used before is to add to `globals` and pass that as in your current solution. Your other alternative seems more complex, and other things like wrapping lambdas in lambdas to somehow create the local environment we want makes things messier. Changing what the global scope is will probably be safe if people pay attention to good variable names, re-using a module name as a variable name for example could work in normal function code which would look the variable up correctly in `locals` before `globals` but is an obvious code smell.

Nic-Ma: @ericspod Thanks for your comments.
Seems you agree with the `globals` solution "The workaround I've used before is to add to globals and pass that as in your current solution." Do you have any other comments about the PR?
@atbenmurray @drbeh Would be helpful if you guys also share some thoughts.

Thanks in advance.

ericspod (approved):
This change may not be needed if we use other techniques to avoid having global variables but it's much cleaner to do it this way. It's correct that an unbound value in a lambda is a global variable, the only concern is if changing the global dictionary would override an existing value, warning when this happens might be a good addition.

tests/test_config_parser.py
Nic-Ma: Thanks @drbeh for your input.
@wyli I will update the PR to add a warning when globals conflict according to @ericspod 's suggestion.
What do you think?

Thanks.

wyli: sure, I agree with this PR as long as there's no significant side-effect of memory/

Nic-Ma: Added warning and new unit test to cover it.

Thanks.

PR #4410, opened 2022-05-31
Masked Patch WSI Dataset

### LINKED ISSUE #4014: Refactor MaskedInferenceWSIDataset
Refactor and enhance `MaskedInferenceWSIDataset` in pathology, for blending it into core MONAI as laid out in https://github.com/Project-MONAI/MONAI/issues/4005.

https://github.com/Project-MONAI/MONAI/blob/a676e3876903e5799eb3d24245872ea4f3a5b152/monai/apps/pathology/data/datasets.py#L175-L319

### LINKED ISSUE #4398: WSIReader get_patch => ChannelFirst behvaiour should be controlled through argument
https://github.com/Project-MONAI/MONAI/blob/61a0dc35b8a5edf78487b30c02c7aa0e51fc3409/monai/data/wsi_reader.py#L459

We can't force it to convert it to ChannelFirst
This behavior should be configurable.  Can't force the user to use ChannelFirst patch (specially in case of png..)

LoadImaged => MyTransform => EnsureChannelFirst
LoadWSI => MyTransform => EnsureChannelFirst

Little not consistent with LoadImage and LoadWSI ..  in the 2nd case, I can't use my transform same as first case.. so I have to do something like this as a workaround
LoadWSI => AsChannelLast => MyTransform => EnsureChannelFirst

better to be consistent with what LoadImage does..

bhashemian:
Fixes #4014 
Fixes #4398

### Description
This PR implements `MaskedPatchWSIDataset`, which uses a whole slide image at a lower resolution to create a mask and extract patches based on that.
It also updates `WSIReader` to be able to set `channel_last`, and some other updates to make sure all components are properly working together.

### Status
**Ready**

### Types of changes

### DIFF
--- a/docs/source/apps.rst
+++ b/docs/source/apps.rst
@@ -160,12 +160,15 @@ Applications
 .. automodule:: monai.apps.detection.utils.box_coder
     :members:
 
-`Detector utils`
-~~~~~~~~~~~~~~~~
+`Detection Utilities`
+~~~~~~~~~~~~~~~~~~~~~
 .. automodule:: monai.apps.detection.utils.detector_utils
     :members:
 
-`Detector box selector`
-~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: monai.apps.detection.utils.predict_utils
+    :members:
+
+`Inference box selector`
+~~~~~~~~~~~~~~~~~~~~~~~~
 .. automodule:: monai.apps.detection.utils.box_selector
     :members:
--- a/monai/apps/detection/networks/retinanet_detector.py
+++ b/monai/apps/detection/networks/retinanet_detector.py
@@ -1,0 +1,624 @@
+# Copyright (c) MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# =========================================================================
+# Adapted from https://github.com/pytorch/vision/blob/main/torchvision/models/detection/retinanet.py
+# which has the following license...
+# https://github.com/pytorch/vision/blob/main/LICENSE
+
+# BSD 3-Clause License
+
+# Copyright (c) Soumith Chintala 2016,
+# All rights reserved.
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+
+# * Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+
+# * Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+
+# * Neither the name of the copyright holder nor the names of its
+#   contributors may be used to endorse or promote products derived from
+#   this software without specific prior written permission.
+
+"""
+Part of this script is adapted from
+https://github.com/pytorch/vision/blob/main/torchvision/models/detection/retinanet.py
+"""
+
+import warnings
+from typing import Any, Callable, Dict, List, Sequence, Tuple, Union
+
+import torch
+from torch import Tensor, nn
+
+from monai.apps.detection.networks.retinanet_network import RetinaNet, resnet_fpn_feature_extractor
+from monai.apps.detection.utils.anchor_utils import AnchorGenerator
+from monai.apps.detection.utils.box_coder import BoxCoder
+from monai.apps.detection.utils.box_selector import BoxSelector
+from monai.apps.detection.utils.detector_utils import check_training_targets, preprocess_images
+from monai.apps.detection.utils.predict_utils import ensure_dict_value_to_list_, predict_with_inferer
+from monai.data.box_utils import box_iou
+from monai.inferers import SlidingWindowInferer
+from monai.networks.nets import resnet
+from monai.utils import BlendMode, PytorchPadMode, ensure_tuple_rep
+
+
+class RetinaNetDetector(nn.Module):
+    """
+    Retinanet detector, expandable to other one stage anchor based box detectors in the future.
+    An example of construction can found in the source code of
+     :func:`~monai.apps.detection.networks.retinanet_detector.retinanet_resnet50_fpn_detector` .
+
+    The input to the model is expected to be a list of tensors, each of shape (C, H, W) or  (C, H, W, D),
+    one for each image, and should be in 0-1 range. Different images can have different sizes.
+    Or it can also be a Tensor sized (B, C, H, W) or  (B, C, H, W, D). In this case, all images have same size.
+
+    The behavior of the model changes depending if it is in training or evaluation mode.
+
+    During training, the model expects both the input tensors, as well as a targets (list of dictionary),
+    containing:
+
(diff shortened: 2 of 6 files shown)


drbeh:
\black

drbeh:
@Nic-Ma Please merge https://github.com/Project-MONAI/MONAI/pull/4382 first. Thaks

Nic-Ma:
Hi @drbeh ,

I see the PR status is "draft" and "hold", but you added me to review, is it ready for review?

Thanks.

drbeh:
Hi @Nic-Ma,
Sorry for the confusion. It was ready for review but since the rest of PR's are inter-dependent, I decided to include the other PR that I'm working on to this one to avoid wait for one to be merged for testing and updating.
I'll let you know when this PR is ready to review. 
Thanks

drbeh:
@Nic-Ma it is ready for review. Do you have any idea what's the issue with flake8 tests? It seems that they are failing for an unrelated issue to this PR, doesn't it?

monai/data/utils.py
```diff
@@ -285,19 +285,28 @@ def iter_patch(
     patch_size_ = get_valid_patch_size(arr.shape, patch_size)
     start_pos = ensure_tuple_size(start_pos, arr.ndim)
 
+    # set padded flag to false if pad mode is None
+    padded = True if mode else False
```
Nic-Ma, line 289: Could you please help add some description in the doc-string of `model` arg for the `model == None` case?

Thanks.

monai/data/wsi_reader.py
```diff
@@ -56,9 +56,11 @@ class BaseWSIReader(ImageReader):
     supported_suffixes: List[str] = []
     backend = ""
 
-    def __init__(self, level: int, **kwargs):
+    def __init__(self, level: int, channel_last: bool = False, **kwargs):
```
Nic-Ma, line 59: I would suggest to use an arg `channel_dim: int = 0` directly instead of `channel_last`.
And add the doc-string for these 2 args.

Thanks.

monai/data/wsi_reader.py
```diff
@@ -56,9 +56,11 @@ class BaseWSIReader(ImageReader):
...
         super().__init__()
         self.level = level
+        self.channel_last = channel_last
+        self.channel_dim = -1 if self.channel_last else 0
```
Nic-Ma, line 63: And here we verify the supported `channel_dim` values so far (like 0, -1, 2, -3).
What do you think?

Thanks.

monai/transforms/spatial/array.py
```diff
@@ -2647,42 +2649,41 @@ class GridPatch(Transform):
...
         pad_mode: Union[NumpyPadMode, PytorchPadMode, str] = NumpyPadMode.CONSTANT,
         **pad_kwargs,
     ):
         self.patch_size = ensure_tuple(patch_size)
-        self.offset = ensure_tuple(offset)
-        self.pad_mode: NumpyPadMode = convert_pad_mode(dst=np.zeros(1), mode=pad_mode)
+        if offset:
+            self.offset = ensure_tuple(offset)
+        else:
+            self.offset = (0,) * len(self.patch_size)
```
Nic-Ma, line 2664: I would suggest to simplify this code to:
```py
self.offset = ensure_tuple(offset) if offset else (0,) * len(self.patch_size)
```

Thanks.

monai/transforms/spatial/array.py
```diff
@@ -2637,7 +2637,9 @@ class GridPatch(Transform):
             which are, respectively, the minimum and maximum of the sum of intensities of a patch across all dimensions
             and channels. Also "random" creates a random order of patches.
             By default no sorting is being done and patches are returned in a row-major order.
-        pad_mode: refer to  NumpyPadMode and PytorchPadMode. Defaults to ``"constant"``.
+        threshold: a value to keep only the patches whose sum of intensities are less than the threshold.
```
Nic-Ma, line 2640: Maybe we can change this `threshold` arg to `filter_fn` directly? And default to `None`?
It's more flexible and the code logic can be easier, refer to:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/croppad/array.py#L691

What do you think?

Thanks.

monai/transforms/spatial/dictionary.py
```diff
@@ -2368,7 +2376,7 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> List[Dict]:
                 new_dict[k] = deepcopy(d[k])
             # fill additional metadata
             new_dict["original_spatial_shape"] = original_spatial_shape
-            new_dict["slices"] = patch[0][1]  # use the coordinate of the first item
+            new_dict["patch_location"] = patch[0][1]  # use the starting coordinate of the first item
```
Nic-Ma, line 2379: I feel maybe we should also define all these keys of `new_dict` in some Enum to make them easier to use and maintain? I see you already defined something similar, like `WSIPatchKeys`.
CC @ericspod 

Thanks.

monai/utils/enums.py
```diff
@@ -332,23 +334,55 @@ class BoxModeName(Enum):
     CCCWHD = "cccwhd"  # [xcenter, ycenter, zcenter, xsize, ysize, zsize]
 
 
-class ProbMapKeys(Enum):
+class ProbMapKeys(str, Enum):
```
Nic-Ma, line 337: May I know why inherit from `str`?

Thank.

monai/transforms/spatial/array.py
drbeh: This is how I implemented in previous commits but the problem is that here we need to pass the threshold to the function, which will cause problem in distributed trainings since it need to be pickled and it cannot be a lambda function or a function defined locally in the worker. Then it will create a hassle for the user on where and how to create such filter function. So it would better to make it simple and use threshold for its most common use case.

monai/transforms/spatial/dictionary.py
drbeh: Sure, I already have them.

monai/data/utils.py
drbeh: Added!

monai/transforms/spatial/dictionary.py
drbeh: I updated it with `WSIPatchKeys`.

monai/transforms/spatial/array.py
Nic-Ma: Hi @drbeh ,

I think it's not a problem to pass a function as the arg in multi-gpu program, many MONAI APIs do that, and I remember it only affects the `spawn` multi-processing?
To support pickle, you can just define the function and pass it as the arg.
The benefit of using `filter_fn` is that it can simplify the code logic and support more kinds of filters in the future.
Please feel free to correct me if I misunderstood the logic.
@ericspod @wyli @rijobro What do you think?

Thanks.

drbeh: @Nic-Ma I know that, and if you check the previous commits of this PR, you can see the exact implementation that you say. But consider the MIL pipeline that we have: https://github.com/Project-MONAI/tutorials/blob/main/pathology/multiple_instance_learning/panda_mil_train_evaluate_pytorch_gpu.py

It uses `spawn` and the way that the arguments are passed, you cannot create such filtering function based on the input args. I know that we can change the pipeline but the point is users need to know too much for that too work. It's a simple and common functionality to threshold and I don't think that we should make it more difficult for the user to use it.

ericspod: One of the issues here is that arguments get pickled when passed to a spawned process and a pickle of a function only contains a reference to it's name and not the bytecode of the function itself. If a function defined with a name that can be imported is passed to a spawned process that process will import the function and get the reference to it, but a function defined locally (eg. in a notebook) cannot be imported and will cause an error. So this sometimes works and sometimes doesn't for non-obvious reasons, so it's probably best to use the threshold value instead of a callable. We've seen this before under Windows with some transforms working inside dataloaders with multiple workers sometimes and sometimes not.

Nic-Ma: OK, let's keep this implementation for now, thanks for the discussion.

Nic-Ma (approved):
Thanks for the quick update.
Looks good to me now.

Issue #233, opened 2020-03-31
Set up an explicit backward compatibility policy


tvercaut:
A discussed in #197, we agreed on following [semver](https://semver.org/) for version numbering. This implicetly means we agree on maintaining backward compatibility until major version changes.

It might be good to make the backward compatibility policy more explicit. By means of example, TensorFlow documents their backwards compatibility policy quite well: https://www.tensorflow.org/guide/versions

For MONAI, enforcing bacward compatibility acording to semver might also mean introducing an "experimental" folder or similar within MONAI where features that are ready for real-world testing but not yet for backward compatibility can be placed.

tvercaut:
A bit of a side question in this issue but note that semver and PEP440 are incompatible when it comes to pre-release numbering (see also https://github.com/semver/semver/issues/483). Given the pythonic ambition of MONAI, having PEP440 prevail over semver for this particular point would probably make sense.

IsaacYangSLA:
Proposed statement on backward compatibility policy:

MONAI versioning policy follows PEP440 as follows
```
X.YaN   # Alpha release
X.YbN   # Beta release
X.YrcN  # Release Candidate
X.Y     # Final release
```
The common understanding on backward compatibility applies.

- Backward compatibility across MAJOR (X in the example above) is not guaranteed.
- Backward compatibility in private API across MINOR (Y in the example above) is not guaranteed.

For releases with MAJOR == 0, public API can change at any time.

tvercaut:
Having a third component would make it closer aligned with semver and should remain compatible with PEP440. I would suggest following the  modified semver called "Linux/Python Compatible Semantic Versioning 3.0.0" from openstack:
https://docs.openstack.org/pbr/3.1.1/semver.html

Releases would be tagged with
```
X.Y.Z.aN             # Alpha release
X.Y.Z.aN.devM   # Dev version for an alpha release
X.Y.Z.bN             # Beta release
X.Y.Z.rcN            # Release Candidate
X.Y.Z                   # Final release
X.Y.Z.devM         # Dev version for a final release
```

tvercaut:
Related discussion in #2158

rijobro:
Also related PR: https://github.com/Project-MONAI/MONAI/pull/2171.

ericspod:
Another PR: #2338

wyli:
Here are some discussion points for the potential feature classifications:

Stable
Feature with code, documentation, tutorial ready, available in the milestone releases
[...]

PR #4415, opened 2022-05-31
Deprecate Old Pathology Components

### LINKED ISSUE #4226: Deprecating Old Pathology Components
Following our discussion in #4216 about wether we should deprecate the old pathology components (including the components in `monai/app/pathology` and elsewhere), or we should simply remove them and mention it in the release note.

My suggestion would be to deprecate since we have users that are continuously building pipelines based on our pathology components but I see Nic's point (https://github.com/Project-MONAI/MONAI/pull/4216#issuecomment-1116222932) about making a breaking change too. Any Thought?

@Nic-Ma @wyli @ericspod

drbeh:
Fixes #4226 

This PR deprecates datasets and append the previous test cases with `_deprecated`.

### Status
**Ready**

### Types of changes


tests/test_patch_wsi_dataset_deprecated.py
```diff
@@ -0,0 +1,256 @@
+# Copyright (c) MONAI Consortium
```
Nic-Ma, line 1: I think this file should be a "rename" instead of a totally new file?
Maybe you can try with `git mv` ?

Thanks.

Nic-Ma (approved):
Looks good to me, just some comments about the test files.
@ericspod Should we maintain the tests for new and deprecated APIs in this solution or any other better idea?

Thanks in advance.

ericspod:
So long as we're supporting deprecated code we should keep the tests for it.

Nic-Ma:
Hi @drbeh ,

I think @ericspod and I only have a comment left now:
Please use `git mv` to rename the files to avoid so much diff.
1. Rename `test_patch_wsi_dataset.py` to `test_patch_wsi_dataset_deprecated.py`.
2. Rename `test_patch_wsi_dataset_new.py` to `test_patch_wsi_dataset.py`.

Thanks.

tests/test_patch_wsi_dataset_deprecated.py
drbeh: @Nic-Ma I have already used `git mv` everywhere but I don't know why it is considering new file for this file but not for the other!
Anyways, even if someone use the regular `mv`, once it is staged, git automatically detects that it's a rename! I don't know why it's not happening here.

Nic-Ma: Hi @ericspod @wyli ,

Do you know any possible reason for this problem?

Thanks in advance.

ericspod:
Changing the names looks fine to me. If we can fix the renaming issue that would be good but I'm not worried if not. Maybe the issue is trying to rename an existing file to a new name then renaming another existing file to another existing name in one commit. What you might have to do is keep a copy of the existing files you're moving around, revert changes so that the files marked `deprecated` are back to their original names, change the name to add `_deprecated` again, commit, then move the `_new` files to not have the prefix.

ericspod:
I'm fine merging this one if there's no conflicts, the file names look right in the changes.

Nic-Ma (approved):
Looks good to me.

Hi @KumoLiu ,

@wyli Do you have any other comments?

Thanks in advance.

KumoLiu:
looks good to me!

wyli:
with the renaming approach, we lose track of the corresponding code change history within the files, and it's not easy to understand the diff contents. could we keep the original names and only add a comment in the test files to describe which one is newer? alternatively, we can remove the deprecated components as we are releasing v1.0. I think we should try to avoid situations that having many lines of changes but effectively little work being done.

drbeh:
Hi, I think we have spent too much time on this PR. It's been two months since this PR was created and we have already discussed many times about it. I believe we should move on and focus on more important things. Thanks.

Nic-Ma (approved):

wyli (dismissed):
I think it's the first time we have a rearrangement/renaming of some released components it's worth taking some time/effort to make the process good, and document it in the contributing guide. at the moment most of us have concern that in the current approach we lose track of the line-by-line corresponding unit test changes, and this approach also introduces some unnecessary code changes.

Nic-Ma:
Hi @drbeh ,

As discussed in https://github.com/Project-MONAI/MONAI/issues/233#issuecomment-1200887224, I think maybe we can:
1. Combine the tests for the latest module and the deprecated module in only one python file `test_wsireader.py`, same for `test_patch_wsi_dataset`.
2. We will drop the tests for deprecated module in the future version and give error message instead of warning message.

The benefit of this method is to keep tracking the change of tests in one python code file and easy to check the history.
What do you think?

Thanks.

Nic-Ma:
Hi @ericspod @rijobro ,

What do you guys think about the idea?
Thanks in advance.

rijobro:
I think this should be a note to ourselves not to call things `_new`, as they only stay new for a short while!

ericspod:
Definitely should avoid using "new" or "old", when a feature is removed should be mandatory and we remove the feature and its tests at the stated version. Perhaps the tests to be removed should be marked as deprecated as well.

Nic-Ma:
> Definitely should avoid using "new" or "old", when a feature is removed should be mandatory and we remove the feature and its tests at the stated version. Perhaps the tests to be removed should be marked as deprecated as well.

Thanks @ericspod @rijobro ,

That's a good point!

@drbeh What do you think? Could you please help merge the test cases and also mark the tests for deprecated APIs as "deprecated"?

Thanks in advance.

drbeh:
That is fine with me and I agree that we should not use words `new` and `old`. However, usually we don't encounter situations to have components with the same name but the problem arises because we called the test file of components in the apps folder similar to the ones in the core. It could have easily avoided by calling the test file for `monai.apps.xxx.yyy`, for instance, as `test_xxx_yyy.py` and reserve `test_yyy.py` for components in the core.

What do you think? @Nic-Ma @ericspod @rijobro @wyli

Nic-Ma:
Hi @drbeh ,

As I described in the previous comment, I think one of the main benefits to merge the test cases is that: it can help keep tracking the change of tests in one python test code file and easy to check the history.
What do you think?

Thanks.

drbeh:
> Hi @drbeh ,
> 
> As I described in the previous comment, I think one of the main benefits to merge the test cases is that: it can help keep tracking the change of tests in one python test code file and easy to check the history. What do you think?
> 
> Thanks.

Hi @Nic-Ma, maybe I am missing something but in this way we will lose the track of the tests for the newer component, don't we?

Anyways, if it makes this PR acceptable to merge, I'll do it because I don't think that it is that important. Just to clarify, when you say to mark it as deprecated, do you mean adding deprecated to the test name or use deprecated decorator?

Nic-Ma:
Hi @drbeh ,

Sorry maybe I didn't describe it clearly.
I think @ericspod @rijobro and I were talking about this idea:
1. Write all the **tests** for both the deprecated component and the new component in the same python file.
2. Mark the tests for the deprecated component with `deprecated` decorator (maybe defining the test program for the deprecated component in a unique class with a decorator).

What do you think?
@ericspod @rijobro Please feel free to correct me if I misunderstand anything.

Thanks in advance.

drbeh:
Hi, I have merged the unittest files according to what we have discussed. Please let me know if you have any other comment.

wyli (approved):
Thanks, it looks good to me.

Nic-Ma (approved):
Thanks for the quick update!
Looks good to me.

PR #4479, opened 2022-06-10
4477 4478 iterables enhancements

### LINKED ISSUE #4477: patch dataset `transform` should be applied to the patches only
**Describe the bug**
```py
import monai
from monai.transforms import DataStatsd
import numpy as np

img = [{"img": np.zeros((1, 30, 40, 3))}]
patch_func = monai.data.PatchIterd(keys=["img"], patch_size=(30, 40, 1))
transform = DataStatsd(keys=["img"])

for x in monai.data.GridPatchDataset(img, patch_iter=patch_func, transform=transform):
    print(x[0]["img"].shape)

```
should generate slices in shape `(30, 40, 1)`, 
but the `transform` is mistakenly applied in the base class:
```
Data statistics:
Type: <class 'numpy.ndarray'> float64
Shape: (1, 30, 40, 3)
Value range: (0.0, 0.0)
Data statistics:
Type: <class 'numpy.ndarray'> float64
Shape: (1, 30, 40, 1)
Value range: (0.0, 0.0)
(1, 30, 40, 1)
Data statistics:
Type: <class 'numpy.ndarray'> float64
Shape: (1, 30, 40, 1)
Value range: (0.0, 0.0)
(1, 30, 40, 1)
Data statistics:
Type: <class 'numpy.ndarray'> float64
Shape: (1, 30, 40, 1)
Value range: (0.0, 0.0)
(1, 30, 40, 1)
```

**Expected behavior**
the `super().__iter__()` shouldn't trigger the patch transform
https://github.com/Project-MONAI/MONAI/blob/750452652a12b28eb50c92c956323214de358809/monai/data/grid_dataset.py#L194

### LINKED ISSUE #4478: shufferbuffer is not compatible with the dataloader
**Describe the bug**
```py
import monai

buffer = monai.data.ShuffleBuffer([1,2,3,4])
for x in monai.data.DataLoader(buffer, batch_size=2):
    print(x)
```

```
    for x in monai.data.DataLoader(buffer, batch_size=2):
  File "/Users/Documents/MONAI/monai/data/dataloader.py", line 78, in __init__
    set_rnd(dataset, int(_seed))
  File "/Users/Documents/MONAI/monai/data/utils.py", line 624, in set_rnd
    obj.set_random_state(seed=seed % MAX_SEED)
  File "/Users/Documents/MONAI/monai/data/iterable_dataset.py", line 119, in set_random_state
    raise NotImplementedError(f"`set_random_state` is not available in {self.__class__.__name__}.")
NotImplementedError: `set_random_state` is not available in ShuffleBuffer.
```

wyli:

Fixes #4477 
Fixes #4478

### Description
- skip `ShuffleBuffer.set_random_state` method (the seed controls the state)
- `GridPatchDataset`'s self.transform is left empty, because it's shared with the base class.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/grid_dataset.py
+++ b/monai/data/grid_dataset.py
@@ -187,7 +187,7 @@ class GridPatchDataset(IterableDataset):
     ) -> None:
         super().__init__(data=data, transform=None)
         self.patch_iter = patch_iter
-        self.transform = transform
+        self.patch_transform = transform
         self.with_coordinates = with_coordinates
 
     def __iter__(self):
@@ -195,13 +195,17 @@ class GridPatchDataset(IterableDataset):
             if not self.with_coordinates:
                 for patch, *_ in self.patch_iter(image):  # patch_iter to yield at least 1 item: patch
                     out_patch = (
-                        patch if self.transform is None else apply_transform(self.transform, patch, map_items=False)
+                        patch
+                        if self.patch_transform is None
+                        else apply_transform(self.patch_transform, patch, map_items=False)
                     )
                     yield out_patch
             else:
                 for patch, slices, *_ in self.patch_iter(image):  # patch_iter to yield at least 2 items: patch, coords
                     out_patch = (
-                        patch if self.transform is None else apply_transform(self.transform, patch, map_items=False)
+                        patch
+                        if self.patch_transform is None
+                        else apply_transform(self.patch_transform, patch, map_items=False)
                     )
                     yield out_patch, slices
 
--- a/monai/data/grid_dataset.py
+++ b/monai/data/grid_dataset.py
@@ -192,22 +192,14 @@ class GridPatchDataset(IterableDataset):
 
     def __iter__(self):
         for image in super().__iter__():
-            if not self.with_coordinates:
-                for patch, *_ in self.patch_iter(image):  # patch_iter to yield at least 1 item: patch
-                    out_patch = (
-                        patch
-                        if self.patch_transform is None
-                        else apply_transform(self.patch_transform, patch, map_items=False)
-                    )
+            for patch, *others in self.patch_iter(image):
+                out_patch = patch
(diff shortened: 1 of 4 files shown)


monai/data/iterable_dataset.py
```diff
@@ -116,7 +116,7 @@ def randomize(self, size: int) -> None:
         self._idx = self.R.randint(size)
 
     def set_random_state(self, seed: Optional[int] = None, state: Optional[np.random.RandomState] = None):
-        raise NotImplementedError(f"`set_random_state` is not available in {self.__class__.__name__}.")
```
ericspod, line 119: Should just remove this method altogether since it doesn't do anything to the inherited one.

wyli: that's right, let me update this..

monai/data/grid_dataset.py
```diff
@@ -187,21 +187,25 @@ def __init__(
     ) -> None:
         super().__init__(data=data, transform=None)
         self.patch_iter = patch_iter
-        self.transform = transform
+        self.patch_transform = transform
         self.with_coordinates = with_coordinates
 
     def __iter__(self):
```
ericspod, line 193: I'd suggest reducing the whole method a bit more:

```python
def __iter__(self):
    for image in super().__iter__():
        for patch, *others in self.patch_iter(image):  # patch_iter to yield at least 1 item: patch
            out_patch = patch
            if self.patch_transform is not None:
                out_patch = apply_transform(self.patch_transform, patch, map_items=False)

            if self.with_coordinates and len(others) > 0:
                yield out_patch, others[0]
            else:
                yield out_patch
```

wyli: makes sense, thanks!

ericspod (approved):

Issue #4462, opened 2022-06-06
MetaTensor: re-enable full tests


wyli:
**Is your feature request related to a problem? Please describe.**
Follow-up of prototype https://github.com/Project-MONAI/MONAI/pull/4371

This ticket is to track the necessary breaking changes on branch [feature/MetaTensor](https://github.com/Project-MONAI/MONAI/tree/feature/MetaTensor) when rolling out the MetaTensor support for all the `monai.transforms.croppad`. 

These tests/checks are temporarily muted and should be re-enabled before merging the new features  to the `dev` branch:

- [x] monai/apps/detection/transforms/dictionary.py (type checks) in `ZoomBoxd` 
- [x] monai/metrics/utils.py (type checks) in `get_mask_edges`
- [x] monai/transforms/spatial/array.py (type checks) in `Zoom`
- [x] monai/transforms/spatial/dictionary.py (type checks) in `Zoomd`
- [x] tests/test_box_transform.py(https://github.com/Project-MONAI/MONAI/pull/4729)
- [x] tests/test_crop_base.py (`test_slice_calculation`)
- [x] tests/test_inverse.py (https://github.com/Project-MONAI/MONAI/commit/fb054359c852b6d9d62356077a2d99748c6137c1)
- [x] tests/test_invertd.py
- [x] tests/test_testtimeaugmentation.py
- [x] tests/test_integration_fast_train.py (addressed https://github.com/Project-MONAI/MONAI/pull/4506)
- [x] tests/test_integration_workflows.py ([the connected component not yet support metatensor](https://github.com/Project-MONAI/MONAI/blob/36dc12620f70ee19d16fd0b99a349df383def7ce/tests/test_integration_workflows.py#L246-L248))

cc @rijobro 

(latest successful build is at 60a22ffa467f280c99d3b4a3e6d75617c77aa98c)

- [x] tests/test_dataset_summary.py and dataset summary with metatensor (https://github.com/Project-MONAI/MONAI/pull/4728)

wyli:
`tests/test_integration_fast_train.py` is not currently leveraging MetaTensor, an `Ensuretyped` was included to convert the metatensors into regular tensors:
https://github.com/Project-MONAI/MONAI/blob/3583ea0677652bd8161626e57dcb3d3fa11d2068/tests/test_integration_fast_train.py#L97

The primary issue was in network forward performance, I looked into a smaller example to replicate it:

```py
[...]

PR #4506, opened 2022-06-15
enhance copy in metatensor

wyli:

addresses https://github.com/Project-MONAI/MONAI/issues/4462#issuecomment-1155128136

### Description
- the `_copy_meta` is currently making `outputs = model(image)` slow because of the meta attributes deepcop, this PR updates it to use a shallow copy, now the fast training pipeline can achieve the expected speed
- the PR makes `get_default_meta` static to have some minor speed up.

### Status

### Types of changes

### DIFF
--- a/monai/data/meta_obj.py
+++ b/monai/data/meta_obj.py
@@ -80,7 +80,7 @@ class MetaObj:
         self._is_batch: bool = False
 
     @staticmethod
-    def flatten_meta_objs(args: Iterable):
+    def flatten_meta_objs(*args: Iterable):
         """
         Recursively flatten input and yield all instances of `MetaObj`.
         This means that for both `torch.add(a, b)`, `torch.stack([a, b])` (and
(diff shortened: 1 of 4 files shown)


rijobro:
Hm, although disabling deep copying will make things faster, we obviously need to be careful of accidentally modifying an image's meta data when modifying another image that is a shallow copy. I suppose `id(self) != id_in` is limited in that with `a = b`, both `a` and `b` have different ID's, but they are shallow copies of one another and so here a shallow copy would suffice.

I'm not sure what the easy answer here is, we need a cleverer mechanism for figuring out when to copy and when not.

rijobro:
I don't think this is necessarily a good idea, but we could have a context manager:

```python
with MetaTensor.deep_copy_meta(False):
    out = model(image
```

wyli:
> I don't think this is necessarily a good idea, but we could have a context manager:
> 
> ```python
> with MetaTensor.deep_copy_meta(False):
>     out = model(image
> ```

I'm trying to set shallow copy when `is_batch` is True, essentially when it's a batch, it's within a network forward and the meta info are not updated within it.

PR #4544, opened 2022-06-20
Adding transforms for signal (ECG, EEG for instance)

doursand:
Fixes # .

### Description
This changes introduces the following transforms for 1d signal (example ECG , EEC, etc)

  "SignalRandDrop",
  "SignalRandScale",
  "SignalRandShift",
  "SignalResample",
  "SignalRandAddSine",
  "SignalRandAddSquarePulse",
  "SignalRandAddGaussianNoise",
  "SignalRandAddSinePartial",
  "SignalRandAddSquarePulsePartial",
  "SignalNormalize",
  "SignalStandardize",
  "SignalZeroPad",
  "SignalFillEmpty",
  "SignalRemoveFrequency",
  "SignalShortTimeFourier",
  "SignalContinousWavelet",

### Status
**Work in progress**

### Types of changes


doursand:
@wyli 

I made some corrections and went thru a new run for the checks, however it is now failing on a file which is not part of my PR

notation for "outputs"? Suggestion: "List[Union[ndarray[Any, Any], Tensor]]"
monai/networks/blocks/feature_pyramid_network.py:196:42: error: Argument 1 to "kaiming_uniform_" has incompatible type "Union[Tensor, Module]"; expected "Tensor"  [arg-type]

what is the strategy to remediate that ?

wyli:
that's from torch 1.12 update, i'm working on a fix https://github.com/Project-MONAI/MONAI/issues/4606

rijobro:
Thanks @doursand ! I can't see where `best_metric_model.pth` or the two `nii.gz` files are used. Are they used? If so, would it be possible to minimise the size of the files added? Perhaps generate the model on the fly (doesn't need to be good quality), and use test images that we already have present?

doursand:
@rijobro you are absolutely right , these 2 files are not used at all. I believe that this is because I did a full test locally and it may have downloaded those files in my local repo. I will remove them and resubmit

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+
+class SignalResample(Transform):
+    """
+    Resample signal to the target sampling rate.
+    """
+
+    backend = [TransformBackends.NUMPY]
+
+    def __init__(
+        self, method: str = "interpolation", current_sample_rate: int = 500, target_sample_rate: int = 250
```
rijobro, line 64: 500 and 250 seem to me to be completely arbitrary, is that correct? If so, I think the defaults should be removed.

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+        """
+        Args:
+            method: which method to be used to resample the signal, default is interpolation.
+            Available options : {``"interpolation"``, ``"polynomial"``, ``"fourier"``}
+            current_sample_rate: initial sampling rate of the signal
+            target_sample_rate: target signal sample rate
+        """
+        if method is None and method not in ["interpolation", "polynomial", "fourier"]:
+            raise ValueError("Incompatible values: method needs to be either interpolation, polynomial or fourier.")
+        if current_sample_rate < target_sample_rate:
```
rijobro, line 75: Downsampling seems to be fairly simple to me, is that not the case?

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+        self.current_sample_rate = current_sample_rate
+        self.target_sample_rate = target_sample_rate
+
+    def __call__(self, signal: np.ndarray) -> np.ndarray:
+        """
+        Args:
+            signal: input 1 dimension signal to be resampled
+        """
+        if len(signal.shape) > 1:
+            inputs_channels = signal.shape[0]
```
rijobro, line 89: All of our transforms assume the channel dimension to be present, so maybe these transforms should make the same assumption to be consistent with them.

monai/transforms/signal/array.py
doursand: yes you're correct, i'll fix that

monai/transforms/signal/array.py
doursand: fair enough

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+        """
+        Args:
+            mode: define how the extension of the input array is done beyond its boundaries, see for more details :
+            https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.shift.html.
+            filling: value to fill past edges of input if mode is ‘constant’. Default is 0.0. see for mode details :
+            https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.shift.html.
+            boundaries: list defining lower and upper boundaries for the signal shift, example : ``[-1.0, 1.0]``
+        """
+        super().__init__()
+        if boundaries is None or None in boundaries:
```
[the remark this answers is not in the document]
rijobro: Could also check like this: `isinstance(boundaries, Sequence) and len(boundaries) == 2 and all(isinstance(i, float) for i in boundaries)`

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+
+
+class SignalRandScale(RandomizableTransform):
+    """
+    Apply a random rescaling on a signal
+    """
+
+    backend = [TransformBackends.NUMPY]
+
+    def __init__(self, boundaries: Optional[List[float]] = None, *args, **kwargs) -> None:
```
rijobro, line 166: Same comment here about boundaries default value (won't mention it again, but if changes are made, it makes sense to change all following instances).

monai/transforms/signal/array.py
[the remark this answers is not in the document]
rijobro: I see that boundaries are a common input argument, so you could lift out a small function to perform the checking?

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+
+    backend = [TransformBackends.NUMPY]
+
+    def __init__(
+        self,
+        mode: Optional[str] = "wrap",
+        filling: Optional[float] = 0.0,
+        boundaries: Optional[List[float]] = None,
+        *args,
+        **kwargs,
```
rijobro, line 123: `args` and `kwargs` are present in a lot of the constructors but are never used, so should be deleted.

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+        loc = np.random.choice(range(length))
+        signal = self._paste(signal, squaredpulse_partial, (loc,))
+
+        if len(signal.shape) == 1:
+            signal = np.expand_dims(signal, axis=0)
+
+        return signal
+
+
+class SignalNormalize(Transform):
```
rijobro, line 559: I actually don't see what this transform does that `ScaleIntensity` doesn't already do, so I would suggest removing this one.

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+            signal: signal to be normalized
+        """
+        if len(signal.shape) == 1:
+            signal = np.expand_dims(signal, axis=0)
+
+        signal = (signal - np.min(signal)) / (np.max(signal) - np.min(signal))
+        return signal
+
+
+class SignalStandardize(Transform):
```
rijobro, line 578: This is same as `NormalizeIntensity` with default values for `subtrahend` and `divisor` – can be deleted.

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+            signal: signal to be standardized
+        """
+        if len(signal.shape) == 1:
+            signal = np.expand_dims(signal, axis=0)
+
+        signal = (signal - signal.mean()) / signal.std()
+        return signal
+
+
+class SignalZeroPad(Transform):
```
rijobro, line 597: Think the same could be achieved with `SpatialPad(spatial_size=2000, method="end", mode="constant")`. Admittedly, `SpatialPad` is an ill-fitting name here.

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+            quality_factor: quality factor for notch filter
+            see : https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.iirnotch.html
+            sampling_freq: sampling frequency of the input signal
+        """
+        super().__init__()
+        self.frequency = frequency
+        self.quality_factor = quality_factor
+        self.sampling_freq = sampling_freq
+
+    def __call__(self, signal: Any) -> Any:
```
rijobro, line 674: `Any` is used here, but `np.ndarray` is used elsewhere.

rijobro:
Looks great, thanks for all the hard work. A few points:

1. There are a few classes here that already exist in a more general form in our code base. I think the current MONAI transforms should be used in favour (e.g., `ScaleIntensity` instead of `SignalNormalize`), so some of these transforms can be deleted.
1. There is repetition of methods such as `_paste_slices`. This implies to me that there should be more class inheritance.
1. By default you've used `TransformBackends.NUMPY`, which makes sense when `scikit` is used. But when the transforms is doing simple maths, I think they could be ambivalent to numpy/torch input. I'd then love to see CPU and GPU tensors tested.
1. It would be great if you could create visualisations for these so that they can be rendered in our documentation. See [SpatialPad](https://docs.monai.io/en/stable/transforms.html#spatialpad) for example. Images are stored in another github repository: https://github.com/Project-MONAI/DocImages.

Once these are done, we'll have to do equivalent dictionary transforms, though they should be trivial.

doursand:
Thanks a lot for the review ,I'll try to allocate time in the coming weeks to progress on this. It should be faster as I get a (little) bit  more familiar with monai in general. Again thanks for your time and valuable inputs!

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+zoom, has_zoom = optional_import("scipy.ndimage", name="zoom")
+resample_poly, has_resample_poly = optional_import("scipy.signal", name="resample_poly")
+fft, has_resample_fft = optional_import("scipy.signal", name="resample")
+shift, has_shift = optional_import("scipy.ndimage.interpolation", name="shift")
+square, has_square = optional_import("scipy.signal", name="square")
+iirnotch, has_iirnotch = optional_import("scipy.signal", name="iirnotch")
+filtfilt, has_filtfilt = optional_import("scipy.signal", name="filtfilt")
+spectrogram, has_spectrogram = optional_import("scipy.signal", name="spectrogram")
+central_frequency, has_central_frequency = optional_import("pywt", name="central_frequency")
+cwt, has_cwt = optional_import("pywt", name="cwt")
```
Nic-Ma, line 33: Hi @rijobro ,

As you mentioned in the comments, I agree MONAI already supports many operations.
I think @doursand may be not familiar with what exists in MONAI.
Could you please help confirm the optional imports here and let's only keep the "must-have"?
For example, we can use MONAI transforms for zoom, torch API for fft?

Thanks in advance.

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+
+    backend = [TransformBackends.NUMPY]
+
+    def __init__(
+        self, boundaries: Optional[List[float]] = None, frequencies: Optional[List[float]] = None, *args, **kwargs
+    ) -> None:
+        """
+        Args:
+            boundaries: list defining lower and upper boundaries for the square pulse magnitude,
+            lower and upper values need to be positive example : ``[0.2, 0.6]``
```
Nic-Ma, line 319: please `tap` for the second line of arg doc-string.

Thanks.

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+        super().__init__()
+        if boundaries is None or None in boundaries:
+            raise ValueError("Incompatible values: boundaries needs to be a list of float.")
+        if (boundaries is None) or (None in boundaries) or (any(x < 0 for x in boundaries)):
+            raise ValueError("Incompatible values: boundaries needs to be a list of positive float.")
+
+        self.boundaries = boundaries
+        self.frequencies = frequencies
+
+    def _randomize(self):
```
Nic-Ma, line 332: Maybe change to override `randomize()` directly?
Same for all the other transforms.

Thanks.

Nic-Ma (commented):
Thanks for the great contribution! It will definitely help a lot by adding the signal support.
I see most of the transforms are using `numpy` backend, is it possible to update them to `torch` APIs instead? Then we can even enable `MetaTensor` for all these transforms?
@wyli @rijobro @ericspod What do you guys think?

Thanks in advance.

monai/transforms/signal/array.py
Can-Zhao: There are fft utils in monai, not sure whether they are useful here. https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/fft_utils.py

monai/transforms/signal/array.py
doursand: used self.randomize() directly

monai/transforms/signal/array.py
doursand: removed

monai/transforms/signal/array.py
doursand: removed

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,773 @@
...
+        super().randomize(None)
+        self.magnitude = self.R.uniform(low=self.boundaries[0], high=self.boundaries[1])
+
+    def __call__(self, signal: np.ndarray) -> np.ndarray:
+        """
+        Args:
+            signal: input 1 dimension signal to be shifted
+        """
+        self._randomize()
+        if len(signal.shape) == 1:
```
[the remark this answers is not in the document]
doursand: removed , assumed channel is already present

monai/transforms/signal/array.py
doursand: removed , assumed channel is already there

monai/transforms/signal/array.py
doursand: fixed

monai/transforms/signal/array.py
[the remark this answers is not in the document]
doursand: fixed, function added to utils to perform the check

monai/transforms/signal/array.py
doursand: fixed

monai/transforms/signal/array.py
doursand: @Nic-Ma indeed I just realized that Zoom does the same as SignalResample, so I m removing it from my code ;-)
For numpy backend, it was done on purpose , I wanted to have something working initially on numpy and them update to torch as well.

@Can-Zhao Thanks for the tip, indeed I don't know well enough monai and all the stuffs which are already there, so I will try to use that one first

monai/transforms/signal/array.py
doursand: SignalResample is doing the same as zoom, so I ll remove that one from my code

monai/transforms/utils.py
```diff
@@ -1660,5 +1661,65 @@ def sync_meta_info(key, data_dict, t: bool = True):
...
+
+
+def squarepulse(sig, duty: float = 0.5):
+    """
+    compute squarepulse using pytorch
+    equivalent to numpy implementation from
+    https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.square.html
+    """
+    t, w = torch.tensor(sig), torch.tensor(duty)
+    w = torch.tensor(w + (t - t))
```
ericspod, line 1709: Is the `t - t` part here needed? If there's a rationale please comment on it. Also using `torch.tensor` directly throws up the warning `To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).` You can use `torch.as_tensor` when making tensors for arguments but on this line you don't need anything, `w+(t-t)` produces a tensor. If you did want to copy then `clone` is what to use.

doursand: indeed (t - t) is 100% useless :-) and thanks for the hints on tensor stuffs
By the way, is the location of the functions `check_boundaries`, `paste`, `paste_slices` and `squarepulse` correct or is there a better place to store them in ? Those are helper functions for the signal transforms, I don't think they can be used for anything else

ericspod: For now we can have these functions here, `utils.py` is getting rather large so we should consider a refactor in the future however. Your functions are specific to your transforms but it's possible you could use existing general purpose functions to do the same job (`copypaste_arrays` perhaps could work in place on your `paste`). I wouldn't worry too much about that now since again we can refactor functions in the future to reduce code.

doursand: right thanks. Last thing, as I m looking at fourier transform, I realize that in fact we could use the existing function in `torchaudio` , is `torchaudio` considered as part of monai "core", as it is part of the pytorch ecosystem or should I consider it as an optional dependency here ?

doursand: @ericspod back on this topic ... I think I can simply remove this STFT transform for now as again there is already an equivalent transform in torchaudio and we can easily chain it with the signal transforms like for instance :

`tfs = Compose([SignalRandAddGaussianNoise(boundaries=(0.001,0.02)),
                Spectrogram(n_fft=64,
                win_length=64,
                hop_length=32,
                center=True,
                pad_mode="reflect",
                power=1.0,
                normalized=True)
               ])`

monai/transforms/signal/array.py
```diff
@@ -0,0 +1,499 @@
...
+        """
+        super().__init__()
+        self.replacement = replacement
+
+    def __call__(self, signal: NdarrayOrTensor) -> NdarrayOrTensor:
+        """
+        Args:
+            signal: signal to be filled
+        """
+        signal = torch.nan_to_num(convert_to_tensor(signal), nan=self.replacement)
```
wyli, line 381: this `torch.nan_to_num` is not available in torch 1.7, perhaps 
```py
ERROR: test_correct_parameters_multi_channels (tests.test_signal_fillempty.TestSignalRandDrop)
  File "/home/runner/work/MONAI/MONAI/tests/test_signal_fillempty.py", line 28, in test_correct_parameters_multi_channels
    fillemptysignal = fillempty(sig)
  File "/home/runner/work/MONAI/MONAI/monai/transforms/signal/array.py", line 381, in __call__
    signal = torch.nan_to_num(convert_to_tensor(signal), nan=self.replacement)
AttributeError: module 'torch' has no attribute 'nan_to_num'

```

you can put a decorator in the test to skip the case, for example: https://github.com/Project-MONAI/MONAI/blob/1516ca758090dd8ed2890b64554a77f681a0e224/tests/test_reg_loss_integration.py#L43-L45

monai/transforms/utils.py
ericspod: You can if you like but we'd be adding torchaudio as an optional dependency rather than having everything inclusive in our own library.

monai/transforms/signal/array.py
wyli: it seems `torch.pi` is also not available in the previous versions https://github.com/Project-MONAI/MONAI/runs/7920616362?check_suite_focus=true

doursand:
@wyli : the remaining errors are not related to my PR (tests.test_slice_inferer.TestSliceInferer) ...

wyli (approved):
Thanks, it looks good to me.

bruAristimunha:
Hello guys!  I loved this contribution.  Are there any plans to do a transformation to EEG reading?

PR #4536, opened 2022-06-20
[don't merge] this PR is for dev code comparisons/discussions

wyli:
for the convenience of comparing `dev` vs `feature/MetaTensor`

### Status
**Ready/Work in progress/Hold**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick --unittests  --disttests`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

PR #4548, opened 2022-06-21
Move metatensor support into dev branch (crop/pad)

Nic-Ma:
### Description
This PR added `MetaTensor` support for all the crop & pad transforms.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @wyli , @rijobro ,

Thanks for sharing the implementation in the metatensor branch, I refer to the PR #4536 and made the change for `Pad` transform in this PR, could you please help take a look first then I will dive into more transforms?
The main differences between this PR and the implementation in #4536:
1. The transform can accept `tensor`, `metatensor` or `numpy`, but the computation backend is always tensor.
2. The output of the transform is always `tensor` or `metatensor`, I see the typehints in metatensor branch are `torch.Tensor` for transforms input/output now.
3. The value modification are in-place for the image and metadata, same as our current existing transforms.

Thanks.

Nic-Ma:
Hi @wyli @rijobro ,

I have integrated all the crop and pad transforms and migrated all the crop & pad unit tests from https://github.com/Project-MONAI/MONAI/pull/4536.
Please note that:
1. Some CI tests can't pass so far because having some other dependencies need to update.
2. I dropped the `inverse` operation of "crop samples" transforms.

Thanks.

monai/transforms/croppad/array.py
```diff
@@ -84,63 +84,110 @@ class Pad(Transform):
...
+                    if get_track_meta():
+                        out = MetaTensor(out, meta=img_t.meta, applied_operations=img_t.applied_operations)
+        else:
+            out = img_t
+        if get_track_meta():
+            self._update_meta(tensor=out, to_pad=to_pad_)  # type: ignore
+            self.push_transform(out, extra_info={"padded": to_pad_})
+        return out
+
+    def _update_meta(self, tensor: MetaTensor, to_pad: List[Tuple[int, int]]):
```
wyli, line 167: could you rename `_update_meta` -> `update_meta`, this method is useful in general

monai/transforms/croppad/array.py
```diff
@@ -160,62 +207,42 @@ class SpatialPad(Transform):
 
     """
 
-    backend = Pad.backend
-
     def __init__(
         self,
         spatial_size: Union[Sequence[int], int],
         method: Union[Method, str] = Method.SYMMETRIC,
```
wyli, line 213: please help simplify the type hint in other places as well.
```suggestion
        method: str = Method.SYMMETRIC,
```

monai/transforms/croppad/array.py
```diff
@@ -300,70 +298,146 @@ class DivisiblePad(Transform):
     def __init__(
         self,
         k: Union[Sequence[int], int],
-        mode: Union[NumpyPadMode, PytorchPadMode, str] = NumpyPadMode.CONSTANT,
         method: Union[Method, str] = Method.SYMMETRIC,
+        mode: str = PytorchPadMode.CONSTANT,
```
wyli, line 302: the position of `method` and `mode` are swapped, perhaps we keep the original ones to avoid breaking change.

monai/transforms/croppad/array.py
```diff
@@ -644,15 +697,19 @@ def set_random_state(
...
         pass
 
-    def __call__(self, img: NdarrayOrTensor) -> List[NdarrayOrTensor]:
+    def __call__(self, img: torch.Tensor) -> List[torch.Tensor]:
         """
         Apply the transform to `img`, assuming `img` is channel-first and
         cropping doesn't change the channel dim.
         """
-        return [self.cropper(img) for _ in range(self.num_samples)]
+        ret = [self.cropper(img) for _ in range(self.num_samples)]
```
wyli, line 705: I think this will require deepcopy of meta dict?

monai/transforms/croppad/array.py
```diff
@@ -84,63 +84,110 @@ class Pad(Transform):
...
+
+        img_t = convert_to_tensor(data=img, track_meta=get_track_meta())
+
+        # all zeros, skip padding
+        if np.asarray(to_pad_).any():
+            try:
+                mode_ = convert_pad_mode(dst=img_t, mode=mode_).value
+                out = self._pt_pad(img_t, pad_width=to_pad_, mode=mode_, **kwargs_)
+            # but if mode or args don't exist in pytorch, use numpy instead
+            except (ValueError, TypeError) as err:
```
wyli, line 152: the exception handling will be slow, if we know the user-specified mode is numpy only, perhaps we switch to numpy before the exception is raised? this will be efficient in typical use cases.

monai/transforms/croppad/dictionary.py
```diff
@@ -1079,22 +793,12 @@ class RandCropByPosNegLabeld(Randomizable, MapTransform, InvertibleTransform):
...
-            used to add `patch_index` to the meta dict.
         allow_smaller: if `False`, an exception will be raised if the image is smaller than
             the requested ROI in any dimension. If `True`, any smaller dimensions will be set to
             match the cropped size (i.e., no cropping in that dimension).
         allow_missing_keys: don't raise exception if key is missing.
-
-    Raises:
-        ValueError: When ``pos`` or ``neg`` are negative.
+        padcropper = ResizeWithPadOrCrop(spatial_size=spatial_size, method=method, **pad_kwargs)
+        super().__init__(keys, padder=padcropper, mg`` are negative.
```
wyli, line 801: this looks like typos

monai/transforms/croppad/dictionary.py
```diff
@@ -103,24 +107,69 @@
     "RandCropByLabelClassesDict",
 ]
 
-PadModeSequence = Union[Sequence[Union[NumpyPadMode, PytorchPadMode, str]], NumpyPadMode, PytorchPadMode, str]
 DEFAULT_POST_FIX = PostFix.meta()
```
wyli, line 110: this is not needed any more ?

wyli (commented):
thanks, I put some initial comments inline, please help fix the docs error: https://github.com/Project-MONAI/MONAI/runs/7087792214?check_suite_focus=true

monai/transforms/croppad/array.py
Nic-Ma: updated in the latest commit.

Thanks.

monai/transforms/croppad/array.py
Nic-Ma: Sure, restored in the latest commit.

Thanks.

monai/transforms/croppad/array.py
Nic-Ma: Hi @wyli ,

Actually, I asked you about this problem offline, you told me `MetaTensor` will automatically deepcopy when cropping if not a batch data?
Let me add the deepcopy logic if necessary.

Thanks.

monai/transforms/croppad/array.py
Nic-Ma: Hi @wyli ,

That's the same logic from #4536 , I think it's not easy to detect whether the args are for numpy.
For example, both numpy and torch have the `constant` mode, but the `kwargs` can be different.

Thanks.

monai/transforms/croppad/dictionary.py
Nic-Ma: OK, removed in the latest commit.

Thanks.

monai/transforms/croppad/array.py
wyli: My understanding is for the numpy only modes we can decide before raising an exception.

monai/transforms/croppad/array.py
wyli: By default it'll be automatic but in this pr you have made most changes in-plcae as you commented earlier https://github.com/Project-MONAI/MONAI/pull/4548#issuecomment-1161268944

monai/transforms/croppad/array.py
Nic-Ma: OK, updated in the latest commit.

Thanks.

monai/transforms/croppad/array.py
Nic-Ma: Hi @wyli ,

I think the `slices` operation on a `MetaTensor` in the crop transforms will automatically deepcopy the content (including meta), just as we discussed offline before.
I added more unit tests to cover it.
It also can be verified by below program:
```py
from monai.utils import convert_to_tensor

a = convert_to_tensor([1, 2, 3], track_meta=True, wrap_sequence=True)
b = a[1:]
b.meta["patch_index"] = 1
b.affine[0, 1] = 2
```
```
>>> print(b)
tensor([2, 3])
MetaData
        affine: tensor([[1., 2., 0., 0.],
        [0., 1., 0., 0.],
        [0., 0., 1., 0.],
        patch_index: 1

Applied operations

Is batch?: False
>>> 
>>> print(a)
tensor([1, 2, 3])
MetaData
        [0., 1., 0., 0.],
        [0., 0., 1., 0.],

Applied operations

Is batch?: False
>>>
```
What do you think?

Thanks.

wyli: I see, it looks good here.

wyli (approved):
looks good to me

PR #4561, opened 2022-06-22
Wrap `Type` to the typing of `writer` parameter of `SaveImage`

### LINKED ISSUE #4560: Wrong typing for `writer` parameter of `SaveImage`
Currently, the typing for `writer` parameter of `SaveImage` is `Union[image_writer.ImageWriter, str, None]`. But it is supposed to be some class, thus the `image_writer.ImageWriter` part in the Union should be `Type[image_writer.ImageWriter]` where `Type` is from `typing` module.

I can make a simple fix and submit a PR. Maybe we could check if there's any similar issue in other parts of the code base.

function2-llx:
Fixes #4560.

### Status
**ready for review**

### Types of changes

### DIFF
--- a/monai/transforms/io/array.py
+++ b/monai/transforms/io/array.py
@@ -20,7 +20,7 @@ import traceback
 import warnings
 from pathlib import Path
 from pydoc import locate
-from typing import Dict, List, Optional, Sequence, Union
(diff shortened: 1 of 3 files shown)


wyli (approved):
Thanks for the fix, it looks good to me.

I'll merge this one if you mark it 'ready for review'.

Issue #4565, opened 2022-06-22
TORCH backend for RandHistogramShift transform


dyollb:
**Is your feature request related to a problem? Please describe.**
Currently the only transform in our augmentation pipeline that doesn't have a TORCH backend is the `RandHistogramShiftd`.
I have not profiled the speedup and GPU memory for the augmentation transforms, but I guess if we could do most of it on the GPU it should improve performance.

**Describe the solution you'd like**
Implement TORCH backend for `RandHistogramShift`.

**Describe alternatives (solution) you've considered**
I would propose to
- not change `RandHistogramShift.randomize`, i.e. control points remain `np.ndarray`
- cast the control points to the input `NdarrayOrTensor` type
- implement a function with the functionality/API of `np.interp` ([used in `__call__`](https://github.com/Project-MONAI/MONAI/blob/669bddf581201f994d1bcc0cb780854901605d9b/monai/transforms/intensity/array.py#L1400)) that supports `NdarrayOrTensor` 

An example implementation for `np.interp`-like function using `torch.Tensor` is mentioned here: https://github.com/pytorch/pytorch/issues/50334

If you agree to this strategy I will submit a PR.

dyollb:
How about something like this?:
```
class RandHistogramShift(RandomizableTransform):

    backend = [TransformBackends.TORCH, TransformBackends.NUMPY]

    def __init__(self, num_control_points: Union[Tuple[int, int], int] = 10, prob: float = 0.1) -> None:
        ...

    def randomize(self, data: Optional[Any] = None) -> None:
        ...

    def interp(x: NdarrayOrTensor, xp: NdarrayOrTensor, fp: NdarrayOrTensor) -> NdarrayOrTensor:

        m = (fp[1:]- fp[:-1]) / (xp[1:]- xp[:-1])
        b = fp[:-1] - (m * xp[:-1])

        indices = (x.reshape(-1, 1) >= xp.reshape(1, -1)).sum(1) - 1
        indices = indices.clip(0, len(m) - 1)

        return (m[indices] * x.reshape(-1) + b[indices]).reshape( x.shape)

    def __call__(self, img: NdarrayOrTensor, randomize: bool = True) -> NdarrayOrTensor:
        if randomize:
            self.randomize()

        if not self._do_transform:
            return img

        if self.reference_control_points is None or self.floating_control_points is None:
            raise RuntimeError("please call the `randomize()` function first.")
        
[...]

PR #4567, opened 2022-06-23
add TORCH backend for RandHistogramShift

dyollb:
Signed-off-by: Bryn Lloyd <lloyd@itis.swiss>

Fixes # .

### Description
Currently the only transform in our augmentation pipeline that doesn't have a TORCH backend is the RandHistogramShiftd.
I have not profiled the speedup and GPU memory for the augmentation transforms, but I guess if we could do most of it on the GPU it should improve performance.

Solves https://github.com/Project-MONAI/MONAI/issues/4565

### Status
**Ready**

### Types of changes


monai/transforms/intensity/array.py
```diff
@@ -1368,8 +1368,25 @@ def __init__(self, num_control_points: Union[Tuple[int, int], int] = 10, prob: f
...
+    def interp(self, x: NdarrayOrTensor, xp: NdarrayOrTensor, fp: NdarrayOrTensor) -> NdarrayOrTensor:
+        ns = torch if isinstance(x, torch.Tensor) else np
+        if isinstance(x, np.ndarray):
+            # approx 2x faster than code below for ndarray
+            return np.interp(x, xp, fp)
+
+        m = (fp[1:] - fp[:-1]) / (xp[1:] - xp[:-1])
+        b = fp[:-1] - (m * xp[:-1])
+
+        indices = ns.searchsorted(xp.reshape(-1), x.reshape(-1), side="left") - 1
```
wyli, line 1383: thanks, the solution looks nice, just this function has a compatibility issue with the previous pytorch versions, I think it could be addressed with some version check like this one:
https://github.com/Project-MONAI/MONAI/blob/669bddf581201f994d1bcc0cb780854901605d9b/monai/transforms/spatial/array.py#L232-L234

dyollb: `[torch, numpy].searchsorted(..., side='side')` is the default. So if older torch versions don't have the `side` argument but use `right: bool = False` instead we can simply remove the argument.

Also, for the use case here it doesn't seem relevant: histogram control points with repeated x-values probably don't make much sense.

wyli: still the previous versions of torch has some issue
```
ERROR: test_interp (tests.test_rand_histogram_shift.TestRandHistogramShift)
  File "/home/runner/work/MONAI/MONAI/tests/test_rand_histogram_shift.py", line 60, in test_interp
    yi = tr.interp(array_type([0, 2, 4, 8, 10]), x, y)
  File "/home/runner/work/MONAI/MONAI/monai/transforms/intensity/array.py", line 1387, in interp
    f[x < xp[0]] = fp[0]  # type: ignore
RuntimeError: Index put requires the source and destination dtypes match, got Float for the destination and Long for the source.

```

btw I suspect the same indices calculation could be done with `torch.bucketize` and maybe it's efficient https://pytorch.org/docs/stable/generated/torch.bucketize.html

wyli: I think the error is about assigning `fp` (in long) to the larger tensor `f` (in float). perhaps naturally raising this runtime error is not a bad thing to have here...

PR #4598, opened 2022-06-27
Fix the location type returned in GridPatch when padding

### LINKED ISSUE #4597: GridPatch returns inconsistent type location when padding
GridPatch adds patches of constant value and 0, 0 location when the number of required patches is bigger than the patches that can be extracted from the image. The patch location returned in this case should be a List `[0, 0]` ,to be consistent with the type of the otherwise returned types, while it currently is a List of Lists `[[0], [0]]`.

The reason of this behavior can be seen in this line https://github.com/Project-MONAI/MONAI/blob/23d9b5217f0b367e3bedf24d6e508131e66a0cd1/monai/transforms/spatial/array.py#L2773

The above line should be changed to `np.zeros((len(self.patch_size)))`

vale-salvatelli:
Fixes #4597  .

### Description
The location type returned in GridPatch when padding should be List to be consistent with the locations returned when not padding

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -2770,7 +2770,7 @@ class GridPatch(Transform):
             patch = convert_to_dst_type(
                 src=np.full((array.shape[0], *self.patch_size), self.pad_kwargs.get("constant_values", 0)), dst=array
             )[0]
-            start_location = convert_to_dst_type(src=np.zeros((len(self.patch_size), 1)), dst=array)[0]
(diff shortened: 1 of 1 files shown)


vale-salvatelli:
Sorry for the slow answer. I signed my last commit - hope it's OK now

wyli:
> Sorry for the slow answer. I signed my last commit - hope it's OK now @wyli

thanks! I'll merge this one soon.

PR #4633, opened 2022-07-05
Permanent dataset hashes transforms

### LINKED ISSUE #999: Control Consistency of PersistentDataset
Already included as a warning in the source code, the `PersistentDataset` does not check if any of the transformations changed, and thus after modifying the transformations, the cache is outdated.

Calculate the hash of the very first element of the dataset (after calling `_pre_first_random_transform`) and store this in the cache folder. In case the stored hash matches with the one calculated during `__init__` of `PersistentDataset` the cache is still valid and no deterministic transformations have changed. 
However, if the hashes do not match, either a new cache folder could be created or the old cache can be overwritten.

Alternatively, the deterministic transformations could be hashed, but I am not sure how that would work. Might be more memory efficient, but potentially harder to implement due to `TypeError: Object of type *** is not JSON serializable` for most transforms.

rijobro:
### Description
With this change, `PermanentDataset` creates a hash of the deterministic transforms and adds this to the filename. Hence, as transforms change, so to should the filename. This reduces errors due to transforms changing during experiments.

fixes #999

### Status
**Ready**

### Types of changes


wyli (approved):
thanks, it looks good to me, I guess it'll close https://github.com/Project-MONAI/MONAI/issues/999?

rijobro:
Fixes #999

monai/data/dataset.py
```diff
@@ -246,6 +247,28 @@ def __init__(
                 self.cache_dir.mkdir(parents=True, exist_ok=True)
             if not self.cache_dir.is_dir():
                 raise ValueError("cache_dir must be a directory.")
+        self.set_transform_hash()
```
Nic-Ma, line 250: Thanks for the good feature.
I am afraid it may affect the performance of PerisistentDataset, especially when not JSON serializable.
Could you please help make it optional and default to True?
CC @wyli @ericspod 

Thanks.

ChenglongWang:
I'm wondering what's the definition of "transforms change". Is it possible to hash the argument changes? Perhaps more detailed docs/examples are required. I think users need to know exactly the data they are processing.

KumoLiu:
Hi @ChenglongWang, 
        The same transforms with different argument would also make hash change!
Thanks!

Nic-Ma:
Thanks @KumoLiu for your explanation.
I think most of the transforms are JSON serializable, here only the exceptions will only hash with the transform name.

Thanks.

monai/data/dataset.py
wyli: agreed, @rijobro could you please make it optional? then I'll merge this one soon. Thanks!

PR #4635, opened 2022-07-05
Fix casting of location dtype in GridPatch 

### LINKED ISSUE #4596: Wrong type casting in GridPatch produces wrong locations for patches
**Description**
In the class GridPatch (monai/transforms/spatial/array.py) the type of the locations is inferred looking at the type of image array. This is erroneous because the type of the image array is related to the range of **intensity** value in the image, while the location type is related to the **size** of the image. For example, the image might contain values in the range 0-255 and be casted to uint8, while the image size might span range 0 -1000 (or more for pathology images). This forced conversion leads to completely erroneous returned locations in output.

Below a screenshot of where the problem is. The issue can be reproduced by:
* loading images in uint8
* applying GridPatch transform to the image
* looking at the returned type (and values) of the locations

MicrosoftTeams-image

Link to the line that creates the issue https://github.com/Project-MONAI/MONAI/blob/23d9b5217f0b367e3bedf24d6e508131e66a0cd1/monai/transforms/spatial/array.py#L2765

To be noted also that the test of this class run an assertion only on the returned patch, not the location https://github.com/Project-MONAI/MONAI/blob/23d9b5217f0b367e3bedf24d6e508131e66a0cd1/tests/test_grid_patch.py#L76

vale-salvatelli:
Fixes #4596 

Locations dtype in GridPatch was erroneously always inferred from the image dtype. This led to wrong locations values when the image was casted to small integers (eg. pixels values span 0 -255, locations values can span 0 -5000). This behavior is now fixed, locations will always be casted to be regular int


monai/transforms/spatial/array.py
```diff
@@ -3128,8 +3128,16 @@ def __call__(self, array: NdarrayOrTensor):
...
+        if isinstance(array, torch.Tensor):
+            location_dtype = torch.int
+        elif isinstance(array, np.ndarray):
+            location_dtype = np.int
+        else:
+            location_dtype = None
         output = list(
-            zip(convert_to_dst_type(src=patched_image, dst=array)[0], convert_to_dst_type(src=locations, dst=array)[0])
+            zip(convert_to_dst_type(src=patched_image, dst=array)[0],
+                convert_to_dst_type(src=locations, dst=array, dtype=location_dtype)[0])
```
wyli, line 3140: thanks for the patch, I think 
```py
convert_to_dst_type(src=locations, dst=array, dtype=int)[0]
```
is good enough in this case, because the function will handle `dtype` properly based on the `dst`:

```py
>>> convert_to_dst_type(torch.zeros((2,2)), dst=numpy.zeros((2, 2)), dtype=int)[0]
array([[0, 0],
       [0, 0]])
>>> convert_to_dst_type(torch.zeros((2,2)), dst=torch.zeros((2,2)), dtype=int)[0]
tensor([[0, 0],
        [0, 0]])
>>> convert_to_dst_type(numpy.zeros((2,2)), dst=torch.zeros((2,2)), dtype=int)[0]
tensor([[0, 0],
        [0, 0]])
>>> convert_to_dst_type(numpy.zeros((2,2)), dst=numpy.zeros((2,2)), dtype=int)[0]
array([[0, 0],
       [0, 0]])
```

vale-salvatelli: Brilliant - I updated accordingly

wyli (approved):
Thanks, it looks good to me.

PR #4638, opened 2022-07-06
4636 4637 backward compatible types

### LINKED ISSUE #4636: metatensor supports `astype(dtype, device)`
would be great to support `astype` API like numpy's astype https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html

`astype(dtype)` the dtype could be numpy based (and detach if needed when it's from a torch.tensor), as well as torch based.

### LINKED ISSUE #4637: backward compatible `EnsureType`
the current `EnsureType` removes any metadata
https://github.com/Project-MONAI/MONAI/blob/99b9dd7cf0667113e56f0dd90d40699115048d59/monai/transforms/utility/array.py#L464

given it's extensively used in the existing tutorials, would be great to enable metatensor output support and remove the inverse to preserve metadata when inverting the transform.
https://github.com/Project-MONAI/MONAI/blob/99b9dd7cf0667113e56f0dd90d40699115048d59/monai/transforms/utility/dictionary.py#L570-L574

### LINKED ISSUE #4639: convert_to_numpy issue
**Describe the bug**
```py
>>> from monai.utils import convert_to_numpy
>>> import torch
>>> import numpy as np
>>> convert_to_numpy(data=torch.zeros((2,2),dtype=torch.long), dtype=np.uint16)
  File "monai/utils/type_conversion.py", line 175, in convert_to_numpy
    data = data.detach().to(dtype=get_equivalent_dtype(dtype, torch.Tensor), device="cpu").numpy()
  File "monai/utils/type_conversion.py", line 79, in get_equivalent_dtype
    return dtype_numpy_to_torch(dtype)
  File "monai/utils/type_conversion.py", line 58, in dtype_numpy_to_torch
    return torch.from_numpy(np.empty([], dtype=dtype)).dtype
TypeError: can't convert np.ndarray of type numpy.uint16. The only supported types are: float64, float32, float16, complex64, complex128, int64, int32, int16, int8, uint8, and bool.
>>> 

```

wyli:

Fixes #4636
Fixes #4637
Fixes #4639

### Description
- adds `track_meta` to `EnsureType(d)`
- adds `astype(...)` to `MetaTensor`

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/utility/dictionary.py
+++ b/monai/transforms/utility/dictionary.py
@@ -62,7 +62,7 @@ from monai.transforms.utility.array import (
 )
 from monai.transforms.utils import extreme_points_to_image, get_extreme_points
 from monai.transforms.utils_pytorch_numpy_unification import concatenate
-from monai.utils import convert_to_numpy, deprecated, deprecated_arg, ensure_tuple, ensure_tuple_rep
+from monai.utils import deprecated, deprecated_arg, ensure_tuple, ensure_tuple_rep
 from monai.utils.enums import PostFix, TraceKeys, TransformBackends
 from monai.utils.type_conversion import convert_to_dst_type
 
@@ -519,7 +519,7 @@ class ToTensord(MapTransform, InvertibleTransform):
         return d
 
 
-class EnsureTyped(MapTransform, InvertibleTransform):
+class EnsureTyped(MapTransform):
     """
     Dictionary-based wrapper of :py:class:`monai.transforms.EnsureType`.
 
@@ -541,6 +541,7 @@ class EnsureTyped(MapTransform, InvertibleTransform):
         dtype: Union[DtypeLike, torch.dtype] = None,
         device: Optional[torch.device] = None,
         wrap_sequence: bool = True,
+        track_meta: Optional[bool] = None,
         allow_missing_keys: bool = False,
     ) -> None:
         """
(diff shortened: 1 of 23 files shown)


Nic-Ma:
Hi @KumoLiu ,

Could you please help take a look at this PR according to your observation in the tutorials?

Thanks in advance.

monai/transforms/utility/dictionary.py
```diff
@@ -552,28 +553,21 @@ def __init__(
...
-    def inverse(self, data: Mapping[Hashable, Any]) -> Dict[Hashable, Any]:
-        d = deepcopy(dict(data))
-        for key in self.key_iterator(d):
-            # FIXME: currently, only convert tensor data to numpy array or scalar number,
-            # need to also invert numpy array but it's not easy to determine the previous data type
-            d[key] = convert_to_numpy(d[key])
-            # Remove the applied transform
-            self.pop_transform(d, key)
-        return d
-
```
KumoLiu, line 576: I know it would be great to enable metatensor output support and remove the inverse to preserve metadata when inverting the transform. But will it cause incompatibility if a user used EnsureTyped.Inverse() with old version?

monai/utils/type_conversion.py
```diff
@@ -239,8 +239,8 @@ def convert_data_type(
```
KumoLiu, line 239: Maybe line238 need to modify to 'Convert to `MetaTensor` /`torch.Tensor`/`np.ndarray` from `torch.Tensor`/`np.ndarray`/`float`/`int` etc.'

monai/transforms/utility/dictionary.py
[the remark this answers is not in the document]
KumoLiu: Hi @Nic-Ma,
Sorry maybe I didn't explain it well. In this case, inverse() can not use anymore because it has been deleted. Maybe someone had used it may get an error？
Thanks in advance.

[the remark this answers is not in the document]
wyli: thanks, I put an inverse with the copy of data. I was not sure about it because of the `wrap_sequence` recursive cases.

monai/utils/type_conversion.py
wyli: thanks, this is updated.

monai/transforms/utility/dictionary.py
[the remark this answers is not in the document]
Nic-Ma: Hi @KumoLiu ,

I see your point. The `inverse()` operation of `EnsureType` is to convert a Tensor back to numpy array, because we usually use numpy array for transforms.
But now from this version we suppose to use `MetaTensor` for transforms, so I think we don't need to support the inverse support anymore.
And people usually call the inverse operation for the transform chain, deleting the `inverse()` of `EnsureType` will not break the inverse of the chain.

Thanks.

monai/utils/type_conversion.py
```diff
@@ -172,7 +172,7 @@ def convert_to_numpy(data, dtype: DtypeLike = None, wrap_sequence: bool = False)
             E.g., `[1, 2]` -> `[array(1), array(2)]`. If `True`, then `[1, 2]` -> `array([1, 2])`.
     """
     if isinstance(data, torch.Tensor):
-        data = data.detach().to(dtype=get_equivalent_dtype(dtype, torch.Tensor), device="cpu").numpy()
+        data = np.asarray(data.detach().to(device="cpu").numpy(), dtype=get_equivalent_dtype(dtype, np.ndarray))
```
Nic-Ma, line 175: I think maybe the previous code has a better performance?
Especially when converting a GPU tensor to numpy array, it converted to the expected `dtype` on CPU directly.
But this new code makes it into 2 stages.
What do you think?

Thanks.

Nic-Ma (approved):
Looks good to me except for 1 minor comment inline.

Thanks.

monai/utils/type_conversion.py
wyli: this is to fix a bug https://github.com/Project-MONAI/MONAI/issues/4639

Nic-Ma (approved):

PR #4651, opened 2022-07-07
Metric: r2

rijobro:
### Description
Implements r2 metric to compare similarities between two tensors. This is a useful metric as it can be expressed as a percentage, as opposed to MSE for example which is arbitrary.

Useful in image regression.

### Status
**Ready**

### Types of changes


Nic-Ma (approved):
Looks good to me.
@KumoLiu Could you please help double confirm this regression metric computation of R2?

Thanks in advance.

KumoLiu:
Hi @Nic-Ma,
        I wonder why assign y.sum(1) to two different variables (n and sum_y).
       And I didn't find the definition of  the mean of the observed data which needed by R2 metric definition. I'm not sure whether it is a variation of the original formula?
Thanks.

KumoLiu:
Hi @rijobro,
I wonder why assign y.sum(1) to two different variables (n and sum_y).
And I didn't find the definition of the mean of the observed data which needed by R2 metric definition. I'm not sure whether it is a variation of the original formula or maybe just I misunderstood.
Thanks.

PR #4653, opened 2022-07-07
4646 fastmri reader

mersad95zd:
Fixes # .

### Description
FastMRI reader for MONAI

### Status
**Work in progress**

### Types of changes


mersad95zd:
ToDo: Write and add its unittest.

monai/apps/reconstruction/fastMRIreader.py
```diff
@@ -0,0 +1,98 @@
...
+from monai.utils.type_conversion import convert_to_tensor
+
+h5py, has_h5py = optional_import("h5py")
+
+
+@require_pkg(pkg_name="h5py")
+class FastMRIReader(ImageReader):
+    """
+    Load fastMRI files with '.h5' suffix.
+    """
```
Can-Zhao, line 30: Could you add more details on how fastMRI h5 data look like? Thanks!

monai/apps/reconstruction/fastMRIreader.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def __init__(self):
+        super().__init__()
+
+    def verify_suffix(self, filename: Union[Sequence[PathLike], PathLike]) -> bool:
```
Can-Zhao, line 35: This func is not used in this PR. Please double check it will be used in somewhere else.

Can-Zhao:
Could you resolve the pakage error by adding empty `_init__.py` to the folder?

monai/apps/reconstruction/fastMRIreader.py
mersad95zd: Yes, `LoadImage` transform https://github.com/Project-MONAI/MONAI/blob/112d36c659b0cae510f44f78edfddea93ba314d3/monai/transforms/io/array.py#L89 calls ``verify_suffix`.

Every `ImageReader` class https://github.com/Project-MONAI/MONAI/blob/7f80288381575121e670a4e184d45b01813a51be/monai/data/image_reader.py#L66 should define a `verify_suffix`.

Can-Zhao: Great! Thanks!

monai/apps/reconstruction/fastMRIreader.py
```diff
@@ -0,0 +1,109 @@
...
+            data = data[0]
+
+        with h5py.File(data, "r") as f:
+            # extract everything from the ht5 file
+            dat = dict(
+                [(key, f[key][()]) for key in f]
+                + [(key, f.attrs[key]) for key in f.attrs]
+                + [("filename", str(data).split("/")[-1])]
+            )
+        f.close()
```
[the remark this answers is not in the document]
mersad95zd: Not necessarily, good point.
The only thing is for ipython users. If they call it and f.close() is not there, it'll cause a memory crash for them because ipyhton saves the object in memory if it's not explicitly closed.

Should we care about this use case? If not, I can just remove that. Please let me know what you think, thanks.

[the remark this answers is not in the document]
wyli: I think this could be removed, the context manager should already close `f` properly...also, I think `str(data).split("/")[-1]` could be more robust with `os.path.basename(data)`? I'm not familiar with the dataset so I might be wrong

mersad95zd:
Thanks @Nic-Ma for your comments. Let me know if any further edits are needed, thanks!

PR #4714, opened 2022-07-16
simplify copy attr  speed up metatensor

wyli:
### Description
remove the recursive deepcopy to speed up operations

this PR:
Screenshot 2022-07-16 at 13 27 09

dev branch:
Screenshot 2022-07-16 at 06 57 36

0.9.0:
Screenshot 2022-07-16 at 07 08 34

### Status

### Types of changes

### DIFF
--- a/monai/data/meta_obj.py
+++ b/monai/data/meta_obj.py
@@ -11,6 +11,7 @@
 
 from __future__ import annotations
 
+import time, pickle
 import itertools
 import pprint
 from copy import deepcopy
@@ -126,7 +127,8 @@ class MetaObj:
         found = [False] * len(attributes)
         for i, (idx, a) in itertools.product(input_objs, enumerate(attributes)):
             if not found[idx] and hasattr(i, a):
-                setattr(self, a, MetaObj.copy_items(getattr(i, a)) if deep_copy else getattr(i, a))
+                attr = getattr(i, a)
+                setattr(self, a, MetaObj.copy_items(attr)) if deep_copy else attr
                 found[idx] = True
             if all(found):
                 return
@@ -141,7 +143,7 @@ class MetaObj:
         if isinstance(data, (int, float, str, bool, Enum)):
             return data
         if isinstance(data, np.ndarray):
-            return np.array(data, copy=True)
+            return data.copy()
         if isinstance(data, torch.Tensor):
             return data.detach().clone()
         if isinstance(data, (list, tuple)):
--- a/monai/data/meta_obj.py
+++ b/monai/data/meta_obj.py
@@ -103,33 +103,30 @@ class MetaObj:
             elif isinstance(a, MetaObj):
                 yield a
 
-    def _copy_attr(self, attributes: list[str], input_objs) -> None:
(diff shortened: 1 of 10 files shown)


monai/data/utils.py
```diff
@@ -588,11 +588,11 @@ def decollate_batch(batch, detach: bool = True, pad=True, fill_value=None):
         if isinstance(batch, MetaObj):
             for t, m in zip(out_list, decollate_batch(batch.meta)):
                 if isinstance(t, MetaObj):
-                    t.meta = m
+                    t.meta = deepcopy(m)
```
Nic-Ma, line 591: May I know why add `deepcopy` here? I think the decollated items of `batch.meta` should be dispatched to every item in `out_list`? Or Maybe I misundertand this logic here.

Thanks.

monai/data/meta_obj.py
```diff
@@ -103,71 +103,30 @@ def flatten_meta_objs(*args: Iterable):
...
         """
-        Copy metadata from an iterable of `MetaObj` instances. For a given attribute, we copy the
-        adjunct data from the first element in the list containing that attribute.
+        Copy metadata from a `MetaObj` or an iterable of `MetaObj` instances.
 
         Args:
             input_objs: list of `MetaObj` to copy data from.
-
+            copy_attr: whether to copy each attribute with `MetaObj.copy_item`.
+                note that if the attribute is a nested list or dict, only a shallow copy will be done.
```
Nic-Ma, line 122: Seems even if `copy_attr = True`, it still shallow copy the numpy array or tensor?
So if I really want to deepcopy, can I use `a = MetaTensor(...), MetaTensor b = deepcopy(a)` or `b.meta = deepcopy(a.meta)`?

Thanks.

wyli: that's a good point, now it's updated to deepcopy in two cases when a is a MetaTensor, `b = MetaTensor(a)` and `b = a.clone()`, what do you think?

monai/data/utils.py
wyli: this is to ensure all the samples in a batch don't share the meta info (to be consistent with applied_operations). I think it's robust, and doesn't require much cpu time as it's done only once per tensor.

monai/data/meta_obj.py
Nic-Ma: OK, the solution looks good to me. Please document it clearly, I feel it's easy to be confused..

Thanks.

monai/data/utils.py
wyli: thanks -- that's removed in the latest commit

Nic-Ma (approved):

rijobro:
Hi, it seems like the difference in your screenshot is around 0.1s  between slowest and fastest (1.93s vs 2.04s). Are we sure this isn't random fluctuations?

wyli:
Yes @rijobro, I run `python tests/test_integration_fast_train.py` the total time is 30s vs 20s with my workstations, it's also the case for the relevant demos: https://github.com/Project-MONAI/tutorials/pull/786#issuecomment-1186073996.

PR #4732, opened 2022-07-20
add user warnings in AsChannelFirst and AddChannel

KumoLiu:
Fixes # .

### Description
add user warnings in `AsChannelFirst` and `AddChannel` to recommend users use `EnsureChannelFirst`.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/utility/array.py
+++ b/monai/transforms/utility/array.py
@@ -112,8 +112,8 @@ class Identity(Transform):
         return img
 
 
+@deprecated(since="0.9", msg_suffix="use monai.transforms.EnsureChannelFirst instead.")
 class AsChannelFirst(Transform):
-    warnings.warn('recommend to use `EnsureChannelFirst` to ensure that the channel is in the first dimension')
     """
     Change the channel dimension of the image to the first dimension.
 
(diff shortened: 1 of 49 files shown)


monai/transforms/utility/array.py
```diff
@@ -112,6 +112,7 @@ def __call__(self, img: NdarrayOrTensor) -> NdarrayOrTensor:
         return img
 
 
+@deprecated(since="0.9", msg_suffix="use monai.transforms.EnsureChannelFirst instead.")
```
Nic-Ma, line 115: Hi @wyli ,

I think the `since` should be `0.8`, could you please help confirm it?

Thanks in advance.

Nic-Ma (approved):

monai/transforms/utility/array.py
Nic-Ma: About the message can be:
msg_suffix="use `MetaTensor` data type and `monai.transforms.EnsureChannelFirst` instead."
@wyli What do you think?

Thanks.

wyli: I think 0.9 here is fine because it's a warning for 0.9.1... Do you think there are use cases that addchannel can achieve but ensure channel first can't, should we mention it?

Nic-Ma: Hi @wyli ,

`AddChannel` can work for PyTorch Tensor and numpy array directly while `EnsureChannelFirst` can't, I am not sure whether users still want to use it after we introduced 0.9.1 with `MetaTensor`...

Thanks.

wyli: ok thanks,then I think this PR is ready, I'll merge it.

Nic-Ma:
Hi @wyli ,

Should we also deprecate the dict version transforms?

Thanks.

wyli:
@Nic-Ma I think calling the dictionary version should alreadyraise the array level warning?

PR #4768, opened 2022-07-26
Implement GridSplit using torch.as_strided

### LINKED ISSUE #4286: `GridSplit` torch vs numpy performance
`GridSplit` has two separate implementations for `torch.Tensor` and `numpy.ndarray`. Although the torch implementation using `unfold` is specific to torch, both should be able to use the numpy implementation using `as_strided`. Thus, it would be nice to compare the performance of these two implementation and if `as_strided` is better, we can unify the implementation. Other factor to consider would be ability to expand to tensors with three spatial dimensions.

drbeh:
Fixes #4286 

### Description
This PR implements `GridSplit` using `torch.as_strided` instead of `torch.unfold`, which provides significant speedup.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -2989,7 +2989,10 @@ class GridSplit(Transform):
         # Flatten the first two dimensions
         strided_image = strided_image.reshape(-1, *strided_image.shape[2:])
         # Make a list of contiguous patches
-        patches = [np.ascontiguousarray(p) for p in strided_image]
+        if isinstance(image, torch.Tensor):
+            patches = [p.contiguous() for p in strided_image]
+        elif isinstance(image, np.ndarray):
+            patches = [np.ascontiguousarray(p) for p in strided_image]
 
         return patches
 
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -15,11 +15,10 @@ https://github.com/Project-MONAI/MONAI/wiki/MONAI_Design
 import warnings
 from copy import deepcopy
 from enum import Enum
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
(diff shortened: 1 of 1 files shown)


KumoLiu (approved):

Nic-Ma:
Hi @drbeh ,

Thanks for the investigation and test.
I see some related discussion in PyTorch, `torch.as_strided` returns the view of the original Tensor, should we test again with `contiguous()`? PyTorch collate_fn needs the data to be contiguous.
https://github.com/pytorch/pytorch/issues/60466#issuecomment-992969680

Thanks.

monai/transforms/spatial/array.py
```diff
@@ -2970,31 +2969,28 @@ def __call__(
...
+        n_channels = image.shape[0]
+        strided_image = as_strided_func(
+            image,
+            (*self.grid, n_channels, split_size[0], split_size[1]),
+            (x_stride * x_step, y_stride * y_step, c_stride, x_stride, y_stride),
+        )
+        # Flatten the first two dimensions
+        strided_image = strided_image.reshape(-1, *strided_image.shape[2:])
+        # Make a list of contiguous patches
+        patches = [np.ascontiguousarray(p) for p in strided_image]
```
wyli, line 2992: @Nic-Ma I think ascontiguous is done here before returning `patches`  but the issue is `np.ascontiguousarray` won't work for cuda torch tensors.

drbeh: That's right, @Nic-Ma! the contiguous is already being done for both.
Good catch, @wyli! Thanks, I fixed it and it seems to be faster now.
```
> numpy:
2.3 ms ± 385 µs per loop (mean ± std. dev. of 10 runs, 100 loops each)
> torch:
296 µs ± 65.1 µs per loop (mean ± std. dev. of 10 runs, 100 loops each)
```

Nic-Ma:
Looks good to me now, thanks for the update.

PR #4792, opened 2022-07-29
4791 Adapt affine arg to the Spacing transform

### LINKED ISSUE #4791: Unused 'affine' argument in Spacing transform
**Describe the bug**
The `Spacing` transform accepts an `affine` argument, but  the argument is unused by the `__call__` function and has no effect on the output. 

**To Reproduce**

```python
from monai import transforms as trans

def test_monai_utils():

    # fake data
    x = np.ones((1, 64, 64, 64))

    # fake affines
    affine_1 = np.eye(4)
    affine_2 = np.eye(4) * 2

    resamp = trans.Spacing(pixdim=[1, 1, 1], diagonal=False, mode='bilinear')

    # using MetaTensor works as expected
    y_1 = resamp(MetaTensor(x, affine=affine_1))
    y_2 = resamp(MetaTensor(x, affine=affine_2))
    assert not all(i == j for i, j in zip(y_1.shape, y_2.shape))

    # using affine has no effect
    y_1 = resamp(x, affine=affine_1)
    y_2 = resamp(x, affine=affine_2)

    # these shapes should be different!
    assert not all(i == j for i, j in zip(y_1.shape, y_2.shape))
```

**Expected behavior**
Either have the `affine` argument override the input `data_array.affine`, or remove the `affine` argument altogether in favor of `data_array.affine`.

Nic-Ma:
Fixes #4791 .

### Description
This PR adapted the deprecated `affine` arg to the Spacing transform.
Similar to the AsDiscrete transform:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/post/array.py#L244

### Status
**Ready**

### Types of changes


monai/transforms/spatial/array.py
```diff
@@ -515,15 +515,18 @@ def __call__(
...
         if sr <= 0:
             raise ValueError("data_array must have at least one spatial dimension.")
+        input_affine: NdarrayOrTensor
         affine_: np.ndarray
-        affine_np: np.ndarray
-        if isinstance(data_array, MetaTensor):
-            affine_np, *_ = convert_data_type(data_array.affine, np.ndarray)
-            affine_ = to_affine_nd(sr, affine_np)
+        if affine is not None:
+            warnings.warn("`affine` argument is deprecated, please use the affine of MetaTensor in data_array.")
```
wyli, line 521: Shall we remove the 'affine' arg in this method for v1.0? It's deprecated and will generate the warning message already via 'deprecated_arg'. If we keep it, I hope metatensor.affine would have a higher priority.

Nic-Ma: I think we should remove this arg in 1.0, I add the functionality back here just because it's important argument and may help users smoothly migrate to MetaTensor.
Updated the priority in the latest commit.

Thanks.

wyli: So this is only temporarily for the users installing from the dev branch before we remove the args in v1.0? I feel this might cause more confusions...

wyli (approved):
The implementati logic is fine but we may discuss whether to remove the deprecated args in the upcoming release.

monai/transforms/spatial/array.py
Nic-Ma: Hi @wyli ,

I think it's for the weekly build releases.
If more users face this problem, at least we can suggest them to pip install the weekly build and then may migrate to MetaTensor if they are interested.
What do you think?

Thanks.

wyli: so why not recommend using the metatensor to store the affine instead of installing the weekly build? the metatensor usage will be the v1.0 way to do it https://github.com/Project-MONAI/MONAI/issues/4791#issuecomment-1198920001

Nic-Ma: OK, let me set up an offiline discussion on Monday.

Thanks.

PR #4793, opened 2022-07-29
Update patch merging with a for loop

function2-llx:
Fixes #4735

### Description
Update the patch merging process with a for loop to make it less error-prone and more readable.

### Status
**Ready**

### Types of changes


function2-llx:
I've come up with one issue: in case someone's using 2D Swin Transformer pre-trained model, I've swapped the order of `i` and `j` in the code. It's unnecessary to do the same for 3D since original implementation should be used for existed pre-trained model.

wyli (approved):

PR #4825, opened 2022-08-03
Add Time Profiler Class

ericspod:
### Description
This adds a time profile class to measure how much recorded time various components of workflows consume.  There is a [tutorial notebook here](https://github.com/ericspod/tutorials/blob/profiling_tutorial/modules/workflow_profiling.ipynb) that will be PR'd to the tutorials repo once this is accepted. 

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/utils/profiling.py
+++ b/monai/utils/profiling.py
@@ -154,7 +154,7 @@ class WorkflowProfiler:
     This profiler must be used only within its context because it uses an internal thread to read results from a
     multiprocessing queue. This allows the profiler to function across multiple threads and processes, though the
     multiprocess tracing is at times unreliable and not available in Windows at all.
-    
+
     The profiler uses `sys.settrace` and `threading.settrace` to find all calls to profile, this will be set when
     the context enters and cleared when it exits so proper use of the context is essential to prevent excessive
     tracing. Note that tracing has a high overhead so times will not accurately reflect real world performance
--- a/monai/utils/profiling.py
+++ b/monai/utils/profiling.py
@@ -323,7 +323,7 @@ class WorkflowProfiler:
         """Wrapper around anything iterable to profile how long it takes to generate items."""
 
         class _Iterable:
-            def __iter__(_self):  # noqa: B902, N805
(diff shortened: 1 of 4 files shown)


Nic-Ma:
Add @binliunls as he was doing something similar.
@binliunls could you please help share the GPU time measure tutorial from TensorRT and review this PR?

Thanks in advance.

monai/utils/profiling.py
```diff
@@ -102,9 +125,294 @@ def __init__(self):
...
+    infinite loops and would be terribly slow even if not. This selector is a callable accepting a `call` trace
+    frame and returns True if the call should be traced. The dedault is `select_transform_call` which will return
+    True for `Transform.__call__` calls only.
+
+    Example showing use of all profiling functions:
+
+    .. code-block:: python
+
+        import monai.transform as mt
+        comp=mt.Compose([mt.ScaleIntensity(),mt.RandAxisFlip(0.5)])
```
wyli, line 171: cool example, it works fine but requires these imports as I test it:
```py
import monai.transforms as mt
from monai.utils import WorkflowProfiler
import torch
```

ericspod: Added in.

binliunls:
> 

Ok, I am glad to review this PR. 
And the link below is the GPU time measure tutorial from TensorRT on a classification task using efficientnet.
https://pytorch.org/TensorRT/_notebooks/EfficientNet-example.html

ericspod:
There's something really odd about multiprocess behaviour in Darwin and in one the Nvidia images used for testing (nvcr.io/nvidia/pytorch:22.07-py3). I tried reproducing with Pytorch 1.13 but the test passes locally on Ubuntu 20.04. I think I need to disable multiprocess tracing test and state it's unreliable.

wyli (approved):
Thanks, it looks good to me.

ericspod:
Hi @binliunls we merged this PR but any feedback you have on this would still be great. I had avoided doing profiling of CPU, GPU, or memory since there's the Pytorch profiler for that and TensorRT.

PR #4838, opened 2022-08-05
4848 849 4718 misc updates - adds device tests, docstrings

### LINKED ISSUE #4718: review `img.meta['affine']` and `img.affine`
writing the affine could currently be achieved with two methods:
`img.meta['affine']=matrix` and `img.affine=matrix`, the property setter is safer with an additional type constraint.

existing usages should be reviewed and made consistent.

### LINKED ISSUE #4848: removed -- optional argument in `deprecated_arg`
the docstring: 

> removed: version at which the argument was removed and no longer usable.

should be more like:

> remove: version at which the argument **will be** removed and no longer usable.

_Originally posted by @rijobro in https://github.com/Project-MONAI/MONAI/issues/233#issuecomment-1203772856_

wyli:
- unit tests part of https://github.com/Project-MONAI/tutorials/issues/849
- fixes #4848
- fixes #4718 

### Description
- fixes rocauc
- adds unit tests

### Status
**Ready**

### Types of changes


monai/metrics/rocauc.py
```diff
@@ -81,7 +81,7 @@ def _calculate(y_pred: torch.Tensor, y: torch.Tensor) -> float:
         return float("nan")
 
     n = len(y)
-    indices = y_pred.argsort()
+    indices = y_pred.argsort().cpu()
     y = y[indices].cpu().numpy()
```
Nic-Ma, line 85: I think the problem is that why `y` and `y_pred` are not on the same device in the latest tutorial?
If both are on GPU, `y[indices]` may be faster than CPU indices? So I think we should change the tutorial instead of here?

Thanks.

wyli: sure, I revert the change here

Nic-Ma (approved):
Looks good to me now.

Thanks.

PR #4842, opened 2022-08-05
Change to EnsureChannelFirst to include behaviour from deprecated transforms

ericspod:

### Description
`EnsureChannelFirst` should be used in place of `AddChannel`, however it requires that metadata be provided as MetaTensor inputs or separate dictionaries. This doesn't substitute for `AddChannel` which required neither of these things and so can't be used in the situation where one has data known to have no channel dimension nor metadata and we just want to hard-code adding a 1-sized channel dimension. This is useful in a lot of examples that use generated data and not loaded files that come with metadata, and avoids forcing the user to use MetaTensor.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/utility/array.py
+++ b/monai/transforms/utility/array.py
@@ -245,12 +245,12 @@ class EnsureChannelFirst(Transform):
                     raise ValueError(msg)
                 warnings.warn(msg)
                 return img
-            meta_dict["original_channel_dim"] = channel_dim = "no_channel"
+            meta_dict["original_channel_dim"] = channel_dim = "no_channel"   # type: ignore
 
         if channel_dim == "no_channel":
             result = img[None]
         else:
-            result = moveaxis(img, channel_dim, 0)
+            result = moveaxis(img, channel_dim, 0)   # type: ignore
 
         return convert_to_tensor(result, track_meta=get_track_meta())  # type: ignore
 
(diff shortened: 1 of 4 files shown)


rijobro (approved):

monai/transforms/utility/array.py
```diff
@@ -202,43 +202,57 @@ def __call__(self, img: NdarrayOrTensor) -> NdarrayOrTensor:
...
     backend = [TransformBackends.TORCH, TransformBackends.NUMPY]
 
-    def __init__(self, strict_check: bool = True):
-        """
-        Args:
-            strict_check: whether to raise an error when the meta information is insufficient.
-        """
+    def __init__(self, strict_check: bool = True, add_channel_default=True):
         self.strict_check = strict_check
+        self.add_channel_default = add_channel_default
```
KumoLiu, line 220: I think the argument name `add_channel_default` may be a little confused, perhaps name such as  `force_add_channel` maybe more  easier to understand.

wyli: I think it provides a default value only if channel_dim is None, so it's not forcing the value...

ericspod: It doesn't force adding the channel no, but if that was desired for some reason we should undeprecate `AddChannel`.

PR #4843, opened 2022-08-05
Adding __str__ to MetaTensor

### LINKED ISSUE #4840: MetaTensor Needs `__str__` Method
When a `MetaTensor` is printed with `print` it produces its metadata information in addition to what a tensor would normally print. This is usually too verbose for just printing to screen, and differs from default behaviour from Pytorch. The design intent was to not require users to have to deal with MetaTensors, it's meant to be opt-in, but this change has a visible difference the unexpecting user would have to dig into to solve. 

Add a `__str__` method to `MetaTensor` to preserve the printing behaviour of regular tensors but leave `__repr__` for those wanting larger outputs. This can be prototyped with:

```python
class MetaTensor1(monai.data.MetaTensor):
    def __str__(self):
        return str(self.as_tensor())
    
t=torch.rand(10)
print(t.min())  # expected
print(monai.data.MetaTensor(t).min())  # too verbose
print(MetaTensor1(t).min())  # as expected
```

The alternative to get default behaviour is `print(t.as_tensor())` but that is now forcing users to opt-in to MetaTensor and makes code non-portable, ie. won't work with regular Pytorch tensors if used somewhere without MONAI

**Additional context**
This [comment](https://stackoverflow.com/a/2626364/2788477) summarises usage of `__str__` and `__repr__`. I think this approach makes `__str__` readable and leaves `__repr__` unambiguous as it is now.

ericspod:

Fixes #4840 .

### Description
Add `__str__` to `MetaTensor` to keep similar behavior with Pytorch tensors.

### Status
**Ready**

### Types of changes


tests/test_meta_tensor.py
```diff
@@ -422,21 +422,7 @@ def test_decollate(self, dtype):
...
-            + "\taffine: 1\n"
-            + "\tspace: RAS\n"
-            + "\n"
-            + "Applied operations\n"
-            + "[]\n"
-            + "Is batch?: False"
-        )
-        for s in (s1, s2):
-            self.assertEqual(s, expected_out)
+        self.assertEqual(str(t), "tensor([1.])")
```
wyli, line 425: thanks, I forgot that the tutorial relies on this to print some meta information https://colab.research.google.com/drive/1T4iAys-cC2qL80oJkIbAXAPlWNPwp4H7#scrollTo=hQv51ORy6_oV, so what would be your recommendation of print those? cc @rijobro

ericspod: In the tutorials we can do `print(repr(t))` to get the same behaviour.

rijobro (approved):
Do you think you could make the relevant change in the tutorial? Otherwise, all good. Thanks!

ericspod:
I can get to the notebook. Which one is it? The colab link doesn't say where it is in the repo.

wyli:
I'll update the tutorial https://colab.research.google.com/drive/1T4iAys-cC2qL80oJkIbAXAPlWNPwp4H7

ericspod:
The way tensors are printed was [changed in Pytorch 1.12](https://github.com/pytorch/pytorch/blame/v1.12.0/torch/_tensor_str.py#L312) which needs to be taken into account in the tests.

wyli:
with this PR, print(metatensor) will output exactly the same string as print(tensor)...we lose all the meta info in the string, without this PR, print(metatensor) is verbose, any solution in the middle cc @Nic-Ma @rijobro

Nic-Ma:
Hi @ericspod @wyli ,

Actually, I still don't quite understand the target goal of this PR, usually the output of `__str__` is for human reading, just in the format of a regular string, why not add some meta information in it?

Thanks.

wyli:
> Hi @ericspod @wyli ,
> 
> Actually, I still don't quite understand the target goal of this PR, usually the output of `__str__` is for human reading, just in the format of a regular string, why not add some meta information in it?
> 
> Thanks.

Hi Nic, the issue is clearly described in #4840, the pr is trying to address 4840.

Nic-Ma:
Hi @wyli ,

I checked the original ticket this morning, I don't quite understand it...
Especially: "The design intent was to **not require users to have to deal with MetaTensors**, it's meant to be opt-in, but this change has a visible difference the unexpecting user would have to dig into to solve."
I think we want to highlight `MetaTensor` in MONAI release and let users to use it, why hide the meta information in the `print`?

Thanks.

ericspod:
The issue is that a lot of code will have print statements using tensor directly, with normal Pytorch tensors this isn't a problem but with MetaTensor it's way too verbose. It'll break some people's code I image who rely on output being in a specific format. It's just too surprising to use code that does known things in Pytorch then all of a sudden with MONAI produces a huge output with information that isn't likely relevant at the time. Users will be confused about where the metadata stuff came from, we want people to not have to deal with MetaTensor if they don't want but this forces them to do just that.

ericspod:
> Hi @ericspod ,
> 
> OK, I see your point. Could you please help add clear doc-string to the `__str__` function and `__repr__` function to let users or developers know the usage?
> 
> Thanks.

Done.

Nic-Ma (approved):
Thanks for the quick update.
Looks good to me.

PR #4857, opened 2022-08-08
233 - set up a backward compatibility policy

### LINKED ISSUE #233: Set up an explicit backward compatibility policy

wyli:

Fixes #233

### Status
**Ready**

### Types of changes


CONTRIBUTING.md
```diff
@@ -283,10 +283,21 @@ for example, ``import monai.transforms.Spacing`` is the equivalent of ``monai.tr
...
+Utility functions are provided in `monai.utils.deprecated` to help migrate from the deprecated to new APIs. The use of these utilities is encouraged.
+The pull request [template contains checkboxes](https://github.com/Project-MONAI/MONAI/blame/dev/.github/pull_request_template.md#L11-L12) that
+the contributor should use accordingly to clearly indicate breaking changes.
+
+The process of releasing backwards incompatible API changes is as follows:
+1. discuss the breaking changes during pull requests or in dev meetings with a feature proposal if needed.
+1. add a warning message in the upcoming release (version `X.Y`), the warning message should include a forecast of removing the deprecated API in:
+   1. `X+1.0` -- major version `X+1` and minor version `0` the next major version if it's a significant change,
+   1. `X.Y+2` -- major version `X` and minor version `Y+2` (the minor version after the next one), if it's a minor API change.
+   1. Note that the versioning policy is similar to PyTorch's approach, however, does not follow [the semantic versioning](https://semver.org/).
```
ericspod, line 296: "Note that the versioning policy is similar to PyTorch's approach, however, does not follow [the semantic versioning](https://semver.org/)."
->
"Note that the versioning policy is similar to PyTorch's approach which does not precisely follow [the semantic versioning](https://semver.org/) definition. Major version numbers are instead used to represent major product version (which is not anticipated to be anything other an 1), minor version for both compatible and incompatible, and patch version for bug fixes."

CONTRIBUTING.md
```diff
@@ -283,10 +283,21 @@ for example, ``import monai.transforms.Spacing`` is the equivalent of ``monai.tr
...
+1. add new test cases by extending the existing unit tests to cover both the deprecated and updated APIs.
+1. collect feedback from the users during the subsequent few releases, and reconsider step 1 if needed.
+1. before each release, review the deprecating APIs, and clean up the removed APIs described in step 2.
+
```
ericspod, line 300: We should add some note when referring to another construct to use in place of a deprecated construct. In this case the referred to construct needs to provide exact feature-like behaviour otherwise users will have a harder time migrating.

CONTRIBUTING.md
```diff
@@ -283,10 +283,25 @@ for example, ``import monai.transforms.Spacing`` is the equivalent of ``monai.tr
...
+   1. `X+1.0` -- major version `X+1` and minor version `0` the next major version if it's a significant change,
+   1. `X.Y+2` -- major version `X` and minor version `Y+2` (the minor version after the next one), if it's a minor API change.
+   1. Note that the versioning policy is similar to PyTorch's approach which does not precisely follow [the semantic versioning](https://semver.org/) definition.
+      Major version numbers are instead used to represent major product version (which is currently not planned to be greater than 1),
+      minor version for both compatible and incompatible, and patch version for bug fixes.
+   1. when recommending new API to use in place of a deprecated API, the recommended version should
+      provide exact feature-like behaviour otherwise users will have a harder time migrating.
+1. add new test cases by extending the existing unit tests to cover both the deprecated and updated APIs.
+1. collect feedback from the users during the subsequent few releases, and reconsider step 1 if needed.
+1. before each release, review the deprecating APIs, and clean up the removed APIs described in step 2.
```
Nic-Ma, line 303: Maybe add: "and also remove the associated deprecated test cases"?

Thanks.

Nic-Ma (approved):
Looks good to me.
Put minor comments inline.

Thanks.

PR #4873, opened 2022-08-09
4867 flexible torchvision fc models

### LINKED ISSUE #4867: Question about TorchVisionFC model
### Discussed in https://github.com/Project-MONAI/MONAI/discussions/4865

<sup>Originally posted by **ctestagrose** August  8, 2022</sup>
I am interested in using the TorchVisionFC model. I see that the default model is resnet. Is there any way to use another torchvision model that is specified https://pytorch.org/vision/stable/models.html ? I would like to replace resnet with inception v3. 

When I replace resnet with inception v3 I am receiving the following error:

```
RuntimeError: Expected 4-dimensional input for 4-dimensional weight [192, 768, 1, 1], but got 2-dimensional input of size [16, 1000] instead
```
I am thinking this is an issue with either the way I am loading the data or possibly a misinterpretation on my part in regard to the way the TorchVisionFC model works.

Thank you, I am extremely appreciative of any help.
</div>

wyli:

Fixes #4867 

### Description
- enhance the `model.fc` replacement in netadapter
- tested inception_v3 as a torchvision fc model

### Status
**Ready**

### Types of changes


monai/networks/nets/netadapter.py
```diff
@@ -71,33 +71,42 @@ def __init__(
             in_channels_ = in_channels
 
         if pool is None:
-            # remove the last layer
-            self.features = torch.nn.Sequential(*layers[:-1])
+            # remove the last layer or replace it with an identity
+            if hasattr(model, "fc"):  # assuming fc is the last layer
```
ericspod, line 75: Is there any value in making the name a constructor argument rather than hard-coding "fc", or is this class used only in situations that would follow this pattern?

vikashg:
So @wyli, This patch works on models like Inception_V3 and produces the desired output. However, it fails for models like VisionTransformer `torchnvision.models.vit_b_16` with the following error
```
  File "/raid/Vikash/Tools/review-MONAI/MONAI/monai/networks/nets/torchvision_fc.py", line 91, in __init__
    raise ValueError(f"Model ['{model_name}'] does not have a Linear layer at the end.")
ValueError: Model ['vit_b_16'] does not have a Linear layer at the end.

```
The error is obvious as there is the startswith Linear condition check. It might work for Inception and Resnet as the tail of the model looks like this 
```
  )
  (avgpool): AdaptiveAvgPool2d(output_size=(1, 1))
  (dropout): Dropout(p=0.5, inplace=False)
  (fc): Linear(in_features=2048, out_features=1000, bias=True)
)
```
But does not work if the model terminates in a Linear layer wrapped in a Sequential module as follows
```
          (4): Dropout(p=0.0, inplace=False)
        )
      )
    )
    (ln): LayerNorm((768,), eps=1e-06, elementwise_affine=True)
  )
  (heads): Sequential(
    (head): Linear(in_features=768, out_features=1000, bias=True)
  )
)
```
So, if we can expand the `if ` statement it will cover majority of the models in `torchvision.models`

vikashg (changes requested):
When I try to use the VIT (`torchvision.models.vit_b_16`), I get the following error after the recent commit 
```
  File "test.py", line 6, in <module>
    model = TorchVisionFCModel(
  File "/raid/Vikash/Tools/review-MONAI/MONAI/monai/utils/deprecate_utils.py", line 224, in _wrapper
    return func(*args, **kwargs)
  File "/raid/Vikash/Tools/review-MONAI/MONAI/monai/networks/nets/torchvision_fc.py", line 93, in __init__
    if look_up_named_module(fc_name, model) is None:
  File "/raid/Vikash/Tools/review-MONAI/MONAI/monai/networks/utils.py", line 65, in look_up_named_module
    name_str = look_up_option(name, {n[0] for n in mod.named_modules()}, print_all_options=print_all_options)
  File "/raid/Vikash/Tools/review-MONAI/MONAI/monai/utils/module.py", line 120, in look_up_option
    raise ValueError(
ValueError: By 'fc', did you mean ''?
'fc' is not a valid value.
```

wyli:
thanks @vikashg the comments are helpful, I spend some time on this, it's not easy to be that flexible at the same time backwards compatible... but for the basic use cases it's ok now:

```py
import torch
from torchvision.models import Swin_T_Weights, vit_b_16

from monai.networks.nets import TorchVisionFCModel

model = TorchVisionFCModel(
    "vit_b_16", num_classes=4, use_conv=False, pool=None, in_channels=768, fc_name="heads.head")
output = model.forward(torch.randn(2, 3, 224, 224))
print(output.shape)  # torch.Size([2, 4])

model = TorchVisionFCModel(
    "swin_t",
    num_classes=4,
    pool="",
    use_conv=True,
    in_channels=768,
    node_name="permute",
    weights=Swin_T_Weights.IMAGENET1K_V1,
)
output = model.forward(torch.randn(2, 3, 500, 500))
print(output.shape)  # torch.Size([2, 4, 16, 16])

```

vikashg (approved):
Thanks @wyli for supplying an example code snippet. It all works well now. IT should be compatible with 0.9 right ?

wyli:
> Thanks @wyli for supplying an example code snippet. It all works well now. IT should be compatible with 0.9 right ?

yes it's non-breaking, the previous test cases all work fine.

ctestagrose:
Will these changes be available if we install monai via pip 0.8.1? Because we are using NVFlare for federated learning following this [tutorial](https://github.com/ctestagrose/tutorials/tree/main/federated_learning/breast_density_challenge) which uses `pip install monai=0.8.1` to build the docker packages. I wonder if these changes will be pushed to the pip package manager.

wyli:
Hi @ctestagrose, there's currently no plan to backport this to v0.8, it will be available in the weekly dev release https://pypi.org/project/monai-weekly/#history (next one on 14th/Aug). So, early next week, you can uninstall monai with `pip uninstall -y monai` and run `pip install monai-weekly` to get this feature. This version will be very different from v0.8.1 https://github.com/Project-MONAI/MONAI/compare/0.8.1...dev including some breaking changes so the existing 0.8.1 pipelines may or may not work.

Issue #4754, opened 2022-07-23
Why does get_default_applied_operations for MetaObj only return an empty list?


ltsc256:
I have been getting this error (see below) and found that the error was caused by an empty list for all_transforms. Looking at the code in inverse.py, the problem was in line 186 where all_transforms = data.get(self.trace_key(key), MetaTensor.get_default_applied_operations()). But get_default_applied_operations() only returns an empty list according to the source code provided here: https://docs.monai.io/en/stable/_modules/monai/data/meta_obj.html#MetaObj.get_default_applied_operations

How can I resolve this issue? (Config also pasted below)

Traceback (most recent call last):
  File "/home/user1/monai/MONAI/monai/transforms/transform.py", line 90, in apply_transform
    return _apply_transform(transform, data, unpack_items)
  File "/home/user1/monai/MONAI/monai/transforms/transform.py", line 54, in _apply_transform
    return transform(parameters)
  File "/home/user1/monai/MONAI/monai/transforms/utility/dictionary.py", line 523, in inverse
    self.pop_transform(d, key)
  File "/home/user1/monai/MONAI/monai/transforms/inverse.py", line 210, in pop_transform
    return self.get_most_recent_transform(data, key, check, pop=True)
  File "/home/user1/monai/MONAI/monai/transforms/inverse.py", line 192, in get_most_recent_transform
    self.check_transforms_match(all_transforms[-1])
IndexError: **list index out of range**

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/user1/monai/MONAI/monai/transforms/transform.py", line 90, in apply_transform
    return _apply_transform(transform, data, unpack_items)
  File "/home/user1/monai/MONAI/monai/transforms/transform.py", line 54, in _apply_transform
    return transform(parameters)
  File "/home/user1/monai/MONAI/monai/transforms/post/dictionary.py", line 664, in __call__
    inverted = self.transform.inverse(input_dict)
  File "/home/user1/monai/MONAI/monai/transforms/compose.py", line 183, in inverse
    data = apply_transform(t.inverse, data, self.map_items, self.unpack_items, self.log_stats)
  File "/home/user1/monai/MONAI/monai/transforms/transform.py", line 114, in apply_transform
    raise RuntimeError(f"applying transform {transform}") from e
RuntimeError: applying transform <bound method ToTensord.inverse of <monai.transforms.utility.dictionary.ToTensord object at 0x7f04da6b8cd0>>

[...]

Issue #1333, opened 2020-12-09
The prediction has different direction with the original val dataset


GOZGOXGOYGO:
Hi,

I am using NiftiSaver to save the  prediction of val dataset. However, the predict result has different direction with the original val dataset. Could you please help me find what is wrong? My code is here. It is almost the same with the code in tutorial.

```
    monai.config.print_config() 
    logging.basicConfig(stream=sys.stdout, level=logging.INFO) 
    
    datadir = '../data/'
    images_val = sorted(glob(os.path.join(datadir, './*.nii.gz')))
    segs_val = sorted(glob(os.path.join(datadir, './*.nii.gz')))  
    print('validation num is ', len(images_val))
    val_files = [{"img": img, "seg": seg} for img, seg in zip(images_val, segs_val)]

    # define transforms for image and segmentation
    val_transforms = Compose(
        [
            LoadImaged(keys=["img", "seg"]),
            AddChanneld(keys=["img", "seg"]),
            ScaleIntensityd(keys="img"),
            ToTensord(keys=["img", "seg"]),
        ]
    )

    '''
    Validation
    '''    
    # create a validation data loader
    val_ds = monai.data.Dataset(data=val_files, transform=val_transforms)
    val_loader = DataLoader(val_ds, batch_size=1, num_workers=4, collate_fn=list_data_collate)
    dice_metric = DiceMetric(include_background=True, reduction="mean")
    post_trans = Compose([Activations(sigmoid=True), AsDiscrete(threshold_values=True)])
    
    # create UNet, DiceLoss and Adam optimizer
    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
    model = monai.networks.nets.UNet(
        dimensions=2,
        in_channels=1,
        out_channels=1,
        channels=(16, 32, 64, 128, 256),
        strides=(2, 2, 2, 2),
        num_res_units=2,
    ).to(device)
    model.load_state_dict(torch.load("./results/best_metric_model.pth"))
    
    
    model.eval()
    print('evaluating...')
    with torch.no_grad():
        metric_sum = 0.0
        metric_count = 0
        saver = NiftiSaver(output_dir="./val_predict")
        for val_data in val_loader:
            val_images, val_labels = val_data["img"].to(device), val_data["seg"].to(device)
            # define sliding window size and batch size for windows inference
            roi_size = (96, 96)
            sw_batch_size = 4
[...]

PR #4874, opened 2022-08-09
4754 default ToTensor.track_meta to None

wyli:

addresses
- #4754 
- https://github.com/Project-MONAI/MONAI/issues/1333#issuecomment-1208826068
- https://github.com/Project-MONAI/model-zoo/issues/66

### Description
given that `ToTensor` is very popular in early versions of MONAI, the current default value should be updated in favor of using metatensor.

### Status
**Ready**

### Types of changes


monai/transforms/utility/array.py
```diff
@@ -415,7 +415,7 @@ class ToTensor(Transform):
         device: target device to put the converted Tensor data.
         wrap_sequence: if `False`, then lists will recursively call this function, default to `True`.
             E.g., if `False`, `[1, 2]` -> `[tensor(1), tensor(2)]`, if `True`, then `[1, 2]` -> `tensor([1, 2])`.
-        track_meta: whether to convert to `MetaTensor`, default to `False`, output type will be `torch.Tensor`.
+        track_meta: whether to convert to `MetaTensor`, default to `None`, output type will be `torch.Tensor`.
```
Nic-Ma, line 418: I think the change of doc-string is not correct here, should be: default to `None`, use the return value of ``get_track_meta``.

Thanks.

Nic-Ma (commented):
Hi @wyli ,

In my previous PR, I initially planned to set the default value to `None`, but we discussed and aligned that: Set to `False` to be compatible with the previous versions and seems `ToTensor` defaults to PyTorch Tensor is reasonable.
Why you changed the idea? Default to `None` may break many programs.

Thanks.

wyli:
> Hi @wyli ,
> 
> In my previous PR, I initially planned to set the default value to `None`, but we discussed and aligned that: Set to `False` to be compatible with the previous versions and seems `ToTensor` defaults to PyTorch Tensor is reasonable. Why you changed the idea? Default to `None` may break many programs.
> 
> Thanks.

sure, `ToTensor` is used in many programs, if it defaults to regular tensor, all the postprocessings will not have the metadata. In my understanding, changing this default will only break one use case -- previously using pytorch's collate_fn should be replaced with monai's list_collate_fn, such as this line
https://github.com/Project-MONAI/MONAI/blob/6da25ae54996a7be8e06b9e23a18a25ea44d70f2/tests/test_deepedit_interaction.py#L65 apart from this, metatensor is a subclass of tensor, should work fine with the previous tensor based pipelines.

Nic-Ma:
Yes, I faced the same issue in the deepedit unit tests as you changed, so we decided to use `False` as default.
I am very sure about the idea now.. @ericspod @rijobro What do you guys think? Should we change the default?

Thanks in advance.

wyli:
Hi Nic, the error was incompatibility of metatensor + pytorch default_collate. our current solution is to convert it to regular tensors then use default collate. 

I'm suggesting that the solution should be metatensor + monai's collate which encourages the use of meta info in the deep learning pipeline. In my opinion in v1.0 this should be one of the most important concepts.

Nic-Ma (approved):
Let's try with this default first, if any unknown issue is reported later, we can revert it.
@ericspod @rijobro What do you think?

Thanks.

monai/transforms/utility/array.py
```diff
@@ -415,8 +415,8 @@ class ToTensor(Transform):
         device: target device to put the converted Tensor data.
         wrap_sequence: if `False`, then lists will recursively call this function, default to `True`.
             E.g., if `False`, `[1, 2]` -> `[tensor(1), tensor(2)]`, if `True`, then `[1, 2]` -> `tensor([1, 2])`.
-        track_meta: whether to convert to `MetaTensor`, default to `False`, output type will be `torch.Tensor`.
-            if `None`, use the return value of ``get_track_meta``.
+        track_meta: whether to convert to `MetaTensor` or regular tensor, default to `None`,
```
ericspod, line 418: I would clarify the description a bit: "track_meta: if True convert to `MetaTensor`, otherwise to Pytorch Tensor, if None behave according to return value of `get_track_meta`."

ericspod:
I'm good with this with a minor docstring comment, though I hope @rijobro had a chance to look at this.

PR #4888, opened 2022-08-10
4563 Enhance the stats information of workflows

### LINKED ISSUE #4563: Consider enhancing get_train_stats()
Currently, the Trainer's [get_train_stats](https://github.com/Project-MONAI/MONAI/blob/d34fa14eed3191126fc1bc1a9751e69e0396f78b/monai/engines/trainer.py#L60)() function is returning a fixed set of stats

```
    def get_train_stats(self) -> dict[str, float]:
        return {"total_epochs": self.state.max_epochs, "total_iterations": self.state.epoch_length}
```

Make it configurable true input arguments.

Alternatively, anyone needing train stats could access `trainer.state` directly. However, having a general `get_train_stats()` call might have more utilities.

**Additional context**
Training stats could be useful in contexts such as federated learning.

Nic-Ma:
Fixes #4563 .

### Description
This PR enhanced the stats information of trainer and evaluator workflows.

### Status
**Ready**

### Types of changes


monai/engines/trainer.py
```diff
@@ -57,8 +58,27 @@ def run(self) -> None:
         self.scaler = torch.cuda.amp.GradScaler() if self.amp else None
         super().run()
 
-    def get_train_stats(self) -> dict[str, float]:
-        return {"total_epochs": self.state.max_epochs, "total_iterations": self.state.epoch_length}
+    def get_train_stats(self, *others) -> dict[str, float]:
```
ericspod, line 61: Unless I'm out of the loop on why, I would call both methods just `get_stats` so that code can be written to use either class.

monai/engines/trainer.py
```diff
@@ -57,8 +58,27 @@ def run(self) -> None:
...
+
+        """
+        stats = {
+            StatsKeys.RANK: self.state.rank,
+            StatsKeys.CURRENT_EPOCH: self.state.epoch,
+            StatsKeys.CURRENT_ITERATION: self.state.iteration,
+            StatsKeys.TOTAL_EPOCHS: self.state.max_epochs,
+            StatsKeys.TOTAL_ITERATIONS: self.state.epoch_length,
+        }
+        for k in others:
```
ericspod, line 79: Furthermore I would put in `Workflow`:

````python 
def get_stats(self, *others):
    return {getattr(self.state, k, None) for k in others}
```

and call that in the overrides in `Trainer` and `Evaluator`.

monai/engines/trainer.py
Nic-Ma: Updated in the latest commit.

Thanks.

monai/engines/trainer.py
Nic-Ma: That's a good point, updated in the latest commit.

Thanks.

monai/engines/evaluator.py
```diff
@@ -146,7 +147,28 @@ def run(self, global_epoch: int = 1) -> None:
...
+        stats = {
+            ESKeys.RANK: self.state.rank,
+            ESKeys.BEST_VALIDATION_EPOCH: self.state.best_metric_epoch,
+            ESKeys.BEST_VALIDATION_METRTC: self.state.best_metric,
+        }
+        for k in vars:
+            stats[k] = getattr(self.state, k, None)
+        return stats
+
+    @deprecated(since="0.8", msg_suffix="please use the `get_stats()` API instead.")
```
wyli, line 170: this should be since 0.9?

Nic-Ma: Sure, updated in the latest commit.

Thanks.

holgerroth (dismissed):
LGTM

holgerroth (commented):
It looks good but I'm getting an error with the nvflare integration now:

```
  File "/home/hroth/Code2/nvflare/sim-project/yhwen-simulator/nvflare/app_common/pt/pt_file_model_persistor.py", line 199, in load_model
    data = self.model.state_dict() if self.model is not None else OrderedDict()
AttributeError: 'dict' object has no attribute 'state_dict'
```

Nic-Ma:
Hi @holgerroth ,

Thanks for your review and test.
I see the error message is in NVFlare, could you please help provide more information on MONAI side?

Thanks.

holgerroth:
The issue came from using an old UNet style config in the NVFlare configuration. 
```
>>> net = UNet(
...     dimensions=2,
...     in_channels=1,
...     out_channels=2,
...     channels=(16, 32, 64, 128, 256),
...     strides=(2, 2, 2, 2),
...     num_res_units=2,
...     norm="batch"
... )
TypeError: __init__() missing 1 required positional argument: 'spatial_dims'
```
Changing `dimensions` to `spatial_dims` solves the issue.

holgerroth (approved):

PR #4902, opened 2022-08-12
[WIP] 4901 keychain

### LINKED ISSUE #4901: Add methods to get all key chains, and convert it from/to list format in ConfigParser
Currently one can recursively get value from a chain of keys separated by special characters. 
For example, one can do the following
```python
cfg = ConfigParser(config)
key_chain = "key1#key2#key3"
cfg.get(key_chain)
```
The problem is format may not interface with other code/module well. One example is Auto3D which format key chain in a list ['key1', 'key2', 'key3']. It would be good to add some convenient functions in the ConfigParser class to support easy interface

Two new functions
1. convert a list of keys to the key chain
2. recursively find out all possible key_chains in the self.config of ConfigParser

mingxin-zheng:
Fixes #4901

### Description
Add functions in ConfigParser 

### Status
WIP

### Types of changes

### DIFF
--- a/monai/bundle/config_parser.py
+++ b/monai/bundle/config_parser.py
@@ -13,7 +13,7 @@ import json
 import re
 from copy import deepcopy
 from pathlib import Path
-from typing import Any, Dict, Optional, Sequence, Tuple, Union
+from typing import Any, Dict, Optional, Sequence, Tuple, Union, List
 
 from monai.bundle.config_item import ComponentLocator, ConfigComponent, ConfigExpression, ConfigItem
 from monai.bundle.reference_resolver import ReferenceResolver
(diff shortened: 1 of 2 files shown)


monai/bundle/config_parser.py
```diff
@@ -463,3 +463,50 @@ def resolve_relative_ids(cls, id: str, value: str) -> str:
...
+
+        Return:
+            a formated string
+
+        """
+        id_str = ""
+        for key in key_chain:
+            id_str += key + ID_SEP_KEY
+        id_str = id_str[:-1]
+        return id_str
```
wyli, line 484: I think this could be `ID_SEP_KEY.join(key_chain)`?

Nic-Ma:
Hi @mingxin-zheng @dongyang0122 @wyli ,

I have a question about the feature in this PR:
As we already have the config management and parsing logic in bundle, can we drop the "key chain" in Auto3D?
Unify to use the same representation (`AAA#BBB#CCC#2#DDD`) for nested keys as bundle?
What do you think?

Thanks.

monai/bundle/config_parser.py
mingxin-zheng: Thanks for the suggestion. Improved in 47fe5f393de15cb6ee2acdf5e34f46dcf9b2a705.

mingxin-zheng:
@Nic-Ma I have the same goal. After the auto3dseg is merged, I will use recursive_get_key_chains() (which outputs "AAA#BBB#CCC" lists that replace key chain) with other recursive get/set in ConfigParser. 

Meanwhile, I guess some users may still use the key chain. So I wanted to provide a convenient function if they need to convert from the key chain to ConfigParser ("convert_key_chain_to_id")

Nic-Ma:
Hi @mingxin-zheng ,

Actually, I feel we should not expect users to use both `key chain` and `bundle config ID`.
Better to unify to "AAA#BBB#CCC"?
@wyli What do you think?

Thanks.

mingxin-zheng:
> Hi @mingxin-zheng ,
> 
> Actually, I feel we should not expect users to use both `key chain` and `bundle config ID`. Better to unify to "AAA#BBB#CCC"? @wyli What do you think?
> 
> Thanks.

A minor point: I don't know how to name the format of "AAA#BBB#CCC", so I just name it `id` in the function name `convert_key_chain_to_id`. It shouldn't be confused with other things, and name suggestion is welcomed.

PR #4911, opened 2022-08-15
[Prototype] 4855 compose with lazy resampling

wyli:
Fixes #4855

### Description
The goal of this PR is to add a `lazy_resample: bool` option to `Compose`, so
that consecutive spatial transforms could be applied with a single step image resampling.
This will have several benefits:
- reduce the memory footprint
- speed up the compose transform
- reduce interpolation errors

The implementation mainly consists of three types of non-breaking API modifications:
- make sure all spatial transforms have an option to skip the array resampling step, and not skipping the metadata update step. this is achieved with a new base class (`LazyTransform` with class variable `eager_mode: bool`).
- the MetaTensor must have a mechanism to track lazily applied changes, this is achieved by adding a `MetaTensor.evaluated` flag and `spatial_shape` properties to track the latest expected output spatial shape
- the `Compose` transform's `__call__` must be able to find out when to evaluate the fused transforms eagerly in a sequence of transforms.

Some examples of improving the composed transfom quality, based on this sequence:
```py
xforms = [
    mt.LoadImageD(keys, ensure_channel_first=True),
    mt.Orientationd(keys, "RAS"),
    mt.SpacingD(keys, (1.5, 1.5, 1.5)),
    mt.CenterScaleCropD(keys, roi_scale=0.9),
    # mt.CropForegroundD(keys, source_key="seg", k_divisible=5),
    mt.RandRotateD(keys, prob=1.0, range_y=np.pi / 2, range_x=np.pi / 3),
    mt.RandSpatialCropD(keys, roi_size=(76, 87, 73)),
    mt.RandScaleCropD(keys, roi_scale=0.9),
    mt.Resized(keys, (30, 40, 60)),
    # mt.NormalizeIntensityd(keys),
    mt.ZoomD(keys, 1.3, keep_size=False),
    mt.FlipD(keys),
    mt.Rotate90D(keys),
    mt.RandAffined(keys),
    mt.ResizeWithPadOrCropd(keys, spatial_size=(32, 43, 54)),
    mt.DivisiblePadD(keys, k=3),
]
xform = mt.Compose(xforms, lazy_resample=True, mode='nearest')
xform.set_random_state(0)
```

before this PR (`xform = mt.Compose(xforms)`):

Screenshot 2022-08-15 at 11 33 06

Screenshot 2022-08-15 at 11 34 08

after this PR (`xform = mt.Compose(xforms, lazy_resample=True, mode='nearest')`):

Screenshot 2022-08-15 at 11 34 30

[...]

Issue #4855, opened 2022-08-08
fusing spatial transforms


wyli:
**Is your feature request related to a problem? Please describe.**
Follow up of https://github.com/Project-MONAI/MONAI/discussions/4198, https://github.com/Project-MONAI/MONAI/issues/112, when there are various spatial transforms such as `Spacing` `RandZoom` being used in `Compose`, automatically fusing them to run one step resampling would speed up the process and reduce the resampling errors by repeated resamplings.

atbenmurray:
@wyli Do you mind if I add some more detail to this feature request?

wyli:
no, please feel free to add more info @atbenmurray

myron:
This is great effort, to interpolate only once, it would avoid the repeated interpolation artifacts and should speed up the transforms pipeline, and use less memory, in theory.

At the same time, it's very difficult to implement for any general transforms, and may result in many potential bugs. Perhaps we can start by supporting only a subset of transforms (Spacing, Cropping and fused Affine (which includes Rotation, Scaling, Affine)).

And may be it's better for a user to explicitly indicate which transforms to fuse together, e.g. 

```
fused_transform = transforms.FusedSpatialTranform([SpacingD(), RandomCroppingD(), RandAffineD()])

transforms = transforms.Compose[LoadImageD(), ToTensor(), fused_transform,  NormalizeIntensities()]
```

It's one extra step for the user, but it could be easier to test, and returns the error right away if the transforms can not be fused. 

In this case each transform can have a method  to accept a grid (X, Y, Z coordinates) for interpolation, and each transform will update it (e.g. crop grid, or rotate grid or scale). 

and the transforms.FusedSpatialTranform() class will call all those spatial grid manipulations and interpolate once at the end

atbenmurray:
> This is great effort, to interpolate only once, it would avoid the repeated interpolation artifacts and should speed up the transforms pipeline, and use less memory, in theory.
> 
> At the same time, it's very difficult to implement for any general transforms, and may result in many potential bugs. Perhaps we can start by supporting only a subset of transforms (Spacing, Cropping and fused Affine (which includes Rotation, Scaling, Affine)).
> 
> And may be it's better for a user to explicitly indicate which transforms to fuse together, e.g.
> 
> ```
> fused_transform = transforms.FusedSpatialTranform([SpacingD(), RandomCroppingD(), RandAffineD()])
> 
[...]

PR #4922, opened 2022-08-17
[Early WIP] - Lazy resampling

atbenmurray:
# Lazy Resampling - deep refactor (see #4855)

The goal is to provide less memory-intensive, faster and higher quality spatial resampling by adopting the best practice of traditional graphics pipelines.

This change impacts on most of the spatial transform code, and so is also an opportunity to refactor said code to detangle what have become quite complex spatial transforms. While #4911 demonstrates that #4855 is possible, this PR is trying to find a way to decouple elements of this solution from one another and dramatically simplify the task of implementing new transforms in future and by third parties.

# Transform refactoring

Due to the complexity of existing transforms and the need to avoid adding more complexity, this refactor looks at the transforms from first principles, attempting to decouple the definition of the transform from its application.  This can be seen in the three layers of transform implementation, functional, array and dictionary:

#### Functional
Functional transforms now have a single responsibility, to generate the definition of the transform.

#### Array
Array transforms now wrap the Functional transforms, and then decide whether to push the transform to a pending queue or apply it immediately. They do this by descending from LazyTransform, which has a flag `lazy_evaluation`. This could be taken further; the pattern of whether to push to pending or apply immediately could be part of LazyTransform functionality.

#### Dictionary
Dictionary transforms wrap Array transforms, applying them once for each key. There is potential here to further reduce the code by having a standard 'apply to each specified key' function that can be called by subclasses.

The amount of code is dramatically reduced as a result, at least in the transforms reimplemented so far.

# Compose compilers

As part of looking at this refactor, it is becoming obvious that there are a few places where the list of transforms to be executed are manipulated as they are being executed. Each place in the code that does this does so with the assumption that no-one else is doing it. As such, one compose manipulator will "win" and the others do not appear to be applied at all.

One way around this is to specify "compose compilers". These are functions that will take an input list of transforms and compile it to a new list of transforms. The benefit of compose compilers is twofold:
1. They reduce the number of places where handling of the compose is customised
2. They can be chained in order to meet the demands of multiple transform list customisations (lazy resampling and cached datasets, for example)

### Types of changes

# TODO:

### Spatial transform implementation

See **Table: Implementation by Spatial Transform** for detailed progress:
 - [ ] Implement functional versions of all spatial transforms
 - [ ] Implement array versions of all spatial transforms
 - [ ] Implement dictionary versions of all spatial transforms

### LazyTransform and RandomTransform interfaces for 3rd party use 
Interfaces that indicate capability without needing to adopt implementations:
 - [x] Implement LazyTransform interface
 - [x] Optionally reimplement RandomTransform interface for 3rd party use

### Compose compilation

This is done if we go with compose compilers as a unified mechanism for lazy resampling, smart datasets, etc.:
 - [x] Compose compiler implementation
 - [ ] Compose accepts compose compilers
 - Compose compiler transforms
   - [x] Cached / Persistent dataset transforms
   - [x] Multi-sample transforms

### MetaTensor updates
Optionally make MetaTensors able to share read only access to underlying Tensor for multi-sampling transforms:
 - [ ] MetaTensor sharing underlying tensor 

### Lazy resampling compose
This is done if we go with augmenting compose and leaving smart datasets alone:
 - [ ] Compose performs lazy resampling
 
### Table: Implementation by spatial transform
```
#                                                                                               R
#                                                                                               a
#                                                                                               n
#                          S  R                                                                 d
#                          p  e                                         R  R                 G  G
#                          a  s                                         a  a           R  R  r  r        R
#                          t  a                       R        R        n  n           a  a  i  i        a
#                          i  m     O                 a        a        d  d           n  n  d  d        n
#                          a  p     r                 n  R     n     A  A  D        R  d  d  D  D        d
#                          l  l     i                 d  a     d     f  f  e        a  2  3  o  i  G  G  G       T
#                          R  e     e              R  R  n  R  A  R  f  f  f  R     n  D  D  s  s  r  r  r    I  r
#                          e  T  S  n              o  o  d  a  x  a  i  i  o  e     d  E  E  t  t  i  i  i    d  a
#                          s  o  p  t     R  R     t  t  R  n  i  n  n  n  r  s  A  A  l  l  o  o  d  d  d    e  n
#                          a  M  a  a     e  o     a  a  o  d  s  d  e  e  m  a  f  f  a  a  r  r  S  P  S    n  s
#                          m  a  c  t  F  s  t  Z  t  t  t  F  F  Z  G  G  G  m  f  f  s  s  t  t  p  a  p    t  l
#                          p  t  i  i  l  i  a  o  e  e  a  l  l  o  r  r  r  p  i  i  t  t  i  i  l  t  l    i  a
#                          l  c  n  o  i  z  t  o  9  9  t  i  i  o  i  i  i  l  n  n  i  i  o  o  i  c  i    t  t
#                          e  h  g  n  p  e  e  m  0  0  e  p  p  m  d  d  d  e  e  e  c  c  n  n  t  h  t    y  e

Functional                       X  .  X  X  X  X  X  n  n  n  n  n  n  n  n  n  .  n  n  n  .  n  .  .  n    X  X
Array forward                    X  .  X  X  X  X  X  X  X  X  X  .  .  .  .  n  .  .  .  .  .  .  .  .  .    X  X
Dictionary forward               X  .  .  X  X  X  .  .  .  .  .  .  .  .  .  n  .  .  .  .  .  .  .  .  .    .  X
Test forward                     f  .  .  f  f  f  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .    .  .
Array inverse                    .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .    .  .
Dictionary inverse               .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .    .  .
Test inverse                     .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .    .  .
```

### Table: Implementation by Crop / Pad transform
```
#                                                  R           R
#                                                  a        R  a
#                                                  n        a  n
#                                                  d        n  d  R
#                                                  S        d  C  e
#                                      C           p        C  r  s
#                                      e           a     R  r  o  i 
#                                      n  C  R     t     a  o  p  z
#                                      t  e  a     i  C  n  p  B  e
#                                      e  n  n  R  a  r  d  B  y  W
#                             D        r  t  d  a  l  o  W  y  L  i  B
#                             i     S  S  e  S  n  C  p  e  P  a  t  o
#                       S     v     p  p  r  p  d  r  F  i  o  b  h  u
#                       p  B  i     a  a  S  a  S  o  o  g  s  e  P  n
#                       a  o  s     t  t  c  t  c  p  r  h  N  l  a  d
#                       t  r  i     i  i  a  i  a  S  e  t  e  C  d  i    C
#                       i  d  b     a  a  l  a  l  a  g  e  g  l  O  n    r
#                       a  e  l     l  l  e  l  e  m  r  d  L  a  r  g    o
#                       l  r  e  C  C  C  C  C  C  p  o  C  a  s  C  R    p
#                    P  P  P  P  r  r  r  r  r  r  l  u  r  b  s  r  e    P
#                    a  a  a  a  o  o  o  o  o  o  e  n  o  e  e  o  c    a
#                    d  d  d  d  p  p  p  p  p  p  s  d  p  l  s  p  t    d

Functional           .  .  .  .  .  .  .  .  n  n  n  .  n  n  n  .  .    X
Array forward        .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .    X
Dictionary forward   .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .    X
Test forward         .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .    .
Array inverse        .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .    .
Dictionary inverse   .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .    .
Test inverse         .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .    .
```


monai/transforms/atmostonce/functional.py
```diff
@@ -0,0 +1,295 @@
...
+    if diagonal is True:
+        raise ValueError("'diagonal' value of True is not currently supported")
+
+    mode_ = look_up_option(mode, GridSampleMode)
+    padding_mode_ = look_up_option(padding_mode, GridSamplePadMode)
+    dtype_ = get_equivalent_dtype(dtype or img.dtype, torch.Tensor)
+    zoom_factors = [i / j for i, j in zip(src_pixdim_, pixdim_)]
+
+    transform = MatrixFactory.from_tensor(img).scale(zoom_factors)
+    im_extents = extents_from_shape(img.shape)
```
wyli, line 66: when we have a chain of lazy transforms, `img.shape` will become outdated, correct? because we don't actually run `img_t = apply(img_t)`. 

so I think accessing `img.shape` at this point is inaccurate, this should be something like `metadata['spatial_shape']` where `metadata` should come from a previous lazy transform or `img.shape` if the previous transform evaluates eagerly.

wyli: and it seems `img` is not really needed in this function (and in the other functionals as well), just need the spatial shape information of it.

atbenmurray: Yep, I'm aware of the issue (chain of lazy transforms). I'm still trying to figure out the best way to handle when / how transforms to shape / extents occur. I'll hopefully have it sorted by the end of the day; just experimenting at the moment.

atbenmurray: I think that we should be able describe the extent change by concatenating the individual extent change operations at the point we perform the stacking of transforms; details to follow

atbenmurray: I've gone the way @wyli went in PR #4911. I thought I could do this by applying the accumulated transform to the extents as they are defined at that point in the chain, but there are a couple of scenarios that seem to break that.

PR #4933, opened 2022-08-18
4931- provide an initial channel dim 

### LINKED ISSUE #931: Add 2D segmentation example
Recently, we have received request for 2D segmentation example for internal and external users, need to add it.

wyli:

Fixes #4931

### Description
adds the flexibility of providing a channel dim.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/transforms/utility/array.py
+++ b/monai/transforms/utility/array.py
@@ -209,23 +209,25 @@ class EnsureChannelFirst(Transform):
 
     Args:
         strict_check: whether to raise an error when the meta information is insufficient.
-        add_channel_default: If True and the input image `img` is not a MetaTensor and `meta_dict` is not given,
-            or `img` is a MetaTensor but doesn't specify channel dimension, use the value is `no_channel`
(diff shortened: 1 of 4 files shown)


monai/transforms/utility/array.py
```diff
@@ -239,13 +241,13 @@ def __call__(self, img: torch.Tensor, meta_dict: Optional[Mapping] = None) -> to
         channel_dim = meta_dict.get("original_channel_dim")  # type: ignore
```
ericspod, line 241:
```suggestion
        channel_dim = meta_dict.get("original_channel_dim", None)  # type: ignore
```
If `self.input_channel_dim` is not `None` but `img` was not a `MetaTensor`, it'll get converted into one with an empty meta dict, or if `img` is a `MetaTensor` without this value.

wyli: updated -- btw if it's a regular dictionary the get method by default return None for the missing key https://docs.python.org/3/library/stdtypes.html?highlight=get#dict.get

ericspod (approved):

PR #4934, opened 2022-08-18
Improve data type conversion performance in convert_to_numpy

### LINKED ISSUE #4900: `WSIReader` does not support `TiffFile` backend
`image_reader.WSIReader` supports tifffile (https://github.com/Project-MONAI/MONAI/blob/e4624678cc15b5a5670776715c935279f7837c89/monai/data/image_reader.py#L1257), but it does not supported in `wsi_reader.WSIReader`. Can I ask what is the purpose here to remove the supports?

I'm not sure but it seems `tifffile` loads faster in some cases.
For example:
Screen Shot 2022-08-12 at 6 18 58 PM

gigony:
Signed-off-by: Gigon Bae <gbae@nvidia.com>

#4900

### Description

Improve type conversion(`convert_to_numpy()`) performance by converting to contiguous array first.

```bash
# Without patch
  Thread elapsed time (cucim loader) : 14.059558755951002
    type: <class 'monai.data.meta_tensor.MetaTensor'>, shape: (3, 60797, 34007), dtype: torch.float32

# With patch
  Thread elapsed time (cucim loader) : 10.268190009985119
    type: <class 'monai.data.meta_tensor.MetaTensor'>, shape: (3, 60797, 34007), dtype: torch.float32
```

test code
```python
from contextlib import ContextDecorator
from time import perf_counter

import monai
from monai.transforms import LoadImage

loader_cucim = LoadImage(
    reader=monai.data.wsi_reader.WSIReader, backend="cucim", image_only=True)


class Timer(ContextDecorator):
    def __init__(self, message):
        self.message = message
        self.end = None

    def elapsed_time(self):
        self.end = perf_counter()
        return self.end - self.start

    def __enter__(self):
        self.start = perf_counter()
        return self

    def __exit__(self, exc_type, exc, exc_tb):
        if not self.end:
            self.elapsed_time()
        print("{} : {}".format(self.message, self.end - self.start))


image_file = "/home/gbae/repo/cucim/notebooks/input/006388_0.tif"

with Timer("  Thread elapsed time (cucim loader)") as timer:
    image = loader_cucim(image_file)
print(f"    type: {type(image)}, shape: {image.shape}, dtype: {image.dtype}")
```

### Status
Ready

### Types of changes

### DIFF
--- a/monai/utils/type_conversion.py
+++ b/monai/utils/type_conversion.py
@@ -175,6 +175,13 @@ def convert_to_numpy(data, dtype: DtypeLike = None, wrap_sequence: bool = False)
     elif has_cp and isinstance(data, cp_ndarray):
         data = cp.asnumpy(data).astype(dtype, copy=False)
     elif isinstance(data, (np.ndarray, float, int, bool)):
+        # Convert into a contiguous array first if the current dtype's size is smaller than the target dtype's size.
+        # This help improve the performance because (convert to contiguous array) -> (convert dtype) is faster
+        # than (convert dtype) -> (convert to contiguous array) when src dtype (e.g., uint8) is smaller than
+        # target dtype(e.g., float32) and we are going to convert it to contiguous array anyway later in this
+        # method.
+        if data.ndim > 0 and data.dtype.itemsize < np.dtype(dtype).itemsize:
+            data = np.ascontiguousarray(data)
         data = np.asarray(data, dtype=dtype)
     elif isinstance(data, list):
         list_ret = [convert_to_numpy(i, dtype=dtype) for i in data]
--- a/monai/utils/type_conversion.py
+++ b/monai/utils/type_conversion.py
@@ -180,7 +180,7 @@ def convert_to_numpy(data, dtype: DtypeLike = None, wrap_sequence: bool = False)
         # than (convert dtype) -> (convert to contiguous array) when src dtype (e.g., uint8) is smaller than
         # target dtype(e.g., float32) and we are going to convert it to contiguous array anyway later in this
         # method.
-        if data.ndim > 0 and data.dtype.itemsize < np.dtype(dtype).itemsize:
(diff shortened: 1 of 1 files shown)


ericspod (approved):

monai/utils/type_conversion.py
```diff
@@ -175,6 +175,13 @@ def convert_to_numpy(data, dtype: DtypeLike = None, wrap_sequence: bool = False)
     elif has_cp and isinstance(data, cp_ndarray):
         data = cp.asnumpy(data).astype(dtype, copy=False)
     elif isinstance(data, (np.ndarray, float, int, bool)):
+        # Convert into a contiguous array first if the current dtype's size is smaller than the target dtype's size.
+        # This help improve the performance because (convert to contiguous array) -> (convert dtype) is faster
+        # than (convert dtype) -> (convert to contiguous array) when src dtype (e.g., uint8) is smaller than
+        # target dtype(e.g., float32) and we are going to convert it to contiguous array anyway later in this
+        # method.
+        if data.ndim > 0 and data.dtype.itemsize < np.dtype(dtype).itemsize:
```
grlee77, line 183: CI is failing due to this call to ndim when `data` may not be a ndarray
```
monai/utils/type_conversion.py:183:12: error: Item "float" of "Union[ndarray[Any, Any], float]" has no attribute "ndim"  [union-attr]
monai/utils/type_conversion.py:183:30: error: Item "float" of "Union[ndarray[Any, Any], float]" has no attribute "dtype"  [union-attr]
```

grlee77, line 183:
```suggestion
        if isinstance(data, np.ndarray) and data.ndim > 0 and data.dtype.itemsize < np.dtype(dtype).itemsize:
```

gigony: Thanks @grlee77 ! 
You're right, data can be `float, int, bool` so need your suggested change.

drbeh (approved):

PR #4944, opened 2022-08-19
Workflow Fix

ericspod:
### Description
Very minor fix to how workflows decide what the best metric result is such that the first epoch always chooses whatever its metric was. `BaseWorkflow` was also removed since it wasn't used anywhere, this can be put back if somewhere else needs it.

### Status
**Ready**

### Types of changes

### DIFF
--- a/docs/source/engines.rst
+++ b/docs/source/engines.rst
@@ -16,11 +16,6 @@ Workflows
 
 .. currentmodule:: monai.engines
 
-`BaseWorkflow`
-~~~~~~~~~~~~~~
-.. autoclass:: BaseWorkflow
-    :members:
-
 `Workflow`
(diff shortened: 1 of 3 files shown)


wyli (approved):

Nic-Ma:
Hi @ericspod ,

The `BaseWorkflow` is useful in the NVIDIA Clara Train project to ask developers to follow the signature if they want to develop their own workflow.

Thanks.

ericspod:
Could that be put into the clara train package then? If it's not used in MONAI I would say it shouldn't be here, but I can put it back in another PR.

Nic-Ma:
Hi @ericspod ,

I think we are fine now, Clara Train supposes to work with fixed MONAI version.

Thanks.

PR #4945, opened 2022-08-19
Compute HoVer Maps

### LINKED ISSUE #4932: Compute horizontal vertical maps
HoVerNet pipeline requires to compute horizontal and vertical maps from instance segmentations, which are based on the horizontal and vertical distances of each pixel in a nucleus to the center of mass of each nucleus.

drbeh:
Fixes #4932 

### Description
Implements a transform to compute horizontal and vertical maps, which is used in HoVerNet pipeline.

### Status
**Ready**

### Types of changes


monai/transforms/intensity/array.py
```diff
@@ -2301,3 +2302,41 @@ def __call__(self, image: NdarrayOrTensor):
...
+
+    def __init__(self, dtype: DtypeLike = "float32") -> None:
+        super().__init__()
+        self.dtype = dtype
+
+    def __call__(self, mask: NdarrayOrTensor):
+        instance_mask = convert_data_type(mask, np.ndarray)[0]
+
+        h_map = instance_mask.astype(self.dtype)
+        v_map = instance_mask.astype(self.dtype)
```
wyli, line 2326: These two maps might share the same memory, the in-place changes in line 2338/2339 might be incorrect

tests/test_compute_ho_ver_maps.py
```diff
@@ -0,0 +1,73 @@
...
+from parameterized import parameterized
+
+from monai.transforms.intensity.array import ComputeHoVerMaps
+from monai.utils import min_version, optional_import
+from tests.utils import TEST_NDARRAYS, assert_allclose
+
+skimage, has_skimage = optional_import("skimage", "0.19.0", min_version)
+
+
+INSTANCE_MASK = np.zeros((16, 16), dtype="int16")
```
wyli, line 25: please use https://github.com/Project-MONAI/MONAI/blob/1941283b1862e3f578ccfeb2bdfa864171125029/monai/data/synthetic.py#L21 to write more compact use cases.

monai/transforms/intensity/array.py
drbeh: Good catch! Right, if the the instance_mask has the same type as output masks, there won't be any copy. will fix it. Thank

tests/test_compute_ho_ver_maps.py
drbeh: That's good idea but here I need to manually calculate the normalized distances to the center of mass of each region, so I cannot work with a randomly generated mask. Does it make sense?

wyli (approved):
Looks good to me, but would be great if we can simplify the test cases, any idea? Cc @Nic-Ma @ericspod @rijobro

drbeh:
Hi @wyli @Nic-Ma @ericspod @rijobro,
Any input here? It would be great if we can merge this PR.

Issue #4967, opened 2022-08-23
AddRandomGuidanceDeepEditd transform did not successfully add Guidance


wojiazaiyugang:
**Describe the bug**
AddRandomGuidanceDeepEditd transform did not successfully add Guidance
The Guidance that needs to be added is calculated in AddRandomGuidanceDeepEditd, but when the transform ends, the calculated Guidance is not added to d
![image](https://user-images.githubusercontent.com/14771322/186106300-2c98d58b-d48e-4510-bd88-ddccdd9ca686.png)

**To Reproduce**
2. Install '....'
3. Run commands '....'

**Expected behavior**

**Screenshots**

**Environment**

Ensuring you use the relevant python executable, please paste the output of:

```
python -c 'import monai; monai.config.print_debug_info()'
```

**Additional context**

PR #4970, opened 2022-08-23
4967 fix deepedit AddRandomGuidanceDeepEditd transform

wojiazaiyugang:
Fixes https://github.com/Project-MONAI/MONAI/issues/4967

### Description

### Status

### Types of changes


diazandr3s (approved):
Looks good to me. Thanks!

Edited: Sorry, I didn't realise you've added this line `d[self.guidance_key] = self.guidance`. 
Have you tried to train DeepEdit on MONAI Label? These transforms are interconnected and followed a way of passing the guidance.

wyli:
please sign off the commit messages (https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#signing-your-work)

monai/apps/deepedit/transforms.py
```diff
@@ -595,20 +595,20 @@ def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.nda
...
+                                self.guidance[key_label] = []
                         logger.info(f"Number of simulated clicks: {counter}")
                         break
 
                 # Breaking once all labels are covered
                 if len(keep_guidance) == len(d["label_names"].keys()):
                     logger.info(f"Number of simulated clicks: {counter}")
                     break
-
+        d[self.guidance_key] = self.guidance
```
diazandr3s, line 611: I'd suggest we first perform training and inference on MONAI Label. This might break the training

diazandr3s (changes requested):

monai/apps/deepedit/transforms.py
wojiazaiyugang: image

this is from training

wojiazaiyugang: image

this is from monai label inference, when inference, AddRandomGuidanceDeepEditd is not used.
AddRandomGuidanceDeepEditd only used in training to add some random simulation guidance

wojiazaiyugang: image

this is from monai label deepedit training, same as train in monai

wojiazaiyugang:
I noticed that there is one unit test (tests/test_deepedit_transforms.py/TestAddRandomGuidanceCustomd) that fails, but I don't know how to fix it. My current thinking is that AddRandomGuidanceDeepEditd should add some random points, but these are randomly generated so I can't unit test. Please point out if my understanding is wrong, thanks

diazandr3s:
> I noticed that there is one unit test (tests/test_deepedit_transforms.py/TestAddRandomGuidanceCustomd) that fails, but I don't know how to fix it. My current thinking is that AddRandomGuidanceDeepEditd should add some random points, but these are randomly generated so I can't unit test. Please point out if my understanding is wrong, thanks

Sorry, I'm a bit lost here. Why are we modifying the DeepEdit transforms? What's your use case?

wojiazaiyugang:
> > I noticed that there is one unit test (tests/test_deepedit_transforms.py/TestAddRandomGuidanceCustomd) that fails, but I don't know how to fix it. My current thinking is that AddRandomGuidanceDeepEditd should add some random points, but these are randomly generated so I can't unit test. Please point out if my understanding is wrong, thanks
> 
> Sorry, I'm a bit lost here. Why are we modifying the DeepEdit transforms? What's your use case?

Please refer to this issue
https://github.com/Project-MONAI/MONAI/issues/4967

To sum up, I tracked the code when I trained DeepEdit and found that the AddRandomGuidanceDeepEditd transform did not add some random guidance as expected. I think may forgot to update `d` at the end.

diazandr3s:
> > > I noticed that there is one unit test (tests/test_deepedit_transforms.py/TestAddRandomGuidanceCustomd) that fails, but I don't know how to fix it. My current thinking is that AddRandomGuidanceDeepEditd should add some random points, but these are randomly generated so I can't unit test. Please point out if my understanding is wrong, thanks
> > 
> > 
> > Sorry, I'm a bit lost here. Why are we modifying the DeepEdit transforms? What's your use case?
> 
> Please refer to this issue #4967
> 
> To sum up, I tracked the code when I trained DeepEdit and found that the AddRandomGuidanceDeepEditd transform did not add some random guidance as expected. I think may forgot to update `d` at the end.

Thanks for this, @wojiazaiyugang 
I've completely missed this bit. This will significantly help DeepEdit performance

wyli:
could you please have a look @SachidanandAlle? it seems there's still an issue with the windows tests (https://github.com/Project-MONAI/MONAI/runs/8096979079?check_suite_focus=true) @diazandr3s

diazandr3s:
> could you please have a look @SachidanandAlle? it seems there's still an issue with the windows tests (https://github.com/Project-MONAI/MONAI/runs/8096979079?check_suite_focus=true) @diazandr3s

Just fix the issue. I think this time it will pass :)

PR #4972, opened 2022-08-23
Implement TiffFile backend for WSIReader

### LINKED ISSUE #4900: `WSIReader` does not support `TiffFile` backend

drbeh:
Fixes #4900 

### Description
This PR implements `TiffFileWSIReader` as the backend of `monai.dadta.wsi_reader.WSIReader`.

### Status
**Ready**

### Types of changes


Nic-Ma:
Hi @yiheng-wang-nv ,

I think this PR should be useful for your pathology challenge pipeline.

Thanks in advance.

yiheng-wang-nv:
> Hi @yiheng-wang-nv ,
> 
> I think this PR should be useful for your pathology challenge pipeline. Could you please help review it?
> 
> Thanks in advance.

Hi @Nic-Ma , I'm not an expert of pathology, and may not be able to review this PR. BTW, I think it may be transplanted from the deprecated class, thus should not have other issues. Could you please invite someone else to review it? Thanks!

Nic-Ma (approved):
Looks good to me.
@KumoLiu Could you please help also review it?

Thanks in advance.

monai/data/wsi_reader.py
```diff
@@ -270,11 +271,13 @@ class WSIReader(BaseWSIReader):
...
-        self.reader: Union[CuCIMWSIReader, OpenSlideWSIReader]
+        self.reader: Union[CuCIMWSIReader, OpenSlideWSIReader, TiffFileWSIReader]
         if self.backend == "cucim":
             self.reader = CuCIMWSIReader(level=level, channel_dim=channel_dim, **kwargs)
         elif self.backend == "openslide":
             self.reader = OpenSlideWSIReader(level=level, channel_dim=channel_dim, **kwargs)
+        elif self.backend == "tifffile":
+            self.reader = TiffFileWSIReader(level=level, channel_dim=channel_dim, **kwargs)
         else:
             raise ValueError(f"The supported backends are cucim and openslide, '{self.backend}' was given.")
```
KumoLiu, line 282: raise ValueError(f"The supported backends are cucim, openslide and tifffile ....

Nic-Ma:
Hi @wyli ,

I tried to run the package tests several times, but it always fails with no clear error.
Do you have any idea what happened or saw this before?

Thanks in advance.

wyli:
looks like the something is using too much memory and the premerge is cancelled `exit code 137` https://github.com/Project-MONAI/MONAI/runs/8084831205?check_suite_focus=true

drbeh:
> looks like the something is using too much memory and the premerge is cancelled `exit code 137` https://github.com/Project-MONAI/MONAI/runs/8084831205?check_suite_focus=true
Hi @Nic-Ma @wyli,
That's right! It seems that on this test, there is no cuCIM nor OpenSlide so they will be skipped but TiffFile is installed and it will load the image, which is big but not that big. Do you know what is the memory limit of this instance?

wyli:
probably it's about 7GB https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

drbeh:
> probably it's about 7GB https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

I see, then it should be it. Tifffile backend needs to load the entire image into memory when extracting small patches. I'll find a way to fix this test.

drbeh:
@wyli @Nic-Ma the memory overuse of tifffile backend is now fixed and the packaging test is passed.

PR #4973, opened 2022-08-23
Fix For RandSmoothDeform Component Order

ericspod:

### Description
One of the issues with `torch.nn.functional.grid_sample` is its definition of grid coordinate space versus dimensional ordering in inputs. The Pytorch ordering is HW or DHW which in coordinate terms is YX or ZYX, this is inverse to what the function actually expects in the  component ordering of the grid given as input. The solution is to reverse the order of the component dimension. Added tests check that an image with a pattern is effectively unchanged with a deformation having almost 0 magnitude to verify images are not transposed in one dimension or another which was occurring without this fix.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/data/synthetic.py
+++ b/monai/data/synthetic.py
@@ -48,6 +48,9 @@ def create_test_image_2d(
         channel_dim: if None, create an image without channel dimension, otherwise create
             an image with channel dimension as first dim or last dim. Defaults to `None`.
         random_state: the random generator to use. Defaults to `np.random`.
+
+    Returns:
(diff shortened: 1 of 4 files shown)


wyli (approved):
there were some discussions in pytorch as well https://discuss.pytorch.org/t/surprising-convention-for-grid-sample-coordinates/79997

ericspod:
> there were some discussions in pytorch as well https://discuss.pytorch.org/t/surprising-convention-for-grid-sample-coordinates/79997

Yes, I saw that earlier. So it's not just me!

PR #5002, opened 2022-08-26
5001 fix convert global weights

### LINKED ISSUE #5001: No error raised if no global weights are converted in MonaiAlgo
**Describe the bug**
In `convert_global_weights`, no error raised if no global weights are converted in MonaiAlgo

**To Reproduce**
Server uses a different model configuration than client. Mismatching keys cause `convert_global_weights` to not convert any global weights causing later errors.

**Expected behavior**
Raise error if no weights are converted.

**Screenshots**
n/a

**Environment**

```
```

**Additional context**
dev branch

holgerroth:
Fixes #5001.

### Description
Raise error if no weights are converted.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/fl/client/monai_algo.py
+++ b/monai/fl/client/monai_algo.py
@@ -47,6 +47,7 @@ def convert_global_weights(global_weights, local_var_dict):
     """Helper function to convert global weights to local weights format"""
     # Before loading weights, tensors might need to be reshaped to support HE for secure aggregation.
     model_keys = global_weights.keys()
+    n_converted = 0
     for var_name in local_var_dict:
         if var_name in model_keys:
             weights = global_weights[var_name]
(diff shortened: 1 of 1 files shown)


monai/fl/client/monai_algo.py
```diff
@@ -55,9 +56,12 @@ def convert_global_weights(global_weights, local_var_dict):
                 weights = torch.reshape(torch.as_tensor(weights), local_var_dict[var_name].shape)
                 # update the local dict
                 local_var_dict[var_name] = weights
+                n_converted += 1
             except Exception as e:
                 raise ValueError(f"Convert weight from {var_name} failed.") from e
-    return local_var_dict
+    if n_converted == 0:
+        raise ValueError(f"No global weights converted! Received weight dict keys are {list(global_weights.keys())}")
```
wyli, line 63: perhaps this could be a warning saying received weight dict but no matched parameter keys?

holgerroth: Right. Continuing here would cause an error later when computing weight differences because the local and global weights end up on different devices.

Is there a use case where train() is called but provides empty weights in the `data`? It would be unusual for the standard federated averaging workflow we want to implement here. @habibcomsats , @Warvito, any opinions?

holgerroth: Actually, we can support this by moving the weight diff computation to the CPU() for both local and global model. Then just printing the warning should be sufficient.

holgerroth: @wyli, I made some changes that move all returned weights to CPU, allowing also to compute differences if non have been converted. Now, only a warning is printed in that case.

wyli (approved):
/black
Thanks, it looks good to me.

PR #4765, opened 2022-07-26
Auto3D Task Module 1 (DataAnalyzer) for #4743

mingxin-zheng:
### Description
Implemented a DataAnalyzer class to encapsulate data analysis module. As the beginning part of the auto3d/automl pipeline, the module shall find data and label from user inputs and generate a summary (dictionary) of data stats. The summary includes
	- file names, list, number of files;
	- dataset summary (basic information, image dimensions, number of classes, etc.);
	- individual data information (spacing, image size, number and size of the regions, etc.).
The summary can be exported as a YAML file and a dictionary variable for use in Python

Example Usage:
```python
from monai.apps.auto3d.data_analyzer import DataAnalyzer

datalist = {
    "testing": [{"image": "image_003.nii.gz"}],
    "training": [
        {"fold": 0, "image": "image_001.nii.gz", "label": "label_001.nii.gz"},
        {"fold": 0, "image": "image_002.nii.gz", "label": "label_002.nii.gz"},
        {"fold": 1, "image": "image_001.nii.gz", "label": "label_001.nii.gz"},
        {"fold": 1, "image": "image_004.nii.gz", "label": "label_004.nii.gz"},
    ],
}

dataroot = '/datasets' # the directory where you have the image files (in this example we're using nii.gz)
analyser = DataAnalyzer(datalist, dataroot)
datastat = analyser.get_all_case_stats() # it will also generate a data_stats.yaml that saves the stats
```

### Status
Ready for Review 
Reference issue https://github.com/Project-MONAI/MONAI/issues/4743.

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [x] New tests added to cover the changes.
- [x] Quick tests passed locally by running `./runtests.sh --quick --unittests  --disttests`.
- [x] In-line docstrings updated.
- [x] Documentation updated, tested `make html` command in the `docs/` folder.

PR #5003, opened 2022-08-26
4996 Enhance `ConfigParser` to support nested new ID

### LINKED ISSUE #4996: `ConfigParser` can't set new nested ID
Currently, our `ConfigParser` can't set new nested ID, for example "key1#2#key2" if even "key1" doesn't exist:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/config_parser.py#L159

Would be nice to support new nested ID.

Nic-Ma:
Fixes #4996 .

### Description
This PR enhanced the `ConfigParser` to support nested new ID.

### Status
**Work in progress**

### Types of changes


mingxin-zheng:
Looks good to me!

monai/bundle/config_parser.py
```diff
@@ -154,9 +154,25 @@ def __setitem__(self, id: Union[str, int], config: Any):
...
+        conf_ = self.get()
+        if recursive:
+            if conf_ is None:
+                conf_ = [] if keys[0] == "0" else {}
+            for i, k in enumerate(keys[:-1]):
+                if isinstance(conf_, dict):
+                    if k not in conf_:
+                        conf_[k] = [] if keys[i + 1] == "0" else {}
+                if isinstance(conf_, list):
+                    if int(k) == len(conf_):
```
wyli, line 166: I think this way of appending new elements to a list by index is quite confusing...

Nic-Ma: Hi @wyli ,

Thanks for your review.
I added docstring to describe the feature, it's useful in the data analysis of Auto3D to predefine the report template: https://github.com/Project-MONAI/MONAI/pull/4765#discussion_r955224503, CC @mingxin-zheng .
And another typical use case is to add a transform or handler when overriding a config file.
Refer to Hydra argument: https://github.com/facebookresearch/hydra/blob/v1.2.0/hydra/_internal/instantiate/_instantiate2.py#L155

Thanks.

wyli: it doesn't make sense in terms of the python style...
it's not ok to set an array with a new index
```py
array = [0, 1]
array[2] = 2  # IndexError: list assignment index out of range
```
while it's ok to set a dictionary with a new key:
```py
d = {"a": 0}
d['b'] = 1
```

Nic-Ma: Hi @wyli ,

Maybe I didn't describe the PR clearly.
The usage is not `array[2] = 2`, which is against python style.
Instead, I just added the `recursive` option to the `set()` function, similar to the enhancement on `get()`, we added support to give the default value in the `get()` function.
So the usage will be:
```py
array = ConfigParser([0, 1])
array.set(id="2") = 2

array[3] = 3  # IndexError
```
What do you think?

Thanks.


Thanks.

Nic-Ma: Simplified to only support nested dicts.

Thanks.

Issue #5008, opened 2022-08-26
Add ensemble module for Auto3Dseg 


mingxin-zheng:
**Is your feature request related to a problem? Please describe.**
This feature is part of the Auto3Dseg development 

**Describe the solution you'd like**
An algorithm ensemble that supports
- loading inferer and configs from folders
- perform inference on the loaded models
- user can select the method of ensemble (mean of best 5, for example)

**Describe alternatives you've considered**
Reference for other ensemble
- sklean: https://github.com/automl/auto-sklearn/tree/master/autosklearn/ensembles
- pytorch: https://github.com/automl/Auto-PyTorch/tree/master/autoPyTorch/ensemble

PR #5012, opened 2022-08-26
Ensemble 5008

mingxin-zheng:
Initial draft for #5008 .

### Description
Ensemble models for Auto3Dseg and external algorithm

### Status
**Ready**

### Types of changes


monai/apps/auto3dseg/data_analyzer.py
```diff
@@ -41,7 +41,8 @@ def strenum_representer(dumper, data):
     return dumper.represent_scalar("tag:yaml.org,2002:str", data.value)
 
 
-config_parser.yaml.SafeDumper.add_multi_representer(StrEnum, strenum_representer)
+if optional_import("yaml")[1]:
+    config_parser.yaml.SafeDumper.add_multi_representer(StrEnum, strenum_representer)
```
Nic-Ma, line 45: @wyli I think maybe we should not include this YAML feature in this release?

Thanks.

mingxin-zheng: Let me double check @Nic-Ma . Thanks for bringing it up

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -0,0 +1,227 @@
...
+    def set_infer_files(self, dataroot: str, data_src_cfg_file: str):
+        """
+        Set the files to perform model inference.
+
+        Args:
+            dataroot: the path of the files
+            data_src_cfg_file: the data source file path
+        """
+        with open(data_src_cfg_file) as f:
+            datalist = json.load(f)
```
KumoLiu, line 75: Maybe it seems like `data_list_file_path` instead of `data_src_cfg_file `?

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -0,0 +1,370 @@
...
+        else:
+            n_devices, devices_str = torch.cuda.device_count(), ""
+        if n_devices > 1:
+            cmd = f"torchrun --nnodes={1:d} --nproc_per_node={n_devices:d} "
+        else:
+            cmd = "python "  # TODO: which system python?
+        cmd += base_cmd
+        if params and isinstance(params, Mapping):
+            for k, v in params.items():
+                cmd += f" --{k}={v}"
```
wyli, line 204: not sure how good it is for passing args with special characters... may need some tests for file paths.

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -0,0 +1,370 @@
...
+        # handling scripts files
+        output_scripts_path = os.path.join(write_path, "scripts")
+        if os.path.exists(output_scripts_path):
+            shutil.rmtree(output_scripts_path)
+        if self.scripts_path is not None and os.path.exists(self.scripts_path):
+            shutil.copytree(self.scripts_path, output_scripts_path)
+        # handling config files
+        output_config_path = os.path.join(write_path, "configs")
+        if os.path.exists(output_config_path):
+            shutil.rmtree(output_config_path)
```
wyli, line 161: this `rmtree` might remove existing experiments and results...kind of dangerous here but I don't have a better idea for now.

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -0,0 +1,370 @@
...
+    """
+
+    def __init__(self, algos=None, data_stats_filename=None, data_src_cfg_name=None):
+        self.algos = []
+        if algos is None:
+            # trigger the download process
+            sys.path.insert(0, "..")
+            algo_compressed_file = os.path.join("..", "algo_templates.tar")
+            download_and_extract(algo_zip, algo_compressed_file, "..", md5)
+            algos = copy(default_algos)
```
wyli: perhaps consider putting the downloaded files to a subfolder of the torch hub folder: https://pytorch.org/docs/stable/hub.html?highlight=get_dir#torch.hub.get_dir

I used the directory previously for models, code here:
https://github.com/Project-MONAI/MONAI/blob/f9e8e10fab3a0c0b87708222db96571236c9e764/monai/apps/mmars/mmars.py#L134-L138

so maybe we can create `algo_dir = Path(get_dir()) / "algo_templates"`

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -0,0 +1,370 @@
...
+            fold_idx: an index to append to the name of the algorithm to
+        """
+        for algo in self.algos:
+            for f_id in ensure_tuple(fold_idx):
+                data_stats = self.get_data_stats()
+                data_src_cfg = self.get_data_src(fold_idx)
+                gen_algo = deepcopy(algo)
+                gen_algo.set_data_stats(data_stats)
+                gen_algo.set_data_source(data_src_cfg)
+                name = f"{gen_algo.name}_{f_id}"
```
wyli, line 387: might need a UUID here to have a unique bundle name?

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -0,0 +1,227 @@
...
+
+        super().__init__()
+        self.n_best = n_best
+
+    def sort_score(self):
+        """sort the best_metrics"""
+        scores = concat_val_to_np(self.algos, [ScriptEnsembleKeys.SCORE])
+        return np.argsort(scores).tolist()
+
+    def rank_algos(self):
```
wyli, line 112:
```suggestion
    def rank_algos(self, n_best: = -1):
```

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -0,0 +1,227 @@
...
+
+        for i in range(len(self.infer_files)):
+            preds = []
+            infer_filename = self.infer_files[i]
+            for algo in self.ranked_algo:
+                infer_instance = algo[ScriptEnsembleKeys.ALGO]
+                param.update({"files": [infer_filename]})
+                pred = infer_instance.predict(param)
+                preds.append(pred[0])
+            output = sum(preds) / len(preds)
```
wyli: this is an average, how about majority voting which is also common...

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -0,0 +1,227 @@
...
+        self.ensemble: ScriptEnsemble
+        self.data_src_cfg = ConfigParser(globals=False)
+
+        if data_src_cfg_filename is not None and os.path.exists(str(data_src_cfg_filename)):
+            self.data_src_cfg.read_config(data_src_cfg_filename)
+
+        for h in history:
+            # load inference_config_paths
+            # raise warning/error if not found
+            if len(h.keys()) > 1:
```
wyli, line 197: could be simplified:
```suggestion
            if len(h) > 1:
```

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -0,0 +1,227 @@
...
+                raise ValueError(f"{h} should only contain one set of genAlgo key-value")
+
+            name = list(h.keys())[0]
+            gen_algo = h[name]
+            best_metric = gen_algo.get_score()
```
wyli, line 202: hope this could support a tuple of numbers, for example, this might include model size and inference speed, they could also be part of the model ranking.

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -0,0 +1,227 @@
...
+            algo_path = gen_algo.output_path
+            infer_path = os.path.join(algo_path, "scripts", "infer.py")
+
+            if not os.path.isdir(algo_path):
+                raise ValueError(f"{gen_algo.output_path} is not a directory. Please check the path.")
```
wyli, line 207: perhaps put warnings in this method instead of `ValueError`, this method doesn't strictly require these files and folders..

wyli (commented):
thanks, I put some comments inline...please feel free to ignore any of them

monai/apps/auto3dseg/data_analyzer.py
wyli: it is needed for exporting analyzer's yaml output, I added `if optional_import("yaml")[1]:` otherwise it can't pass the min tests for some reason.

mingxin-zheng: The multi_representer is a necessary line to avoid the str(MetaKeys) calls across all the scripts. 

The added optional_import seems to be a fix to tell if pyyaml is available. I remember the StrEnum works okay with json so pure JSON use case is possible.

monai/apps/auto3dseg/ensemble_builder.py
mingxin-zheng: Fixed. Thanks!

monai/apps/auto3dseg/bundle_gen.py
mingxin-zheng: Yes, the downloaded files need to be managed better. 

My latest commit put the downloaded files to `algo_path/algorithm_templates`, which user can specify `algo_path` or it will be the user's current working directory. 

If we put the to a subfolder of the torch hub, the process will be a bit more invisible (ofc we can log info and let user know too). I personally likes more transparency but don't know which way is better.

```python
class BundleGen(AlgoGen):
    def __init__(self, algo_path: str = ".", algos=None, data_stats_filename=None, data_src_cfg_name=None):
```

monai/apps/auto3dseg/ensemble_builder.py
Nic-Ma: I think we can support the `ensemble method` as a callable argument?

Thanks.

monai/apps/auto3dseg/bundle_gen.py
wyli: ok sure, I think user's working directory is better, here I was mainly thinking about caching the original system default values somewhere to avoid repeated downloading..

monai/apps/auto3dseg/bundle_gen.py
mingxin-zheng: @Nic-Ma and I happened to discuss UUID on another day. Looks like a single f_id satisfy the need of multi-fold, but when HPO comes in, better identifier is needed, and it would be best if the name can tell the experiment details.

monai/apps/auto3dseg/ensemble_builder.py
mingxin-zheng: Fixed. Thanks!

monai/apps/auto3dseg/ensemble_builder.py
mingxin-zheng: That's right. It was from the previous version when I wasn't using BundleAgo

monai/apps/auto3dseg/ensemble_builder.py
mingxin-zheng: Incorporated the change by manual replacing. thanks

monai/apps/auto3dseg/ensemble_builder.py
mingxin-zheng: Or just uses a dict for the scores?

monai/apps/auto3dseg/ensemble_builder.py
mingxin-zheng: Okay, I will make the ‘predict’ to a ‘__call__’ function.

mingxin-zheng: In the upcoming commit, I added the majority voting, made the BundleEnsemble a callable and some additional parameters to define the prediction.

monai/apps/auto3dseg/bundle_gen.py
mingxin-zheng: I added a few line to copy the existing experiments and results to a backup (.bak) and a warning message.

monai/apps/auto3dseg/bundle_gen.py
mingxin-zheng: I did some simple tests. It seems overall consistent with Python behavior ("aaa/bbb", "aaa\\bbb" okay but not "a\b"). Such test seems standalone from the ensemble module and we can construct something else to test the subprocess call.

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -0,0 +1,389 @@
...
+
+
+# path to download the algo_templates
+algo_zip = "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/algo_templates.tar.gz"
+
+md5 = "67f434ce3c44800cfb6341d6278d635b"
+
+# default algorithms
+default_algos = {
+    "unet": dict(
```
dongyang0122, line 279: is this unet the same as the one in gitlab repo?

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -0,0 +1,389 @@
...
+
+    def get_data_src(self, fold_idx=0):
+        """Get the data source filename"""
+        return self.data_src_cfg_filename
+
+    def get_history(self, *args, **kwargs) -> List:
+        """get the history of the bundleAlgo object with their names/identifiers"""
+        return self.history
+
+    def generate(self, output_folder=".", fold_idx: Union[int, Sequence[int]] = 0):
```
dongyang0122, line 378: it is better to have number of folds as input here. for now, we are experimenting with 5-fold. users may use it for different number of folds.

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -0,0 +1,251 @@
...
+        for i in range(len(files)):
+            print(i)
+            preds = []
+            infer_filename = self.infer_files[i]
+            for algo in self.algo_ensemble:
+                infer_instance = algo[BundleEnsembleKeys.ALGO]
+                param.update({"files": [infer_filename]})
+                pred = infer_instance.predict(param)
+                preds.append(pred[0])
+            outputs.append(sum(preds) / len(preds))
```
dongyang0122: we can design additional class method for aggregation of all probability maps. now the default way is averaging. we should be able to let user change aggregation methods for their own purpose. and we can even add a tutorial to demonstrate that capacity.

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -0,0 +1,251 @@
...
+        Set the files to perform model inference.
+
+        Args:
+            dataroot: the path of the files
+            data_src_cfg_file: the data source file path
+        """
+        with open(data_src_cfg_file) as f:
+            datalist = json.load(f)
+
+        for d in datalist["testing"]:
```
dongyang0122, line 82: please add key of the datalist as a method argument.

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -0,0 +1,251 @@
...
+        self.n_best = n_best
+
+    def sort_score(self):
+        """
+        Sort the best_metrics
+        """
+        scores = concat_val_to_np(self.algos, [BundleEnsembleKeys.SCORE])
+        return np.argsort(scores).tolist()
+
+    def rank_algos(self):
```
dongyang0122: we should have an additional method for ranking in each individual folds. we chose best model for each cross-validation fold in previous public challenges.

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -0,0 +1,79 @@
...
+    def select_ensemble_method(self, ensemble: EnsembleMethods):
+        """Select the ensemble method"""
+
+        self.ensemble = ensemble
+
+
+    def __call__(self, **kwargs):
+        """Get the score if it hasn't yet, and ensemble."""
+        for (name, model) in self.validate_tasks:
+            self.best_metrics.update({name, model(self.datsets['validate'])})
```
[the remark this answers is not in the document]
mingxin-zheng: Hi @dongyang0122 , this is marked as "outdated" so it is from a few versions ago. The scripts was very different back then and `datsets` was dropped

monai/apps/auto3dseg/__main__.py
```diff
@@ -10,10 +10,19 @@
 # limitations under the License.
```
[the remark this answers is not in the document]
mingxin-zheng: This script is actually essential. If we move it to tutorial, then we lose the ability to call these functions in the command lines via Fire.

monai/apps/auto3dseg/bundle_gen.py
mingxin-zheng: I did not validate but tried my best to replicate the all the details.

monai/apps/auto3dseg/bundle_gen.py
mingxin-zheng: Fixed. Thanks!

monai/apps/auto3dseg/ensemble_builder.py
mingxin-zheng: Fixed. Thanks!

monai/apps/auto3dseg/ensemble_builder.py
mingxin-zheng: Added a new class to address the need. Thanks for the suggestion!

```python
class BundleEnsembleBestByFold(BundleEnsemble):
    """
    Ensemble method that select the best models that are the tops in each fold.

    Args:
        n_fold: number of cross-valiation folds used in training
    """

    def __init__(self, n_fold: int = 5):
```

tangy5 (commented):
The current codes looks good. 
Curious about whether the ensemble_builder.py is tested with 5 folds cross validation? 
Might need to test the ensemble after merging the PR and test with real tasks.

wyli (approved):
thanks, merging this as an initial implementation

PR #5015, opened 2022-08-26
Sobel Gradients Operator

### LINKED ISSUE #4997: Sobel gradient calculation transform
HoVerNet losses requires calculation of numerical gradient of HoVerMaps in horizontal and vertical direction. This can be achieved by a Sobel kernel. skimage implementation uses 3x3 Sobel kernel but this pipeline requires a 5x5 kernel. Thus having a Sobel filter to calculate gradients with a flexible kernel size would be very helpful.

drbeh:
Fixes #4997

### Description
Implements Sobel gradients operators for a given size of kernel.

### Types of changes


KumoLiu (approved):
It looks good to me, thanks!

drbeh:
> thanks, I thought sobel kernel is always 3x3, why is there a size parameter, and it's unclear to me what sobel gradient is...perhaps a reference here?

You are right, Sobel kernels are 3x3 but they can be generalized to a larger kernel size with different approaches. You can find a nice writeup on how to derive larger size kernels following the same property of 3x3 Sobel kernels: https://stackoverflow.com/a/41065243

Following the original implementation of HoVerNet, we are going to use 5x5 Sobel kernels in HoVerNet pipeline, and they have implemented kernels similar to this PR. Here is their implementation:
https://github.com/vqdang/hover_net/blob/842964dc5d26ffe196d236d16fad16643a103850/models/hovernet/utils.py#L124-L145
https://github.com/vqdang/hover_net/blob/842964dc5d26ffe196d236d16fad16643a103850/models/hovernet/utils.py#L150

PR #5030, opened 2022-08-29
Update HoVerNet model with enum keys

### LINKED ISSUE #5029: HoVerNet Model branch naming
HoVerNet model has hard coded branch names that makes it hard to use these branch names across other components, for instance when using in a loss function. It would be great to have enum keys to define these branch names to be usable elsewhere.

drbeh:
Fixes #5029

### Description
Add enums to define different branches of HoVerNet model.

### Status
**Ready**

### Types of changes


monai/networks/nets/hovernet.py
```diff
@@ -391,6 +391,19 @@ class Mode(Enum):
...
+        Three branches of HoVerNet model, which results in three outputs:
+        `HOVER` is horizontal and vertical regressed gradient map of each nucleus,
+        `NUCLEUS` is the segmentation of all nuclei, and
+        `TYPE` is the type of each nucleus.
+
+        """
+
+        HOVER = "horizontal_vertical"
+        NUCLEUS = "nucleus_prediction"
+        TYPE = "type_prediction"
```
KumoLiu, line 405: I would suggest the name to be consistent with the paper which users can easily find assistance from the paper.
```
HOVER = "horizontal_vertical"
NP = "nuclear_pixel"
NC = "nuclear_classification"
```

KumoLiu:
Looks good to me except for the leave comment, thanks!

monai/networks/nets/hovernet.py
JHancox: I like the idea of referring to the original paper but the reason for using keys, rather than just numeric indices, is to make clear what each branch does. NC, NP and HOVER are not very clear in their meaning, whereas the the text is. If we are going to use keys then I think drBeh's are clearer.

drbeh: Thanks @KumoLiu and @JHancox for your comments. 
In my opinion, we should either be descriptive or minimalistic, so I would suggest to take the following keys, which are easier to use, have the same structure (abbreviation of the descriptive names), and mostly matches the paper. What do you think?
```python
HV = "horizontal_vertical"
NP = "nuclear_pixel"
NC = "nuclear_classification"
```

drbeh: For the description to use, for instance "type_prediction" or "nuclear_classification", I would rely on @JHancox's opinion since he has more experience with this model.

KumoLiu: I agree with the change of **key** and will be willing to listen to @JHancox's opinion on the description!

drbeh: This information is represented as a class instance of `HoVerNet` model, and not being reflected in MetaTensor. MetaTensor carries information about the image. Anyways, the end user does not need to deal with the branch names and needs to work only with high-level API but developers can access the branch names to use it in other related components like for testing:
https://github.com/Project-MONAI/MONAI/blob/67cee72eaf68f5b13bfa0326eb8253e00d4cab22/tests/test_hovernet.py#L51

PR #5038, opened 2022-08-31
Make the flexible unet more extensible to different backbones

### LINKED ISSUE #5011: Make the Flexible U-Net easily extensible for new backbones
Currently, the FlexibleUNet structure only supports the efficient-net series as the backbone and it is hard to extend to other network structures. A more extensible and convenient way to add more benchmark or user defined backbones will make it more flexible.   

1. A base backbone/encoder class that defines interfaces like the number of output feature maps, a list of output feature map channels, string names of backbones and so on.

2. A register who can add and update backbones informations to a dict.

With these two components, users can implement their own backbones class derived from the base backbone class and regist them to the backbone dict. Then Flexible U-Net will use this dict to generate backbones/encoders.

binliunls:

Fixes #5011  .

### Description
Currently, the FlexibleUNet structure only supports the efficient-net series as the backbone and it is hard to extend to other network structures. A more extensible and convenient way to add more benchmark or user defined backbones will make it more flexible. I plan to make it by doing steps below

1. A base backbone/encoder class that defines interfaces like the number of output feature maps, a list of output feature map channels, string names of backbones and so on.

2. A register who can dynamically regist backbones to a dict that will be used during the flexible unet initialization.

### Status
**Work in progress**

### Types of changes


monai/networks/blocks/encoder.py
```diff
@@ -0,0 +1,61 @@
...
+    @abstractmethod
+    def get_output_feature_number_list(cls) -> List[int]:
+        """
+        Get number of output feature.
+        """
+        pass
+
+    @classmethod
+    @abstractmethod
+    def get_encoder_name_string_list(cls) -> List[str]:
```
Nic-Ma, line 57: I think better to change to `get_encoder_names(cls) -> List[str]`.
No need the `string` in function name and should explain why it's a list.
What do you think?

Thanks.

monai/networks/nets/flexible_unet.py
```diff
@@ -243,26 +249,26 @@ def __init__(
...
-        encoder_channels = _get_encoder_channels_by_backbone(backbone, in_channels)
-        self.encoder = EfficientNetBNFeatures(
-            model_name=model_name,
-            pretrained=pretrained,
-            in_channels=in_channels,
-            spatial_dims=spatial_dims,
-            norm=norm,
-            adv_prop=adv_prop,
+        encoder_parameters = encoder["parameter"]
+        assert (
```
Nic-Ma, line 262: `assert` is not recommended to use in python, please use below logic instead of `assert`:
```py
if "spatial_dims" not in encoder_parameters or "in_channels" not in encoder_parameters or "pretrained" not in encoder_parameters:
    raise ValueError("XXXX")
```

monai/networks/nets/flexible_unet.py
```diff
@@ -9,49 +9,55 @@
...
+        self.register_dict = {}
 
-    Returns:
-        A tuple of output feature map channels' length .
-    """
-    encoder_channel_tuple = encoder_feature_channel[backbone]
-    encoder_channel_list = [in_channels] + list(encoder_channel_tuple)
-    encoder_channel = tuple(encoder_channel_list)
-    return encoder_channel
+    def regist_class(self, name: Type):
```
Nic-Ma, line 36: Please consider to support `str` as the arg, in case we use it in the JSON config of bundle.

Thanks.

wyli:
please also help address these style issues https://deepsource.io/gh/Project-MONAI/MONAI/run/27214ec8-bf5b-4c2d-92d2-450b10308d91/python/

monai/networks/blocks/encoder.py
```diff
@@ -0,0 +1,84 @@
...
+
+from abc import ABCMeta, abstractmethod
+from typing import Dict, List, Tuple
+
+__all__ = ["BasicEncoder"]
+
+
+class BasicEncoder(metaclass=ABCMeta):
+    """
+    Abstract class defines inferface of encoders in flexible unet.
```
wyli, line 20: I think the class name could be `BaseEncoder`? 

typo "inferface" => "interface"

monai/networks/nets/flexible_unet.py
```diff
@@ -9,49 +9,55 @@
...
+        """
+        Regist a given class to the encoder dict. Please notice that input class must be a
+        subclass of BasicEncoder.
+        """
+        if not issubclass(name, BasicEncoder):
+            raise Exception("An encoder must derive from BasicEncoder class.")
```
wyli, line 42: perhaps not a fatal error if the user doesn't use the base class

monai/networks/nets/flexible_unet.py
binliunls: Hi @Nic-Ma ,
     I have added string input to regist_class function.
Thanks,
Bin

tests/test_flexible_unet.py
```diff
@@ -10,23 +10,161 @@
...
+        feature_list.append(x)
+        x = self.layer3(x)
+        feature_list.append(x)
+        x = self.layer4(x)
+        feature_list.append(x)
+
+        return feature_list
+
+
+BACKBONE.regist_class(ResNetEncoder)
```
Nic-Ma, line 156: Could you please help add a test case to ensure the `FlexibleUNet` with registered `ResNetEncoder` can do `forward()` computation?

Thanks in advance.

Nic-Ma (approved):

monai/networks/blocks/encoder.py
```diff
@@ -0,0 +1,84 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import ABCMeta, abstractmethod
+from typing import Dict, List, Tuple
+
+__all__ = ["BaseEncoder"]
+
+
+class BaseEncoder(metaclass=ABCMeta):
```
Nic-Ma, line 18: `encoder` seems like too general a name to me, I would suggest `FlexibleUNetEncoder`.
@wyli What do you think?

Thanks.

wyli: I'm fine with either...the methods here look very general though

tests/test_flexible_unet.py
binliunls: Hi @Nic-Ma ,
      I have added the resnet test in [test cases] 
(https://github.com/binliunls/MONAI/blob/d981093403bdad9abfca5ddbbb9c8035ba4e9664/tests/test_flexible_unet.py#L218) and will update the comment. I think this test should be able to verify forward function of  resnet encoder.
Thanks,
Bin

monai/networks/nets/__init__.py
```diff
@@ -38,10 +38,11 @@
     EfficientNet,
     EfficientNetBN,
     EfficientNetBNFeatures,
+    EfficientNetEncoder,
     drop_connect,
     get_efficientnet_image_size,
 )
-from .flexible_unet import FlexibleUNet
+from .flexible_unet import BACKBONE, FlexibleUNet, Register
```
wyli, line 45: `monai.networks.nets.Register` sounds very general, but the actual implementation is only for the FlexibleUNet... this might be an issue cc @Nic-Ma

Nic-Ma: Hi @wyli ,

I didn't put comment for this one because it's within the `flexible_unet.py` file, but I agree a more specific name would be better.

Thanks.

Nic-Ma: Maybe we can call it `EncoderRegister` or some other name?

Thanks.

wyli: ok, or removing `from .flexible_unet import BACKBONE, Register` is also fine?

Nic-Ma: I would suggest: EncoderRegister or make a alias for the `FlexibleUNet` vs `FlexUNet`?
@wyli What do you think?

Thanks.

Nic-Ma: And `FLEXUNET_BACKBONE`。

binliunls: Hi @Nic-Ma ,
      I have added the alias and changed the backbone name.
Thanks.

Nic-Ma (approved):

Issue #4891, opened 2022-08-11
Auto3D algorithm selector and configer 


mingxin-zheng:
**Is your feature request related to a problem? Please describe.**
This is a part of the overall ticket https://github.com/Project-MONAI/MONAI/issues/4743
Task
- Implement a algorithm selector acting as a placeholder for future smarter algorithm selection
- Implement a config generator to create monai.bundle config files from templates

**Describe the solution you'd like**
N/A

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #5042, opened 2022-08-31
4891 base class algo generator

wyli:
### Description
part of #4891 with some misc fixes and tests.
adds:
- Algo class
- AlgoGen class

### Status
**Readyd**

### Types of changes

### DIFF
--- a/monai/auto3dseg/algo_gen.py
+++ b/monai/auto3dseg/algo_gen.py
@@ -1,0 +1,110 @@
+# Copyright (c) MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from monai.transforms import Randomizable
+
+
+class Algo:
+    """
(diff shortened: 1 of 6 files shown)


monai/auto3dseg/algo_gen.py
```diff
@@ -0,0 +1,108 @@
...
+            params: key-value pairs of input parameters for the predicting pipeline.
+        """
+        pass
+
+    def get_score(self, *args, **kwargs):
+        """Returns the model quality measurement based on training and validation datasets."""
+        pass
+
+
+class AlgoGen(Randomizable):
```
dongyang0122, line 48: why the algogen is randomizable?

monai/auto3dseg/algo_gen.py
```diff
@@ -0,0 +1,108 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from monai.transforms import Randomizable
+
+
+class Algo:
```
dongyang0122, line 15: if algo and algogen are independent classes, shall we separate them into two files?

monai/auto3dseg/algo_gen.py
wyli: so that it's possible to draw random samples from a probabilistic search space

monai/auto3dseg/algo_gen.py
wyli: there are cases where we put the base classes in the same file, for example https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/transform.py. but I think it's just a style preference, I'm fine with either way. @Nic-Ma @mingxin-zheng

Nic-Ma (approved):

PR #5064, opened 2022-09-01
LogfileHandler

ericspod:

### Description
Adds a handler for directing logging information to a file once an `Engine` object starts and then doing the cleanup when it completes.

### Status
**Ready**

### Types of changes


monai/handlers/logfile_handler.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def _start(self, engine: Engine) -> None:
+        if self.create_dir and not os.path.exists(self.output_dir):
+            os.makedirs(self.output_dir, exist_ok=True)
+
+        self.handler = logging.FileHandler(os.path.join(self.output_dir, self.filename))
+        self.handler.setLevel(self.loglevel)
+        self.handler.setFormatter(logging.Formatter(self.formatter))
+        assert self.logger is not None
+        self.logger.addHandler(self.handler)
```
Nic-Ma, line 78: Hi @wyli @ericspod ,

I remember we discussed the policy of stats handler, we should leverage the standard way of python logging:
Let the users to set log handler (not ignite handler) to the logger outside of the `StatsHandler`, then we can easily log info to files or console:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/handlers/stats_handler.py#L114
So we removed all the `handler` arg in ignite handlers, suppose to use `logging.getLogger(name)` to control it.
What do you think?

Thanks.

ericspod: I was thinking that different engines might end up with different logger objects so doing it this way ensured you got the right one, it would be compatible with vanilla Ignite engine objects. The problem I encountered was to do the log file handling in bundles wasn't straight forward.

wyli (approved):
looks good to me, could you replace `assert` with valueerror+some meaningful error messages?

ericspod:
Actually the asserts were for mypy to figure out what state the handler would be in when the Ignite events occur. It's doing some sort of data flow analysis and using the asserts as hints about state.

wyli:
I think these recommended methods should be used for type narrowing https://mypy.readthedocs.io/en/latest/type_narrowing.html

ericspod:
The page does also mention using assert for this which seems to be less code, the changes I did here are move verbose now.

Nic-Ma (approved):
Thanks for the update.
The use case makes sense to me.
Please try to use it in the tutorial examples or some model-zoo bundles in the future.

Thanks.

PR #5071, opened 2022-09-01
Use Ignite's interrupt api in MonaiAlgo

### LINKED ISSUE #4554: Add interrupt/terminate option to Trainers & Evaluators
It should be possible to abort/finalize a running Trainer by calling the API (rather than ctr+C). This will be helpful if the Trainer needs to be executed remotely, such as in federated learning (FL) scenarios.

Add abort() and finalize() functions to the [Trainer](https://github.com/Project-MONAI/MONAI/blob/d34fa14eed3191126fc1bc1a9751e69e0396f78b/monai/engines/trainer.py#L45) class (or potentially its base class). Note, finalize() should terminate the training completely, while abort() should allow later continue of where it was aborted(), by calling run() again.

For example, an ignite-based Trainer support abort() and finalize() calls could be implemented as such (Currently used in MONAI-FL's [MonaiAlgo class](https://github.com/Project-MONAI/monai-fl/blob/main/monaifl/monai_algo.py); private repo - contact me if you need access)
```
    def abort(self):
        self.trainer.terminate()
        # save current iteration for next round
        setattr(self.trainer.state, "dataloader_iter", self.trainer._dataloader_iter)

        if self.trainer.state.iteration % self.trainer.state.epoch_length == 0:
            # if current iteration is end of 1 epoch, manually trigger epoch completed event
            self.trainer._fire_event(Events.EPOCH_COMPLETED)

    def finalize(self):
        self.trainer.terminate()
```

n/a

**Additional context**
n/a

holgerroth:
Fixes #4554. 

Requires latest Ignite RC.

Also improves ClientAlgo docstrings formatting.

### Description

### Status
**Ready**

### Types of changes


monai/fl/client/client_algo.py
```diff
@@ -20,43 +20,69 @@ class ClientAlgo(abc.ABC):
...
+        Train network and produce new network from train data.
+
+        Args:
+            data: ExchangeObject containing current network weights to base training on.
+            extra: optional extra information
+
+        Returns:
+            None
         """
         raise NotImplementedError
```
Nic-Ma, line 71: I would suggest to add some error message, like: https://github.com/Project-MONAI/MONAI/blob/dev/monai/engines/workflow.py#L293

Thanks.

monai/fl/client/client_algo.py
```diff
@@ -20,43 +20,69 @@ class ClientAlgo(abc.ABC):
...
+    of the class object.
     """
 
     def initialize(self, extra=None):
-        """call to initialize the ClientAlgo class"""
+        """
+        Call to initialize the ClientAlgo class
+
+        Args:
+            extra: optional extra information, e.g. dict of `ExtraItems.CLIENT_NAME` and/or `ExtraItems.APP_ROOT`
```
Nic-Ma, line 36: Should we require the type of `extra` in the base class?
For example: `Optional[dict]`.

Thanks.

Nic-Ma:
Hi @holgerroth @wyli ,

I think we should update the `min_version` of ignite in the monai_algo.py file to `0.4.10`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/fl/client/monai_algo.py#L41
Other places keep the original min version.
Then when ignite released 0.4.10 next Monday, let's update the ignite version in all the docker and CI configs.
Now the min required version is `0.4.4`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/config/deviceconfig.py#L261
And the docker and CI version is `0.4.9`:
https://github.com/Project-MONAI/MONAI/blob/dev/requirements-dev.txt#L3

What do you think?

Thanks.

holgerroth:
> Hi @holgerroth @wyli ,
> 
> I think we should update the `min_version` of ignite in the monai_algo.py file to `0.4.10`: https://github.com/Project-MONAI/MONAI/blob/dev/monai/fl/client/monai_algo.py#L41 Other places keep the original min version. Then when ignite released 0.4.10 next Monday, let's update the ignite version in all the docker and CI configs. Now the min required version is `0.4.4` for all the places: https://github.com/Project-MONAI/MONAI/blob/dev/monai/config/deviceconfig.py#L261 And the docker and CI version is `0.4.9`: https://github.com/Project-MONAI/MONAI/blob/dev/requirements-dev.txt#L3
> 
> What do you think?
> 
> Thanks.

Actually, we don't need to import Ignite anymore in this class. I removed it.

monai/fl/client/client_algo.py
holgerroth: Good point. I added it.

monai/fl/client/client_algo.py
holgerroth: Thanks. I added the typing info.

vfdev-5:
FYI, we just released [0.4.10](https://github.com/pytorch/ignite/releases/tag/v0.4.10), so you can check and also update MONAI deps.

Nic-Ma (approved):
I have updated the ignite dependency in the PR.

Thanks.

Nic-Ma:
Hi @wyli ,

Could you please help double confirm it?

Thanks in advance.

wyli:
> Hi @wyli ,
> 
> Could you please help double confirm it?
> 
> Thanks in advance.

I see there are docs-test errors: https://github.com/Project-MONAI/MONAI/runs/8197608122?check_suite_focus=true

monai/fl/client/client_algo.py
```diff
@@ -20,60 +21,88 @@ class ClientAlgo(abc.ABC):
...
+
+        Args:
+            data: ExchangeObject containing current network weights to base training on.
+            extra: optional extra information
+
+        Returns:
+            None
         """
-        raise NotImplementedError
+        raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
```
vfdev-5, line 72: I wonder whether this error will be ever seen as method is decorated with `@abc.abstractmethod` ?
As we can't instantiate `ClientAlgo`, the only way as far as I can see to raise this error is to call the method directly as static : `ClientAlgo.train(None, None)`...

Nic-Ma: Hi @vfdev-5 ,

That's a good point.
I removed the decorator in the latest commit, it's same as the `Workflow` class now:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/engines/workflow.py#L280
And tested as below:
```py
>>> from monai.fl.client.client_algo import ClientAlgo
>>>
>>>
>>> class Test(ClientAlgo):
...     pass
...
>>> t = Test()
>>> t.train(None)
  File "/workspace/data/medical/MONAI/monai/fl/client/client_algo.py", line 71, in train
NotImplementedError: Subclass Test must implement this method.
```

Thanks.

vfdev-5: I see. So, in this case is there a reason to derive `ClientAlgo` from `abc.ABC` ?

Nic-Ma: You are right, also removed `ABC`.

Thanks.

PR #5075, opened 2022-09-02
5061 align corners and recompute_affine for spacing

### LINKED ISSUE #5061: Spacing transform changes physical extent of images
**Describe the bug**
After applying the Spacing transform, the physical extent that an image covers is changing.

**To Reproduce**
```
import numpy as np
from monai.transforms import Spacing

pixdim_new = [2.0, 1.0, 0.2]
pixdim_old = [1.0, 1.0, 1.0]
affine = np.diag(pixdim_old + [1])
data = np.random.randn(1, 128, 128, 32)

img = MetaTensor(data, affine=affine).to("cpu")
res: MetaTensor = Spacing(pixdim=pixdim_new)(img)

np.testing.assert_allclose(
    np.array(img.shape[1:]) * pixdim_old,
    np.array(res.shape[1:]) * pixdim_new
)
```

**Expected behavior**
Resampled image covers same physical space as original volume, in particular if it is possible because the resampling factors are integers. In fact, during upsampling (last dimension), the physical extent is too small.
I might understand the behavior now if `align_corners` is true, in the sense that the centers of the corner pixels define the extent and not the borders.
From `align_corners` documentation: "if set to False, they are instead considered as referring to the corner points of the input’s corner pixels, making the sampling more resolution agnostic"

**Screenshots**
none

**Environment**

```
MONAI version: 0.9.1
Numpy version: 1.21.6
Pytorch version: 1.12.1+cpu
MONAI rev id: 356d2d2f41b473f588899d705bbc682308cee52c

```

**Additional context**

wyli:

Fixes #5061

### Description
computing output shape considering `align_corners`, recompute the output affine based on output shapes.
a new `scale_extent` parameter is included to choose between the two resampling coordinate conventions.

### Status
**Ready**

### Types of changes

### DIFF
--- a/monai/apps/reconstruction/networks/nets/complex_unet.py
+++ b/monai/apps/reconstruction/networks/nets/complex_unet.py
@@ -83,7 +83,8 @@ class ComplexUnet(nn.Module):
             params = [p.shape for p in conv_net.parameters()]
             if params[0][1] != 2:
                 raise ValueError(f"in_channels should be 2 but it's {params[0][1]}.")
-            self.unet = conv_net  # type: ignore
+            self.unet = conv_net
+
         self.pad_factor = pad_factor
 
     def forward(self, x: Tensor) -> Tensor:
--- a/monai/auto3dseg/utils.py
+++ b/monai/auto3dseg/utils.py
@@ -168,7 +168,8 @@ def concat_val_to_np(
     elif ragged:
         return np.concatenate(np_list, **kwargs)  # type: ignore
     else:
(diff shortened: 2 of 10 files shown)


ericspod (approved):
It looks correct to me given the issue it addresses though @yanlend may want to try running the code to double check for his use case.

yanlend:
I ran the code with the new spatial offset computation and it behaves as expected. However, we still disagree about the meaning of the `align_corners` parameter.

I have not checked the `recompute_affine` part.

PR #5079, opened 2022-09-02
Default prepare batch update

ericspod:
### Description
This allows `default_prepare_batch` to handle batches containing a single tensor or a pair of tensors. Normally if one uses array transforms you'll get an error saying the default batch function needs dictionaries so this will allow such transform sequences to be used without modification.

### Status
**Work in progress**

### Types of changes


Nic-Ma (approved):
Looks good to me.
Should we also update the `PrepareBatch` class and its subclasses?
https://github.com/Project-MONAI/MONAI/blob/dev/monai/engines/utils.py#L120

Thanks.

ericspod:
I updated the classes with more explanatory commentary and correct type annotations.

Nic-Ma (approved):

PR #5092, opened 2022-09-06
Remove Status from PR Template

drbeh:
### Description
This PR remove "Status" from PR template. The Status was very useful before GitHub introduced *Draft* PR. I believe right now it is redundant to have this and sometimes conflicting (e.g., Status Ready <> Draft PR), so I believe it is time to remove the "Status" from PR template.

### Types of changes


drbeh:
@Nic-Ma @wyli @ericspod @rijobro, what do you think?

wyli (approved):
agreed, I think we can simplify the form

ericspod:
Fair enough, we should also not put "WIP" or something similar in titles either.

PR #5093, opened 2022-09-06
Updated Contrastive Loss batch Size

### LINKED ISSUE #5070: ContrastiveLoss with dynamic batchsize
When we instantiate an instance of type class `ContrastiveLoss` we need to pass in the parameter for `batch_size`. This assumes that the batch size remains constant all through the training(which is not always the case). 

As we know and have fixed in the [docs](https://docs.monai.io/en/latest/losses.html#monai.losses.ContrastiveLoss) that the forward accepts inputs of shape `B[F]`, why not pick up the Batch size dynamically in the forward. Is there a reason why it is not designed in that fashion?

I would be happy to put a pull request if it is okay.

a-parida12:
Fixes #5070 .

### Description

The batch size for the contrastive loss is estimated automatically

### Types of changes


wyli (approved):
Thanks, it looks good to me. merging this PR if all tests work fine cc @finalelement .

PR #5104, opened 2022-09-07
adds a debugging flag for bundle parser

wyli:

### Description
adds a `_debug_` keyword, so that when it's set to true, the program stops at the component init stage and enter the debug mode , usage:

```yaml
inferer:
  _target_: SlidingWindowInferer
  _debug_: true
  roi_size: [96, 96, 96]
  sw_batch_size: 4
  overlap: 0.5
```

alternatively at the CLI set `MONAI_DEBUG_CONFIG=1`, for example:

```bash
MONAI_DEBUG_CONFIG=1 python -m my_test_5 run --config_file=my_test_5.yaml
```

### Types of changes

### DIFF
--- a/monai/bundle/config_item.py
+++ b/monai/bundle/config_item.py
@@ -11,7 +11,6 @@
 
 import ast
 import inspect
-import os
 import sys
 import warnings
 from abc import ABC, abstractmethod
@@ -19,7 +18,7 @@ from importlib import import_module
 from typing import Any, Dict, List, Mapping, Optional, Sequence, Union
 
 from monai.bundle.utils import EXPR_KEY
-from monai.utils import ensure_tuple, first, instantiate, optional_import
+from monai.utils import ensure_tuple, first, instantiate, optional_import, run_debug, run_eval
 
 __all__ = ["ComponentLocator", "ConfigItem", "ConfigExpression", "ConfigComponent", "Instantiable"]
 
@@ -316,7 +315,7 @@ class ConfigExpression(ConfigItem):
     """
 
     prefix = EXPR_KEY
-    run_eval = os.environ.get("MONAI_EVAL_EXPR", "1") != "0"
+    run_eval = run_eval
 
     def __init__(self, config: Any, id: str = "", globals: Optional[Dict] = None) -> None:
         super().__init__(config=config, id=id)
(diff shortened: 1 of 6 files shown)


wyli:
two ways to enter the debug mode:

- adds `_debug_: true` to the component in the config
- run with a system `MONAI_DEBUG_CONFIG` flag when starting a bundle.configparser from CLI:  
```bash
MONAI_DEBUG_CONFIG=0 python -m my_test_5 run --config_file=my_test_5.yaml
```

monai/bundle/config_item.py
```diff
@@ -365,7 +364,15 @@ def evaluate(self, globals: Optional[Dict] = None, locals: Optional[Dict] = None
...
                     warnings.warn(f"the new global variable `{k}` conflicts with `self.globals`, override it.")
                 globals_[k] = v
-        return eval(value[len(self.prefix) :], globals_, locals)
+        if not run_debug:
+            return eval(value[len(self.prefix) :], globals_, locals)
+        warnings.warn(
+            f"\n\npdb: value={value}\n"
+            f"See also Debugger commands documentation: https://docs.python.org/3/library/pdb.html\n"
+        )
+        import pdb
```
Nic-Ma, line 373: Should we import `pdb` at the beginning of the file?

Thanks.

monai/utils/module.py
```diff
@@ -220,6 +228,14 @@ def instantiate(path: str, **kwargs):
...
         raise ModuleNotFoundError(f"Cannot locate class or function path: '{path}'.")
     try:
+        if kwargs.pop("_debug_", False) or run_debug:
+            warnings.warn(
+                f"\n\npdb: instantiating component={component}\n"
+                f"See also Debugger commands documentation: https://docs.python.org/3/library/pdb.html\n"
+            )
+            import pdb
+
+            pdb.set_trace()
```
Nic-Ma, line 238: Is it possible to add some unit tests to cover the behavior here?
I didn't test `pdb` locally, I think it's an interactive operation?

Thanks.

monai/bundle/config_item.py
wyli: don't want to introduce this tool to the module, it's mostly not used, if we find it's frequently I'll update this.

Nic-Ma (approved):
Thanks for the quick update, looks good to me.
Could you please help also update the `get_started` tutorial of bundle to show this feature?

Thanks in advance.

PR #5106, opened 2022-09-08
Fix PatchMerging bug

hxhxhx88:
### Description

Fix 3D PatchMerging which has incorrect indices.


wyli:
this version of patch merging is intentionally kept in case of breaking existing models https://github.com/Project-MONAI/MONAI/blob/c0c5f28a2b5fe6c2d53da18495c2c883026ee50e/monai/networks/nets/swin_unetr.py#L729

I'm closing PR, but thanks for looking into this

PR #5111, opened 2022-09-08
Implements SplitDimToListd

### LINKED ISSUE #5109: SplitDimToListd to store list of dictionaries for SplitDim
Our multiple instance learning uses a list of patches (a bag of images) for training the model. Although we have a transform that splits the image into patches for our pipeline, this pipeline cannot be used for other use cases that do not rely on this transform, like this conversation. `SplitDim` is useful here but `SplitDimd` splits and stores into dictionary keys instead of list of dictionaries with the same key, which is required for this pipeline.

drbeh:
Fixes #5109 

### Description

This PR implements `SplitDimToListd` that uses `SplitDim` and for a given dictionary of tensors, split the tensor along the given dimension and store it as a list of dictionaries with the same keys but having the a split of the tensor.

### Types of changes


drbeh:
closing in favor of other solution #5115

PR #5115, opened 2022-09-09
Update `SplitDimd` to output lists

### LINKED ISSUE #5109: SplitDimToListd to store list of dictionaries for SplitDim

drbeh:
Fixes #5109 

### Description
Updates `SplitDimd` to have the option of outputting a list of dictionaries.

### Types of changes


drbeh:
@Nic-Ma @KumoLiu for reviewing this PR, could you please just select the first two commits, and let me know your comments. Thanks

KumoLiu (dismissed):

Nic-Ma:
Hi @drbeh @wyli ,

@KumoLiu already approved the PR, so I think we can close it now as it's already merged in another PR?

Thanks.

wyli:
Closing, there is no code change in this pr.

monai/transforms/utility/dictionary.py
```diff
@@ -399,15 +400,34 @@ def __init__(
                 dimension will be squeezed.
             update_meta: if `True`, copy `[key]_meta_dict` for each output and update affine to
                 reflect the cropped image
+            list_output: it `True`, the output will be a list of dictionaries with the same keys as original.
```
KumoLiu, line 403: Hi @drbeh, if `list_output` is `True`, it will return a list of dictionaries in your design. Do we need just one dictionary with orginal key and a list of data?

drbeh: Hi @KumoLiu, we want to have a list of dictionaries all with the same keys, so the transforms are being applied individually on each of them.

PR #5132, opened 2022-09-12
4456 optional import issue on top-level modules -- monai.handlers/monai.engines

### LINKED ISSUE #4456: improve error msg when optional packages not available
```py
                                                                                                                               
config = {"training": "$monai.handlers"}                                                                                       
parser.get_parsed_content("training")   
```

when ignite is not installed, the error is:
```
  File "test.py", line 5, in <module>
    parser.get_parsed_content("training")
  File "/Users/Documents/MONAI/monai/bundle/config_parser.py", line 237, in get_parsed_content
    return self.ref_resolver.get_resolved_content(id=id, **kwargs)
  File "/Users/Documents/MONAI/monai/bundle/reference_resolver.py", line 182, in get_resolved_content
  File "/Users/Documents/MONAI/monai/bundle/reference_resolver.py", line 166, in _resolve_one_item
    item.evaluate(globals={f"{self._vars}": self.resolved_content}) if run_eval else item
  File "/Users/Documents/MONAI/monai/bundle/config_item.py", line 367, in evaluate
    return eval(value[len(self.prefix) :], globals_, locals)
  File "<string>", line 1, in <module>
AttributeError: module 'monai' has no attribute 'handlers'
```

cc @dongyang0122 @Nic-Ma

wyli:
Fixes #4456

### Description
make sure that if the top-level modules miss dependencies, the optional import error raises properly.

### Types of changes

### DIFF
--- a/monai/__init__.py
+++ b/monai/__init__.py
@@ -11,6 +11,7 @@
 
 import os
 import sys
+import types
 
 from ._version import get_versions
 
@@ -45,7 +46,18 @@ excludes = "(^(monai.handlers))|(^(monai.bundle))|(^(monai.fl))|((\\.so)$)|(^(mo
 load_submodules(sys.modules[__name__], False, exclude_pattern=excludes)
 
 # load all modules, this will trigger all export decorations
-load_submodules(sys.modules[__name__], True, exclude_pattern=excludes)
+_, err_mod = load_submodules(sys.modules[__name__], True, exclude_pattern=excludes)
+
+if isinstance(sys.modules.get("monai", ""), types.ModuleType) and err_mod:  # project-monai/monai#4456
+    for x in err_mod:
+        if x.startswith("monai.") and "utils" not in x:
+            from monai.utils.module import optional_import
+
+            name = x.split(".", 1)
+            if len(name) != 2:
+                continue  # don't understand more than one dot in the erroneous module name
+            setattr(sys.modules["monai"], name[1], optional_import(x)[0])
+
 
 __all__ = [
     "apps",
--- a/monai/__init__.py
+++ b/monai/__init__.py
@@ -11,7 +11,6 @@
 
 import os
 import sys
-import types
 
 from ._version import get_versions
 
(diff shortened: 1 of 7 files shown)


wyli:
hi @ericspod could you please help review this PR? this is to correct the error message when top-level modules (`monai.engines` and `monai.handlers`) are not imported by `load_modules`, it now raises an optional import error instead of an attribute error. All integration tests work fine.

ericspod:
The way of loading modules needs to be reconsidered a bit here since the two stage process was from some very old code of mine. We aren't using the `@export` decorator which it was intended for. The immediate issue here however is that `Metric` object [here](https://github.com/Project-MONAI/MONAI/blob/41950d4a3c6498d7a4e26a8ae08b0fbb2246cad0/monai/handlers/ignite_metric.py#L22) raises the exception when it is used as a parent class, I would say it should raise the exception when one tries to instantiate it instead. Because the exception is raised the module doesn't get created, your solution here doesn't raise the error but will return a `LazyRaise` object which is a bit confusing as well. Could we change the behaviour of `LazyRaise` so that `monai.handlers` is an actual module with classes which raises exceptions on instantiation?

wyli:
thanks @ericspod, it seems it's possible to achieve what you suggested. I've added an `as_type` bool to `optional_import`, when it's true, it returns the class instead of an instance so other types can inherit. I've updated the PR, please let me know what you think...

ericspod:
The code from issue #4456 now works in that it gives me a module object back as expected, however it permits me to instantiate `monai.handlers.IgniteMetric` without Ignite being installed, it does raise an exception when I try to call any of its methods. If that's expected then it works and I'm good with the changes, I would still suggest some tweak that will raise the exception on instantiation however. This may require a different class from `LazyRaise` which has a constructor accepting any arguments and doing nothing but raising the exception.

wyli:
updated to raise an error when calling `__init__`, the only corner case is the decorator usage:

https://github.com/Project-MONAI/MONAI/blob/d64c6d2e88a658ec02d5fd23559513dec24153ff/monai/handlers/ignite_metric.py#L63

it'll also call the constructor, so I add some check for this case. @ericspod

ericspod (approved):
Works for me as expected now, thanks!

PR #5154, opened 2022-09-15
5150 fixes integration error  reading stdout visible devices

### LINKED ISSUE #5150: transformer 4.22.0 causes integration issues
```
The cache for model files in Transformers v4.22.0 has been updated. Migrating your old cache. This is a one-time only operation. You can interrupt this and resume the migration later on by calling `transformers.utils.move_cache()`.

  0%|          | 0/1 [00:00<?, ?it/s]
100%|██████████| 1/1 [00:00<00:00,  3.10it/s]
100%|██████████| 1/1 [00:00<00:00,  3.10it/s]
Moving 1 files to the new cache system 2,1
  File "<string>", line 2, in <module>
  File "/opt/conda/lib/python3.8/site-packages/torch/cuda/__init__.py", line 217, in _lazy_init
    torch._C._cuda_init()
RuntimeError: No CUDA GPUs are available
1.13.0a0+d321be6
0 of GPUs available
  File "<string>", line 1, in <module>
  File "/opt/conda/lib/python3.8/site-packages/torch/cuda/__init__.py", line 217, in _lazy_init
    torch._C._cuda_init()
RuntimeError: No CUDA GPUs are available
```

wyli:

Fixes #5150

### Description
`python -m tests.utils` may print additional contents from 3rd party packages,
only the last line for the stdout should be used

### Types of changes

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -42,7 +42,7 @@ jobs:
         echo "Sleep $LAUNCH_DELAY"
         sleep $LAUNCH_DELAY
         nvidia-smi
-        export CUDA_VISIBLE_DEVICES=$(python -m tests.utils)
+        export CUDA_VISIBLE_DEVICES=$(python -m tests.utils | tail -n 1)
         echo $CUDA_VISIBLE_DEVICES
         trap 'if pgrep python; then pkill python; fi;' ERR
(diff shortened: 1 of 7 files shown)


ericspod (approved):
I'm good with this as a quick fix if that's needed right now. I would suggest for the future having a separate script file in tests that does this for us without importing anything extra, something like this or something calling nvidia-smi:

```python
if __name__ == "__main__":
    import torch
    print(",".join(map(str,range(torch.cuda.device_count()))))
```

Nic-Ma (approved):

PR #5169, opened 2022-09-16
allow bbox in occ sens constructor

rijobro:
### Description

If bounding box is always the same, it makes sense to give the user the option to set it in the constructor.

If bbox given in `__call__` method use that. Else use bbox given in constructor. Else use `None`.

### Types of changes

### DIFF
--- a/monai/visualize/occlusion_sensitivity.py
+++ b/monai/visualize/occlusion_sensitivity.py
@@ -152,6 +152,7 @@ class OcclusionSensitivity:
         per_channel: bool = True,
         upsampler: Optional[Callable] = default_upsampler,
         verbose: bool = True,
+        bbox: Optional[Sequence] = None
     ) -> None:
         """Occlusion sensitivity constructor.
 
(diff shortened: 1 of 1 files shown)


ericspod (approved):

monai/visualize/occlusion_sensitivity.py
```diff
@@ -298,6 +306,7 @@ def __call__(self, x: torch.Tensor, b_box: Optional[Sequence] = None) -> Tuple[t
                 * The most probable class when the corresponding part of the image is occluded (``argmax(dim=-1)``).
             Both images will be cropped if a bounding box used, but voxel sizes will always match the input.
         """
+        bbox = bbox if bbox is not None else self.bbox
```
wyli, line 309:
```suggestion
        bbox = b_box if b_box is not None else self.bbox
```

garciadias:
Hi @rijobro, could you please resolve the conflicts with the dev branch?

ericspod:
Hi @garciadias I think this one is so old now that we are better closing it and revisiting this change if it's still desired.

PR #5186, opened 2022-09-21
Add Classification Result Post Process for HoVerNet

### LINKED ISSUE #5180: The whole process pipeline

KumoLiu:

Fixes #5180.

### Description
Add classification post-process script for HoVerNet.

### Types of changes


drbeh (changes requested):
Hi @KumoLiu, thanks for submitting this PR. it seems that this PR is still not ready for reviewing yet. Can you remove redundant files, review and address the checklist in the PR description, resolve the failed tests, and update docs before moving forward with review? Thanks

KumoLiu:
Hi @drbeh, I have reviewed items in this PR description, check everything that I have already done, and removed irrelevant lines. Some tests will continue to be unsuccessful since #5173 hasn't been merged. But could you please help review this PR first and let me know if you have any advice? Thanks in advance!

docs/source/apps.rst
```diff
@@ -126,6 +126,17 @@ Applications
 .. autoclass:: TileOnGridd
     :members:
 
+.. automodule:: monai.apps.pathology.transforms.post.array
+.. autofunction:: _coords_to_pixel
+.. autofunction:: _dist_from_topleft
+.. autofunction:: _sp_contours_to_cv
```
drbeh, line 132: You don't need to add the private functions to the documents. Thanks

monai/apps/pathology/transforms/post/array.py
```diff
@@ -0,0 +1,336 @@
...
+from monai.utils.enums import TransformBackends
+
+find_contours, _ = optional_import("skimage.measure", name="find_contours")
+moments, _ = optional_import("skimage.measure", name="moments")
+
+
+__all__ = ["PostProcessHoVerNetOutput"]
+
+
+def _coords_to_pixel(current, previous) -> Sequence[int]:
```
drbeh, line 33: I know that the name of these functions come from the original script but they are not very descriptive so reduce the readability of the code. Can you improve their names if possible and update the docstring?

monai/apps/pathology/transforms/post/array.py
```diff
@@ -0,0 +1,336 @@
...
+        row = int(current[0])
+        col = int(current[1] + 0.5)
+    elif p_delta == (-0.5, 0.5):
+        row = int(current[0] + 0.5)
+        col = int(current[1] + 0.5)
+
+    return row, col
+
+
+def _dist_from_topleft(sequence, h, w) -> int:
```
drbeh, line 63: same here.

monai/apps/pathology/transforms/post/array.py
```diff
@@ -0,0 +1,336 @@
...
+        distance = w + first[0]
+    elif first[0] == h - 1:
+        distance = 2 * w + h - first[1]
+    else:
+        distance = 2 * (w + h) - first[0]
+
+    return distance
+
+
+def _sp_contours_to_cv(contours, h, w) -> np.array:
```
drbeh, line 95: Doesn't make sense to create a new components for this? Can it be useful in other applications?

monai/apps/pathology/transforms/post/array.py
```diff
@@ -0,0 +1,336 @@
...
+        self.get_instance_level_seg_map = GetInstanceLevelSegMap(
+            threshold_pred=threshold_pred,
+            threshold_overall=threshold_overall,
+            min_size=min_size,
+            sigma=sigma,
+            kernel_size=kernel_size,
+            radius=radius,
+        )
+
+    def __call__(self, np_pred: NdarrayOrTensor, hv_pred: NdarrayOrTensor, nc_pred: Optional[NdarrayOrTensor] = None):
```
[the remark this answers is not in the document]
KumoLiu: Hi @drbeh, the dictionary transform already accepts input of the dictionary type, do you mean that we only need to write the dictionary transform? Or if this also becomes a dictionary as input is it confusing with `PostProcessHoVerNetOutputd`? Thanks!

docs/source/apps.rst
KumoLiu: Removed it, thanks!

KumoLiu: Removed it, thanks!

monai/apps/pathology/transforms/post/array.py
KumoLiu: Have updated it, thanks!

monai/apps/pathology/transforms/post/array.py
KumoLiu: Have updated it, thanks!

monai/apps/pathology/transforms/post/array.py
KumoLiu: I was going to add this in another pr, but now I've created it in this [new commit](https://github.com/Project-MONAI/MONAI/pull/5186/commits/b253302a7a58e1ae73826e1de68eb0153d9c56cb) in this pr. Thanks!

monai/apps/pathology/transforms/post/dictionary.py
```diff
@@ -0,0 +1,152 @@
...
+        allow_missing_keys: don't raise exception if key is missing.
+    """
+
+    backend = PostProcessHoVerNetOutput.backend
+
+    def __init__(
+        self,
+        keys: KeysCollection = HoVerNet.Branch.NP.value,
+        hv_pred_key: str = HoVerNet.Branch.HV.value,
+        nv_pred_key: str = HoVerNet.Branch.NC.value,
```
drbeh, line 96: Hi @wyli @ericspod @rijobro @Nic-Ma,

I would like to know your opinion on what is the best way to create map transforms for transforms that rely on multiple keys at the same time to do the transforms. Here is an example where the output of HoVerNet model is a dictionary with three keys (one for each of three branches). This post-processing transform need to accept three keys but our `MapTransform` only accepts one set of independent keys `keys`.

This also has been the case in deep grow transform, where the dictionary transform inherit `Transform` instead of `MapTransform` and create its own signature.
https://github.com/Project-MONAI/MONAI/blob/ba670c9ba78f9f4d76eb12b4d33f245b5774d462/monai/apps/deepgrow/transforms.py#L79-L81

It seems that it is not a one-off thing and may happen more, thus I believe it worth to discuss and come to an agreement on a "right way" to do that in MONAI. What do you thing?

rijobro: This seems fine to me. `keys` can be a sequence of keys on which the transform will be applied. Other references to keys (could be single keys as in this case, could feasibly be sequence of keys (e.g., meta_key prior to us implementing MetaTensor) will be used as supporting information. `CropForegroundd` ([link](https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/croppad/dictionary.py#L621)), for example, will crop all images given as `keys`, but will use the `source_key` as the basis for deciding what constitutes foreground and background.

This seems to be the way we do it throughout, do you have any examples of us doing it differently? Or suggestions on how to improve upon it?

drbeh: Thanks @rijobro! The situations that you mentioned make sense and it seems that `keys` argument are the primary keys and the other keys are kind of auxiliary or complementary.

Our situation is that all the keys have equal value, so even the decision of which one should be `keys` and which one `other_name_keys` is arbitrary. This might cause a confusion.

The other way that we have done in MONAI (and it might be better for these situations) is in deepgrow transforms, where we have inherited `Transform` instead of `MapTransform` and made equally important key arguments with different names, e.g., [AddInitialSeedPointd](https://github.com/Project-MONAI/MONAI/blob/ba670c9ba78f9f4d76eb12b4d33f245b5774d462/monai/apps/deepgrow/transforms.py#L68).

rijobro: Ah I see. In the [proposed](https://github.com/Project-MONAI/MONAI/pull/5186/files/48f796f8e7bd49b3ddbb606c24332b1298e42071#diff-980b604c7a16c6f010fea135e006ad1f7b65a9744a311df9421e9feb74e78b10R145) implementation of `PostProcessHoVerNetOutputd`, even though different keys might be equally important in the calculation, the value in `[key]` is the one that is changed, so I think the way it's currently suggested is correct. You also update a dictionary info key, but I would argue that it's still secondary information compared to the image that is being replaced with `[key]`.

Nic-Ma: Hi @wyli @ericspod ,

What do you guys think about this question?

Thanks in advance.

drbeh: @rijobro you are right, in the current implementation the key in `keys` is the one that is changed but it can be either of branches (NP or HV), or better not to be any of them and create a new key.

```python
        keys: KeysCollection = HoVerNet.Branch.NP.value,
        hv_pred_key: str = HoVerNet.Branch.HV.value,
```

rijobro: Ah, I see...

KumoLiu: Hi @drbeh, in my mind, in our current case, the NP branch should be the one that is changed, as HV branch is actually used to assist in generating the instance-level segmentation. NP branch's output is indeed the segmentation result. Thanks!

drbeh: @KumoLiu you are right but actually none of them is instance segmentation. Even if it was the case, the approach that we took in deepgrow is more readable and less confusing for the user. You just need to inherit from Transform and give the argument for each key the appropriate name (e.g., `np_pred_key` and `hv_pred_key` as you did). What do you think?
https://github.com/Project-MONAI/MONAI/blob/ba670c9ba78f9f4d76eb12b4d33f245b5774d462/monai/apps/deepgrow/transforms.py#L87-L90

KumoLiu:
Hi @drbeh @Nic-Ma  @wyli, I refactored the whole pipeline to several small components without the whole transformation. The whole transformation will be added to the tutorials. I want to know your suggestions if this refactor is more reasonable. If it is ok, I will add unit tests then. Thanks in advance!

monai/apps/pathology/transforms/post/array.py
```diff
@@ -320,3 +327,300 @@ def __call__(self, mask: NdarrayOrTensor, instance_border: NdarrayOrTensor) -> N
...
+    Converts Scipy-style contours(generated by skimage.measure.find_contours) to a more succinct version which only includes
+    the pixels to which lines need to be drawn (i.e. not the intervening pixels along each line).
+
+    Args:
+        height: height of bounding box, used to detect direction of line segment.
+        width: width of bounding box, used to detect direction of line segment.
+
+    Returns:
+        the pixels that need to be joined by straight lines to describe the outmost pixels of the foreground similar to
+            OpenCV's cv.CHAIN_APPROX_SIMPLE (anti-clockwise)
```
drbeh, line 343:
```suggestion
            OpenCV's cv.CHAIN_APPROX_SIMPLE (counterclockwise)
```

drbeh (commented):
Are you using the testing data? `tests/testing_data/hovernet_test_data_raw.npz`
If not, please remove it. Thanks

KumoLiu:
> Are you using the testing data? `tests/testing_data/hovernet_test_data_raw.npz` If not, please remove it. Thanks

Yes, thanks for pointing it out, I removed it in the latest commit!

drbeh:
@JHancox do you have any comments on this PR?

Nic-Ma:
Hi @drbeh ,

I see @JHancox approved the PR, and @KumoLiu is developing the whole wrapper transform in #5462 .
Do you still have any other concerns? If not, I will try to merge this PR ASAP.

Thanks in advance.

drbeh (approved):

PR #5197, opened 2022-09-22
Add `GetForegroundPatches` and `GetForegroundPatchesd`

### LINKED ISSUE #4181: Transform to Threshold Image for Tissue
A transform that given a RGB image (a tile from histopathology image or a WSI at low resolution), apply thresholding algorithm to distinguish the tissue from the background, and outputs a binary mask and some statistics.

KumoLiu:

Fixes #4181 .

### Description
Add `GetForegroundPatches` and `GetForegroundPatchesd` to meet the requirements in #4181. The shape of input image patches is [N, C, H, W, [D]] where N denotes the patch/tile count, and returns foreground patches/tiles and masks.

### Types of changes


drbeh:
Hi @KumoLiu, thank you very much for creating this PR.

We can extract masked patches in the dataset like this:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/wsi_datasets.py#L316
But I am not sure how this components will be used in a pipeline. Is it based on PR#4181? 
@JHancox can you take a quick look and see is it what you have requested for? 

I know it takes a lot of time and effort to write the code and unit tests, so let's make sure first that the requirements and feature requests are clear.

KumoLiu:
Hi @drbeh @JHancox,
I have just checked [`MaskedPatchWSIDataset`](https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/wsi_datasets.py#L316). I'll briefly summarise how it takes all the foreground patches. 
- First, it ingests the wsi in the `mask_level`(which may be a lower resolution)
- Then at this level, it uses `ForegroundMask` to get all foreground pixel locations
- Then it will take all the foreground pixel locations of this `mask_level` back to level 0 and use these coordinates as the initial or central coordinates for extracting the foreground patches later

In this way, users may need to notice that  `patch_size` >= “magnification between final used level and `mask_level`” which ensures all the patches cover the foreground. If I understand correctly, `MaskedPatchWSIDataset` can indeed meet the requirements in #4181. @JHancox, could you please see the steps I wrote above to double-check whether it can meet your use case? Thanks!

JHancox:
It sounds fine. The only other feature, which I mentioned recently, was the ability to clean up the patch selection, based on whether is has any neighbours, but this can also be added later.

KumoLiu:
Hi @JHancox, thanks for your reply. I will close this PR, and if we have any further requirements, we can discuss them later. Thank!

PR #5220, opened 2022-09-29
get_norm_layer default to affine=True

myron:
https://github.com/Project-MONAI/MONAI/issues/5222

Sets the default normalization parameter (affine) to True. 

Most normalizations in pytorch by default estimates the affine parameters (e.g. Batchnorm, Groupnorm). 
One notable exception is Instancenorm (for which default is affine=False), and if we don't specify it manually, it won't estimate these parameters, e.g. this won't have any learnable parameters
``
get_norm_layer(name="instance", spatial_dims=2)
``
and one needs to use a verbose input
``
get_norm_layer(name=("instance", {"affine":True}, spatial_dims=2)
``

this change makes affine (parameters) default to True for all normalizations, unless it's explicitly set to False. 
This is the preferred way and  should be the default for all segmentation/classification tasks.

### Description

### Types of changes

### DIFF
--- a/monai/networks/layers/utils.py
+++ b/monai/networks/layers/utils.py
@@ -47,6 +47,8 @@ def get_norm_layer(name: Union[Tuple, str], spatial_dims: Optional[int] = 1, cha
         kw_args["num_features"] = channels
     if has_option(norm_type, "num_channels") and "num_channels" not in kw_args:
         kw_args["num_channels"] = channels
+    if has_option(norm_type, "affine") and "affine" not in kw_args:
+        kw_args["affine"] = True
     return norm_type(**kw_args)
 
 


myron:
Looks like it failed a few tests. One of them is the number of parameters for Unet

Looks like our Unet implementation has Instancenorm as the default normalization
https://github.com/Project-MONAI/MONAI/blob/dev/monai/networks/nets/unet.py#L124

as well as many other places (e.g Convolution helper func)
https://github.com/Project-MONAI/MONAI/blob/ba670c9ba78f9f4d76eb12b4d33f245b5774d462/monai/networks/blocks/acti_norm.py#L105

but this seems wrong , Instancenorm in Unet must have learnable (affine) parameters, and current implementation does not.

monai/networks/layers/utils.py
```diff
@@ -47,6 +47,8 @@ def get_norm_layer(name: Union[Tuple, str], spatial_dims: Optional[int] = 1, cha
         kw_args["num_features"] = channels
     if has_option(norm_type, "num_channels") and "num_channels" not in kw_args:
         kw_args["num_channels"] = channels
+    if has_option(norm_type, "affine") and "affine" not in kw_args:
+        kw_args["affine"] = True
```
wyli, line 51: this change will break some existing models... I think we should warn in this version with something like:

```suggestion
        warnings.warn(f"`affine` is not set in {norm_type}, in future releases the default will be changed to `affine=True`.")
```

and change it in v1.2

wyli (commented):
(this would need more discussions @ericspod @rijobro @Nic-Ma @mingxin-zheng about compatibility)

monai/networks/layers/utils.py
ericspod: This will break models. I had used instancenorm as the default because it gave better results and explicitly because it didn't have learnable parameters, it was a simple transform which reduced the number of parameters a network would need. We have an issue that, yes, a number of networks used instancenorm and its default of not having parameters so this will be a breaking change for tested models and those in the model zoo. I feel we shouldn't make this change now but we could introduce a "instancenorm_affine" item into the `Norm` factory to simplify doing this for future network definitions.

myron:
Alright, looks like this doesn't have support for legacy reasons, I'm going to close this

PR #5221, opened 2022-09-29
UpSample optional kernel_size for deconv mode

myron:
Adds (optional) kernel_size parameter to UpSample, used for deconv (convolution transpose up-sampling). 

This allows to upsample, e.g to upscale to 2x with a kernel_size 3. (currently the default is to upscale to 2x with a kernel size 2)

if this parameter is not set, the behavior is the same as before

### Description

### Types of changes

### DIFF
--- a/monai/networks/blocks/upsample.py
+++ b/monai/networks/blocks/upsample.py
@@ -40,6 +40,7 @@ class UpSample(nn.Sequential):
         in_channels: Optional[int] = None,
         out_channels: Optional[int] = None,
         scale_factor: Union[Sequence[float], float] = 2,
+        kernel_size: Optional[Union[Sequence[float], float]] = None,
         size: Optional[Union[Tuple[int], int]] = None,
         mode: Union[UpsampleMode, str] = UpsampleMode.DECONV,
         pre_conv: Optional[Union[nn.Module, str]] = "default",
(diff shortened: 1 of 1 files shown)


monai/networks/blocks/upsample.py
```diff
@@ -83,13 +85,21 @@ def __init__(
...
             self.add_module(
                 "deconv",
                 Conv[Conv.CONVTRANS, spatial_dims](
                     in_channels=in_channels,
                     out_channels=out_channels or in_channels,
-                    kernel_size=scale_factor_,
+                    kernel_size=kernel_size_,
                     stride=scale_factor_,
+                    padding=padding,
+                    output_padding=padding,
```
wyli, line 102: these paddings couldn't pass the test case:
https://github.com/Project-MONAI/MONAI/blob/ba670c9ba78f9f4d76eb12b4d33f245b5774d462/tests/test_upsample_block.py#L45-L47 perhaps add a `'same'` padding mode?

KumoLiu:
Hi @myron, you can set up padding and output_padding separately like this, which might solve the test error of mismatched test output sizes.
```
output_padding = [0, ] * spatial_dims
for i, _scale_factor_ in enumerate(scale_factor_):
    if kernel_size_[i] % 2 == 0:
        output_padding[i] = _scale_factor_ - 2
    else:
        output_padding[i] = _scale_factor_ - 1

padding = tuple((i-1)//2 for i in kernel_size_)

self.add_module(
    "deconv",
    Conv[Conv.CONVTRANS, spatial_dims](
        in_channels=in_channels,
        out_channels=out_channels or in_channels,
        kernel_size=kernel_size_,
        stride=scale_factor_,
        padding=padding,
        output_padding=tuple(output_padding),
        bias=bias,
    ),
)
```
Thanks!

myron:
ok, I believe I fixed that, and added another unit test. thanks you

wyli (approved):
looks good to me, just to note that when the kernel size is not divisible by the scale_factor, this will introduce some checkerboard artefacts.

myron:
> looks good to me, just to note that when the kernel size is not divisible by the scale_factor, this will introduce some checkerboard artefacts.

Thanks. is it a known fact? I use it with scale 2, kernel 3, I don't observe any artifacts.

wyli:
the idea is presented here https://distill.pub/2016/deconv-checkerboard/, I guess it'll be less noticeable for segmentation tasks..

Issue #3032, opened 2021-09-27
Interpretability of Segmentation UNet


wyli:
### Discussed in https://github.com/Project-MONAI/MONAI/discussions/3005

<div type='discussions-op-text'>

<sup>Originally posted by **joho84** September 22, 2021</sup>
Hi!
First of all, thank you very much for your work and the great support you offer!
I have a question concerning the interpretability of a segmentation Unet. I am trying to get some more insight into the functioning of my Unet trained on multi-channel data. In especially I would be interested in which channels and which voxels contribute the most to the decision of the trained Unet for the segmentation of a specific class. Something similar to this: https://arxiv.org/abs/2002.11434
Is there already a way to achieve this in monai? Or could you point me in the right direction to adapt the existing interpretability code to a segmentation task?
Thank you in advance!</div>

(converting the discussion into a feature request)
see also https://github.com/kiraving/SegGradCAM

Nic-Ma:
Hi @ericspod ,

I think you are expert on the UNet, could you please help share some comments?

Thanks in advance.

ericspod:
I'm afraid I don't have much experience with this specific sort of interpretability. @rijobro already weighed in about gradcam and what we've done with it, but for this problem you're looking at discerning the contribution of specific parts of the input not just regions. I could guess at a few strategies to try which aren't specific in any way to what network you have. Adding noise to one channel at a time to see what causes performance to decrease the most during inference, or even blacking out a channel, applying occlusion mapping to one channel at a time (occlusion is like noise), etc. Essentially making one channel worse so that it's contribution is decreased to the final result. If the channels are co-dependent this might be hard to see in which case you'd have decreasing performance no matter which channel you attack, then it could be argued all channels are equally important.

PR #5230, opened 2022-09-29
Occlusion sensitivity to use slidiing_window_inference

rijobro:
### Description

Occlusion sensitivity currently works by setting the logits at the center of the occluded region. Even though neighbouring voxels are occluded at the same time, this is not recorded. A better way to do it would be to do an average each time a voxel is occluded. This is taken care of by switching to using `sliding_window_inference`. 

We had to do a bit of hacking to get this to work. `sliding_window_inference` normally takes a subset of the whole image and infers that. We want `sliding_window_inference` to tell us which part of the image we should occlude, then we occlude it and infer the **whole** image. To that end, we use a meshgrid to tell us the coordinates of the region `sliding_window_inference` wants us to occlude. We occlude, infer and then crop back to the size of the region that was requested by `sliding_window_inference`. 

This PR also allows for different occlusion kernels. We currently have:
    - gaussian (actually inverse gaussian): the center of the occluded region is zero, and towards the edge of the image is unchanged. This doesn't introduce hard edges into the image, which might undermine the visualisation process.
    - mean_patch: the occluded region is replaced with the mean of the patch it is occluding.
    - mean_img: the occluded region is replaced with the mean of the whole image (current implementation).

## Changes to input arguments
This PR is backwards incompatible, as using `sliding_window_inference` means changing the API significantly.
    - pad_val: now determined by `mode`
    - stride: `overlap` used instead
    - per_channel: all channels are done simultaneously
    - upsampler: image is no longer downsampled

## Changes to output
Output previously had the shape `B,C,H,W,[D],N` where `C` and `N` were the number of input and output channels of the network, respectively. Now, we output the shape `B,N,H,W,[D]` as the `per_channel` feature is no longer present.

Columns 2-4 are occlusion sensitivity done with Gaussian, mean of patch and mean of image:

vis

### Types of changes

### DIFF
--- a/monai/visualize/occlusion_sensitivity.py
+++ b/monai/visualize/occlusion_sensitivity.py
@@ -291,7 +291,7 @@ class OcclusionSensitivity:
 
         # check that the grid is bigger than the mask size
         if any(m > g for g, m in zip(grid.shape[2:], mask_size)):
-            raise ValueError(f"Image (after cropping with bounding box) should be bigger than mask.")
+            raise ValueError("Image (after cropping with bounding box) should be bigger than mask.")
 
         # get additive and multiplicative factors if they are unchanged for all patches (i.e., not mean_patch)
         add: Optional[Union[float, torch.Tensor]]
--- a/monai/visualize/occlusion_sensitivity.py
+++ b/monai/visualize/occlusion_sensitivity.py
@@ -10,88 +10,17 @@
 # limitations under the License.
 
 from collections.abc import Sequence
-from functools import partial
 from typing import Callable, Optional, Tuple, Union
 
 import numpy as np
 import torch
 import torch.nn as nn
 
+from monai.data.meta_tensor import MetaTensor
 from monai.networks.utils import eval_mode
+from monai.transforms import Compose, GaussianSmooth, Lambda, ScaleIntensity, SpatialCrop
+from monai.utils import deprecated_arg, ensure_tuple_rep
 from monai.visualize.visualizer import default_upsampler
-
-try:
-    from tqdm import trange
-
-    trange = partial(trange, desc="Computing occlusion sensitivity")
-except (ImportError, AttributeError):
-    trange = range
-
-# For stride two (for example),
-# if input array is:     |0|1|2|3|4|5|6|7|
-# downsampled output is: | 0 | 1 | 2 | 3 |
-# So the upsampling should do it by the corners of the image, not their centres
-default_upsampler = partial(default_upsampler, align_corners=True)
-
-
-def _check_input_image(image):
-    """Check that the input image is as expected."""
-    # Only accept batch size of 1
-    if image.shape[0] > 1:
-        raise RuntimeError("Expected batch size of 1.")
-
-
-def _check_input_bounding_box(b_box, im_shape):
-    """Check that the bounding box (if supplied) is as expected."""
-    # If no bounding box has been supplied, set min and max to None
-    if b_box is None:
-        b_box_min = b_box_max = None
-
-    # Bounding box has been supplied
-    else:
-        # Should be twice as many elements in `b_box` as `im_shape`
-        if len(b_box) != 2 * len(im_shape):
-            raise ValueError("Bounding box should contain upper and lower for all dimensions (except batch number)")
-
-        # If any min's or max's are -ve, set them to 0 and im_shape-1, respectively.
-        b_box_min = np.array(b_box[::2])
-        b_box_max = np.array(b_box[1::2])
-        b_box_min[b_box_min < 0] = 0
-        b_box_max[b_box_max < 0] = im_shape[b_box_max < 0] - 1
-        # Check all max's are < im_shape
-        if np.any(b_box_max >= im_shape):
-            raise ValueError("Max bounding box should be < image size for all values")
-        # Check all min's are <= max's
-        if np.any(b_box_min > b_box_max):
-            raise ValueError("Min bounding box should be <= max for all values")
-
-    return b_box_min, b_box_max
-
-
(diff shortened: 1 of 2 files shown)


monai/visualize/occlusion_sensitivity.py
```diff
@@ -142,139 +71,184 @@ class OcclusionSensitivity:
         - :py:class:`monai.visualize.occlusion_sensitivity.OcclusionSensitivity.`
     """
 
+    @deprecated_arg(
+        name="pad_val",
+        since="1.0",
+        removed="1.1",
```
wyli, line 77: we agreed to remove in the x+2 version, so I think this should be `1.2`: https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#backwards-compatibility
```suggestion
        removed="1.2",
```

monai/visualize/occlusion_sensitivity.py
```diff
@@ -142,139 +71,184 @@ class OcclusionSensitivity:
...
+        kernel = torch.zeros((x.shape[1], *mask_size), device=x.device, dtype=x.dtype)
+        spatial_shape = kernel.shape[1:]
+        # all channels (as occluded shape already takes into account per_channel), center in spatial dimensions
+        center = [slice(None)] + [slice(s // 2, s // 2 + 1) for s in spatial_shape]
+        # place value of 1 at center
+        kernel[center] = 1.0
+        # Smooth with sigma equal to quarter of image, flip +ve/-ve so largest values are at edge
+        # and smallest at center. Scale to [0, 1].
+        gaussian = Compose(
+            [GaussianSmooth(sigma=[b // 4 for b in spatial_shape]), Lambda(lambda x: -x), ScaleIntensity()]
```
wyli, line 151: why choose this sigma as `b//4`? this might need some comments

rijobro: Empirically I found this to give good results. Below I've attached imshows of the kernels for mask sizes of 15 and 31. I don't think this parameter should be offered to the user to change, the goal is simply to give a smoothly varying kernel. What do you think? I can add to the documentation that it is empirically determined and give the code to do the imshow?

image

wyli (approved):
cool, please help address the minor comments inline...

looks like this sliding window implementation can close https://github.com/Project-MONAI/MONAI/issues/3032?

monai/visualize/occlusion_sensitivity.py
wyli: thanks, I think your comment here is enough as a kind of documentation... so marking this as resolved.

rijobro: Few more plots, sigmas of 1, 0.25, 0.1 and mask sizes of 15 and 31. Think you'll agree that 0.25 looks best.
image

monai/visualize/occlusion_sensitivity.py
```diff
@@ -301,29 +277,73 @@ def __call__(
...
             Both images will be cropped if a bounding box used, but voxel sizes will always match the input.
         """
+        if x.shape[0] > 1:
+            raise ValueError("Expected batch size of 1.")
+
+        sd = x.ndim - 2
+        mask_size = ensure_tuple_rep(self.mask_size, sd)
+
+        # get the meshgrid (so that sliding_window_inference can tell us which bit to occlude)
+        grid: MetaTensor = MetaTensor(
```
wyli, line 287: just another thought that by defining a binary mask to indicate the area to occlude, along with this `grid` variable, it seems it'll be possible to make occlusion of arbitrary shapes (not necessarily a set of rectangular bounding boxes).

ericspod (approved):

ericspod:
This can close #3032 though one thing mentioned there is determining the importance of individual channels, this may be important to support with an example of how it can be shown to work. I'm not sure if cats and dogs would be suitable unless the network fixates on specific coat colours or something like that.

monai/visualize/occlusion_sensitivity.py
```diff
@@ -142,139 +71,184 @@ class OcclusionSensitivity:
...
+        if callable(activate):
+            out = activate(out)
+        # else if True (should be boolean), sigmoid if n_chan == 1 else softmax
+        elif activate:
+            out = out.sigmoid() if x.shape[1] == 1 else out.softmax(1)
+
+        # the output will have shape [B,C] where C is number of channels output by model (inference classes)
+        # we need to return it to sliding window inference with shape [B,C,H,W,[D]], so add dims and repeat values
+        for m in mask_size:
+            out = torch.repeat_interleave(out.unsqueeze(-1), m, dim=-1)
```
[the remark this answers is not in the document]
rijobro: This will return the image as `BCH'W'D'`, where `H'W'D'` are the size of the mask, which is smaller than the image. If the device is big enough to handle `BCHWD`, then `BCH'W'D'` should be fine. The for loop is unfortunate but will only be run two or three times depending on the number of spatial dimensions in the image.

monai/visualize/occlusion_sensitivity.py
rijobro: Yes! The Gaussian is already arguably circular/spherical as the voxels are affected radially from the centre of the occluded region.

rijobro:
I don't think this solves #3032 – it still remains in the domain of interpretability of image classification, not segmentation. The output of `predictor` assumes the network gives the shape `BC`, not `BCHWD`.

wyli:
> 

I see, for unet, it'll require some spatial pooling after this line to make it from BCHWD -> BC... 
https://github.com/Project-MONAI/MONAI/blob/e427a00babaa9c74cda7080b04a83e3a330799cd/monai/visualize/occlusion_sensitivity.py#L215
could be a very interesting future PR :)

PR #5278, opened 2022-10-06
[ReadyForReview] Auto3DSeg DataAnalyzer OOM and other minor issue

### LINKED ISSUE #5277: Auto3DSeg DataAnalyzer Out-of-memory (OOM)
**Describe the bug**
Auto3DSeg would go out-of-memory during data analysis

**Expected behavior**
- Auto3DSeg uses CPU to analyze datasets by default. 
- When it is using GPU but unable to allocate memory, it can provide a warning to user to change the analysis to CPU-mode.

mingxin-zheng:
Fixes #5277 .

### Updated results
In my local test env, I have the following results:
- The change of GPU memory before/after DataAnalyzer is less than 5MB after the fix. Previously, there are lots of cached PyTorch tensors and CuPy variables that are not released for trainings that takes up to several GBs of GPU mem.
- DataAnalyzer can also process larger images now because leaks are fix (3D image with a size 512x512x512 passed for 12GB RTX 3080Ti) 

### Description

Auto3DSeg DataAnalyzer occupied a large trunk of memory and was unable to release them during the training. The reasons behind are possibly due to:
- Training are done by subprocess call, and PyTorch in the subprocess is unable to find the memory pool allocated by the main process
- GPU memory leakage ( DataAnalyzer math operations uses torch functions and CuPy)
plus test functions need improvements and AutoRunner needs to expose the API call to change device of DataAnalyzer

### Types of changes


monai/auto3dseg/analyzer.py
```diff
@@ -229,29 +229,31 @@ def __call__(self, data):
...
         d = dict(data)
         start = time.time()
-        ndas = data[self.image_key]
-        ndas = [ndas[i] for i in range(ndas.shape[0])]
-        if "nda_croppeds" not in d:
-            nda_croppeds = [get_foreground_image(nda) for nda in ndas]
 
-        # perform calculation
-        report = deepcopy(self.get_report_format())
+        with torch.no_grad():
```
wyli, line 233: I'm not sure if it's needed to add `no_grad()`  for each analyzer, it looks like lots of lines changed. how about we add it here in the summarizer? if it doesn't work, we make further changes?
https://github.com/Project-MONAI/MONAI/blob/1f9e5ce34691af8be20a84a673ecf5c7535e6b66/monai/auto3dseg/seg_summarizer.py#L181

mingxin-zheng: Thanks @wyli for the suggestion. Lots of line changes are due to the indentation. To reduce the number of line changes, I would replace `with torch.no_grad()` with

```
torch.set_grad_enabled(mode)
```

Adding it in the summarizer also looked like a good idea to me first, but later I found a problem: the summarizer inherits `Compose`. The computation happens in the `__call__` functions of `Compose` and adding `torch.no_grad()` there seemed a bad idea.

wyli (approved):
Thanks, it looks good to me.

PR #5292, opened 2022-10-08
fixes DiceCELoss for multichannel targets

myron:
Fixes DiceCELoss for multichannel targets.
Currently if "target" (ground truth label) is provided as a multichannel data (each channel is binary or float), then current DiceCELoss attempts to convert it to 1-channel using argmax (which could be impossible with overlapping labels).  There is no need for argmax, since pytorch's cross entropy can handle multi-channel targets already. 

### Types of changes


monai/losses/dice.py
```diff
@@ -701,12 +701,10 @@ def ce(self, input: torch.Tensor, target: torch.Tensor):
 
         """
         n_pred_ch, n_target_ch = input.shape[1], target.shape[1]
-        if n_pred_ch == n_target_ch:
```
wyli, line 704: The changes break unit tests https://github.com/Project-MONAI/MONAI/actions/runs/3208441811/jobs/5244297528

myron: it fails on old pytorch versions, so it seems on Old pytorch version, CE didn't support multichannel input. 
do we need to support those old pytorch versions?

myron:
added check for pytorch version

monai/losses/dice.py
```diff
@@ -701,12 +701,16 @@ def ce(self, input: torch.Tensor, target: torch.Tensor):
...
-            # target is in the one-hot format, convert to BH[WD] format to calculate ce loss
-            target = torch.argmax(target, dim=1)
-        else:
+        if n_pred_ch != n_target_ch and n_target_ch == 1:
             target = torch.squeeze(target, dim=1)
-        target = target.long()
+            target = target.long()
+        else:
+            v = torch.__version__.split(".")[:2]
+            if int(v[0]) < 1 or (int(v[0]) == 1 and int(v[1]) < 10):
```
wyli, line 709: Looks good to me, please update to use this version comparison https://github.com/Project-MONAI/MONAI/blob/4ad790cc5f6022a92e0585c34f2f50fb6fa7fd62/monai/utils/module.py#L538

myron: pytorch_after() has unnecessary overhead, it imports "pkg_resources" package, which will do parsing of version string, creating a Version() class, just to compare the versions.  We call the loss (DiceCELoss) very frequently, often sub-second calls. I'm worried that calling this pytorch version parsing with pytorch_after() creates unnecessary overhead (on Cpu) at each backpropagration step, which will delay gpu execution.

wyli: if indeed it's called frequently and has an impact on the efficiency, shouldn't we define these in the constructor? for example:
```py
class DiceCELoss:
  def __init(...):
     self.loss_ce = loss_ce_v2 if pytorch_after(1,10) else loss_ce_v1
  def forward(input, ...):
     self.loss_ce(input, ...)
```

myron: sounds good. Thank you. I've added the version check in the __init__.  please check

wyli (approved):

myron:
some checks are failing, but it's something unrelated

PR #5297, opened 2022-10-08
SlidingWindowInferer: option to adaptively stitch in cpu memory for large images

myron:
SlidingWindowInferer: option to adaptively stitch in cpu memory for large images.

This adds an option to provide maximum input image volume (number of elements) to dynamically change stitching to cpu memory (to avoid gpu memory crashes).  For example with `cpu_thresh=400*400*400`, all input images with large volume will be stitched on cpu.  

At the moment, a user must decide beforehand, to stitch ALL images on cpu or gpu (by specifying the 'device' parameter). But in many datasets,  only a few large images  require device==cpu, and running inference on cpu for ALL will be unnecessary slow. 

It's related to 
https://github.com/Project-MONAI/MONAI/discussions/4625
https://github.com/Project-MONAI/MONAI/discussions/4495
https://github.com/Project-MONAI/MONAI/discussions/3497
https://github.com/Project-MONAI/MONAI/discussions/4726
https://github.com/Project-MONAI/MONAI/discussions/4588 

### Types of changes


monai/inferers/inferer.py
```diff
@@ -122,6 +122,8 @@ class SlidingWindowInferer(Inferer):
             `inputs` and `roi_size`. Output is on the `device`.
         progress: whether to print a tqdm progress bar.
         cache_roi_weight_map: whether to precompute the ROI weight map.
+        cpu_thresh: when provided, adaptively switch to stitching on cpu (to save gpu memory)
```
Nic-Ma, line 125: Thanks for the feature enhancement, I would suggest to make this `cpu_thresh` to work with `device=None` only to avoid unexpected behavior.
What do you think?
CC @wyli @rijobro @ericspod 

Thanks.

ericspod: I'm not sure what the concern is? If `device` is any GPU moving the computation to CPU and the results back to it should be fine, so I think the behaviour of the class is that it returns results on the original device regardless. Also it should be state that the `cpu_thresh` value is is number of pixels/voxels, not bytes (which might be a better idea).

myron: thanks. I've updated the doc to indicate pixels/voxels. I also think that it should just default to "device" when image is smaller then threshold, we don't need to check for device=None condition

wyli: agree with Eric that we should use bytes to make it invariant to different dtypes.

wyli: I think Nic's idea is that if `device="cuda"` the user would expect the output is always on the cuda device. But with `cpu_thresh` the output may quietly go to cpu, slow down the inference/postprocessing.

myron: 1) I think providing the volume size in bytes will confuse  users. Instead of providing the volume size, they will need to remember to multiply it by 4 (for floats). It's uncommon that the input option for sizes/volumes is in bytes

2) It's true that the output may quietly go to Cpu (but this is intentional to avoid crashes due to GPU memory limit). Currently our inference will fail if at least 1 image in the dataset is abnormally large.  

so what do you guys propose?  to only have this options when device==None (default)?  but this still produces outputs on  both gpu / cpu depending on input image size.  
At the moment it switches to CPU (when image is large), but otherwise respects the user input for "device".

myron: so what is the consensus here, is anyone against  specifying the volume in voxels (not in bytes)? if not let's merge it for now

Nic-Ma: Hi @myron ,

I would suggest to make it only work with `device=None` to avoid unexpected or unknown behavior.
If we get more positive user feedback later, we can try to enable the feature for all `device` values.
What do you think?
CC @ericspod @rijobro 

Thanks.

monai/inferers/inferer.py
```diff
@@ -122,6 +122,9 @@ class SlidingWindowInferer(Inferer):
             `inputs` and `roi_size`. Output is on the `device`.
         progress: whether to print a tqdm progress bar.
         cache_roi_weight_map: whether to precompute the ROI weight map.
+        cpu_thresh: when provided, dynamically switch to stitching on cpu (to save gpu memory)
+                    when input image volume is larger than this threshold (in pixels/volxels).
```
Nic-Ma, line 126: The indent of doc-string is not correct here.

Thanks.

monai/inferers/inferer.py
myron: Alright, I've changed it, so that the new logics works only if device==None

monai/inferers/inferer.py
myron: I've updated this doc-string

Nic-Ma (approved):

monai/inferers/inferer.py
Nic-Ma: Thanks for your quick update, looks good to me now.

PR #5307, opened 2022-10-10
Adding a median filter for 2/3D images

### LINKED ISSUE #5264: Add median image filter as a transform
One problem is explained in #3178.

Add a median filter based on @ebrahimebrahim's [medianblur_with_gpu](https://github.com/ebrahimebrahim/abcd-registration-experiments/blob/main/medianblur_with_gpu.py?rgh-link-date=2022-10-03T16%3A21%3A07Z)

#3178 mentions some alternatives, such as mean filtering. Those are suboptimal for many use cases.

**Additional context**
Model it after `GaussianSmooth`: https://github.com/Project-MONAI/MONAI/blob/ef68f16dde33172b2eeae5397ac75aa717808875/monai/transforms/intensity/array.py#L1139-L1171

dzenanz:
Co-authored-by: @ebrahimebrahim. Closes #5264.

### Types of changes


dzenanz:
@wyli and @ebrahimebrahim can you provide some feedback? Or take over the PR?

monai/networks/layers/simplelayers.py
```diff
@@ -434,6 +435,109 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    kernel: torch.Tensor = torch.zeros(window_range, 1, window_size[0], window_size[1], window_size[2])
+    for i in range(window_size[0]):
+        for j in range(window_size[1]):
+            for k in range(window_size[2]):
+                l: int = i * window_size[2] * window_size[1] + j * window_size[2] + k
+                kernel[l, 0, i, j, k] += 1.0
+    return kernel
+
+
+def _compute_zero_padding(kernel_size: Sequence[int]) -> List[int]:
```
ericspod, line 452: Is this doing the same as `same_padding` ?

ebrahimebrahim:
I think it looks good; thanks for working on this!

Some useful features, which could potentially be put off to future work: 
- be generic to spatial dimensionality. there is no reason it has to focus on 3D images only
- add `padding_mode` options similar to those supported by other monai filters. `torch.nn.functional.conv3d` only does zero or "same" padding I think, so `torch.nn.functional.pad` may need to get involved to provide other options such as padding by a nonzero constant

tests/test_median_filter.py
```diff
@@ -0,0 +1,69 @@
...
+
+from monai.networks.layers import MedianFilter
+
+
+class MedianFilterTestCase(unittest.TestCase):
+    def test_3d(self):
+        a = torch.ones(1, 1, 4, 3, 4)
+        g = MedianFilter(1).to(torch.device("cpu:0"))
+
+        expected = np.array(
```
ebrahimebrahim, line 25: this is the failing test, right?

I think the expected outcome is different.

thinking about a 3x3x3 kernel, it has 27 positions, and
- on a corner of the image 19 of them (majority) are 0 due to padding
- on an edge of the image 15 of them (still majority) are 0 due to padding
- in the middle of a face 9 of them (not majority) are 0 due to padding

when the unpadded image consists of 1s, this means that when we apply the 3x3x3 median filter we expect:
- 0 at image corners
- 0 at image edges
- 1 in the middle of image faces
- 1 in the interior of the image volume

monai/networks/layers/simplelayers.py
```diff
@@ -434,6 +435,109 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    If the window size is HxWxD will create a (H*W*D)xHxWxD kernel.
+    """
+    window_range: int = window_size[0] * window_size[1] * window_size[2]
+    kernel: torch.Tensor = torch.zeros(window_range, 1, window_size[0], window_size[1], window_size[2])
+    for i in range(window_size[0]):
+        for j in range(window_size[1]):
+            for k in range(window_size[2]):
+                l: int = i * window_size[2] * window_size[1] + j * window_size[2] + k
+                kernel[l, 0, i, j, k] += 1.0
+    return kernel
```
ericspod, line 449:
```suggestion
def get_binary_kernel3d(window_size: Sequence[int]) -> torch.Tensor:
    r"""Create a binary kernel to extract the patches.
    If the window size is HxWxD will create a (H*W*D)xHxWxD kernel.
    """
    prod=torch.prod(torch.tensor(window_size))
    return torch.diag(torch.ones(prod)).reshape(prod,1,*window_size))
```

Is this equivalent?

monai/networks/layers/simplelayers.py
```diff
@@ -434,6 +435,109 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    # prepare kernel
+    if kernel is None:
+        kernel: torch.Tensor = get_binary_kernel3d(kernel_size).to(in_tensor)
+    else:
+        kernel: torch.Tensor = kernel.to(in_tensor)
+    c, h, w, d = in_tensor.shape
+
+    # map the local window to single vector
+    features: torch.Tensor = F.conv3d(in_tensor.reshape(c, 1, h, w, d), kernel, padding=padding, stride=1)
+    features = features.view(c, -1, h, w, d)  # Cx(K_h * K_w * K_d)xHxWxD
```
ericspod, line 491:
```suggestion
        kernel: torch.Tensor = get_binary_kernel(kernel_size).to(in_tensor)
    else:
        kernel: torch.Tensor = kernel.to(in_tensor)
    c, *sshape = in_tensor.shape

    # map the local window to single vector
    features: torch.Tensor = F.conv3d(in_tensor.reshape(c, 1, *sshape), kernel, padding=padding, stride=1)
    features = features.view(c, -1, *sshape)  # Cx(K_h * K_w * K_d)xHxWxD
```
I think we can get this to work in 2D as well.

monai/networks/layers/simplelayers.py
dzenanz: The tests pass with this change.

monai/networks/layers/simplelayers.py
dzenanz: The tests pass with this change.

monai/networks/layers/simplelayers.py
dzenanz: It turns out it is. The latest commit makes the change.

dzenanz:
I am stuck again. Judging by debugger's inspection, we are passing both 4D tensor and 4D kernel for 2D test case to `F.conv3d()` on line 475.

dzenanz:
What we really want is [pad with 'replicate' mode](https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html?highlight=pad#torch.nn.functional.pad). Otherwise zeroes creep in all around the edges. With the small test images that is super-obvious, but it might be hard to debug with realistic-sized images later on, especially for other people.

@wyli If you can take a crack at this modification, that would be great. Otherwise I could give it a shot later, perhaps in a new PR.

wyli:
> What we really want is [pad with 'replicate' mode](https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html?highlight=pad#torch.nn.functional.pad). Otherwise zeroes creep in all around the edges. With the small test images that is super-obvious, but it might be hard to debug with realistic-sized images later on, especially for other people.
> 
> @wyli If you can take a crack at this modification, that would be great. Otherwise I could give it a shot later, perhaps in a new PR.

sure, I'll stop here with constant padding 0 for a basic implementation...

wyli:
added an example here https://github.com/Project-MONAI/DocImages/pull/16

wyli (approved):
looks good to me, padding mode and separable approximations are left for future PRs.

Issue #3843, opened 2022-02-23
PersistentDataset, CacheDataset improvements


myron:
It would nice to have several improvements to caching of data on disk and in memory. 

For caching on disk with **PersistentDataset**, 
-  the first epoch takes a long time, because saving is synchronous. I suggest saving on background queue, and returning data right away. This has a small probability that the same image might be saved twice, but on average it will have a speed up
- the documentation description implies that that cache_dir is optional (but it's not) "..If specified, this is the location for persistent storage .."
- The cached data is much larger on disk than the originals (and takes a lot of space). This is due to the original .nii.gz image is usually uint8/16 for images and uint8 for labels and gziped inside of nii.gz.  I wonder if there is a way to do similar for the cache

For caching in memory with **CacheDataset**
- on multigpu machine, each process gets its own cache, and they don't have access to any shared cache. It seems as of python 3.8 we can create a shared memory object https://docs.python.org/3/library/multiprocessing.shared_memory.html, so that all process can use it. it would be nice to have this, otherwise it's not very practical to use CacheDataset on multigpu machine. I know there is a way to manually partition data between processes, but then we need to worry about accuracy due to less random data sampling. 
- In memory caching runs before any training starts, which takes time. Can we do it on the fly (similar to PersistentDataset) as we iterate the first epoch, we add data into memory cache. 

Thank you

Nic-Ma:
Hi @myron ,

Thanks very much for your detailed feedback.
1. I can try to investigate the `multi-thread` saving logic later, see whether the total training time is really faster.
2. For the `cache_dir` arg, I submitted a closed PR before: https://github.com/Project-MONAI/MONAI/pull/3453, @wyli described some opinions in that PR.
3. If you are OK to save the data in `uint8`, you can put a `CastToType` transform before caching:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/utility/array.py#L313

4. Let me investigate this new shared object in python 3.8, the reason we didn't use shared memory in `CacheDataset` is that the IPC of shared memory makes training much slower.
5. If computing cache on the fly, we may need to use shared memory first, that's same question as above.
Please note that the multi-gpu case is "multi-processing in multi-processing" because PyTorch DataLoader is based on python `multiprocessing` and distributed launch is based on python `subprocess`.

Thanks.

myron:
>If you are OK to save the data in uint8, you can put a CastToType transform
[...]

PR #5308, opened 2022-10-11
SharedCacheDataset

myron:
This introduces a new SharedCacheDataset(Dataset) class to address  https://github.com/Project-MONAI/MONAI/issues/3843

It has several improvements / changes compared to the CacheDataset https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/dataset.py#L690

- It caches into a shared memory in RAM, so all processes in multigpu run with DistributedDataParallel can read/write to the same shared memory (globally accessible)
- it works on the fly  (there is non need to wait to pre-cache the data),  it will cache it on the first data access, and then use it from the cache

If tested it and used for many  tasks already in DDP / multigpu
I'll add unit tests too if it looks good

### Types of changes

### DIFF
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -20,6 +20,7 @@ import threading
 import time
 import warnings
 from copy import copy, deepcopy
+from multiprocessing.managers import ListProxy
 from multiprocessing.pool import ThreadPool
 from pathlib import Path
 from typing import IO, TYPE_CHECKING, Any, Callable, Dict, List, Optional, Sequence, Union
@@ -868,6 +869,126 @@ class CacheDataset(Dataset):
         data = self._cache[index_]
         if not isinstance(self.transform, Compose):
             raise ValueError("transform must be an instance of monai.transforms.Compose.")
+        for _transform in self.transform.transforms:
+            if start_run or isinstance(_transform, Randomizable) or not isinstance(_transform, Transform):
+                # only need to deep copy data on first non-deterministic transform
+                if not start_run:
+                    start_run = True
+                    if self.copy_cache:
+                        data = deepcopy(data)
+                data = apply_transform(_transform, data)
+        return data
+
+
+class SharedCacheDataset(Dataset):
+    """
+    Dataset with a shared cache among the processes. Particularly useful in DistributedDataParallel
+    multigpu run, when each process is able to read/write to the same shared cache list and
+    collectively cache the whole dataset in RAM.
+
+    Leading subset of non-random transforms are cached to accelerate the training data pipeline.
+    The transforms which are supposed to be cached must implement the `monai.transforms.Transform`
+    interface and should not be `Randomizable`. This dataset will cache the outcomes before the first
+    `Randomizable` `Transform` within a `Compose` instance.
+    So to improve the caching efficiency, please always put as many as possible non-random transforms
+    before the randomized ones when composing the chain of transforms.
+
+    For example, if the transform is a `Compose` of::
+
+        transforms = Compose([
+            LoadImaged(),
+            EnsureChannelFirstd(),
+            Spacingd(),
+            Orientationd(),
+            ScaleIntensityRanged(),
+            RandCropByPosNegLabeld(),
+            ToTensord()
+        ])
+
+    when `transforms` is used in a multi-epoch training pipeline, before the first training epoch,
+    this dataset will cache the results up to ``ScaleIntensityRanged``, as
+    all non-random transforms `LoadImaged`, `EnsureChannelFirstd`, `Spacingd`, `Orientationd`, `ScaleIntensityRanged`
+    can be cached. During training, the dataset will load the cached results and run
+    ``RandCropByPosNegLabeld`` and ``ToTensord``, as ``RandCropByPosNegLabeld`` is a randomized transform
+    and the outcome not cached.
+
+    Note:
+        Unlike `CacheDataset` class, `SharedCacheDataset` caches the full dataset in the shared memory (via ListProxy),
+        so each process within DistributedDataParallel can access items previously cached by other processes.
+        Data is cached on the fly, so there is no need to wait to cache all the data beforehand.
+    """
+
+    def __init__(
+        self,
+        data: Sequence,
+        transform: Union[Sequence[Callable], Callable],
+        copy_cache: bool = False,
+        as_contiguous: bool = True,
+        cache_list: Optional[ListProxy] = None,
+        use_cache: bool = True,
+    ) -> None:
+        """
+        Args:
+            data: input data to load and transform to generate dataset for model.
+            transform: transforms to execute operations on input data.
+            copy_cache: whether to `deepcopy` the cache content before applying the random transforms,
+                default to `False`.
+            as_contiguous: whether to convert the cached NumPy array or PyTorch tensor to be contiguous.
+                it may help improve the performance of following logic.
+            cache_list: ListProxy instance created by a master process, to hold the shared memory list
+                It must be crated on the master process e.g. as cache_list = multiprocessing.Manager().list()
+                before spawning/forking the subprocesses. If no cache_list is provided, a local (per process)
+                non-shared cache list will be created (could be sufficient in single gpu environment)
+            use_cache: whether to use cache mechanism. Defaults to True.  When False, the logic becomes equivalent
+                to a Dataset super class (no caching)
+        """
+        if not isinstance(transform, Compose):
+            transform = Compose(transform)
+        super().__init__(data=data, transform=transform)
+        self.copy_cache = copy_cache
+        self.as_contiguous = as_contiguous
+
+        if cache_list is None and use_cache:
+            cache_list = torch.multiprocessing.Manager().list()
+
+        if cache_list is not None:
+            cache_list[:] = [None] * len(data)
+
+        self._cache = cache_list
+
+    def _load_cache_item(self, idx: int):
+        """
+        Args:
+            idx: the index of the input data sequence.
+        """
+        item = self.data[idx]
+        for _transform in self.transform.transforms:  # type:ignore
+            # execute all the deterministic transforms
+            if isinstance(_transform, Randomizable) or not isinstance(_transform, Transform):
+                break
+            _xform = deepcopy(_transform) if isinstance(_transform, ThreadUnsafe) else _transform
+            item = apply_transform(_xform, item)
+        if self.as_contiguous:
+            item = convert_to_contiguous(item, memory_format=torch.contiguous_format)
+        return item
+
+    def _transform(self, index: int):
+        """
+        Args:
+            index: the index of the input data sequence.
+        """
+        if self._cache is None:
+            return super()._transform(index=index)
+
+        data = self._cache[index]
+
+        # if data is not in cache yet, transform (non-randoms) and cache it
+        if data is None:
+            data = self._load_cache_item(index)
+            self._cache[index] = data
+
+        # proceed with randomizable transforms
+        start_run = False
         for _transform in self.transform.transforms:
             if start_run or isinstance(_transform, Randomizable) or not isinstance(_transform, Transform):
                 # only need to deep copy data on first non-deterministic transform


myron:
I've added unit tests. please review. thanks

Nic-Ma:
Hi @myron ,

Thanks very much for the implementation of this feature.
1. I see much duplicated code if compared to the existing `CacheDataset`, is it possible to enhance the original `CacheDataset` directly? Or inherit from it?
2. Have you tested it with different multi-processing methods? like `spawn` and `fork`?
https://pytorch.org/docs/stable/multiprocessing.html
3. Please also test the case that we cache data on GPU (with `ToDevice` transform), refer to:
https://github.com/Project-MONAI/tutorials/blob/main/acceleration/fast_model_training_guide.md#4-cache-io-and-transforms-data-to-gpu
5. I remember `shared memory` is much slower, that's why we cache duplicated data for every rank in multi-gpu training for relatively small dataset. Could you please help provide some speed comparison when doing multi-gpu training with `CacheDataset` and this `SharedCacheDataset`?

Thanks in advance.

myron:
Hi @Nic-Ma , thank you for the feedback

>     1. I see much duplicated code if compared to the existing `CacheDataset`, is it possible to enhance the original `CacheDataset` directly? Or inherit from it?

There is only one similar method.  Sublclassing would be too messy  and confusing, since the use case is different. Modifying CacheDataset will make it very long (and prone to  bugs). 

> 
>     2. Have you tested it with different multi-processing methods? like `spawn` and `fork`?
>        https://pytorch.org/docs/stable/multiprocessing.html

Yes, I have been using in both spawn and fork modes. 

> 
>     3. Please also test the case that we cache data on GPU (with `ToDevice` transform), refer to:
>        https://github.com/Project-MONAI/tutorials/blob/main/acceleration/fast_model_training_guide.md#4-cache-io-and-transforms-data-to-gpu

This class is designed to for caching on CPU only,  it is not designed for GPU, I don't see a reason to test for it. 
We can add an assert statement to exit if the data is on gpu, if you'd like  

>     4. I remember `shared memory` is much slower, that's why we cache duplicated data for every rank in multi-gpu training for relatively small dataset. Could you please help provide some speed comparison when doing multi-gpu training with `CacheDataset` and this `SharedCacheDataset`?

It's true there is an overhead with shared memory.  It's definitely faster vs not-cashing at all.   
I don't have the bandwidth to write this code for speed comparisons at the moment.  Perhaps someone else would like to do this exercise.   Otherwise, in the interest of time, let's merge it for now, and later we can do further improvements / tuning.  thank you

myron:
> > 4. I remember `shared memory` is much slower, that's why we cache duplicated data for every rank in multi-gpu training for relatively small dataset. Could you please help provide some speed comparison when doing multi-gpu training with `CacheDataset` and this `SharedCacheDataset`?
> > ```
> 
> It's true there is an overhead with shared memory. It's definitely faster vs not-cashing at all. I don't have the bandwidth to write this code for speed comparisons at the moment. Perhaps someone else would like to do this exercise. Otherwise, in the interest of time, let's merge it for now, and later we can do further improvements / tuning. thank you

I ran the comparison in terms of time SharedCacheDataset vs CacheDataset, using a 3D segmentation task (416 images training, 104 validation) - evenly divisible data size for 8 gpu training. 

Set up on 8gpu (Nvidia A40 48gb x 8, ampere, latest pytorch, monai 1.0.0 dev branch):

- For CacheDataset pre-slit dataset into non-overlapping subsets.  Then each process caches it's subset before proceeding to training (using random sampling without DistributedSampler)
- For SharedCacheDataset,  (no extra setup , just using as a standard Dataset with DistributedSampler). It cashes images on the fly during first epoch. 

Timing:

First epoch:

- CachDataset: 20 sec to pre-cache everything, then train loop 45sec, validation loop 21sec
- SharedCacheDataset: trainloop 49sec, validation loop 22 sec (slightly longer train loop but, it includes caching, so no time wasted to wait for caching beforehand)

Second epoch (everything non-random is already cached)

- CachDataset: train loop 39 sec, validation loop 21sec
- SharedCacheDataset:  train loop 39 sec, validation loop 21sec

no difference in time on subsequent epochs.  If there is any overhead with using shared memory, it's probably hidden within DataLoader parallel loads (that are preloading in parallel to gpu runs). 

 So overall it seems SharedCacheDataset has all the advantages: caches on the fly, no need for special pre-splitting of data (that limits randomness). PS: this an experiment on particular 3D tumor segmentation task from CT, on other tasks timing could be slightly different.

wyli:
agree with @Nic-Ma, if it's always better the changes should be made to `CachDataset` instead of creating another class.

Nic-Ma:
Hi @myron ,

Cool, glad to see your test results, could you please help change it to update the `CacheDataset` directly?
And maybe you can help also double confirm whether different `num_workers` of DataLoader will affect the test results.

Thanks in advance.

myron:
> Cool, glad to see your test results, could you please help change it to update the `CacheDataset` directly? And maybe you can help also double confirm whether different `num_workers` of DataLoader will affect the test results.
> 
> Thanks in advance.

@Nic-Ma , thanks, CacheDataset is quite a bit different, if I update it it will break many old use cases.  Also perhaps someone prefers to  use CacheDataset as is (as it works for gpu caching too).  For num_workers, I used 3, but, we could investigate it's performance later. 

Can we merge this please, to unblock the overall progress. This is a new class, so it won't break anything. In the future we can consider more performance tests, or think of merging with other classes.

wyli:
this is a nice feature, all of us don't want to block it... @Nic-Ma and I just wanted to ensure the API is clean and maintainable. simply copying and modifying existing classes will add to the technical debts. hope we are not creating trouble for you. perhaps we can invite other people to collaborate on this if you don't have time.

myron:
> this is a nice feature, all of us don't want to block it... @Nic-Ma and I just wanted to ensure the API is clean and maintainable. simply copying and modifying existing classes will add to the technical debts. hope we are not creating trouble for you. perhaps we can invite other people to collaborate on this if you don't have time.

@wyli  I understand what you're saying, and I agree with you that we need a clean maintainable API.  So if you'd like to invite other people to collaborate here and they do it, it'd be great.  My only request is that we merge it soon , as we need it for Auto3Dseg to unblock that work as soon as possible.  PS: this PR has been waiting for 7 days already.  

The related issue https://github.com/Project-MONAI/MONAI/issues/3843 has been open for 8 months (since February), it doesn't look like we have much interest from other people to collaborate or contribute to it. I want to avoid the situation when it will take forever to bring any novel developments into our system. Since this class is very useful and not conflicting with anything existing, my preference is to merge it now, and after that, for sure, people can have a thorough discussion how to improve the cleanness of the API or unify some classes if they want.

Nic-Ma:
Hi @myron ,

OK, let me try to investigate more about this feature, can I commit to this PR directly? Or I create another PR based on this.

Thanks.

Nic-Ma:
Hi @myron ,

I took a deep look at the solution of this PR, and got some questions:
1. Seems it requires the master process to provide shared-memory list, how can we use it with `torchrun` or `torch.launch`?
2. If you use the `torch.multiprocessing` method, can you use `fork` and initialize `CacheDataset` before launching multiprocessing? Then I think all the subprocesses will share the master's list because it's `copy-on-write`. Same as the multiprocessing logic in `CacheDataset + DataLoader` when `num_worker > 0`?

I am checking: https://pytorch.org/docs/stable/multiprocessing.html?highlight=torch%20multiprocessing#module-torch.multiprocessing for more information.

CC @ericspod @wyli for more discussion.

Thanks in advance.

monai/data/dataset.py
```diff
@@ -878,6 +879,127 @@ def _transform(self, index: int):
...
+        """
+        Args:
+            data: input data to load and transform to generate dataset for model.
+            transform: transforms to execute operations on input data.
+            copy_cache: whether to `deepcopy` the cache content before applying the random transforms,
+                default to `False`.
+            as_contiguous: whether to convert the cached NumPy array or PyTorch tensor to be contiguous.
+                it may help improve the performance of following logic.
+            cache_list: ListProxy instance created by a master process, to hold the shared memory list
+                It must be crated on the master process e.g. as cache_list = multiprocessing.Manager().list()
```
ericspod, line 938: "crated" -> "created". Should also mention users shouldn't provide this argument and so maybe should be last in the argument list.

monai/data/dataset.py
```diff
@@ -878,6 +879,127 @@ def _transform(self, index: int):
...
+                before spawning/forking the subprocesses. If no cache_list is provided, a local (per process)
+                non-shared cache list will be created (could be sufficient in single gpu environment)
+            use_cache: whether to use cache mechanism. Defaults to True.  When False, the logic becomes equivalent
+                to a Dataset super class (no caching)
+        """
+        if not isinstance(transform, Compose):
+            transform = Compose(transform)
```
ericspod, line 945: If `transform` is not a sequence an exception should be raised here.

monai/data/dataset.py
```diff
@@ -878,6 +879,127 @@ def _transform(self, index: int):
...
+        super().__init__(data=data, transform=transform)
+        self.copy_cache = copy_cache
+        self.as_contiguous = as_contiguous
+
+        if cache_list is None and use_cache:
+            cache_list = torch.multiprocessing.Manager().list()
```
ericspod, line 951: I think the `Manager` instance should be kept as a member variable, it's not meant to be used in a throwaway fashion like this though the list you get probably retains a reference to it so everything works.

ericspod:
> 2\. can you use `fork` and initialize `CacheDataset`

Do be aware that `fork` is not present on Windows so special care has to be taken so that any multiprocessing semantics implemented here is going to work across platforms. Shared memory and other facilities are present in Windows but handles to such things have to be passed to subprocesses, [Process](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process) does that for you as does other multiprocessing facilities in the standard library and Pytorch. If what you want to do is cache in multiple processes and then use those processes again in the `Dataloader` I don't think that's going to be compatible.

Nic-Ma:
Hi @ericspod ,

I synced with @myron , I will take over the development of this feature in https://github.com/Project-MONAI/MONAI/pull/5365 and let's keep this PR open for comparison.

Thanks.

myron:
> Hi @myron ,
> 
> I took a deep look at the solution of this PR, and got some questions:
> 
>     1. Seems it requires the master process to provide shared-memory list, how can we use it with `torchrun` or `torch.launch`?

I don't know, I don't use torch.launch personally, I always start DistributedDataParaller from within the code itself, so that I can allocated the shared memory and provide it to the individual/spawned processes.

>     2. If you use the `torch.multiprocessing` method, can you use `fork` and initialize `CacheDataset` before launching multiprocessing? Then I think all the subprocesses will share the master's list because it's `copy-on-write`. Same as the multiprocessing logic in `CacheDataset + DataLoader` when `num_worker > 0`:
>        https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py#L1022
I'm not sure either. I do use 'fork' usually with DDP,   but I don't initialize anything major before launching DDP. 

> CC @ericspod @wyli for more discussion.
> 
> Thanks in advance.

monai/data/dataset.py
myron: I'm not sure I understand. It's the opposite, the users Should provide this argument (cache_list), and should be initialized on the master process.

monai/data/dataset.py
myron: sounds good. thank you

monai/data/dataset.py
myron: okay, we can do that, it was working fine without the  member variable, but I see your point.

monai/data/dataset.py
ericspod: Ah ok nevermind then.

myron:
> Seems it requires the master process to provide shared-memory list, how can we use it with `torchrun` or `torch.launch`?

I found a way to initialize cache without master process, by initializing it on first rank , and broadcasting
see here (I pushed the update to demostrate) https://github.com/Project-MONAI/MONAI/pull/5308/files#diff-e682e654b07b9d753cac73f5a782e70644c82a84d30ebc757ec294a261845108R961-R966

caveat, but this will only works if this was called before
`torch.cuda.set_device(device)`
(it is required to be called before creating the SharedDataset)

caveat2, in multi-gpu, all processes must create the SharedDataset, to be able to broadcast the cache, otherwise there will a deadlock.

but with this approach , we don't need to initialize cache in master process, and it doesn't need to be an input parameter tat all.

@Nic-Ma what do you think?

Nic-Ma:
Hi @myron ,

Thanks for the sharing, sounds interesting, let me test it and update PR https://github.com/Project-MONAI/MONAI/pull/5365 to include it.

Thanks.

PR #5317, opened 2022-10-12
Add generic kernel transform with support for multiple kernels

### LINKED ISSUE #3178: Antialiasing for labels
Low resolution labels are often too jagged.
image

Originally I hoped that the `Spacingd` transform would just have some antialiasing.
But a separate transform for this might be a better solution anyway.

I implemented a simple transform using a gaussian filter combined with a threshold on every label.

```python
class Antialiasingd(MapTransform):

    def __init__(
        self,
        keys: KeysCollection,
        sigma: Union[Sequence[float], float] = 1.0,
        approx: str = "erf",
        threshold: float = 0.5,
    ) -> None:
        self.sigma = sigma
        self.approx = approx
        self.threshold = threshold

    def __call__(self, data: Mapping[Hashable, NdarrayTensor]) -> Dict[Hashable, NdarrayTensor]:
        d = dict(data)
            img = d[key]

            gaussian_filter = GaussianFilter(img.ndim - 1, self.sigma, approx=self.approx)

            labels = torch.unique(img)[1:]
            new_img = torch.zeros_like(img)
            for label in labels:
                label_mask = (img == label).to(torch.float)
                blurred = gaussian_filter(label_mask.unsqueeze(0)).squeeze(0)
                new_img[blurred > self.threshold] = label
            d[key] = new_img
        return d
```

What do you think?

**Here some results**

| Parameters      | Result |
| :---        |    :----:   | 
| Original     | image    |
| Sigma=1.0, Threshold=0.5   |  image      |
| Sigma=4.0, Threshold=0.5   |  image   |
| Sigma=4.0, Threshold=0.4   |  image     |

kbressem:
Signed-off-by: kbressem <kenobressem@gmail.com>

Fixes #3178

### Description
As discussed in #3178, this PR adds a more general interface to apply simple convolutions to images. This PR (re-)implements 2D and 3D versions for mean filtering, edge detection, outline detection, mask dilatation and image sharpening. 

### Detailed description
The main transform is `ImageFilter`. It comes with some preset filters that can be initialized using a string. Available presets are `["mean", "laplace", "elliptical", "sobel", "sharpen", "median", "gauss", "savitzky_golay"]`. For example, a transformation for mean filtering can be created with: 

```python
mean_filter = ImageFilter("mean", 3)
```
`ImageFilter` automatically detects whether the input is 2D or 3D and creates the appropriate filter on the first call. This filter will then be used in future calls of the transformation, so switching between 2D and 3D images is not supported. 

The initialization form string is only a convenience function. All preset filters are also available in `monai.networks.layers.simplelayers` as `nn.Module` subclasses. ImageFilter` can also be created directly from an `nn.Module`. For example, to create a transformation for mean filtering, the following is also possible: 

```python
filter_3d = MeanFilter(spatial_dims=3, size=3)
mean_filter = ImageFilter(filter_3d)
```

In this example, however, the filter size is predefined and is not automatically derived from the input. 
In addition, it is also possible to use custom filters with `ImageFilter` using a `torch.Tensor` or `numpy.ndarray`. For example: 

```python
filter_3d = torch.ones(3,3,3)
mean_filter = ImageFilter(filter_3d)
```

`RandImageFilter` is a randomizable variant of `ImageFilter` that executes with probability `p` on each call. Both `ImageFilter` and `RandImageFilter` also have dictionary versions. 

### Types of changes


wyli:
thank you, for the first issue I think it'll be good to use `kwargs`, and in the docstring explain that the corresponding required arguments are described in each `create_[name]_kernel` function...

also precompute the kernel func would save a few cycles during `__call__`:
```py
   def __init__(self, kernel, spatial_dim, k_size, dtype, **kernel_args):
       if isinstance(kernel, str):
           func = self._create_kernel_from_string(name)
       self.kernel = func(spatial_dim, k_size, **kernel_args)
       self.dtype = dtype

  def __call__(self, ...):
      self.kernel.to(device=input.device, dtype=self.dtype or input.dtype)
```

for the 2nd issue, would be great to have a second issue/PR...

kbressem:
> also precompute the kernel func would save a few cycles during `__call__`:

The reason this is not done, is that I do not know if the image will be 2D or 3D until the function is called (see line 1564 in the array.py, where I extract the `ndim` from the image). 

However, I could precompute 2D and 3D kernels and just use the appropriate kernel at runtime?


> thank you, for the first issue I think it'll be good to use kwargs, and in the docstring explain that the corresponding required arguments are described in each create_[name]_kernel function (or saying k_size will be interpreted differently by each create kernel function)...

Ok, then I will use kwargs and add the information to the main docstring as well as the `each create_[name]_kernel` function

wyli:
> 
> However, I could precompute 2D and 3D kernels and just use the appropriate kernel at runtime?
> 

sure, i think that's a nice idea

ericspod:
I would suggest putting the kernel creation functions into a utilities file in transforms or networks as regular functions so that they are accessible to other if we wanted to use them for other purposes.

kbressem:
@wyli @ericspod I've updated the PR incorporating your suggestions. Sorry for the delay.

wyli (approved):
Thanks, it looks good to me.

PR #5338, opened 2022-10-15
Deep supervision loss wrapper class

myron:
This adds a DeepSupervisionLoss  wrapper class around the main loss function to accept a list of tensors returned from a deeply supervised networks. The final loss is computed as the sum of weighted losses for each of deep supervision levels (accounting for potential differences in shapes between targets and ds outputs)

The wrapper class is designed to work with arbitrary existing loss,e.g.
```
    loss = DiceCELoss(to_onehot_y=True, softmax=True) 
    ds_loss = DeepSupervisionLoss(loss)
```

Whereas the existing loss accepts the input as a single Tensor, ds_loss accepts the input as a list of Tensors (for each output of a deeply supervised network).  If only a simple Tensor input is provided, ds_loss behaves exactly the same as the underlying loss.

I added unit tests too.

### Types of changes


wyli (commented):
could you please help add an antialiasing option for the downsampling? (https://github.com/Project-MONAI/MONAI/issues/3178) `F.interpolate` has the option for 2D as well. the 3d could be done with the recent gaussian/median smoothing https://github.com/Project-MONAI/MONAI/blob/da1c32abbd77f9a8c2674f764be78aeb4828c1b6/monai/transforms/intensity/array.py#L59-L60

wyli (approved):

PR #5350, opened 2022-10-17
A small update to to ensure ds_loss returns a const

myron:
A small fix to followup on https://github.com/Project-MONAI/MONAI/pull/5338
to ensure  ds_loss returns a constant, not an array

### Types of changes


monai/losses/ds_loss.py
```diff
@@ -71,10 +71,8 @@ def get_loss(self, input: torch.Tensor, target: torch.Tensor):
     def forward(self, input: Union[torch.Tensor, List[torch.Tensor]], target: torch.Tensor):
 
         if isinstance(input, (list, tuple)):
-            loss = torch.zeros(1, dtype=torch.float, device=target.device)
-            for l in range(len(input)):
-                loss += self.get_loss(input[l].float(), target) * self.get_weight(l)
-            return loss
+            l = torch.stack([self.get_loss(input[l].float(), target) * self.get_weight(l) for l in range(len(input))])
```
wyli, line 74: the sum could be done in `torch.sum(self.get_loss...)` before stacking to have a smaller memory footprint? 

also at each training iteration, for each level, `get_weight` will do some string comparisons, could you move the comparisons to `__init__`?

mingxin-zheng:
Hi @myron I saw this was added to the Auto3DSeg milestone. I am having some trouble to understand how it relates to Auto3DSeg framework. Could you explain a bit? Thanks

myron:
> Hi @myron I saw this was added to the Auto3DSeg milestone. I am having some trouble to understand how it relates to Auto3DSeg framework. Could you explain a bit? Thanks

we used segresnet with deepsupervision, and this loss is to support deep supervision.

monai/losses/ds_loss.py
myron: @wyli , thanks for the comments

> the sum could be done in torch.sum(self.get_loss...) before stacking to have a smaller memory footprint?

self.get_loss() already returns a single number (nothing to sum). We're stacking a list of losses from each scale (e.g. 4x1  for 4 scales) and then summing over it.  Previously I had it as 

`sum(self.get_loss() for i in range(4))`, but torchscript does not like it, so I changed it to 
`torch.sum(torch.stack([self.get_loss() for i in range(4)]))` form, which does the same.

> also at each training iteration, for each level, get_weight will do some string comparisons, could you move the comparisons to __init__ to save some cycles?

That's true if we put it in __init__ , then we need to provide the number of scale levels beforehand ,and it will be less flexible (if a user needs to change that dynamically during training). 
I did a small optimization, and returning a list of weights now (so there will be only 1 string comparison instead of 4 (for 4 scales)).   I think the overhead of this is negligible.  

let's please merge it. thank you

wyli: I don't think it's necessary to assume `self.get_loss` returns a single number. most of the pytorch loss function supports different 'reduction' mode: https://github.com/Project-MONAI/MONAI/blob/d327088e690ec0899dda09db1f83d0bb8971a425/monai/losses/dice.py#L53
this wrapper should be generic.

myron: okay, but if the main loss() function doesn't return a single number, then we do not know what to do with it.  
We can do sum() or mean() or something else , perhaps loss() returns a tuple/list of  several tensors.  we do not know how to handle such unreduced outputs. 

 It think it should be a job of main loss() function to reduce the output before returning.  We could add it do docs/description to indicate that DeepSupervisionLoss expects loss to return reduced output (float)? would you like that?
 
 (also this small PR, is just a fix to the previous PR that is already merged, it doesn't change the functionality,  can we please merge it, and then think how to generalize it). thank you

wyli: the current dev's implementation happens to be compatible with any loss shape, I thought it's by your design...

myron: I see, it is somewhat a coincidence, let me take a look

myron: Alright, I've updated the code, I kept the loop summation (as it was before). I've avoided returning array by initializing accumulator as `torch.zeros([])` instead of `torch.zeros([1])`.
This works for techscript too. @wyli  plz check

myron: hmm, so som PyT versions didn't like it either for torchscript. I've updated it again to `torch.tensor(0)`

Issue #3989, opened 2022-03-23
Add warmup_start_lr and eta_min to WarmupCosineSchedule


holgerroth:
The current implementation of [WarmupCosineSchedule](https://github.com/Project-MONAI/MONAI/blob/fbd09a10cbdebff332aed4f95fd61dd28f0aee14/monai/optimizers/lr_scheduler.py#L59) will start the learning rate from 0. If step() is only called after the first epoch as [here](https://github.com/Project-MONAI/tutorials/blob/master/multimodal/openi_multilabel_classification_transchex/transchex_openi_multilabel_classification.ipynb), it causes the entire first epoch to be executed with a learning rate of zero.

Alternative implementations allow the user to set a `warmup_start_lr` and `eta_min` to avoid starting from or going down to zero, respectively, for example [here](https://lightning-bolts.readthedocs.io/en/stable/deprecated/schedulers/warmup_cosine_annealing.html#linear-warmup-cosine-annealing).

wyli:
(reminds me that I had a similar comment a while ago for the upstream implementation... https://github.com/Project-MONAI/MONAI/issues/2183#issuecomment-842978322)

ahatamiz:
The intended implementation requires `step()` to be called once before training initializes. Adding `eta_min` will alter the scheduler  behavior, hence is not desired. The proposed scheduler in my opinion better be added separately to avoid such issues.

holgerroth:
You can still default eta_min to 0. Hence, the behavior of the scheduler would not change. Adding those arguments would just make a more flexible implementation available in monai.

holgerroth:
I created a related issue for the tutorial https://github.com/Project-MONAI/tutorials/issues/630

ahatamiz:
Hi @holgerroth,

Thanks for the comment. I respectfully disagree. The scheduler is generic and not designed for epoch-based or iteration-based training. In case of iteration-based, it works as desired. For epoch-based, user needs to call it once before training. Adding an extra parameter `eta_min` would deviate from original implementation and would not solve the epoch-based issue, if the default is set to be `0`. The user still needs to adjust `eta_min` , if the desired behavior is to start with a non-zero learning rate. 

I strongly recommend submitting a new PR for a scheduler specifically designed for epoch-based training. A good candidate maybe [here](https://github.com/PyTorchLightning/Lightning-Bolts/blob/master/pl_bolts/optimizers/lr_scheduler.py#L10-L120). 

[...]

PR #5351, opened 2022-10-17
Enhancement for WarmupCosineSchedule to specify the beginning fraction of the linear warmup

myron:
small enhancement to WarmupCosineSchedule input, to optionally specify the beginning of the linear warmup from something above 0 ( e.g from a fraction 0.1 * initial_lr). 

a unit test is added too.

### Types of changes

### DIFF
--- a/monai/optimizers/lr_scheduler.py
+++ b/monai/optimizers/lr_scheduler.py
@@ -62,7 +62,13 @@ class WarmupCosineSchedule(LambdaLR):
     """
 
     def __init__(
-        self, optimizer: Optimizer, warmup_steps: int, t_total: int, cycles: float = 0.5, last_epoch: int = -1
+        self,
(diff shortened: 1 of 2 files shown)


wyli:
looks like this option will address https://github.com/Project-MONAI/MONAI/issues/3989, could you confirm?

myron:
> looks like this option will address #3989, could you confirm?

@wyli @holgerroth 
Yes, I looked at it, and it will solve the issue in https://github.com/Project-MONAI/MONAI/issues/3989.
It's to prevent the same situation when the first iterations to be zero (or almost zero) during linear warmup. 
The change is non-breaking, and the code will work as before, unless a user specifies this option.  I don't think we need a whole new Scheduler for such a small change.

wyli (approved):

PR #5365, opened 2022-10-20
5308 Add runtime cache support to `CacheDataset`

Nic-Ma:
Fixes #5308 .

### Description

This PR added runtime cache support in the `CacheDataset`.

### Types of changes


Nic-Ma:
Mark: current implementation will fail with GPU cache even when `num_workers=0` in `DataLoader`:
```
NotImplementedError: sharing CUDA metatensor across processes not implemented

  File "/workspace/data/medical/MONAI/tests/utils.py", line 601, in _wrapper
    raise RuntimeError(res.traceback) from res
RuntimeError: Traceback (most recent call last):
  File "/workspace/data/medical/MONAI/tests/utils.py", line 552, in run_process
    output = func(*args, **kwargs)
  File "/workspace/data/medical/MONAI/tests/utils.py", line 633, in _call_original_func
    return f(*args, **kwargs)
  File "/workspace/data/medical/MONAI/tests/test_integration_fast_train.py", line 146, in test_train_timing
    train_ds = CacheDataset(data=train_files, transform=train_transforms, cache_rate=1.0, num_workers=8)
  File "/workspace/data/medical/MONAI/monai/data/dataset.py", line 794, in __init__
    self.set_data(data)
  File "/workspace/data/medical/MONAI/monai/data/dataset.py", line 819, in set_data
    self._cache = Manager().list(_compute_cache())
  File "/opt/conda/lib/python3.8/multiprocessing/managers.py", line 740, in temp
    token, exp = self._create(typeid, *args, **kwds)
  File "/opt/conda/lib/python3.8/multiprocessing/managers.py", line 625, in _create
    id, exposed = dispatch(conn, None, 'create', (typeid,)+args, kwds)
  File "/opt/conda/lib/python3.8/multiprocessing/managers.py", line 87, in dispatch
    c.send((id, methodname, args, kwds))
  File "/opt/conda/lib/python3.8/multiprocessing/connection.py", line 206, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/opt/conda/lib/python3.8/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
  File "/workspace/data/medical/MONAI/monai/data/__init__.py", line 128, in reduce_meta_tensor
    raise NotImplementedError("sharing CUDA metatensor across processes not implemented")
NotImplementedError: sharing CUDA metatensor across processes not implemented
```

Thanks.

monai/data/dataset.py
```diff
@@ -789,8 +794,10 @@ def __init__(
         if self.num_workers is not None:
             self.num_workers = max(int(self.num_workers), 1)
         self.cache_num = 0
-        self._cache: Union[List, Dict] = []
+        self._cache: Union[List, Dict, DictProxy, ListProxy] = []
+        self.runtime_cache = runtime_cache
         self.set_data(data)
+        self.set_multiprocessing_cache(runtime_cache)
```
wyli, line 800: perhaps `set_data` should run after `set_multiprocessing_cache`?
```suggestion
        self.set_multiprocessing_cache(runtime_cache)
        self.set_data(data)
```

monai/data/dataset.py
```diff
@@ -773,6 +776,8 @@ def __init__(
                 the dataset has duplicated items or augmented dataset.
             hash_func: if `hash_as_key`, a callable to compute hash from data items to be cached.
                 defaults to `monai.data.utils.pickle_hashing`.
+            runtime_cache: whether to compute cache at the runtime, default to `False` to prepare
+                the cache content at initializaiton.
```
wyli, line 780: may need some docstring about the dataloader's `num_workers`, if `num_workers==0`, this `set_multiprocessing_cache(False)` is needed..

wyli (approved):
Integration, platform tests all work fine

monai/data/dataset.py
Nic-Ma: Hi @wyli , if `num_workers==0`, no need to call `set_multiprocessing_cache(False)`. It's only necessary when caching on GPU memory, and the `monai.data.DataLoader` will seamlessly take care of it in this PR.

Thanks.

monai/data/dataset.py
Nic-Ma: updated in the latest commit.

Thanks.

monai/data/dataset.py
Nic-Ma: added doc-string in the latest commit.

Thanks.

Nic-Ma:
Hi @myron ,

I tested the `on-the-fly` caching method vs caching before the training with the spleen fast-training pipeline, both with on 1 GPU:
1. With `num_workers=0` in DataLoader and caching on **GPU**
`on-the-fly`: total time = 300s, `pre caching`: total time = 209s + 51s(pre caching)
2. With `num_workers=1` in DataLoader (`fork` method multiprocessing shared cache) and caching on **CPU**
`on-the-fly`: first epoch = 135s, every following epoch  = 3.1s
`pre caching`: pre caching = 42s, every epoch = 2.92s

Seems the `on-the-fly` caching is slower than the pre-caching in single GPU training, I think that because the multithread pre-caching is faster in this case.

Then I tested with `torch.launch` for 2 GPUs with the Brats pipeline, `num_workers=4` for DataLoader and cache rate=0.2 due to my limited GPU memory:
`on-the-fly`: first epoch = 208s, every following epoch  = 188s
`pre caching`: pre caching = 27s, every epoch = 189s
Seems `on-the-fly` is slightly faster in this case.

And if want to use the `torch.multiprocessing.spawn()` method to avoid duplicated caching in each process, you can try below code (I tested locally):
```py
train_ds = CacheDataset(...)
val_ds = CacheDataset(...)
torch.multiprocessing.spawn(main_worker, nprocs=2, args=(args, train_ds, val_ds))
```

Thanks.

monai/data/dataset.py
Nic-Ma: Hi @wyli ,

Actually, here we should call `set_data()` first, the logic is to convert the set data to multiprocessing list or regular list.

Thanks.

wyli:
might be useful to update the docstring or the tutorial to explain the recommended usage? currently it's unclear to me when to use which option..

Nic-Ma:
Hi @wyli ,

I think `runtime_cache=True` is not always better, depends on dataset, transforms, and num workers, and GPU / CPU hardware, etc.
Prefer to non-breaking change.
What do you think?

Thanks.

myron:
I tried your code and I get multiple errors (I can't get it to work):
`AttributeError: Can't pickle local object ....`


both in spawn and fork mode in multiprocessing. 
 Passing the whole CacheDataset object to torch.multiprocessing.spawn  would require it all to be pickled/serialized, and that is a large object that includes all the chain of transforms inside.
This changes the logic too, now we'll have to create the Dataset in the master process, and will not to be able to customize it for each rank / sub-process. 


What is the goal of this PR? on the top it says it fixes SharedCacheDataset PR https://github.com/Project-MONAI/MONAI/pull/5308?

The SharedCacheDataset was specifically to use shared cache in **multiprocessing multi-gpu** environment,
 so if this PR is to accomplish the same it should be tested in multi-gpu (not on 1 GPU) on several tasks to measure memory consumption, and testing if various processes indeed share the same shared cache.

Nic-Ma:
Hi @myron ,

Thanks for your testing.
Sorry I think the PR description was copied from somewhere and not accurate, I updated it now.
And I understand your feature request of supporting the shared memory object as args, I didn't get a perfect solution, so I implemented this `runtime cache` feature first, will think about it again.

Thanks.

Nic-Ma:
> I tried your code and I get multiple errors (I can't get it to work): `AttributeError: Can't pickle local object ....`
> 
> both in spawn and fork mode in multiprocessing. Passing the whole CacheDataset object to torch.multiprocessing.spawn would require it all to be pickled/serialized, and that is a large object that includes all the chain of transforms inside. This changes the logic too, now we'll have to create the Dataset in the master process, and will not to be able to customize it for each rank / sub-process.
> 
> What is the goal of this PR? on the top it says it fixes SharedCacheDataset PR #5308?
> 
> The SharedCacheDataset was specifically to use shared cache in **multiprocessing multi-gpu** environment, so if this PR is to accomplish the same it should be tested in multi-gpu (not on 1 GPU) on several tasks to measure memory consumption, and testing if various processes indeed share the same shared cache.

Hi @myron ,

With PR https://github.com/Project-MONAI/MONAI/pull/5398, it can unify the cache to `list`, then maybe we can add another arg `shared_list` for `CacheDataset` later as your original PR to avoid passing the whole CacheDataset object.
What do you think?

Thanks.

myron:
> > I tried your code and I get multiple errors (I can't get it to work): `AttributeError: Can't pickle local object ....`
> > both in spawn and fork mode in multiprocessing. Passing the whole CacheDataset object to torch.multiprocessing.spawn would require it all to be pickled/serialized, and that is a large object that includes all the chain of transforms inside. This changes the logic too, now we'll have to create the Dataset in the master process, and will not to be able to customize it for each rank / sub-process.
> > What is the goal of this PR? on the top it says it fixes SharedCacheDataset PR #5308?
> > The SharedCacheDataset was specifically to use shared cache in **multiprocessing multi-gpu** environment, so if this PR is to accomplish the same it should be tested in multi-gpu (not on 1 GPU) on several tasks to measure memory consumption, and testing if various processes indeed share the same shared cache.
> 
> Hi @myron ,
> 
> With PR #5398, it can unify the cache to `list`, then maybe we can add another arg `shared_list` for `CacheDataset` later as your original PR to avoid passing the whole CacheDataset object. What do you think?
> 
> Thanks.

Well, yes, it's possible. Thank you for suggesting.  I've re-read your PR #5398 about "unify the cache to `list`", and it seems it's implementation currently has issues, and it's not trivial to improve it (I've updated my review comments there, please check)

Alternative:  it might be easier to keep supporting List and Dict cache as is, but have a  parameter `shared_cache: Union[ListProxy, DictProxy] `, and it will be up to a user to supply the correct "shared cache" object in case of "hashing by key" (if we keep it)

myron:
> Hi @myron ,
> 
> Thanks for your suggestions. My concern is that: currently, users don't need to care about the internal caching format. If we let users prepare the `list` or `dict` differently according to the `hash_as_key` arg, I feel it will make the API harder to use. So I think if we always use `list` as the cache content, it's slightly easier for users at least. What do you think?
> 
> Thanks.

yes, it sound good. I think it's easier , but adds a little bit of convoluted logic for that hash_by_key scenario.

Nic-Ma:
@myron shared a method to share cache content across ranks:
```py
if dist.is_initialized():
    print('using shared cache via broadcasting in DDP, torch.cuda.set_device(device) must be set before')
    object_list=[cache_list,]
    dist.broadcast_object_list(object_list, src=0)
    cache_list = object_list[0]
```
I will test it and enhance the PR, still don't quite understand why it must execute before `set_device()`.

Thanks.

monai/data/dataset.py
```diff
@@ -773,6 +778,12 @@ def __init__(
                 the dataset has duplicated items or augmented dataset.
             hash_func: if `hash_as_key`, a callable to compute hash from data items to be cached.
                 defaults to `monai.data.utils.pickle_hashing`.
+            runtime_cache: whether to compute cache at the runtime, default to `False` to prepare
+                the cache content at initializaiton, if `True`, it will cache during the first epoch
+                of model training, so it can start the first mini-batch earlier.
+                to execute runtime cache on GPU memory, must co-work with `monai.data.DataLoader`.
+            cache_rank: if in distributed data parallel, only compute the cache on `cache_rank` and
```
myron, line 785: I don't think we need to expose cache_rank as a user option. We can always use rank 0. Regardless of who allocates it, the shared cpu cache in cpu RAM, shared memory.  I do not see a use case when a user needs to change it

monai/data/dataset.py
```diff
@@ -807,6 +821,20 @@ def set_data(self, data: Sequence):
...
+                cache = self._fill_cache(indices)
+            if self.runtime_cache:
+                cache = self._set_multiprocessing_cache(cache, mp_cache=True)
+
+            if self._is_dist:
+                obj_list = [cache]
+                dist.broadcast_object_list(obj_list, src=self.cache_rank)
+                cache = obj_list[0]
+            return cache
+
```
myron, line 837: - so here, if we don't do runtime_cache , but is_dist==True and rank!=cache_rank,  the cache list will be all Nones (without precomputing). I don't think it's correct. 
When runtime_cache==False, we should always precompute on all ranks.  

- we can remove cache_rank user option.
- check for broadcast should be only under runtime_cache, since we use shared list only in that case

I'd change to:
 ```
  def _compute_cache(indices=None):

    if self.runtime_cache:
        cache = self._set_multiprocessing_cache([None,]*self.cache_num, mp_cache=True)
        if self._is_dist:
            obj_list = [cache]
            dist.broadcast_object_list(obj_list)
            cache = obj_list[0]

    else:
        cache = self._fill_cache(indices)
   
    return cache
```

File: monai/data/dataloader.py [hunk 5]
```diff
@@ -80,6 +80,12 @@ def __init__(self, dataset: Dataset, num_workers: int = 0, **kwargs) -> None:
             init_seed = _g.initial_seed()
             _seed = torch.empty((), dtype=torch.int64).random_(generator=_g).item()
             set_rnd(dataset, int(_seed))
+            # disable unnecessary multiprocessing caching
+            from monai.data.dataset import CacheDataset  # avoid circular import
+
+            if isinstance(dataset, CacheDataset) and dataset.runtime_cache:
+                dataset._cache = dataset._set_multiprocessing_cache(dataset._cache, False)
+
```
myron, line 88: - another option is to use `hasattr(dataset, '_set_multiprocessing_cache')`, then we don't need to worry about cicular import

- `dataset._cache = dataset._set_multiprocessing_cache(dataset._cache, False)` this seems like a call to a private api (both `._set_multiprocessing_cache and ._cache` are private)
May be we can do it as `dataset.ensure_non_shared()`, which will ensure that cache is List and not ListProxy inside

myron (changes requested):

monai/data/dataset.py
Nic-Ma: Hi @myron ,

For the "not runtime cache" case, I think you want to broadcast the precomputed cache content to every rank to avoid duplicated computation and cache, is it right?

Thanks.

myron: @Nic-Ma no, here we're only broadcasting the **Empty** ProxyList (pointer), so that all processes know where to read/write to shared cache.  This broadcasting should be done before any data caching begins

Nic-Ma: Updated to only broadcast for the `runtime cache` case.

Thanks.

monai/data/dataset.py
Nic-Ma: Sure, removed this arg in the latest commit.

Thanks.

monai/data/dataloader.py
Nic-Ma: Sure, I simplified the API to avoid private functions and private variables in the latest commit.
But I didn't use the "hasattr" method, I think `if isinstance(dataset, CacheDataset)` is more robust.

Thanks.

Nic-Ma:
Hi @myron ,

And I added a new unit tests for distributed training with `DistributedSampler`, seems this `broadcast` method can't work well with "multiprocessing for GPUs + multiprocessing for workers in dataloader" case. Could you please help take a look or verify the unit test locally?
The error message is:
```
  File "/workspace/data/medical/MONAI/tests/test_sampler_dist.py", line 60, in test_cachedataset
    for d in dataloader:
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 681, in __next__
    data = self._next_data()
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1374, in _next_data
    return self._process_data(data)
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1400, in _process_data
    data.reraise()
  File "/opt/conda/lib/python3.8/site-packages/torch/_utils.py", line 543, in reraise
    raise exception
FileNotFoundError: Caught FileNotFoundError in DataLoader worker process 0.
Original Traceback (most recent call last):
  File "/opt/conda/lib/python3.8/multiprocessing/managers.py", line 827, in _callmethod
    conn = self._tls.connection
AttributeError: 'ForkAwareLocal' object has no attribute 'connection'

During handling of the above exception, another exception occurred:

  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/worker.py", line 302, in _worker_loop
    data = fetcher.fetch(index)
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/fetch.py", line 49, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/fetch.py", line 49, in <listcomp>
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "/workspace/data/medical/MONAI/monai/data/dataset.py", line 107, in __getitem__
    return self._transform(index)
  File "/workspace/data/medical/MONAI/monai/data/dataset.py", line 906, in _transform
    data = self._cache[cache_index]
  File "<string>", line 2, in __getitem__
  File "/opt/conda/lib/python3.8/multiprocessing/managers.py", line 831, in _callmethod
    self._connect()
  File "/opt/conda/lib/python3.8/multiprocessing/managers.py", line 818, in _connect
    conn = self._Client(self._token.address, authkey=self._authkey)
  File "/opt/conda/lib/python3.8/multiprocessing/connection.py", line 502, in Client
    c = SocketClient(address)
  File "/opt/conda/lib/python3.8/multiprocessing/connection.py", line 630, in SocketClient
    s.connect(address)
FileNotFoundError: [Errno 2] No such file or directory
```

Thanks in advance.

Nic-Ma:
Hi @myron ,

I updated the PR according to your comments and also added new unit tests for the typical use-case of caching on the fly with multi-gpu and `DistributedSampler`.

Thanks in advance.

monai/apps/datasets.py
```diff
@@ -223,6 +227,8 @@ class DecathlonDataset(Randomizable, CacheDataset):
             may set `copy=False` for better performance.
         as_contiguous: whether to convert the cached NumPy array or PyTorch tensor to be contiguous.
             it may help improve the performance of following logic.
+        runtime_cache: whether to compute cache at the runtime, default to `False` to prepare
+            the cache content at initializaiton.
```
myron, line 231: We can (optionally) provide more information here. 

When using this option in multi-gpu distributed training, `torch.cuda.set_device(local_rank)` must be called before initializing this class.

myron (approved):

monai/apps/datasets.py
Nic-Ma: Sure, updated in the latest commit.

Thanks.

Nic-Ma:
I can't reproduce the CI error locally, let me change to a multi-gpu V100 server and test again.

Thanks.

myron:
looks great! thank you

PR #5377, opened 2022-10-21
5344 implement pq metric

### LINKED ISSUE #5344: Panoptic Quality metric for instance segmentation
Panoptic Quality is described here: https://openaccess.thecvf.com/content_CVPR_2019/papers/Kirillov_Panoptic_Segmentation_CVPR_2019_paper.pdf

This is a useful metric for instance segmentation and it would be great to have it in the HoVerNet pipeline.

Also used in this challenge for nuclei identification: https://conic-challenge.grand-challenge.org/Evaluation/

yiheng-wang-nv:
Fixes #5344 .

### Description

This PR implements the metric of Panoptic Quality.

### Types of changes


yiheng-wang-nv:
Update:
this draft PR is WIP, and so far I've implemented the class of the metric.
The implementation refers to:
https://github.com/TissueImageAnalytics/CoNIC/blob/main/misc/utils.py
and fixes some of its bugs, such as calculate when the predictions or labels do not have background (instance id = 0).

I discussed with @KumoLiu , and according to my understand of the [original paper](https://openaccess.thecvf.com/content_CVPR_2019/papers/Kirillov_Panoptic_Segmentation_CVPR_2019_paper.pdf), the predictions should only have 1 channel, since for each pixel, it should only belong to one instance. Therefore, the ground truth of an image with shape [H, W] must have the shape [H, W], thus to be consistent with other metrics in MONAI, I limit the input into shape [B1HW], am I understand right? Hi @drbeh , may need your help to confirm it, thanks!

drbeh:
> Update: this draft PR is WIP, and so far I've implemented the class of the metric. The implementation refers to: https://github.com/TissueImageAnalytics/CoNIC/blob/main/misc/utils.py and fixes some of its bugs, such as calculate when the predictions or labels do not have background (instance id = 0).
> 
> I discussed with @KumoLiu , and according to my understand of the [original paper](https://openaccess.thecvf.com/content_CVPR_2019/papers/Kirillov_Panoptic_Segmentation_CVPR_2019_paper.pdf), the predictions should only have 1 channel, since for each pixel, it should only belong to one instance. Therefore, the ground truth of an image with shape `[H, W]` must have the shape `[H, W]`, thus to be consistent with other metrics in MONAI, I limit the input into shape `[B1HW]`, am I understand right? Hi @drbeh , may need your help to confirm it, thanks!


Hi @yiheng-wang-nv, thanks for implementing PQ metrics.

You are right, in the paper, the PQ metrics is calculated per class for the input is one dimensional for each PQ calculation; however, these PQs are averaged afterwards to create the metrics that we can denote as `mPQ`, and this is the actual metrics that we need.

Unlike the rest of MONAI metrics that are designed for semantic segmentation, this metrics is used for instance segmentation and they can have multiple channels. Having said that, it still depends how the data is being represented, for instance look at the ground truth of this dataset, where one channel is for instances and one channel for classes: https://conic-challenge.grand-challenge.org/Data/

I believe we can have two classes, one for binary PQ metrics where it assumes only one-channel inputs, and another for multi-class PQ where it accepts two channels (one for instance and one for classes) and you call the binary PQ inside it for each class.
OR, maybe it is better to have one class but accepting two one-channel tensors (instance map and class map) so for the binary one the second tensor is not needed.

Please let me know if you have any question.

monai/metrics/panoptic_quality.py
```diff
@@ -0,0 +1,224 @@
...
+        match_iou: float = 0.5,
+        smooth_nr: float = 1e-6,
+    ) -> None:
+        super().__init__()
+        self.reduction = reduction
+        self.remap = remap
+        self.match_iou = match_iou
+        self.smooth_nr = smooth_nr
+
+    def _compute_tensor(self, y_pred: torch.Tensor, y: torch.Tensor):  # type: ignore
```
JHancox, line 68: I wonder whether we can assume that a torch tensor will be the input. Instance segmentation will need to have been computed before this metric can be calculated and this uses non-torch tensors. Should we be accepting ndarrays? 
Perhaps @drbeh has a view.

monai/metrics/panoptic_quality.py
```diff
@@ -0,0 +1,224 @@
...
+        """
+        Args:
+            y_pred: input data to compute, typical panoptic segmentation model output.
+                It must be in the form of B1HW[D] and has integer type.
+            y: ground truth. It must has the same shape as `y_pred` and has integer type.
```
JHancox, line 73: "It must be in the form of B1HW[D] and **have** integer type.
y: ground truth. It must **have** the same shape as `y_pred` and **have** integer type."

monai/metrics/panoptic_quality.py
```diff
@@ -0,0 +1,224 @@
...
+        data = self.get_buffer()
+        if not isinstance(data, torch.Tensor):
+            raise ValueError("the data to aggregate must be PyTorch Tensor.")
+
+        # do metric reduction
+        f, _ = do_metric_reduction(data, reduction or self.reduction)
+        return f
+
+
+def compute_panoptic_quality(
```
JHancox, line 121: I wondered why this is not a method of the PanopticQualityMetric class, which would reduce the need to pass all the parameters in. Is there is a reason to decouple these?

monai/metrics/panoptic_quality.py
```diff
@@ -0,0 +1,224 @@
...
+    https://openaccess.thecvf.com/content_CVPR_2019/papers/Kirillov_Panoptic_Segmentation_CVPR_2019_paper.pdf
+
+    Example of the typical execution steps of this metric class follows :py:class:`monai.metrics.metric.Cumulative`.
+
+    Args:
+        reduction: define mode of reduction to the metrics,
+            available reduction modes: {``"none"``, ``"mean"``, ``"sum"``, ``"mean_batch"``, ``"sum_batch"``},
+            default to ``"mean"``. if "none", will not do reduction.
+        remap: whether to remap `y_pred` and `y` to ensure contiguous ordering of instance id.
+        match_iou: IOU threshould to determine the paring between `y_pred` and `y`. Usually,
```
JHancox, line 47: "match_iou: IOU threshould to determine the **pairing** between `y_pred` and `y`. Usually,"
There are a couple of other instances where **paring** is used instead of **pairing**

monai/metrics/panoptic_quality.py
```diff
@@ -0,0 +1,224 @@
...
+        for b in range(batch):
+            outputs[b] = compute_panoptic_quality(
+                pred=y_pred[b, 0], gt=y[b, 0], remap=self.remap, match_iou=self.match_iou, smooth_nr=self.smooth_nr
+            )
+
+        return outputs
+
+    def aggregate(self, reduction: Union[MetricReduction, str, None] = None):
+        """
+        Execute reduction logic for the output of `compute_meandice`.
```
JHancox, line 104: Execute reduction logic for the output of `**compute_meandice**`. should be '**compute_panoptic_quality**'

monai/metrics/panoptic_quality.py
yiheng-wang-nv: Hi @JHancox , this part is to align with other metrics, such as: https://github.com/Project-MONAI/MONAI/blob/f407fcce1c32c050f327b26a16b7f7bc8a01f593/monai/metrics/meandice.py#L106

monai/metrics/panoptic_quality.py
yiheng-wang-nv: Hi @JHancox , thanks for posting this question, let me just modify the function to support both torch tensor and numpy array.

monai/metrics/panoptic_quality.py
```diff
@@ -0,0 +1,224 @@
...
+from .metric import CumulativeIterationMetric
+
+linear_sum_assignment, _ = optional_import("scipy.optimize", name="linear_sum_assignment")
+
+__all__ = ["PanopticQualityMetric", "compute_panoptic_quality"]
+
+
+class PanopticQualityMetric(CumulativeIterationMetric):
+    """
+    Compute panoptic quality between two tensors.
```
ericspod, line 28: Could we described briefly what the metric actually computes and what applications it would be for?

drbeh:
Hi @yiheng-wang-nv, please let me know when this PR is ready for review. Thanks

drbeh:
@yiheng-wang-nv, it would be great if this can output SQ (segmentation quality) and RQ (recognition quality) in addition to PQ which is the combination of them. Thanks

monai/metrics/panoptic_quality.py
yiheng-wang-nv: Hi @JHancox and @drbeh , sorry that now `CumulativeIterationMetric` requires the input to be tensor only, thus in order to unify with other monai metrics, PQ will stick to support torch tensor only.

yiheng-wang-nv:
Hi @drbeh , I modified the metric implementation to align with CoNiC challenge's version, and if calling `compute_panoptic_quality` directly, users can also do binary PQ calculation. As for the MPQ, we can use `PanopticQualityMetric` class and with the default setting: `reduction = "mean_batch"`, the metric will return three (1, N) tensors which stand for PQ, SQ and PR, where N equals to number of classes. The calculation is the same as in CoNiC's function.

Could you please help to re-review it? Thanks, and if the format meets the requirements, I'll add unit tests asap.

monai/metrics/panoptic_quality.py
```diff
@@ -0,0 +1,263 @@
...
+from monai.utils import MetricReduction, optional_import
+
+linear_sum_assignment, _ = optional_import("scipy.optimize", name="linear_sum_assignment")
+
+__all__ = ["PanopticQualityMetric", "compute_panoptic_quality"]
+
+
+class PanopticQualityMetric(CumulativeIterationMetric):
+    """
+    Compute Panoptic Quality between two tensors, except Panoptic Quality, two related metrics Segmentation Quality (SQ)
```
dzenanz, line 27:
```suggestion
    Compute Panoptic Quality between two tensors. Two related metrics Segmentation Quality (SQ)
````

Comment (yiheng-wang-nv) [reply monai/metrics/panoptic_quality.py]: Thanks @dzenanz !

File: docs/source/metrics.rst [hunk 8]
```diff
@@ -104,6 +104,13 @@ Metrics
 .. autoclass:: SurfaceDiceMetric
     :members:
 
+`PanopticQualityMetric`
+-----------------------
+.. autofunction:: compute_panoptic_quality
+
+.. autoclass:: PanopticQualityMetric
+    :members:
+
```
wyli, line 113:
```suggestion


```

monai/metrics/panoptic_quality.py
```diff
@@ -0,0 +1,282 @@
...
+    match_iou: float = 0.5,
+    smooth_nr: float = 1e-6,
+    output_confusion_matrix: bool = False,
+):
+    """Computes Panoptic Quality (PQ). If specifying `metric_name` to "SQ" or "RQ",
+    Segmentation Quality (SQ) or Recognition Quality (RQ) will be returned instead.
+
+    In addition, if `output_confusion_matrix` is True, the function will return a tensor with shape 4, which
+    represents the true positive, false positive, false negative and the sum of iou. These four values are used to
+    calculate PQ, and return them directly is able for further calculation over all images.
```
wyli, line 160:
```suggestion
    calculate PQ, and returning them directly enables further calculation over all images.
```

Nic-Ma (approved):

drbeh (changes requested):
Based on our today's conversation, I am reviewing it. Please let me finish it first.

monai/metrics/panoptic_quality.py
```diff
@@ -0,0 +1,282 @@
...
+            inter = (t_mask * p_mask).sum()
+            iou = inter / (total - inter)
+            pairwise_iou[true_id - 1, pred_id - 1] = iou
+
+    return pairwise_iou, true_id_list, pred_id_list
+
+
+def _get_paired_iou(pairwise_iou: torch.Tensor, match_iou: float = 0.5, device: Union[str, torch.device] = "cpu"):
+    if match_iou >= 0.5:
+        paired_iou = pairwise_iou[pairwise_iou > match_iou]
```
drbeh, line 254: It seems that this line is not used and can be deleted, right?

monai/metrics/panoptic_quality.py
```diff
@@ -0,0 +1,282 @@
...
+        paired_pred += 1
+
+        return paired_iou, paired_true, paired_pred
+
+    pairwise_iou = pairwise_iou.cpu().numpy()
+    paired_true, paired_pred = linear_sum_assignment(-pairwise_iou)
+    paired_iou = pairwise_iou[paired_true, paired_pred]
+    paired_true = torch.as_tensor(list(paired_true[paired_iou > match_iou] + 1), device=device)
+    paired_pred = torch.as_tensor(list(paired_pred[paired_iou > match_iou] + 1), device=device)
+    paired_iou = paired_iou[paired_iou > match_iou]
```
drbeh, line 268: Shouldn't they be first filtered for `paired_iou > match_iou` and then use `linear_sum_assignment`?
I haven't mathematically worked it out but it seems that the values less than the threshold value (e.g., 0.4) add to the cost while it doesn't matter what value they have (e.g., it does matter it is 0.01 or 0.03 because they won't be a match any way). Any thought?

monai/metrics/panoptic_quality.py
```diff
@@ -0,0 +1,282 @@
...
+            If set `match_iou` < 0.5, this function uses Munkres assignment to find the
+            maximal amout of unique pairing.
+        smooth_numerator: a small constant added to the numerator to avoid zero.
+
+    """
+
+    def __init__(
+        self,
+        num_classes: int,
+        metric_name: str = "pq",
```
drbeh, line 60: In this way if we want all the three outputs (`PQ`, `SQ` and `RQ`), we should run it three times? Would it be possible to return all of them too (or accept the list of str with any combination)?

monai/metrics/panoptic_quality.py
```diff
@@ -0,0 +1,282 @@
...
+    instance segmentation (detect and segment each object instance). Compared with semantic segmentation, panoptic
+    segmentation distinguish different instances that belong to same class.
+    Compared with instance segmentation, panoptic segmentation does not allow overlap and only one semantic label and
+    one instance id can be assigned to each pixel.
+    Please refer to the following paper for more details:
+    https://openaccess.thecvf.com/content_CVPR_2019/papers/Kirillov_Panoptic_Segmentation_CVPR_2019_paper.pdf
+
+    Different from the original panoptic quality, this class refers to the evaluation metric of CoNiC Challenge 2022,
+    a Nuclei instance segmentation and classification task. The referred implementation can be found in:
+    https://github.com/TissueImageAnalytics/CoNIC
```
drbeh, line 41: Could you please let me know what is the main difference? Thanks

drbeh (changes requested):
Hi @yiheng-wang-nv, nice work! I left few comments and it would be great if you can address them. Thanks

monai/handlers/panoptic_quality.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    Computes Panoptic quality from full size Tensor and collects average over batch, class-channels, iterations.
+    """
+
+    def __init__(
+        self,
+        num_classes: int,
+        metric_name: str = "pq",
+        reduction: Union[MetricReduction, str] = MetricReduction.MEAN_BATCH,
+        match_iou: float = 0.5,
```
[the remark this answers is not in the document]
yiheng-wang-nv: Thanks!

monai/metrics/panoptic_quality.py
yiheng-wang-nv: Hi @drbeh , it is mainly for the input side. CoNIC version defines the input to be a (batch, 2, H, W) format, and the 1st channel and 2nd channel represent instance map and class map respectively. For the original version, it only focus on how to calculate on (H, W) input.

drbeh: I see. Then maybe you can remove this comment (or elaborate on the difference).

monai/metrics/panoptic_quality.py
yiheng-wang-nv: Hi @drbeh , I get your point here, but I'm not sure if "filtering first, then mapping" is equivalent to "mapping first, then filtering". Therefore, I suggest here we follow the order of the original implementation (https://github.com/TissueImageAnalytics/CoNIC/blob/main/metrics/stats_utils.py#L143), what do you think?

Issue #5378, opened 2022-10-22
DiceMetric refactoring


myron:
Dice metric is probably the most popular accuracy metric used for medical images.  

Our current implementation is lacking features, slow and bug-prone since it's very convoluted coding. 
https://github.com/Project-MONAI/MONAI/blob/dev/monai/metrics/meandice.py#L22

Issues

1) Long subclass chain: DiceMetric  class -> CumulativeIterationMetric class -> Cumulative class + IterationMetric class -> Metric class. For something simple to compute such as Dice, it takes time understand it or debug 

2) The implementation is not intuitive, when calling class instance
For example
```
metric = DiceMetric(reduction = 'mean')
dice = metric(x,y)
```
Usually one would expect a mean dice value, instead we return an un-reduced  result. Futhermore we add intermediate result to "buffers", which is hidden from the the user and non-obvious
to get my mean dice, I'll need to do these steps
```
metric = DiceMetric(reduction = 'mean')
metric(x,y)
dice = metric.aggregate()
```
but even worse, if within my training loop I call this metric several times, I need to reset the buffers manually:
```
metric = DiceMetric(reduction = 'mean')
for i in range(100):
  metric(x[i],y[i])
  dice = metric.aggregate()
  metric.reset()
```
If we don't reset , then the buffer will keep accumulating data, and the aggregate result will be an overall aggregate (not from the current images).   But then we have a second issue, if we .reset() then we loose the ability to find  the total mean dice at the end of the loop.  

----
3) The hidden buffers.  with each call to DiceMetric() instance, we keep adding intermediate results (BatchxChannels) to a buffer (usually on GPU) which growth with each call.  
And we  return the cloned buffers 

https://github.com/Project-MONAI/MONAI/blob/c27ab360212ad122f5ca41fc69b8c2f47676b809/monai/metrics/metric.py#L269

So we're growing gpu memory unnecessarily, and cloning at each aggregate call. 

And in DDP (on multi gpu) it's even slower
https://github.com/Project-MONAI/MONAI/blob/c27ab360212ad122f5ca41fc69b8c2f47676b809/monai/metrics/metric.py#L253

we gather all buffers from all gpus,  this requires Allocating new gpu memory (the same size as buffers) to synchronize.
The syncronization itself is using our own function evenly_divisible_all_gather, which adds further overhead. 

[...]

PR #5379, opened 2022-10-22
Dicemetric updates draft

myron:
Draft PR for DiceMetric https://github.com/Project-MONAI/MONAI/issues/5378

1) add a simple option, so that the "_call_(): returns a simple mean dice value without using any buffers  and properly reduced
```
metric = DiceMetric(simple=True, reduction='mean')
for i in range(1000):
  dice = metric(x,y)
```
this can be called many times in loop without buffer overhead, and return an intuitive result

2) Alternatively we can add  DiceMetricSimple() class, if we don't want to create a convoluted logic by further extending DiceMetric. (it's also in this PR)

3) adding softmax, sigmoid options to compute_meandice() to  mirror the DiceLoss implementation, and remove the need for additional user-coding preprocessing

### DIFF
--- a/monai/metrics/meandice.py
+++ b/monai/metrics/meandice.py
@@ -13,7 +13,9 @@ from typing import Union
 
 import torch
 
+from monai.config import TensorOrList
 from monai.metrics.utils import do_metric_reduction, ignore_background, is_binary_tensor
+from monai.networks.utils import one_hot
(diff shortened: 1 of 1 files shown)


wyli:
mostly addressed by https://github.com/Project-MONAI/MONAI/commit/5ce8a10efb7e00637211a590d9b0efecff70c41d

PR #5380, opened 2022-10-23
AverageMeter utility 

myron:
Adds a a utility class AverageMeter to keep track of average accumulating values. For example during training/validation loop, we need to accumulate the per-batch metrics and calculate the final average value for the whole dataset. When training in multi-gpu environment, with DistributedDataParallel, it will average across the multi-gpu processes.

```
from monai.metrics import AverageMeter

run_avg = AverageMeter()
for i in range(len(train_set)):
    ...
    val = calc_metric(x,y) #some metric value
    run_avg.update(val)

val_avg = run_avg.get_avg() #average value

```

it supports input in various formats as constant, lists, arrays

   ```
  # a simple constant tracking
  avg_meter= AverageMeter()
  avg_meter.update(0.6)
  avg_meter.update(0.8)
  print(avg_meter.get_avg()) #prints 0.7

  # an array tracking, e.g. metrics from 3 classes
  avg_meter= AverageMeter()
  avg_meter.update([0.2, 0.4, 0.4])
  avg_meter.update([0.4, 0.6, 0.4])
  print(avg_meter.get_avg()) #prints [0.3, 0.5. 0.4]

  # different contributions / counts
  avg_meter= AverageMeter()
  avg_meter.update(1, count=4) #avg metric 1 coming from a batch of 4
  avg_meter.update(2, count=6) #avg metric 2 coming from a batch of 6
  print(avg_meter.get_avg()) #prints 1.6 == (1*4 +2*6)/(4+6)

  # different contributions / counts
  avg_meter= AverageMeter()
  avg_meter.update([0.5, 0.5, 0], count=[1, 1, 0]) # last elements count is zero to ignore it
  avg_meter.update([0.5, 0.5, 0.5], count=[1, 1, 1]) #
  print(avg_meter.get_avg()) #prints [0.5, 0.5, 0,5] == ([0.5, 0.5, 0] + [0.5, 0.5, 0.5]) / ([1, 1, 0] + [1, 1, 1])

```

This class is of general purpose.  Some monai metrics (e.g. DiceMetric) support reduction across gpu (by being a subclass of Cumulative that maintains gpu buffers for all data points).  But there are multiple issues with such approach  https://github.com/Project-MONAI/MONAI/issues/5378 , and for any new custom metric (that is not in monai) a user will have to write their own multi-gpu reduction or try to sublclass from Cumulative, which is not straightforward. 

This class adds a utility function for arbitrary custom metric to compute proper average (on one or more gpus/processes).  And for existing monai metrics, it could help (in the future) to disentangle the logic of metric computation from the multi-gpu reduction handling.

### Types of changes

### DIFF
--- a/monai/metrics/average_meter.py
+++ b/monai/metrics/average_meter.py
@@ -151,9 +151,10 @@ class AverageMeter:
         if torch.any(nfin):
             # non-finite numbers may indicate some errors in the user code
             print("non-finite numbers received", val, count)
-            val = torch.where(nfin, 0, val)
-            count = torch.where(nfin, 0, count)
-            val_count = torch.where(nfin, 0, val_count)
+            zero = torch.tensor(0).to(val)
+            val = torch.where(nfin, zero, val)
+            count = torch.where(nfin, zero, count)
+            val_count = torch.where(nfin, zero, val_count)
 
         self.val = val
         self.count += count
--- a/monai/metrics/average_meter.py
+++ b/monai/metrics/average_meter.py
@@ -1,0 +1,160 @@
+# Copyright (c) MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+import torch.distributed as dist
+
+__all__ = ["AverageMeter"]
+
+
+class AverageMeter:
+    """
+    A utility class to keep track of average values. For example during training/validation loop,
+    we need to accumulate the per-batch metrics and calculate the final average value for the whole dataset.
+    When training in multi-gpu environment, with DistributedDataParallel, it will average across the processes.
+
+    Example:
+
+    .. code-block:: python
+
+        from monai.metrics import AverageMeter
+
+        run_avg = AverageMeter()
+        for i in range(len(train_set)):
+            ...
+            val = calc_metric(x,y) #some metric value
+            run_avg.update(val)
+
+        val_avg = run_avg.get_avg() #average value
+
+    """
+
+    def __init__(self, ensure_cpu_group: bool = False) -> None:
+        """
+        Args:
+            ensure_cpu_group: applies only to multi-gpu distributed setup. Optionally request to operate on CPU,
+                even if the default DistributedDataParallel group is on GPU. This will create an additional
+                CPU group with GLOO backend to reduce across processes on CPU (and maintain all intermediate data on CPU).
+                Defaults to False: use the default process group created during DistributedDataParallel initialization.
+        """
+
+        self.is_distributed = dist.is_available() and dist.is_initialized()
+        self.group = None
+        self.device = None
+
+        if self.is_distributed:
+            dist_cuda = dist.get_backend() == dist.Backend.NCCL
+            if ensure_cpu_group and dist_cuda:
+                self.group = dist.new_group(backend="gloo")  # create a new cpu group
+            elif dist_cuda:
+                self.device = torch.device("cuda", dist.get_rank())
+
+        self.val = 0
+        self.sum = 0
+        self.count = 0
+
+    def proper_tensor(self, x) -> torch.Tensor:
+        """
+        Ensure torch.Tensor float format and optionally copy to the proper gpu device
+        """
+        x = x.detach() if isinstance(x, torch.Tensor) else torch.tensor(x)
+
+        if self.device is not None:
+            x = x.to(device=self.device)
+        else:
+            x = x.cpu()
+
+        return x.float()
+
+    def reduce(self, x: torch.Tensor, avg: bool = False) -> torch.Tensor:
+        """
+        Reduce across processes if in DDP
+        """
+        if self.is_distributed:
+            if avg:
+                x = x / dist.get_world_size()
+            else:
+                x = x.clone()
+            dist.all_reduce(x, group=self.group)
+
+        return x
+
+    def get_current(self, np: bool = True):
+        """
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Nic-Ma** [comment]
Hi @myron ,

We have something similar in MONAI which supports multi-gpu:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/metrics/cumulative_average.py

Thanks.

**Comment by myron** [comment]
> Hi @myron ,
> 
> We have something similar in MONAI which supports multi-gpu: https://github.com/Project-MONAI/MONAI/blob/dev/monai/metrics/cumulative_average.py Could you please have a try?
> 
> Thanks.

So it seems that CumulativeAverage() class was actually added due to my older request https://github.com/Project-MONAI/MONAI/issues/3233 here. But the way it was implemented by sub-classing from Cumulative()) creates unnecessary overhead (extra gpu buffers, extra memory cloning, extra allocations and synchronizations in evenly_divisible_all_gather). 
So we can add this extra class AverageMeter, or I can modify CumulativeAverage() (basically replace CumulativeAverage with this PR's AverageMeter).  I don't think anyone is using CumulativeAverge and no other classes subclass from it, so it should be acceptable). Would you like me to replace CumulativeAverage ?

**Comment by myron** [comment]
a new PR that modifies the CumulativeAverage class instead of adding AverageMeter : https://github.com/Project-MONAI/MONAI/pull/5383

@Nic-Ma , if you think it's better let's merge that PR (instead of this one)

thank you

**Comment by Nic-Ma** [comment]
> a new PR that modifies the CumulativeAverage class instead of adding AverageMeter : #5383
> 
> @Nic-Ma , if you think it's better let's merge that PR (instead of this one)
> 
> thank you

Thanks for your update, CC @ericspod @wyli to help discuss and review.
I will also take a look when I completed the SharedCacheDataset testing.

Thanks.

**Comment by Nic-Ma** [comment]
Hi @myron ,

BTW, I still don't quite understand the issue you said in the current implementation: "unnecessary overhead (extra gpu buffers, extra memory cloning, extra allocations and synchronizations in evenly_divisible_all_gather)"
Could you please help share more information? speed too slow? memory too big? or missing some functionality?

Thanks in advance.

**Comment by myron** [comment]
> Hi @myron ,
> 
> 
> 
> BTW, I still don't quite understand the issue you said in the current implementation: "unnecessary overhead (extra gpu buffers, extra memory cloning, extra allocations and synchronizations in evenly_divisible_all_gather)"
> 
> Could you please help share more information? speed too slow? memory too big? or missing some functionality?
> 

I'm referring to CumulativeAverage(), since you mentioned it. It's current implementation is not efficient.

a = CumulativeAverage()
- If you call it a.append(val) in the loop it will grow the Gpu memory, with each call. If the loop is long enough it will probably crash Gpu, and if it doesn't , it will take Gpu memory unnecessarily. A workaround is to call .aggregate() after each.append() call to clear buffers but that creates multi-Gpu extra syncs with all_gather

- on .aggregate() it clones all Gpu buffers. Since the size of buffer could be large, this is an additional overhead 

- in multigpu there are overheads due to all_gather and evenly_divisible_all_gather. In that function we need to allocate new memory (same size as buffer, which could be large) and gather all buffers (not simple averages). Furthermore it does another memory allocation and synchronization to account for different buffer length.  All of that is unnecessary if instead we simply do "all_reduce" on an average tensor (we don't need all the history buffers to keep track of an average value)

- And the missing functionality. It's often necessary to provide the count number (if metric mean value came from a batch of 8 or 16 or 10) to properly calculate the total average.

**Comment by Nic-Ma** [comment]
Thanks for the detailed explanation.
CC @wyli @ericspod also for more analysis.

Thanks.

**Comment by myron** [comment]
> Thanks for the detailed explanation. CC @wyli @ericspod also for more analysis.
> 
> Thanks.

you're welcome. So perhaps we can merge this other PR https://github.com/Project-MONAI/MONAI/pull/5383
that updates CumulativeAverage  (in that case we don't need this AverageMeter PR)

File: monai/metrics/average_meter.py [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+        if self.device is not None:
+            x = x.to(device=self.device)
+        else:
+            x = x.cpu()
+
+        return x.float()
+
+    def reduce(self, x: torch.Tensor, avg: bool = False) -> torch.Tensor:
+        """
+        Reduce across processes if in DDP
```
ericspod, line 79: We should document what the arguments mean in each method.

monai/metrics/average_meter.py
```diff
@@ -0,0 +1,161 @@
...
+    def get_current(self, np: bool = True):
+        """
+        return most recent value (averaged across processes)
+        """
+        x = self.reduce(self.val, avg=True)
+        if np:
+            x = x.cpu().numpy()
+        return x
+
+    def get_avg(self, np: bool = True):
```
ericspod, line 99: Maybe `to_numpy` instead of `np` to be more descriptive in addition to comments.

monai/metrics/average_meter.py
```diff
@@ -0,0 +1,161 @@
...
+        self.group = None
+        self.device = None
+
+        if self.is_distributed:
+            dist_cuda = dist.get_backend() == dist.Backend.NCCL
+            if ensure_cpu_group and dist_cuda:
+                self.group = dist.new_group(backend="gloo")  # create a new cpu group
+            elif dist_cuda:
+                self.device = torch.device("cuda", dist.get_rank())
+
```
ericspod, line 59: if `device` is `None` at this point it could be set to `torch.device("cpu")` since code elsewhere uses  `device == None` to check whether to move to CPU.

ericspod:
This looks good, as discussed with doing evaluation with odd-sized batches we could possibly put here some mechanism for `reduce` where not every rank is going to respond.

myron:
@ericspod Thank you very much for review.  I think we want to use another PR https://github.com/Project-MONAI/MONAI/pull/5383
instead of this one (that is to modify CumulativeAverage class insdead of adding a new class). Can you please comment and review that other PR.  It  is essentially the same code, but replaces CumulativeAverage

I will update it according to your comments. Thank you

ericspod:
@myron Will do, there were a few things I mentioned here that are relevant there as well.

myron:
Closing in favor of https://github.com/Project-MONAI/MONAI/pull/5383

PR #5382, opened 2022-10-23
5381 and deprecate `compute_meandice` `compute_meaniou`

### LINKED ISSUE #5381: `compute_meandice` and `compute_meaniou` should be `compute_dice` and `compute_iou`
**Describe the bug**
the basic logic in
https://github.com/Project-MONAI/MONAI/blob/bc09a4ce642b7ae32d0264e9025938b416fd56a3/monai/metrics/meandice.py#L106-L108

https://github.com/Project-MONAI/MONAI/blob/bc09a4ce642b7ae32d0264e9025938b416fd56a3/monai/metrics/meaniou.py#L106-L108

don't include any metrics reduction logic, so their names shouldn't contain 'mean'.

wyli:

Fixes #5381

### Description
- non-breaking changes to rename `compute_meandice` to `compute_dice`
- non-breaking changes to rename `compute_meaniou` to `compute_iou`

### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 * Automatically infer device from the first item in random elastic deformation dict
 * Add channel dim in `ComputeHoVerMaps` and `ComputeHoVerMapsd`
 * Remove batch dim in `SobelGradients` and `SobelGradientsd`
+### Deprecated
+* Deprecating `compute_meandice`, `compute_meaniou` in `monai.metrics`, in favor of
+`compute_dice` and `compute_iou` respectively.
 
 ## [1.0.0] - 2022-09-16
 ### Added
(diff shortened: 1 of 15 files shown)


ericspod:
If we're looking at `__all__` we should consider changing `__init__.py` files to  use `from whatever import *` rather than restating the same list of definitions we have in `__all__`. We've assumed that modules defining code include a `__all__` value stating what should be public, so `__init__.py` files at the same level don't need to work any of that out so don't need `__all__` themselves and can just import with `*`. Eg. `transforms/__init__.py` I think can be changed to replace the list of definitions with `*`.

mingxin-zheng:
Looks good to me.
`compute_meandice` was used across dints, unetr, segresnet, segresnet2d in Auto3Dseg templates. I guess we need to update them after the merge. @dongyang0122 @tangy5 @KumoLiu

mingxin-zheng (approved):

wyli:
> Looks good to me. `compute_meandice` was used across dints, unetr, segresnet, segresnet2d in Auto3Dseg templates. I guess we need to update them after the merge. @dongyang0122 @tangy5 @KumoLiu

thanks Mingxin, this patch will be included in v1.0.1, and agree that we should update the templates. It's also fine if we don't have time to update the template hash value for v1.0.1, because by default the deprecation warning will be printed only on the first occurrence within each run.

PR #5383, opened 2022-10-23
CumulativeAverage refactor

myron:
This refactors CumulativeAverage() by removing subclassing from Cumulative() to have much less overhead, and adds additionan functionality to specify count. 
(as discussed here https://github.com/Project-MONAI/MONAI/pull/5380)

CumulativeAverage previously was not used almost anywhere, except for MIL tutorial. Name of class methods .aggregate() and .append() are preserved, so it should work the same way as before (non-breaking).

New tests are added (especially to test in multi-gpu DDP)

### Types of changes

### DIFF
--- a/monai/metrics/cumulative_average.py
+++ b/monai/metrics/cumulative_average.py
@@ -10,63 +10,161 @@
 # limitations under the License.
 
 import torch
-
-from monai.transforms import isnan
-from monai.utils import convert_data_type
-
-from .metric import Cumulative
+import torch.distributed as dist
 
 
-class CumulativeAverage(Cumulative):
+class CumulativeAverage:
     """
-    Cumulatively record data value and aggregate for the average value.
-    It supports single class or multi-class data, for example,
-    value can be 0.44 (a loss value) or [0.3, 0.4] (metrics of two classes).
-    It also supports distributed data parallel, sync data when aggregating.
-    For example, recording loss values and compute the overall average value in every 5 iterations:
+    A utility class to keep track of average values. For example during training/validation loop,
+    we need to accumulate the per-batch metrics and calculate the final average value for the whole dataset.
+    When training in multi-gpu environment, with DistributedDataParallel, it will average across the processes.
+
+    Example:
 
     .. code-block:: python
 
-        average = CumulativeAverage()
-        for i, d in enumerate(dataloader):
-            loss = ...
-            average.append(loss)
-            if i % 5 == 0:
-                print(f"cumulative average of loss: {average.aggregate()}")
-        average.reset()
+        from monai.metrics import CumulativeAverage
+
+        run_avg = CumulativeAverage()
+        for i in range(len(train_set)):
+            ...
+            val = calc_metric(x,y) #some metric value
+            run_avg.append(val)
+
+        val_avg = run_avg.aggregate() #average value
 
     """
 
-    def __init__(self) -> None:
-        super().__init__()
-        self.sum = None
-        self.not_nans = None
+    def __init__(self, ensure_cpu_group: bool = False) -> None:
+        """
+        Args:
+            ensure_cpu_group: applies only to multi-gpu distributed setup. Optionally request to operate on CPU,
+                even if the default DistributedDataParallel group is on GPU. This will create an additional
+                CPU group with GLOO backend to reduce across processes on CPU (and maintain all intermediate data on CPU).
+                Defaults to False: use the default process group created during DistributedDataParallel initialization.
+        """
+
+        self.is_distributed = dist.is_available() and dist.is_initialized()
+        self.group = None
+        self.device = None
+
+        if self.is_distributed:
+            dist_cuda = dist.get_backend() == dist.Backend.NCCL
+            if ensure_cpu_group and dist_cuda:
+                self.group = dist.new_group(backend="gloo")  # create a new cpu group
+            elif dist_cuda:
+                self.device = torch.device("cuda", dist.get_rank())
+
+        self.reset()
 
     def reset(self):
         """
-        Reset all the running status, including buffers, sum, not nans count, etc.
+        Reset all the running status
 
         """
-        super().reset()
-        self.sum = None
-        self.not_nans = None
+        self.val: torch.Tensor = 0
+        self.sum: torch.Tensor = 0
+        self.count: torch.Tensor = 0
 
-    def aggregate(self):
+    def proper_tensor(self, x) -> torch.Tensor:
         """
-        Sync data from all the ranks and compute the average value with previous sum value.
+        Ensure torch.Tensor float format and optionally copy to the proper gpu device
+        """
+        if isinstance(x, torch.Tensor):
+            x = x.detach()
+        else:
+            x = torch.tensor(x)
+
+        if self.device is not None:
+            x = x.to(device=self.device, dtype=torch.float)
+        else:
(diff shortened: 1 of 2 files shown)


myron:
seems it passed all the tests  @Nic-Ma

myron:
@wyli @Nic-Ma  any objections to merging it? (this class was originally created due to my request, and this PR improves it)

monai/metrics/cumulative_average.py
```diff
@@ -9,64 +9,162 @@
...
+        """
+
+        self.is_distributed = dist.is_available() and dist.is_initialized()
+        self.group = None
+        self.device = None
+
+        if self.is_distributed:
+            dist_cuda = dist.get_backend() == dist.Backend.NCCL
+            if ensure_cpu_group and dist_cuda:
+                self.group = dist.new_group(backend="gloo")  # create a new cpu group
```
wyli, line 57: no sure if this class should manage the process group, this should be decoupled. also `new_group` takes additional parameters, it's not easy to make a suitable assumption about the system environment and have a robust implementation within this class

myron: @wyli good point, thank you. I've updated the code,  to have the process group as an option (if a user wants to provide a custom group).

myron:
@ericspod  I've updated it according to your comments in https://github.com/Project-MONAI/MONAI/pull/5380
Can you please check

monai/metrics/cumulative_average.py
```diff
@@ -9,64 +9,160 @@
...
+        self.is_distributed = dist.is_available() and dist.is_initialized()
+        self.device = None
+
+        if self.is_distributed:
+            self.group = group
+            if dist.get_backend(group=group) == dist.Backend.NCCL:
+                self.device = torch.device("cuda", dist.get_rank(group=group))
+        else:
+            self.group = None
+
```
ericspod, line 58:
```suggestion
if self.device is None:
    self.device=torch.device("cpu")
```

monai/metrics/cumulative_average.py
```diff
@@ -9,64 +9,160 @@
...
         """
-        Reset all the running status, including buffers, sum, not nans count, etc.
+        Ensure torch.Tensor float format and optionally copy to the proper gpu device
+        """
+        x: torch.Tensor = a.detach() if isinstance(a, torch.Tensor) else torch.tensor(a)
+
+        if self.device is not None:
+            x = x.to(device=self.device, dtype=torch.float)
+        else:
+            x = x.cpu().float()
```
ericspod, line 79: If `self.device` isn't ever `None`:
```suggestion
        x = x.to(device=self.device, dtype=torch.float)
```

monai/metrics/cumulative_average.py
```diff
@@ -9,64 +9,170 @@
...
 
-        return self.sum / self.not_nans
+        val = self.proper_tensor(val)
+        count = self.proper_tensor(count)
+
+        val_count = val * count
+        nfin = ~torch.isfinite(val_count)
+        if torch.any(nfin):
+            # non-finite numbers may indicate some errors in the user code
+            print("non-finite numbers received", val, count)
```
ericspod, line 170: We shouldn't have print statements in regular code since at times we would want programs to be silent. Instead if it's important to state this we should write to the log or issue a warning with `warnings.warn`.

monai/metrics/cumulative_average.py
myron: updated

monai/metrics/cumulative_average.py
myron: updated

monai/metrics/cumulative_average.py
myron: okay, I 've changed it to warnings. thank you

myron:
@ericspod thank you for  the quick review. I've updated the code according to your comments. plz check again.

monai/metrics/cumulative_average.py
```diff
@@ -9,64 +9,168 @@
...
+    def reset(self) -> None:
         """
-        Reset all the running status, including buffers, sum, not nans count, etc.
+        Reset all the running status
 
         """
-        super().reset()
-        self.sum = None
-        self.not_nans = None
+        self.val: torch.Tensor = 0  # type: ignore
```
ericspod, line 70: It's slightly inefficient but using actual tensors here instead of 0 would be more consistent and not need the type ignore comment, but it's not critical.

ericspod (approved):

ericspod:
I'm good with this if others are?

monai/metrics/cumulative_average.py
myron: yes, I tried it, using actual torch tensor 0, but we don't know the shape of val, sum, count before stating to accumulate, and later self.sum += some_tensor fails, since in place increment not possible with different tensor shapes (even if incrementing 0 dim tensor).  So we'd have to use self.sum = self.sum + some_tensor, out of place increment, which is slightly bigger overhead. It's seems simpler to have it as is.

monai/metrics/cumulative_average.py
```diff
@@ -9,64 +9,168 @@
...
+        self.sum: torch.Tensor = 0  # type: ignore
+        self.count: torch.Tensor = 0  # type: ignore
 
-    def aggregate(self):
+    def proper_tensor(self, a: Any) -> torch.Tensor:
         """
-        Sync data from all the ranks and compute the average value with previous sum value.
+        Ensure torch.Tensor float format and optionally copy to the proper gpu device
+        """
+        x: torch.Tensor = a.detach() if isinstance(a, torch.Tensor) else torch.tensor(a)
```
wyli, line 78: Normally the metrics module shouldn't change the device of the input if we follow pytorch's api style. The user should prepare the input beforehand accordingly in a compatible shape/type/device... If this is 100% needed, we may have to change all the relevant APIs just to me consistent

myron: this line is just to ensure the the avergage value is on the correct device ( CPU vs GPU).  If a user provide it as a simple constant or a list (on cpu and we're in DDP on gpu), it'll be converted to proper GPU device. And otherwise if we're not in DDP, it will ensure we're maintaining average on CPU.  

many pytorch api accepts input as either a constant or a tensor, and support shape broadcasting. In either case this  self.proper_tensor()  is just an internal sub-function.

wyli: sorry, the comment is meant to be on the the `self.device` usage in general in this class

myron: @wyli thank you for your comment. the self.device is set according to the processing group. If the user specified group on CPU, then device is on CPU, if the user specified it on GPU (or it's default group on GPU/ NCCL) then the device is GPU.  We need to insure the proper device to be able to run multiprocessing reduction on proper Processing group in DDP.  Please let me know if there is something specific you'd like me to change in the code.  thank you

myron: @wyli , should I mark this conversation as resolved? thank you

wyli: Thanks, I don't think this class should assign the devices, it doesn't even consider the difference between rank vs local rank.

myron: @wyli alright, I've removed self.device , it will be up to a user now to use the correct device.  Let me know if you have any other objections.

PR #5388, opened 2022-10-24
5385-enhance-mlflow-handler

### LINKED ISSUE #5385: Enhance MLFlowHandler

MLFlowHandler is being used to log DL experiments with PyTorch Ignite for MONAI bundles. We can further enhance the module by adding these four features:
 P0:
- API for users to add experiment/run name in MLFlow
- API for users to log customized params for each run
- Methods to log json file content
- Methods to log result images
- Methods to log optimizer params 
- Make sure this handler works in multi-gpu & multi-node environment
- Make sure this handler works in all existed bundles

P1:
- Methods to log dataset
- Enhance the `tracking_uri` parameter for convenient modification
- 

Use `ignite.contrib.handlers.mlflow_logger`

https://pytorch.org/ignite/generated/ignite.contrib.handlers.mlflow_logger.html#module-ignite.contrib.handlers.mlflow_logger

binliunls:

Fixes #5385  .

### Description

This PR is about to enhance the mlflow handler in monai to track more details in the experiment. Here are a few enhancements 
that needs to be added through this PR.

- API for users to add experiment/run name in MLFlow
- API for users to log customized params for each run
- Methods to log result images
- Methods to log optimizer params
- (optional) additional metric_names as a user argument to override the default engine.state.metrics to instruct MLFlow about metrics to log

After adding these enhancements, some tests listed below should be excuted.
- Make sure this handler works in multi-gpu environment
- Make sure this handler works in all existed bundles

### Types of changes

- [x] API for users to add experiment/run name in MLFlow
- [x] API for users to log customized params for each run
- [x] Methods to log result images
- [x] Methods to log optimizer params
- [x] Make sure this handler works in multi-gpu environment
- [ ] Make sure this handler works in all existed bundles


monai/handlers/mlflow_handler.py
```diff
@@ -86,6 +93,11 @@ def __init__(
         global_epoch_transform: Callable = lambda x: x,
         state_attributes: Optional[Sequence[str]] = None,
         tag_name: str = DEFAULT_TAG,
+        experiment_name: str = "default_experiment",
+        run_name: str = "test_run",
```
mingxin-zheng, line 97: How about automatically generating the `run_name` by datetime/model name, if user did not specify it in the `__init__`? For example, the user doesn't need to provide inputs but they can finally see their experiments in different `run_name`, instead of "test_run" in the console.

monai/handlers/mlflow_handler.py
```diff
@@ -86,6 +93,11 @@ def __init__(
...
+        experiment_param: Optional[Dict] = None,
+        artifacts: Optional[Sequence[str]] = None,
```
mingxin-zheng, line 99: Maybe we can use `pathlib.Path` and `PathLike` for type hints.

monai/handlers/mlflow_handler.py
```diff
@@ -70,6 +70,13 @@ class MLFlowHandler:
         state_attributes: expected attributes from `engine.state`, if provided, will extract them
             when epoch completed.
         tag_name: when iteration output is a scalar, `tag_name` is used to track, defaults to `'Loss'`.
+        experiment_name: name for an experiment, defaults to `default_experiment`.
+        run_name: name for run in an experiment, defaults to `test_run`.
+        experiment_param: a dict recording parameters which will not change through whole experiment,
+            like torch version, cuda version and so on.
+        artifacts: paths to images that need to be recorded after a whole run.
```
mingxin-zheng, line 77: The experiment management tutorial needs to cover how these artifacts are generated because the usage is quite unclear.

monai/handlers/mlflow_handler.py
```diff
@@ -196,4 +260,18 @@ def _default_iteration_log(self, engine: Engine) -> None:
         if not isinstance(loss, dict):
             loss = {self.tag_name: loss.item() if isinstance(loss, torch.Tensor) else loss}
 
-        mlflow.log_metrics(loss, step=engine.state.iteration)
+        if isinstance(engine, Trainer):
```
mingxin-zheng, line 263: Would that be a problem if the trainer is not from `monai.engine.Trainer`? For example, cell 7 in the [segmentation ignite notebook](https://github.com/Project-MONAI/tutorials/blob/main/3d_segmentation/unet_segmentation_3d_ignite.ipynb)

```python
trainer = ignite.engine.create_supervised_trainer(
    net, opt, loss, device, False
)
```

binliunls: Hi @mingxin-zheng ,
     Currently there will be a problem if the input engine is not a Trainer in monai engine. I will add some code here to ensure the engine type to be Trainer.
Thanks,
Bin

monai/handlers/mlflow_handler.py
```diff
@@ -113,14 +158,56 @@ def attach(self, engine: Engine) -> None:
...
         """
+        mlflow.set_experiment(self.experiment_name)
         if mlflow.active_run() is None:
-            mlflow.start_run()
+            if self.run_name:
+                mlflow.start_run(run_name=self.run_name)
+            else:
+                cur_time = time.strftime("%Y%m%d_%H%M%S")
+                run_name = "run_" + cur_time
+                mlflow.start_run(run_name=run_name)
```
Nic-Ma, line 176: I would suggest to simplify the logic to:
```py
run_name = f"run_{time.strftime("%Y%m%d_%H%M%S")}" if self.run_name is None else self.run_name
mlflow.start_run(run_name=run_name)
```

monai/handlers/mlflow_handler.py
```diff
@@ -174,6 +261,16 @@ def _default_epoch_log(self, engine: Engine) -> None:
         current_epoch = self.global_epoch_transform(engine.state.epoch)
         mlflow.log_metrics(log_dict, step=current_epoch)
 
+        events = engine._event_handlers
+
+        for e in events:
+            for handler, _, _ in engine._event_handlers[e]:
+                if isinstance(handler, ValidationHandler):
+                    evaluator_state = getattr(handler.validator, "state", None)
```
Nic-Ma, line 269: We usually attach handlers to `trainer` or `evaluator` separately to track `loss` and `val metrics` during training:
https://github.com/Project-MONAI/tutorials/blob/main/modules/engines/unet_training_dict.py#L125
Here the logic is to get `evaluator` from the `trainer` handler, I feel it's different from our `Statshandler` and `TensorBoardHandler` behavior. But seems `ignite` example has some logic similar in theory?
https://github.com/pytorch/ignite/blob/master/examples/mnist/mnist_with_tensorboard.py#L116
@wyli @vfdev-5 What do you think?

Thanks in advance.

monai/handlers/mlflow_handler.py
Nic-Ma: Hi @mingxin-zheng ,

I think `MLFlow` only supports local image files with provided path. So maybe it's OK to save the prediction in files then visualize here? The question I have is how to distinguish the images of different experiments if we always saving the prediction to the same path to override previous results?

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -70,6 +75,13 @@ class MLFlowHandler:
         state_attributes: expected attributes from `engine.state`, if provided, will extract them
             when epoch completed.
         tag_name: when iteration output is a scalar, `tag_name` is used to track, defaults to `'Loss'`.
+        experiment_name: name for an experiment, defaults to `default_experiment`.
+        run_name: name for run in an experiment, defaults to `test_run`.
+        experiment_param: a dict recording parameters which will not change through whole experiment,
+            like torch version, cuda version and so on.
+        artifacts: paths to images that need to be recorded after a whole run.
+        optimizer_param_names: parameters' name in optimizer that need to be record during runing,
+            defaults to `["lr"]`.
```
Nic-Ma, line 84: I think we need a more general way to track `optimizer` parameters, especially when there are multiple optimizers and we may also need to track `lr_scheduler`.
@wyli @ericspod Do you experts have any suggestions?

Thanks in advance.

Nic-Ma (commented):
Thanks for the update, put some comments inline.
Need @wyli @ericspod @vfdev-5 for more design discussion.

Thanks in advance.

Nic-Ma:
Please ensure we can call these `get_config_values()` and `get_optional_config_values()` functions to track the system env info in MLFlow:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/config/deviceconfig.py#L48-L81

Thanks.

monai/bundle/utils.py
```diff
@@ -108,16 +108,27 @@
         # MLFlowHandler config for the trainer
         "trainer": {
             "_target_": "MLFlowHandler",
+            "_disabled_": "$torch.distributed.get_rank() > 0 if torch.distributed.is_initialized() else False",
```
Nic-Ma, line 111: Please also check "dist.is_available()", refer to:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/dataset.py#L808
More details:
https://pytorch.org/docs/stable/distributed.html#torch.distributed.is_available

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -72,11 +76,21 @@ class MLFlowHandler:
...
 
     For more details of MLFlow usage, please refer to: https://mlflow.org/docs/latest/index.html.
 
     """
 
+    default_log_param_list = ["network", "device", "optimizer", "loss_function"]
+    default_attr_name = ["seed", "max_epochs", "epoch_length"]
```
Nic-Ma, line 92: The `engine.state.seed` is deprecated in ignite and MONAI, I think no need to track it.
CC @vfdev-5 to help confirm it.

Thanks in advance.

monai/bundle/utils.py
```diff
@@ -108,16 +108,27 @@
         # MLFlowHandler config for the trainer
```
Nic-Ma, line 108: Please add `experiment_name` and `run_name` here with default values (like the `tracking_uri`) and refer to them in the handlers, as these 2 vars are frequently override by end users.

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -115,14 +183,61 @@ def attach(self, engine: Engine) -> None:
...
 
         """
+        mlflow.set_experiment(self.experiment_name)
         if mlflow.active_run() is None:
-            mlflow.start_run()
+            run_name = f"run_{time.strftime('%Y%m%d_%H%M%S')}" if self.run_name is None else self.run_name
+            mlflow.start_run(run_name=run_name)
+
+        if self.experiment_param:
+            # avoid to recording same parameters in the same run
```
Nic-Ma, line 200: "recording" -> "record"

monai/handlers/mlflow_handler.py
```diff
@@ -115,14 +183,61 @@ def attach(self, engine: Engine) -> None:
...
+            self._delete_exist_param_in_dict(self.experiment_param)
```
Nic-Ma, line 201: I don't quite understand why we need to delete the existing values?
I think the values in `experiment_param` are const, right?
And the same engine object may `start -> end` several times during training.

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -115,14 +183,61 @@ def attach(self, engine: Engine) -> None:
...
+                        artifact_list.append(file_path)
+        return artifact_list
+
+    def complete(self) -> None:
+        """
+        Handler for train or validation/evaluation completed Event.
+        """
+        # If just input one path, convert it to list.
+        if isinstance(self.artifacts, str):
+            self.artifacts = [self.artifacts]
```
Nic-Ma, line 236: Please avoid to changing member variable during training, I would suggest to set it in the `__init__()`:
```py
self.artifacts = ensure_tuple(artifacts)
```

monai/handlers/mlflow_handler.py
```diff
@@ -198,4 +313,18 @@ def _default_iteration_log(self, engine: Engine) -> None:
         if not isinstance(loss, dict):
             loss = {self.tag_name: loss.item() if isinstance(loss, torch.Tensor) else loss}
 
-        mlflow.log_metrics(loss, step=engine.state.iteration)
+        if isinstance(engine, Trainer):
+            mlflow.log_metrics(loss, step=engine.state.iteration)
+
+        # If there is optimizer attr in engine, then record parameters specified in init function.
+        try:
+            cur_optimizer = engine.optimizer  # type: ignore
```
Nic-Ma, line 321: Please note that not all the MONAI Trainer object has `optimizer` var, let's don't use `try - except` to check it, maybe `hasattr()` is better.
@wyli @ericspod Could you python experts help share some ideas?

Thanks in advance.

monai/handlers/mlflow_handler.py
```diff
@@ -72,11 +76,21 @@ class MLFlowHandler:
...
+            like torch version, cuda version and so on.
+        artifacts: paths to images that need to be recorded after a whole run.
+        optimizer_param_names: parameters' name in optimizer that need to be record during runing,
+            defaults to `["lr"]`.
 
     For more details of MLFlow usage, please refer to: https://mlflow.org/docs/latest/index.html.
 
     """
 
+    default_log_param_list = ["network", "device", "optimizer", "loss_function"]
```
Nic-Ma, line 91: I don't get what we will log for these objects?

Thanks.

monai/bundle/utils.py
```diff
@@ -19,7 +19,7 @@
 
 yaml, _ = optional_import("yaml")
 
-__all__ = ["ID_REF_KEY", "ID_SEP_KEY", "EXPR_KEY", "MACRO_KEY"]
+__all__ = ["ID_REF_KEY", "ID_SEP_KEY", "EXPR_KEY", "MACRO_KEY", "DEFAULT_MLFLOW_SETTINGS"]
```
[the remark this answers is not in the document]
binliunls: Thanks, I will add this var.

monai/bundle/utils.py
binliunls: Thanks Nic, I will add this check in the initial config.

monai/handlers/mlflow_handler.py
binliunls: Hi @Nic-Ma 
It tries to log the name of these objects' type to make users check their basic workflow easily. For example, if UNet were used, it will log `network:<class 'monai.networks.nets.unet.UNet'>`.  As we will log all the meta info, maybe these infos are not necessary. What would you think?
Thanks,
Bin

monai/handlers/mlflow_handler.py
Nic-Ma: I think we don't suppose to attach the same handler object to more than 1 engine, do we still need to worry about this problem?

monai/handlers/mlflow_handler.py
Nic-Ma: I would suggest to remove this feature directly.

Thanks.

monai/handlers/mlflow_handler.py
binliunls: If it can be ensured that the `experiment_param` in trainer and validator won't have any same key, then we can remove this part.
Thanks,
Bin

binliunls:
Hi @Nic-Ma @wyli ,
      I encountered this problem when running the ci/cd test. However, I didn't modify any corresponding code. Could you please help me to check why this happened.

```
  File "/tmp/tmp.Ly6otvIpdJ/tests/test_scale_intensity_range_percentiles.py", line 60, in test_relative_scaling
    assert_allclose(
  File "/tmp/tmp.Ly6otvIpdJ/tests/utils.py", line 120, in assert_allclose
    np.testing.assert_allclose(actual, desired, *args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/site-packages/numpy/testing/_private/utils.py", line 1527, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/site-packages/numpy/testing/_private/utils.py", line 844, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=0.0001, atol=0
```
Thanks,
Bin

monai/bundle/utils.py
```diff
@@ -105,19 +105,43 @@
     "handlers_id": DEFAULT_HANDLERS_ID,
     "configs": {
         "tracking_uri": "$@output_dir + '/mlruns'",
+        "experiment_name": "default_experiment",
+        "run_name": "$None",
```
Nic-Ma, line 109: Please use `"run_name": None,` directly.

Thanks.

monai/bundle/utils.py
```diff
@@ -105,19 +105,43 @@
...
+        "not_rank0": (
```
Nic-Ma, line 110: I would suggest "is_not_rank0", follows the style:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/config_item.py#L258
What do you think?

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -88,6 +101,11 @@ def __init__(
         global_epoch_transform: Callable = lambda x: x,
         state_attributes: Optional[Sequence[str]] = None,
         tag_name: str = DEFAULT_TAG,
+        experiment_name: str = "default_experiment",
+        run_name: Optional[str] = None,
+        experiment_param: Optional[Dict] = None,
+        artifacts: Optional[Sequence[Path]] = None,
+        optimizer_param_names: Tuple[str, ...] = ("lr",),
```
Nic-Ma, line 108: Please use `ensure_tuple` to simplify the arg value, for example:
```py
optimizer_param_names: Union[str, Sequence[str]] = "lr"
```
Then:
```py
self.optimizer_param_names = ensure_tuple(optimizer_param_names)
```

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -88,6 +101,11 @@ def __init__(
         global_epoch_transform: Callable = lambda x: x,
         state_attributes: Optional[Sequence[str]] = None,
         tag_name: str = DEFAULT_TAG,
+        experiment_name: str = "default_experiment",
+        run_name: Optional[str] = None,
+        experiment_param: Optional[Dict] = None,
+        artifacts: Optional[Sequence[Path]] = None,
```
Nic-Ma, line 107: Same as the comments for `optimizer_param_names`.

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -198,4 +304,15 @@ def _default_iteration_log(self, engine: Engine) -> None:
         if not isinstance(loss, dict):
             loss = {self.tag_name: loss.item() if isinstance(loss, torch.Tensor) else loss}
 
-        mlflow.log_metrics(loss, step=engine.state.iteration)
+        if isinstance(engine, Trainer):
```
Nic-Ma, line 307: I don't get why we need to check whether the engine is `Trainer` here?

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -100,6 +118,55 @@ def __init__(
...
         self.state_attributes = state_attributes
         self.tag_name = tag_name
+        self.experiment_name = experiment_name
+        self.run_name = run_name
+        self.experiment_param = experiment_param
+        self.artifacts = ensure_tuple(artifacts)
+        self.optimizer_param_names = optimizer_param_names
+        self.client = mlflow.MlflowClient()
+
+    def _try_log_param(self, engine: Engine, attr: str) -> None:
```
Nic-Ma, line 128: I don't see where you use this function, please delete it.

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -115,14 +182,53 @@ def attach(self, engine: Engine) -> None:
...
+        mlflow.set_experiment(self.experiment_name)
         if mlflow.active_run() is None:
-            mlflow.start_run()
+            run_name = f"run_{time.strftime('%Y%m%d_%H%M%S')}" if self.run_name is None else self.run_name
+            mlflow.start_run(run_name=run_name)
+
+        if self.experiment_param:
+            mlflow.log_params(self.experiment_param)
+
+        attrs = {attr: getattr(engine.state, attr, None) for attr in self.default_attr_name}
```
Nic-Ma, line 201: Can we allow user to specify other attrs of `engine.state.XXX` to log except the default ones?

Thanks.

monai/handlers/mlflow_handler.py
binliunls: The check here is to avoid that users do transform on outputs of validator or evaluator which may influence the loss record.
Thanks,
Bin

Nic-Ma: I think we don't need to take care of this logic here, user should set `iteration_log=False` for validator or evaluator: https://github.com/Project-MONAI/MONAI/blob/dev/monai/handlers/mlflow_handler.py#L83

Thanks.

monai/handlers/mlflow_handler.py
binliunls: Yes, this line is for some default paramerters. Users can specified their own attrs throught the line below.
https://github.com/binliunls/MONAI/blob/cabc6cc56e78cf54c640ec3e22efa35eb6dc7da1/monai/handlers/mlflow_handler.py#L286
Thanks
Bin

Nic-Ma: I would suggest to put the `["max_epochs", "epoch_length"]` as default value of the `state_attributes` arg.
What do you think?

Thanks.

monai/bundle/utils.py
```diff
@@ -105,19 +105,43 @@
     "handlers_id": DEFAULT_HANDLERS_ID,
     "configs": {
         "tracking_uri": "$@output_dir + '/mlruns'",
+        "experiment_name": "default_experiment",
```
wyli, line 108:
```suggestion
        "experiment_name": "monai_experiment",
```
or default_monai_experiment (so that it's clear that it's from monai default API instead of mlflow's)

monai/bundle/utils.py
```diff
@@ -105,19 +105,43 @@
...
+        "run_name": "None",
+        "is_not_rank0": (
+            "$torch.distributed.get_rank() > 0 \
+                if (torch.distributed.is_available() and torch.distributed.is_initialized()) \
+                else False"
```
wyli, line 113:
```suggestion
            "$torch.distributed.is_available() and torch.distributed.is_initialized() and torch.distributed.get_rank() > 0"
```

monai/handlers/mlflow_handler.py
```diff
@@ -100,6 +117,42 @@ def __init__(
...
+        Check whether a parameter has already been logged in current mlflow run.
+
+        Args:
+            param_name: name of parameter to check in mlflow run.
+
+        Return:
+            If the given parameter already was logged to mlflow.
+        """
+        cur_run = mlflow.active_run()
+        log_data = self.client.get_run(cur_run.info.run_id).data
```
wyli, line 138: this active_run/get_run are called for every key in key_list, perhaps merge`_is_param_exists` into `_delete_exist_param_in_dict`

wyli (commented):
thanks, I put some minor comments inline

wyli (approved):
looks good to me, @Nic-Ma pls let me know if you still have comments otherwise I'll merge this later today

monai/bundle/utils.py
```diff
@@ -105,19 +105,42 @@
     "handlers_id": DEFAULT_HANDLERS_ID,
     "configs": {
         "tracking_uri": "$@output_dir + '/mlruns'",
+        "experiment_name": "monai_experiment",
+        "run_name": "None",
```
Nic-Ma, line 109: According to my previous comment, I think it should be `None` instead of string "None".

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -72,11 +75,20 @@ class MLFlowHandler:
         state_attributes: expected attributes from `engine.state`, if provided, will extract them
             when epoch completed.
         tag_name: when iteration output is a scalar, `tag_name` is used to track, defaults to `'Loss'`.
+        experiment_name: name for an experiment, defaults to `default_experiment`.
+        run_name: name for run in an experiment.
+        experiment_param: a dict recording parameters which will not change through whole experiment,
+            like torch version, cuda version and so on.
+        artifacts: paths to images that need to be recorded after a whole run.
+        optimizer_param_names: parameters' name in optimizer that need to be record during runing,
+            defaults to `["lr"]`.
```
Nic-Ma, line 84: defaults to `"lr"`.

monai/handlers/mlflow_handler.py
```diff
@@ -88,6 +100,11 @@ def __init__(
         global_epoch_transform: Callable = lambda x: x,
         state_attributes: Optional[Sequence[str]] = None,
```
Nic-Ma, line 101: I didn't see you change the default value of this arg to `["max_epochs", "epoch_length"]` as my previous comment. Do you agree with that?

Nic-Ma (changes requested):

Nic-Ma:
Could you please help also add the `experiment_name` and `run_name` in this doc-string example:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/scripts.py#L549

Thanks.

monai/bundle/scripts.py
```diff
@@ -547,17 +547,36 @@ def run(
                     },
                     "configs": {
                         "tracking_uri": "<path>",
+                        "experiment_name": "monai_experiment",
+                        "run_name": None,
+                        "is_not_rank0": (
```
Nic-Ma, line 552: Hi @wyli , I think we should simplify this example config here to make the doc-string clear, and we will also have a very detailed tutorial link soon.
My proposal example:
```py
tracking = {
    "handlers_id": {"evaluator": {"id": "evaluator", "handlers": "handlers"}},
    "configs": {
        "tracking_uri": "<path>",
        "run_name": "my_test_run",
        "evaluator": {
            "_target_": "MLFlowHandler",
            "tracking_uri": "@tracking_uri",
            "experiment_name": "@experiment_name",
            "run_name": "@run_name",
            "iteration_log": False,
        },
    },
}
```
I am not very sure, what do you think from users' point view?

Thanks.

monai/handlers/mlflow_handler.py
binliunls: Hi @Nic-Ma ,
     In my opinion, the `default_attr_name` should be some parameters that would not change through the whole experiment, therefore they should be recorded at the beginning of the training as parameters, as what i did in this code.  And the attributes specified in `state_attributes` are kind of epoch depending things that would expect to change every epoch. Thus, these parameters shoud be recorded every epoch as metrics. I would suggest we keep these two things split. What would you like?

Thanks,
Bin

monai/bundle/scripts.py
wyli: (it seems all the bundles have the same set of logging config:
https://github.com/Project-MONAI/model-zoo/blob/dev/models/brats_mri_segmentation/configs/logging.conf maybe we should consider improving the default file as well...)

monai/handlers/mlflow_handler.py
wyli: the values might be changed during FL  https://github.com/Project-MONAI/MONAI/blob/e2fc703ad9467eb2ffe28e0a5c51d252e22cef64/monai/fl/client/monai_algo.py#L495-L496 so I agree with @Nic-Ma

monai/bundle/scripts.py
Nic-Ma: okok, then I am OK with the change.

Thanks.

monai/handlers/mlflow_handler.py
Nic-Ma: Hi @wyli ,

The values might be changed during FL but this PR records the value at "START" event of every FL round, I think it's OK? I think maybe we just need to change the var name to `default_tracking_params` to resolve the confusion?
What do you think?

Thanks.

Nic-Ma (approved):

PR #5398, opened 2022-10-25
5390 Simplify cache content to only list type

### LINKED ISSUE #5390: Remove hashing by key from CacheDatast  (hash_as_key)
In the CacheDataset we have an option to hash by key (hash_as_key with hash_func=pickle_hashing).  It looks like , it is not used anywhere in the code or in any tutorials. 

Historically, the option was added due to @dongyang0122 request
 https://github.com/Project-MONAI/MONAI/issues/3734
 https://github.com/Project-MONAI/MONAI/pull/3739
 I've just synced with @dongyang0122 and he is not using this option either (and never used it)   

Hashing by key also seems not practical,  computing hash on 3D image inquires overhead (and it needs to be done every time when trying to access hash , both for read/write).  Finally other Datasets sub-classes don't have this option at all, which makes it inconsistent.   

I don't really see a good use-case to keep it. If there are duplicate images in the dataset , and they have to be exactly identical to a voxel for it to work,  then it will be probably faster to just process a dubplicate image, then checking for hash for all images.  Or even better to run some check for dublicates beforehand. 

If we remove it completely,  we'll simplify quite a bit our code in CacheDataset.  There is already a refactoring PR https://github.com/Project-MONAI/MONAI/pull/5365, perhaps we can remove this hashing by key too @Nic-Ma

Nic-Ma:
Fixes #5390 .

### Description

This PR simplified the `hash key` caching mode in the `CacheDataset` to use `list` as cache type, same as regular cache.

### Types of changes


myron:
@Nic-Ma , I do appreciate you putting time into creating this PR so quickly , but this does not address https://github.com/Project-MONAI/MONAI/issues/5390 . That issue proposes to remove caching by key completely, since it is not necessary and there is no use cases to keep it.  

we should probably first agree if we're removing caching_by_key completely or not first.

monai/data/dataset.py
```diff
@@ -810,12 +811,11 @@ def _compute_cache():
             # only compute cache for the unique items of dataset
             mapping = {self.hash_func(v): v for v in data}
             self.data = list(mapping.values())
```
myron, line 813: So here self.data becomes of shorter length (with duplicates removed).  Then on line 817 we call _compute_cache() which starts parallel Threads for each item in self.data (via _load_cache_item func).    

But on line 818 self.data = data becomes the full list again (with duplicates).  If this assignment happens before paraller Threads are executed, wouldn't it potentially have unpredictable behaviour of caching wrong data for each hash.  A list of returned hashed data could be with duplicates, and the list of hashes (self._hash_keys) will be without.  Then using such hashes will return wrong result (potentially). 

Recommendations: you can try to deepcopy(data) and pass it as a parameter to _compute_cache() and it's sub-functions.  Also unify some sub-fuctions (_compute_cache is called only once, there is no need for it to be a subfunction).   Alternatively  we could keep existing logic (without this PR), (and for shared memory, pass it as Union[ListProxy, DictProxy] single parameter. It will be up to a user to supply the correct cashe object in case of using hash.

monai/data/dataset.py
```diff
@@ -850,21 +850,20 @@ def _load_cache_item(self, idx: int):
...
         if self.hash_as_key:
             key = self.hash_func(self.data[index])
-            if key in self._cache:
+            if key in self._hash_keys:
                 # if existing in cache, get the index
-                index_ = key  # if using hash as cache keys, set the key
+                index = self._hash_keys.index(key)
 
-        if isinstance(index_, int) and index_ % len(self) >= self.cache_num:  # support negative index
+        if index % len(self) >= self.cache_num:  # support negative index
```
myron, line 859: this is a bit confusing. at this point the  variable "index" becomes  an index within cached array (potentially much shorter than the full self.data), we should probably call it something else "index_hash" to avoid later confusion with an actual index in self.data.  And retrieve the cache here
right away data = self._cache[index_hash]

monai/data/dataset.py
```diff
@@ -850,21 +850,20 @@ def _load_cache_item(self, idx: int):
...
             # no cache for this index, execute all the transforms directly
-            return super()._transform(index_)
+            return super()._transform(index)
```
myron, line 861: it seems, at this point "index" could be in the hash space (if hashing by key), so this _transform(index) will run on the wrong element. see comment above for using index_hash

monai/data/dataset.py
```diff
@@ -850,21 +850,20 @@ def _load_cache_item(self, idx: int):
...
         # load data from cache and execute from the first random transform
         start_run = False
         if self._cache is None:
```
myron, line 864: this line is not part of the PR, but out of curiosity, do we expect _cache to be None at this point?   I thought we always set it in init / set_data at the beginning.

monai/data/dataset.py
```diff
@@ -850,21 +850,20 @@ def _load_cache_item(self, idx: int):
...
             self._cache = self._fill_cache()
```
myron, line 865: if hashing by key, this will cache wrong things, since self.data  includes all the duplicates

myron (changes requested):

monai/data/dataset.py
Nic-Ma: You are right, you can see I already did the change in https://github.com/Project-MONAI/MONAI/pull/5365, so I don't want to make duplicated change in this PR.
But anyway, let me also update it in this PR.

Thanks.

monai/data/dataset.py
Nic-Ma: You are right, you can see I already did the change in https://github.com/Project-MONAI/MONAI/pull/5365, so I don't want to make duplicated change in this PR.
Anyway, I already updated it in the latest commit.

Thanks.

monai/data/dataset.py
Nic-Ma: Hi @myron ,

Maybe the logic here is a little bit confusing, actually, this line only executes when **the target item is not cached**, we don't change any cache content during training. So the `index` is certainly the original index instead of the hash list, here we call `super()` to treat it just as a regular `Dataset` item to run transforms.

Thanks.

monai/data/dataset.py
Nic-Ma: I agree the logic here is a little bit confusing.
Same as above comment, if the target item is cached, so we update the index to the hashed list, in this case, it definitely will not call `super()._transform(index)`. That's 2 different cases: fetch `cached item` vs `uncached item`.

Thanks.

monai/data/dataset.py
```diff
@@ -810,12 +811,11 @@ def _compute_cache():
             # only compute cache for the unique items of dataset
             mapping = {self.hash_func(v): v for v in data}
             self.data = list(mapping.values())
-            cache_ = _compute_cache()
-            self._cache = dict(zip(list(mapping)[: self.cache_num], cache_))
-            self.data = data
+            self._hash_keys = list(mapping)[: self.cache_num]
```
myron, line 814: self.cache_num seems to be always 0 at this point, so your self._hash_keys will be always empty. !?

monai/data/dataset.py
myron: changing index  ( to hash list space) can lead to  future bugs that is difficult to detect, if in the future  someone else edits this class  they will need to remember that "index" could be in different spaces. you can use a new variable "index_"  (or return cache sooner).  Also the way we detect if it is  in the cache is also confusing `if index % len(self) >= self.cache_num`,  we rely on self.cache_num to be correctly set and index to be in the correct (non-hashed space).  I think more explicit check will be more robust

```
if index < len(self._cache):
  data = self._cache[index]
 else:
  return super()._transform(index) 
```

monai/data/dataset.py
myron: yes, you're correct that line happened to only execute when target item is not cached.  But this is very confusing,  and if someone changes anything e.g. self.cache_num calculation or it can lead to bugs that will be very hard to detect (it will return wrong image, without showing any errors)

monai/data/dataset.py
Nic-Ma: Thanks for your finding, yes, I made a mistake here.
I updated in the latest commit and added a new unit test to cover it.

Thanks.

monai/data/dataset.py
myron: is this not an issue in the most recent update?

monai/data/dataset.py
Nic-Ma: Sure, I updated all this part logic in the latest commit to make it more straightforward.
And `self.cache_num` should always represent the length of cache content, we may use a larger buffer as `self._cache` in future.
thanks for your suggestion.

monai/data/dataset.py
Nic-Ma: Sure, I updated all this part logic in the latest commit to make it more straightforward.

Thanks for your feedback.

monai/data/dataset.py
Nic-Ma: Hi @myron ,

The multi-thread `imap` logic here is `blocking`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/dataset.py#L825
Refer to: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.map
So I think the assignment will not happen before parallel threads are executed.

Thanks.

myron:
it looks better, and works good, I tested it.  I'd still suggest explicitly providing data to _fill_cache() , instead of modifying self.data twice. It will be more explicit and more compact.  you can use these functions below

```
   def set_data(self, data: Sequence):
        """
        Set the input data and run deterministic transforms to generate cache content.
        Note: should call this func after an entire epoch and must set `persistent_workers=False`
        in PyTorch DataLoader, because it needs to create new worker processes based on new
        generated cache content.
        """

        self.data = data
        self._cache = []
        self._hash_keys = []
        self.cache_num = min(int(self.set_num), int(len(self.data) * self.set_rate), len(self.data))

        if self.cache_num <= 0:
            return

        if self.hash_as_key:
            # only compute cache for the unique items of dataset
            mapping = {self.hash_func(v): v for v in data}
            data = list(mapping.values())[: self.cache_num]
            self._hash_keys = list(mapping)[: self.cache_num]

        self._cache = self._fill_cache(data)

    def _fill_cache(self, data) -> List:
        if self.cache_num <= 0:
            return []
        if self.progress and not has_tqdm:
            warnings.warn("tqdm is not installed, will not show the caching progress bar.")
        with ThreadPool(self.num_workers) as p:
            if self.progress and has_tqdm:
                return list(
                    tqdm(
                        p.imap(self._load_cache_item, data),
                        total=len(data),
                        desc="Loading dataset",
                    )
                )
            return list(p.imap(self._load_cache_item, data))

    def _load_cache_item(self, item):
        """
        Args:
            item: element from the input data sequence.
        """
        for _transform in self.transform.transforms:  # type:ignore
            # execute all the deterministic transforms
            if isinstance(_transform, Randomizable) or not isinstance(_transform, Transform):
                break
            _xform = deepcopy(_transform) if isinstance(_transform, ThreadUnsafe) else _transform
            item = apply_transform(_xform, item)
        if self.as_contiguous:
            item = convert_to_contiguous(item, memory_format=torch.contiguous_format)
        return item
```

Nic-Ma:
Hi @myron ,

Thanks for your suggestion.
I updated the PR to avoid changing `self.data`, I kept the `_compute_cache` function because this is a minor issue in your code that we should compute `self.cache_num` on `data` instead of `self.data`.

Thanks.

myron:
> Hi @myron ,
> 
> During deep tests, I found that if the `data` in `p.imap(self._load_cache_item, data)` is complicated `iterable` object (like another `Dataset` object), this multi-thread imap may do wrong slicing for the items. So I reverted the change, can we try to optimize the `self.data` logic later when we got more clear idea? I think that's not in the scope of this PR.
> 
> Thanks in advance.

Thank you for the reply.  So if data is a complicated iterable and p.imap does not handle its slicing, perhaps we can compute and list of non-repeated data **indices** and iterate on them. It seems like a more robust solution, and we don't need to a workaround to change self.data temporarily.   Please see below.  If this solution does not work for any reason , I'm okay with your PR as is.  thank you

   ```
 def set_data(self, data: Sequence):
        """
        Set the input data and run deterministic transforms to generate cache content.
        Note: should call this func after an entire epoch and must set `persistent_workers=False`
        in PyTorch DataLoader, because it needs to create new worker processes based on new
        generated cache content.
        """

        self.data = data
        self._cache = []
        self._hash_keys = []
        self.cache_num = min(int(self.set_num), int(len(self.data) * self.set_rate), len(self.data))

        if self.cache_num <= 0:
            return

        if self.hash_as_key:
            # only compute cache for the unique items of dataset
            mapping = {self.hash_func(data[i]): i for i in range(len(data))}
            self._hash_keys = list(mapping)[: self.cache_num]
            indices = list(mapping.values())[: self.cache_num]
        else:
            indices = list(range(self.cache_num))
  
        self._cache = self._fill_cache(indices)

    def _fill_cache(self, indices=None) -> List:

        if self.cache_num <= 0:
            return []

        if indices is None:
            indices = list(range(self.cache_num))

        if self.progress and not has_tqdm:
            warnings.warn("tqdm is not installed, will not show the caching progress bar.")
        with ThreadPool(self.num_workers) as p:
            if self.progress and has_tqdm:
                return list(
                    tqdm(
                        p.imap(self._load_cache_item, indices),
                        total=len(indices),
                        desc="Loading dataset",
                    )
                )
            return list(p.imap(self._load_cache_item, indices))
```

Nic-Ma:
Hi @myron ,

Thanks for your suggestion.
I have updated the PR according to your proposal, could you please help review it again?

Thanks in advance.

myron (approved):

Issue #5387, opened 2022-10-24
Add pretrained weights support for HoverNet


yiheng-wang-nv:
HoverNet uses resnet50 based imagenet pretrained weights for the encoder (see: https://github.com/vqdang/hover_net#training). I think we should also enable this feature.

PR #5405, opened 2022-10-26
5386 Update HoverNet structure

### LINKED ISSUE #5386: Inconsistent places of HoverNet
When I compared our implementation of HoverNet and the [referred version](https://github.com/vqdang/hover_net/blob/master/models/hovernet/net_desc.py), I found a different place:
for each residual block, the kernel size of the second conv layer is 3 (see [here](https://github.com/vqdang/hover_net/blob/master/models/hovernet/net_desc.py#L36)). However, our implementation uses the value 1 (see [here](https://github.com/Project-MONAI/MONAI/blob/4b41ad906030477c277c2b82835058ed71a4a9c0/monai/networks/nets/hovernet.py#L187)) .

In addition, there is a small issue that the type of `drop_first_norm_relu` should be `bool` rather than `int`.

yiheng-wang-nv:

Fixes #5386 #5387 .

### Description

This PR modifies HoverNet via change the kernel size and padding size in order to be consistent with the reference source.

This PR also enables to output same shape as the input when using fast mode (takes the conic branch version (see: https://github.com/vqdang/hover_net/tree/conic) for reference).

What's more, the functionality of loading pretrained weights is added, users need to add the url which contains the weights that follow the format of the encoder.

Finally, this PR adds the `np_out_channels` arg to control the output channels of the nucleus prediction branch.

### Types of changes


yiheng-wang-nv:
The changed HoverNet is still not 100% the same as the reference version on the encoder side, since we use conv layers to do the padding but the referred version defines `TFSamepaddingLayer ` (see [here](https://github.com/vqdang/hover_net/blob/842964dc5d26ffe196d236d16fad16643a103850/models/hovernet/net_utils.py#L39)) and employ it before the conv layer.

Hi @JHancox , could you please share some experiences of the difference here? In my point of view, it's more convenient to use your method since we do not need to maintain more layers within the network.

Nic-Ma (approved):

Nic-Ma:
Hi @KumoLiu ,

Could you please help verify the PR in your draft training pipeline?
If everything works fine, then let's invite more reviewers.

Thanks in advance.

monai/networks/nets/hovernet.py
```diff
@@ -27,19 +27,26 @@
...
 from monai.networks.layers.factories import Conv, Dropout
 from monai.networks.layers.utils import get_act_layer, get_norm_layer
 from monai.utils.enums import HoVerNetBranch, HoVerNetMode, InterpolateMode, UpsampleMode
 from monai.utils.module import export, look_up_option
 
-__all__ = ["HoVerNet", "Hovernet", "HoVernet", "HoVerNet"]
+__all__ = ["HoVerNet", "Hovernet", "HoVernet", "HoVerNet", "HOVERNET_MODELS"]
+
+HOVERNET_MODELS = {
+    "preact-resnet50": "https://drive.google.com/u/1/uc?id=1KntZge40tAHgyXmHYVqZZ5d2p_4Qr2l5&export=download"
```
wyli, line 48: could you move this to https://github.com/Project-MONAI/MONAI-extra-test-data/releases/tag/0.8.1 which is more stable. and if this is trained from imagenet we may need comment on the license.

yiheng-wang-nv: Thanks @wyli , this link comes from the referred HoverNet repo: https://github.com/vqdang/hover_net#data-format
the author uploaded the weights into the google drive and I just copy the google drive link. Could we download and copy to releases? Or let me just add some doc strings?

JHancox:
> The changed HoverNet is still not 100% the same as the reference version on the encoder side, since we use conv layers to do the padding but the referred version defines `TFSamepaddingLayer ` (see [here](https://github.com/vqdang/hover_net/blob/842964dc5d26ffe196d236d16fad16643a103850/models/hovernet/net_utils.py#L39)) and employ it before the conv layer.
> 
> Hi @JHancox , could you please share some experiences of the difference here? In my point of view, it's more convenient to use your method since we do not need to maintain more layers within the network.

Thanks @yiheng-wang-nv. In their PyTorch code, they tried to emulate their original Tensorflow code, which used Tensorflow's simpler padding/cropping implementation. As you say, this resulted in additional layers that I felt could be eliminated in a more elegant way. So, although there are differences, it's the consistent output that is important.

monai/networks/nets/hovernet.py
```diff
@@ -172,22 +179,24 @@ def __init__(
         dropout_type: Callable = Dropout[Dropout.DROPOUT, 2]
 
         if not drop_first_norm_relu:
-            self.layers.add_module("preact_norm", get_norm_layer(name=norm, spatial_dims=2, channels=in_channels))
```
JHancox, line 175: I see that you renamed some of the layers to match the original. I guess this is to allow the pretrained weights to be loaded. I changed the layer names to be somewhat more interpretable whilst allowing for some tweaking of the network. Do we want to change some of them back? If not, then you could just use a dictionary to map the old to the new. 
I'm open to there being a better layer naming convention. The main point, I feel, is that we do it with a purpose and are consistent. (e.g. If we decide to keep the original layer names then we should probably do it for all layers).

yiheng-wang-nv: Thanks @JHancox , this does introduce some breaking changes. Hi @wyli @ericspod , could you please help to make some comment here? The main purpose of changing the name is to simplify the mapping of pretrained weights. If we need to use non-breaking and more interpretable names, I will change them back and add more rules to map the weights.

monai/networks/nets/hovernet.py
```diff
@@ -27,19 +27,26 @@
...
 
+from monai.apps.utils import download_url
 from monai.networks.blocks import UpSample
 from monai.networks.layers.factories import Conv, Dropout
 from monai.networks.layers.utils import get_act_layer, get_norm_layer
 from monai.utils.enums import HoVerNetBranch, HoVerNetMode, InterpolateMode, UpsampleMode
 from monai.utils.module import export, look_up_option
 
-__all__ = ["HoVerNet", "Hovernet", "HoVernet", "HoVerNet"]
+__all__ = ["HoVerNet", "Hovernet", "HoVernet", "HoVerNet", "HOVERNET_MODELS"]
```
drbeh, line 45: Maybe we don't need "HOVERNET_MODELS" here as it is only used by the classes here and no need to be imported.

monai/networks/nets/hovernet.py
drbeh: Hi @yiheng-wang-nv, I don't think that we can add their weight to our release since it is trained on a data with "CC BY-NC" license, which is more restrictive than MONAI license (Apache 2.0). I suggest to check with @daguangxu on that.

monai/networks/nets/hovernet.py
drbeh: I agree that it is better to be consistent with the naming across MONAI's networks. If we can have a more consistent naming, maybe it can justify the backward incompatibility in the name of layers. @wyli @ericspod @Nic-Ma what do you think?

yiheng-wang-nv: BTW, I think the whole encoder part can be enhanced by making it flexible: not limit to fixed resnet50, and can use more advanced backbones such as seresnext, efficientnet and so on. For example, in the `conic` branch of the referenced repo, the author uses torchvision's resnet, rather than reimplement it (see: https://github.com/vqdang/hover_net/blob/conic/models/hovernet/net_desc.py#L56), it makes the whole network more elegant, and is able to load the torch's pretrained weights directly.
Therefore, in my point of view, considering the possibility of this kind of enhancement, we will introduce many breaking changes.

JHancox: I like the idea of making the encoder interchangeable - makes sense. However, I suspect that @drbeh et al have a deadline to meet so that's probably a decision for him - as it may add complexity. 
As far as the layer naming is concerned, my preference is for whatever is most consistent with the rest of Monai.  Making breaking changes doesn't matter too much at the moment, since we haven't released anything significant yet. The only thing that would break I know of is the new unit test I added to check the kernel consistency with the original model, but I can hold off the PR for this until we have agreed the naming.

Nic-Ma: Hi @JHancox ,

Thanks for your review and discussion here, I agree with you, I feel it's OK to make breaking change for this network.

drbeh:
@JHancox will give @yiheng-wang-nv new unittests by Nov 1 to be added to this PR.

yiheng-wang-nv:
> @JHancox will give @yiheng-wang-nv new unittests by Nov 1 to be added to this PR.

Thanks @JHancox @drbeh . I already received the code of unit tests from Jonny, I will update them into this PR soon.

monai/networks/nets/hovernet.py
```diff
@@ -52,6 +62,7 @@ def __init__(
         act: Union[str, tuple] = ("relu", {"inplace": True}),
         norm: Union[str, tuple] = "batch",
         kernel_size: int = 3,
+        padding: int = 0,
```
KumoLiu, line 65: Is it different with the docstring?

monai/networks/nets/hovernet.py
```diff
@@ -331,11 +365,12 @@ def __init__(
         )
 
         self.output_features.add_module(f"decoderblock{_i + 1}", _seq_block)
+        self.same_padding = same_padding
```
KumoLiu, line 368: Maybe we can remove this line? There seems no place has used it.

yiheng-wang-nv: Thanks! @KumoLiu !

monai/networks/nets/hovernet.py
yiheng-wang-nv: Thanks @KumoLiu , I made a mistake here.

monai/networks/nets/hovernet.py
JHancox: It's what enables this possibility (from _init_ docstring):

> decoder_padding: whether to do padding on convolution layers in the decoders. In the referred repository,
>             the architecture is changed to do padding on convolution layers in order to get the same output size
>             as the input. This changed version is used on CoNIC challenge.

Having said that, I don't think it's been tested and suspect that enabling it might break a few things, so perhaps it would be better to remove it for now.

yiheng-wang-nv: Hi @JHancox , I can add the unit tests and actually @KumoLiu has validated this option locally when she is trying to reproduce the performance on CoNIC dataset. What do you think?

yiheng-wang-nv: Hi @JHancox , thanks for posting this question, I double checked the network of conic version, and it only supports to produce same input size on "fast" mode. Therefore, I did some changes in the new commit and supplement some unit tests. Could you please help to recheck it? Thanks!

JHancox (approved):

drbeh (changes requested):
@Nic-Ma let's wait on merging this PR till tomorrow as I am figuring out if we are allowed to have the link to weights or not. Thanks.

KumoLiu:
Hi @yiheng-wang-nv @JHancox @drbeh, maybe we should remove this line in the network to make it more flexible? Generally, we don't normalize in the network.
https://github.com/Project-MONAI/MONAI/blob/52525a815f27a1af09e92b880ddbf99e0137e1ca/monai/networks/nets/hovernet.py#L550

KumoLiu:
Also for the output channel in NP branch, it seems to be hard-coded in 2. Maybe we should make it not hard-code here?
https://github.com/Project-MONAI/MONAI/blob/52525a815f27a1af09e92b880ddbf99e0137e1ca/monai/networks/nets/hovernet.py#L520-L523
Thanks!

JHancox:
Hi @KumoLiu 
> Hi @yiheng-wang-nv @JHancox @drbeh, maybe we should remove this line in the network to make it more flexible?

- I think that's a valid point. It would make more sense to be in the training pipeline

> Also for the output channel in NP branch, it seems to be hard-coded in 2. Maybe we should make it not hard-code here?

- Are you talking about the number of channels as 2, the number of branches or something else?

KumoLiu:
Hi @JHancox 
> * Are you talking about the number of channels as 2, the number of branches or something else?
Yes, I am talking about the number of channels in the NP branch. I think it can be not only 2, but also 1 or other values.
Thanks!

JHancox:
> Hi @JHancox
> 
> > * Are you talking about the number of channels as 2, the number of branches or something else?
> 
> Yes, I am talking about the number of channels in the NP branch. I think it can be not only 2, but also 1 or other values.
> 
> Thanks!

I don't think that there is a need to generalize this, because the purpose of the branch is pretty specific. However, as you mentioned previously, there is a good argument to simply make it a 1 channel output and then apply a sigmoid to the output, rather than having 2 channels with softmax.

drbeh:
> > Hi @JHancox
> > > * Are you talking about the number of channels as 2, the number of branches or something else?
> > 
> > 
> > Yes, I am talking about the number of channels in the NP branch. I think it can be not only 2, but also 1 or other values.
> > Thanks!
> 
> I don't think that there is a need to generalize this, because the purpose of the branch is pretty specific. However, as you mentioned previously, there is a good argument to simply make it a 1 channel output and then apply a sigmoid to the output, rather than having 2 channels with softmax.

@KumoLiu @yiheng-wang-nv, I agree with @JHancox here since we already have another branch (`NC`) that classify further this prediction into subclasses, so the purpose of `NP` would be a binary prediction (unlike other use cases that we might have multi-class segmentation as one output).

KumoLiu:
> > Hi @JHancox
> > > * Are you talking about the number of channels as 2, the number of branches or something else?
> > 
> > 
> > Yes, I am talking about the number of channels in the NP branch. I think it can be not only 2, but also 1 or other values.
> > Thanks!
> 
> I don't think that there is a need to generalize this, because the purpose of the branch is pretty specific. However, as you mentioned previously, there is a good argument to simply make it a 1 channel output and then apply a sigmoid to the output, rather than having 2 channels with softmax.

Even if there is only a choice of 1 or 2, maybe we can also make it not hard-code here?
@yiheng-wang-nv, @drbeh, @Nic-Ma, @wyli, what do you think?

drbeh:
@KumoLiu, having 2 classes where they are mutually exclusive is identical to having one class. I believe it has been a mistake that the original authors have used 2 classes.

monai/networks/nets/hovernet.py
drbeh: @yiheng-wang-nv, please remove the link to gdrive from here and just keep the comment about the existence of pretrained weights with the link to the website. You can add the link to gdrive in the tutorial with appropriate comments about the license as we discussed.

yiheng-wang-nv:
> > > Hi @JHancox
> > > > * Are you talking about the number of channels as 2, the number of branches or something else?
> > > 
> > > 
> > > Yes, I am talking about the number of channels in the NP branch. I think it can be not only 2, but also 1 or other values.
> > > Thanks!
> > 
> > 
> > I don't think that there is a need to generalize this, because the purpose of the branch is pretty specific. However, as you mentioned previously, there is a good argument to simply make it a 1 channel output and then apply a sigmoid to the output, rather than having 2 channels with softmax.
> 
> Even if there is only a choice of 1 or 2, maybe we can also make it not hard-code here? @yiheng-wang-nv, @drbeh, @Nic-Ma, @wyli, what do you think?

Hi, I think add an arg to control the output channel of NP branch makes sense, for reproducing the original result, just need to keep the default setting.

yiheng-wang-nv:
Update: according to the latest discussion, I the [current version ](https://github.com/Project-MONAI/MONAI/pull/5405/files#diff-a7ff37e4794b36ea32f15803f354b295d6143abd0f1c914de6cb35f16ad9d642) will be updated and the load pretrained weights part will be removed due to the license issue.

KumoLiu (approved):
Thanks for the update, looks good to me now!

Nic-Ma (approved):

monai/networks/nets/hovernet.py
```diff
@@ -424,7 +418,7 @@ class HoVerNet(nn.Module):
             get the same output size as the input, and this changed version is used on CoNIC challenge.
             Please note that to get consistent output size, `HoVerNetMode.FAST` mode should be employed.
         dropout_prob: dropout rate after each dense layer.
-        pretrained: whether to load pretrained weights for encoder.
+        pretrained_url: if specifying, will loaded the pretrained weights of preact resnet50 downloaded from the url.
```
drbeh, line 421: Each user is responsible for checking the content of model/datasets and the applicable licenses and determining if suitable for the intended use.

drbeh (approved):

PR #5410, opened 2022-10-26
RandomizableTransformType, LazyTransformType, and MultiSampleTransformType

atbenmurray:
…mpleTransform. RandomizableTransform now inherits from IRandomizableTransform

Fixes # .

### Description

### Types of changes

### DIFF
--- a/monai/transforms/transform.py
+++ b/monai/transforms/transform.py
@@ -26,7 +26,17 @@ from monai.utils import MAX_SEED, ensure_tuple, first
 from monai.utils.enums import TransformBackends
 from monai.utils.misc import MONAIEnvVars
 
-__all__ = ["ThreadUnsafe", "apply_transform", "Randomizable", "RandomizableTransform", "Transform", "MapTransform"]
+__all__ = [
+    "ThreadUnsafe",
+    "apply_transform",
+    "LazyTransformType",
+    "RandomizableTransformType",
+    "MultiSampleTransformType",
+    "Randomizable",
+    "RandomizableTransform",
+    "Transform",
+    "MapTransform",
+]
 
 ReturnType = TypeVar("ReturnType")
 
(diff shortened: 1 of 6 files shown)


monai/transforms/transform.py
```diff
@@ -118,6 +120,79 @@ def _log_stats(data, prefix: Optional[str] = "Data"):
...
+        """
+        Set whether lazy_evaluation is enabled for this transform instance.
+
+        Args:
+            enabled: True if the transform should operate in a lazy fashion, False if not.
+        """
+        raise NotImplementedError()
+
+
+class IRandomizableTransform:
```
wyli, line 157: I understand that we need `IMultiSampleTransform`, but I'm not sure about `IRandomizableTransform` or `ILazyTransform`, these are not required concepts at the moment according to the prototypes... I can redirect my PR (https://github.com/Project-MONAI/MONAI/pull/5407) to your fork if that addresses any potential merging conflicts

ericspod:
Prefixing interfaces with `I` isn't really Pythonic style so could we remove that please?

atbenmurray:
> Prefixing interfaces with `I` isn't really Pythonic style so could we remove that please?

It's neither pythonic nor un-pythonic, as there isn't really anything either way on how to name interface style classes. That said, I get your point. The problem is that we have both IRandomizableTransform (the signature) and RandomizableTransform (some implementation). I would definitely suggest keeping them separate if at all possible

monai/transforms/transform.py
atbenmurray: If you look at the dictionary transform implementations on #4922, they make use of IRandomizableTransform.

The point of IRandomizableTransform is twofold:
1. It can be the way that a transform is detected as randomizable or not, regardless of whether it is implemented in MONAI or extended from another library
2. Randomizable dictionary transforms inherit from RandomizableTransform but then create a randomizable array transform to do the heavy lifting. However, they only delegate _most_ of the randomization to the array class, whereas
they could delegate all of it. IRandomizableTransform would allow that to be fixed as the dictionary transform can just say "hey, I'm randomizable" and then delegate all of its work to the array transform rather than having an implementation that is only partially used

The addition of this interface is partially motivated by the way I've implemented RandRotated in the big PR #4922 .

https://github.com/Project-MONAI/MONAI/blob/62f8172f020c3c8b6ac182b6b4ca5c3805f9877b/monai/transforms/atmostonce/dictionary.py#L174

I might have gone too far in creating a separate Randomizer class that can be reused across the dictionary and array implementations, but I think it results in pretty clean code.

atbenmurray:
> Prefixing interfaces with `I` isn't really Pythonic style so could we remove that please?

```
IRandomizableTransform -> RandomizableTransformInterface  ?
```

monai/transforms/transform.py
wyli: I see, that looks like another refactoring effort introducing new concepts of `Randomizer` and `IRandomizableTransform`... how about we limit the scope for lazy resampling feature, and all the other nice-to-have features stay on a feature branch for now? 

really hope we can ship the lazy resampling soon. The other refactorings add much risks but at the moment they are not critical from a user experience's perspective...

atbenmurray:
Not going ahead with this one; having issues with signing. Started a duplicate PR #5418

PR #5418, opened 2022-10-27
Type classes for lazy resampling

atbenmurray:
Signed-off-by: Ben Murray <ben.murray@gmail.com>

Fixes # .

### Description

### Types of changes


wyli (approved):
Thanks, it looks good to me... cc @ericspod @Nic-Ma @Nic-Ma  if you don't have any other concerns I'll merge this one today

Nic-Ma:
Hi @wyli @atbenmurray ,

I see you 2 experts submitted several PRs today, I was busy with other tasks, will try to review all of them tomorrow.
Please keep it open for 1 more day.

Thanks.

Nic-Ma (commented):
May I know why use the class name with "XXXType"? `LazyTransform` may be enough? Like other classes in the codebase.

Thanks.

wyli:
> I don't know why people don't do this in python in general; our lives would be much easier if pytorch provided a Tensor interface that their Tensor implementation inherits, for example.

numpy and pytorch defines the `__array_function__` protocol instead of defining abstract classes which is in my opinion more flexible duck typing pattern...
- https://pytorch.org/docs/stable/notes/extending.html#extending-torch
- https://numpy.org/neps/nep-0018-array-function-protocol.html

atbenmurray:
> > I don't know why people don't do this in python in general; our lives would be much easier if pytorch provided a Tensor interface that their Tensor implementation inherits, for example.
> 
> numpy and pytorch defines the `__array_function__` protocol instead of defining abstract classes which is in my opinion more flexible duck typing pattern...
> 
> * https://pytorch.org/docs/stable/notes/extending.html#extending-torch
> * https://numpy.org/neps/nep-0018-array-function-protocol.html

This kind of dispatch method is certainly a useful way of solving issues with operator overloading, but it doesn't solve every problem with subtyping ndarrays or tensors. For example, I would have preferred to solve the multi sampling issue by having metatensor keep an internal data tensor that could be shared between multiple metatensor instances, which would be possible with a TensorType interface to inherit from but not possible as metatensor _is_ a tensor, rather than just appearing to be a tensor.

I agree that a pure duck typing approach is quite pythonic, but the big libraries use isinstance a lot to validate their inputs which means that you can't get away with a class that walks and quacks like a tensor but isn't of Tensor type. I think people are generally more used to that than they are that checking whether a method has certain attributes or not.

Nic-Ma:
Hi @ericspod ,

May I know your opinion about this naming question?
I prefer to any method that is consistent with our existing classes in the codebase as we are already 1.0..

Thanks in advance.

atbenmurray:
It is saying the blossom-ci is failing, but when I check it, is says that it succeeded, but I suspect part of it is not running

wyli:
> It is saying the blossom-ci is failing, but when I check it, is says that it succeeded, but I suspect part of it is not running

I guess torch 1.13 released a few hours ago changed some of the typing. probably unrelated to this PR...

atbenmurray:
Nice!

Issue #4604, opened 2022-06-28
Compose deformation vector fields (DVFs)


xiaogengen-007:
**Is your feature request related to a problem? Please describe.**
Does MONAI have code to compose 2 DVFs together? In image registration, we often encounter multi-stage registration. A series of DVF will be generated to warp the moving image sequentially. However, each warping will blur the image a bit due to interpolation. The best approach is to compose the DVF first and warp the image only once.

**Describe the solution you'd like**
The TensorFlow version of voxelmorph already has this feature implemented. But not in PyTorch. It would be great if it is available in MONAI in PyTorch.
The following 3 functions in voxelmorph and neurite are related to the task:
https://github.com/voxelmorph/voxelmorph/blob/dev/voxelmorph/tf/utils/utils.py: def compose(transforms, interp_method='linear', shift_center=True, indexing='ij')
https://github.com/voxelmorph/voxelmorph/blob/dev/voxelmorph/tf/utils/utils.py: def transform(vol, loc_shift, interp_method='linear', indexing='ij', fill_value=None)
https://github.com/adalca/neurite/blob/dev/neurite/tf/utils/utils.py: def interpn(vol, loc, interp_method='linear', fill_value=None)

wyli:
thanks, we take this feature request into consideration when creating https://github.com/Project-MONAI/MONAI/pull/5436, we'll follow up on this with those trainable use cases

ebrahimebrahim:
What would it mean to compose DVFs?

Composing DDFs makes sense because the composite of two displacement fields is a displacement field. Resampling/warping an image via DDF $\vec{v}$ and then via DDF $\vec{u}$ is the same as resampling the image once via their composite. In MONAI this composite can be computed by
```py
u + monai.networks.blocks.Warp(v,u)
```
(e.g. [here](https://github.com/ebrahimebrahim/abcd-registration-experiments/blob/58b1b7e5118729ad22fe6ba8bbcf87c252af26eb/util.py#L255-L263))

[...]

PR #5436, opened 2022-10-30
Replacement Apply and Resample

atbenmurray:
Signed-off-by: Ben Murray <ben.murray@gmail.com>

### Description

This is part of the work towards #4855. It adds:
 - a lazy `apply` method
 - A transform-like wrapper for `apply` called `Apply`
 ~- `MetaMatrix` and related functionality to represent abstracted grid and matrix transforms with metadata~
 - A universal `resample` function that can be used to apply grid / matrix transforms
 ~- Functional spatial and croppad implementations that define but don't apply transforms~

### Types of changes


monai/transforms/lazy/functional.py
```diff
@@ -0,0 +1,201 @@
...
+    cumulative_matrix_ = matrix_from_matrix_container(cumulative_matrix)
+
+    # print(f"applying with cumulative matrix\n {cumulative_matrix_}")
+    a = Affine(norm_coords=False, affine=cumulative_matrix_, spatial_size=cur_shape[1:], normalized=False, **kwargs)
+    data, tx = a(img=data)
+    if isinstance(data, MetaTensor):
+        data.clear_pending_transforms()
+        for p in pending_:
+            data.affine = p.matrix.data
+            data.push_applied_operation(p)
```
wyli, line 199: thanks for the updates, could you create an array transform example based on this? the previous #4922's transform seems to be out of date (https://github.com/Project-MONAI/MONAI/blob/62f8172f020c3c8b6ac182b6b4ca5c3805f9877b/monai/transforms/atmostonce/array.py)

wyli:
thanks, I think this PR introduces quite a few APIs at a time and some of them (`MatrixFactory`/`MetaMatrix`) need more discussions... is it possible to make it smaller, for example, only including `croppad`, `apply`, and an example of `SpatialCrop` transform, so that we can create a working example of lazy cropping, then finalise the `functional` and `apply` API?  cc @ericspod @rijobro @Nic-Ma

atbenmurray:
> thanks, I think this PR introduces quite a few APIs at a time and some of them (`MatrixFactory`/`MetaMatrix`) need more discussions... is it possible to make it smaller, for example, only including `croppad`, `apply`, and an example of `SpatialCrop` transform, so that we can create a working example of lazy cropping, then finalise the `functional` and `apply` API? cc @ericspod @rijobro @Nic-Ma

The problem is that MatrixFactory/MetaMatrix are the classes by which a lot of the internal stuff for lazy resampling is done. They could certainly change before the PR goes in, but that would necessitate some plumbing to happen and I don't know if I'll have time before I am on holiday. The key thing here at the moment is that this PR doesn't change any existing behaviour so if we want to refactor we certainly can.

What they do:
 - MatrixFactory is an attempt to wrap up the various affine operations in a way that it can be simply got from a tensor and used in an OO way.
 - MetaTensor is the transformation + metadata that accompanies it. It contains either a Matrix or a Grid object, depending

In theory, one or both could go, but it might make everything somewhat verbose. Neither are supposed to be user-facing

Might be a good topic for tomorrow, but I am trying to get the PRs as populated as possible for tonight.

monai/transforms/meta_matrix.py
```diff
@@ -0,0 +1,292 @@
...
+
+    def __rmatmul__(self, other):
+        if isinstance(other, MetaMatrix):
+            other_ = other.matrix
+        else:
+            other_ = other
+        return MetaMatrix(other_ @ self.matrix)
+
+
+def matmul(
```
wyli, line 177: this util almost addresses https://github.com/Project-MONAI/MONAI/issues/4604 if the interface is simpler and considers a batch of transforms... any interests in tackling that ticket within this implementation? @atbenmurray

atbenmurray: @wyli sorry, access has been patchy. Let me take a look

atbenmurray: @wyli Yes, I think so. I haven't understood yet how to do it so if there are well defined methods that is great!

atbenmurray: @wyli Hmm, it does appear to be the same principle. Well, at least we are pytorch and they are tensorflow. I did look across a number of libraries but I focused on the pytorch-based ones of course

atbenmurray: @wyli hmm, it seems to be more focused on providing these capabilities for registration based deep learning, as opposed to preprocessing

wyli: if this `matmul` could be the util to implement the key idea and shared between preprocessing and as a differentiable module , that would be nice...

monai/transforms/lazy/utils.py
```diff
@@ -0,0 +1,123 @@
...
+        if isinstance(data, Affine):
+            return True
+        if isinstance(data, DDF):
+            return False
+        if not hasattr(data, "shape") or len(data.shape) < 2:
+            return False
+        return data.shape[-1] in (3, 4) and data.shape[-2] in (3, 4) and data.shape[-1] == data.shape[-2]
+
+
+class DDF:
```
wyli, line 44: `class DisplacementField`

wyli (approved):
@atbenmurray please make more changes if needed and let me know when you think it's ready.. cc @Nic-Ma this mainly introduces necessary APIs `monai.transforms.lazy.*`

PR #5464, opened 2022-11-03
Close image object in `LoadImage`

### LINKED ISSUE #5461: tests.test_wsireader unclosed file
```
2022-11-03T08:33:03.4930436Z test_read_malformats_0 (tests.test_wsireader.TestTiffFileDeprecated) ... ok
2022-11-03T08:33:03.4941640Z test_read_malformats_1 (tests.test_wsireader.TestTiffFileDeprecated) ... ok
2022-11-03T08:33:03.4983021Z test_read_malformats_2 (tests.test_wsireader.TestTiffFileDeprecated) ... unclosed file <_io.FileIO name='/home/runner/work/MONAI/MONAI/tests/testing_data/temp_CMU-1.tiff.tiff' mode='rb' closefd=True>
unclosed file <_io.BufferedReader name='/home/runner/work/MONAI/MONAI/tests/testing_data/temp_CMU-1.tiff.tiff'>
2022-11-03T08:33:04.4968625Z unclosed file <_io.FileIO name='/home/runner/work/MONAI/MONAI/tests/testing_data/temp_CMU-1.tiff.tiff' mode='rb' closefd=True>
```

log: https://github.com/Project-MONAI/MONAI/actions/runs/3379844814/jobs/5612814925 (probably this is introduced much earlier, I only noticed just now)

cc @drbeh

drbeh:
Fixes #5461 

### Description

It closes image objects that are created in `LoadImage` when it has `close` attribute. Also since `img` variable has been over use for both the image array and the image object, it changes `img` to `img_obj` when it is referred to the object.

### Types of changes


monai/transforms/io/array.py
```diff
@@ -263,7 +263,14 @@ def __call__(self, filename: Union[Sequence[PathLike], PathLike], reader: Option
             )
 
         img_array: NdarrayOrTensor
-        img_array, meta_data = reader.get_data(img)
+        img_array, meta_data = reader.get_data(img_obj)
+        # Close image objects if they have close attribute
```
wyli, line 267: this could be done in the reader's `get_data()` implementation? also I'm not sure if the readers want to close objects at every iteration

drbeh: @wyli, I see your point but implementing in `get_data` will reduce the flexibility of the reader component. However, here, if you look at line 239, the image object is being read at each call, so closing it at the end of the call should be fine since we are not returning the image object. Do you agree?

wyli: I don't think within this generic LoadImage should understand any attributes of `img_obj`

drbeh: Then do you suggest to leave the warning as is and close the issue?

wyli: not sure which reader is creating this obj, if it's in the WSIreader, perhaps close it right after this line? https://github.com/Project-MONAI/MONAI/blob/9efc6e9a966751a91f8eaac36e2d205f40686b6c/monai/data/image_reader.py#L1346

drbeh: It is only happening for TiffFile backend. Do you agree that for reading an image (WSI, PNG, PIL, etc.) a user might want to create an image object and call `get_data` multiple times? This is currently a supported behavior and I believe we should not change that.

On the other hand, modifying `LoadImage` to close the image object at the end of the call [if possible] does not change ANY behavior of `LoadImage`, does it?

cc @Nic-Ma @ericspod

wyli: yeah, the current PR breaks the LoadImage API...decoupling the backend objects from LoadImage call is important, maybe you can create some other APIs for wsireader users who want to get_data but not close the object if there are feature requests

wyli: or if it's just a warning maybe we can ignore it... I just found the warning message when I was looking at something else, there's no other issues at the moment

drbeh: Maybe I am missing something but LoadImage API accepts the image reader as input and not the image object, so the object is being created in the call and then abandoned when the call is over. Then for the next time, a new object needs to be created again, right?

If so, the object is not being re-used and closing the object would be actually a desired behavior. Am I missing something here?

wyli: the idea is that the image reader should define the backend object details. so if there's really a common requirement among the backends to postprocess the objects, the LoadImage should have some command like `reader.finalize(img_obj)`

drbeh: That I agree! It would be nice to have a method common to all readers inherited from `ImageReader` so that it closes the image object if possible. So do you suggest to add such method to `ImageReader` abstract class?

wyli: I don't think so, because the other 6 reader implementations don't need this interface. I'd suggest an update in wsireader with the additional arg `wsireader.get_data(...keep_obj=False)`, so the user specifically set `keep_obj=True` the object will be left open after this call

drbeh: It is not a requirement for all WSI reader (only TiffFile). Also that breaks the abstraction as they are inherited `ImageReader`.

I believe the solution in this PR is non-intrusive and does not change any API or intended behavior of ANY component in MONAI as I described above. The other proposed solution is breaking abstraction, and causing backward incompatibility for other use cases (for instance in `PatchWSIReader` where we are caching image objects).

@Nic-Ma @ericspod any thought here?

wyli: for my suggested solution, you can change the call here to use `keep_obj=True`: https://github.com/Project-MONAI/MONAI/blob/27362e482761985daac8d5ef732eb57b50c741f1/monai/data/wsi_datasets.py#L146

it's non-breaking because  `PatchWSIReader._get_data` is a private method

monai/transforms/io/array.py
```diff
@@ -263,7 +263,14 @@ def __call__(self, filename: Union[Sequence[PathLike], PathLike], reader: Option
...
+        if isinstance(img_obj, Sequence):
+            for each_obj in img_obj:
+                if hasattr(each_obj, "close"):
+                    each_obj.close()
+        elif hasattr(img_obj, "close"):
+            img_obj.close()
```
ericspod, line 273:
```suggestion
        for each_obj in ensure_tuple(img_obj):
            if hasattr(each_obj, "close"):
                each_obj.close()
```

monai/transforms/io/array.py
Nic-Ma: Hi @drbeh @wyli ,

I think the idea of adding param in `get_data()` is OK, but the arg name `keep_obj=True` may be not clear enough.

Thanks.

ericspod: I don't feel this is really the correct solution. `LoadImage` isn't concerned with how the readers read and present data, so details like file closing or other finalisation is left up to the reader side. Adding something to close objects in `LoadImage` spreads concern about file handling between these two classes. If we can change the concerned readers to have an added constructor/method argument to control whether it's methods should close files or not then this code wouldn't needed here.

Nic-Ma: Hi @drbeh , @ericspod , @wyli ,

If I understand the problem correctly, how about this: add some parameter to the `WSIReader` init function to control whether we should close the file after reading. Then we can pass additional args to the reader in `LoadImage` by the `arg` or `kwargs`: https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/io/array.py#L143
What do you think?

Thanks.

drbeh: @Nic-Ma I agree that solution would work but it make the `WSIReader` API unnecessarily crowded, considering that in WSI images, we usually don't want to close the image object  and also we rarely use LoadImage (except for loading in a low resolution level like MIL example).

If everyone agrees, I would not change anything here and let the warning of TiffFile be there in the tests.
cc @Nic-Ma @wyli @ericspod

Issue #5394, opened 2022-10-25
Python 3.11 support


rijobro:
Python 3.11 has just been released and looks to be significantly [faster than its predecessors](https://towardsdatascience.com/python-3-11-is-indeed-faster-than-3-10-1247531e771b). Pytorch support for python 3.11 [hasn't been released yet](https://github.com/pytorch/pytorch/issues/86566), but when it has we should add 3.11 to our test scripts. Numpy already supports python 3.11 so all good on that front.

rijobro:
@wyli can we add something like this to our testing?
```bash
conda install -c conda-forge python==3.11 -y
pip install --pre torch --extra-index-url https://download.pytorch.org/whl/nightly/cu117
pip install numpy
```

wyli:
current status:

- deployed cpu minimal tests: https://github.com/Project-MONAI/MONAI/blob/e7bfac61f286782756c083f992e8efc2b14f18b1/.github/workflows/pythonapp-min.yml#L78
- jit testing error skipped https://github.com/Project-MONAI/MONAI/blob/e7bfac61f286782756c083f992e8efc2b14f18b1/tests/utils.py#L712-L715

PR #5481, opened 2022-11-07
5394  adds py311 min tests

wyli:

part of #5394

### Types of changes

cc @rijobro


.github/workflows/pythonapp-min.yml
```diff
@@ -98,6 +98,9 @@ jobs:
       with:
         path: ${{ steps.pip-cache.outputs.dir }}
         key: ubuntu-latest-latest-pip-${{ steps.pip-cache.outputs.datew }}
+    - name: install prerelease and py311
+      if: matrix.python-version == '3.11'
+      run: pip install --pre torch --extra-index-url https://download.pytorch.org/whl/nightly/cpu
```
rijobro, line 103: Is there no GPU available for these tests?

wyli: no, this is the minimum cpu tests, I think this should first work fine because we invest GPU resources.

rijobro: Ok, great. On the pytorch website they mention that some features like JIT aren't there yet, so maybe we could decorate those tests to skip if python version == 3.11. Other tests I think we can just fix so that they work with all versions of python3.

rijobro: Or maybe even better we can figure out whether jit works for a given version and store as global variable:
```python
try:
    # try to do some jitting
    CAN_JIT = True
except SomeException:
    CAN_JIT = False

class TestSomething(TestClass):
    @skipUnless(CAN_JIT)
    def test_something_with_jit(self, ...):
        pass
```

wyli: the only remaining error is this one
```py
  File "/opt/hostedtoolcache/Python/3.11.0/x64/lib/python3.11/site-packages/torch/serialization.py", line 1136, in _load
    result = unpickler.load()
             ^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.0/x64/lib/python3.11/site-packages/torch/_tensor.py", line 70, in _rebuild_from_type_v2
    ret = torch._utils._set_obj_state(ret, state)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.0/x64/lib/python3.11/site-packages/torch/_utils.py", line 369, in _set_obj_state
    for k, v in dict_state.items():
                ^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'items'

```

any idea?

rijobro: If `torch.save` doesn't successfully save anything, then maybe `torch.load` will return `None`?

wyli: the stable release version works fine...

tests/utils.py
```diff
@@ -697,16 +697,21 @@ def test_script_save(net, *inputs, device=None, rtol=1e-4, atol=0.0):
...
+                model=net,
+                filename_or_obj=os.path.join(tempdir, "model.ts"),
+                verify=True,
+                inputs=inputs,
+                device=device,
+                rtol=rtol,
+                atol=atol,
+            )
+    except (RuntimeError, AttributeError):
+        if sys.version_info.minor == 11 and sys.version_info.major == 3:
```
rijobro, line 712:
```suggestion
        if sys.version_info.major == 3 and sys.version_info.minor == 11:
```

rijobro (approved):

tests/utils.py
rijobro: Trivial suggestion, looks more readable as maj.minor.

PR #5492, opened 2022-11-08
Install dependencies on Mac

drbeh:
### Description

Few MONAI dependencies are flagged to be installed only on a Linux system while they are available on Macintosh systems too. This PR fix that to include `imagecodecs` and `tifffile`.

### Types of changes

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -33,8 +33,8 @@ sphinx-autodoc-typehints==1.11.1
 sphinx-rtd-theme==0.5.2
 cucim==22.8.1; platform_system == "Linux"
 openslide-python==1.1.2
-imagecodecs; platform_system == "Linux"
-tifffile; platform_system == "Linux"
+imagecodecs; platform_system == "Linux" or platform_system == "Darwin"
+tifffile; platform_system == "Linux" or platform_system == "Darwin"
 pandas
 requests
 einops


drbeh:
@wyli @Nic-Ma do these changes looks good to you?

wyli:
could you please re-enable the relevant tests to confirm the usage?

drbeh:
> could you please re-enable the relevant tests to confirm the usage?

@wyli may I know which tests do you mean?

drbeh:
We have excluded tests based on the existing of a library and not based on the platform, so they should be enabled automatically.

wyli:
> We have excluded tests based on the existence of a library and not based on the platform, so they should be enabled automatically.

so which tests are enabled now?

> We have excluded tests based on the existence of a library and not based on the platform, so they should be enabled automatically.

ok, may I know which tests are enabled?

drbeh:
> > We have excluded tests based on the existence of a library and not based on the platform, so they should be enabled automatically.
> 
> so which tests are enabled now?

`tests/test_wsireader.py` for `TiffFile` backend should be enabled on Mac.

drbeh:
https://github.com/Project-MONAI/MONAI/actions/runs/3429146901/jobs/5714369965#step:11:19135

wyli (approved):

PR #5495, opened 2022-11-09
5493 Enhance meandice

### LINKED ISSUE #5493: Implement Ensemble Dice
Reference:
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6454006/pdf/fbioe-07-00053.pdf

This metric is helpful for hovernet pipeline

cc @KumoLiu @drbeh

yiheng-wang-nv:

Fixes #5493 .

### Description

### Types of changes

### DIFF
--- a/monai/metrics/meandice.py
+++ b/monai/metrics/meandice.py
@@ -9,14 +9,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Union
+from typing import Tuple, Union
 
 import torch
 
+from monai.metrics.metric import CumulativeIterationMetric
 from monai.metrics.utils import do_metric_reduction, ignore_background, is_binary_tensor
 from monai.utils import MetricReduction, deprecated
-
-from .metric import CumulativeIterationMetric
 
 
 class DiceMetric(CumulativeIterationMetric):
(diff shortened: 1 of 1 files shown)


yiheng-wang-nv:
Hi @wyli , @KumoLiu @Nic-Ma , may need your help to have a quick a look at this draft PR, I'm not confident with the naming of the changes, thanks!

monai/metrics/meandice.py
```diff
@@ -44,6 +43,11 @@ class DiceMetric(CumulativeIterationMetric):
         ignore_empty: whether to ignore empty ground truth cases during calculation.
             If `True`, NaN value will be set for empty ground truth cases.
             If `False`, 1 will be set if the predictions of empty ground truth cases are also empty.
+        compute_sample: if ``True``, each sample's dice metric will be computed before reduction.
```
Nic-Ma, line 46: Maybe call it `sample_wise`? @wyli What do you think?

Thanks.

wyli: perhaps we just modify the existing `reduction` to support this sample mode? also for the typo, `dice` should be changed to `Dice`

drbeh: I agree with @wyli on just modifying `reduction` method and having something like `reduction_sample` and use the `filename` attribute to accumulate the metrics.

myron (changes requested):
This PR adds extra code to compute_dice() and DiceMetric to satisfy an edge case of non-standard metric calculation. 

If want to suggest to create a new class and call it something else , instead of adding more options and crowding the code of  compute_dice and DiceMetric

yiheng-wang-nv:
Hi @myron @wyli @drbeh @Nic-Ma , thanks for the comments, I will close this PR and submit a new one later. I will create a new sub class rather than change the diceMetric

Issue #5498, opened 2022-11-09
__call__() and _compute_metric() in SSIMMetric  


Joeycho:
**Describe the bug**
Hello, I cannot store values as buffers in SSIMMetric, for calculating the metric later.

```py
self.ssim_metric._compute_metric(x,y) # By _compute_metric() it works.
self.ssim_metric(x,y) # It does not work.

```

**To Reproduce**

1.
```py

x = torch.ones([2,1,10,10,10])/2 # ground truth
y = torch.ones([2,1,10,10,10])/2 # prediction
self.ssim_metric = SSIMMetric(data_range =torch.tensor([255.0], device="cpu") ,spatial_dims= 3)
           
# the following line should print 1.0 (or 0.9999)
print(self.ssim_metric(x,y))
```

**Expected behavior**
```
tensor(1.)
```
**Screenshots**
```
*** TypeError: split expects at least a 1-dimensional tensor. `data` should be a batch-first tensor or a list of channel-first tensors, got <class 'torch.Tensor'>
```
**Environment**

Ensuring you use the relevant python executable, please paste the output of:

```
================================
Printing MONAI config...
================================
MONAI version: 1.0.0
Numpy version: 1.23.1
Pytorch version: 1.12.1
MONAI flags: HAS_EXT = False, USE_COMPILED = False, USE_META_DICT = False
MONAI rev id: 170093375ce29267e45681fcec09dfa856e1d7e7
MONAI __file__: /u/home/xxx/.conda/envs/unet_torch/lib/python3.10/site-packages/monai/__init__.py

Optional dependencies:
Pytorch Ignite version: NOT INSTALLED or UNKNOWN VERSION.
Nibabel version: 4.0.2
scikit-image version: 0.19.3
Pillow version: 9.2.0
Tensorboard version: 2.10.1
gdown version: NOT INSTALLED or UNKNOWN VERSION.
TorchVision version: 0.13.1
tqdm version: 4.64.1
lmdb version: NOT INSTALLED or UNKNOWN VERSION.
psutil version: 5.9.2
pandas version: 1.5.0
einops version: NOT INSTALLED or UNKNOWN VERSION.
transformers version: NOT INSTALLED or UNKNOWN VERSION.
mlflow version: NOT INSTALLED or UNKNOWN VERSION.
pynrrd version: NOT INSTALLED or UNKNOWN VERSION.

For details about installing the optional dependencies, please visit:
[...]

PR #5502, opened 2022-11-09
#5498 issues fix pull request for SSIMMetric

Joeycho:
Fixes https://github.com/Project-MONAI/MONAI/issues/5498

### Description

simply add arguments (reduction and get_not_nans) for parent class and update ssim loss function to inherit loss, not nn.module.

### Types of changes


Joeycho:
@wyli Sure, I just added test part.

https://github.com/Project-MONAI/MONAI/pull/5502/commits/a444ea822f7b6431b8054c0c845d7909ab3f2ab2

monai/losses/ssim_loss.py
```diff
@@ -11,12 +11,12 @@
...
 import torch
 import torch.nn.functional as F
-from torch import nn
+from torch.nn.modules.loss import _Loss
 
 from monai.utils.type_conversion import convert_to_dst_type
 
 
-class SSIMLoss(nn.Module):
+class SSIMLoss(_Loss):
```
wyli, line 19: the `forward` is currently hard-coded to return the mean value which is not compatible with the metrics... could you help add the `reduction` option to this class?

Joeycho: [a2a351db00b91a3bd229a1c546dec48d5d76466f](https://github.com/Project-MONAI/MONAI/pull/5502/commits/a2a351db00b91a3bd229a1c546dec48d5d76466f)

Yes, I think metric should report for all elements in each batch, but SSIMLoss return a single value across in each batch. That was an issue.

Temporarily, I used for-loop and if-else, in the batch, but in long term, I think we should break the dependency between SSIMLoss and SSIM Metric, which is also true in DiceLoss and DiceMetric.

Joeycho: Oh, so, rather than editing SSIMLoss,  I edited SSIMMetric, since current SSIMLoss makes sense for loss function.

wyli: makes sense, thanks!

wyli (approved):

PR #5512, opened 2022-11-11
5510 flexible device in Invert

### LINKED ISSUE #5510: InvertD() transform should preserve device of the output
https://github.com/Project-MONAI/MONAI/blob/e44f3b904c8c599e46fdfa22758639c63b47ee59/monai/transforms/post/dictionary.py#L703-L706

Currently the inverse transform Invertd() always changes the output tensor device to `self.device` which is `cpu` (default initialization).  And there is no way to preserve the original device (if it's not know in advance).  

- I recommend the default parameter  "device=None", which is to do NO device conversion. 

- Also the default option post_func is `post_func = lambda x: x
 `, which is a redundant call. Perhaps we can do post_func = None as default, and skip it in that case

wyli:

Fixes #5510

### Description
this makes `device` and `post_func` optional 

### Types of changes

### DIFF
--- a/monai/transforms/post/array.py
+++ b/monai/transforms/post/array.py
@@ -27,6 +27,7 @@ from monai.networks import one_hot
 from monai.networks.layers import GaussianFilter, apply_filter, separable_filtering
 from monai.transforms.inverse import InvertibleTransform
 from monai.transforms.transform import Transform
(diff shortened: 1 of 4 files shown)


Nic-Ma (approved):

myron:
thanks you!

PR #5513, opened 2022-11-11
3694 loss function as cumulative metric

### LINKED ISSUE #3694: Generalize Metric to any loss function
If I introduce a new loss function, I have to change the metric as well. For example: `monai.losses.Dice` and `monai.metrics.DiceMetric`. Most of the losses have not a metric associated.

I wonder if the metrics can be made general to any loss function. Where the loss function would be an argument of the metric. Other metric arguments would be the reduction method and similar things.

The alternative is to develop my own Metric every time I change the loss function. But not ideal imo.

wyli:

Fixes #3694

### Description
adds a wrapper for loss function:
```py
dice_loss = DiceLoss(include_background=True)
loss_metric = LossMetric(loss_fn=dice_loss)
```
so that `loss_metric` it can be used as a metric in:
https://github.com/Project-MONAI/MONAI/blob/b030839e98e6a00c1ab0e53545f60b62dc4da4a4/monai/handlers/ignite_metric.py#L32

### Types of changes


Nic-Ma:
Hi @wyli ,

Users can use the ignite metric directly:
https://github.com/pytorch/ignite/blob/master/ignite/metrics/loss.py
Just like how we use the `Accuracy` of ignite.
CC @vfdev-5 .

Thanks.

wyli:
@Nic-Ma yes I'm aware of that, the assumptions on loss_fn is stronger there: -- requiring both y_pred and y, and -- the output must be a single value and --must be used with an ignite engine... 

In our case, previously I thought the cumulative average is good enough, https://github.com/Project-MONAI/MONAI/blob/89684d677f6c93c6fbd023a614c07485c5223557/monai/metrics/cumulative_average.py#L21
But it doesn't seem to work with IgniteMetric handler.

myron (changes requested):
> assumptions on loss_fn is stronger there: -- requiring both y_pred and y, and -- the output must be a single value and --must be used with an ignite engine...

@wyli  thanks for the PR , but I don't think this is what people are asking for in the issue https://github.com/Project-MONAI/MONAI/issues/3694, they did not ask for Ignite support. (unless I misunderstand).  

And currently we can currently use 

```
ac = CumulativeAverage()
loss_value = someloss(x,y)
ac.append(ac)
```

this will work even if loss_value is an array/list/tensor (and removes nans).  I'll clarify what people in the issue are actually asking, unless this PR addresses something else

wyli:
Hi @myron, as I mentioned in the description of this PR, it's about using the loss value within `monai.handlers.IgniteMetric`. `CumulativeAverage` is not compatible within that class. If you are not sure about the PR, we can discuss, but please don't block it using your admin permission.

wyli:
Hi @Nic-Ma  do you have any further comments here?

Nic-Ma (approved):

monai/metrics/loss_as_metric.py
```diff
@@ -0,0 +1,106 @@
+# Copyright (c) MONAI Consortium
```
Nic-Ma, line 1: As the metric name is `LossMetric`, I would suggest to change the file name to `loss_metric.py`.
What do you think?

Thanks.

Nic-Ma:
> @Nic-Ma yes I'm aware of that, the assumptions on loss_fn is stronger there: -- requiring both y_pred and y, and -- the output must be a single value and --must be used with an ignite engine...
> 
> In our case, previously I thought the cumulative average is good enough,
> 
> https://github.com/Project-MONAI/MONAI/blob/89684d677f6c93c6fbd023a614c07485c5223557/monai/metrics/cumulative_average.py#L21
> 
> 
> But it doesn't seem to work with IgniteMetric handler.

OK, that makes sense.

Thanks.

myron (approved):

myron:
> Hi @myron, as I mentioned in the description of this PR, it's about using the loss value within `monai.handlers.IgniteMetric`. `CumulativeAverage` is not compatible within that class. If you are not sure about the PR, we can discuss, but please don't block it using your admin permission.

thank you for the reply. In the PR description on the top , it says that it fixes https://github.com/Project-MONAI/MONAI/issues/3694
In my understanding  it does not fix that issue, since people were asking for something else.  Hence, I left the review and requested the change.  Do you prefer me not leaving reviews at all or not requesting changes?  ps: I changed the review status to approve.

PR #5518, opened 2022-11-14
5515 Adjust freeze encoder of hovernet

### LINKED ISSUE #5515: Adjust the encoder freeze function on HoverNet
In the original paper (https://arxiv.org/pdf/1812.06499.pdf), the freeze way is to freeze the whole encoder.

However, during their official implementation, the freeze way is different on the first residual block (see: https://github.com/vqdang/hover_net/blob/842964dc5d26ffe196d236d16fad16643a103850/models/hovernet/net_desc.py#L107)

We may need to adjust the freeze function to align with their implementation.

cc @KumoLiu @JHancox

yiheng-wang-nv:

Fixes #5515  .

### Description

This PR adjusts the freeze encoder function, and also fix the issue of un-trigger some of the unit tests.

### Types of changes

### DIFF
--- a/monai/networks/nets/hovernet.py
+++ b/monai/networks/nets/hovernet.py
@@ -243,6 +243,8 @@ class _ResidualBlock(nn.Module):
         dropout_prob: float = 0.0,
         act: Union[str, tuple] = ("relu", {"inplace": True}),
         norm: Union[str, tuple] = "batch",
+        freeze_dense_layer: bool = False,
+        freeze_block: bool = False,
     ) -> None:
         """Residual block.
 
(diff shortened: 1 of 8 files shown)


monai/networks/nets/hovernet.py
```diff
@@ -423,6 +433,7 @@ class HoVerNet(nn.Module):
             repository, each user is responsible for checking the content of model/datasets and the applicable licenses
             and determining if suitable for the intended use. please check the following link for more details:
             https://github.com/vqdang/hover_net#data-format
+        freeze_encoder: whether to freeze the encoder of the network.
```
Nic-Ma, line 436: "whether to freeze the encoder of the network when executing the backpropagation."
What do you think?

Thanks.

yiheng-wang-nv: Hi @Nic-Ma , IMO, we may not need to bind "freeze" with "executing the backpropagation". I did not see this kind of description in other places.

Nic-Ma: OK, I see.

Nic-Ma (approved):

PR #5531, opened 2022-11-15
Implement `SlidingWindowHoVerNetInferer`

### LINKED ISSUE #5521: HoVerNet ROI-based Inference
The output of HoVerNet patch-based inference (80x80) does not match its input (270x270) and it cannot be corrected by resizing since it is the result of the center crop of the input (due to handling of borders). This will pose a problem when we want inference on larger ROI (>270) since using the current MONAI `SlidingWindowInferer`, we need to take care of each patch's borders (270-80) as well as the border of the giving ROI.

The goal here is that giving an ROI of 1000x1000 for instance return instance segmentation results with size of 1000x1000.

drbeh:
Fixes #5521 

### Description

This PR implement `SlidingWindowHoVerNetInferer` that with HoVerNet model, where the output size is different than the input but cannot be scaled and should be padded and then cropped.

### Types of changes


myron (changes requested):
@drbeh thanks for the PR, but this adds a lot of extra code to cover some special edge case.  I don't think it should be inside of SlidingWindowInferrer (which job should be just to slide over image, and return the result)

I suggest to either a) create a subclass/another-class for hovernet b) or add an options to SlidingWindowInferrer e.g. callback: callable = None.   and if that function is provided, it'll be applied to every patch, then from hover net you can define that callback to do your padding.

drbeh:
> @drbeh thanks for the PR, but this adds a lot of extra code to cover some special edge case. I don't think it should be inside of SlidingWindowInferrer (which job should be just to slide over image, and return the result)
> 
> I suggest to either a) create a subclass/another-class for hovernet b) or add an options to SlidingWindowInferrer e.g. callback: callable = None. and if that function is provided, it'll be applied to every patch, then from hover net you can define that callback to do your padding.

@myron thanks for your comment. I see your point but the current `SlidingWindowInference` also handles the situation when the output is not the same as input but can be corrected by scaling. This one adds the correction by padding.

The weird design of this class (implementing everything in one single function and then calling it) does not allow any subclassing or modular use of it, so all the code should be copies for any extension, which will create a lot of duplicity and make the maintenance hard.

Regarding the use of handlers, as we want to use this in a bundle, it seems that they way that bundles are consumed by MONAI Label [and MOANI Deploy] does not like use of handlers and they will be ignored: https://github.com/Project-MONAI/MONAILabel/blob/809b8d20168ef5043490d55b53f284c15aeadedb/monailabel/tasks/infer/bundle.py

drbeh:
@myron I totally agree with you that we should separate the functionalities and a refactoring is definitely needed.

Unfortunately, handlers wouldn't work for inference in MONAI Label (where our main use is). What I have done instead is to create another class `SlidingWindowHoVerNetInferer` which calls `sliding_window_inferer` but handles some input paddings and cropping in the class. What do you think?

monai/inferers/utils.py
```diff
@@ -193,9 +197,30 @@ def sliding_window_inference(
...
+        else:
+            importance_map = importance_map_
+
         # for each output in multi-output list
         for ss, seg_prob in enumerate(seg_prob_tuple):
             seg_prob = seg_prob.to(device)  # BxCxMxNxP or BxCxMxN
+            if pad_output:
+                seg_prob = F.pad(seg_prob, pad=tuple(window_pad_size), mode="constant", value=cval)
 
             # compute zoom scale: out_roi_size/in_roi_size
```
myron, line 225: @drbeh thank you for updating, looks much better, but I think this is still a lot of extra code just for hovernet case here. and it's not general, what if someone wants to pad with "border" mode, or do something different. 

I suggest to only modify sliding_window_inference by adding on line 281 (just before the seg_prob_tuple for-loop)
```
if callback is not None:
    seg_prob_tuple, importance_map = callback(seg_prob_tuple, important_map)
```
and in the callback function  you can update both importance_map and seg_prob_tuple (you can define the callback from SlidingWindowHoVerNetInferer). But may be other people have a better suggestion, cc @wyli @Nic-Ma

myron (changes requested):

monai/inferers/utils.py
drbeh: @myron thanks for you suggestion. I have implemented in the PR. Would be great if you can review. thanks

myron: @drbeh thank you so much for taking time to do these changes. to me it looks great.  I can't comment on the logic of this class for hovernet, or what this new addition is doing. but do appreciate that it does not clutter the existing class.

myron (approved):

monai/inferers/inferer.py
```diff
@@ -213,6 +213,7 @@ def __call__(
             device,
             self.progress,
             self.roi_weight_map,
+            None,
```
Nic-Ma, line 216: May I know why adding this `None` value here?

Thanks.

monai/apps/pathology/inferers/inferer.py
```diff
@@ -0,0 +1,224 @@
...
+import torch.nn.functional as F
+
+from monai.inferers import Inferer
+from monai.inferers.utils import compute_importance_map, sliding_window_inference
+from monai.utils import BlendMode, PytorchPadMode, ensure_tuple, look_up_option
+
+__all__ = ["SlidingWindowHoVerNetInferer"]
+
+
+class SlidingWindowHoVerNetInferer(Inferer):
```
Nic-Ma, line 26: Is it possible to inherit from `SlidingWindowInferer` to simplify the code?

Thanks.

tests/test_sliding_window_hovernet_inference.py
```diff
@@ -0,0 +1,362 @@
...
+_, has_tqdm = optional_import("tqdm")
+
+TEST_CASES_PADDING = [
+    [None, (1, 3, 16, 8), (4, 4), 7, 0.5, "constant", torch.device("cpu:0"), None, True],
+    ["hover", (1, 3, 16, 8), (4, 4), 7, 0.5, "constant", torch.device("cpu:0"), None, True],
+    [None, (1, 3, 16, 8), (4, 4), 7, 0.5, "constant", torch.device("cpu:0"), (1,) * 4, True],
+    ["hover", (1, 3, 16, 8), (4, 4), 7, 0.5, "constant", torch.device("cpu:0"), (1,) * 4, True],
+]
+
+TEST_CASES = [
```
wyli, line 34: the cases could be imported from test_sliding_window_inference.py

monai/inferers/utils.py
```diff
@@ -44,10 +44,11 @@ def sliding_window_inference(
     sigma_scale: Union[Sequence[float], float] = 0.125,
     padding_mode: Union[PytorchPadMode, str] = PytorchPadMode.CONSTANT,
     cval: float = 0.0,
-    sw_device: Union[torch.device, str, None] = None,
-    device: Union[torch.device, str, None] = None,
+    sw_device: Optional[Union[torch.device, str]] = None,
```
wyli, line 47: `Optional[X]` is equivalent to `Union[X, None]`. https://docs.python.org/3/library/typing.html#typing.Optional so there's no need to update these typing

tests/test_sliding_window_hovernet_inference.py
```diff
@@ -0,0 +1,362 @@
...
+        result = sliding_window_inference(inputs.to(device), roi_shape, sw_batch_size, compute, overlap, mode=mode)
+        np.testing.assert_string_equal(device.type, result.device.type)
+        np.testing.assert_allclose(result.cpu().numpy(), expected_val)
+
+        result = SlidingWindowHoVerNetInferer(roi_shape, sw_batch_size, overlap, mode)(inputs.to(device), compute)
+        np.testing.assert_string_equal(device.type, result.device.type)
+        np.testing.assert_allclose(result.cpu().numpy(), expected_val)
+
+    @parameterized.expand([[x] for x in TEST_TORCH_AND_META_TENSORS])
+    def test_default_device(self, data_type):
```
wyli, line 117: this test is duplicated and doesn't test any new code changes (please also review the other ones)

monai/apps/pathology/inferers/inferer.py
```diff
@@ -0,0 +1,224 @@
...
+            pad_half_2 = pad_width - pad_half_1
+            window_pad_size.extend([pad_half_1, pad_half_2])
+            window_pad_slices.append(slice(pad_half_1, window_s - pad_half_2))
+
+        # Make the padding area of the importance map zero
+        importance_map = torch.zeros(window_shape)
+        importance_map[window_pad_slices] = importance_map_[window_pad_slices]
+
+        seg_prob_tuple = tuple(
+            F.pad(seg_prob, pad=tuple(window_pad_size), mode="constant") for seg_prob in seg_prob_tuple
```
wyli, line 144: please adjust the API to make the `mode` tunable.

wyli (commented):
thanks, I put some comments inline, I haven't checked this PR as part of any inference pipeline.

monai/inferers/inferer.py
drbeh: The `sliding_window_inferer` has an extra argument now (`process_fn`), which `SlidingWindowInferer` is not using it, so it should be set to None.

Nic-Ma: I see, thanks.

wyli (approved):
thanks, this looks good to me, @KumoLiu please let me know if you are testing this otherwise I'll merge this before next week.

drbeh:
@wyli, it has been used in this inference pipeline: https://github.com/Project-MONAI/tutorials/blob/658244251c56dba2ed0e88d067cda0b6d411bc96/pathology/hovernet/infer_roi.py

Nic-Ma (approved):

KumoLiu (approved):

PR #5533, opened 2022-11-15
5532 tensor.shape in tensor.Size

### LINKED ISSUE #5532: bug: MetaTensor .shape() returns Tuple instead of torch.Size
MetaTensor .shape() returns tuple instead of torch.Size. If MetaTensor is subclass of torch.Tensor, it should not change it methods

```
import torch
from monai.data import MetaTensor

a = torch.rand(1,2,3)
print(a.shape, type(a.shape)) # prints torch.Size([1, 2, 3]) <class 'torch.Size'>

a_meta = MetaTensor(a)
print(a_meta.shape, type(a_meta.shape)) # prints (1, 2, 3) <class 'tuple'>

print('numel of last two dims', a.shape[1:].numel()) # prints 6
print('numel of last two dims', a_meta.shape[1:].numel()) # crashes

AttributeError: 'tuple' object has no attribute 'numel'
```

wyli:

Fixes #5532 

### Description
returns early if it's torch.Size

### Types of changes

### DIFF
--- a/monai/data/meta_tensor.py
+++ b/monai/data/meta_tensor.py
@@ -11,6 +11,7 @@
 
 from __future__ import annotations
 
+import functools
 import warnings
 from copy import deepcopy
 from typing import Any, Sequence
@@ -27,6 +28,25 @@ from monai.utils.enums import LazyAttr, MetaKeys, PostFix, SpaceKeys
 from monai.utils.type_conversion import convert_data_type, convert_to_numpy, convert_to_tensor
 
 __all__ = ["MetaTensor"]
+
+
+@functools.lru_cache(None)
+def _get_named_tuple_like_type(func):
+    if (
+        hasattr(torch, "return_types")
+        and hasattr(func, "__name__")
+        and hasattr(torch.return_types, func.__name__)
+        and isinstance(getattr(torch.return_types, func.__name__), type)
+    ):
+        return getattr(torch.return_types, func.__name__)
+    return None
+
+
+def _not_requiring_metadata(ret):
+    return isinstance(ret, (int, str, bytes, torch.Size, torch.dtype, torch.device, np.ndarray)) or not (
+        isinstance(ret, MetaTensor) or (isinstance(ret, Sequence) and any(isinstance(x, MetaTensor) for x in ret))
+    )
+
 
 class MetaTensor(MetaObj, torch.Tensor):
     """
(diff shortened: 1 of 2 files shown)


monai/data/meta_tensor.py
```diff
@@ -253,6 +253,8 @@ def __torch_function__(cls, func, types, args=(), kwargs=None) -> Any:
         # we might have 1 or multiple outputs. Might be MetaTensor, might be something
         # else (e.g., `__repr__` returns a string).
         # Convert to list (if necessary), process, and at end remove list if one was added.
+        if isinstance(ret, torch.Size):  # tensor.shape/tensor.size()
```
myron, line 256: so this works, but is only a workaround for torch.Size,  there might be other bugs of other Tensor return types.

I think the culprit of the problem is update_meta() function
https://github.com/Project-MONAI/MONAI/blob/82852d89e950d1583f8f787996e71ba9aec8d1c1/monai/data/meta_tensor.py#L276

https://github.com/Project-MONAI/MONAI/blob/82852d89e950d1583f8f787996e71ba9aec8d1c1/monai/data/meta_tensor.py#L242

`isinstance(rets, tuple)` returns True for torch.Size (since it's a subclass of tuple) and then, we explicitely re-cast it to tuple()

we could change this line to 
`return type(rets)(out) if isinstance(rets, tuple) else out`

to preserve the exact sub-class type (such as torch.Size),  

but overall I don't understand why we modify the return types of at all, and check for tuples there,  I dont have a good understanding of this function.  so perhaps there is a better , more general fix

wyli: The idea here is that if a torch function returns a torch.Size already, there's no need to try to attach meta data to the return value.

The tuple check and creation you mentioned is indeed suspicious, probably it will cause bugs for the ret if it's a tuple subclass such as namedtuple. But I'm not aware of any function to trigger this bug, maybe it's ok to keep it simple for now.

myron: so I'm okay we your fix as is, this solves my particular issue.  But I agree with your conclusion that this Bug will reappear again for some other return type.

I checked `MetaTensor.update_meta` and it's only applicable for a list of MetaTensors, so may be we can simply check if ret is not MetaTensor or not a sequence of MetaTensors then return ret right away?

Nic-Ma (approved):
Hi @myron , do you have any other comments?

Thanks.

myron:
> Hi @myron , do you have any other comments?
> 
> Thanks.

this looks very good, thanks @wyli , it's a more general fix

myron (approved):

PR #5534, opened 2022-11-16
Add `RandomOrder`

KumoLiu:
Part of #5517.

### Description

When we have a list of transforms, we may want these transforms to do in random order.  Add `RandomOrder` which is a subclass of `Compose` to provide the ability to apply a list of transformations in random order.

### Types of changes


wyli (commented):
code looks good to me, I'm not sure about the class name `RandomOrder`, to me the operation is similar with np.random.shuffle https://numpy.org/doc/stable/reference/random/generated/numpy.random.shuffle.html

KumoLiu:
> code looks good to me, I'm not sure about the class name `RandomOrder`, to me the operation is similar with np.random.shuffle https://numpy.org/doc/stable/reference/random/generated/numpy.random.shuffle.html

Yes, actually I'm referring to the name of torchvision.
https://pytorch.org/vision/stable/_modules/torchvision/transforms/transforms.html#RandomOrder

wyli:
> > code looks good to me, I'm not sure about the class name `RandomOrder`, to me the operation is similar with np.random.shuffle https://numpy.org/doc/stable/reference/random/generated/numpy.random.shuffle.html
> 
> Yes, actually I'm referring to the name of torchvision. https://pytorch.org/vision/stable/_modules/torchvision/transforms/transforms.html#RandomOrder

Ok, that's also implemented with a 'shuffle'. Any particular reason that permutation is used in this implementation?

tests/test_random_order.py
```diff
@@ -0,0 +1,147 @@
...
+
+from parameterized import parameterized
+
+from monai.data import MetaTensor
+from monai.transforms import EnsureChannelFirst, InvertibleTransform, RandomOrder, TraceableTransform, Transform
+from monai.transforms.compose import Compose
+from monai.transforms.transform import MapTransform
+
+
+class X(Transform):
```
wyli, line 23: I think these class test cases could be imported from test_one_of.py if possible.

KumoLiu:
> > > code looks good to me, I'm not sure about the class name `RandomOrder`, to me the operation is similar with np.random.shuffle https://numpy.org/doc/stable/reference/random/generated/numpy.random.shuffle.html
> > 
> > 
> > Yes, actually I'm referring to the name of torchvision. https://pytorch.org/vision/stable/_modules/torchvision/transforms/transforms.html#RandomOrder
> 
> Ok, that's also implemented with a 'shuffle'. Any particular reason that permutation is used in this implementation?

Yes, since `self. transforms` is tuple object, which does not support item assignment.
https://github.com/Project-MONAI/MONAI/blob/624e832e0c871fc18c8032c87ac9346f2074dc07/monai/transforms/compose.py#L126
Thanks!

tests/test_random_order.py
KumoLiu: Thanks! I will update it!

wyli:
> > > > code looks good to me, I'm not sure about the class name `RandomOrder`, to me the operation is similar with np.random.shuffle https://numpy.org/doc/stable/reference/random/generated/numpy.random.shuffle.html
> > > 
> > > 
> > > Yes, actually I'm referring to the name of torchvision. https://pytorch.org/vision/stable/_modules/torchvision/transforms/transforms.html#RandomOrder
> > 
> > 
> > Ok, that's also implemented with a 'shuffle'. Any particular reason that permutation is used in this implementation?
> 
> Yes, since `self. transforms` is tuple object, which does not support item assignment.
> 
> https://github.com/Project-MONAI/MONAI/blob/624e832e0c871fc18c8032c87ac9346f2074dc07/monai/transforms/compose.py#L126
> 
> 
> Thanks!

Since the transforms are changed at every iteration, perhaps change it to a list and then shuffle the lost is more efficient

monai/transforms/compose.py
```diff
@@ -283,3 +283,36 @@ def inverse(self, data):
         _transform = self.transforms[index]
         # apply the inverse
         return _transform.inverse(data) if isinstance(_transform, InvertibleTransform) else data
+
+
+class RandomOrder(Compose):
```
wyli, line 288: The compose's inverse is not compatible with this class, please double check

KumoLiu: Sorry, not sure what did you mean. Seems all the behaviors can be the same as before except for the order of the transforms. Thanks!

wyli: for example with:
```py
fwd_data_1 = transform(data1)
fwd_data_2 = transform(data2)
fwd_inv = transform.inverse(fwd_data_1)
```

`transform(data2)` will shuffle `self.transforms`, and then calling `transform.inverse` it'll use the current `self.transforms` instead of inverting the original ones.. cc @Nic-Ma

monai/transforms/compose.py
```diff
@@ -158,7 +158,7 @@ def flatten(self):
         """
         new_transforms = []
         for t in self.transforms:
-            if isinstance(t, Compose) and not isinstance(t, OneOf):
+            if isinstance(t, Compose) and not isinstance(t, OneOf) and not isinstance(t, RandomOrder):
```
KumoLiu, line 161: Seems this line will cause a length error since it treats `RandomOrder` as an entire object. cc @wyli

wyli: yeah, ideally the `Compose` as a base class shouldn't have the logic checking its subclasses. I think `OneOf` and `RandomOrder` should implement `flatten()` respectively instead of modifying `Compose.flatten()` directly. please help revise this if you know what I mean. otherwise I can do it later

wyli (approved):
Thanks, it looks good to me.

PR #5535, opened 2022-11-16
Add `clip_range` args in `ShiftIntensity` and `AdjustContrast`

KumoLiu:

Part of https://github.com/Project-MONAI/MONAI/issues/5517.

### Description

If we have an RGB image, after shifting intensity, the intensity will go beyond the original range. Users may want to clip the data to the 0-255 range. This PR adds `clip_range` argument in `ShiftIntensity` and `AdjustContrast` to help do that.

### Types of changes

### DIFF
--- a/docs/source/metrics.rst
+++ b/docs/source/metrics.rst
@@ -46,6 +46,11 @@ Metrics
 .. autoclass:: CumulativeIterationMetric
     :members:
 
+`LossMetric`
+------------
+.. autoclass:: LossMetric
+    :members:
+
 `Mean Dice`
 -----------
 .. autofunction:: compute_meandice
--- a/docs/source/modules.md
+++ b/docs/source/modules.md
@@ -264,7 +264,7 @@ A typical bundle example can include:
      ┗━ *license.txt
 ```
 Details about the bundle config definition and syntax & examples are at [config syntax](https://docs.monai.io/en/latest/config_syntax.html).
-A step-by-step [get started](https://github.com/Project-MONAI/tutorials/blob/master/modules/bundles/get_started.ipynb) tutorial notebook can help users quickly set up a bundle. [[bundle examples](https://github.com/Project-MONAI/tutorials/tree/main/bundle), [model-zoo](https://github.com/Project-MONAI/model-zoo)]
+A step-by-step [get started](https://github.com/Project-MONAI/tutorials/blob/master/bundle/get_started.md) tutorial notebook can help users quickly set up a bundle. [[bundle examples](https://github.com/Project-MONAI/tutorials/tree/main/bundle), [model-zoo](https://github.com/Project-MONAI/model-zoo)]
 
 ## Federated Learning
 
--- a/monai/apps/pathology/losses/hovernet_loss.py
+++ b/monai/apps/pathology/losses/hovernet_loss.py
@@ -61,15 +61,21 @@ class HoVerNetLoss(_Loss):
 
         self.dice = DiceLoss(softmax=True, smooth_dr=1e-03, smooth_nr=1e-03, reduction="sum", batch=True)
         self.ce = CrossEntropyLoss(reduction="mean")
-        self.sobel = SobelGradients(kernel_size=5)
+        self.sobel_v = SobelGradients(kernel_size=5, spatial_axes=0)
+        self.sobel_h = SobelGradients(kernel_size=5, spatial_axes=1)
 
     def _compute_sobel(self, image: torch.Tensor) -> torch.Tensor:
+        """Compute the Sobel gradients of the horizontal vertical map (HoVerMap).
+        More specifically, it will compute horizontal gradient of the input horizontal gradient map (channel=0) and
+        vertical gradient of the input vertical gradient map (channel=1).
 
-        batch_size = image.shape[0]
-        result_h = self.sobel(torch.squeeze(image[:, 0], dim=1))[batch_size:]
-        result_v = self.sobel(torch.squeeze(image[:, 1], dim=1))[:batch_size]
+        Args:
+            image: a tensor with the shape of BxCxHxW representing HoVerMap
 
-        return torch.cat([result_h[:, None, ...], result_v[:, None, ...]], dim=1)
+        """
+        result_h = self.sobel_h(image[:, 0])
+        result_v = self.sobel_v(image[:, 1])
+        return torch.stack([result_h, result_v], dim=1)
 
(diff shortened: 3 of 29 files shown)


Nic-Ma:
Should we consider channel-wise values to clip?

Thanks.

wyli:
I think the range 0-255 comes from the dtype uint8, perhaps what's missing is a `dtype` argument instead of clipping? torchvision preserves the input dtype:
```py
>>> torchvision.transforms.functional.adjust_contrast((torch.rand([1, 3, 3])*512).to(torch.uint8), 1)
tensor([[[232,  40, 211],
         [ 65, 238, 172],
         [198,   2,  93]]], dtype=torch.uint8)
```

KumoLiu:
> 

Hi @wyli, thanks for your suggestions, I'm confused about why casting to uint8 can replace clipping. It seems the return values of the two operations are different. Am I missing anything?
```
>>> a = torch.tensor([260, 250])
>>> b = a.to(torch.uint8)
a: tensor([260, 250])
b: tensor([  4, 250], dtype=torch.uint8)
clip a: tensor([255, 250])
```
Thanks!

KumoLiu:
> > I think the range 0-255 comes from the dtype uint8, perhaps what's missing is a `dtype` argument instead of clipping? torchvision preserves the input dtype:
> 
> Casting to uint8 would effectively clip but we want to be able to do it in floating point anyway. There's a number of times where I've used `ScaleIntensity` to ensure a particular value range but having a `ClipIntensity` transform perhaps makes more sense than adding clipping to just these transforms.

Hi @ericspod, thanks for your suggestions! I think having a `ClipIntensity` transform would be a good solution. Just one thing I am concerned that if the user has several transforms that need to clip at the same time (Just like what I did in the HoVerNet pipeline :) ). It may need to call `ClipIntensity` several times. So maybe we can keep this PR and create another one to add `ClipIntensity`. What do you think?
Thanks!

ericspod:
> It may need to call `ClipIntensity` several times. So maybe we can keep this PR and create another one to add `ClipIntensity`. What do you think?

I don't think having a transform multiple times in a sequence is a problem unless there's performance reasons making that slow. I have had multiples of the same transform in the past. So if there's no overriding performance concern motivating keeping `clip_range` in the transforms you've modified here, I would put this into a new transform. Thanks!

wyli:
sure, just to clarify my previous comment, the main motivation of clipping 0-255 is to simulate training examples in uint8, as a data augmentation method. if we add a dtype preserving option, then the problem is solved.

preserving the types is not necessarily just `to(uint8)`, we can use similar logic like this one from torchvision:

https://github.com/pytorch/vision/blob/ad128b753c7e8cc0c600dfddac22ff48fc73c9d9/torchvision/transforms/functional_tensor.py#L258-L259

```py
    bound = _max_value(img1.dtype)
    return (ratio * img1 + (1.0 - ratio) * img2).clamp(0, bound).to(img1.dtype)
```

Nic-Ma:
@ericspod @KumoLiu ,

Maybe you can also consider the existing `ThresholdIntensity` transform?
 https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/intensity/array.py#L708

Thanks.

KumoLiu:
Hi @wyli, I see your point! 
Maybe we could use a `preserve_range` argument to help users to choose whether preserve the original dtype. Not sure if there are any user cases that do not want to preserve the original dtype.
And for `ClipIntensity`, maybe calling the existing `ThresholdIntensity` one twice can help do it.
What do you think? @wyli @ericspod @Nic-Ma

wyli:
> Not sure if there are any user cases that do not want to preserve the original dtype.

it seems the current transform doesn't preserve the original dtype, and it's not possible to specify output dtype...maybe we can improve this, but at the moment the resampling backend doesn't support many dtypes anyway. so I don't have a strong opinion about this PR, was just mentioning the potential of adding `dtype` option.

myron:
> > I think the range 0-255 comes from the dtype uint8, perhaps what's missing is a `dtype` argument instead of clipping? torchvision preserves the input dtype:
> 
> Casting to uint8 would effectively clip but we want to be able to do it in floating point anyway. There's a number of times where I've used `ScaleIntensity` to ensure a particular value range but having a `ClipIntensity` transform perhaps makes more sense than adding clipping to just these transforms.

I agree with @ericspod , I recommend have a separate clipping transform, instead of cluttering existing transforms. Adding clipping option only to 2 of these augmentation transforms seems somewhat arbitrary, and adding it to All augmentation transforms is too much.

KumoLiu:
I think in `ShiftIntensity`, we need to add a `preserve_range` flag or `clip_range` args since now it forced converts data dtype to the original one which will cause an issue when the intensity goes beyond the original range. But for `AdjustContrast` and other intensity transforms, we can discuss later whether add another clip transform.
https://github.com/Project-MONAI/MONAI/blob/624e832e0c871fc18c8032c87ac9346f2074dc07/monai/transforms/intensity/array.py#L238
Thanks!

wyli:
> MONAI/monai/transforms/intensity/array.py

agreed, I think this is a bug https://github.com/Project-MONAI/MONAI/blob/624e832e0c871fc18c8032c87ac9346f2074dc07/monai/transforms/intensity/array.py#L238 which doesn't take into consideration the dtype overflow...

I think we should do similarly this logic: https://github.com/pytorch/vision/blob/ad128b753c7e8cc0c600dfddac22ff48fc73c9d9/torchvision/transforms/functional_tensor.py#L64

drbeh:
@KumoLiu these are great improvements that you are making. Thanks but for hovernet pipeline, please use colorjitter: https://github.com/Project-MONAI/tutorials/blob/7eccb858f0286a40fdd0a2fd4bfbb0702af8e6ef/pathology/tumor_detection/ignite/camelyon_train_evaluate.py#L89-L91

KumoLiu:
> @KumoLiu these are great improvements that you are making. Thanks for putting effort on this, but for hovernet pipeline, please use colorjitter instead as it should include all these color augmentations: https://github.com/Project-MONAI/tutorials/blob/7eccb858f0286a40fdd0a2fd4bfbb0702af8e6ef/pathology/tumor_detection/ignite/camelyon_train_evaluate.py#L89-L91

Yes, I have used it. Thanks!

KumoLiu:
After discussion, I will close this PR and create another PR later.

function2-llx:
FYI, torchvision clips value to [0, 1] for floating points or [0, 255] for uint8 type in any blending operation.
https://github.com/pytorch/vision/blob/a192c95e77a4a4de3a8aeee45130ddc4d2773a83/torchvision/transforms/_functional_tensor.py#L256-L259

Issue #5430, opened 2022-10-29
DataAnalyser should be run on GPU by default


myron:
auto3dseg.DataAnalyser should run on GPU by default. Currently it always runs on Cpu

I see PRs https://github.com/Project-MONAI/MONAI/pull/5278, related to some CuPy bugs, as far as I know there should not be any CuPy operations there at all. 

The default behavior should be to run on GPU if available.

wyli:
closing this for now, please feel free to create a new ticket for the histogram stats minor enhancements.

PR #5536, opened 2022-11-17
DataAnalyzer fixes

myron:
This adds several non-breaking fixes to DataAnalyzer

- do_ccp option defaults to False (as we don't currently use it, and it's pretty slow)
- on GPU, dataloading is in parallel (num_workers = self.workers), previously it was only synchronous
- when label for one case is completely empty (no foreground), it crashes, so this PR fixes, by putting all zeros in label status

this PR  partially addresses https://github.com/Project-MONAI/MONAI/issues/5430

### Types of changes

### DIFF
--- a/monai/apps/auto3dseg/data_analyzer.py
+++ b/monai/apps/auto3dseg/data_analyzer.py
@@ -261,6 +261,9 @@ class DataAnalyzer:
 
         result[DataStatsKeys.SUMMARY] = summarizer.summarize(result[DataStatsKeys.BY_CASE])
 
+        if not self._check_data_uniformity([ImageStatsKeys.SPACING], result):
+            print("Data spacing is not completely uniform. MONAI transforms may provide unexpected result")
+
         if self.output_path:
             ConfigParser.export_config_file(result, self.output_path, fmt=self.fmt, default_flow_style=None)
 
--- a/monai/apps/auto3dseg/data_analyzer.py
+++ b/monai/apps/auto3dseg/data_analyzer.py
@@ -125,7 +125,7 @@ class DataAnalyzer:
         dataroot: str = "",
         output_path: str = "./data_stats.yaml",
         average: bool = True,
-        do_ccp: bool = True,
+        do_ccp: bool = False,
         device: Union[str, torch.device] = "cpu",
         worker: int = 2,
         image_key: str = "image",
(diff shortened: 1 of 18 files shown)


monai/apps/auto3dseg/data_analyzer.py
```diff
@@ -215,18 +213,17 @@ def get_all_case_stats(self, key="training", transform_list=None):
...
-                EnsureTyped(keys=keys, data_type="tensor"),
                 Lambdad(keys=self.label_key, func=_argmax_if_multichannel) if self.label_key else None,
                 SqueezeDimd(keys=self.label_key, dim=0) if self.label_key else None,
-                ToDeviced(keys=keys, device=self.device),
             ]
-        transform_list.append(summarizer)
 
         transform = Compose(transforms=list(filter(None, transform_list)))
 
+        torch.multiprocessing.set_start_method("fork", force=True)
```
wyli, line 225: I think Windows doesn't natively support fork

myron: okay, I removed fork setting

monai/apps/auto3dseg/data_analyzer.py
```diff
@@ -257,7 +252,7 @@ def get_all_case_stats(self, key="training", transform_list=None):
         result[DataStatsKeys.SUMMARY] = summarizer.summarize(result[DataStatsKeys.BY_CASE])
 
         if not self._check_data_uniformity([ImageStatsKeys.SPACING], result):
-            logger.warning("data spacing is not completely uniform. MONAI transforms may provide unexpected result")
+            print("Data spacing is not completely uniform. MONAI transforms may provide unexpected result")
```
mingxin-zheng, line 255: @myron how about using `logger.info`, if this shouldn't be a warning to the user? 

This check originated from a chat with @dongyang0122 for reasons I don't exactly remember. If this is no longer helpful for the analysis, I am okay not to perform a check of the uniformity.

mingxin-zheng (approved):
Thanks @myron for fixing the data analyzer.  It looks good to me.

I made some minor comments above and please feel free to skip them. I marked them as "resolved" because neither of it should be blocking the merge.

monai/auto3dseg/analyzer.py
```diff
@@ -448,7 +463,21 @@ def __call__(self, data):
 
         ndas = [d[self.image_key][i] for i in range(d[self.image_key].shape[0])]
         ndas_label = d[self.label_key]  # (H,W,D)
+
+        if ndas_label.shape != ndas[0].shape:
+            # if image and label shapes are different, check if they are close
+            if np.allclose(list(ndas_label.shape), list(ndas[0].shape), atol=10):
```
mingxin-zheng, line 469: It may be useful to either 
- Mention about this logic in the class docstring, or
- Make `atol` as an optional argument in the `__init__`

PR #5561, opened 2022-11-22
Fix wrong algorithm folder name parsing in ensemble after HPO in Auto3DSeg

### LINKED ISSUE #5558: Auto3DSeg ensembler module doesn't work with Bundle Algorithms generated by the HPO Module
**Describe the bug**

Auto3DSeg HPO module generates bundle algos and append the hyperparameters in the name of the bundle root folder

For examples, to perform a grid_search on learning rate (lr), folders will be like
. 
├── dints_0_lr_0.0001
├── dints_0_lr_0.001
├── dints_0_lr_0.01
└── dints_0_lr_0.1

In the integration test and some possible usage scenarios, these bundle algorithms need to be ensembled by fold, and the HPO folder name would break Auto3DSeg AutoRunner.

Related line: https://github.com/Project-MONAI/MONAI/blob/72130602340847e15f2042b2bb1b64f6f46eeb93/monai/apps/auto3dseg/ensemble_builder.py#L234

Related issue: #5523 #5555

mingxin-zheng:
Signed-off-by: Mingxin Zheng <18563433+mingxin-zheng@users.noreply.github.com>

Fixes #5558 .

### Description

### Types of changes

- [x] Integration tests passed locally by running `python tests/test_integration_autorunner.py`.

### DIFF
--- a/monai/apps/auto3dseg/ensemble_builder.py
+++ b/monai/apps/auto3dseg/ensemble_builder.py
@@ -222,7 +222,8 @@ class AlgoEnsembleBestByFold(AlgoEnsemble):
             best_score = -1.0
             best_model: Optional[BundleAlgo] = None
             for algo in self.algos:
-                identifier = algo[AlgoEnsembleKeys.ID].split("_")[-1]
+                # algorithm folder: {net}_{fold_index}_{other}
+                identifier = algo[AlgoEnsembleKeys.ID].split("_")[1]
                 try:
                     algo_id = int(identifier)
                 except ValueError as err:


wyli (approved):

myron:
mmm , I'm okay with this change. 
But now we're assuming the algorithm name has no underscores "_", otherwise it'll get a wrong a result.
But it's still better then to assume that fold number is always the last number (as currently). 

I think we can merge it, and think of a better way to track the fold number , instead of parsing the name of the folder/algorithm
@mingxin-zheng

myron (approved):

PR #5570, opened 2022-11-23
Enhance the handler `ProbMapProducer` to use `FolderLayout`

### LINKED ISSUE #5566: Enhance the handler `ProbMapProducer` on checking/making output dir
Currently, if `output_dir` is not existing, an error will be raised. I think a better way to address it is to do sth like the `SaveImaged` transform does (see: https://github.com/Project-MONAI/MONAI/blob/a8f67e4278d46b49a87488a41ba38431883e8907/monai/data/folder_layout.py#L95)

drbeh:
Fixes #5566 

### Types of changes


wyli (approved):

wyli:
thanks, I'm pushing the changes from #5576 here to unblock the pre-merge test issue

Issue #5596, opened 2022-11-28
regular conda tests don't work


wyli:
**Describe the bug**
resolving the dependencies took too long time.

https://github.com/Project-MONAI/MONAI/actions/runs/3485067482/jobs/5830262146
```
# packages in environment at /usr/share/miniconda/envs/test:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                        main  
_openmp_mutex             5.1                       1_gnu  
ca-certificates           2022.10.11           h06a4308_0  
certifi                   2022.9.24        py38h06a4308_0  
ld_impl_linux-64          2.38                 h1181459_1  
libffi                    3.4.2                h295c915_4  
libgcc-ng                 11.2.0               h1234567_1  
libgomp                   11.2.0               h1234567_1  
libstdcxx-ng              11.2.0               h1234567_1  
ncurses                   6.3                  h5eee18b_3  
openssl                   1.1.1s               h7f8727e_0  
pip                       22.2.2           py38h06a4308_0  
python                    3.8.15               h3fd9d12_0  
readline                  8.2                  h5eee18b_0  
setuptools                65.5.0           py38h06a4308_0  
sqlite                    3.39.3               h5082296_0  
tk                        8.6.12               h1ccaba5_0  
wheel                     0.37.1             pyhd3eb1b0_0  
xz                        5.2.6                h5eee18b_0  
zlib                      1.2.13               h5eee18b_0  
Collecting package metadata (repodata.json): ...working... done
/home/runner/.profile: line [40](https://github.com/Project-MONAI/MONAI/actions/runs/3485067482/jobs/5830262146#step:5:41):  1912 Killed                  "$CONDA_EXE" $_CE_M $_CE_CONDA "$@"
Solving environment: ...working... 
Error: Process completed with exit code 137.
```

see also https://github.com/Project-MONAI/MONAI/issues/5088

Issue #5088, opened 2022-09-05
Conda Environment Config Very Slow, Need Lightweight Versions


ericspod:
**Is your feature request related to a problem? Please describe.**
The included Conda specification file `environment-dev.yml` is very slow to resolve packages, taking 12 minutes on a reasonably modern workstation with the latest conda libraries. The mlflow, einops, and transformers libraries appear to be the main culprits. For slightly older versions of conda or some other environment change the package resolution may not complete at all. 

**Describe the solution you'd like**
A simpler dev YAML file with optional luxuries stripped out would be more convenient for quick setup. Additionally a separate file for a minimal environment, and a runtime environment without development libraries, would be nice. This however introduces the need to make sure newly added dependencies are added to all files, but we have this issue with the requirements-*.txt files already.

**Describe alternatives you've considered**
Installing things manually.

**Additional context**

Minimal environment:

```yaml
name: monai
channels:
  - pytorch
  - defaults
  - conda-forge
dependencies:
  - numpy>=1.17
  - pytorch>=1.6
  - cudatoolkit=11.6
```

Runtime environment:

```yaml
name: monai
channels:
  - pytorch
  - defaults
  - conda-forge
dependencies:
  - numpy>=1.17
  - pytorch>=1.6
  - setuptools>=50.3.0,!=60.0.0
  - ignite==0.4.8
  - gdown>=4.4.0
  - scipy
  - nibabel
  - pillow!=8.3.0  # https://github.com/python-pillow/Pillow/issues/5571
  - scikit-image>=0.14.2
  - tqdm>=4.47.0
  - python-lmdb
  - torchvision
  - psutil
  - pandas
  - requests
  - einops  # slow
  - transformers  # slow
  - pyyaml
  - fire
  - jsonschema
  - pynrrd
  - pydicom
  - h5py
  - pip
  - pip:
    # pip for itk as conda-forge version only up to v5.1
    - itk>=5.2
    # OS-specific needs to be done via pip:
    #   https://github.com/conda/conda/issues/8089
    - openslide-python==1.1.2
    - cucim>=21.8.2; platform_system == "Linux"
    - imagecodecs; platform_system == "Linux"
    - tifffile; platform_system == "Linux"
    - matplotlib!=3.5.0
```

PR #5597, opened 2022-11-28
5596 simplify conda env

wyli:
Fixes https://github.com/Project-MONAI/MONAI/issues/5596
Fixes https://github.com/Project-MONAI/MONAI/issues/5088

### Description
simplify the conda env

### Types of changes


ericspod:
I've noticed that `mlflow` is super slow at times to install using the existing environment, and `einops` and `transformers` are really slow as well. I hadn't figured out why but removing these made installing everything else reasonable. I had issues on Windows that the environment would also default to CPU Pytorch for some reason. On the Pytorch site the conda command now includes `pytorch-cuda` as a package with a specification for a version of CUDA so that possibly should be added here as well.

A bare-minimum environment that appears to work is:

```yaml
name: monai
channels:
  - pytorch
  - defaults
  - nvidia
dependencies:
  - numpy>=1.17
  - pytorch>=1.6
  - pytorch-cuda
```

ericspod:
I also created a runtime version of the environment with every package for development removed, this was a little slimmer for just deploying MONAI somewhere, but its contents will change now with the new way Pytorch is doing its packages.

ericspod:
In the logs I see that the Windows build chose the `cpuonly` package by default. This is specified in `conda.yml` but I've seen this happening as a default thing when installing in Windows and I've always had to setup environments manually there.

wyli:
> In the logs I see that the Windows build chose the `cpuonly` package by default. This is specified in `conda.yml` but I've seen this happening as a default thing when installing in Windows and I've always had to setup environments manually there.

I think manually installing pytorch is the recommended way described here https://pytorch.org/get-started/locally/#start-locally let me try to include the cuda requirement for the ci..

wyli:
I've added the cuda dependency following your minimum environment example @ericspod, it works fine with ubuntu and windows: https://github.com/Project-MONAI/MONAI/actions/runs/3566365419/jobs/5992720823
but is not compatible with macos, so I drop the macOS test. 

Apart from the downloading took some time, it looks good to me. Any further suggestions? @ericspod @ericspod. Dropping these pip package requirements in the yaml will make it faster, 
```
  - pip:
    - -r requirements-dev.txt
```
I can do that, just wanted to get your opinion as it'll very different from the current behaviour of `conda update -f environment-dev.yml`.

ericspod:
I've run with the environment file on my Windows box and it's installed things as I expected with the GPU version of Pytorch.

ericspod (approved):

PR #5602, opened 2022-11-29
Generic input interface for Algo base class

myron:
Making Algo base to have more generic inputs.


myron:
@wyli  I remember you commented that Algo base class should have more generic inputs for predict(). Please take a look.

wyli (approved):

PR #5612, opened 2022-11-30
5611 iter_patch no padding for 0/None patch size

### LINKED ISSUE #5611: iter_patch memory consumption
### Discussed in https://github.com/Project-MONAI/MONAI/discussions/5610

<sup>Originally posted by **schellenchris** November 30, 2022</sup>
Hello,

I was playing around with the GridPatchDataset / training with smaller Slices than the original volume, when I ran into some weird RAM OOM problems. After some investigation i noticed that line 290 in data/utils.py cause the problem:
https://github.com/Project-MONAI/MONAI/blob/d0db5fd9d4da3bc7027d22c52bb9c405c3b9e879/monai/data/utils.py#L288-L290

is there a reason to pad the image that much (especially padding the channels doesn't make sense for me)?
I managed to reduce the memory by 15! GB by modifying the code a bit and just padding the axis slices are generated from. I was just wondering if I am missing out on something.
</div>

wyli:

Fixes #5611

### Description
skip the padding to make smaller memory footprint when the patch size is 0 or None for a particular dimension

### Types of changes

### DIFF
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -287,13 +287,15 @@ def iter_patch(
     padded = bool(mode)
     # pad image by maximum values needed to ensure patches are taken from inside an image
     if padded:
-        arrpad = np.pad(arr, tuple((p, p) for p in patch_size_), look_up_option(mode, NumpyPadMode).value, **pad_opts)
+        is_v = [not bool(p) for p in ensure_tuple_size(patch_size, arr.ndim)]  # no padding for (0 or None) patch_size
+        _pad_size = tuple(p if is_v else 0 for p, v in zip(patch_size_, is_v))
+        arrpad = np.pad(arr, tuple((p, p) for p in _pad_size), look_up_option(mode, NumpyPadMode).value, **pad_opts)
         # choose a start position in the padded image
-        start_pos_padded = tuple(s + p for s, p in zip(start_pos, patch_size_))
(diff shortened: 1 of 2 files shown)


monai/data/utils.py
```diff
@@ -287,13 +288,15 @@ def iter_patch(
     padded = bool(mode)
     # pad image by maximum values needed to ensure patches are taken from inside an image
     if padded:
-        arrpad = np.pad(arr, tuple((p, p) for p in patch_size_), look_up_option(mode, NumpyPadMode).value, **pad_opts)
+        is_v = [not bool(p) for p in ensure_tuple_size(patch_size, arr.ndim)]  # no padding for (0 or None) patch_size
+        _pad_size = tuple(p if is_v else 0 for p, v in zip(patch_size_, is_v))
```
ericspod, line 292:
```suggestion
        _pad_size = tuple(p if v else 0 for p, v in zip(patch_size_, is_v))
```
Could also have ` _pad_size = tuple(p * v for p, v in zip(patch_size_, is_v))` but that's less clear.

wyli:
maybe with this PR the `GridPatch` will require less mem or faster, but it shouldn't change the outcome cc@KumoLiu

KumoLiu (approved):
Thanks for update, looks good to me!

Issue #5613, opened 2022-11-30
CacheDataset shared memory,  during torchrun multi-gpu training


myron:
CacheDataset with shared memory crashes during torchrun distributed training

steps to reproduce, run `torchrun --nproc_per_node=2 main.py`

```
import torch, os
import torch.distributed as dist
import torch.multiprocessing as mp
from monai.data import CacheDataset


rank = int(os.getenv("LOCAL_RANK"))
dist.init_process_group(backend="nccl", init_method="env://") 
torch.cuda.set_device(rank)

dataset = CacheDataset(data=[1,2,3], runtime_cache=True)
```

this errors doesn't accure if spawning multi-gpu processes manually (instead of torchrun). 
this issue is  because "broadcast_object_list" checks for authkey in each process, and  torchrun has these keys different.

PR #5615, opened 2022-11-30
fixing broadcast_object_list during torchrun

myron:
Fixes # https://github.com/Project-MONAI/MONAI/issues/5613

This ensures that each process has the same authkey (same as processes 0) during torchrun multi-gpu, to be able to call broadcast_object_list for shared memory.
Here we broadcast the key from the 0 process  too all other processes.  

### Types of changes

### DIFF
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -829,9 +829,9 @@ class CacheDataset(Dataset):
                 if self._is_dist:
                     # ensure each process has the same authkey, otherwise broadcast_object_list fails
                     authkey = bytearray(mp.current_process().authkey).ljust(32)[:32]
-                    ak_tensor = torch.frombuffer(authkey, dtype=torch.uint8)
-                    if dist.get_backend()==dist.Backend.NCCL:
-                        ak_tensor=ak_tensor.cuda(torch.cuda.current_device())
+                    ak_tensor = torch.from_numpy(np.frombuffer(authkey, dtype=np.uint8))
+                    if dist.get_backend() == dist.Backend.NCCL:
+                        ak_tensor = ak_tensor.cuda(torch.cuda.current_device())
                     dist.broadcast(ak_tensor, src=0)
                     mp.current_process().authkey = ak_tensor.cpu().numpy().tobytes()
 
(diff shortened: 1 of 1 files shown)


monai/data/dataset.py
```diff
@@ -824,6 +825,10 @@ def _compute_cache_num(data_len: int):
 
         def _compute_cache(indices=None):
             if self.runtime_cache:
+                if self._is_dist:
+                    # ensure each process has the same authkey, otherwise broadcast_object_list fails
+                    mp.current_process().authkey = np.arange(32, dtype=np.uint8).tobytes()
```
Nic-Ma, line 830: I personally feel this method of changing the global status of python `multi-processing` is risky, let's investigate other "safer" solutions or disable the shared-memory support (runtime cache feature) for this case directly.
Add @ericspod @wyli also for more discussion as they are python and pytorch experts.

Thanks in advance.

Nic-Ma:
BTW, could you please help test whether our runtime-cache mechanism can work with other `start method` of the PyTorch DataLoader if not using the default:
https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py#L188
I think it can be set by the `multiprocessing_context` arg?

Thanks in advance.

monai/data/dataset.py
myron: I think it should be okay, but  another solution is to broadcast authkey from first process to all others (which should be the same already). I had it here in the initial version, I can put it back

```
    # ensure each process has the same authkey, otherwise broadcast_object_list fails
    authkey = bytearray(mp.current_process().authkey).ljust(32)[:32]
    ak_tensor = torch.from_numpy(np.frombuffer(authkey, dtype=np.uint8))
    if dist.get_backend() == dist.Backend.NCCL:
        ak_tensor = ak_tensor.cuda(torch.cuda.current_device())
    dist.broadcast(ak_tensor, src=0)
    mp.current_process().authkey = ak_tensor.cpu().numpy().tobytes()
```

another approach is to go back to the SharedDataCache PR that I proposed initially, where the shared cache was initialized the master process and passed an an input option.

ericspod: I'm not sure what the implication could be for changing the auth key in this way. I don't see what other code would do if this were changed but it's possible if this is used in a larger program there may be other concurrent code that is affected, but this would be very rare. If we can do with broadcasting the key afterwards rather than changing global data it's probably better, plus it's more likely to not cause weird issues in Windows where we don't have fork semantics.

myron:
Since https://github.com/Project-MONAI/MONAI/pull/5630 removed most of the shared cache handling logic, this PR is no longer applicable.

PR #5620, opened 2022-12-01
Add `safe_dtype_convert`

### LINKED ISSUE #5621: Intensity overflow in `ShiftIntensity`
**Describe the bug**
Now in `ShiftIntensity`, the default is to use the input dtype as the output dtype. As we discussed in https://github.com/Project-MONAI/MONAI/pull/5535, it will cause an issue when the intensity goes beyond the original range. 
https://github.com/Project-MONAI/MONAI/blob/4cf6cf846468bc325d1695e9a13137f88d31daae/monai/transforms/intensity/array.py#L238

I followed @wyli's suggestion and tried to add `convert_data_dtype` referring to https://github.com/pytorch/vision/blob/ad128b753c7e8cc0c600dfddac22ff48fc73c9d9/torchvision/transforms/functional_tensor.py#L64, but I find that what they did may be different with what we want.

```
>>> a = torch.tensor([175, 175], dtype=torch.uint8)
>>> b = convert_image_dtype(a, dtype=torch.float32)
>>> print(b, b.dtype)
tensor([0.6863, 0.6863]) torch.float32
```

I try to add some warning in this [draft pr](https://github.com/Project-MONAI/MONAI/pull/5620). But that still doesn't solve the problem. Because even if the user sees the warning, the user cannot modify it because the `ShiftIntensity` does not have the `dtype` parameter passed in. And if we add `convert_data_dtype` for each transform, it still doesn't look right, because we change the intensity inside the transform, which may not be what the user wants. 
Like what @ericspod and @myron said in https://github.com/Project-MONAI/MONAI/pull/5535, adding a clip transform is indeed a good solution, but it does not solve the current bug, because the data dtype has been changed inside the transform, that is, the intensity has been changed.
At the moment, I have not thought of a better way than adding a WARNING. 
Do you have any suggestions? cc @wyli @ericspod @myron @Nic-Ma

### LINKED ISSUE #5685: test_query_memory doesn't work for certain hardware config
```
ERROR: test_output_str (tests.test_query_memory.TestQueryMemory)
  File “/tmp/MONAI/tests/test_query_memory.py”, line 19, in test_output_str
    self.assertTrue(isinstance(query_memory(2), str))
  File “/tmp/MONAI/tests/utils.py”, line 728, in query_memory
    free_memory = np.asarray(free_memory, dtype=float).T
ValueError: could not convert string to float: ‘[N/A]’
```

KumoLiu:
Fixes #5621 .

this also fixes #5685 by https://github.com/Project-MONAI/MONAI/pull/5620/commits/0cf152ec27301c880422517831d7a9fa25a9ab3f

### Description
As stated in the [issue](https://github.com/Project-MONAI/MONAI/issues/5621), a warning is currently added to remind the user of the intensity overflow.
Add `safe_dtype_convert`
Add `safe` flag in `convert_data_type`, `convert_to_tensor`, `convert_to_numpy`, `convert_to_cupy`

### Types of changes


monai/utils/type_conversion.py
```diff
@@ -366,3 +385,72 @@ def convert_to_list(data: Union[Sequence, torch.Tensor, np.ndarray]) -> list:
...
+    else:
+        is_floating_point = get_equivalent_dtype(dtype, torch.Tensor).is_floating_point
+    dtype = get_equivalent_dtype(dtype, np.array)
+    if is_floating_point:
+        return (np.finfo(dtype).min, np.finfo(dtype).max)  # type: ignore
+    else:
+        return (np.iinfo(dtype).min, np.iinfo(dtype).max)
+
+
+def safe_dtype_convert(data: Any, dtype: Union[DtypeLike, torch.dtype] = None):
```
wyli, line 409: this introduces some duplicated logic with the existing code. could this function simply do the data value change but not change the dtype?

KumoLiu: Hi @wyli, docstring is not quite right, this function actually only does data value change, not change dtype and type.

wyli: I see, perhaps change the function name to `safe_dtype_range` as well and please write a new test class for this function as it might be used by many components in the future. for example that np.clip might not work with cupy arrays.

KumoLiu: Hi @wyli, I have added new unit tests and found that when python version > 3.10, the value cannot be converted to type uint8 without overflow which caused CI to fail.
https://github.com/Project-MONAI/MONAI/actions/runs/3626898100/jobs/6116284269

Not sure about this case, maybe In addition to adding the safe flag as I am doing now, we should first give a warning if it overflows but still can't solve the error unless forced safe convert. Do you have any suggestions? Thanks!

wyli: which specific case is this?

wyli: raising an runtime error when safe=False is actually a better handling, we can add the same exception for the unsafe cases or treat the runtime error as expected.

wyli (approved):

PR #5630, opened 2022-12-02
revise cachedataset runtime_cache modes

wyli:
Fixes #5613

### Description

- `runtime_cache=False`: the default v1.0.1 behaviour
- `runtime_cache=True` or `"thread"`: single process, for caching cuda tensors
- `runtime_cache="process"`: single process workflow + multiprocess dataloader
- `runtime_cache=` user-provided object, could be used to pass a container shared among processes

I feel in this way the user can determine what to use instead of guessing and providing an automated solution... let me know what you think @myron @Nic-Ma, I'm fine if this is eventually merged or not merged

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [x] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [x] Quick tests passed locally by running `./runtests.sh --quick --unittests  --disttests`.
- [x] In-line docstrings updated.
- [x] Documentation updated, tested `make html` command in the `docs/` folder.

PR #5632, opened 2022-12-02
CacheDataset preserve cache_num 

myron:
to partially address: https://github.com/Project-MONAI/MONAI/issues/5633

Since CacheDataset accepts a ListProxy object directly now as runtime_cache option
https://github.com/Project-MONAI/MONAI/pull/5630

we need to make sure it's length is "cache_num", (an this length is not always known to a user beforehand), this way a user can provide an empty cache, which will be set to a correct length

```
shared_cache= Manager().list()
CacheDataset(runtime_cache=shared_cache)
```

### Types of changes

### DIFF
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -838,16 +838,15 @@ class CacheDataset(Dataset):
 
         if self.runtime_cache in (False, None):  # prepare cache content immediately
             self._cache = self._fill_cache(indices)
-            return
-        if isinstance(self.runtime_cache, str) and "process" in self.runtime_cache:
+        elif isinstance(self.runtime_cache, str) and "process" in self.runtime_cache:
             # this must be in the main process, not in dataloader's workers
             self._cache = Manager().list([None] * self.cache_num)
-            return
-        if (self.runtime_cache is True) or (isinstance(self.runtime_cache, str) and "thread" in self.runtime_cache):
+        elif (self.runtime_cache is True) or (isinstance(self.runtime_cache, str) and "thread" in self.runtime_cache):
             self._cache = [None] * self.cache_num
-            return
-        self._cache = self.runtime_cache  # type: ignore
-        return
+        else:
+            self._cache = self.runtime_cache  # type: ignore
+            if len(self._cache) == 0:
+                self._cache[:] = [None] * self.cache_num
 
     def _fill_cache(self, indices=None) -> List:
         """
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -845,7 +845,7 @@ class CacheDataset(Dataset):
             self._cache = [None] * self.cache_num
         else:
             self._cache = self.runtime_cache  # type: ignore
-            if len(self._cache) == 0:
+            if isinstance(self._cache, (list, ListProxy)) and len(self._cache) == 0:
                 self._cache[:] = [None] * self.cache_num
 
     def _fill_cache(self, indices=None) -> List:


monai/data/dataset.py
```diff
@@ -838,16 +838,15 @@ def _compute_cache_num(data_len: int):
...
-        if (self.runtime_cache is True) or (isinstance(self.runtime_cache, str) and "thread" in self.runtime_cache):
+        elif (self.runtime_cache is True) or (isinstance(self.runtime_cache, str) and "thread" in self.runtime_cache):
             self._cache = [None] * self.cache_num
-            return
-        self._cache = self.runtime_cache  # type: ignore
-        return
+        else:
+            self._cache = self.runtime_cache  # type: ignore
+            if len(self._cache) == 0:
+                self._cache[:] = [None] * self.cache_num
```
wyli, line 849: I think you should just use len(data) outside of this class to prepare the cache. Given the hashing mechanism, the actual length is not known anyway. But it'll be at most len(data).

myron: unfortunately, a user may not know the len(data) at the moment of creation of ListProxy (if it's created in master process as  Manager().list()) only in the child processes when the datalist is available, we know len(data).  it will be unnecessary extra coding on a user-side to mitigate it.

and if len(self._cache) is different from self.cache_num, it can create unforeseen bugs in the future.   Currently it'll probably crash if len(self._cach)==0, and self.cache_num>0.

wyli: the runtime_cache and `data: Sequence` must be both provided when creating the instance, why is it not available?  the `len` method is not necessarily implemented by the user-provided object and indeed is not required by the cache reading logic.

myron: 1) there reason why len(data) is not available is because a master process needs to create a shared list before anything else

```
def subfunction(rank, shared_list):
  #mainlogic to load data
  data = ...
  CacheData(data=data, runtime_cache=shared_list)

shared_list=Manager().list()
np.swawn(subfunction, nproc=8, args(shared_list))
```

2)   in your definition of `runtime_cache: Union[bool, str, List, ListProxy] = False,`,  so at the moment when len(self._cache) is called it can only be List or ListProxy.  we can also check for isinstance(ListProxy) if you'd like

wyli: Within the subfunction, the length of the shared list can be changed, right? I mean if we make less assumptions, the user will have more flexibility

myron: I've added to check for type isinstance(self._cache, (list, ListProxy))

myron: > Within the subfunction, the length of the shared list can be changed, right? I mean if we make less assumptions, the user will have more flexibility

we need to simplify things for a user, it was easy to use it before this PR https://github.com/Project-MONAI/MONAI/pull/5630

now as it stands,  a user needs to remember to initialize it's own cache, and make sure it's length is correct  (and no shorter than num_cache) otherwise it will crash. which is very inconvenient.

if nothing else, this PR does not hurt anything, and avoid the crash if empty list is provided.  is there any reason why not merge it?

wyli (changes requested):
please see my reply https://github.com/Project-MONAI/MONAI/issues/5633#issuecomment-1335967445

myron:
> please see my reply [#5633 (comment)](https://github.com/Project-MONAI/MONAI/issues/5633#issuecomment-1335967445)

I saw your reply.  This current PR address only  a bug / crash when a user supplies an empty ListProxy.   It says "Changes requested", which changes are you requesting?

wyli:
> > please see my reply [#5633 (comment)](https://github.com/Project-MONAI/MONAI/issues/5633#issuecomment-1335967445)
> 
> I saw your reply. This current PR address only a bug / crash when a user supplies an empty ListProxy. It says "Changes requested", which changes are you requesting?

Change to close this pr...

myron:
> > > please see my reply [#5633 (comment)](https://github.com/Project-MONAI/MONAI/issues/5633#issuecomment-1335967445)
> > 
> > 
> > I saw your reply. This current PR address only a bug / crash when a user supplies an empty ListProxy. It says "Changes requested", which changes are you requesting?
> 
> Change to close this pr...

Thank you for the reply, I'm not sure I understand why you're asking to close it. This PR has no downside, avoids a crash,  simplifies user effort, and makes len(self._cache) and self.cache_num consistent (which was always the case before). Perhaps somebody else can review it as well.

wyli:
> > > > please see my reply [#5633 (comment)](https://github.com/Project-MONAI/MONAI/issues/5633#issuecomment-1335967445)
> > > 
> > > 
> > > I saw your reply. This current PR address only a bug / crash when a user supplies an empty ListProxy. It says "Changes requested", which changes are you requesting?
> > 
> > 
> > Change to close this pr...
> 
> Thank you for the reply, I'm not sure I understand why you're asking to close it. This PR has no downside, avoids a crash, simplifies user effort, and makes len(self._cache) and self.cache_num consistent (which was always the case before). Perhaps somebody else can review it as well.

It's not as a simple change as you are imagining. for example, if multiple processes concurrently query the length of the shared list and all try to expand the list, the program may potentially run into strange bugs.

myron:
> > > > > please see my reply [#5633 (comment)](https://github.com/Project-MONAI/MONAI/issues/5633#issuecomment-1335967445)
> > > > 
> > > > 
> > > > I saw your reply. This current PR address only a bug / crash when a user supplies an empty ListProxy. It says "Changes requested", which changes are you requesting?
> > > 
> > > 
> > > Change to close this pr...
> > 
> > 
> > Thank you for the reply, I'm not sure I understand why you're asking to close it. This PR has no downside, avoids a crash, simplifies user effort, and makes len(self._cache) and self.cache_num consistent (which was always the case before). Perhaps somebody else can review it as well.
> 
> It's not as a simple change as you are imagining. for example, if multiple processes concurrently query the length of the shared list and all try to expand the list, the program may potentially run into strange bugs.

as far as I understand ListProxy objects are specifically designed for concurrent access during multiprocessing. In fact we've been already using this pattern cache[:] = [None]*n without issues (and even without checking for len(cache)).  So multiple processes successfully called it simultaneously without issues.  we can also check for **rank==0** , just in case, but this seems unnecessary.

It seems you don't want to have this logic inside of CacheDataset. As an alternative proposal I can create another class
SharedCacheDataset(CacheDataset), and inside of subclassed self.set_data() method, have all these checks and even broadcasting for shared memory objects. Would that be better?

wyli:
> > > > > > please see my reply [#5633 (comment)](https://github.com/Project-MONAI/MONAI/issues/5633#issuecomment-1335967445)
> > > > > 
> > > > > 
> > > > > I saw your reply. This current PR address only a bug / crash when a user supplies an empty ListProxy. It says "Changes requested", which changes are you requesting?
> > > > 
> > > > 
> > > > Change to close this pr...
> > > 
> > > 
> > > Thank you for the reply, I'm not sure I understand why you're asking to close it. This PR has no downside, avoids a crash, simplifies user effort, and makes len(self._cache) and self.cache_num consistent (which was always the case before). Perhaps somebody else can review it as well.
> > 
> > 
> > It's not as a simple change as you are imagining. for example, if multiple processes concurrently query the length of the shared list and all try to expand the list, the program may potentially run into strange bugs.
> 
> as far as I understand ListProxy objects are specifically designed for concurrent access during multiprocessing. In fact we've been already using this pattern cache[:] = [None]*n without issues (and even without checking for len(cache)). So multiple processes successfully called it simultaneously without issues. we can also check for **rank==0** , just in case, but this seems unnecessary.
> 
> It seems you don't want to have this logic inside of CacheDataset. As an alternative proposal I can create another class SharedCacheDataset(CacheDataset), and inside of subclassed self.set_data() method, have all these checks and even broadcasting for shared memory objects. Would that be better?

The understanding is not correct and adding a rank check would start to guess the environment again. When the understanding is incomplete, lets please not rush into writing code. My suggestion would be, please focus on the high priority tasks that are assigned to you. And for the other type of ideas, please file bug reports and feature requests, there's no need to speculate the root cause and/or propose workarounds.

myron:
> > It seems you don't want to have this logic inside of CacheDataset. As an alternative proposal I can create another class SharedCacheDataset(CacheDataset), and inside of subclassed self.set_data() method, have all these checks and even broadcasting for shared memory objects. Would that be better?
> 
> The understanding is not correct and adding a rank check would start to guess the environment again. When the understanding is incomplete, lets please not rush into writing code. My suggestion would be, please focus on the high priority tasks that are assigned to you. And for the other type of ideas, please file bug reports and feature requests, there's no need to speculate the root cause and/or propose workarounds.

Checking for a rank (and DDP) and it's not a random guessing of the environment, it's a check for a common workflow.  we're are supporting  multi-gpu processing in DDP, in fact it is an essential part of our training. As it stands, we're putting extra burden of these checks (and making sure the length is correct) on the user, which will require more unnecessary user coding, and may result in a crash. 

Since you're blocking this PR, I'm going to close it, and allow you add a proper code solution to issues here: https://github.com/Project-MONAI/MONAI/issues/5633. thank you.

PR #5656, opened 2022-12-05
5649 auto close mlflow run

### LINKED ISSUE #5649: Close MLFlow run automatically when workflow is end
The MLFlow will start a [run](https://mlflow.org/docs/latest/tracking.html#where-runs-are-recorded) in the backend to record experiments' details like parameters, metrics and images.  Currently, the mlflow handler works fine if just run one round of several epoch training like an 100 epochs segmentation task, since the run will be killed as the process end. However, tasks like active learning, network architecture search and automl try to run a few rounds of several epochs training at one time. For exampe, an active learning may run 10 rounds of 100 epochs training in one process. This may lead to a problem that experiment details of all rounds would be recorded to the same run due to the check logic [here](https://github.com/Project-MONAI/MONAI/blob/7614aca16617a8f0350a206dd57f7a9bfca66fd4/monai/handlers/mlflow_handler.py#L166). From the users' experience point of view, these rounds of experiments should be recorded to different runs (one run for one round). Therefore, it would be better that we could choose to shut down the current run after one round of training.

- [ ] Add an `auto_close:bool` parameter which decides whether to close current `run` at the `complete` phase of the workflow, default to False. 
- [ ] Make the default `auto_close` parameter of trainer and evaluator to True to close the run when training and inference is completed.

binliunls:
Fixes #5649 .

### Description

Add the `auto_close` parameter to mlflow_handler so that it can close the mlflow run according to given value.
Update the default handler parameter, so that trainer and evaluator could have `True` value on `auto_close`.

### Types of changes


Nic-Ma (approved):
Looks good to me, please change the unit test.

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -82,6 +82,7 @@ class MLFlowHandler:
         artifacts: paths to images that need to be recorded after a whole run.
         optimizer_param_names: parameters' name in optimizer that need to be record during runing,
             defaults to "lr".
+        auto_close: whether to close the mlflow run in `complete` phase in workflow, default to False.
```
wyli, line 85: argument name could be `close_on_exit` or `close_on_complete`

PR #5659, opened 2022-12-05
import cv2 inside class

rijobro:
The line `optional_import("cv2")` causes the security light of my macOS to light up in green, even for the lowest level command e.g., `import monai`. I was worried I had a virus until I realised what was causing it! To avoid worrying other users, I suggest we only import when a user is using the `VideoDataset`.

also addresses #5660

### Types of changes


Nic-Ma:
Hi @rijobro ,

May I know why you changed the `SmartCacheDataset` logic in this PR? Is there any bug?

Thanks.

monai/data/dataset.py
```diff
@@ -1173,7 +1170,7 @@ def shutdown(self):
         # wait until replace mgr is done the current round
         while not self._try_shutdown():
             time.sleep(0.01)
-        self._replace_mgr.join()
+        self._replace_mgr.join(300)
```
Nic-Ma, line 1173: May I know what issue this `300` change to solve?

Thanks.

wyli: Yes, the 'join' blocks the entire workflow https://github.com/Project-MONAI/MONAI/issues/5660 I've seen it randomly many times, so trying to add a timeout to avoid waiting indefinitely. It's safe to do so because we always check if there is a thread alive and always create a new one with daemon=True.

Issue #4057, opened 2022-04-04
Support string config starts with bundle syntax or expression contains "@"


Nic-Ma:
**Is your feature request related to a problem? Please describe.**
Thanks for @gigony 's feedback, in some corner cases, maybe a string in the bundle config contains our predefined syntax.
For example:
```
"username": "%%%%cool"
```
Or:
```
"money": "$$$$$"
```
Or:
```
"comment": "#####good"
```
And even in an expression, may contain unexpected "@" syntax:
```
"outout_dir": "$@root_dir + '~/nic-ma@nvidia/temp' + 'model.pt'"
```

So it would be nice to have a way to support the syntax in the string config.

holgerroth:
We also have this issue with nvflare where the path to downloaded model files contain the email address of the admin, e.g. `/t1_workspace/admin@nvflare.com/transfer/7def72a7-a224-403c-9969-2eb9da02db7b/workspace/app_server/FL_global_model.pt`

This causes a reference ID error with bundle.

```
packages/monai/bundle/reference_resolver.py", line 231, in update_refs_pattern
    raise KeyError(msg)
KeyError: "can not find expected ID 'nvflare' in the references."
```

Nic-Ma:
Hi @holgerroth ,

Thanks for your report, please note that: Something like: `{"test": "admin@nvflare.com"}` can work in the bundle, We just treat it as a reference in the 2 cases: (1) a string starts with "@", (2) a string starts with "$" and contains "@" in it.

Thanks.

Nic-Ma:
The logic in the dev is not correct, because it always try to check reference even not started with "@" or "$":
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/reference_resolver.py#L225
Should only do it when: `if value_is_expr or value == item`.

Thanks.

PR #5666, opened 2022-12-06
5648 track bundle config in MLFlow

### LINKED ISSUE #5648: Add support to record the executed bundle config content in MLFlow
Currently, we already added support to enable the MLFlow tracking in the bundle execution CLI command.
Would be nice to add the support to record the executed bundle config content.

Nic-Ma:
Fixes #5648 .

### Description

This PR added support to track the executed bundle config in the MLFlow, also fixed https://github.com/Project-MONAI/MONAI/issues/4057#issuecomment-1342618517.

### Types of changes


Nic-Ma:
Hi @wyli , @binliunls ,

I tried to track the executed bundle config as a string param in MLFlow, but unfortunately MLFlow has some limitation:
```
mlflow.exceptions.MlflowException: Param value *** had length 2005, which exceeded length limit of 500
```
So I plan to save the executed bundle config in a file and put it as `artifacts`.
What do you think?

Thanks in advance.

monai/bundle/scripts.py
```diff
@@ -489,6 +490,12 @@ def patch_bundle_tracking(parser: ConfigParser, settings: dict):
                     handlers.append(v)
         elif k not in parser:
             parser[k] = v
+    # save the executed config into file
+    timestamp = time.strftime('%Y%m%d_%H%M%S')
+    filename = parser["output_dir"] + f"/config_{timestamp}.json"
```
binliunls, line 495: Hi @Nic-Ma ,
    The filename var would be ` $@bundle_root + '/eval'/config_20221207_064559.json`, if run with spleen_ct bundle. And in line 498, it would try to save the config with this name.

Thanks,
Bin

monai/bundle/utils.py
```diff
@@ -107,6 +107,7 @@
         "tracking_uri": "$@output_dir + '/mlruns'",
         "experiment_name": "monai_experiment",
         "run_name": None,
+        "execute_config": "placeholder, may fill at runtime",
         "is_not_rank0": (
```
binliunls, line 111: Hi @Nic-Ma ,
When running the bundle, in the init function of mlflow handler the artifacts var input value if `None`.

Thanks,
Bin

monai/bundle/scripts.py
```diff
@@ -489,6 +490,12 @@ def patch_bundle_tracking(parser: ConfigParser, settings: dict):
...
+    # experiment management tools can refer to this config item to track the config info
+    parser["execute_config"] = filename
+    parser.export_config_file(parser.get(), filename)
```
binliunls, line 498: Hi @Nic-Ma ,
     A `path doesn't exist` error would be reported, if the `eval` folder doesn't exists. And the original spleen bundle downloaded from model zoo doesn't have this folder.

Thanks,
Bin

Nic-Ma: Updated in the latest commit, could you please help try it again?

Thanks.

monai/bundle/scripts.py
```diff
@@ -489,6 +490,15 @@ def patch_bundle_tracking(parser: ConfigParser, settings: dict):
                     handlers.append(v)
         elif k not in parser:
             parser[k] = v
+    # save the executed config into file
+    default_name = f"config_{time.strftime('%Y%m%d_%H%M%S')}.json"
+    filepath = parser.get("execute_config", None)
+    if filepath is None:
+        # experiment management tools can refer to this config item to track the config info
+        parser["execute_config"] = os.path.join(parser["output_dir"], default_name)
```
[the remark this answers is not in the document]

Thanks.

binliunls (approved):

monai/bundle/reference_resolver.py
```diff
@@ -221,21 +221,23 @@ def update_refs_pattern(cls, value: str, refs: Dict) -> str:
...
-                else:
-                    raise KeyError(msg)
-            if value_is_expr:
-                # replace with local code, `{"__local_refs": self.resolved_content}` will be added to
-                # the `globals` argument of python `eval` in the `evaluate`
-                value = value.replace(item, f"{cls._vars}['{ref_id}']")
-            elif value == item:
-                # the whole content is "@XXX", it will avoid the case that regular string contains "@"
-                value = refs[ref_id]
+            # only update reference when string starts with "$" or the whole content is "@XXX"
```
wyli, line 224: is this to skip getting id from strings with special characters?

Nic-Ma: Yes, same as this logic:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/reference_resolver.py#L202
I think we forgot to add it for the `update refs` function before.

Thanks.

wyli:
> Windows issue fixed.
> 
> Thanks.

do you know what the root cause is?

monai/utils/misc.py
```diff
@@ -584,3 +585,14 @@ def prob2class(x, sigmoid: bool = False, threshold: float = 0.5, **kwargs):
...
+
+def path_to_uri(path: PathLike) -> str:
+    """
+    Convert a file path to URI.
+
+    Args:
+        path: input file path to convert, can be a string or `Path` object.
+
+    """
+    return Path(path).as_uri()
```
wyli, line 598: as_uri requires absolute path, I think we run into the issue before, should this `path` be converted to an absolute path?

Nic-Ma: That's a good point, added in the latest commit.

Thanks.

Nic-Ma:
> > Windows issue fixed.
> > Thanks.
> 
> do you know what the root cause is?

The root cause is that the previous test code changed `output_dir` from regular path to URI on windows:
https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_integration_bundle_run.py#L108
Then can't use it to dump the config content to a file.
Already updated in the latest commit.

Thanks.

Issue #5653, opened 2022-12-05
utility function to warn changes in default arguments


wyli:
**Is your feature request related to a problem? Please describe.**
https://github.com/Project-MONAI/MONAI/blob/e8c624769b30f55f265995e1c4ff899ead7ff19e/monai/utils/deprecate_utils.py#L119-L127
the `deprecated_arg` can trigger warnings when a deprecating argument is provided by the user,
it however doesn't warn when a default value of an argument has been/will be changed.

for example it's currently not possible to warn about these changes:
- https://github.com/Project-MONAI/MONAI/issues/5652
- https://github.com/Project-MONAI/MONAI/issues/5007

wyli:
agreed, please feel free to create a separate decorator for this feature request.

Shadow-Devil:
My current idea how one could implement this decorator is as follows:
```py
@deprecated_arg_default("name", old_default="test", new_default="new_name", since="1.1.0", changed="1.2.0")
def test(name: str = "test"):
    print(name)
```
`name`, `old_default` and `new_default` are all required.
The decorators only purpose is to validate if the deprecation is valid and to print a warning if `version` >= `since`, 
it doesn't change the default argument by itself (would be too much "magic" in the background IMO and would be confusing if in the function signature the old default is still present, but it's changed by the decorator). 

It will print different messages if `since <= version < changed`,  `version >= changed` and if `since` and `changed` are not present (adopted from `deprecated_default`).

Two open questions I still have:
1. Should I include an "end of warning version"? Otherwise, all future versions will print the deprecation warning if no explicit argument is used.
2. What if a parameters default is changed multiple times? Should I also consider this scenario in my PR?

wyli:
it looks good to me, there's no need to add end of warning or consider multiple changes.

we may consider another simple case of changing the default behaviour but using the same signature:

```diff
def test(name=None):
   _name = "test"
   if name is None:
-      _name = "old_default"
+      _name = "new_default"
   return _name
```

PR #5673, opened 2022-12-07
Mypy check untyped functions

### LINKED ISSUE #5657: Add mypy typchecking of untyped functions for source code (not tests)
MyPy currently only checks functions with type definitions. It would be great to add more type annotations to untyped functions, so they are also type checked.

Following config can be added to the `setup.cfg` so that MyPy also checks untyped functions:
```py
check_untyped_defs = True
```
But running MyPy with this setting will show ~1000 type errors.

Adding type annotations to the codebase so that MyPy runs without errors even when enabling:
```py
check_untyped_defs = True
```

Since most of the type errors happen inside test cases this issue will only include the main source code. Test code can remain dynamically typed without any problems.

**Additional context**
Follow up of #5605.

Shadow-Devil:
Fixes #5657.

### Description

Enable config to also type check untyped functions (not in tests).
- Type annotations were added
- Explicit casts were added
- Variable redefinitions with other types were refactored 
- If no other option was available (or it would be overly verbose) `# type: ignore` was added

### Types of changes


Shadow-Devil:
I'm a bit confused by the auto fix by pre-commit.ci. I thought that type unions with `A | B` syntax was only added in python 3.10 bit MONAI is targetting versions >= 3.7.
See also here: https://docs.python.org/3/library/typing.html#typing.Union

wyli:
> I'm a bit confused by the auto fix by pre-commit.ci. I thought that type unions with `A | B` syntax was only added in python 3.10 but MONAI is targeting versions >= 3.7. See also here: https://docs.python.org/3/library/typing.html#typing.Union

that particular file has `from __future__ import annotations` which triggers pyupgrade of pre-commit.ci to update the type annotations

Shadow-Devil:
Is it ok to add venv to the exclude path of black or should it be in another PR with an extra issue?
It was just quite annoying since my whole pc froze as black tried to format all dependencies, but I can remove it from this PR.

wyli:
> Is it ok to add venv to the exclude path of black or should it be in another PR with an extra issue? It was just quite annoying since my whole pc froze as black tried to format all dependencies, but I can remove it from this PR.

sure. could you please also avoid `assert` https://deepsource.io/gh/Project-MONAI/MONAI/run/66546629-66ba-4424-8758-2fe9f51ac2ed/python/BAN-B101?listindex=0 you can use `isinstance` to check the types

wyli (approved):

monai/visualize/class_activation_maps.py
```diff
@@ -204,12 +205,9 @@ def compute_map(self, x, class_idx=None, layer_idx=-1):
 
     def _upsample_and_post_process(self, acti_map, x):
         # upsampling and postprocessing
-        if self.upsampler:
-            img_spatial = x.shape[2:]
-            acti_map = self.upsampler(img_spatial)(acti_map)
-        if self.postprocessing:
-            acti_map = self.postprocessing(acti_map)
-        return acti_map
+        img_spatial = x.shape[2:]
```
wyli, line 208: this change will require the previous user's input `upsampler=None` being changed to something like `upsampler=torch.nn.Identity`, but I think it should be a minor change.

wyli:
thanks @Shadow-Devil, I can see that the type annotation enhancements addressed a few potential bugs, that's great.

Shadow-Devil:
@wyli Thank you for apprechiating my work! I would like to go on and as my next type checking enhancement type the deprecated & deprecated_args decorators, since otherwise all functions annotated with them lose type information. Do you think this is a good next step? If yes, I will create an issue and work on it😄

wyli:
thanks, I think those utilities will still need some refactoring (such as for https://github.com/Project-MONAI/MONAI/issues/5653), type checking support could come later after that. But it's a good idea and please help create a feature request...

PR #5693, opened 2022-12-08
adds release note v1.1

wyli:

part of #5626
changelog updates

### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ## [1.1.0] - 2022-12-19
 ### Added
 * Hover-Net based digital pathology workflows including new network, loss, postprocessing, metric, training, and inference modules
-* Various enhancements for the Auto3dSeg APIs including default transforms for `DataAnalyzer`, local folder and algo subsets for `AutoRunner`
+* Various enhancements for Auto3dSeg including default transforms for `DataAnalyzer`, folder and algo subsets for `AutoRunner`
 * New bundle API and configurations to support experiment management including `MLFlowHandler`
 * New `bundle.script` API to support model zoo query and download
 * `LossMetric` metric to compute loss as cumulative metric measurement
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+## [1.1.0] - 2022-12-19
+### Added
+### Fixed
+
 ## [1.0.1] - 2022-10-24
 ### Fixes
 * DiceCELoss for multichannel targets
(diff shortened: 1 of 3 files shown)


mingxin-zheng:
Hi @wyli , thanks for drafting the notes. I see there are two items for Auto3DSeg. Maybe we can expand them a bit more to provide more contexts in the release note. Below is a draft of mine (cc @myron @dongyang0122 for visibility and comments)

```markdown
Added
- * Auto3DSeg AutoRunner added option self.algos to manually specify algorithms to run (instead of always downloading a zip)
- * Auto3DSeg AutoRunner added input options train, algo_gen, analyze will default to None, indicating to automatically decide whether this step is necessary based on cache (history).
- * Auto3DSeg AutoRunner added method set_analyze_params() to set DataAnalyser parameters (e.g. run it on gpu)
- * Auto3DSeg AutoRunner added a dry-run mode for hyper-parameter optimization(HPO) module so that users can get the commands to run NNI without actually using them.
(pending)
- * Auto3DSeg BundleGen added a new feature that automatically optimizes the parameter of model training configuration and maximize GPU memory utilization according to the devices in the house.
(also pending to merge because of the update of ALGO_HASH)
- * Auto3DSeg SegResNet is refracted to be more efficient, modular, with added support on minimal custom component (BYOC). Currently testing performance. The minimal user input config for the hecktor22 dataset would include a BYOC transform.

Fixed
- * Auto3DSeg AutoRunner will no longer fail if the users first set their ensemble method to AlgoEnsembleBestByFold and then change the num_fold by set_num_fold
- * Auto3DSeg DataAnalyzer now accepts "None" as the label key to indicate there is no label in the command line mode
- * Auto3DSeg DataAnalyzer will no longer crash when label image is empty (no foreground voxels are in this case, previously it crashed)
- * Auto3DSeg ensemble module can find the algorithm folders generated by the hyper-parameter optimization module.
- * Auto3DSeg ensemble module currently can use the voting method under sigmoid mode (one-hot encoding).
- * Auto3DSeg ensemble method AlgoEnsembleBestN will no longer fail if the number of trained algorithms is less than the pre-defined N best algorithms.


Changed
- * Auto3DSeg AutoRunner changed self.cache dict to be a single level (since it only has 4 items), instead of a nested dictionary
- * Auto3DSeg AutoRunner removed default setting sigmoid=True in set_prediction_params(), since it's somewhat arbitrary
- * Auto3DSeg AutoRunner will raise errors if previous steps are not completed (e.g. ensemble requires train step)
- * Auto3DSeg AutoRunner changed its defaults to run data analysis on GPU (previously it was CPU)
- * Auto3DSeg AutoRunner will save ensemble prediction labels in uint8 (unless specified; previously it was floats)
- * Auto3DSeg AutoRunner will print training progress in Console (previously no output was produced at all for a very long time)
- * Auto3DSeg AutoRunner will execute training in sorted order (fold0, fold1, fold2..), Previously it was in random order
- * Auto3DSeg AutoRunner now picks the best algorithm in each folder (AlgoEnsembleBestByFold) to ensemble algorithm by default, instead of finding the best N algorithms (AlgoEnsembleBestN)
- * Auto3DSeg DataAnalyzer uses GPU to improve speed by running data loading in parallel
Auto3DSeg DataAnalyzer can automatically fix shape mismatch (instead of crashing) when label is slightly different in size compared to image
- * Auto3DSeg DataAnalyzer saves summary stats on the top in datastats.yaml (previously it was at the very bottom)
- * Auto3DSeg DataAnalyzer's connected component stastical analysis switched its default to false, since it’s currently not used, and takes a long time to comput)

(also pending to merge because of the update of ALGO_HASH)
- * All default algorithm templates in Auto3DSeg are unified to use epochs to mark the progress of training and validation.
```

wyli:
thanks @mingxin-zheng, I'll update the relevant DataAnalyzer changes. for monai.apps.auto3dseg it's unstable anyway, I'll omit the details

Nic-Ma (approved):

PR #5732, opened 2022-12-14
 Change default dtype value in `Watershed` and `GenerateWatershedMarkers`

KumoLiu:
Fixes #.

### Description

- When the image is large, the number of instances may exceed np.uint8, so the default dtype should set to `np.int64`
- a workaround way to add `FillHoles` before `generate_instance_contour`, may prevent the issue when instances contain hole

### Types of changes


drbeh (approved):
Looks good to me. Thanks

Issue #5564, opened 2022-11-22
Autorunner AlgoEnsemble  improvements


myron:
we need to make a few fixes to autorunner and ensembler

 1. ~~BUG: AutoRunner().set_num_fold() does not update the enemble method num_folds.  So e.g. if the default method is AlgoEnsembleBestByFold (it'll be initialized with 5 folds by default), but when a user calls AutoRunner().set_num_fold(2), it only updates number of folds for training, and the ensembling will still look for 5 folds, which results in some algorithms being None.  I believe this error https://github.com/Project-MONAI/MONAI/pull/5523#issuecomment-1323319137 was due to this, and not only due to https://github.com/Project-MONAI/MONAI/issues/5558~~

2. ~~change the default ensemble method to AlgoEnsembleBestByFold in AutoRunner().set_ensemble_method()~~

2b. improve AlgoEnsembleBestByFold to return N best from each fold. Currently it returns only 1 best result, but we need it to return for  e.g. N=2, or N=3)

3. ~~VoteEnsemble() currently does not work for (sigmoid==True) mode~~

4. remove **pickling** of algorithms class instances, and re-initialize them in standard way

5.  ~~in class AlgoEnsembleBestN() line 196 , raises error if I select only 1 algorithm and 1 fold for training.  The default n_best is 2, since we trained only 1 fold/1alg, len(ranks)==1~~

6.  in the ensembling step we accumulate all results in memory, and then save them to disk. this can create OOM if RAM is small. I think we should save to disk each result as soon as we get it. https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/auto3dseg/ensemble_builder.py#L162

@mingxin-zheng

myron:
first issue is addressed in https://github.com/Project-MONAI/MONAI/pull/5667

mingxin-zheng:
Hi @myron , for the Item 3, I am wondering where the change should happen. Do you think it should be on `monai.transforms.VoteEnsemble`?

myron:
> `VoteEnsemble`

I AlgoEnemble.ensemble_pred()
```
            classes = [prob2class(p, dim=0, keepdim=True, sigmoid=False) for p in preds]
            return VoteEnsemble(num_classes=preds[0].shape[0])(classes)
```
it converts to classes fo sigmoid==False (hardcoded).  
 there should be an alternative for sigmoid==True

[...]

PR #5734, opened 2022-12-14
Remove hard-coded sigmoid for voting ensemble method in Auto3DSeg

mingxin-zheng:
Signed-off-by: Mingxin Zheng <18563433+mingxin-zheng@users.noreply.github.com>

Fixes item 3 and 5 in #5564.

### Description

The `vote` method in `ensemble_pred` currently does not work for under sigmoid mode, because the function overrides the argument to False before the `VoteEnsemble`.

Also, if the user only trains a small number of algorithm (1 fold for 1 algo) and forgets the update the `n_best` (default is 5) in `AlgoEnsembleBestN` , instead of throwing an error, the fix will automatically use all available algos after posting a warning.

### Types of changes

### DIFF
--- a/monai/apps/auto3dseg/ensemble_builder.py
+++ b/monai/apps/auto3dseg/ensemble_builder.py
@@ -197,7 +197,7 @@ class AlgoEnsembleBestN(AlgoEnsemble):
 
         ranks = self.sort_score()
         if len(ranks) < n_best:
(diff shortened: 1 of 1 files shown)


wyli:
is it ready? or you plan to add more updates? @mingxin-zheng

wyli (approved):

Issue #5740, opened 2022-12-14
MetricsReloaded support


brudfors:
The MetricsReloaded package under Project-MONAI (https://github.com/Project-MONAI/MetricsReloaded), developed by @csudre, is a recommendation framework for biomedical image analysis validation. As already discussed with @wyli, @Nic-Ma, @csudre and others; being able to use this framework in MONAI would be very valuable. The aim of this issue is to integrate MetricsReloaded support into MONAI.

wyli:
closing this by

- https://github.com/Project-MONAI/MONAI/pull/5741
- https://github.com/Project-MONAI/MONAI/pull/5920

the other packaging/demo tasks are tracked https://github.com/Project-MONAI/MONAI/milestone/91 with separate tickets.

PR #5741, opened 2022-12-14
5740 metrics reloaded support

brudfors:
Signed-off-by: Mikael Brudfors [mbrudfors@nvidia.com](mailto:mbrudfors@nvidia.com)

part of #5740

### Description

Makes the binary and categorical metrics from MetricsReloaded available in MONAI via a wrapper module (`monai/metrics/wrapper.py`). 

This module allows to use the MetricsReloaded metrics as, e.g.:
```py
import torch
from monai.metrics import MetricsReloadedBinary

metric_name = "Cohens Kappa"
metric = MetricsReloadedBinary(metric_name=metric_name)

# first iteration
# shape [batch=1, channel=1, 2, 2]
y_pred = torch.tensor([[[[1.0, 0.0], [0.0, 1.0]]]])
y = torch.tensor([[[[1.0, 0.0], [1.0, 1.0]]]])
print(metric(y_pred, y))

# second iteration
# shape [batch=1, channel=1, 2, 2]
y_pred = torch.tensor([[[[1.0, 0.0], [0.0, 0.0]]]])
y = torch.tensor([[[[1.0, 0.0], [1.0, 1.0]]]])
print(metric(y_pred, y))

# aggregate
# shape ([batch=2, channel=1])
print(metric.aggregate(reduction="none"))  # tensor([[0.5], [0.2]])

# reset
metric.reset()
```

Tests of all metrics are in `tests/test_metrics_reloaded.py`.

Note that MetricsReloaded is an optional dependency of MONAI; so in order to use the wrapper, MONAI needs to be installed with:
```sh
pip install '.[metricsreloaded]'
```

### Types of changes


monai/metrics/wrapper.py
```diff
@@ -0,0 +1,298 @@
...
+
+__all__ = ["MetricsReloadedBinary", "MetricsReloadedCategorical"]
+
+
+def torch2numpy(x):
+    return x.cpu().detach().numpy()
+
+
+def numpy2torch(x, device):
+    return torch.as_tensor(x, device=device)
```
wyli, line 34: there are generic utilities in monai.utils https://github.com/Project-MONAI/MONAI/blob/c10cdd226d94f06600a929de13c5d78bda6e8a10/monai/utils/type_conversion.py#L34-L35

wyli:
it looks great, any feedback and things we can improve for better integration? I think once we confirm `MetricsReloaded`'s API is stable, this PR can be merged

monai/metrics/wrapper.py
brudfors: Great, I changed to using those.

brudfors:
> it looks great, any feedback and things we can improve for better integration (from both monai and MetricsReloaded perspectives)? I think once we confirm `MetricsReloaded`'s API is stable, this PR can be merged

Sounds good!

This PR is limited to supporting the semantic segmentation metrics from MetricsReloaded and for that I cannot, for now, think of any improvements on the MONAI side. This might change when we integrate the instance segmentation and object detection metrics, later down the line. On the MetricsReloaded side, there are still some improvements that can be made, but I don't think they would require changing the API that MONAI uses.

requirements-dev.txt
```diff
@@ -52,3 +52,4 @@ pydicom
 h5py
 nni
 optuna
+git+https://github.com/Project-MONAI/MetricsReloaded@monai-support#egg=MetricsReloaded
```
Nic-Ma, line 55: @brudfors @wyli ,do we plan to release `MetricsReloaded` with versions? I think it's better to integrate with stable versions.

Thanks.

monai/metrics/wrapper.py
```diff
@@ -0,0 +1,290 @@
...
+            raise ValueError(f"Unsupported metric: {self.metric_name}")
+
+        # Compute metric
+        metric = bpm.metrics[self.metric_name]()
+
+        # Put back singleton channel dimension
+        metric = metric[..., None]
+
+        # Return metric as numpy array
+        return convert_to_tensor(metric, device=device)
```
Nic-Ma, line 290: The comment is "as numpy array" but the function is "to_tensor".

Thanks.

monai/metrics/wrapper.py
```diff
@@ -0,0 +1,290 @@
...
+MultiClassPairwiseMeasures, _ = optional_import(
+    "MetricsReloaded.metrics.pairwise_measures", name="MultiClassPairwiseMeasures"
+)
+
+__all__ = ["MetricsReloadedBinary", "MetricsReloadedCategorical"]
+
+
+class MetricsReloadedMetric(CumulativeIterationMetric):
+    """Base class for defining MetricsReloaded metrics as a
+    CumulativeIterationMetric in MONAI"""
```
Nic-Ma, line 31: Could you please help add doc-string for the args?

Thanks.

monai/metrics/wrapper.py
```diff
@@ -0,0 +1,290 @@
...
+
+    def __init__(
+        self,
+        metric_name: str,
```
Nic-Ma, line 35: Is there a way to search all the supported metric names? Better to put a link of metrics list or some Enum?

Thanks.

monai/metrics/wrapper.py
```diff
@@ -0,0 +1,290 @@
...
+
+BinaryPairwiseMeasures, _ = optional_import("MetricsReloaded.metrics.pairwise_measures", name="BinaryPairwiseMeasures")
+MultiClassPairwiseMeasures, _ = optional_import(
+    "MetricsReloaded.metrics.pairwise_measures", name="MultiClassPairwiseMeasures"
+)
+
+__all__ = ["MetricsReloadedBinary", "MetricsReloadedCategorical"]
+
+
+class MetricsReloadedMetric(CumulativeIterationMetric):
```
Nic-Ma, line 29: I would suggest class name to `MetricsReloadedWrapper`, what do you guys think?

Thanks.

brudfors: Good suggestion, renamed!

monai/metrics/wrapper.py
brudfors: Done!

monai/metrics/wrapper.py
brudfors: Fixed, thanks!

requirements-dev.txt
brudfors: I agree that pointing to a released version of MetricsReloaded would be prefered. I raised this with @csudre before the holiday break, maybe it is in the to-do list?

monai/metrics/wrapper.py
brudfors: How about adding two enums to `wrapper.py` with the binary and categorical metrics from MetricsReloaded that MONAI supports? Or should they be in `enums.py` maybe?

Nic-Ma: I would suggest to put the enums in the `MetricsReloaded` project and put a link here, because I suspect we may frequently add new metrics in `MetricsReloaded`, need to maintain the list out of MONAI?

Thanks.

brudfors: Okay, that sounds like a better idea, let's do that.

wyli (approved):
We'll add the ignite wrapper in a separate pr

PR #5757, opened 2022-12-15
1840 add tbf layer

### LINKED ISSUE #1840: Additional features for the BilateralFilter layer
For using the BilateralFilter layer (by @charliebudd ) more flexibly, I would really appreciate two extensions to the current filter implementation.

1. It would be nice to implement the backpropagation of a loss into the filter parameters (spatial and range sigma) to be able to learn them directly.
2. In addition, independent spatial filter kernel sizes (sigma) for different spatial dimensions would be very interesting to have (Gaussian filter kernel can have a different size in different dimensions).

faebstn96:
Fixes #1840 .

### Description

I integrated the trainable bilateral filter layer (TBF) in the MONAI repository as a new PyTorch filter layer. The TBF contains an analytical gradient derivation toward its filter parameters and its noisy input image which enables gradient-based optimization within the PyTorch graph. See [here](https://doi.org/10.1002/mp.15718) for more details on the gradient derivation. Unit tests were added that check the filter output as well as the gradient computation.

### Types of changes

### DIFF
--- a/monai/csrc/filtering/trainable_bilateral/bf_layer_gpu_backward.cu
+++ b/monai/csrc/filtering/trainable_bilateral/bf_layer_gpu_backward.cu
@@ -1,0 +1,272 @@
+/*
+Copyright (c) MONAI Consortium
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <cuda.h>
+#include <cuda_runtime.h>
+
+#include "trainable_bilateral.h"
+//#include "../utils/cuda_error_check.h"
+#include "utils/meta_macros.h"
+
+__constant__ int cBatchStrideBack;
+__constant__ int cColorStrideBack;
+
+__constant__ int cSizesBack[3];
+__constant__ int cStridesBack[3];
+
+__constant__ int cKernelSizesBack[3];
+__constant__ int cHalfWindowSize_arrBack[3];
+__constant__ float cGaussianKernel_xBack[256];
+__constant__ float cGaussianKernel_yBack[256];
+__constant__ float cGaussianKernel_zBack[256];
+__constant__ float cXDistanceSquaredBack[256];
+__constant__ float cYDistanceSquaredBack[256];
+__constant__ float cZDistanceSquaredBack[256];
+__constant__ float cColorExponentConstantBack;
+__constant__ float cSigma_xBack;
+__constant__ float cSigma_yBack;
+__constant__ float cSigma_zBack;
+__constant__ float cColorSigmaBack;
+
+
+template <typename scalar_t, int C>
+__global__ void BilateralFilterCudaKernel3DBackward(scalar_t* gradientInputTensor,
+                                                    scalar_t* gradientOutputTensor,
+                                                    scalar_t* inputTensor,
+                                                    scalar_t* outputTensor,
+                                                    scalar_t* outputWeightsTensor,
+                                                    scalar_t* dO_dx_ki) {
+  int homeOffset = blockIdx.x * blockDim.x + threadIdx.x;
+  int batchOffset = blockIdx.y * cBatchStrideBack;
+
+  if (homeOffset >= cColorStrideBack)
+    return;
+
+  int homeX = homeOffset / cStridesBack[0];
+  int homeY = (homeOffset - homeX * cStridesBack[0]) / cStridesBack[1];
+  int homeZ = (homeOffset - homeX * cStridesBack[0] - homeY * cStridesBack[1]) / cStridesBack[2];
+  int homeIndex[] = {homeX, homeY, homeZ};
+
+  // Zero kernel aggregates.
+  scalar_t valueSum = 0;
+  scalar_t weightSum = 0;
+
+  for (int kernelX = 0; kernelX < cKernelSizesBack[0]; kernelX++) {
+    int neighbourX = max(0, min(homeX + (kernelX - cHalfWindowSize_arrBack[0]), cSizesBack[0] - 1));
+    scalar_t gaussianX = cGaussianKernel_xBack[kernelX];
+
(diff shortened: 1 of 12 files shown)


wyli:
to compile and run the tests of this PR:

```
git clone https://github.com/faebstn96/MONAI.git
cd MONAI/
pip install -r requirements-min.txt
BUILD_MONAI=1 python setup.py develop
BUILD_MONAI=1 python -m tests.test_trainable_bilateral
```

monai/csrc/filtering/trainable_bilateral/bf_layer_gpu_forward.cu
```diff
@@ -0,0 +1,313 @@
...
+    yDistanceSquared[i] = distance * distance;
+  }
+  for (int i = 0; i < windowSize_z; i++) {
+    int distance = i - halfWindowSize_z;
+    gaussianKernel_z[i] = exp(distance * distance * spatialExpConstant_z);
+    zDistanceSquared[i] = distance * distance;
+  }
+
+  // Writing constant memory.
+  cudaMemcpyToSymbol(cBatchStride, &desc.batchStride, sizeof(int));
```
zehuanw, line 237: Scaler and array with known (compiler known e.g. 3 here) length will be passed through constant cache (const memory) by default. So if you want to leverage constant cache here, you can pass desc.batchStride directly to the CUDA kernel by function argument.

monai/csrc/filtering/trainable_bilateral/bf_layer_gpu_backward.cu
```diff
@@ -0,0 +1,282 @@
...
+                       inputTensor[batchOffset + homeOffset + c * cColorStrideBack] * colorDistance /
+                           (cColorSigmaBack * cColorSigmaBack)); // inputTensorData[homeOffset] !!
+            } else {
+              filter_kernel_back = dO_dx_ki[batchOffset + homeOffset + c * cColorStrideBack];
+            }
+
+            valueSum += gradientInputTensor[batchOffset + neighbourOffset + c * cColorStrideBack] * filter_kernel_back;
+          }
+
+          weightSum += totalWeight;
```
zehuanw, line 134: weightSum seems not used in backward path, can be removed?

zehuanw (commented):
Hi @faebstn96 , I read your code in the past two weeks and it looks great! Just left two comments directly on the code. 

One more question: It seems that most of the logics in the backprop is the same to forward (gaussianKernel_xyz are also the same?). Does it make sense to generate filter_kernel_back (in bf_layer_gpu_backward.cu) directly in forward and compute gradientOutputTensor with gradientInputTensor && the filter? If feasible, backward code will be much simpler.

monai/csrc/filtering/trainable_bilateral/bf_layer_gpu_forward.cu
faebstn96: Thank you so much for your comments and having a look at the pull request, @zehuanw!
I am not sure if I understand you right here. Do you really think this would improve performance if we pass the desc.batchStride via the kernel arguments? Usually it should be more efficient to keep it in the constant cache for repeated kernel calls, right? Here, it should not make much of a difference anyways because of the small size of desc.batchStride, right? But please correct me if I am wrong.

faebstn96:
Hi @zehuanw,
Thank you for looking into the code in such detail!

> One more question: It seems that most of the logics in the backprop is the same to forward (gaussianKernel_xyz are also the same?). Does it make sense to generate filter_kernel_back (in bf_layer_gpu_backward.cu) directly in forward and compute gradientOutputTensor with gradientInputTensor && the filter? If feasible, backward code will be much simpler.

You are right, the calculation of the spatial Kernels gaussianKernel_xyz is the same in the forward and backward and could be calculated in the forward only and then reused in the backward. However, the intensity range kernel needs to be calculated on the fly as the input intensities change locally (this is the more complex part of the calculation within the CUDA kernels).
I don't think that removing the spatial kernel computation from the backward pass is a change that would notably improve the performance of the code. Usually, the filter size is small in comparison to the filtered image and the spatial filter computation is only done once per filter backward. Please let me know what you think. :)

monai/csrc/filtering/trainable_bilateral/bf_layer_gpu_backward.cu
faebstn96: Hi @zehuanw,
Unfortunately, I forgot to sign my commit. However, adding a remediation commit did not work as the DCO still complains. Also, rebasing and searching on Google did not work. Can you help me with signing the commits? Sorry, I am not super familiar with signed commits/DCO. Thank you for your help!

wyli: Hi @faebstn96 that's fine I'll handle it when merging this PR.

monai/csrc/filtering/trainable_bilateral/bf_layer_gpu_forward.cu
zehuanw: > Do you really think this would improve performance if we pass the desc.batchStride via the kernel arguments? 

It may not improve the performance. I meant that it will do no harm to your performance but just keep the code cleaner if you pass desc.batchStride directly to kernel with arguments. It's not a big problem anyway, so it's just FYI and please be free to adopt this idea or not.

>Usually it should be more efficient to keep it in the constant cache for repeated kernel calls, right?
Constant Cache is a on Chip cache (like shared memory) and you cannot keep something in it across different kernel launch (but L2 can). What you can do is copy data into Constant memory (on GPU memory) once and reuse in different kernels (cache will be cleared after kernel end). From the code it looks like you will call `cudaMemcpyToSymbol` for every forward and each forward has one kernel launch in this layer. So it looks like even we cannot save the memory copy here and reuse in different forwards with the same copy of constants？

faebstn96: Okay, I see. Thank you for your explanation! You are right, there is only one kernel launch per forward/backward. I tried to use the shared memory the same way it was used in the [existing bilateral filter implementation](https://github.com/Project-MONAI/MONAI/blob/734d4ffc521adf3359b22b7a38c4eb7d3966c223/monai/csrc/filtering/bilateral/bilateralfilter_cuda.cu#216). If you are fine with it then I would keep the current implementation. But thank you again for pointing that out, I did not really think about the memory access here!

wyli (approved):
thanks @faebstn96 and the comments from @zehuanw, the PR looks nice and I'm merging it soon. the fact that the variable notations are consistent with the descriptions in the paper makes the code logic easy to follow...

PR #5787, opened 2022-12-26
5775 Fix `_get_latest_bundle_version` issue on Windows

### LINKED ISSUE #5775: `_get_latest_bundle_version` issue on Windows
https://github.com/Project-MONAI/MONAI/blob/69463a69637d4383056455b1b12bdff58294363e/monai/bundle/scripts.py#L185

On Windows, `os.path.join` will produce backslash rather than slash. Therefore, the created url will be wrong (`os.path.join(repo_owner, repo_name)`)
May need to handle this to avoid issues for Windows users.

yiheng-wang-nv:

Fixes #5775 .

### Description

This PR fixes the issue of `_get_latest_bundle_version` on Windows (`os.path.join` will produce backslash which will create a wrong url) 
Thanks @SachidanandAlle for finding this issue.

### Types of changes

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -182,7 +182,7 @@ def _get_latest_bundle_version(source: str, name: str, repo: str):
         return None
     elif source == "github":
         repo_owner, repo_name, tag_name = repo.split("/")
-        return get_bundle_versions(name, repo=os.path.join(repo_owner, repo_name), tag=tag_name)["latest_version"]
+        return get_bundle_versions(name, repo=f"{repo_owner}/{repo_name}", tag=tag_name)["latest_version"]
     else:
         raise ValueError(f"To get the latest bundle version, source should be 'github' or 'ngc', got {source}.")
 
--- a/tests/test_bundle_download.py
+++ b/tests/test_bundle_download.py
@@ -58,7 +58,6 @@ TEST_CASE_5 = [
 ]
 
 
-@skip_if_windows
 class TestDownload(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


SachidanandAlle (approved):

Nic-Ma (approved):
Hi @yiheng-wang-nv ,

Could you please help remove this line to allow the windows tests:
https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_bundle_download.py#L61
Actually, other bundle tests cover windows, like:
https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_integration_bundle_run.py#L108

Thanks.

yiheng-wang-nv:
> Hi @yiheng-wang-nv ,
> 
> Could you please help remove this line to allow the windows tests: https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_bundle_download.py#L61 Actually, other bundle tests cover windows, like: https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_integration_bundle_run.py#L108
> 
> Thanks.

Thanks Nic for the advice. I removed the line you mentioned. However, here I wish to double confirm that if `windows` and `macOS` will execute the unit tests that has `skip_if_quick`. Initially I followed `test_mmar_download` and avoid download and load tests if quick. I'm not sure if it is necessary to remove this decorator as well to ensure`windows` test is able to run.

Hi @wyli , could you please help to give some suggestions? Thanks!

wyli:
there's only quick tests for windows/macos, because we don't have windows+gpu full tests anywhere, so please verify this manually.

PR #5813, opened 2023-01-05
5804 Simplify the configparser usage to get parsed attributes easily

### LINKED ISSUE #5804: simplify the configparser usage to get parsed attributes easily

the current usage looks tedious (https://github.com/Project-MONAI/research-contributions/blob/07da7ad36a0b218556a427f3bc5292a772c30f60/auto3dseg/algorithm_templates/segresnet2d/scripts/infer.py#L34-L50)

```py
parser = ConfigParser()
parser.read_config(config_file_)
parser.update(pairs=_args)

data_file_base_dir = parser.get_parsed_content("data_file_base_dir")
data_list_file_path = parser.get_parsed_content("data_list_file_path")
self.fast = parser.get_parsed_content("infer")["fast"]
self.num_adjacent_slices = parser.get_parsed_content("num_adjacent_slices")
self.num_sw_batch_size = parser.get_parsed_content("num_sw_batch_size")
self.output_classes = parser.get_parsed_content("output_classes")
self.overlap_ratio = parser.get_parsed_content("overlap_ratio")
self.patch_size_valid = parser.get_parsed_content("patch_size_valid")
softmax = parser.get_parsed_content("softmax")

ckpt_name = parser.get_parsed_content("infer")["ckpt_name"]
data_list_key = parser.get_parsed_content("infer")["data_list_key"]
output_path = parser.get_parsed_content("infer")["ouptut_path"]
```

this is a feature request of simplifying the APIs to support more user-friendly calls
to get the parsed config as attributes directly:

```py
cfg = ConfigParser()
cfg.read_config(config_file_)
cfg.update(pairs=_args)

data_file_base_dir = cfg.data_file_base_dir
data_list_file_path = cfg.data_list_file_path
self.fast = cfg.infer.fast
...
```

cc @Project-MONAI/core-reviewers

Shadow-Devil:
part of #5804.

### Description

This adds a simple attribute access for ConfigParser so

```python
parser = ConfigParser({"a":"b"})
a = parser.get_parsed_content("a")
```
can be rewritten to
```python
parser = ConfigParser({"a":"b"})
a = parser.a
```
This only works for variables/methods that are not included in ConfigParser so e.g. `parser.config` would still get the whole config.
This PR only supports shallow attribute accesses, since the returned value will be a `dict` where you need to use `["key"]` to get the content.

### Types of changes

### DIFF
--- a/docs/source/bundle.rst
+++ b/docs/source/bundle.rst
@@ -32,6 +32,7 @@ Model Bundle
 ---------------
 .. autoclass:: ConfigParser
     :members:
+    :special-members:
 
 `Scripts`
 ---------
--- a/monai/bundle/config_parser.py
+++ b/monai/bundle/config_parser.py
@@ -118,11 +118,14 @@ class ConfigParser:
 
     def __getattr__(self, id):
         """
-        Get the config by id.
+        Get the parsed result of ``ConfigItem`` with the specified ``id``
+        with default arguments (e.g. ``lazy=True``, ``instantiate=True`` and ``eval_expr=True``).
 
         Args:
-            id: id to specify the expected position. See also :py:meth:`__getitem__`.
-
+            id: id of the ``ConfigItem``.
+        
+        See also:
+             :py:meth:`get_parsed_content`
         """
         return self.get_parsed_content(id)
 
--- a/monai/bundle/config_parser.py
+++ b/monai/bundle/config_parser.py
@@ -116,6 +116,16 @@ class ConfigParser:
     def __repr__(self):
         return f"{self.config}"
 
+    def __getattr__(self, id):
+        """
+        Get the config by id.
+
+        Args:
+            id: id to specify the expected position. See also :py:meth:`__getitem__`.
+
+        """
+        return self[id]
+
     def __getitem__(self, id: Union[str, int]):
         """
         Get the config by id.
--- a/monai/bundle/config_parser.py
+++ b/monai/bundle/config_parser.py
@@ -123,7 +123,7 @@ class ConfigParser:
 
         Args:
             id: id of the ``ConfigItem``.
-        
+
         See also:
              :py:meth:`get_parsed_content`
         """
--- a/monai/bundle/config_parser.py
+++ b/monai/bundle/config_parser.py
@@ -124,7 +124,7 @@ class ConfigParser:
             id: id to specify the expected position. See also :py:meth:`__getitem__`.
 
         """
-        return self[id]
+        return self.get_parsed_content(id)
 
     def __getitem__(self, id: Union[str, int]):
         """
--- a/tests/test_auto3dseg_ensemble.py
+++ b/tests/test_auto3dseg_ensemble.py
@@ -139,7 +139,7 @@ class TestEnsembleBuilder(unittest.TestCase):
         builder = AlgoEnsembleBuilder(history, data_src_cfg)
         builder.set_ensemble_method(AlgoEnsembleBestN(n_best=1))
         ensemble = builder.get_ensemble()
-        pred_param["network#init_filter"] = 8  # segresnet
+        pred_param["network#init_filters"] = 8  # segresnet
         preds = ensemble(pred_param)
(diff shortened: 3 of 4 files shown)


monai/bundle/config_parser.py
```diff
@@ -116,6 +116,16 @@ def __init__(
...
 
+    def __getattr__(self, id):
+        """
+        Get the config by id.
+
+        Args:
+            id: id to specify the expected position. See also :py:meth:`__getitem__`.
+
+        """
+        return self[id]
```
wyli, line 127: have you tried returning `self.get_parsed_content("id")` here? maybe it's better to return the parsed content cc @Nic-Ma

Nic-Ma: I agree with @wyli , this implementation doesn't match the initial feature request, we already have `dict` style `get / set` APIs to handle the config "text", this feature request is to **simplify the API of getting "parsed content"**, for example the instantiated objects ('dataloader', 'network', etc.) and also want to support nested items if the "parsed content" is still a `dict`.
Please check this recursive function for more details:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/config_parser.py#L359
Sorry for the confusion.

Thanks.

Shadow-Devil: Sorry, I fixed it, so it will now return the parsed content. It currently still only works one layer deep, so no nested items.

wyli (approved):
Thanks, it looks good to me.

Shadow-Devil:
One Idea was to change the `get_config` of `ConfigItem` like this:
```python
    def get_config(self) -> Any:
        """
        Get the config content of current config item.
        If the content is a dict it will wrap it with `DictWrapper` to allow attribute like access.

        """
        
        return self.config if not isinstance(self.config, dict) else DictWrapper(self.config)

class DictWrapper(UserDict):
    """
    Wrapper class to support attribute-style access to the dict items.
    """

    def __getattr__(self, key: str) -> Any:
        if key in self.data:
            item = self.data[key]
            return item if not isinstance(item, dict) else DictWrapper(item)
        raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{key}'")
```
This implementation wraps each `dict` into this `DictWrapper` which will in itself wrap each `dict` again into a `DictWrapper`.

Shadow-Devil:
The problem with this implementation is, that nested content is not parsed/instantiated...

wyli:
I'll merge this PR for now... for the nested cases, perhaps we'll introduce a new class for `ConfigParser.config`? do you plan to continue working on this @Shadow-Devil? otherwise @Nic-Ma and I can take over and enhance. we are fine with either way.

Shadow-Devil:
Thanks @wyli, I think it would be better if you continue since I'm not sure how this new class should look like and there are some design decisions to make, which should be in line with the rest of the project. If you have a general plan on how this should look like, I'm happy to implement it :)

monai/bundle/config_parser.py
```diff
@@ -116,6 +116,16 @@ def __init__(
     def __repr__(self):
         return f"{self.config}"
 
+    def __getattr__(self, id):
+        """
+        Get the config by id.
+
+        Args:
+            id: id to specify the expected position. See also :py:meth:`__getitem__`.
```
Nic-Ma, line 124: Hi @Shadow-Devil ,

I think the doc-string should also be updated?
It's not `__getitem__ `.

Thanks.

Shadow-Devil: sorry, I forgot

docs/source/bundle.rst
```diff
@@ -32,6 +32,7 @@ Model Bundle
 ---------------
 .. autoclass:: ConfigParser
     :members:
+    :special-members:
```
Nic-Ma, line 35: May I know why add this line?

Thanks.

wyli: special member functions are not in the autogenerated doc by default https://docs.monai.io/en/stable/bundle.html#monai.bundle.ConfigParser

PR #5824, opened 2023-01-08
MyPy disallow untyped decorators

### LINKED ISSUE #5823: MyPy disallow untyped decorators
There are some decorators that are untyped e.g. `deprecated` and `deprecated_arg`.
See also here: https://github.com/Project-MONAI/MONAI/blob/dev/monai/utils/deprecate_utils.py
These untyped decorators break the type checking of MyPy, so any function annotated with these decorators is not type checked at all.

Add a type signature to these decorators, so the decorated functions are type checked.
First Proposal:
```python
T = TypeVar('T', type, Callable)
def deprecated(
    since: Optional[str] = None,
    removed: Optional[str] = None,
    msg_suffix: str = "",
    version_val: str = __version__,
    warning_category=FutureWarning,
) -> Callable[[T], T]:
...
```
Another idea would be to set the return type to `Callable[[T], Union[T, NoReturn]]` since in some circumstances the decorator throws an exception and doesn't even reach the decorated code.
Also, enable `disallow_untyped_decorators = True` inside `setup.cfg` so we can be certain that all decorators are typed.

**Additional context**
MyPy documentation: https://mypy.readthedocs.io/en/stable/generics.html?highlight=decorator#declaring-decorators

Shadow-Devil:
Fixes #5823.

### Description
This PR types following decorators:
- `deprecated`
- `deprecated_arg`
- `deprecated_arg_default`

Found [here](https://github.com/Project-MONAI/MONAI/blob/dev/monai/utils/deprecate_utils.py). 
It also fixes any typing issues that occurred because of the additional typing.

Secondly, it fixes untyped decorator usage of the `ignite` library (because of `optional_import`) for [Workflow](https://github.com/Project-MONAI/MONAI/blob/34d713f9887a85f140630a75e9261b89f0005c84/monai/engines/workflow.py#L45) and [IgniteMetric](https://github.com/Project-MONAI/MONAI/blob/f9d472af05ceb56513771ea2faccebb79edc1a7a/monai/handlers/ignite_metric.py#L36). 

### Types of changes

### DIFF
--- a/monai/apps/deepedit/interaction.py
+++ b/monai/apps/deepedit/interaction.py
@@ -96,4 +96,4 @@ class Interaction:
 
         # first item in batch only
         engine.state.batch = batchdata
-        return engine._iteration(engine, batchdata)
+        return engine._iteration(engine, batchdata)  # type: ignore[arg-type]
--- a/monai/apps/deepgrow/interaction.py
+++ b/monai/apps/deepgrow/interaction.py
@@ -85,4 +85,4 @@ class Interaction:
             # collate list into a batch for next round interaction
             batchdata = list_data_collate(batchdata_list)
 
-        return engine._iteration(engine, batchdata)
+        return engine._iteration(engine, batchdata)  # type: ignore[arg-type]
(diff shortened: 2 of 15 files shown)


wyli (approved):
Thanks, it looks good to me.

PR #5836, opened 2023-01-09
Consistent usage of annotation syntax

### LINKED ISSUE #5835: Consistent usage of annotation syntax
There are already some files using the new syntax for annotations. 
https://github.com/Project-MONAI/MONAI/search?q=from+__future__+import+annotations
The old type aliases from the `typing` module are deprecated starting with python 3.9,
 so to be future-prove it makes sense to migrate to the new style.

Use `from __future__ import annotations` in all source files using annotations, so we have consistent usage (and not only in ~5 files).
The new annotations are also more readable e.g. `Optional[Union[List[str], str]]` vs. `list[str] | str | None`.

Secondly, this issue includes changing `from typing import Callable, Sequence, ...` to `from collections.abc import Callable, Sequence, ...` since the ones in the `typing` module are deprecated. They are interchangable even for `isinstance` checks (see also [this stackoverflow thread](https://stackoverflow.com/questions/62547848/should-isinstance-check-against-typing-or-collections-abc)).

**Additional context**
- [PEP 585](https://peps.python.org/pep-0585/)
- [Documentation to the typing module](https://docs.python.org/3.10/library/typing.html)
- Old discussion about this topic: https://github.com/Project-MONAI/MONAI/pull/4150#issuecomment-1104115319

Shadow-Devil:
Fixes #5835.

### Description
Use `from __future__ import annotations` in all source files using annotations, so we have consistent usage (and not only in ~5 files).
The new annotations are also more readable e.g. `Optional[Union[List[str], str]]` vs. `list[str] | str | None`.

Secondly, this issue includes changing `from typing import Callable, Sequence, ...` to `from collections.abc import Callable, Sequence, ...` since the ones in the `typing` module are deprecated. They are interchangeable even for `isinstance` checks (see also [this stackoverflow thread](https://stackoverflow.com/questions/62547848/should-isinstance-check-against-typing-or-collections-abc)).

### Types of changes


Shadow-Devil:
There are some special cases, since these subscriptable types are only allowed as annotations but not inside casts. Therefore, `cast(Iterable[str], x)` fails with `collections.abc.Iterable` but works for `typing.Iterable`. For those cases I change it back to `typing.Iterable`.

wyli:
thanks, could you please also review the pyupgrade configs
https://github.com/Project-MONAI/MONAI/blob/dev/.pre-commit-config.yaml#L33
maybe there are some options that could be enabled after this PR
https://github.com/asottile/pyupgrade#pep-585-typing-rewrites

drbeh:
> There are some special cases, since these subscriptable types are only allowed as annotations but not inside casts. Therefore, `cast(Iterable[str], x)` fails with `collections.abc.Iterable` but works for `typing.Iterable`. For those cases I change it back to `typing.Iterable`.

I see! I believe it is better to ignore those typings instead of keeping a deprecated type.

Shadow-Devil (commented):
I think we still need the deprecated types here, since the new ones only work inside annotations, but this is creating a type alias.
https://github.com/Shadow-Devil/MONAI/blob/9f4c8059491af22c7d89f02f8d8ed804206868e3/monai/transforms/spatial/array.py#L111

Also here: https://github.com/Shadow-Devil/MONAI/blob/feature_use_new_annotation_syntax/monai/config/type_definitions.py

monai/transforms/spatial/array.py
```diff
@@ -12,11 +12,15 @@
...
+
+from __future__ import annotations
+
 import warnings
+from collections.abc import Callable
 from copy import deepcopy
 from enum import Enum
 from itertools import zip_longest
-from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union, cast
+from typing import Any, Optional, Sequence, Tuple, Union, cast
```
Shadow-Devil, line 23: I think we still need the deprecated types here, since the new ones only work inside annotations, but this is creating a type alias.
```python
RandRange = Optional[Union[Sequence[Union[Tuple[float, float], float]], float]]
```

wyli:
looks like it's feasible to ensure the `"from __future__ import annotations"` via isort config, https://github.com/Project-MONAI/MONAI/blob/6060a4731f087d9947236f312a388cd65f7ff8f4/setup.cfg#L157

could you please add this:
```
add_imports = ["from __future__ import annotations"]
append_only = true
```

wyli:
thanks, I think it's good as long as the tools are configured to always enforce the style consistently

wyli (approved):

drbeh:
> I think we still need the deprecated types here, since the new ones only work inside annotations, but this is creating a type alias. https://github.com/Shadow-Devil/MONAI/blob/9f4c8059491af22c7d89f02f8d8ed804206868e3/monai/transforms/spatial/array.py#L111
> 
> Also here: https://github.com/Shadow-Devil/MONAI/blob/feature_use_new_annotation_syntax/monai/config/type_definitions.py

We have until 2025 to update this!😉 This behavior is accepted in 3.10 where we can define `numeric = int | float` for instance. Maybe by 2025 we can drop support for <3.10.

PR #5845, opened 2023-01-11
5844 improve metatensor slicing

### LINKED ISSUE #5844: Slicing a MetaTensor: `TypeError: only integer tensors of a single element can be converted to an index`
The goal is to slice the tensor in half. For some reason that I cannot comprehend, it doesn't work on a particular tensor:

```python
>>> type(x)
<class 'monai.data.meta_tensor.MetaTensor'>

>>> x.shape
torch.Size([128, 1024, 8, 64])

>>> x[slice(0, 64, None), ...]
*** TypeError: only integer tensors of a single element can be converted to an index

# OK with [0:16]:
>>> x[slice(0, 16, None), ...].shape
torch.Size([16, 1024, 8, 64])

# NOT OK with [0:32]:
>>> x[slice(0, 32, None), ...]
*** TypeError: only integer tensors of a single element can be converted to an index

>>> x[0:16].shape
torch.Size([16, 1024, 8, 64])

# ARGGGGGHHHH
>>> x[0:32].shape
*** TypeError: only integer tensors of a single element can be converted to an index

>>> x[0:17].shape
*** TypeError: only integer tensors of a single element can be converted to an index

# What if we convert to Numpy? (result: it works fine)
>>> x.detach().numpy()[0:64].shape
(64, 1024, 8, 64)

```

What could be going on here?

If it is of any use, the error occurs at this specific line: https://github.com/pashtari/factorizer/blob/fa7f03dfe8c390f0993cbc0d161922d2de819a1d/factorizer/factorization/operations.py#L331

**To Reproduce**

```python
# This works fine, I cannot reproduce in this way
>>> from monai.data import MetaTensor
>>> x = MetaTensor(np.random.rand(128, 1024, 8, 64))
>>> x[slice(0, 64, None), ...].shape
torch.Size([128, 1024, 8, 64])
```

How is it possible that a specific tensor would cause such error?

**Environment**

```
Ubuntu 20.04
python 3.10

monai                     1.1.0              pyhd8ed1ab_0    conda-forge

pytorch                   1.13.1          py3.10_cuda11.7_cudnn8.5.0_0    pytorch
pytorch-cuda              11.7                 h67b0de4_1    pytorch
pytorch-lightning         1.8.1              pyhd8ed1ab_0    conda-forge
pytorch-mutex             1.0                        cuda    pytorch
torchmetrics              0.11.0             pyhd8ed1ab_0    conda-forge
torchvision               0.14.1              py310_cu117    pytorch
```

wyli:

Fixes #5844

- improves the error message when slicing data with inconsistent metadata
- fixes an indexing case:
```
x = MetaTensor(np.zeros((10, 3, 4)))
x[slice(1, 0)]  # should return zero length data
```

### Types of changes


wyli:
could you please review @Nic-Ma?

monai/data/meta_tensor.py
```diff
@@ -235,11 +235,18 @@ def update_meta(rets: Sequence, func, args, kwargs) -> Sequence:
...
                             ret_meta = decollate_batch(args[0], detach=False)[batch_idx]
-                            if isinstance(ret_meta, list):  # e.g. batch[0:2], re-collate
-                                ret_meta = list_data_collate(ret_meta)
-                            else:  # e.g. `batch[0]` or `batch[0, 1]`, batch index is an integer
+                            if isinstance(ret_meta, list) and ret_meta:  # e.g. batch[0:2], re-collate
+                                try:
+                                    ret_meta = list_data_collate(ret_meta)
+                                except (TypeError, ValueError, RuntimeError, IndexError) as e:
+                                    raise ValueError(
+                                        "Inconsistent metadata when slicing a MetaTensor, please convert it into"
```
Nic-Ma, line 243: I feel the error message is not very straightforward to me, I read it 3 times to get the info..
Could you please help enhance it? Mainly about "what is inconsistent".

Thanks in advance.

Nic-Ma (approved):

PR #5860, opened 2023-01-17
4855 lazy resampling impl -- Compose

wyli:
part of https://github.com/Project-MONAI/MONAI/issues/4855

upgrade https://github.com/Project-MONAI/MONAI/pull/4911 to use the latest dev API

### Description
e.g., for a sequence of spatial transforms

```py
xforms = [
    mt.LoadImageD(keys, ensure_channel_first=True),
    mt.Orientationd(keys, "RAS"),
    mt.SpacingD(keys, (1.5, 1.5, 1.5)),
    mt.CenterScaleCropD(keys, roi_scale=0.9),
    # mt.CropForegroundD(keys, source_key="seg", k_divisible=5),
    mt.RandRotateD(keys, prob=1.0, range_y=np.pi / 2, range_x=np.pi / 3),
    mt.RandSpatialCropD(keys, roi_size=(76, 87, 73)),
    mt.RandScaleCropD(keys, roi_scale=0.9),
    mt.Resized(keys, (30, 40, 60)),
    # mt.NormalizeIntensityd(keys),
    mt.ZoomD(keys, 1.3, keep_size=False),
    mt.FlipD(keys),
    mt.Rotate90D(keys),
    mt.RandAffined(keys),
    mt.ResizeWithPadOrCropd(keys, spatial_size=(32, 43, 54)),
    mt.DivisiblePadD(keys, k=3),
]
lazy_kwargs = dict(mode=("bilinear", 0), padding_mode=("border", "nearest"), dtype=(torch.float32, torch.uint8))
xform = mt.Compose(xforms, lazy_evaluation=True, overrides=lazy_kwargs, override_keys=keys)
xform.set_random_state(0)
```
lazy_evaluation=True preserves more details
Screenshot 2023-01-17 at 00 31 40
compared with the regular compose
Screenshot 2023-01-17 at 00 31 43

### Types of changes


monai/data/meta_tensor.py
```diff
@@ -479,10 +479,19 @@ def peek_pending_shape(self):
...
+            if next_matrix is None:
+                continue
+            res = convert_to_dst_type(res, next_matrix)[0]
+            res = monai.transforms.lazy.utils.combine_transforms(res, next_matrix)
+        return res
+
+    def peek_pending_rank(self):
+        a = self.pending_operations[-1].get(LazyAttr.AFFINE, None) if self.pending_operations else self.affine
+        r = max(1, len(a) - 1)
+        return convert_to_dst_type(r, self.affine)[0]
```
KumoLiu, line 494: Do we have a special reason to convert this rank to "dst type"( maybe tensor) which may cause an error when we use this rank?

monai/transforms/croppad/functional.py
```diff
@@ -0,0 +1,84 @@
...
+    extra_info = {"padded": to_pad_}
+    img_size = img.peek_pending_shape() if isinstance(img, MetaTensor) else img.shape[1:]
+    spatial_rank = img.peek_pending_rank() if isinstance(img, MetaTensor) else torch.tensor(3.0, dtype=torch.double)
+    do_pad = np.asarray(to_pad_).any()
+    if do_pad:
+        to_pad_ = list(to_pad_)
+        if len(to_pad_) < len(img.shape):
+            to_pad_ = list(to_pad_) + [(0, 0)] * (len(img.shape) - len(to_pad_))
+        to_shift = [-s[0] for s in to_pad_[1:]]  # skipping the channel pad
+        xform = convert_to_dst_type(create_translate(spatial_rank, to_shift), spatial_rank)[0]
```
KumoLiu, line 41: As mentioned before, the `spatial_rank` here needs to be int.
https://github.com/Project-MONAI/MONAI/blob/f4902b220538f02f3621b37c3e1f2067c830e0fe/monai/transforms/utils.py#L860-L861

monai/data/meta_tensor.py
wyli: this is a good question, currently the affine can be float32/float64, so we need `peek_pending_affine_dtype`. in practice I think it's more convenient to merge `peek_pending_affine_dtype` into `peek_pending_rank`, because both functions are mostly used together...

thinking about this again, maybe we should always use float64 for the affine matrix, float32 is not precise enough in most cases. I'll update this PR to use float64 affine.

monai/transforms/croppad/functional.py
```diff
@@ -0,0 +1,84 @@
...
+        affine=convert_to_dst_type(create_translate(spatial_rank, to_shift), spatial_rank)[0],
+        extra_info=extra_info,
+        orig_size=img_size,
+        transform_info=transform_info,
+        lazy_evaluation=transform_info.get(TraceKeys.LAZY_EVALUATION, False),
+    )
+    out = convert_to_tensor(img.as_tensor() if isinstance(img, MetaTensor) else img, track_meta=get_track_meta())
+    if transform_info.get(TraceKeys.LAZY_EVALUATION, False):
+        return out.copy_meta_from(meta_info) if isinstance(out, MetaTensor) else meta_info
+    out = out[slices]
```
wyli, line 83: if `out.pending_operations` is not empty at this line (doing the non-lazy transform), it should raise an error or do `out = moani.transforms.lazy.functional.apply_transforms(out)`. But we don't know how to choose the mode/padding mode parameters for the `apply_transforms`. So maybe raising an error is better..

monai/data/meta_tensor.py
wyli: added a new commit to use double for the affine https://github.com/Project-MONAI/MONAI/pull/5860/commits/4a46fda6b44a6ccdb200d2cb94cb3ed50f30667b

monai/transforms/utils.py
```diff
@@ -1676,14 +1677,14 @@ def scale_affine(affine, spatial_size, new_spatial_size, centered: bool = True):
         Scaled affine matrix.
 
     """
+    r = int(spatial_rank)
     if spatial_size == new_spatial_size:
-        return affine
-    r = len(affine) - 1
+        return np.eye(r + 1)
```
yiheng-wang-nv, line 1682: here should we add a check to ensure the value `spatial_rank ` matches `spatial_size` and `new_spatial_size`?

wyli: It seems spatial rank is redundant... We can infer it from len(new_spatial_size), let's remove the spatial_rank arg

KumoLiu: Another question here, if we don't pass `affine` in this function, does it mean we only record something like "affine_scale_ratio" in the `TraceableTransform.track_transform_meta`? Am I understand correctly?

wyli: Correct, the result is then combined with the original affine in https://github.com/Project-MONAI/MONAI/blob/59f42f234bf04e8d001fb77d8813e7efd23d9bd9/monai/transforms/inverse.py#L182

wyli: removed `spatial_rank` from `scale_affine`, now it's computed from `r = max(len(new_spatial_size), len(spatial_size))`, so that when it's 2d->3d or 3d->2d can still work https://github.com/Project-MONAI/MONAI/pull/5860/commits/1957962ff1edf480f6d8692224c7a576cdabee27

tests/test_zoom.py
```diff
@@ -14,19 +14,44 @@
...
+        # non-lazy
+        expected = zoom_fn(im)
+        self.assertIsInstance(expected, MetaTensor)
+        # lazy
+        zoom_fn.lazy_evaluation = True
+        pending_result = zoom_fn(im)
+        self.assertIsInstance(pending_result, MetaTensor)
+        assert_allclose(pending_result.peek_pending_affine(), expected.affine)
+        assert_allclose(pending_result.peek_pending_shape(), expected.shape[1:])
+        result = apply_transforms(pending_result, mode="bilinear", dtype=np.float64, align_corners=True)[0]
```
KumoLiu, line 51: A little confused here, does it mean we only support {``"bilinear"``, ``"nearest"``} in mode and {``"zeros"``, ``"border"``, ``"reflection"``} in padding mode when we doing lazy resampling?
https://github.com/Project-MONAI/MONAI/blob/bf55f22c461bac286e3078a242d9d0b90a43ba00/monai/transforms/lazy/functional.py#L46-L52

wyli: We support all the modes that pytorch grid_sample and scipy map_coordinates support. But the underlying implementations are different, we can't guarantee exactly the same results using these different backends/precisions/devices. This test is only to check that we are using the resampling APIs correctly.

monai/transforms/spatial/array.py
```diff
@@ -396,11 +285,16 @@ def __call__(  # type: ignore
             align_corners=align_corners,
             dtype=dtype,
         )
-        self.update_meta(img, dst_affine=dst_affine, img_dst=img_dst)
+        if isinstance(img, MetaTensor):
+            img.affine = dst_affine
+            if isinstance(img_dst, MetaTensor):
+                original_fname = img.meta.get(Key.FILENAME_OR_OBJ, "resampled_to_match_source")
```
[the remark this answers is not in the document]
wyli: As an output filename, it suggests that the image is an output of the transform. But I don't have a preference here..both look ok.

monai/transforms/spatial/functional.py
```diff
@@ -0,0 +1,405 @@
...
+nib, has_nib = optional_import("nibabel")
+cupy, _ = optional_import("cupy")
+cupy_ndi, _ = optional_import("cupyx.scipy.ndimage")
+np_ndi, _ = optional_import("scipy.ndimage")
+
+__all__ = ["spatial_resample", "orientation", "flip", "resize", "rotate", "zoom", "rotate90", "affine_func"]
+
+
+def spatial_resample(
+    img, dst_affine, spatial_size, mode, padding_mode, align_corners, dtype_pt, transform_info
```
yiheng-wang-nv, line 58: missing doc strings. The original call function has them all thus I will add them in my PR

monai/transforms/croppad/functional.py
```diff
@@ -0,0 +1,83 @@
...
+    out = convert_to_tensor(out, track_meta=get_track_meta())
+    return out.copy_meta_from(meta_info) if isinstance(out, MetaTensor) else out
+
+
+def crop_func(img, slices, transform_info):
+    img_size = img.peek_pending_shape() if isinstance(img, MetaTensor) else img.shape[1:]
+    spatial_rank = img.peek_pending_rank() if isinstance(img, MetaTensor) else torch.tensor(3.0, dtype=torch.double)
+    cropped = np.asarray([[s.indices(o)[0], o - s.indices(o)[1]] for s, o in zip(slices[1:], img_size)])
+    extra_info = {"cropped": cropped.flatten().tolist()}
+    to_shift = [s.start if s.start is not None else 0 for s in ensure_tuple(slices)[1:]]
```
KumoLiu, line 68: In the use of slice, users could specify a negative value so here would generate a wrong shift value. Referring to https://docs.python.org/2/library/stdtypes.html#sequence-types-str-unicode-list-tuple-bytearray-buffer-xrange.

```suggestion
    to_shift = []
    for i, s in enumerate(ensure_tuple(slices)[1:]):
        if s is not None:
            if s.start < 0:
                to_shift.append(img_size[i] + s.start)
            else:
                to_shift.append(s.start)
        else:
            to_shift.append(0)
```

monai/transforms/croppad/array.py
```diff
@@ -472,7 +458,7 @@ def __init__(self, roi_size: Sequence[int] | int) -> None:
         self.roi_size = roi_size
 
     def compute_slices(self, spatial_size: Sequence[int]):  # type: ignore
-        roi_size = fall_back_tuple(self.roi_size, spatial_size)
+        roi_size = fall_back_tuple(self.roi_size, convert_to_numpy(spatial_size, wrap_sequence=True))
```
KumoLiu, line 461: why need add `convert_to_numpy` here?

wyli: I think I ran into a situation that spatial_size is None or an integer, maybe we can remove it for now. Thanks!

monai/transforms/spatial/array.py
```diff
@@ -286,7 +294,7 @@ def __call__(  # type: ignore
         return img
```
yiheng-wang-nv, line 294: Hi @wyli , I'm not sure but I think lines from 288 to 293 may not correct when `lazy_evaluation = True`.
I tried an example like:
```
tr_lazy = ResampleToMatch()
tr_lazy.lazy_evaluation = True
pending_im_mod = tr_lazy(data["im2"], data["im1"])
``` 
and `pending_im_mod.peek_pending_affine()` does not produce expected result.

monai/transforms/croppad/array.py
```diff
@@ -802,16 +804,18 @@ def crop_pad(
...
+        )
         ret = self.padder.__call__(img=cropped, to_pad=pad_width, mode=mode, **pad_kwargs)
         # combine the traced cropping and padding into one transformation
         # by taking the padded info and placing it in a key inside the crop info.
-        if get_track_meta():
-            ret_: MetaTensor = ret  # type: ignore
-            app_op = ret_.applied_operations.pop(-1)
-            ret_.applied_operations[-1][TraceKeys.EXTRA_INFO]["pad_info"] = app_op
+        if get_track_meta() and isinstance(ret, MetaTensor):
+            ret.applied_operations[-1][TraceKeys.EXTRA_INFO]["pad_info"] = ret.applied_operations.pop()
```
KumoLiu, line 818: Seems when `self.lazy_evaluation=True`, here should be `ret.pending_operations`.

monai/transforms/croppad/array.py
```diff
@@ -802,16 +804,18 @@ def crop_pad(
         slices = self.compute_slices(roi_start=box_start, roi_end=box_end)
```
KumoLiu, line 804: If `margin>0`, `roi_start` maybe < 0, but `slices` can only start from 0, then `cropped.peek_pending_shape()` would be the wrong shape.

wyli: maybe after line 814 padding with -box_start, the shape will be correct again?

KumoLiu: No, since line 811-813 was using `cropped.peek_pending_shape()` to compute pad width.

KumoLiu:
Sorry for accidentally clicking "update branch".

monai/transforms/croppad/array.py
KumoLiu: I find the reason, line 818 only updated "pad_info" in `TraceKeys.EXTRA_INFO` for padding transform which may lose `lazy_shape` and `lazy_affine`.

monai/transforms/croppad/array.py
KumoLiu: after a deep look at it, we don't need to update `pending_operations` here since we need lazy info from `pending_operations`. Here may be updated to
```
if get_track_meta() and isinstance(ret, MetaTensor):
    if not self.lazy_evaluation:
        ret.applied_operations[TraceKeys.EXTRA_INFO]["pad_info"] = ret.applied_operations.pop()
return ret
```

monai/transforms/croppad/array.py
KumoLiu: solved when we updated line 818.

monai/transforms/spatial/functional.py
```diff
@@ -158,3 +165,270 @@ def spatial_resample(
         img = img.reshape(full_shape)
     out = convert_to_tensor(img, track_meta=get_track_meta(), dtype=torch.float32)
     return out.copy_meta_from(meta_info) if isinstance(out, MetaTensor) else out  # type: ignore
+
+
```
yiheng-wang-nv, line 169: Hi @wyli ,
https://github.com/Project-MONAI/MONAI/pull/5860/files#diff-442297dd3459f9b82088aaee19c385b18e28099bc22a8353348b178b2094b4c5R138
does line 138 wrong here? should it be: `if isinstance(out, MetaTensor) else out`?

monai/transforms/croppad/array.py
```diff
@@ -1260,14 +1305,29 @@ def __call__(self, img: torch.Tensor, mode: str | None = None, **pad_kwargs) ->
...
+                crop_info = ret_.applied_operations.pop()
+                orig_size = crop_info.get(TraceKeys.ORIG_SIZE)
+                self.push_transform(
+                    ret_, orig_size=orig_size, extra_info={"pad_info": pad_info, "crop_info": crop_info}
+                )
+            else:
+                pad_info = ret_.pending_operations.pop()
+                crop_info = ret_.pending_operations.pop()
+                orig_size = crop_info.get(TraceKeys.ORIG_SIZE)
+                self.push_transform(
```
KumoLiu, line 1323: Do we need push_transform here for lazy mode? We will pop transforms even without applying them.

monai/transforms/croppad/array.py
wyli: Sure, looks good to me

yiheng-wang-nv:
Hi @wyli , @KumoLiu 
Today I got stuck in a test case because of `set_track_meta`. Do you think if we should add a check of `_TRACK_META` when doing lazy resampling? If False, maybe we can raise a warning to let users know that they cannot get expected results.

monai/transforms/croppad/array.py
wyli: > Hi @wyli , @KumoLiu
> Today I got stuck in a test case because of `set_track_meta`. Do you think if we should add a check of `_TRACK_META` when doing lazy resampling? If False, maybe we can raise a warning to let users know that they cannot get expected results.

Yes, if False, we don't update the metatensor's meta info and lazy resampling is not available.

monai/transforms/croppad/array.py
wyli: Yes, I think let's keep it for now, unless you have tested it and it works without these

KumoLiu:
Seems we lose some logic for multisample in `apply_transforms`. For now, we could only apply it in a loop.
https://github.com/Project-MONAI/MONAI/blob/ab800d8413df5680161ea00fb3d6c1a7aa8dd04b/monai/transforms/lazy/functional.py#L64

wyli:
> Seems we lose some logic for multisample in `apply_transforms`. For now, we could only apply it in a loop.
> 
> https://github.com/Project-MONAI/MONAI/blob/ab800d8413df5680161ea00fb3d6c1a7aa8dd04b/monai/transforms/lazy/functional.py#L64

Yes, let's use a for loop for now. this util can handle that use cases, but the design hasn't been discussed yet. https://github.com/Project-MONAI/MONAI/blob/ab6e83ec1ad919876b96fdc070a794d3d8f1ad72/monai/transforms/compose.py#L41

monai/transforms/spatial/functional.py
```diff
@@ -160,3 +167,270 @@ def spatial_resample(
...
+    extra_info = {"original_affine": original_affine}
+
+    shape_np = convert_to_numpy(spatial_shape, wrap_sequence=True)
+    shape_np = shape_np[[i - 1 for i in full_transpose if i > 0]]
+    meta_info = TraceableTransform.track_transform_meta(
+        img,
+        sp_size=shape_np,
+        affine=xform,
+        extra_info=extra_info,
+        orig_size=spatial_shape,
```
yiheng-wang-nv, line 191: Hi @wyli , here `spatial_shape` does not use the same rule as `spatial_resample` does to ensure that the shape is <= 3 dims. Is the rule missing or there are some other considerations? It may impact the `peek_pending_shape`

wyli: we can remove the test case of more than 3d here, I don't think we have a use case in general for 5d affine matrix

yiheng-wang-nv: I see. How about 1d cases?
It seems lazy resampling only handle 2d and 3d cases? It seems `apply_transforms` has the limitation. (see: [here](https://github.com/Project-MONAI/MONAI/blob/ab800d8413df5680161ea00fb3d6c1a7aa8dd04b/monai/transforms/lazy/utils.py#L76) and [here](https://github.com/Project-MONAI/MONAI/blob/ab800d8413df5680161ea00fb3d6c1a7aa8dd04b/monai/transforms/lazy/utils.py#L41))

yiheng-wang-nv:
> Hi @wyli , @KumoLiu Today I got stuck in a test case because of `set_track_meta`. Should we add a check of `_TRACK_META` when doing lazy resampling? If False, maybe we can raise a warning to let users know that they cannot get expected results.

Update: we've aligned with this case, and I'll add a check in https://github.com/Project-MONAI/MONAI/blob/ab800d8413df5680161ea00fb3d6c1a7aa8dd04b/monai/transforms/inverse.py#L138  if False, an error should be raised since lazy resampling is not supported.
The change will be added with the PR: https://github.com/Project-MONAI/MONAI/pull/6080/files

monai/transforms/spatial/functional.py
yiheng-wang-nv: Update: we've discussed and in `peek_pending_affine`, a shape check should be added and give informative warnings if the shape is not supported. I'll address it within the PR: https://github.com/Project-MONAI/MONAI/pull/6080/files

wyli:
> https://github.com/Project-MONAI/MONAI/blob/ab800d8413df5680161ea00fb3d6c1a7aa8dd04b/monai/transforms/inverse.py#L185
> 
> 
> there already has a logic for `get_track_meta == False`. Should we ignore the logic here, or we need to raise error in some other places such as in `apply_transforms`?

maybe we can add a warning message here instead of quietly returning the original default metadata...

monai/transforms/spatial/array.py
```diff
@@ -633,30 +620,15 @@ class Flip(InvertibleTransform):
...
         """
         img = convert_to_tensor(img, track_meta=get_track_meta())
-        axes = map_spatial_axes(img.ndim, self.spatial_axis)
-        out = self.forward_image(img, axes)
-        if get_track_meta():
-            self.update_meta(out, out.shape, axes)
-            self.push_transform(out)
-        return out
+        spatial_shape = img.peek_pending_shape() if isinstance(img, MetaTensor) else img.shape[1:]
+        spatial_chn_shape = [1, *convert_to_numpy(spatial_shape, wrap_sequence=True).tolist()]
```
yiheng-wang-nv, line 630: I think there may no need to define this variable, thus I moved the code into `def flip` (https://github.com/Project-MONAI/MONAI/pull/6080/files#diff-442297dd3459f9b82088aaee19c385b18e28099bc22a8353348b178b2094b4c5R220)
Fix me if I made mistake here.

monai/transforms/spatial/functional.py
```diff
@@ -160,3 +167,270 @@ def spatial_resample(
...
+    im_shape = img.peek_pending_shape() if isinstance(img, MetaTensor) else img.shape[1:]
+    input_ndim = len(im_shape)
+    if input_ndim not in (2, 3):
+        raise ValueError(f"Unsupported image dimension: {input_ndim}, available options are [2, 3].")
+    _angle = ensure_tuple_rep(angle, 1 if input_ndim == 2 else 3)
+    transform = create_rotate(input_ndim, _angle)
+    if output_shape is None:
+        corners = np.asarray(np.meshgrid(*[(0, dim) for dim in im_shape], indexing="ij")).reshape((len(im_shape), -1))
+        corners = transform[:-1, :-1] @ corners  # type: ignore
+        output_shape = corners.ptp(axis=1) + 0.5
```
yiheng-wang-nv, line 282: May still need `output_shape = np.asarray(corners.ptp(axis=1) + 0.5, dtype=int)` to ensure the type, otherwise we may get float values in `peek_pending_shape`. I will address this issue in: https://github.com/Project-MONAI/MONAI/pull/6080

wyli: Sure, please go ahead and fix it. It’s a good point, maybe we don’t need integer shapes during lazy evaluations, we just cast them into integers when resampling, I’ll think about it.

monai/transforms/spatial/array.py
```diff
@@ -1859,6 +1756,7 @@ def __init__(
         padding_mode: str = GridSamplePadMode.BORDER,
         norm_coords: bool = True,
         device: torch.device | None = None,
+        align_corners: bool = True,
```
yiheng-wang-nv, line 1759: This class seems wrong after change.
If adding a test case with `align_corners  = False`, we are able to see errors in `test_resampler.py`.

wyli:
(I couldn’t find where you post the comment) the expected results will be different if you set align_corners=False

On 7 Mar 2023, at 08:33, Yiheng Wang ***@***.***> wrote:



@yiheng-wang-nv commented on this pull request.

________________________________

In monai/transforms/spatial/array.py<https://github.com/Project-MONAI/MONAI/pull/5860#discussion_r1127519082>:

> @@ -1859,6 +1756,7 @@ def __init__(
         padding_mode: str = GridSamplePadMode.BORDER,
         norm_coords: bool = True,
         device: torch.device | None = None,
+        align_corners: bool = True,


This class seems wrong after change.
If adding a test case with align_corners = False, we are able to see errors in test_resampler.py.

tests/test_affine.py
```diff
@@ -60,6 +60,13 @@
...
             ]
         )
+        TESTS.append(
+            [
+                dict(rotate_params=[np.pi / 2], padding_mode="zeros", device=device, align_corners=False),
+                {"img": p(np.arange(4).reshape((1, 2, 2))), "spatial_size": (4, 4)},
+                p(np.array([[[0.0, 0.0, 0.0, 0.0], [0.0, 0.0, 2.0, 0.0], [0.0, 0.0, 3.0, 1.0], [0.0, 0.0, 0.0, 0.0]]])),
+            ]
+        )
         TESTS.append(
```
yiheng-wang-nv, line 70: Hi @wyli , I found that when `align_corners=False`. lazy output always cannot get expected results. I took sometime but cannot find the root cause so far. Could you please help to provide some suggestions? Thanks!

yiheng-wang-nv: `peek_pending_affine` and `peek_pending_shape` are correct for pending output, but after calling `apply_transforms`, the results are different.

KumoLiu: the same issue happens in `zoom` when the mode is "nearest" which can create the correct affine and shape but the results after `apply_transforms` are different.

monai/transforms/spatial/array.py
```diff
@@ -1666,38 +1545,48 @@ def __call__(
...
+                affine @= create_translate(spatial_dims, self.translate_params, device=_device, backend=_b)
             if self.scale_params:
-                affine = affine @ create_scale(spatial_dims, self.scale_params, device=_device, backend=_b)
+                affine @= create_scale(spatial_dims, self.scale_params, device=_device, backend=_b)
         else:
-            affine = self.affine
+            affine = self.affine  # type: ignore
+        affine = to_affine_nd(spatial_dims, affine)
+        if not self.align_corners:
+            affine @= convert_to_dst_type(
```
yiheng-wang-nv, line 1578: `@=` only works on tensor. For numpy, it will raise the following error:
```
TypeError: In-place matrix multiplication is not (yet) supported. Use 'a = a @ b' instead of 'a @= b'.
```
It seems only this place will meet the issue because "self.affine" could be numpy. I will modify this line only in #6080

wyli: Hi @yiheng-wang-nv I'm still working on this, if I change the logic again I'll let you know. for now please skip the `align_corner=True` test cases..

tests/test_affine.py
wyli: Hi @yiheng-wang-nv, I've fixed the `align_corners=False` case, yesterday's commit was wrong, so I force pushed https://github.com/Project-MONAI/MONAI/pull/5860/commits/a43fe9f1c08ade5fa0a201bcca84076340af294b

@KumoLiu could you please let me know which lines cause the issue? I can take another look to double-check.

KumoLiu: Hi @wyli, I just tried to change the mode to `nearest` in L41 and L51, and I find that it could generate the correct shape and affine. But after `apply_transforms`, it can't give the correct output with non-lazy mode. I also output the results to see and discovered that the results will always have a few pixels that are not the same, most at the edge.
https://github.com/Project-MONAI/MONAI/blob/a43fe9f1c08ade5fa0a201bcca84076340af294b/tests/test_zoom.py#L41

tests/test_zoom.py
KumoLiu, line 51: Hi @wyli, I also feel confused here about `align_corners=True`, seems we needn't set `align_corners=True` in Line 41, and by default, it is `False`. I don't know why the test can pass. When I change `align_corners=False`, the test can also pass. Not sure if it's related to your latest commit.

monai/transforms/spatial/dictionary.py
```diff
@@ -743,6 +775,11 @@ def __init__(
         self.mode = ensure_tuple_rep(mode, len(self.keys))
```
yiheng-wang-nv, line 775: here `align_corners ` is not input into `Affine`

wyli: please help revise if needed, it's an input argument in the constructor though https://github.com/Project-MONAI/MONAI/blob/f389d87a6b3f5a0a726c88e0c9fca95f2379f84b/monai/transforms/spatial/dictionary.py#L760-L769

monai/transforms/spatial/dictionary.py
```diff
@@ -756,7 +793,7 @@ def inverse(self, data: Mapping[Hashable, torch.Tensor]) -> dict[Hashable, torch
         return d
 
 
-class RandAffined(RandomizableTransform, MapTransform, InvertibleTransform):
+class RandAffined(RandomizableTransform, MapTransform, InvertibleTransform, LazyTransform):
```
yiheng-wang-nv, line 796: There may has some issues on this class, lazy resampling test passed on `RandAffine` but failed here (I used same input data and parameters to double check)

tests/test_zoom.py
```diff
@@ -14,19 +14,44 @@
...
 VALID_CASES = [(1.5, "nearest"), (1.5, "nearest"), (0.8, "bilinear"), (0.8, "area")]
 
 INVALID_CASES = [((None, None), "bilinear", TypeError), ((0.9, 0.9), "s", ValueError)]
 
 
 class TestZoom(NumpyImageTestCase2D):
+    @parameterized.expand(VALID_CASES)
+    def test_pending_ops(self, zoom, mode):
+        im = MetaTensor(self.imt[0], meta={"a": "b", "affine": DEFAULT_TEST_AFFINE})
+        zoom_fn = Zoom(zoom=zoom, mode="bilinear", keep_size=False, dtype=torch.float64)
```
KumoLiu, line 41: I find tests can only passed when dtype is `float64`. The same issue I found in  `Resize` and `Zoom`.

tests/test_affine.py
wyli: I've finished revising the relevant align corners options, and added consistency tests of Affine vs Rot90 and Resize:

https://github.com/Project-MONAI/MONAI/blob/916f7ced5151d8fb7486aa17667b6162c7889b83/tests/test_rotate90.py#L123-L125

https://github.com/Project-MONAI/MONAI/blob/916f7ced5151d8fb7486aa17667b6162c7889b83/tests/test_affine.py#L188-L191

monai/networks/layers/spatial_transforms.py
```diff
@@ -559,7 +559,7 @@ def forward(
                 affine=theta,
                 src_size=src_size[2:],
                 dst_size=dst_size[2:],
-                align_corners=self.align_corners,
+                align_corners=False,
```
KumoLiu, line 562: Hi @wyli, could you please help me understand here why we need to hard-code this `align_corners` to `False`? Thanks!

wyli: yes, this pytorch function already includes the normalization, if we do it here again, it generates the wrong results... https://github.com/Project-MONAI/MONAI/blob/5cb680ff29ab1a482767b3b29e60cea3cb0d8ab3/monai/networks/layers/spatial_transforms.py#L577

tests/test_zoom.py
wyli: that's probably expected, float32 is faster but less accurate

monai/transforms/spatial/dictionary.py
```diff
@@ -877,26 +921,27 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> dict[Hashable, N
         if do_resampling:  # need to prepare grid
             grid = self.rand_affine.get_identity_grid(sp_size)
             if self._do_transform:  # add some random factors
-                grid = self.rand_affine.rand_affine_grid(grid=grid)
+                grid = self.rand_affine.rand_affine_grid(sp_size, grid=grid)
```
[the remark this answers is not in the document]
wyli: thanks, that's addressed by https://github.com/Project-MONAI/MONAI/pull/5860/commits/717a6fbf613c15cb8b82bd265a7230af851ee29a

monai/networks/layers/spatial_transforms.py
wyli: the current dev branch of this argument is wrong when `normalized=False`, it effectively always generates the same result using `align_corners=False` even though the default is `align_corners=True`. this was confusing and I didn't realise this issue earlier.

tests/test_affine_transform.py
```diff
@@ -350,19 +350,19 @@ def test_forward_2d(self):
         expected = torch.nn.functional.grid_sample(x, grid, align_corners=False)
```
yiheng-wang-nv, line 350: line 226 also needs to change: `align_corners=True`

monai/transforms/spatial/dictionary.py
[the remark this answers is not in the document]
yiheng-wang-nv: Thanks, it fixed.

monai/transforms/spatial/array.py
```diff
@@ -460,12 +473,7 @@ def __call__(
         scale_extent = self.scale_extent if scale_extent is None else scale_extent
         output_shape, offset = compute_shape_offset(data_array.shape[1:], affine_, new_affine, scale_extent)
```
KumoLiu, line 474: Does the shape here `data_array.shape[1:]` be the pending shape or it will compute the output shape based on the original shape?

wyli: you are right, this should be the pending shape

tests/test_affine_transform.py
wyli: thanks, this is now fixed on this branch https://github.com/Project-MONAI/MONAI/pull/5860/commits/f9643c37c4efa999c94939b705826a00dccd2e4c

atbenmurray:
@wyli I have some refactors that I want to make for this PR. I've checked out your fork and implemented the proposed changes that I'd like to discuss. Can I go ahead and commit to this PR branch?

wyli:
Hi @atbenmurray this branch can pass all the unit/integration tests and it’s being used for benchmarking… so it’s fine to change if you have minor bug fixes. 

for design discussions and prototypes please start new pull requests from your fork, feel free to use this branch as a base

atbenmurray:
> 

Umm, I was in the process of going through those branches and ensuring that there was nothing that I needed to port across. Luckily I still have the stuff locally.

wyli:
> Umm, I was in the process of going through those branches and ensuring that there was nothing that I needed to port across. Luckily I still have the stuff locally.

I also backed up the branches to my fork https://github.com/wyli/MONAI/branches in case you need them.

atbenmurray:
Preparing two PRs; splitting the proposed changes into two separate ones

atbenmurray:
@wyli, I've got the first branch ready to be pushed to your fork. Can I get permissions to contribute to it? Thanks!

wyli:
> @wyli, I've got the first branch ready to be pushed to your fork. Can I get permissions to contribute to it? Thanks!

sure, thanks, please see https://github.com/wyli/MONAI/invitations for the permission setting.

tests/test_integration_lazy_samples.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def run_training_test(root_dir, device="cuda:0", cachedataset=0, readers=(None, None), num_workers=4, lazy=True):
+    print(f"test case: {locals()}")
+    images = sorted(glob(os.path.join(root_dir, "img*.nii.gz")))
+    segs = sorted(glob(os.path.join(root_dir, "seg*.nii.gz")))
+    train_files = [{"img": img, "seg": seg} for img, seg in zip(images[:20], segs[:20])]
+
+    # define transforms for image and segmentation
+    lazy_kwargs = dict(mode=("bilinear", 0), padding_mode=("border", "nearest"), dtype=(torch.float32, torch.uint8))
```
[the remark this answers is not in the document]
wyli: I think that's covered here by checking the `transform.keys` https://github.com/Project-MONAI/MONAI/blob/d4b4f5b8fc60124c6f81ba1c8bb021ecd389affd/monai/transforms/compose.py#L66-L67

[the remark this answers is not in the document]
atbenmurray: Not quite. There is an additional change that needs to be made. I'm just trying to get my repo to sync with yours to get the latest commit to 4855-lazy-resampling-impl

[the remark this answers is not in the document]
atbenmurray: https://github.com/wyli/MONAI/pull/21 fixes it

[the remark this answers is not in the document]
wyli: thanks @atbenmurray

monai/transforms/compose.py
```diff
@@ -21,21 +21,64 @@
...
+        if data.has_pending_operations() and ((isinstance(upcoming, (mt.Identityd, mt.Identity))) or upcoming is None):
+            device = overrides.pop("device", None)
+            if device is not None:
+                data = mt.EnsureType(device=device)(data)
+            data, _ = mt.apply_transforms(data, None, overrides=overrides)
+        return data
+    override_keys = ensure_tuple(override_keys)
+    if isinstance(data, dict):
+        if isinstance(upcoming, MapTransform):
+            _keys = [k if k in upcoming.keys and k in data else None for k in override_keys]  # type: ignore
```
wyli, line 67: when override_keys are provided but don't match any of the transform/data keys, this should have a warning message

monai/transforms/compose.py
```diff
@@ -123,6 +177,9 @@ def __init__(
         map_items: bool = True,
         unpack_items: bool = False,
         log_stats: bool = False,
+        lazy_evaluation: bool | None = None,
+        overrides: dict | None = None,
+        override_keys: Sequence[str] | None = None,
```
wyli, line 182: need to provide some good default when `lazy_evaluation=True`

atbenmurray: The default behaviour should be that no overrides are performed, no? The overrides are there for where you have incompatible sets of modes / padding_modes and so forth. I'm working on the PR to reduce the number of cases where this is an issue.

monai/transforms/lazy/functional.py
```diff
@@ -76,30 +79,29 @@ def apply_transforms(
...
-        override_kwargs["resample_mode"] = resample_mode
-    override_kwargs[LazyAttr.DTYPE] = data.dtype if dtype is None else dtype
+    if "mode" in overrides:
+        override_kwargs[LazyAttr.INTERP_MODE] = overrides["mode"]
+    if "padding_mode" in overrides:
+        override_kwargs[LazyAttr.PADDING_MODE] = overrides["padding_mode"]
+    if "align_corners" in overrides:
+        override_kwargs[LazyAttr.ALIGN_CORNERS] = overrides["align_corners"]
+    if "resample_mode" in overrides:
+        override_kwargs["resample_mode"] = overrides["resample_mode"]
```
Nic-Ma, line 89: Should we consider to add `resample_mode` to `LazyAttr`?

Thanks.

tests/test_integration_lazy_samples.py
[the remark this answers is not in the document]
Nic-Ma: Thanks for the quick fix.

monai/transforms/compose.py
Nic-Ma: I agree, a warning message can improve the user experience here.

Thanks.

Nic-Ma (approved):
Looks good to me, looking forward to more tests and some bundle benchmarks before merging.

Thanks.

monai/transforms/compose.py
```diff
@@ -21,21 +21,64 @@
...
+    """
+    Given the upcoming transform ``upcoming``, if lazy_resample is True, go through the MetaTensors and
+    evaluate the lazy applied operations. The returned `data` will then be ready for the ``upcoming`` transform.
+    """
+    if not lazy_evaluation:
+        return data  # eager evaluation
+    overrides = (overrides or {}).copy()
+    if isinstance(data, monai.data.MetaTensor):
+        if data.has_pending_operations() and ((isinstance(upcoming, (mt.Identityd, mt.Identity))) or upcoming is None):
+            device = overrides.pop("device", None)
```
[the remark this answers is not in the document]
atbenmurray: Yes, I think we do.

monai/transforms/compose.py
atbenmurray: Hmmm. Is it actually a warning? You might omit a key because you don't need to specify an override for them. You might need to change ```mode``` for 'image' if it is set to one of the various ```bilinear```-style keys, but ```label``` might already be set to ```nearest```.

wyli:
thanks, I'm working on this branch to address the review comments.

atbenmurray:
> thanks, I'm working on this branch to address the review comments.

Ahh, @wyli I was in the process of adding the missing keywords to the overrides namespace, but if you are already doing it I'll leave it

wyli:
> > thanks, I'm working on this branch to address the review comments.
> 
> Ahh, @wyli I was in the process of adding the missing keywords to the overrides namespace, but if you are already doing it I'll leave it

sure, I'll make minor usability fixes and merge it soon..

monai/transforms/compose.py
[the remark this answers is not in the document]
wyli: EnsureType -- it's mainly used to change the device (used in https://github.com/Project-MONAI/tutorials/blob/main/acceleration/fast_training_tutorial.ipynb), I'll add some tests

monai/transforms/compose.py
wyli: sure, in that logic, empty `override_keys` means check all the data inputs for lazy evaluation, I'll change. previously this variable was `lazy_key`, which means if it's empty, then no input is checked for lazy evaluation. I'll think about it.

PR #5878, opened 2023-01-19
Fix redundancy in test WSI image

drbeh:
### Description

The test WSI image is downloaded to a file with redundant extension `.tiff.tiff` and also the code has unused variables defined like basename. This PR fix that and simplify the WSI name.

### Types of changes


Nic-Ma:
Hi @drbeh ,

May I know which line in the original code causes "redundant extension .tiff.tiff" as you described?
I didn't find it immediately.

Thanks in advance.

drbeh:
> Hi @drbeh ,
> 
> May I know which line in the original code causes "redundant extension .tiff.tiff" as you described? I didn't find it immediately.
> 
> Thanks in advance.

Hi @Nic-Ma, this is the line that generates extra extension:
```python
base_name, extension = os.path.basename(f"{FILE_URL}"), ".tiff"
FILE_PATH = os.path.join(os.path.dirname(__file__), "testing_data", FILE_NAME + extension)
```
`os.path.basename` return the base part of the URL and not the base part of filename. To separate extension, one should use `os.path.splitext`. Since the basename already have `.tiff`, the second line add another extension to it.
```python
base_name, extension = os.path.splitext(os.path.basename(f"{FILE_URL}"))
```
But it is not needed here as we don't need to extract the extension on the first place.

Nic-Ma (approved):

tests/test_masked_inference_wsi_dataset.py
```diff
@@ -29,9 +29,8 @@
 
 FILE_KEY = "wsi_img"
 FILE_URL = testing_data_config("images", FILE_KEY, "url")
-base_name, extension = os.path.basename(f"{FILE_URL}"), ".tiff"
```
wyli, line 32: to me the filename ending with `.tiff.tiff` is somewhat a better test case than a regular `image_name.tiff`, showing that the reader's file extension matching works fine

drbeh: I might be missing something, but in this new way also the extension is explicitly set:
```
FILE_NAME = f"temp_{FILE_KEY}.tiff"
```

drbeh: We can have the base part of the file name instead of `FILE_KEY` without duplicating `.tiff` if you see any benefit to that. But I thought that having `test_{FILE_KEY}.tiff`  is more representative what the file is.

drbeh: @wyli reading your comment again, I realized what you mean and it's a good point. I think a better way to do it is to explicitly mentioning it in the FILE_KEY (i.e. `wsi_tiff` instead of `wsi_image`) and not adding any additional extension. In this way, the developer who is writing the test knows that it is using a tiff file. Otherwise, it is not visible to the developer what the actual format of the file unless check the file that is created in the testing_data.

I will update the PR to address your point.

drbeh: Update the PR.

wyli: thanks for the updates, I checked out the code and ran the test cases, still I don't fully understand why these code changes are needed. were you seeing some unexpected outcomes of the code?

drbeh: @wyli I believe this PR now makes improvement to two concerns:

1. the developer who knows what image is being used (`CMU-1.tiff` from Carnegie Melon University) doesn't expect to deal with `.tiff.tiff` (the original goal of this PR). 
2. the users that just use the `testing_data_config` API to write unittests don't know what's the format of `wsi_img` is. Also WSI images are in multiple format (.tiff, .svs, .mrxs, etc.) so making it explicit about `wsi_image.tiff` would be helpful to avoid confusions.

wyli: thanks, I think in general it's good to simulate some corner cases as the user's input to detect potential issues. if multiple format support is important, shall we focus on generating test cases in different formats (and may be combined with different variants of file extension names)?

drbeh:
closing in favor of #6244.

PR #5880, opened 2023-01-19
5869 adds support of `_mode_` keyword in `instantiate`

### LINKED ISSUE #5869: Supporting callable functions in instantiate
Continuing discussion from #5854 

Callable functions can form a key part of the instantiation pipeline. For example, several functions return object instances, and supporting this would be very useful. 

Currently, using `monai.networks.nets.resnet.resnet50` to create a model instance, I've to do the following,
````
      model: "$@#_requires_()"
      _requires_:
        _target_: monai.networks.nets.resnet.resnet50
        pretrained: False
        n_input_channels: 1
        widen_factor: 2
````        

which gets quite hacky because I've to use the `_requires_` attribute to do this. I would just expect to do this with a `_target_` call (or by having a potential `_call_` attribute set)

This can also be useful to several other monai functionality offered in functions such as
````
monai.optimizers.generate_param_groups # Calling the function to generate param groups during optimizer instantiation
monai.engines.multi_gpu_supervised_trainer.create_multigpu_supervised_evaluator # Instantiating trainer object using this call
````

I would expect `_target_` to invoke default behavior. In the case of a function, this would be calling it rather than creating a callable partial function. Creating a partial fn with a `_partial_` attribute for instantiate to invoke this behaviour would be very easy.

If the partial fn setting of the instantiate is tied to many workflows, having a  `_call_` attribute to offer the inverse functionality would also work. But I think it having both functionalities (the ability to call and to create a partial fn) would be a huge plus. 
Currently, there is no straightforward way to create instances from fn calls. 

Alternative outlined above

### PR DESCRIPTION
**Pull request by wyli**

Fixes #5869 

### Description
support an optional `_mode_` in `"default"`/`"partial"`/`"debug"` when calling `component(**kwargs)`

example config:
```
{"xform": 
    {"_target_": "Compose", 
     "transforms": [], 
     "_mode_": "debug"}}
```

### Types of changes

### CODE REVIEW

File: monai/bundle/config_item.py [hunk 1]
```diff
@@ -177,6 +177,11 @@ class ConfigComponent(ConfigItem, Instantiable):
           but requires the dependencies to be instantiated/evaluated beforehand.
         - ``"_disabled_"`` (optional): a flag to indicate whether to skip the instantiation.
         - ``"_desc_"`` (optional): free text descriptions of the component for code readability.
+        - ``"_mode_"`` (optional): the operating mode for invoking the ``component``:
+
+            - ``"default"``: returns ``component(**kwargs)``
```
Nic-Ma, line 182: Could you please help describe more about the "component" here, it can be a class name or function? It may cause unexpected behavior for users.

Thanks.

monai/utils/module.py
```diff
@@ -214,36 +214,55 @@ def load_submodules(basemod, load_all: bool = True, exclude_pattern: str = "(.*[
...
 
     Args:
-        path: full path of the target class or function component.
-        kwargs: arguments to initialize the class instance or set default args
-            for `partial` function.
+        __path: if a string is provided, it's interpreted as the full path of the target class or function component.
+            If a callable is provided, ``__path(**kwargs)`` or ``functools.partial(__path, **kwargs)`` will be returned.
+        __mode: the operating mode for invoking the ``component`` represented by ``__path``:
+
+            - ``"default"``: returns ``component(**kwargs)``
```
Nic-Ma, line 228: Same comment as the above.

Thanks.

tests/test_config_parser.py
```diff
@@ -36,6 +36,13 @@ def case_pdb(sarg=None):
     parser.get_parsed_content()
 
 
+@TimedCall(seconds=100, force_quit=True)
+def case_pdb_inst(sarg=None):
```
Nic-Ma, line 40: I am not very familiar with pdb test, do we need to check / assert anything in this function?

Thanks.

Nic-Ma (approved):
This PR slightly changed the default behavior of `ConfigComponent` for functions, but I think it's acceptable because now it's same as Hydra and other similar projects:
https://github.com/facebookresearch/hydra/blob/main/hydra/_internal/instantiate/_instantiate2.py#L283

Thanks.

Issue #5884, opened 2023-01-22
Fully type annotate any partially typed functions (mypy)


Shadow-Devil:
**Is your feature request related to a problem? Please describe.**
As a followup of #5823 and #5657 a next step could be to fill out type annotations for any function that already has at least one type annotation.

**Describe the solution you'd like**
Fully type annotate parameters and return types for functions that have at least one type annotation.
This can also be enforced by the mypy config `disallow_incomplete_defs = True`.
**Should not be done in one big PR, but in smaller PRs (maybe for each module or less).**

**Describe alternatives you've considered**
Maybe first get https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_untyped_defs to pass, so for each function you would only need at least one type annotation.

**Additional context**
- mypy documentation: https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_incomplete_defs

Finished Modules:
- [x] _extensions
- [x] apps
- [x] auto3dseg
- [x] bundle
- [x] config
- [ ] data
- [x] engines
- [x] fl
- [x] handlers
- [x] inferers
- [x] losses
- [x] metrics
- [ ] networks
- [x] optimizers
- [ ] transforms
- [x] utils
- [x] visualize

Shadow-Devil:
Still missing following modules:
```
networks   Found 142 errors in 37 files (checked 1048 source files)
transforms Found 187 errors in 23 files (checked 1048 source files)
data       Found 223 errors in 20 files (checked 1048 source files)
```

PR #5885, opened 2023-01-22
Disallow incomplete defs in visualize module

Shadow-Devil:
Part of #5884.

### Description

Fully type annotate any functions with at least one type annotation in module `visualize`.

### Types of changes


wyli (approved):
Thanks, it looks good to me.

PR #5890, opened 2023-01-23
add rand identity

rijobro:
### Description

Adds `RandIdentity` which can be placed in a sequence of transforms to stop the caching of any results beyond this point.

### Types of changes

### DIFF
--- a/monai/transforms/utility/array.py
+++ b/monai/transforms/utility/array.py
@@ -128,6 +128,7 @@ class Identity(Transform):
         """
         return img
 
+
 class RandIdentity(RandomizableTransform):
     """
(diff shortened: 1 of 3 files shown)


monai/transforms/utility/array.py
```diff
@@ -128,6 +129,18 @@ def __call__(self, img: NdarrayOrTensor) -> NdarrayOrTensor:
         return img
 
 
+class RandIdentity(RandomizableTransform):
```
wyli, line 132: `RandomizableTrait` is the minimal class...
```suggestion
class RandIdentity(RandomizableTrait):
```

wyli (approved):

PR #5891, opened 2023-01-23
Complete partially typed signatures (monai.utils)

Shadow-Devil:
Part of #5884.

### Description

Fully type annotate any functions with at least one type annotation in
module `visualize`

### Types of changes

### DIFF
--- a/monai/utils/jupyter_utils.py
+++ b/monai/utils/jupyter_utils.py
@@ -233,7 +233,9 @@ def plot_engine_status(
     return fig, axes
 
 
-def _get_loss_from_output(output: list[torch.Tensor | dict[str, torch.Tensor]] | dict[str, torch.Tensor] | torch.Tensor) -> torch.Tensor:
+def _get_loss_from_output(
+    output: list[torch.Tensor | dict[str, torch.Tensor]] | dict[str, torch.Tensor] | torch.Tensor
+) -> torch.Tensor:
     """Returns a single value from the network output, which is a dict or tensor."""
 
     def _get_loss(data: torch.Tensor | dict[str, torch.Tensor]) -> torch.Tensor:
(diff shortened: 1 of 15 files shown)


Shadow-Devil:
CI failed, because python Literals were introduced in 3.8, but I'm using them inside the code

monai/utils/jupyter_utils.py
```diff
@@ -147,25 +149,25 @@ def tensor_to_images(name: str, tensor: torch.Tensor):
...
-        return tensor[:, dmid].cpu().data.numpy()
+        return tensor[:, dmid].cpu().data.numpy()  # type: ignore[no-any-return]
 
     return None
 
 
 def plot_engine_status(
     engine: Engine,
-    logger,
+    logger: MetricLogger,
```
Shadow-Devil, line 162: Building the docs doesn't work because of this forward reference of `MetricLogger`. I tried putting it into quotes so `logger: "MetricLogger"` but this is automatically reverted by the pre-commit pyupgrade. I also searched for inline comments to locally disable pyupgrade, but this does not exist.

So my solutions would be one of:
- Completely disable pyupgrade for this file
- (Preferred) Use `Any` for the argument

monai/utils/misc.py
```diff
@@ -82,7 +82,20 @@ def star_zip_with(op, *vals):
     return zip_with(op, *vals, mapfunc=itertools.starmap)
 
 
-def first(iterable, default=None):
+T = TypeVar("T")
+
+
+@overload
+def first(iterable: Iterable[T], default: T) -> T:
```
wyli, line 89: just a comment that perhaps we should start to put these definitions in a separate `.pyi` stub file if it doesn't take a lot of effort to maintain...

wyli (approved):

PR #5892, opened 2023-01-24
make path user expansion optional

colobas:
### Description

When using `LoadImage`, file paths get wrapped with `pathlib.Path` and `.expanduser()` is called. This causes paths with a protocol prefix to get corrupted (e.g. in my case paths starting with `s3://` are getting one of the forward slashes removed)

This PR makes the wrapping with `Path` and call to `.expanduser()` optional (but set to `True` by default, to preserve current behavior)

### Types of changes

### DIFF
--- a/monai/transforms/io/array.py
+++ b/monai/transforms/io/array.py
@@ -128,6 +128,7 @@ class LoadImage(Transform):
         simple_keys: bool = False,
         prune_meta_pattern: str | None = None,
         prune_meta_sep: str = ".",
+        expanduser: bool = True,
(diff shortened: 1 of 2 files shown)


wyli:
Thanks, could you please amend the commit to have a sign-off line "git commit --amend --sign-off"? See https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#signing-your-work

colobas:
signed off!

wyli (approved):

Issue #4117, opened 2022-04-12
Move code that converts ddf tensor to itk VectorImage into library?


wyli:
### Discussed in https://github.com/Project-MONAI/MONAI/discussions/4104

<sup>Originally posted by **HastingsGreer** April 11, 2022</sup>
There is a correct way to convert a ddf tensor accepted by Monai's warp block into an itk VectorImage accepted by itk's WarpImageFilter. 

This conversion is defined and tested here:

[https://github.com/Project-MONAI/MONAI/blob/af6560a5facef14483e43cb516d8f07d72aa1ffe/tests/test_warp.py#L178](https://github.com/Project-MONAI/MONAI/blob/af6560a5facef14483e43cb516d8f07d72aa1ffe/tests/test_warp.py#L178)

I would like to use this conversion code, and I think it is generally useful since it correctly handles fiddly details like transposing the tensor and reversing the order of elements in each vector. Would a pull request moving it from the test code into the MONAI library be welcome?

--
see also https://github.com/Project-MONAI/tutorials/blob/master/modules/resample_benchmark.ipynb

Issue #5711, opened 2022-12-12
[Feature request] Plugin/Module to visualize MONAI Transforms


diazandr3s:
**Is your feature request related to a problem? Please describe.**

There is no tool to visualize the effect of MONAI transforms on a custom dataset.

MONAI Label users will highly benefit from a tool/plugin/module that allows them to visualize the effect of the MONAI transforms applied to their dataset. 

I've got lots of questions about which transforms are best to do data augmentation or to preprocess their dataset: window scale intensity, random flipping, random rotation, scaling, affine transform, etc.

The idea of this tool is to facilitate the selection of MONAI transforms that researchers can use [here](https://github.com/Project-MONAI/MONAILabel/blob/main/sample-apps/radiology/lib/trainers/segmentation.py#L74-L98) to do data augmentation. 

Ideally, researchers can use this plugin with individual transform and/or a set of random transformations (Compose).

**Describe the solution you'd like**
A MONAI tool/plugin/module that allows users to visualize the effects of the MONAI transforms applied to their dataset.

dzenanz:
You could get the feel of some data augmentation transforms by using [TorchIO's Slicer plugin](https://torchio.readthedocs.io/interfaces/slicer.html). Maybe MONAI could make a similar thing? The easiest thing would be to add MONAI-specific transforms to the [exiting plugin code](https://github.com/fepegar/SlicerTorchIO). If the author of the plugin @fepegar doesn't like that, a fork could be made.

fepegar:
I think adding MONAI transforms to the TorchIO extension might make things confusing, so forking would possibly make more sense. But I might be wrong!

arpinerap:
I agree with @diazandr3s. Its really hard to manage so huge amount of possibilities in transforms that Monai offer without a graphic interface. TorchIO´s Slicer plugin is a wonderfull tool but with very little options. Developing a good data augmentation strategy without being able to know what transforms exactly do with your images is really hard. Please, reconsider @diazandr3s request, no doubt I´ll be the first user. Thanks in advance!

dzenanz:
If no one else plans to work on this, I could take a crack over the next weeks. @wyli could you assign it to me?

Nic-Ma:
Hi @dzenanz ,

Sure, welcome contribution!

Thanks.

wyli:
given that a monai `Compose` object could be created from a yaml/json config, e.g.
[...]

PR #5897, opened 2023-01-25
Expose ITK Image to MONAI MetaTensor conversion

### LINKED ISSUE #5708: [Feature request] Expose ITK Image to MONAI MetaTensor conversion
I am trying to convert easily between ITK Image and MONAI MetaTensor. Currently, I am using my own simple function that works for my partcular use case:

```
def itk_image_to_metatensor(img):
    reader = ITKReader(affine_lps_to_ras=False)
    img_array, meta_data = reader.get_data(img)
    img_array = convert_to_dst_type(img_array, dst=img_array, dtype=itk.D)[0]
    metatensor = MetaTensor.ensure_torch_and_prune_meta(img_array, meta_data)
    metatensor = EnsureChannelFirst()(metatensor)

    return metatensor
``` 
which more-or-less the code that lives insde `LoadImage`: https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/io/array.py#L273-L289

I would like that MONAI provides two functions e.g. `itk_image_to_metatensor()` and `metatensor_to_itk_image()`. The implementation of the first already exists inside `LoadImage` transform (https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/io/array.py#L273-L289) and of the latter inside ITKWriter: https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/image_writer.py#L470-L517. So, ideally they just need to be exposed in order for them to be more easily usable.

Thanks in advance ;) 
--
xref for us: https://github.com/InsightSoftwareConsortium/ITKElastix/issues/126

Shadow-Devil:
Fixes #5708
Fixes https://github.com/Project-MONAI/MONAI/issues/4117

### Description

This is a migration of the PR https://github.com/InsightSoftwareConsortium/itk-torch-transform-bridge/pull/6 into MONAI.

### Types of changes


Spenhouet:
@Shadow-Devil FYI https://github.com/InsightSoftwareConsortium/itk-torch-transform-bridge/issues/7

monai/data/itk_torch_affine_matrix_bridge.py
```diff
@@ -0,0 +1,306 @@
...
+    "monai_to_itk_affine",
+    "create_itk_affine_from_parameters",
+    "itk_affine_resample",
+    "monai_affine_resample",
+]
+
+
+def metatensor_to_array(metatensor: MetaTensor):
+    metatensor = metatensor.squeeze()
+    metatensor = metatensor.permute(*torch.arange(metatensor.ndim - 1, -1, -1))
```
wyli, line 46: the writer class also does this permute

https://github.com/Project-MONAI/MONAI/blob/50c3f328cb98e4f0c4e6a38a04a98badf0961c18/monai/data/image_writer.py#L507

so this shouldn't be done with the SaveImage to avoid permute twice...

Shadow-Devil: I don't quite understand what you mean with this. Should I remove this function and just use `ITKWriter.create_backend_obj()` instead?

wyli: I think `metatensor_to_array` should be removed... the `meta_tensor.array` property returns the numpy array. any data format change/assumption shouldn't be hard-coded in this way, unless I miss something here cc @ntatsisk

ntatsisk: Looking back I think the naming of the function could be better. The idea is that you need to match the axis of the array the comes from `np.asarray(itk_image)`, and in order to do that you need a squeeze and an axis-swap. Feel free to remove it, but these two operations need to be done then inline before comparing to the array that comes from itk. That at least for testing purposes (no image saving there).

monai/data/itk_torch_affine_matrix_bridge.py
```diff
@@ -0,0 +1,350 @@
...
+
+def monai_affine_resample(metatensor: MetaTensor, affine_matrix: NdarrayOrTensor):
+    # TODO documentation
+    affine = Affine(affine=affine_matrix, padding_mode="zeros", mode="bilinear", dtype=torch.float64, image_only=True)
+    output_tensor = cast(MetaTensor, affine(metatensor))
+
+    return cast(MetaTensor, output_tensor.squeeze().permute(*torch.arange(output_tensor.ndim - 2, -1, -1))).array
+
+
+def monai_to_itk_ddf(image, ddf):
```
wyli, line 279: since this file now also includes some ddf related, the filename could be more generic... please also update the `__init__.py` to have these functions https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/__init__.py

ntatsisk: I don't know if it can help with this error but my idea with the functions `create_itk_affine_from_parameters', 'itk_affine_resample', 'monai_affine_resample', 'monai_wrap', 'itk_wrap' is that they should be part of the testing and not exposed to the user. So, maybe they could be moved to the test files and fix the cyclic imports?

Shadow-Devil: I'm sorry I'll move them. `remove_border` should still be exposed to the user, right?

ntatsisk: It was my fault to begin with of having everything in the same file ;) Also, `remove_border` is part of the testing so not exposed.

wyli: for the circular import we usually avoid it by accessing the name later in the function:

```py
import monai

...
def function_that_uses_warp():
    ...
    x = monai.networks.blocks.Warp()
```

tests/test_itk_torch_bridge.py
```diff
@@ -0,0 +1,440 @@
...
+    monai_to_itk_ddf,
+)
+from monai.networks.blocks import Warp
+from monai.transforms import Affine
+from monai.utils import optional_import
+from tests.utils import skip_if_downloading_fails, testing_data_config
+
+itk, has_itk = optional_import("itk")
+
+TESTS = ["CT_2D_head_fixed.mha", "CT_2D_head_moving.mha", "copd1_highres_INSP_STD_COPD_img.nii.gz"]
```
wyli, line 37: I've uploaded the test cases to 
- https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/CT_2D_head_fixed.mha
- https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/CT_2D_head_moving.mha

could you please update the data_config.json and let the tests download them on the fly? thanks!

Shadow-Devil:
@ntatsisk Thank you for fixing the issues described in your PR! I've added your latest three commits, so https://github.com/InsightSoftwareConsortium/itk-torch-transform-bridge/pull/6/commits/12f4d4bc16411f4a000375cef177ae5232cf968c https://github.com/InsightSoftwareConsortium/itk-torch-transform-bridge/pull/6/commits/37cc4f99b44ab25f0cf5961c07018642ced92b29 and https://github.com/InsightSoftwareConsortium/itk-torch-transform-bridge/pull/6/commits/7a253b242f26fbcf3934f53dd00831de3ad82f68. 
One problem I've found is that sometimes the `test_random_array` fails, maybe I should increase the tolerance?
And is there anything else that I should adjust?

ntatsisk:
Thanks @Shadow-Devil! Yes, increasing the tolerance is one option. You could also copy-paste the output of a random array and keep it fixed, or another way is to set a value for the random seed. I am not aware what is the most common approach in the rest of the monai codebase.

Shadow-Devil:
Since most of this test uses `torch.rand` and `np.random.rand` I decided that setting the random seed of both torch and numpy would be best, or else we need to set specific img, spacing, direction and origin.
@wyli I don't know if this seed will interfere with any other tests or if it should be reset somehow. What do you think?

Another problem I saw within the pipeline is that a testcase is requesting too much memory. Should we skip this test in the quick run? How can I do that?

tests/test_itk_torch_bridge.py
```diff
@@ -0,0 +1,525 @@
...
+    warp = Warp(mode="bilinear", padding_mode="zeros")
+    warped_image = warp(image_tensor.to(torch.float64), ddf_tensor.to(torch.float64))
+
+    return warped_image.to(torch.float32).squeeze().numpy()
+
+
+@unittest.skipUnless(has_itk, "Requires `itk` package.")
+class TestITKTorchAffineMatrixBridge(unittest.TestCase):
+    def setUp(self):
+        torch.manual_seed(RANDOM_SEED)
```
wyli, line 189: please consider these setup and teardown to avoid changing the global random state:

https://github.com/Project-MONAI/MONAI/blob/58c256eae2f1d21436a0b9c0dab5d989b8665e55/tests/test_inverse.py#L367-L371

https://github.com/Project-MONAI/MONAI/blob/58c256eae2f1d21436a0b9c0dab5d989b8665e55/tests/test_inverse.py#L395-L396

Shadow-Devil:
@ntatsisk Didn't you want both ITKImage -> Metatensor and Metatensor -> ITKImage conversions in this PR? 
Currently there is only the function `itk_image_to_metatensor`. Will you also develop a `metatensor_to_itk_image` function in your PR?

wyli:
> @ntatsisk Didn't you want both ITKImage -> Metatensor and Metatensor -> ITKImage conversions in this PR?
> Currently there is only the function `itk_image_to_metatensor`. Will you also develop a `metatensor_to_itk_image` function in your PR?

I think we can create the function, the implementation could be just a thin wrapper of this logic :
```
itk_obj = monai.data.ITKWriter.create_backend_obj(
    meta_tensor.array,
    channel_dim=None,
    affine=meta_tensor.affine,
    affine_lps_to_ras=False,  # False if the affine is in itk convention
)
# itk.imwrite(itk_obj, "output.nii.gz")
```

wyli (approved):
thanks, it looks nice, 
I'll fix a few style issues and merge it soon.

there is a warning message from this call `np.dtype(itk.D)`:

```
in the future the `.dtype` attribute of a given datatype object must be a valid dtype instance. `data_type.dtype` may need to be coerced using `np.dtype(data_type.dtype)`. (Deprecated NumPy 1.20)
```

perhaps we should think about proper methods of getting equivalent dtypes like this https://github.com/Project-MONAI/MONAI/blob/f5708ea89af86ca74f88e93dabb0ec94c4944ac1/monai/utils/type_conversion.py#L29-L39

ntatsisk:
Hi @Shadow-Devil , @wyli. Thanks again for all this amazing work done here! I might visit again my two PRs to see if it is possible:

- To make the affine bridge work for different array shapes between the image to be transformed and the reference image.
- To add a reference image for DDF PR as well

What do you think is preferable: wait for a couple of weeks before merging, or merge now and append any changes later in monai?

wyli:
thanks @ntatsisk, please feel free to create follow-up feature/pull requests... I think this PR is a fairly self-contained starting point.

also, perhaps we should have a meeting to discuss an overall design and roadmap, for example a possible direction is towards an object-oriented approach like https://github.com/InsightSoftwareConsortium/itk-torch-transform-bridge/pull/6#issuecomment-1415954890 please let me know @Shadow-Devil @ntatsisk if you are interested in this topic.

Spenhouet:
> thanks @ntatsisk, please feel free to create follow-up feature/pull requests... I think this PR is a fairly self-contained starting point.
> 
> also, perhaps we should have a meeting to discuss an overall design and roadmap, for example a possible direction is towards an object-oriented approach like [InsightSoftwareConsortium/itk-torch-transform-bridge#6 (comment)](https://github.com/InsightSoftwareConsortium/itk-torch-transform-bridge/pull/6#issuecomment-1415954890) please let me know @Shadow-Devil @ntatsisk if you are interested in this topic.

If there is interest in this, I'm happy to participate and/or share updated code.

dzenanz:
I, too, am glad to see this merged. It will make my life easier with #5711.

Spenhouet:
@ntatsisk Really looking forward to a solution with respect to different image sizes. That's the only bigger limitation I'm aware of right now.

Shadow-Devil:
> also, perhaps we should have a meeting to discuss an overall design and roadmap, for example a possible direction is towards an object-oriented approach like https://github.com/InsightSoftwareConsortium/itk-torch-transform-bridge/pull/6#issuecomment-1415954890 please let me know @Shadow-Devil @ntatsisk if you are interested in this topic.

Sorry for the late response. I won't join this meeting, since this was my first encounter with ITK, so there isn't much I could contribute. I hope you will find a good design/roadmap for further development. 
Thank you for merging this PR, I'm glad that we got a nice first version done.

Issue #5871, opened 2023-01-18
Implement a sliding window Splitter for whole slide images


drbeh:
Similar to #5870, a sliding window class is required to extract patches from whole slide images that cannot fit into memory so the input will be the path to a WSI on the disk and patches are extracted on the fly.

drbeh:
Mistakenly closed instead of #5870.

Issue #5837, opened 2023-01-09
[RFC] Compositional Patch-based Inference


drbeh:
Currently, `SlidingWindowInferer` and ` SlidingWindowHoVerNetInferer` rely on [`sliding_window_inferer`](https://github.com/Project-MONAI/MONAI/blob/9efd54d000d5efd325cda3b752d36651d1796884/monai/inferers/utils.py#L37), which lacks flexibility and structure to generalize to other use cases. For instance, different way of splitting image into patches (e.g., whole slide images or patch filtering), or different way of merging the result patches together.

To make the inference on patches more modular and extensible, I have prepared the following design and would appreciate your feedbacks on it.

In this way, similar behavior to `SlidingWindowInferer`:
```python
SlidingWindowInferer(roi_shape=(256, 256), overlap=0.5, mode="gaussian", sigma_scale=1.0, sw_batch_size=8)
```

can be achieved as following:
```python
PatchInference(splitter=SlidingPatchSplitter(patch_size=(256, 256), overlap=0.5), merger=GaussianMerger(sigma=1.0), batch_size=8)
```
while it is easy to extend this via composition of splitter and merger. 

For instance, use the maximum value for overlapping regions:
```python
PatchInference(splitter=SlidingPatchSplitter(patch_size=(256, 256), overlap=0.5), merger=MaxMerger())
```
or use a whole slide image as an input:
```python
PatchInference(splitter=WSISlidingPatchSplitter(patch_size=(256, 256), overlap=0.5), merger=MaxMerger())
```

Here, I mentioned few concrete classes for `Splitter` and `Merger` as example but many more concrete classes can implemented to extend the usage of `PatchInference`.

![PatchInference](https://user-images.githubusercontent.com/3968947/211410534-a9347e99-0725-4fb2-8745-493f97ded4e2.jpg)

PS: I would suggest to change the name of `Inferer` base class to `Inference` since "inferer" does not exist in English vocabulary.

CC: @Nic-Ma @ericspod @wyli @rijobro

Nic-Ma:
Hi @drbeh ,

To be backward compatible, I would suggest to:
1. Keep the `Inferer` class, Yan, Eric, Wenqi and I defined this signature at the very beginning of MONAI.
2. Keep the `SlidingWindowInferer` as a typical (or simplified) case of `PatchInference`.

@wyli @ericspod What do you guys think?

Thanks.

ericspod:
[...]

PR #5914, opened 2023-01-30
Patch inference

### LINKED ISSUE #5864: Implement `PatchInferer`, `Splitter`, `Merger`
According to the patch-based inference design https://github.com/Project-MONAI/MONAI/issues/5837, implement the following classes:
- `PatchInferer`
- `Splitter`
- `Merger`

drbeh:
Fixes #5864 #5871 #5872 

### Description

This PR implements `PatchInferer` according to #5837, the base classes for `Splitter` and `Merger` along with few concrete classes `SlidingWindowSplitter` and `AvgMerger`. This can provide a sliding window inference for tensor inputs.

### Types of changes


monai/inferers/inferer.py
```diff
@@ -64,6 +67,166 @@ def __call__(self, inputs: torch.Tensor, network: Callable[..., torch.Tensor], *
...
+        self.patch_filter_fn = patch_filter_fn
+
+        # batch size for patches
+        self.batch_size = batch_size
+
+        # model output keys
+        self.output_keys = output_keys
+
+        # check if merger is initialized
+        self.is_merger_initialized = False
```
Nic-Ma, line 131: This flag may cause the thread-safe issue, please double-check the logic.

Thanks.

KumoLiu (commented):
Thanks for the enhancement update, it's quite a large pr, haven't quite review it all. :)

monai/inferers/inferer.py
```diff
@@ -64,6 +67,189 @@ def __call__(self, inputs: torch.Tensor, network: Callable, *args: Any, **kwargs
...
+        # mergers
+        self.mergers: Sequence[Merger] = (AvgMerger(),) if merger is None else ensure_tuple(merger)
+        for m in self.mergers:
+            if not isinstance(m, Merger):
+                raise TypeError(f"'merger' should be a `Merger` object, {type(m)} is given.")
+
+        # pre-processor (process patch before the network)
+        if pre_processor is not None and not callable(pre_processor):
+            raise TypeError(
+                f"'pre_processor' should be a callable object, not None and {type(pre_processor)} is given."
```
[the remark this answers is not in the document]
drbeh: I meant to write `nor` ;)

monai/inferers/splitter.py
```diff
@@ -0,0 +1,199 @@
...
+            n_pos_params = len([v for v in sig.parameters.values() if v.default is _empty])
+            if n_params < 2:
+                raise ValueError(
+                    f"`patch_filter_fn` requires to accept at least two parameters (patch, location)."
+                    f"The provided callable ({filter_fn}) has {n_params} parameters."
+                )
+            elif n_pos_params > 2:
+                raise ValueError(
+                    f"`patch_filter_fn` can have at most two positional parameters (patch, location)."
+                    f"The provided callable ({filter_fn}) has {n_pos_params} positional parameters."
```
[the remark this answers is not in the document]
drbeh: Sure, `filter_fn` are being called at `self.filter_fn(patch, location)` so it has if it has more than two positional parameters, it will fail.

monai/inferers/splitter.py
```diff
@@ -0,0 +1,199 @@
...
+            f"{type(filter_fn)} is given."
+        )
+
+    def _get_valid_patch_size(self, spatial_ndim):
+        return ensure_tuple_rep(self.patch_size, spatial_ndim)
+
+    def _get_valid_overlap(self, spatial_ndim, patch_size):
+        # broadcast overlap is possible
+        overlap = ensure_tuple_rep(self.overlap, spatial_ndim)
+        # keep overlap only in patching dimensions
```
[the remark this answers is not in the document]
drbeh: The input's shape is `BCHWD` but patches are created only along `HWD` dimensions. I called these dimensions patching dimensions.

monai/inferers/splitter.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def _get_valid_offset(self, spatial_shape, spatial_ndim, patch_size):
+        offset = ensure_tuple_rep(self.offset, spatial_ndim)
+        for off, ps, ins in zip(offset, patch_size, spatial_shape):
+            if off < 0 and not self.pad_mode:
+                raise ValueError(
+                    f"Negative `offset` ({off}) requires a valid padding mode, "
+                    f"but `pad_mod` is set to {self.pad_mode}."
+                )
+            if off < -ps:
```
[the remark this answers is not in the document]
drbeh: Since `ps` is a positive number `off < 0 and off < -ps` is equivalent of `off < -ps`.

Nic-Ma:
Hi @drbeh ,

Sorry for the late response, I will also try to review this big PR ASAP tomorrow.

Thanks.

monai/inferers/inferer.py
```diff
@@ -64,6 +67,190 @@ def __call__(self, inputs: torch.Tensor, network: Callable, *args: Any, **kwargs
...
+            Defaults to None, where all the keys are taken if output is a dictionary.
+    """
+
+    def __init__(
+        self,
+        splitter: Splitter | None = None,
+        merger: Merger | Sequence[Merger] | None = None,
+        batch_size: int = 1,
+        pre_processor: Callable | None = None,
+        post_processor: Callable | None = None,
```
Nic-Ma, line 93: I would suggest to use "preprocessing" and "postprocessing" to be consistent with other modules:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/engines/workflow.py#L113

Thanks.

monai/inferers/inferer.py
```diff
@@ -64,6 +67,190 @@ def __call__(self, inputs: torch.Tensor, network: Callable, *args: Any, **kwargs
...
+            patches: a tensor or list of tensors
+
+        Yields:
+            A batch of patches (torch.Tensor or MetaTensor), a sequence of location tuples, and the batch size
+        """
+        if isinstance(patches, MetaTensor):
+            total_size = len(patches)
+            for i in range(0, total_size, self.batch_size):
+                batch_size = min(self.batch_size, total_size - i)
+                yield patches[i : i + batch_size], patches[i : i + batch_size].meta[PatchKeys.LOCATION], batch_size  # type: ignore
```
Nic-Ma, line 140: Do we expect every general patch MetaTensor to have the `LOCATION` field in the `meta`?
CC @wyli 

Thanks.

monai/inferers/inferer.py
```diff
@@ -64,6 +67,190 @@ def __call__(self, inputs: torch.Tensor, network: Callable, *args: Any, **kwargs
...
+                supports callables such as ``lambda x: my_torch_model(x, additional_config)``
+            args: optional args to be passed to ``network``.
+            kwargs: optional keyword args to be passed to ``network``.
+
+        """
+        is_merger_initialized = False
+        ratios = []
+        patch_locations: Iterable[tuple[torch.Tensor, Sequence[int]]]
+        if self.splitter is None:
+            patch_locations = inputs
```
Nic-Ma, line 232: Here the `patch_locations` doesn't match the expected typehint of `_batch_sampler`?

Thanks.

monai/inferers/splitter.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+
+    def __init__(self, patch_size: Sequence[int] | int, device: torch.device | str | None = None) -> None:
+        self.patch_size = patch_size
+        self.device = device
+
+    @abstractmethod
+    def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
+        """
+        Split the input image (or batch of images) into patches.
```
Nic-Ma, line 46: Could you please help add more description about the expected output of this function?

Thanks.

monai/inferers/inferer.py
```diff
@@ -64,6 +67,190 @@ def __call__(self, inputs: torch.Tensor, network: Callable, *args: Any, **kwargs
...
+        else:
+            patch_locations = self.splitter(inputs)
+        for patches, locations, batch_size in self._batch_sampler(patch_locations):
+            # run inference
+            outputs = self._run_inference(network, patches, *args, **kwargs)
+            # initialize the mergers
+            if not is_merger_initialized:
+                ratios = self._initialize_mergers(inputs, outputs, patches, batch_size)
+                is_merger_initialized = True
```
Nic-Ma, line 241: May I know why we don't always call `_initialize_mergers` here?
This `is_merger_initialized` is not thread-safe.

Thanks.

monai/inferers/inferer.py
drbeh: It is not a requirement for using `PatchInferer` (for almost all use cases). This is only required if all the following are satisfied:
1) the data is already patched (for instance if the data is stored in patches on the disk)
2) so the user don't want to use any `Spliter` in `PatchInferer`
3) and the user wants to load all the patches into a `MetaTensor`.

In this case, there should be a standard way to have the MetaTensor of patches being represented, so the user should use `PatchKeys.LOCATION`.
Otherwise, the user still can define the input as a list of (MetaTensor, location), which is generic and does not need `PatchKeys.LOCATION`.

monai/inferers/inferer.py
```diff
@@ -64,6 +67,190 @@ def __call__(self, inputs: torch.Tensor, network: Callable, *args: Any, **kwargs
         raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
 
 
+class PatchInferer(Inferer):
+    """
+    Inference on patches instead of the whole image based on Splitter and Merger.
+    This splits the input image into patches and then merge the resulted patches.
+
+    Args:
+        splitter: a `Splitter` object that split the inputs into patches. Defaults to None.
```
ericspod, line 76:
```suggestion
        splitter: a `Splitter` object that splits the inputs into patches. Defaults to None in which case inputs will be traversed in the batch dimension.
```
We should explain what the default values imply here, hopefully I have it right. Could this argument also just be typed as a callable?

monai/inferers/inferer.py
```diff
@@ -64,6 +67,190 @@ def __call__(self, inputs: torch.Tensor, network: Callable, *args: Any, **kwargs
...
+        # splitter
+        if splitter is not None and not callable(splitter):
+            raise TypeError(f"'splitter' should be a callable object, {type(splitter)} is given.")
+        self.splitter = splitter
+
+        # mergers
+        self.mergers: Sequence[Merger] = (AvgMerger(),) if merger is None else ensure_tuple(merger)
+        for m in self.mergers:
+            if not isinstance(m, Merger):
+                raise TypeError(f"'merger' should be a `Merger` object, {type(m)} is given.")
```
ericspod, line 107:
```suggestion
        if any(not isinstance(m, Merger) for m in self.mergers):
            raise TypeError(f"`merger` should be a `Merger` object or sequence thereof, object of type {type(m)} given.")
```

monai/inferers/inferer.py
drbeh: Thanks, that's correct! As the only interface of `Splitter` is `__call__`, it can be any Callable that act on the input. The input verification already checks for it only to be callable.

monai/inferers/inferer.py
drbeh: Since we want to include `type(m)` in the error message, it might be better to keep it as it to avoid going over mergers to find the erroneous one.

monai/inferers/inferer.py
drbeh: `_initialize_mergers` only need to be called for the first iteration of this loop. Since threading is happing in `dataloader` (or anywhere before calling `PatchInferer`) The entire loop will be run on a single thread, so `is_merger_initialized` should not make it thread unsafe.
Having said that you are right that it is NOT thread safe, and that is because I am using `Merger` object as the input instead of `Merger` class to be instantiated by each thread (but with classes passing arguments is cumbersome). I'll see how to address this but that was a good point anyways. Thanks

monai/inferers/merger.py
```diff
@@ -0,0 +1,186 @@
...
+            location: a tuple/list giving the top left location of the patch in the original image.
+
+        Raises:
+            NotImplementedError: When the subclass does not override this method.
+
+        """
+        raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
+
+    @abstractmethod
+    def finalize(self) -> torch.Tensor:
```
ericspod, line 69: This should state what the expected return value is meant to be. It's clear in `AvgMerger` what it does but it should say here in the general case.

monai/inferers/merger.py
```diff
@@ -0,0 +1,186 @@
...
+from collections.abc import Sequence
+
+import torch
+
+from monai.utils import ensure_tuple_size
+
+__all__ = ["Merger", "AvgMerger"]
+
+
+class Merger(ABC):
```
ericspod, line 24: It's not clear from the docstrings how one would go about implementing a merger class given the methods that need to be implemented. It would need to be explicit here what each method does in more detail and how they get called together. If possible it would be better to provide a default implementation of the methods given so that one need implement only those methods that are totally necessary.

monai/inferers/merger.py
```diff
@@ -0,0 +1,186 @@
...
+        Args:
+            output_shape: the shape of output tensor (B,C,H,W,[D])
+        """
+        # overwrite output_shape if the self.output_shape is set
+        if self.output_shape:
+            output_shape = self.output_shape
+        if output_shape is None:
+            raise ValueError("A valid `output_shape` should be set in the init or initialize method.")
+        self.values = torch.zeros(output_shape, dtype=self.dtype, device=self.device)
+        self.counts = torch.zeros(output_shape, dtype=torch.uint8, device=self.device)
```
ericspod, line 123: Is uint8 enough? Is there ever a situation to use this class with more than 255 values to merge?

monai/inferers/inferer.py
Nic-Ma: OK, now I understand the logic.
Could you please help add some check here for `meta[PatchKeys.LOCATION]`? If users use `MetaTensor` to represent the data by mistake, better to give clear tips.

Thanks.

monai/inferers/merger.py
drbeh: This is essentially the count of overlapping regions, so typycally it should be fine but I can imagine a scenario that someone run infernce with patches of 1000x1000 with 0.999 overlap! I'll change it. Thanks

drbeh: Do you think `torch.int16` (max=32767) would suffice? I want to use the smallest practical dtype as it supposed to be a very large tensor.

monai/inferers/inferer.py
drbeh: I added the check and also unittest. Thanks

monai/inferers/splitter.py
drbeh: Just did!

monai/inferers/inferer.py
drbeh: No `PatchInferer` accepts `merger_cls` (Merger class) instead of `merger` (Merger object), so each call to `PatchInferer` will instantiate a new merger. This should make `PatchInferer` thread safe.

monai/inferers/inferer.py
drbeh: I updated the checks. The splitter should be `Splitter` but since `Splitter` is basically a callable that get tensor and returns pairs of (patch, locations), any other callable that have the same behavior can be used (but not any callable). So to be more clear about it, I kept `Splitter` as the type hint but included `Callable` too. 
```python
splitter: Splitter | Callable | None = None,
```

monai/inferers/inferer.py
drbeh: I changed the input in the last update that makes this irrelevant.

monai/inferers/merger.py
drbeh: I updated the based class and also added more description on what needs to be implemented in concrete classes.

monai/inferers/merger.py
drbeh: I made it customizable but default to `torch.int16`.

monai/inferers/merger.py
drbeh: I update it to be Any for the base class and added more description what is expected to be returned in general.

monai/inferers/inferer.py
drbeh: updated it. thanks

drbeh:
@KumoLiu @Nic-Ma @ericspod, I tried to address all your comments. Please let me know if I have missed anything and if you have any other comments. Thanks

monai/inferers/merger.py
```diff
@@ -0,0 +1,162 @@
...
+
+        Args:
+            values: a tensor of shape BCHW[D], representing the values of inference output
+            location: a tuple/list giving the top left location of the patch in the output.
+
+        Raises:
+            NotImplementedError: When the subclass does not override this method.
+
+        """
+        raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
```
Nic-Ma, line 58: As you already added `@abstractmethod` decorator, this line code will never be executed?

Thanks.

monai/inferers/merger.py
```diff
@@ -0,0 +1,162 @@
...
+        self.is_finalized = False
+
+    @abstractmethod
+    def aggregate(self, values: torch.Tensor, location: Sequence[int]) -> None:
+        """
+        Aggregate values for merging.
+        This method is being called in a loop and should add values to their corresponding location in the merged output results.
+
+        Args:
+            values: a tensor of shape BCHW[D], representing the values of inference output
```
Nic-Ma, line 51: missed "." at the end of the string.

Thanks.

monai/inferers/merger.py
```diff
@@ -0,0 +1,162 @@
...
+        device: the device where Merger tensors should reside.
+    """
+
+    def __init__(self, output_shape: Sequence[int] | None = None, device: torch.device | str | None = None) -> None:
+        self.output_shape = output_shape
+        self.device = device
+        self.is_finalized = False
+
+    @abstractmethod
+    def aggregate(self, values: torch.Tensor, location: Sequence[int]) -> None:
```
Nic-Ma, line 45: Seems this `aggregate()` API is called multiple times, each time to append 1 patch, and the `finalize()` API actually does the real "aggregation" work.
I would suggest to change it to "append" and "finalize", it's also consistent with other MONAI existing module:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/metrics/cumulative_average.py#L80
What do you think?

Thanks.

Nic-Ma (commented):
Thanks for the quick update. I put some minor comments inline, it overall looks good to me.

monai/inferers/inferer.py
```diff
@@ -64,6 +67,227 @@ def __call__(self, inputs: torch.Tensor, network: Callable, *args: Any, **kwargs
...
+            Defaults to None, where all the keys are used.
+        merger_kwargs: arguments to be passed to `merger_cls` for instantiation.
+            `output_shape` is calculated automatically based on the input shape and
+            the output patch shape unless it is passed here.
+    """
+
+    def __init__(
+        self,
+        splitter: Splitter | Callable | None = None,
+        merger_cls: type[Merger] = AvgMerger,
```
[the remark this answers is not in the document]
drbeh: Sure, will implement that.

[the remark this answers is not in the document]
drbeh: Sure, will implement that.

monai/inferers/merger.py
drbeh: It can possibly be called if someone call it from the concrete class using `super()`. I agree that removing it won't cause any harm but the rest of MOANI is like that. Shall we keep it as is and revisit for all the monai base clsses that use both `@abstractmethod` and `raise NotImplementedError`?

https://docs.python.org/3/library/abc.html#abc.abstractmethod
"Note: Unlike Java abstract methods, these abstract methods may have an implementation. This implementation can be called via the `super()` mechanism from the class that overrides it. This could be useful as an end-point for a super-call in a framework that uses cooperative multiple-inheritance."

monai/inferers/merger.py
drbeh: I see your point but `aggregate` as a verb means ["to collect or gather into a mass or whole"](https://www.merriam-webster.com/dictionary/aggregate), so it correctly define the process of collecting values into a result tensor.

Also consider that this API is common for other [not yet implemented] mergers so these methods needs to be generalizable enough. For instance, for `MaxMerger` that aggregates the maximum value instead of average, all the work is happening in `aggregate` method and `finalize` won't do any thing, so I believe replacing it with append and aggregate won't be appropriate.

monai/inferers/merger.py
Nic-Ma: OK, let's leave it as other MONAI modules.

Thanks.

monai/inferers/inferer.py
```diff
@@ -120,7 +121,13 @@ def __init__(
                 )
         self.splitter = splitter
 
-        # mergers
+        # merger
+        if isinstance(merger_cls, str):
+            valid_merger: type[Merger]
+            valid_merger, merger_found = optional_import("monai.inferers.merger", name=merger_cls)  # search mergers
+            if not merger_found:
+                raise ValueError(f"The requested merger ['{merger_cls}'] does not exist.")
```
Nic-Ma, line 129: Could you please help add the support for user customized `Merger`? Refer to:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/io/array.py#L206

Thanks.

drbeh: Sure

monai/inferers/merger.py
Nic-Ma: Hi @ericspod @wyli ,

What do you think about this naming question? I am not very sure about it.

Thanks in advance.

ericspod: I'm fine with the naming as-is. The `aggregate` call in the metrics is collecting data across processes, similarly here `aggregate` means to collect data from sources. The `finalize` part could be a no-op in some cases like @drbeh mentions so `aggregate` then would be doing the real work of combining results like the name would imply.

Nic-Ma: OK, let's keep the current implementation.

Thanks for the discussion.

monai/utils/enums.py
```diff
@@ -462,15 +462,25 @@ def get_sort_fn(sort_fn):
...
+    The keys to be used for metadata of patches extracted from any kind of image
     """
 
     LOCATION = "location"
-    LEVEL = "level"
     SIZE = "size"
     COUNT = "count"
+
+
+class WSIPatchKeys(StrEnum):
```
Nic-Ma, line 475: Can we inherit from `PatchKeys` here?

Thanks.

Nic-Ma (approved):
Looks good to me, thanks for the quick update.

monai/utils/enums.py
drbeh: Unfortunately, enums cannot be inherited.

PR #5916, opened 2023-01-30
4855 5860 update the pending transform utilities

### LINKED ISSUE #5509: LoadImageD and MetaTensor
It's more of question to clarify, not a bug. can you help clarify plz

- LoadImage() and LoadImaged() always return MetaTensor. Is it by design?  I'm okay either way, just want to clarify. So we can't return the original data format ?

- line 281 `return img, img.meta if isinstance(img, MetaTensor) else meta_data`. The img seems to be always a MetaTensor, is that check necessary?

https://github.com/Project-MONAI/MONAI/blob/c27ab360212ad122f5ca41fc69b8c2f47676b809/monai/transforms/io/array.py#L273-L281

- inside of MetaTensor.ensure_torch_and_prune_meta(), we seems to do conversion to Tensor twice on lines 487 and 504.  on 487 we `img = convert_to_tensor(im)` and `return MetaTensor(img, meta=meta) ` will internally call torch.as_tensor on img.  

https://github.com/Project-MONAI/MONAI/blob/c27ab360212ad122f5ca41fc69b8c2f47676b809/monai/data/meta_tensor.py#L487-L504

Thank you

wyli:
one of the sub-PRs from https://github.com/Project-MONAI/MONAI/pull/5860

also fixes #5509, by reviewing the metetensor copying in loadimage

### Description
This PR mainly to enhance the `apply_transforms`, `resample`, `push_transform` APIs
to get prepared for lazy resampling.

### Types of changes


monai/data/meta_tensor.py
```diff
@@ -523,10 +523,10 @@ def ensure_torch_and_prune_meta(
             By default, a `MetaTensor` is returned.
             However, if `get_track_meta()` is `False`, a `torch.Tensor` is returned.
         """
-        img = convert_to_tensor(im)  # potentially ascontiguousarray
+        img = convert_to_tensor(im, track_meta=get_track_meta() and meta is not None)  # potentially ascontiguousarray
```
KumoLiu, line 526: Hi @wyli, I'm a little confused about this line, if im is a MetaTensor, `get_track_meta()` is `True` and meta is `None`, then the meta information will be lost. Is that right?

wyli: Yes, setting meta=None essentially skips the metatensor creation. In that case it'll return a torch tensor.

KumoLiu: Thanks for the explanation, but the case I considering is that when input is already `MetaTensor`, then I can't preserve the affine information unless I reassign it in the meta dict. Maybe it can be `track_meta=get_track_meta() or meta is not None`? Or I misunderstood it?
```
inp = MetaTensor([1,2,3], meta={'affine': affine, 'space': "LAS"})
out = MetaTensor.ensure_torch_and_prune_meta(inp, meta=None) --> affine will be lost
```

monai/transforms/lazy/functional.py
```diff
@@ -11,51 +11,90 @@
```
Nic-Ma, line 11: Do we really need this "lazy" layer folder?
In this case, we will also have some "instant" version `functional.py`?

Thanks.

monai/data/meta_tensor.py
wyli: that's the current `dev` branch's behaviour as well, I didn't try to change it.  I think it makes sense to use `None` to indicate no meta (regular tensor), and `{}` to indicate an empty meta (default affine)

monai/transforms/lazy/functional.py
wyli: maybe we'll need this when we go for a lazy compose implementation... I agree we should review this folder structure before we release the lazy resampling feature.

monai/data/meta_tensor.py
KumoLiu: Ok, got it, thanks!

monai/data/meta_obj.py
```diff
@@ -107,27 +107,35 @@ def flatten_meta_objs(*args: Iterable):
     @staticmethod
     def copy_items(data):
         """returns a copy of the data. list and dict are shallow copied for efficiency purposes."""
+        if isinstance(data, (bool, int, float, str, type(None))):
```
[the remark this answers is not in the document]
wyli: looking at the cpython source code it seems this condition is robust https://github.com/python/cpython/blob/909a6746939ea1d09fab21f26b558cfd7e3e29a0/Lib/copy.py#L104-L108 I'll change to use part of that

monai/data/meta_tensor.py
wyli: thanks, I updated the docstring to make it clear.

monai/data/meta_obj.py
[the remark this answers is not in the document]
Nic-Ma: We may need this check in multiple places, better to define a utility function for it?

Thanks.

monai/transforms/lazy/functional.py
```diff
@@ -11,51 +11,90 @@
...
+from monai.utils import LazyAttr
 
 __all__ = ["apply_transforms"]
 
 
-def apply_transforms(data: torch.Tensor | MetaTensor, pending: list | None = None):
+def apply_transforms(
+    data: torch.Tensor | MetaTensor,
+    pending: list | None = None,
+    mode: str | None = None,
```
yiheng-wang-nv, line 35: here should the type be: `str | int | None`?

PR #5940, opened 2023-02-05
Fully type annotate partially typed functions

Shadow-Devil:
Part of #5884.

### Description

Completes following modules:
- [x] _extensions
- [x] auto3dseg
- [x] config
- [x] engines
- [x] fl
- [x] handlers
- [x] inferers
- [x] losses

Since they all were quite easy to fix, I bundled them into this PR.

### Types of changes


monai/auto3dseg/analyzer.py
```diff
@@ -571,13 +571,13 @@ def __call__(self, data: list[dict]):
...
-            return ValueError(f"Callable {self.__class__} requires list inputs")
+            raise ValueError(f"Callable {self.__class__} requires list inputs")
 
         if len(data) == 0:
-            return ValueError(f"Callable {self.__class__} input list is empty")
+            raise ValueError(f"Callable {self.__class__} input list is empty")
 
         if self.stats_name not in data[0]:
-            return KeyError(f"{self.stats_name} is not in input data")
+            raise KeyError(f"{self.stats_name} is not in input data")
```
Shadow-Devil, line 580: I often found this `return` of an Error inside the auto3dseg module, which didn't seem right.
I changed it to `raise` the Error instead.

monai/auto3dseg/utils.py
```diff
@@ -270,10 +270,10 @@ def verify_report_format(report: dict, report_format: dict):
             else:
                 return False
 
-        return True
+    return True
```
Shadow-Devil, line 273: I think this could have been a bug since the loop would always finish in the first iteration.

monai/auto3dseg/analyzer.py
wyli: Agreed, those look like typos cc @mingxin-zheng

wyli (approved):
I'm merging this if all testing status are green. Please help confirm the auto3dseg changes in case I miss something @mingxin-zheng

mingxin-zheng:
Thanks for the fixes!

I have a question about the type cast. One problem I met in `auto3dseg` was that some outputs from functions can be `numpy` / `torch` arrays or `MetaTensor`. So I did not write the annotations because none of it is correct. It seems the solution in the PR is to make them all `np.ndarray`. Can I do the some practice for similar situations in the future?

Shadow-Devil:
I added these return types since it was stated in the comment that it will return numpy arrays:
https://github.com/Shadow-Devil/MONAI/blob/2a39e6ff3a910a5f02ae1bef2d32e7f59289c5e8/monai/auto3dseg/utils.py#L49-L66

To fix those type issues you have two options:
- You can either use this `cast`, which only fixes type checking and in runtime does nothing
- or you can add a comment `# type: ignore[return-value]` or `# type: ignore[no-any-return]`, depending on the mypy errorcode

See also: https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-return-value-is-compatible-return-value and https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-function-does-not-return-any-value-no-any-return

PR #5965, opened 2023-02-09
Added callable options for iteration_log and epoch_log in StatsHandler

vfdev-5:
Fixes #5964

### Description

Added callable options for iteration_log and epoch_log in StatsHandler.

Ref: https://github.com/Project-MONAI/MONAI/discussions/5958#discussioncomment-4912997

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [x] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick --unittests  --disttests`.
- [x] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

PR #5976, opened 2023-02-10
Added callable options for iteration_log and epoch_log in TensorBoard and MLFlow

vfdev-5:
Follow-up PR on extending `iteration_log`/`epoch_log` functionality

### Description

> Could you please help also enable this feature for other very similar handlers? Like:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/handlers/tensorboard_handlers.py#L94
And:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/handlers/mlflow_handler.py#L101

- https://github.com/Project-MONAI/MONAI/pull/5965#pullrequestreview-1292534057

### Types of changes


monai/handlers/tensorboard_handlers.py
```diff
@@ -91,8 +91,8 @@ def __init__(
         self,
         summary_writer: SummaryWriter | SummaryWriterX | None = None,
         log_dir: str = "./runs",
-        iteration_log: bool = True,
-        epoch_log: bool = True,
+        iteration_log: bool | Callable[[Engine, int], bool] = True,
+        epoch_log: bool | Callable[[Engine, int], bool] = True,
         epoch_event_writer: Callable[[Engine, Any], Any] | None = None,
         epoch_interval: int = 1,
```
Nic-Ma, line 97: I would suggest to add "deprecated" mark for `epoch_interval` and `iteration_interval` args, refer to:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/image_reader.py#L883

Thanks.

Nic-Ma (approved):

monai/handlers/tensorboard_handlers.py
Nic-Ma: Hi @vfdev-5 ,

Could you please help update it? Others look good to me.

Thanks.

vfdev-5: Done

monai/handlers/tensorboard_handlers.py
```diff
@@ -87,12 +87,14 @@ class TensorBoardStatsHandler(TensorBoardHandler):
 
     """
 
+    @deprecated_arg("epoch_interval", since="1.1", removed="1.3")
+    @deprecated_arg("iteration_interval", since="1.1", removed="1.3")
```
Nic-Ma, line 91: Could you please help add some msg_suffix  message here to help user understand?
CC @wyli 

Thanks.

vfdev-5: I added `msg_suffix` as we discussed in the slack but it gives a redundant message like:
```
monai.handlers.tensorboard_handlers TensorBoardStatsHandler.__init__:iteration_interval: Argument `iteration_interval` has been deprecated since version 1.1. It will be removed in vers
ion 1.3. ``iteration_interval`` has been deprecated and will be removed soon. 
```
So, that's why I omitted it. Right now we can't suggest any replacement for the arg.

wyli: thanks both, that looks like a bug in the ci/cd for deprecated_arg, I'll fix it and merge this PR soon.

vfdev-5: Maybe related: https://github.com/python-versioneer/python-versioneer/issues/206

wyli: yes, the version is not well defined in this case, and this could happen in a few use cases as well. I think `deprecated_arg` should assume the arg is deprecated but not removed when monai version is unknown..

PR #5986, opened 2023-02-13
5915-add-export-trt-api

### LINKED ISSUE #5915: Add the trt_export api for bundles
Currently, the MONAI bundle has an `ckpt_export` [API](https://github.com/Project-MONAI/MONAI/blob/50c3f328cb98e4f0c4e6a38a04a98badf0961c18/monai/bundle/scripts.py#L907)  to convert a torch module to a script module.  

In order to leverage the benefit of Nvidia GPUa to get a better performance, a torch module needs to be converted to a trt engine through Torch-TensorRT. This will accelerate models' inference on Nvidia GPUs.

 An `trt_export` API like the `ckpt_export` would be better to standardize this process. 

Add an `trt_export` API to MONAI bundle to export models to trt engines and update the readme file in bundles with the trt converting command.

**Additional context**

binliunls:

Fixes #5915  .

### Description

Add an `trt_export` API to MONAI bundle to export models to trt engines and update the readme file in bundles with the trt converting command.

### Types of changes

- [ ] Add a trt_export api for MONAI bundle.
- [ ] Test this api with current bundles in MONAI model zoo.
- [ ] Try to integrate this api to CI/CD test.


monai/bundle/scripts.py
```diff
@@ -1019,6 +1053,148 @@ def ckpt_export(
     logger.info(f"exported to TorchScript file: {filepath_}.")
 
 
+def trt_export(
```
Nic-Ma, line 1056: I see most of the code in this `trt_export` is same as `trt_export`, so I think you can define a `_export(converter: callable)` function, then define `ckpt_export ` and `trt_export` to call `_export` with different `converter` callback. The converter can take `parser` and `model` as parameters.
What do you think?

Thanks.

monai/bundle/scripts.py
```diff
@@ -822,6 +822,56 @@ def verify_metadata(
...
+            bundle, default to `meta_#network_data_format`.
+        p: power factor to generate input data shape if dim of expected shape is "x**p", default to 1.
+        n: multiply factor to generate input data shape if dim of expected shape is "x*n", default to 1.
+        any: specified size to generate input data shape if dim of expected shape is "*", default to 1.
+
+    Return:
+        a sequence contains the input and output infomation.
+    """
+    if not parser:
+        raise AttributeError("Error parser to parse input shape.")
```
yiheng-wang-nv, line 844: Could the error message here be a bit more informative? Like:
```
    if not isinstance(ConfigParser):
        raise AttributeError(f"'parser' should be a ConfigParser, got {type(parser)}.")
```

monai/bundle/scripts.py
```diff
@@ -822,6 +822,56 @@ def verify_metadata(
...
+    inputs_shape = []
+    try:
+        prefix_key = f"{prefix}#inputs"
+        # prefix_key = key
```
yiheng-wang-nv, line 848: remove?

monai/bundle/scripts.py
binliunls: Thank you for the review. I have updated this part.

monai/bundle/scripts.py
binliunls: Thank you for the review. I have updated this part.

monai/bundle/scripts.py
```diff
@@ -822,6 +822,56 @@ def verify_metadata(
...
+        input_dtype = get_equivalent_dtype(parser[key], torch.Tensor)
+
+        prefix_key = f"{prefix}#outputs"
+        key = f"{prefix_key}#pred#num_channels"
+        output_channels = parser[key]
+        key = f"{prefix_key}#pred#dtype"
+        output_dtype = get_equivalent_dtype(parser[key], torch.Tensor)
+    except KeyError as e:
+        raise KeyError(f"Failed to parse due to missing expected key in the config: {key}.") from e
+    return [input_channels, input_spatial_shape, input_dtype, output_channels, output_dtype, inputs_shape]
```
Nic-Ma, line 872: Please return tuple instead of list.

Thanks.

monai/bundle/scripts.py
```diff
@@ -822,6 +822,56 @@ def verify_metadata(
...
+    Return:
+        a sequence contains the input and output infomation.
+    """
+    if not isinstance(parser, ConfigParser):
+        raise AttributeError(f"'parser' should be a ConfigParser, got {type(parser)}.")
+
+    inputs_shape = []
+    try:
+        prefix_key = f"{prefix}#inputs"
+        inputs_info_dict = parser[prefix_key]
```
Nic-Ma, line 849: I don't quite understand what's this `inputs_info_dict` mean here?

Thanks.

monai/bundle/scripts.py
```diff
@@ -822,6 +822,56 @@ def verify_metadata(
...
+        key = f"{prefix_key}#image#dtype"
+        input_dtype = get_equivalent_dtype(parser[key], torch.Tensor)
+
+        prefix_key = f"{prefix}#outputs"
+        key = f"{prefix_key}#pred#num_channels"
+        output_channels = parser[key]
+        key = f"{prefix_key}#pred#dtype"
+        output_dtype = get_equivalent_dtype(parser[key], torch.Tensor)
+    except KeyError as e:
+        raise KeyError(f"Failed to parse due to missing expected key in the config: {key}.") from e
```
Nic-Ma, line 871: I feel this `KeyError` is too restrictive, maybe better to change all the `parser[key]` to `parser.get(key)`? if None, just return None for this part. For example, an output can be: `(3, (512, 512), float, None, None, None)`.
What do you think?

Thanks.

monai/bundle/scripts.py
```diff
@@ -822,6 +822,56 @@ def verify_metadata(
...
 
+def _get_net_io_info(
+    parser: ConfigParser | None = None, prefix: str = "_meta_#network_data_format", p: int = 1, n: int = 1, any: int = 1
+) -> list[Any]:
+    """
+    Get the input and output information defined in the metadata.
+
+    Args:
+        parser: a parser of the given bundle.
+        prefix: a prefix for the input and output keys, which will be combined as `prefix#input` and
```
Nic-Ma, line 833: I would suggest to use word "id" instead of "key" for the bundle config index, be consistent with all the other MONAI code.

Thanks.

monai/bundle/scripts.py
```diff
@@ -822,6 +822,56 @@ def verify_metadata(
...
+            `prefix#output` to parse the input and output information in the `metadata.json` file of
+            bundle, default to `meta_#network_data_format`.
+        p: power factor to generate input data shape if dim of expected shape is "x**p", default to 1.
+        n: multiply factor to generate input data shape if dim of expected shape is "x*n", default to 1.
+        any: specified size to generate input data shape if dim of expected shape is "*", default to 1.
+
+    Return:
+        a sequence contains the input and output infomation.
```
Nic-Ma: Please clearly describe the return data, because it's very long and the data order is important.

Thanks.

monai/bundle/scripts.py
Nic-Ma: What's the last item "inputs_shape" mean?

Thanks.

monai/networks/utils.py
```diff
@@ -579,13 +583,20 @@ def convert_to_torchscript(
         device: target device to verify the model, if None, use CUDA if available.
         rtol: the relative tolerance when comparing the outputs of PyTorch model and TorchScript model.
         atol: the absolute tolerance when comparing the outputs of PyTorch model and TorchScript model.
+        is_trace: whether to use `torch.jit.trace` to get the torchscript model.
```
Nic-Ma, line 586: "get" -> "export".

Thanks.

monai/networks/utils.py
```diff
@@ -560,6 +563,7 @@ def convert_to_torchscript(
     device: torch.device | None = None,
     rtol: float = 1e-4,
     atol: float = 0.0,
+    is_trace: bool = False,
```
Nic-Ma, line 566: Maybe "use_trace" is more clear a name?

Thanks.

monai/networks/utils.py
```diff
@@ -615,6 +626,101 @@ def convert_to_torchscript(
     return script_module
 
 
+def convert_to_trt(
+    model: nn.Module,
+    precision: str,
+    inputs_shape: Sequence[Sequence[int]],
+    dynamic_batchsize: Sequence[Sequence[int]],
```
Nic-Ma, line 633: I don't quite understand why these 2 args are "sequence in sequence" typehint?

Thanks.

monai/networks/utils.py
```diff
@@ -615,6 +626,101 @@ def convert_to_torchscript(
...
+        model: source PyTorch model to save.
+        precision: the weight precision of converted TensorRT engine based torchscript models. Should be 'fp32' or 'fp16'.
+        inputs_shape: the inputs' shape that is used to generate random input during the convert. Should be a list with elements
+            like [N, C, H, W] or [N, C, H, W, D].
+        dynamic_batchsize: a list contains three number list elements to define the batch size range of inputs for the
+            converted model. In each element, the three number means `MIN_BATCH, OPT_BATCH, MAX_BATCH` in order.
+        ir: the intermediate representation way to transform a pytorch module to a TensorRT engine based torchscript. Could
+            be choose from `(script, trace)`.
+        verify: whether to verify the input and output of TorchScript model.
+            if `filename_or_obj` is not None, load the saved TorchScript model and verify.
```
Nic-Ma, line 654: The docstring is not correct, should be TensorRT instead of TorchScript.
Please double-confirm all the other doc-strings.

Thanks.

monai/networks/utils.py
```diff
@@ -615,6 +626,101 @@ def convert_to_torchscript(
...
+        device: target device to verify the model, if None, use CUDA if available.
```
Nic-Ma, line 655: What will happen if no GPU device? Please ensure to warn the user or gracefully exit.

Thanks.

monai/bundle/scripts.py
```diff
@@ -921,6 +962,78 @@ def verify_net_in_out(
...
 
+def _export(
+    converter: Callable,
+    parser: ConfigParser,
+    net_id: str,
+    filepath: str,
+    ckpt_file: str,
+    config_file: str,
+    key_in_ckpt: str,
+    update_args: dict,
```
Nic-Ma, line 973: I think no need this `update_args` as the passed `parser` should be ready here.

Thanks.

monai/bundle/scripts.py
binliunls: It's a list of shapes of inputs. For example, if you have a model with two inputs, whose shapes are (1, 3, 224, 224) and (1, 3, 256, 256), then this item should be like `[[1,3,224,224], [1,3,256,256]]`.

monai/bundle/scripts.py
binliunls: This parameter is trying to parse all the parameters in `prefix#inputs` in case there are keys in addition to `image` llike the mednist gan with `latent`.

monai/bundle/scripts.py
binliunls: Ok, I will update this part.

monai/bundle/scripts.py
binliunls: Ok, will update this part.

monai/bundle/scripts.py
binliunls: Ok, thanks for your advice.

monai/networks/utils.py
binliunls: Ok, will update this later.

monai/networks/utils.py
binliunls: This is a sequence in sequence type because there may be more than one inputs to the network, therefore I suppose each input should has an input shape and corresponding batch range.

monai/networks/utils.py
binliunls: Thanks, will update this part.

monai/networks/utils.py
binliunls: There is a `torch.cuda.is_availabel` check in line 663, but the device defination in line 669 should be `device = device if device else torch.device("cuda")`. Will update this.

monai/bundle/scripts.py
binliunls: Ok, will delete it.

monai/bundle/scripts.py
```diff
@@ -822,6 +822,59 @@ def verify_metadata(
...
+        parser: a parser of the given bundle.
+        prefix: a prefix for the input and output id, which will be combined as `prefix#input` and
+            `prefix#output` to parse the input and output information in the `metadata.json` file of
+            bundle, default to `meta_#network_data_format`.
+        p: power factor to generate input data shape if dim of expected shape is "x**p", default to 1.
+        n: multiply factor to generate input data shape if dim of expected shape is "x*n", default to 1.
+        any: specified size to generate input data shape if dim of expected shape is "*", default to 1.
+
+    Returns:
+        input_channels: the channel number of the first input.
```
Nic-Ma, line 841: it's not the "first" input, just the item with key "image".

Thanks.

monai/bundle/scripts.py
```diff
@@ -822,6 +822,59 @@ def verify_metadata(
...
+    inputs_shape = []
+    prefix_key = f"{prefix}#inputs"
+    inputs_info_dict = parser.get(prefix_key)
+    for input_name in inputs_info_dict:
+        key = f"{prefix_key}#{input_name}#num_channels"
+        input_channels = parser.get(key)
+        key = f"{prefix_key}#{input_name}#spatial_shape"
+        input_spatial_shape = parser.get(key)
+        spatial_shape = _get_fake_spatial_shape(input_spatial_shape, p=p, n=n, any=any)
+        inputs_shape.append([1, input_channels, *spatial_shape])
```
Nic-Ma, line 860: Here you use `for loop` to iterate a dictionary by keys, but save the results in a list, how to guarantee the order?
To be honest, I feel this logic is very confusing, can we remove all this logic and only support the "image" input for this first version?

Thanks.

monai/bundle/scripts.py
```diff
@@ -982,41 +1092,136 @@ def ckpt_export(
...
+        _args,
+        "filepath",
+        "ckpt_file",
+        "config_file",
+        net_id="",
+        meta_file=None,
+        key_in_ckpt="",
+        precision="fp32",
+        inputs_shape=[],
+        dynamic_batchsize=[[1, 4, 8]],
```
Nic-Ma, line 1194: Why set default value for this arg?

Thanks.

monai/bundle/scripts.py
```diff
@@ -921,6 +965,72 @@ def verify_net_in_out(
...
+    """
+    net = parser.get_parsed_content(net_id)
+    if has_ignite:
+        # here we use ignite Checkpoint to support nested weights and be compatible with MONAI CheckpointSaver
+        Checkpoint.load_objects(to_load={key_in_ckpt: net}, checkpoint=ckpt_file)
+    else:
+        ckpt = torch.load(ckpt_file)
+        copy_model_state(dst=net, src=ckpt if key_in_ckpt == "" else ckpt[key_in_ckpt])
+
+    # convert to TorchScript model and save with metadata, config content
```
Nic-Ma, line 1005: This comment is out-date, not only "TorchScript" anymore.

monai/bundle/scripts.py
binliunls: Ok, thanks. Will update this.

monai/networks/utils.py
```diff
@@ -615,6 +626,103 @@ def convert_to_torchscript(
     return script_module
 
 
+def convert_to_trt(
+    model: nn.Module,
+    precision: str,
+    inputs_shape: Sequence[Sequence[int]],
```
Nic-Ma, line 632: I think this `inputs_shape` arg can be None when converting to `TorchScript` with `torch.jit.script()`, right?
If True, please add None support for this arg in this function and the `export` functions, and unit test for `None`.

Thanks.

monai/bundle/scripts.py
binliunls: Ok, let's focus on one input for this first version.

monai/bundle/scripts.py
binliunls: If users don't specifiy the `inputs_shape` parameter, the default batchsize `inputs_shape` parsed from the `metadata.json` would be 1, which doesn't support slidewindow inference. So by default, I try to make the dynamic batchsize range to be [1, 8] to fit some slidewindow cases.

monai/bundle/scripts.py
binliunls: Ok, thanks.

monai/bundle/scripts.py
Nic-Ma: It looks very hacky, someday we may see a bug: for unknown reason, some sliding window can work, some sliding window with batchsize > 8 can't work.. just because you potentially set a [1, 8] batchsize here. I would suggest to leave it for users to explicitly set.
What do you think?

Thanks.

monai/networks/utils.py
binliunls: Actually, the `inputs_shape` for `convert_to_trt` function cannot be `None`, since it is needed to generate the placeholder for the torch_tensorrt compiler.  But for the `trt_export` function, since we will parse the shape from the parser, it could be None, and the type hint is already supported `None` input. As for the `_export` function, it doesn't have this parameter explictly as input.

Nic-Ma: OK. Thanks.

monai/bundle/scripts.py
binliunls: Ok, I can update this parameter.

monai/bundle/scripts.py
Nic-Ma: I don't see you update these 2 lines.

monai/bundle/scripts.py
```diff
@@ -739,6 +739,51 @@ def verify_metadata(
...
+        p: power factor to generate input data shape if dim of expected shape is "x**p", default to 1.
+        n: multiply factor to generate input data shape if dim of expected shape is "x*n", default to 1.
+        any: specified size to generate input data shape if dim of expected shape is "*", default to 1.
+
+    Returns:
+        input_channels: the channel number of the first input.
+        input_spatial_shape: the spatial shape of the first input.
+        input_dtype: the data type of the `image` input.
+        output_channels: the channel number of the `image` output.
+        output_dtype: the data type of the `image` output.
```
Nic-Ma, line 762: output is not "image"

monai/bundle/scripts.py
```diff
@@ -739,6 +739,51 @@ def verify_metadata(
...
+    spatial_shape = _get_fake_spatial_shape(input_spatial_shape, p=p, n=n, any=any)
+    input_shape = [1, input_channels, *spatial_shape]
+
+    prefix_key = f"{prefix}#outputs"
+    key = f"{prefix_key}#pred#num_channels"
+    output_channels = parser.get(key)
+    key = f"{prefix_key}#pred#dtype"
+    output_dtype = get_equivalent_dtype(parser.get(key), torch.Tensor)
+
+    return (input_shape, input_channels, input_spatial_shape, input_dtype, output_channels, output_dtype)
```
Nic-Ma, line 784: I think `input_shape` is duplicated to `input_channels` and `input_spatial_shape`, maybe we can remove `input_shape` from this function?

Thanks.

Nic-Ma:
Hi @wyli @yiheng-wang-nv ,

Could you please help also review this PR?

Thanks in advance.

monai/bundle/scripts.py
binliunls: This is mainly for the `trt_export`. If necessary, I can move this logic to `trt_export` function.

Nic-Ma: I would suggest to move it to `trt_export`.

Thanks.

monai/bundle/scripts.py
```diff
@@ -739,6 +739,43 @@ def verify_metadata(
...
+    key = f"{prefix_key}#image#dtype"
+    input_dtype = get_equivalent_dtype(parser.get(key), torch.Tensor)
+
+    prefix_key = f"{prefix}#outputs"
+    key = f"{prefix_key}#pred#num_channels"
+    output_channels = parser.get(key)
+    key = f"{prefix_key}#pred#dtype"
+    output_dtype = get_equivalent_dtype(parser.get(key), torch.Tensor)
+
+    return (input_channels, input_spatial_shape, input_dtype, output_channels, output_dtype)
```
Nic-Ma, line 776: No need the "()" here.

Thanks.

monai/bundle/scripts.py
```diff
@@ -899,41 +993,140 @@ def ckpt_export(
...
+        parser.read_meta(f=meta_file_)
+
+    # the rest key-values in the _args are to override config content
+    for k, v in _args.items():
+        parser[k] = v
+
+    if not input_shape_:
+        input_channels, input_spatial_shape, _, _, _ = _get_net_io_info(parser=parser)
+        spatial_shape = _get_fake_spatial_shape(input_spatial_shape, p=1, n=1, any=1)
+        input_shape_ = [1, input_channels, *spatial_shape]
```
Nic-Ma, line 1114: Better to use `tuple` instead of `list`?

Thanks.

monai/networks/utils.py
```diff
@@ -560,6 +563,7 @@ def convert_to_torchscript(
     device: torch.device | None = None,
     rtol: float = 1e-4,
     atol: float = 0.0,
+    use_trace: bool = False,
```
Nic-Ma, line 566: I think you didn't add support for this new arg in the `ckpt_export` API? May also need to add `inputs`?

Thanks.

binliunls: I have added the `ir` and `input_shape` parameters to the `ckpt_export` API. Could you please take a look at it and tell me if it looks fine with you?

Thanks.

yiheng-wang-nv:
> Hi @wyli @yiheng-wang-nv ,
> 
> Could you please help also review this PR?
> 
> Thanks in advance.

Hi @Nic-Ma , sure. Today I'm working working on check this PR, as well as https://github.com/Project-MONAI/model-zoo/pull/325. I'm doing step-by-step experiments to see if there are any places should be modified.

monai/bundle/scripts.py
```diff
@@ -838,13 +866,81 @@ def verify_net_in_out(
...
+
+
 def ckpt_export(
     net_id: str | None = None,
     filepath: PathLike | None = None,
     ckpt_file: str | None = None,
     meta_file: str | Sequence[str] | None = None,
     config_file: str | Sequence[str] | None = None,
     key_in_ckpt: str | None = None,
+    ir: str | None = None,
```
Nic-Ma, line 942: I would suggest to use "use_track" directly instead of this TensorRT style naming.

Thanks.

Nic-Ma (approved):
Thanks for the quick update, looks good to me.

monai/bundle/scripts.py
```diff
@@ -899,41 +1008,144 @@ def ckpt_export(
...
+            value is always serialized in JSON format no matter the original file format is JSON or YAML.
+            it can be a single file or a list of files. if `None`, must be provided in `args_file`.
+        key_in_ckpt: for nested checkpoint like `{"model": XXX, "optimizer": XXX, ...}`, specify the key of model
+            weights. if not nested checkpoint, no need to set.
+        precision: the weight precision of the converted TensorRT engine based torchscript models. Should be 'fp32' or 'fp16'.
+        input_shape: the input shape that is used to convert the model. Should be a list like [N, C, H, W] or
+            [N, C, H, W, D].
+        ir: the intermediate representation way to transform a pytorch module to a TensorRT engine based torchscript. Could
+            be chosen from `(script, trace)`.
+        dynamic_batchsize: a list with three elements to define the batch size range of the input for the model to be converted.
```
yiheng-wang-nv, line 1068: `a sequence with three elements`

monai/networks/utils.py
```diff
@@ -615,6 +626,104 @@ def convert_to_torchscript(
...
+):
+    """
+    Utility to convert a model into a TensorRT engine based torchscript model with optional input / output data verification.
+
+    Args:
+        model: a source PyTorch model to convert.
+        precision: the weight precision of the converted TensorRT engine based torchscript models. Should be 'fp32' or 'fp16'.
+        input_shape: the input shape that is used to convert the model. Should be a list like [N, C, H, W] or
+            [N, C, H, W, D].
+        dynamic_batchsize: a list with three elements to define the batch size range of the input for the model to be converted.
```
yiheng-wang-nv, line 649: same here

monai/networks/utils.py
```diff
@@ -615,6 +626,104 @@ def convert_to_torchscript(
...
+    if not has_torch_tensorrt:
+        raise ImportError("Cannot find the torch_tensorrt module, please install it before converting.")
+
+    if not torch.cuda.is_available():
+        raise Exception("Cannot find any GPU devices.")
+
+    if not input_shape:
+        raise ValueError("Missing the input shape for model convert.")
+
+    if not dynamic_batchsize:
```
yiheng-wang-nv, line 670: suggest to add a length check here, need to be 3, otherwise provide an informative error message

monai/bundle/scripts.py
yiheng-wang-nv: Could you also explain more about this arg. For example, I do not know what is `OPT_BATCH`

binliunls: @yiheng-wang-nv , I have updated it, can you read it again and give me some feedback.

yiheng-wang-nv: Hi @binliunls , thanks for the updates. LGTM

yiheng-wang-nv (changes requested):
I left a comment in:
https://github.com/Project-MONAI/model-zoo/pull/325#issuecomment-1477226686
suggest to pend this PR before solving the issue.

yiheng-wang-nv (approved):
The issue is solved. I don't have other comments on this PR, thanks!

monai/networks/utils.py
```diff
@@ -615,6 +626,109 @@ def convert_to_torchscript(
...
+    # Use the dynamic batchsize range to generate compiler input and compile the model
+    if dynamic_batchsize:
+        min_input_shape = scale_batch_size(input_shape, dynamic_batchsize[0])
+        opt_input_shape = scale_batch_size(input_shape, dynamic_batchsize[1])
+        max_input_shape = scale_batch_size(input_shape, dynamic_batchsize[2])
+    else:
+        min_input_shape = opt_input_shape = max_input_shape = input_shape
+
+    with torch.no_grad():
+        torch_tensorrt.set_device(device)
```
wyli, line 702: ci runners report this issue
```
[2023-03-21T09:11:04.418Z] ======================================================================
[2023-03-21T09:11:04.418Z] FAIL: test_visible_devices (tests.test_set_visible_devices.TestVisibleDevices)
[2023-03-21T09:11:04.418Z] ----------------------------------------------------------------------
[2023-03-21T09:11:04.418Z] Traceback (most recent call last):
[2023-03-21T09:11:04.418Z]   File "/home/jenkins/agent/workspace/MONAI-premerge/monai/tests/test_set_visible_devices.py", line 36, in test_visible_devices
[2023-03-21T09:11:04.418Z]     self.assertEqual(num_gpus_before, num_gpus_after)
[2023-03-21T09:11:04.418Z] AssertionError: 0 != 1
[2023-03-21T09:11:04.418Z] 
[2023-03-21T09:11:04.418Z] ----------------------------------------------------------------------
[2023-03-21T09:11:04.418Z] Ran 12329 tests in 1396.392s
[2023-03-21T09:11:04.418Z] 
[2023-03-21T09:11:04.418Z] FAILED (failures=1, skipped=490)
script returned exit code 1
```
I suspect that this set_device seems to change the system device globally and affects the other unit tests

binliunls: Hi @wyli , 
     I think this is a blocker but need @Nic-Ma  to double confirm. If so, how about I just remove this line and add `currently only support the index 0 GPU` to the `device` doc-string?  This change will make the `trt_export` API only work on GPU0.

Thanks,
Bin

wyli: or somehow reset the device in the unit tests

binliunls: Hi @wyli ,
     I haven't found any `reset` like API in Torch_TensorRT. But find the `torch.cuda.device` context manager in pytorch, which works like the `torch_tensorrt.set_device()`.  I can change the `torch_tensorrt.set_device(device)` to `with torch.cuda.device(device=device):` and may need your help to run the ci test again to see if this issue can be fixed.

Thanks,
Bin

wyli: I edited the unit tests and the prints are not consistent in setup and teardown, so I don't think it works...

Screenshot 2023-03-21 at 13 35 33

wyli: the root cause is that when importing torch_tensorrt, it'll change torch.cuda.device_count, ignoring the environment variable.

Nic-Ma: Hi @wyli ,

Do you think we should report this behavior to `torch TensorRT` project and ask to change?

Thanks.

Issue #5991, opened 2023-02-14
review and implement transform functionals in dev


wyli:
**Is your feature request related to a problem? Please describe.**
part of #4855 

this is to track the progress of implementing transform functionals from the prototype defined in #5860, for logically three parts
- [x] minor utility enhancements in
  - [x] monai.data.metatensor (https://github.com/Project-MONAI/MONAI/pull/6017)
  - [x] monai.transforms.lazy.utils (https://github.com/Project-MONAI/MONAI/pull/6017)
  - [x] monai.transforms.transform (https://github.com/Project-MONAI/MONAI/pull/6017)
- [x] crop and pad transforms with unit tests (monai.transforms.croppad)
  - [x] crop array/dict https://github.com/Project-MONAI/MONAI/pull/6068
  - [x] pad array/dict https://github.com/Project-MONAI/MONAI/pull/6031
  - [x] remaining transforms https://github.com/Project-MONAI/MONAI/pull/6083
- [x] spatial transforms with unit tests (monai.transforms.spatial)
  - [x] spatial_resample https://github.com/Project-MONAI/MONAI/pull/6060
  - [x] the other spatial transforms https://github.com/Project-MONAI/MONAI/pull/6080

monai.data.dataset, monai.transforms.compose changes require more discussions and are not within the scope of this feature request.
please feel free to raise issues, and split them into subtasks if needed cc @yiheng-wang-nv @KumoLiu @Nic-Ma

yiheng-wang-nv:
Hi @wyli , @KumoLiu ,
FYI: I think that the modifications of `minor utility enhancements` is not very complicate relatively, therefore I will start on this part first.

yiheng-wang-nv:
Update: @KumoLiu already started working on crop transforms. I'm now starting on spatial transforms.

PR #6031, opened 2023-02-21
add pad transforms with unit tests for lazy resampling

KumoLiu:
Part of #5991.
### Description

This PR adds the pad transforms from https://github.com/Project-MONAI/MONAI/pull/5860 relate to lazy resampling, and it belongs to the second part mentioned in https://github.com/Project-MONAI/MONAI/issues/5991

### Types of changes

### DIFF
--- a/monai/transforms/croppad/array.py
+++ b/monai/transforms/croppad/array.py
@@ -29,9 +29,10 @@ from monai.config.type_definitions import NdarrayOrTensor
 from monai.data.meta_obj import get_track_meta
 from monai.data.meta_tensor import MetaTensor
 from monai.data.utils import get_random_patch, get_valid_patch_size
+from monai.transforms.croppad.functional import pad_func
 from monai.transforms.inverse import InvertibleTransform, TraceableTransform
 from monai.transforms.traits import MultiSampleTrait
-from monai.transforms.transform import Randomizable, Transform
+from monai.transforms.transform import LazyTransform, Randomizable, Transform
 from monai.transforms.utils import (
     compute_divisible_spatial_size,
     convert_pad_mode,
@@ -82,7 +83,7 @@ __all__ = [
 ]
 
 
-class Pad(InvertibleTransform):
+class Pad(InvertibleTransform, LazyTransform):
     """
     Perform padding for a given an amount of padding in each dimension.
 
@@ -142,6 +143,27 @@ class Pad(InvertibleTransform):
         # torch.pad expects `[B, C, H, W, [D]]` shape
         return pad_pt(img.unsqueeze(0), pt_pad_width, mode=mode, **kwargs).squeeze(0)
 
+    @staticmethod
+    def pad_nd(img_t, to_pad_, mode, **kwargs):
+        """pad with torch or numpy function"""
+        if mode in {"linear_ramp", "maximum", "mean", "median", "minimum", "symmetric", "empty"}:
+            return Pad._np_pad(img_t, pad_width=to_pad_, mode=mode, **kwargs)
+        mode = convert_pad_mode(dst=img_t, mode=mode).value
+        try:
+            _pad = (
+                Pad._pt_pad
+                if mode in {"reflect", "replicate"}
+                and img_t.dtype not in {torch.int16, torch.int64, torch.bool, torch.uint8}
+                else Pad._np_pad
+            )
+            return _pad(img_t, pad_width=to_pad_, mode=mode, **kwargs)
+        except (ValueError, TypeError, RuntimeError) as err:
+            if isinstance(err, NotImplementedError) or any(
+                k in str(err) for k in ("supported", "unexpected keyword", "implemented")
+            ):
+                return Pad._np_pad(img_t, pad_width=to_pad_, mode=mode, **kwargs)
+            raise ValueError(f"{img_t.shape} {to_pad_} {mode} {kwargs} {img_t.dtype} {img_t.device}") from err
+
     def __call__(  # type: ignore
         self, img: torch.Tensor, to_pad: list[tuple[int, int]] | None = None, mode: str | None = None, **kwargs
     ) -> torch.Tensor:
@@ -162,52 +184,14 @@ class Pad(InvertibleTransform):
         """
         to_pad_ = self.to_pad if to_pad is None else to_pad
         if to_pad_ is None:
-            to_pad_ = self.compute_pad_width(img.shape[1:])
+            spatial_shape = img.peek_pending_shape() if isinstance(img, MetaTensor) else img.shape[1:]
+            to_pad_ = self.compute_pad_width(spatial_shape)
         mode_ = self.mode if mode is None else mode
         kwargs_ = dict(self.kwargs)
         kwargs_.update(kwargs)
 
         img_t = convert_to_tensor(data=img, track_meta=get_track_meta())
-        _orig_size = img_t.shape[1:]
-
-        # all zeros, skip padding
-        if np.asarray(to_pad_).any():
-            to_pad_ = list(to_pad_)
-            if len(to_pad_) < len(img_t.shape):
-                to_pad_ = list(to_pad_) + [(0, 0)] * (len(img_t.shape) - len(to_pad_))
-            if mode_ in {"linear_ramp", "maximum", "mean", "median", "minimum", "symmetric", "empty"}:
-                out = self._np_pad(img_t, pad_width=to_pad_, mode=mode_, **kwargs_)
-            else:
-                mode_ = convert_pad_mode(dst=img_t, mode=mode_).value
-                try:
-                    _pad = (
-                        self._pt_pad
-                        if mode_ in {"reflect", "replicate"}
-                        and img_t.dtype not in {torch.int16, torch.int64, torch.bool, torch.uint8}
-                        else self._np_pad
-                    )
-                    out = _pad(img_t, pad_width=to_pad_, mode=mode_, **kwargs_)
-                except (ValueError, TypeError, RuntimeError) as err:
-                    if isinstance(err, NotImplementedError) or any(
-                        k in str(err) for k in ("supported", "unexpected keyword", "implemented")
-                    ):
-                        out = self._np_pad(img_t, pad_width=to_pad_, mode=mode_, **kwargs_)
-                    else:
-                        raise ValueError(
-                            f"{img_t.shape} {to_pad_} {mode_} {kwargs_} {img_t.dtype} {img_t.device}"
-                        ) from err
-        else:
-            out = img_t
-        if get_track_meta():
-            self.update_meta(tensor=out, to_pad=to_pad_)  # type: ignore
-            self.push_transform(out, orig_size=_orig_size, extra_info={"padded": to_pad_})
-        return out
-
-    def update_meta(self, tensor: MetaTensor, to_pad: list[tuple[int, int]]):
-        spatial_rank = max(len(tensor.affine) - 1, 1)
-        to_shift = [-s[0] for s in to_pad[1:]]  # skipping the channel pad
-        mat = create_translate(spatial_rank, to_shift)
-        tensor.affine = tensor.affine @ convert_to_dst_type(mat, tensor.affine)[0]
+        return pad_func(img_t, to_pad_, mode_, kwargs_, self.get_transform_info())  # type: ignore
 
     def inverse(self, data: MetaTensor) -> MetaTensor:
         transform = self.pop_transform(data)
--- a/monai/transforms/croppad/array.py
+++ b/monai/transforms/croppad/array.py
@@ -126,7 +126,7 @@ class Pad(InvertibleTransform, LazyTransform):
         raise NotImplementedError(f"subclass {self.__class__.__name__} must implement this method.")
 
     @staticmethod
-    def _np_pad(img: torch.Tensor, pad_width, mode, **kwargs) -> torch.Tensor:
+    def _np_pad(img, pad_width, mode, **kwargs) -> torch.Tensor:
         img_np = img.detach().cpu().numpy() if isinstance(img, torch.Tensor) else img
         mode = convert_pad_mode(dst=img_np, mode=mode).value
         if mode == "constant" and "value" in kwargs:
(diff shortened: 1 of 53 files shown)


monai/transforms/croppad/array.py
```diff
@@ -142,6 +143,27 @@ def _pt_pad(img: torch.Tensor, pad_width, mode, **kwargs) -> torch.Tensor:
         # torch.pad expects `[B, C, H, W, [D]]` shape
         return pad_pt(img.unsqueeze(0), pt_pad_width, mode=mode, **kwargs).squeeze(0)
 
+    @staticmethod
+    def pad_nd(img_t, to_pad_, mode, **kwargs):
```
wyli, line 147: could you please remove the underscore `_`,  `to_pad_` -> `to_pad` and also add some docstring? I can also push some docstring to your PR directly before merging this PR.

KumoLiu: OK, will do that.

tests/padders.py
```diff
@@ -109,3 +112,27 @@ def pad_test_kwargs(self, unchanged_slices, **input_param):
                             inv = padder.inverse(result)
                             assert_allclose(im, inv, type_test=False)
                             self.assertEqual(inv.applied_operations, [])
+
+    def pad_test_pending_ops(self, input_param, input_shape):
```
wyli, line 116: would be great to have a test about lazily combining multiple crop/pad operations chained in different ordering will give the expected consistent outcomes.

monai/transforms/croppad/functional.py
```diff
@@ -0,0 +1,60 @@
...
+        affine=xform,
+        extra_info=extra_info,
+        orig_size=img_size,
+        transform_info=transform_info,
+        lazy_evaluation=transform_info.get(TraceKeys.LAZY_EVALUATION, False),
+    )
+    out = convert_to_tensor(img.as_tensor() if isinstance(img, MetaTensor) else img, track_meta=get_track_meta())
+    if transform_info.get(TraceKeys.LAZY_EVALUATION, False):
+        return out.copy_meta_from(meta_info) if isinstance(out, MetaTensor) else meta_info
+    out = monai.transforms.Pad.pad_nd(out, to_pad_, mode, **kwargs) if do_pad else out
```
Nic-Ma, line 58: I think `functional.py` should be the lower level operations of transforms, but here `pad_func` calls `Pad`'s static method, then `Pad` also calls `pad_func`, I think we should avoid this `A` <-> `B` logic, can we move the `pad_nd` function of `Pad` to here directly?

Thanks.

monai/transforms/spatial/functional.py
```diff
@@ -0,0 +1,131 @@
...
+    meta_info = TraceableTransform.track_transform_meta(
+        img,
+        sp_size=spatial_size,
+        affine=None if affine_unchanged and not lazy_evaluation else xform,
+        extra_info=extra_info,
+        orig_size=original_spatial_shape,
+        transform_info=transform_info,
+        lazy_evaluation=lazy_evaluation,
+    )
+    img = img.as_tensor() if isinstance(img, MetaTensor) else img
```
Nic-Ma, line 98: Could you please help me understand why we convert `MetaTensor` to `Tensor` here?
If must, can we execute it after line 102, then no need to convert it back to `MetaTensor`?

Thanks.

tests/test_spatial_resample.py
```diff
@@ -142,7 +142,7 @@ def test_flips(self, img, device, data_param, expected_output):
                 img = img.to(device)
             out = SpatialResample()(img=img, **data_param)
             assert_allclose(out, expected_output, rtol=1e-2, atol=1e-2)
-            assert_allclose(out.affine, data_param["dst_affine"])
```
Nic-Ma, line 145: May I know why remove the verification for predefined `data_param["dst_affine"]`?

Thanks.

KumoLiu: Hi @Nic-Ma, we don't remove this verification. We added `to_affine_nd` to make the dimension right. If the data is not metatensor, we will add a fake affine.

monai/transforms/croppad/functional.py
KumoLiu: Hi Nic, if like that, we may also need to move `_np_pad` and `_pt_pad` since `pad_nd` also calls these two functions. What do you think? @wyli

monai/transforms/spatial/array.py
```diff
@@ -340,14 +229,6 @@ class ResampleToMatch(SpatialResample):
     """Resample an image to match given metadata. The affine matrix will be aligned,
     and the size of the output image will match."""
 
-    def update_meta(self, img: torch.Tensor, dst_affine=None, img_dst=None):
-        if dst_affine is not None:
-            super().update_meta(img, dst_affine)
```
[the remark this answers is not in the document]
KumoLiu: Hi @Nic-Ma, do you mean save meta information in a dictionary like `meta_dict`? I think the same logic is retained here as in the original.

monai/transforms/croppad/functional.py
wyli: Sure, I don't have a preference here...

KumoLiu: Ok, will move these functions, and it will also avoid the duplicated doc strings.

monai/transforms/spatial/functional.py
wyli: I think line 102 is a shallow copy, so we should drop the current meta first. Skipping this will cause some bugs.

monai/transforms/croppad/functional.py
```diff
@@ -0,0 +1,110 @@
...
+            (PyTorch) {``"constant"``, ``"reflect"``, ``"replicate"``, ``"circular"``}.
+            One of the listed string values or a user supplied function. Defaults to ``"constant"``.
+            See also: https://numpy.org/doc/1.18/reference/generated/numpy.pad.html
+            https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html
+        transform_info: a dictionary with the relevant information pertaining to an applied transform.
+        kwargs: other arguments for the `np.pad` or `torch.pad` function.
+            note that `np.pad` treats channel dimension as the first dimension.
+    """
+
+    def _np_pad(img: torch.Tensor, pad_width: list[tuple[int, int]], mode: str, **kwargs) -> torch.Tensor:
```
wyli, line 48: This makes the functions not reachable by the users, is it really necessary to hide? Ateast I see pad_nd is a good utility. Previously the user can call Pad.pad_nd

KumoLiu: Can we set default value for `transform_info` and `kwargs` like this `, transform_info: dict = {}, kwargs: dict = {}`? Then users could directly call `pad_func`.

wyli: I don't think they are the same thing, the utilities can keep the data types, right?

KumoLiu: Oh, I see.

KumoLiu: Or we could add a flag in `pad_func`, if true, just return `pad_nd`? I can't find another way to avoid `A` <-> `B` logic.

tests/padders.py
KumoLiu: Hi @wyli, since we don't have the compose logic in this PR, do you mean combining multiple operations manually?

tests/test_spatial_resample.py
```diff
@@ -142,7 +142,7 @@ def test_flips(self, img, device, data_param, expected_output):
...
+            assert_allclose(to_affine_nd(len(out.shape) - 1, out.affine), data_param["dst_affine"])
 
     @parameterized.expand(TEST_4_5_D)
     def test_4d_5d(self, new_shape, tile, device, dtype, expected_data):
```
yiheng-wang-nv, line 148: Hi @wyli , here I'm a bit confused with this test case. 
According to: https://github.com/Project-MONAI/MONAI/pull/6031/files#diff-3a1b9fb96f9cde878a33605fcbd39cf1af51a79c285fe662d80a00f0c43e2d5fL205
the transform does not support >3d input images, is it correct?

monai/transforms/croppad/functional.py
wyli: what is A<->B, do you mean `pad_nd` <-> `pad_func`?

tests/padders.py
wyli: Yes

monai/transforms/croppad/functional.py
wyli: We can have a few functions here:


def _np_pad():
  ...

def _pt_pad():
  ...

def pad_nd():
  ...

def pad_func():
  ...

It’s also fine if you want to put these functions elesewhere. But please make pad_nd and pad_func available to the users. pad_nd is a useful low level implementation, pad_func is the primary functional used by the array transform.

tests/padders.py
KumoLiu: Hi @wyli, I have added a test about lazily combining multiple pad operations chained in different ordering in this commit. https://github.com/Project-MONAI/MONAI/pull/6031/commits/81de9e2f1b6aa82ad3ff3cb9754609a36a22c537
But here I have one question, I only test multiply pad operations with the same pad mode. If with different pad modes, which mode should I specify to `apply_transforms`?

monai/transforms/croppad/functional.py
KumoLiu: Have moved `_np_pad`, `_pt_pad`, and `pad_nd` out of `pad_func` as you suggested!

tests/padders.py
wyli: yes, in theory when we lazily evaluate the pending operations we should check the compatibility of the padding modes across the transforms https://github.com/Project-MONAI/MONAI/blob/b9e17e8186cb25c8781e2c82de3b4915081df48b/monai/transforms/lazy/functional.py#L83-L85 the `is_compatible_apply_kwargs` API is defined, for now we haven't implemented any specific rules. In the first version, I don't thinks it's a high-priority feature..

wyli (approved):

monai/transforms/croppad/functional.py
```diff
@@ -0,0 +1,131 @@
...
+    return out
+
+
+def _pt_pad(img: torch.Tensor, pad_width: list[tuple[int, int]], mode: str, **kwargs) -> torch.Tensor:
+    pt_pad_width = [val for sublist in pad_width[1:] for val in sublist[::-1]][::-1]
+    # torch.pad expects `[B, C, H, W, [D]]` shape
+    return pad_pt(img.unsqueeze(0), pt_pad_width, mode=mode, **kwargs).squeeze(0)
+
+
+def pad_nd(img: torch.Tensor, to_pad: list[tuple[int, int]], mode: str, **kwargs):
```
Nic-Ma, line 48: it seems this `pad_nd` function is only used within this file, should we make it as private? `_pad_nd`.

Thanks.

KumoLiu: Hi @Nic-Ma, Wenqi suggested it to be available to the user in this [reply](https://github.com/Project-MONAI/MONAI/pull/6031#discussion_r1115749051).

monai/transforms/croppad/functional.py
```diff
@@ -0,0 +1,131 @@
...
+        return _pad(img, pad_width=to_pad, mode=mode, **kwargs)
+    except (ValueError, TypeError, RuntimeError) as err:
+        if isinstance(err, NotImplementedError) or any(
+            k in str(err) for k in ("supported", "unexpected keyword", "implemented")
+        ):
+            return _np_pad(img, pad_width=to_pad, mode=mode, **kwargs)
+        raise ValueError(f"{img.shape} {to_pad} {mode} {kwargs} {img.dtype} {img.device}") from err
+
+
+def pad_func(img: torch.Tensor, to_pad: list[tuple[int, int]], mode: str, transform_info: dict, kwargs: dict):
```
Nic-Ma, line 84: I think we should use `dict | None = None` for `kwargs` or no typehint, refer to:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/dataset.py#L548

Thanks.

monai/transforms/spatial/array.py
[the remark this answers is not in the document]
Nic-Ma: Hi @KumoLiu ,

I think the original logic will update meta no matter the image is `MetaTensor` or not, but the changed logic of this PR will only update meta when `if isinstance(img, MetaTensor)` as line 280, right?

Thanks.

monai/transforms/spatial/functional.py
Nic-Ma: Hi @KumoLiu ,

Could you please help add 1 line comment here to explain the code logic? Otherwise, I feel it's very confusing why we convert `MetaTensor` to `Tensor`.

Thanks in advance.

monai/transforms/croppad/functional.py
wyli: this `kwargs` doesn't seem to support `None`

Nic-Ma (commented):
Put some minor comments inline, others look good to me.

Thanks.

monai/transforms/spatial/array.py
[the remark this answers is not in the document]
KumoLiu: Hi @Nic-Ma, the original logic also only update affine when data is `MetaTensor` since we didn't have a meta dict before. If data is `non-MetaTensor`, affine information would be lost. Or did I misunderstand it?
https://github.com/Project-MONAI/MONAI/blob/b9e17e8186cb25c8781e2c82de3b4915081df48b/monai/transforms/spatial/array.py#L189-L190

monai/transforms/spatial/functional.py
KumoLiu: Have added comments in this [commit](https://github.com/Project-MONAI/MONAI/pull/6031/commits/4281a597be03066743ac19577baabaf944f86d5c). Thanks!

monai/transforms/croppad/functional.py
KumoLiu: Have removed typehint for `kwargs`, thanks!

monai/transforms/spatial/array.py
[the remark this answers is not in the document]
Nic-Ma: OK, seems we have bug in the dev, if input is not MetaTensor, it still updates meta.
Thanks for the fix!

monai/transforms/croppad/functional.py
Nic-Ma: OK, I see.

Thanks.

Nic-Ma (approved):
Thanks for the quick update.
Looks good to me now.

monai/transforms/croppad/functional.py
```diff
@@ -0,0 +1,131 @@
...
+    return pad_pt(img.unsqueeze(0), pt_pad_width, mode=mode, **kwargs).squeeze(0)
+
+
+def pad_nd(img: torch.Tensor, to_pad: list[tuple[int, int]], mode: str, **kwargs):
+    """
+    PyTorch/Numpy pad ``img`` with integers ``to_pad`` amounts. Depending on the ``mode`` and input dtype,
+    a suitable backend will be used automatically.
+
+    Args:
+        img: data to be transformed, assuming `img` is channel-first and padding doesn't apply to the channel dim.
```
wyli, line 54: the docstring of `pad_nd` is not correct here, the `to_pad` parameter has to include the amounts along the channel dimension as well.

KumoLiu: Thanks for pointing it out, I will update it in the later pr.

KumoLiu: updated docstring in this [commit](https://github.com/Project-MONAI/MONAI/pull/6068/commits/e13de51ed80c0634a2e6895f65d61d92b80deddf).

PR #5960, opened 2023-02-08
5959 uses logging config at module level

wyli:
Signed-off-by: Wenqi Li <wenqil@nvidia.com>

Fixes #5959

### Description
avoid modifying the system logging config

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick --unittests  --disttests`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

PR #6051, opened 2023-02-22
update statshandler logging message

wyli:
### Description

The recommendation in `StatsHandler` of setting the global logging config may not work in some cases (e.g. google colab and https://github.com/Project-MONAI/MONAI/pull/5960) https://docs.python.org/3/library/logging.html#logging.basicConfig

https://github.com/Project-MONAI/MONAI/blob/3eef61e82883e9bc6c08d26954fc86286df7d1b3/monai/handlers/stats_handler.py#L55

in general we shouldn't recommend changing the root logger directly.

This PR changes the message to tune the module level loggers, updated example message will be:
```
.the effective log level of ignite.engine.engine.Engine higher than INFO, StatsHandler may not generate logs,
please use the following code before running the engine to enable it: 
                                                                                
import logging                                                              
logging.getLogger('ignite.engine.engine.Engine').setLevel(logging.INFO) 
```

the updated commands are tested in the colab example:
https://colab.research.google.com/drive/1boqy7ENpKrqaJoxFlbHIBnIODAs1Ih1T

### Types of changes

### DIFF
--- a/monai/handlers/stats_handler.py
+++ b/monai/handlers/stats_handler.py
@@ -18,6 +18,7 @@ from typing import TYPE_CHECKING, Any
 
 import torch
 
+import monai
 from monai.config import IgniteInfo
 from monai.utils import deprecated_arg_default, is_scalar, min_version, optional_import
 
@@ -39,10 +40,11 @@ class StatsHandler:
     It can be used for any Ignite Engine(trainer, validator and evaluator).
     And it can support logging for epoch level and iteration level with pre-defined loggers.
 
-    Note that if `name` arg is None, will leverage `engine.logger` as default logger directly, otherwise,
-    get logger from `logging.getLogger(name)`, we can setup a logger outside first with the same `name`.
-    As the default log level is `WARNING`, it's recommended to call
-    `logging.getLogger(name).setLevel(logging.INFO)` before running this handler to enable the stats logging.
+    Note that if ``name`` is None, this class will leverage `engine.logger` as the logger, otherwise,
+    ``logging.getLogger(name)`` is used. In both cases, it's important to make sure that the logging level is at least
+    ``INFO``. To change the level of logging, please call ``import ignite; ignite.utils.setup_logger(name)``
+    (when ``name`` is not None) or ``engine.logger = ignite.utils.setup_logger(engine.logger.name)``
+    (when ``name`` is None) before running the engine with this handler attached.
 
     Default behaviors:
         - When EPOCH_COMPLETED, logs ``engine.state.metrics`` using ``self.logger``.
@@ -51,12 +53,14 @@ class StatsHandler:
 
     Usage example::
 
-        logging.getLogger("train_stats").setLevel(logging.INFO)
+        import ignite
+        import monai
 
-        trainer = SupervisedTrainer(...)
-        StatsHandler(name="train_stats").attach(trainer)
+        ignite.utils.setup_logger("train_stats")
+        trainer = ignite.engine.Engine(lambda x, y: [0.0])  # an example trainer
+        monai.handlers.StatsHandler(name="train_stats").attach(trainer)
 
-        trainer.run()
+        trainer.run(range(3), max_epochs=4)
 
     More details of example is available in the tutorial:
     https://github.com/Project-MONAI/tutorials/blob/master/modules/engines/unet_training_dict.py.
@@ -129,7 +133,7 @@ class StatsHandler:
         self.state_attributes = state_attributes
         self.tag_name = tag_name
         self.key_var_format = key_var_format
-        self.logger = logging.getLogger(name)  # if `name` is None, will default to `engine.logger` when attached
+        self.logger = logging.getLogger(name) # if `name` is None, will default to `engine.logger` when attached
         self.name = name
 
     def attach(self, engine: Engine) -> None:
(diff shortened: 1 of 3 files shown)


monai/handlers/stats_handler.py
```diff
@@ -41,9 +41,8 @@ class StatsHandler:
 
     Note that if `name` arg is None, will leverage `engine.logger` as default logger directly, otherwise,
     get logger from `logging.getLogger(name)`, we can setup a logger outside first with the same `name`.
-    As the default log level of `RootLogger` is `WARNING`, may need to call
-    `logging.basicConfig(stream=sys.stdout, level=logging.INFO)` before running this handler to enable
-    the stats logging.
+    As the default log level is `WARNING`, it's recommended to call
+    `logging.getLogger(name).setLevel(logging.INFO)` before running this handler to enable the stats logging.
```
Nic-Ma, line 45: Hi @wyli ,

There is one concern here: in most cases, we want to use the `engine.logger` directly, and if want to enable the logging without changing the root level directly, what's the "name" for `getLogger` we should set here?

Thanks.

wyli: In that case the warning message will automatically suggest the logger name using the value of 'engine.logger.name'.

actually I couldn't find any example in pytorch ignite using engine.logger for printing regular stats. name=None doesn't look like a good default value here.

Nic-Ma: Hi @wyli ,

In that case, user will must run the program first, then got the `engine.logger.name` and change the `name` arg to run again? Seems if using `engine.logger.name`, we can't write the correct `name` directly before running?
@vfdev-5 What do you think about our using of `engine.logger` here?

Thanks.

wyli: > In that case, user will must run the program first, then got the `engine.logger.name` and change the `name` arg to run again? Seems if using `engine.logger.name`, we can't write the correct `name` directly before running?
> 
> 
> 
> 

Yes, if the user want to log using the engine logger, they have to specify the name. It's still better than the current Dev branch of run the program first, then changing the global logging level.

I don't think we should by default use engine.logger in this use case, but changing this behaviour may break some existing applications.

wyli: (I update to change the `name=None` default https://github.com/Project-MONAI/MONAI/pull/6051/commits/170a8caa108302403a4c26f83ee6ec515e43862f)

vfdev-5: Here is an example with logger : 
https://github.com/pytorch/ignite/blob/56c31db08bd42ff66424fbc56642326af83621c6/examples/contrib/cifar10/main.py#L28

https://github.com/pytorch/ignite/blob/56c31db08bd42ff66424fbc56642326af83621c6/examples/contrib/cifar10/main.py#L333

wyli: thanks @vfdev-5 that makes sense. @Nic-Ma I think the recommended usage could also be 
```py
ignite.utils.setup_logger("train_stats")
StatsHandler(name="train_stats").attach(trainer)
```
what do you think?

Nic-Ma: Hi @vfdev-5 @wyli ,

Thanks for the discussion and example, so you mean we should not use the logger of `engine`?

wyli: the ignite examples are essentially setting up a logger externally and then set engine.logger=logger

Nic-Ma (approved):
OK, I think the change is acceptable now. But please **verify it with some bundles and workflow tutorials** before merging.

Thanks.

wyli:
> OK, I think the change is acceptable now. But please **verify it with some bundles and workflow tutorials** before merging.
> 
> Thanks.

thanks, I've tested it in a bundle and tutorials, after merging this PR

this global setting can be avoided:
```py
logging.basicConfig(stream=sys.stdout, level=logging.INFO)
```

new usage would be, either specify a logger name when creating `StatsHandler`:
```py
monai.handlers.StatsHandler(name="train_stats"...)
```
or update the engine logger using `ignite.utils.setup_logger`:
```py
trainer = ignite.engine.Engine(lambda x, y: [0.0])
ignite.utils.setup_logger(trainer.logger.name, reset=True)
monai.handlers.StatsHandler().attach(trainer)
trainer.run(range(3), max_epochs=4)
```

Nic-Ma:
> > OK, I think the change is acceptable now. But please **verify it with some bundles and workflow tutorials** before merging.
> > Thanks.
> 
> thanks, I've tested it in a bundle and tutorials, after merging this PR
> 
> this global setting can be avoided:
> 
> ```python
> logging.basicConfig(stream=sys.stdout, level=logging.INFO)
> ```
> 
> new usage would be, either specify a logger name when creating `StatsHandler`:
> 
> ```python
> monai.handlers.StatsHandler(name="train_stats"...)
> ```
> 
> or update the engine logger using `ignite.utils.setup_logger`:
> 
> ```python
> trainer = ignite.engine.Engine(lambda x, y: [0.0])
> ignite.utils.setup_logger(trainer.logger.name, reset=True)
> monai.handlers.StatsHandler().attach(trainer)
> trainer.run(range(3), max_epochs=4)
> ```

Hi @wyli ,

I think your examples are python level, could you please share some usage examples for the bundle config?
Currently, bundles usually leverage this line config to run with correct logging:
https://github.com/Project-MONAI/model-zoo/blob/dev/models/spleen_ct_segmentation/configs/logging.conf#L11
Please show the recommendation change or how to update the existing bundles?
CC @yiheng-wang-nv .

Thanks in advance.

PR #6055, opened 2023-02-23
GridPatch with both count and threshold filtering

### LINKED ISSUE #6049: GridPatch transform support for max num_patches while filtering

GridPatch is very useful to tile on the fly for a fixed number of patches. One usually uses an estimate of the average of number of tiles per slide. For slide with fewer tissue patches than the requested num_patches, we might end up  with some background tiles, which might be desirable in some cases to ensure an even number of tiles. For other cases, where we try to limit any background tiles, we would like to have a way of filtering out background tiles for a fixed number of patches (this might result in fewer tiles than the requested num_patches).

In monai.tranforms.spatial.array.GridPatch, filtering is done as follow in lines 3279-3283
```        
#Filter patches
        if self.num_patches:
            patched_image, locations = self.filter_count(patched_image, locations)
        elif self.threshold:
            patched_image, locations = self.filter_threshold(patched_image, locations)
```
It would be great to add a flag for filtering background when num_patches > 0, maybe filter_bkgd_for_fixed_num_patches set to False by default

```      
#Filter patches
        if self.num_patches:
            patched_image, locations = self.filter_count(patched_image, locations)
        elif self.threshold or self.filter_bkgd_for_fixed_num_patches:
            assert self.threshold, "threshold must be set when filter_bkgd_for_fixed_num_patches is True"
            patched_image, locations = self.filter_threshold(patched_image, locations)
```

One alternative would be to change elif by if, but it might not be desirable for all use cases.

**Additional context**

drbeh:
Fixes #6049 

### Description
This PR add support to `GridPatch` for both filtering by count and threshold. When filtering by threshold, the `num_patches` will be treated as maximum number of patches.

UPDATE: It also removes the deprecated argument of `skimage.measure.regionprops`, 'coordinates', which was causing some test to fail.

### Types of changes


monai/transforms/spatial/array.py
```diff
@@ -3230,22 +3231,22 @@ def __call__(self, array: NdarrayOrTensor):
         patched_image = np.array(patches[0])
         locations = np.array(patches[1])[:, 1:, 0]  # only keep the starting location
 
-        # Filter patches
-        if self.num_patches:
-            patched_image, locations = self.filter_count(patched_image, locations)
```
kenza-bouzid, line 3235: wondering if the order of these operations matters 🤔 (filter_threshold - filter_count) vs (filter_count - filter_threshold)
I have noticed an increase in memory consumption (dataloader worker killed errors) after enabling filter_threshold on top of filter_count. 
I was able to use all available cpu cores previously... but now I have to restrict it to half the cores

drbeh: That's right, the order matters! Logically it makes sense to have threshold filtering first but it might use more memory.

Consider an image with 200 patches, where 100 are background, and a user sets `num_patches=100` and `sort_fn=None`, if we filter by count first and the apply thresholding, we only needs to apply thresholding on maximum of 100 patches but we might end up with 0 patches since all the foregrounds can be filtered by count filtering (random order).

On the other hand, when we apply thresholding first, we need to filter all the 200 patches first, and then apply count filtering (which in this example won't have any effect since there is only 100 left).

drbeh: Having said that, let me see how I can add that flexibility here.

kenza-bouzid: That would be great, thank you!

drbeh: @kenza-bouzid I added a flag `threshold_first` to handle both use cases. Could you please check with `threshold_first=False` and see if you can use all the available cpu cores? Thanks

kenza-bouzid: Oh it actually failed when I applied filter_count first then filter_threshold. 
I tried a workaround yesterday (by only changing elif by if) which forced me to limit the cpu cores, hence my question as you changed the order! Let me try with threshold first as argsort uses quicksort by default with O(n) space complexity

drbeh: I see! I enhanced the memory management by freeing up the memory for the list of extracted patch. This should help you to use more cores. Please try and let me know.

monai/transforms/spatial/array.py
```diff
@@ -3139,14 +3139,18 @@ class GridPatch(Transform, MultiSampleTrait):
...
+            If the requested number of patches is greater than the number of available patches,
+            padding will be applied to provide exactly `num_patches` patches unless `threshold` is set.
+            Defaults to None, which returns all the available patches.
         overlap: the amount of overlap of neighboring patches in each dimension (a value between 0.0 and 1.0).
             If only one float number is given, it will be applied to all dimensions. Defaults to 0.0.
         sort_fn: when `num_patches` is provided, it determines if keep patches with highest values (`"max"`),
             lowest values (`"min"`), or in their default order (`None`). Default to None.
         threshold: a value to keep only the patches whose sum of intensities are less than the threshold.
             Defaults to no filtering.
+        threshold_first: whether to apply threshold filtering before limiting the number of patches to `num_patches`.
```
kenza-bouzid, line 3152: you might want to propagate this new parameter to child transforms that inherit from GridPatch: RandGridPatch, GridPatchd, RandGridPatchd, ...

drbeh: Thanks for catching that! Could you please check your code with the new update and setting/unsetting `threshold_first` to see if it makes any difference for you?
If it doesn't, I would prefer to remove this flag since it makes more sense to always have thresholding first (unless you have any use for it).

kenza-bouzid: Sure, I have submitted some experiments with the changes you made, will keep you updated!

monai/transforms/spatial/array.py
```diff
@@ -3228,29 +3246,36 @@ def __call__(self, array: NdarrayOrTensor):
...
-                    constant_values=self.pad_kwargs.get("constant_values", 0),
-                )
-                locations = np.pad(locations, [[0, padding], [0, 0]], constant_values=0)
+            # Limit number of patches
+            patched_image, locations = self.filter_count(patched_image, locations)
+            # Pad the patch list to have the requested number of patches
+            if self.threshold is None:
+                padding = self.num_patches - len(patched_image)
+                if padding > 0:
+                    patched_image = np.pad(
```
KumoLiu, line 3264: Hi @drbeh, thanks for your update, I could understand why we applied threshold filtering first. But still a little confused here, if we both set `threshold` and `num_patches`,  and after thresholding filtering, maybe we contain 50 patches, and we set `num_patches=100`, then we may need to pad to have 100 patches which may cause these patches lower than the `threshold` we set?

monai/transforms/spatial/array.py
kenza-bouzid: I'm still unable to use more/all cores. 

I have encountered additional issues: 
Here's my machine specs: 8xA100 80GB RAM each, 96 cores, 1900GB CPU RAM
batch size 1, and num_patches=1700  of 224x224 (this is the max I am able to fit in my GPUs so far)
I am using cudnn.benchmark=False, cudnn.determinitic=True to ensure reproducibility

Using your current code with this setting causes: a runtime error 
`RuntimeError: Unable to find a valid cuDNN algorithm to run convolution`
I believe this usually occurs when the VRAM memory limit is hit, so I reduced the num_patches to 1500 
I still can not use all cores -> dataloader worker killed errors, reduced num_workers=6 per gpu
- For threshold_first = False: 
    Training epoch 0 takes10:07
    Training epoch 1 takes 10:01
- For threshold_first = True:
   Training epoch 0 takes 10:50
   Training epoch 1 takes 10:02

I think the difference is unsignificant, so you can remove the flag threshold_first if you prefer so

Now what is a bit puzzling for me is the 
`RuntimeError: Unable to find a valid cuDNN algorithm to run convolution`
I should be able to use the same num_patches 1700 as before.
I have also tried to set cudnn.benchmark=True, cudnn.deterministic=False (also had to reduce the num_workers to 6)
This increased the estimated total epoch time to
- 55:06 with threshold_first=False -> which now fails at step 20/103 with `CUDA out of memory`
- 61:08 with threshold_first=True -> still running so far reached step 80/103 but extremely slow 60min vs 10min (in older runs)

kenza-bouzid: Probably out of scope of this PR, but I had encountered other issues with this transform in the past.

When using num_workers>1 and pin_memory=True, training time increases exponentially over epochs 
image
image
I had profiled/timed all intermediate steps, and found out that it GridPatch was the guilty one 
image
So I timed all intermediate step in the tranform 
 
image
image

 
image

It turned out that the operation that was taking too long was a memory allocation by np.array
https://github.com/Project-MONAI/MONAI/blob/a2ec3752f54bfc3b40e7952234fbeb5452ed63e3/monai/transforms/spatial/array.py#L3279
I eventually fixed it by setting pin_memory=False, 
which I explain by cuda memory allocation being more expensive as it has to be allocated from the pinned memory

Note that I am dealing with particularly large slides ~50kx50k 

Any thoughts on this?

monai/transforms/spatial/array.py
drbeh: Hi @KumoLiu, that's right but these lines for padding (L3262-L3264) are never reached if threshold is set:
```python
if self.threshold is None:
```

monai/transforms/spatial/array.py
drbeh: I see! Now it makes sense. There are alternatives in MONAI that might work better for your use case but I have few questions to clarify on the situation before giving any remedy:

1. Are you loading the entire slide (50kx50k) into memory first (and then apply GridPatch)?
2. Is your loaded slide (before splitting) in numpy array or pytoch Tensor?
4. Do you have overlap between your patches?

kenza-bouzid: 1. Yes, I am using LoadImaged from monai.tradforms with WSIReader, cucim backend
2. in numpy as outputted from LoadImaged 
3. I am applying some geometric augmentations (from monai) + stain normalization 
4. No 

Thanks

kenza-bouzid: Oh my bad 2. is actually a metatensor with data in uint8

monai/transforms/spatial/array.py
KumoLiu: Hi @drbeh, I know that, but there is no hint that you can't set both `threshold` and `num_patches` at the same time, or maybe we need to explain in the docstring that the return value may not be as expected when you set them at the same time.

monai/transforms/spatial/array.py
drbeh: @kenza-bouzid I was not able to find the reason for the behavior that you have reported. It is very strange the time is increasing in the following epochs. Especially for convert to numpy that should take no time because there shouldn't be any copy involved.
Let's separate this issue from the counting/thresholding filtering here, so we can dig deeper and find the root cause.

kenza-bouzid: Sounds great, thanks! 
Btw I have finally been able to use all cpu cores. I restricted my batch metadata to the bare minimum which freed up some space. Thanks for your help!

kenza-bouzid: Oh and np.array does a copy, did you mean np.asarray? https://stackoverflow.com/questions/14415741/what-is-the-difference-between-np-array-and-np-asarray
I have tried changing that line https://github.com/Project-MONAI/MONAI/blob/a2ec3752f54bfc3b40e7952234fbeb5452ed63e3/monai/transforms/spatial/array.py#L3279 
by np.asarray but it didn't fix the increasing time!

drbeh: Right, I know, I meant `np.asarray` which is used in `convert_data_type(array, np.ndarray)`. I thought convert to numpy array in your profiling is referring to this.
https://github.com/Project-MONAI/MONAI/blob/a2ec3752f54bfc3b40e7952234fbeb5452ed63e3/monai/transforms/spatial/array.py#L3268

But you are referring to `np.array` in https://github.com/Project-MONAI/MONAI/blob/a2ec3752f54bfc3b40e7952234fbeb5452ed63e3/monai/transforms/spatial/array.py#L3279

drbeh: Glad that you can use all cpu cores. One additional thing is that if you don't care about having all the patches in the border (which make sense in you use case as the borders in WSI are usually representing backgrounds), you can set `pad_mode=None` which can save some space and time. I will change the default to `None` as it seems to be more useful.

monai/transforms/spatial/array.py
drbeh: I updated the docstrings with mode explanation about it.

drbeh:
@kenza-bouzid @KumoLiu could you please let me know if you have any other comment? Otherwise, we can merge this PR.  @wyli @Nic-Ma

kenza-bouzid (approved):
LGTM!

wyli (approved):
thanks, I'm merging this for now, please follow up if you have any concerns @Nic-Ma @KumoLiu

monai/transforms/spatial/array.py
```diff
@@ -3115,29 +3140,33 @@ def __call__(self, array: NdarrayOrTensor):
         patches = list(zip(*patch_iterator))
         patched_image = np.array(patches[0])
         locations = np.array(patches[1])[:, 1:, 0]  # only keep the starting location
+        del patches  # it will free up some memory if padding is used.
```
Nic-Ma, line 3143: Hi @ericspod , @wyli ,

Do we really need to manually release the var here? I think we usually don't need to care about it unless it accumulates in some for-loop.

Thanks in advance.

drbeh: Hi @Nic-Ma, that's a fair point but here the patches can be very large (but is not being used anymore). On the other hand, when padding patches_image, a similar size image is being created, so we will end up with 4 large arrays (or list of arrays): array, patches, patches_image, and the padded patches_image. Freeing up one may help to use larger input images (although using del does not guarantee free up the memory as it might not be collected by garbage collector at the same time). Do you see any downside of doing that?

wyli: It's also an antipattern detected here https://deepsource.io/gh/Project-MONAI/MONAI/issue/PTC-W0043/occurrences?listindex=0&page=2

Nic-Ma: Hi @drbeh ,

Thanks for your explanation. As the CI tests also raised it, can we remove it for now? If really has some memory issue, let's add it back.
What do you think?

Thanks.

drbeh: Sure, submitted the PR. Thanks

kenza-bouzid:
@drbeh what is the timeline for this feature to be released?

drbeh:
> @drbeh what is the timeline for this feature to be released? 

Hi @kenza-bouzid, it is merged and is the dev branch of MONAI (and the next weekly release) but it will be released in v1.2 which is due in about a month.

kenza-bouzid:
Hi @drbeh I am trying to install a version of MONAI that includes this fix 
https://github.com/microsoft/hi-ml/pull/834
But my Docker image build fails with the following error: 

```
  Running command git clone -q https://github.com/Project-MONAI/MONAI.git /tmp/pip-req-build-injno4yb
  Running command git checkout -q 958aac78ab8130517638a47175571e081ae03aee
  ERROR: Command errored out with exit status 1:
   command: /azureml-envs/azureml_6f4ca4ccd5cfecbee0a7e6231b99b5fc/bin/python /azureml-envs/azureml_6f4ca4ccd5cfecbee0a7e6231b99b5fc/lib/python3.9/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmpq3ubf06k
       cwd: /tmp/pip-req-build-injno4yb
  Complete output (36 lines):
  Please set environment variable `BUILD_MONAI=1` to enable Cpp/CUDA extension build.
  BUILD_MONAI_CPP=False, BUILD_MONAI_CUDA=False, TORCH_VERSION=0.
    File "/tmp/pip-build-env-axzysf86/overlay/lib/python3.9/site-packages/torch/__init__.py", line 172, in _load_global_deps
      ctypes.CDLL(lib_path, mode=ctypes.RTLD_GLOBAL)
    File "/azureml-envs/azureml_6f4ca4ccd5cfecbee0a7e6231b99b5fc/lib/python3.9/ctypes/__init__.py", line 382, in __init__
      self._handle = _dlopen(self._name, mode)
  OSError: /tmp/pip-build-env-axzysf86/overlay/lib/python3.9/site-packages/torch/lib/../../nvidia/cublas/lib/libcublas.so.11: undefined symbol: cublasLtGetStatusString, version libcublasLt.so.11
  ```
  Any idea how to fix this? I am relying on AzureML environment builds so I can't easily set extra environment variables. 
  
  The latest available release is 1.1.0  https://pypi.org/project/monai/#history  you mentioned a weekly release, is it a different dev package?

drbeh:
Hi @kenza-bouzid, 
I am not sure what's the source of the error that you get but maybe a weekly pypi build would be an easier option for the meantime. You can find the monai-weekly here: https://pypi.org/project/monai-weekly/, which is built every Saturday.
The latest version `monai-weekly==1.2.dev2310` should include the changes that you need Please let me know if you encounter any problem.

kenza-bouzid:
@drbeh, the weekly pipy build is working fine. Thank you!

PR #6072, opened 2023-02-27
6069 Add default metadata and logging values for bundle run

### LINKED ISSUE #6069: Add default meta/logging file inputs on bundle run function
This feature request comes from:
https://github.com/Project-MONAI/model-zoo/issues/319

yiheng-wang-nv:
Fixes #6069 .

### Description

This PR adds commonly used default values for `meta_file` and `logging_file` of `monai.bundle.run` function.

### Types of changes

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -609,15 +609,15 @@ def patch_bundle_tracking(parser: ConfigParser, settings: dict) -> None:
 
 def run(
     runner_id: str | Sequence[str] | None = None,
-    meta_file: str | Sequence[str] | None = None,
+    meta_file: str | Sequence[str] | None = "configs/metadata.json",
     config_file: str | Sequence[str] | None = None,
-    logging_file: str | None = None,
+    logging_file: str | None = "configs/logging.conf",
     tracking: str | dict | None = None,
     args_file: str | None = None,
     **override: Any,
 ) -> list:
     """
-    Specify `meta_file` and `config_file` to run monai bundle components and workflows.
+    Specify `config_file` to run monai bundle components and workflows.
 
     Typical usage examples:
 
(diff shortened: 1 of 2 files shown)


monai/bundle/scripts.py
```diff
@@ -718,8 +720,8 @@ def run(
     if "config_file" not in _args:
         warnings.warn("`config_file` not provided for 'monai.bundle run'.")
     _log_input_summary(tag="run", args=_args)
-    config_file_, meta_file_, runner_id_, logging_file_, tracking_ = _pop_args(
-        _args, config_file=None, meta_file=None, runner_id="", logging_file=None, tracking=None
+    meta_file_, logging_file_, config_file_, runner_id_, tracking_ = _pop_args(
+        _args, "meta_file", "logging_file", config_file=None, runner_id="", tracking=None
```
Nic-Ma, line 724: Hi @yiheng-wang-nv ,

I think we have 2 options here:
(1) Put the default values in the function definition directly as your PR, if so, please also put the `runner_id=""` to there.
(2) Put the default values in this `_pop_args` here, leave the function args as `default = None`.

I am OK with either way, what do you think?
BTW, please update the bundle run integration test to cover this change.

Thanks.

yiheng-wang-nv: Thanks Nic. I modified to use method 2 which is the same as `verify_metadata`

monai/bundle/scripts.py
```diff
@@ -717,9 +719,18 @@ def run(
...
+    if logging_file is None:
+        _args["logging_file"] = None
     _log_input_summary(tag="run", args=_args)
     config_file_, meta_file_, runner_id_, logging_file_, tracking_ = _pop_args(
-        _args, config_file=None, meta_file=None, runner_id="", logging_file=None, tracking=None
+        _args,
+        config_file=None,
+        meta_file="configs/metadata.json",
+        runner_id="",
+        logging_file="configs/logging.conf",
```
wyli, line 732: I think line 737 FileNotFoundError is too restrictive, the training can continue without a logging file.

Nic-Ma: Hi @wyli ,

The bundle can run without logging file, I think this line only raise error when a user actually set the logging config file path and the file doesn't exist.

Thanks.

wyli: now the default is `configs/logging.conf`, I also hope it's ok if by default I don't have this file in a folder...how about raising a warning message if it's the default value and the default config doesn't exist?

Nic-Ma: A warning sounds good to me.

Thanks.

yiheng-wang-nv: Thanks @wyli , it makes sense. I just modified according to the comments.

wyli (approved):

PR #6089, opened 2023-03-01
Remove redundant array copy for TiffFileWSIReader

### LINKED ISSUE #5580: WSIReader, cucim , tifffile, performance regressions
with pathology MIL classification tutorial, as an example
https://github.com/Project-MONAI/tutorials/blob/main/pathology/multiple_instance_learning/panda_mil_train_evaluate_pytorch_gpu.py

WSIReader is now imported from  `from monai.data.wsi_reader import WSIReader`. If changing the import to the old way (now deprecated)  `from monai.data.image_reader import WSIReader`, and setting backend=tifffile, then training is ~10% faster.  seems like a performance regression.

Training time on 8gpu, 1 epoch
- from monai.data.wsi_reader import WSIReader, (backend=cucim)   313 seconds
- from monai.data.wsi_reader import WSIReader, (backend=tiffile)   303 seconds
- from monai.data.image_reader import WSIReader, (backend=tiffile)   283 seconds

drbeh:
Fixes #5580 

### Description
This PR removes an "unnecessary" array copy in WSITiffFileReader, which was causing an slow down in loading whole slide images.

### Types of changes

### DIFF
--- a/monai/data/wsi_reader.py
+++ b/monai/data/wsi_reader.py
@@ -825,7 +825,7 @@ class TiffFileWSIReader(BaseWSIReader):
         # Extract patch
         downsampling_ratio = self.get_downsample_ratio(wsi=wsi, level=level)
         location_ = [round(location[i] / downsampling_ratio) for i in range(len(location))]
-        patch = wsi_image[location_[0] : location_[0] + size[0], location_[1] : location_[1] + size[1], :].copy()
+        patch = wsi_image[location_[0] : location_[0] + size[0], location_[1] : location_[1] + size[1], :]
 
         # Make the channel to desired dimensions
         patch = np.moveaxis(patch, -1, self.channel_dim)


drbeh:
```python
from monai.data.wsi_reader import WSIReader
from monai.data.image_reader import WSIReader as WSIReader2

filename = "temp_CMU-1.tiff"
reader1 = WSIReader(backend="TiffFile")
reader2 = WSIReader2(backend="TiffFile")
obj1 = reader1.read(filename)
obj2 = reader2.read(filename)
```
```python
%%timeit
reader1.get_data(obj1, level=1)
```
current performance: 2.32 s ± 97.7
after fixing the issue: 1.45 s ± 12.9 ms

```python
%%timeit
reader2.get_data(obj2, level=1)
```
current performance:  1.97 s ± 50.2 ms
after fixing the issue:. 1.93 s ± 141 ms

wyli:
thanks, is it possible to add a test to avoid regression?

drbeh:
> thanks, is it possible to add a test to avoid regression?

Since it was a performance regression, I am not sure how we should set the criteria for the test as it is hardware dependent, especially when dealing with variability across different run of the same class. Do you have any suggestion? Maybe we should just check for regression whenever we have a major change.

wyli:
Maybe just check theres no array copy, which means changing the patch would also change the input array?

drbeh:
> Maybe just check theres no array copy, which means changing the patch would also change the input array?

Right but in this case there is mo use for the input array since the array input is loaded from disk and is being prepared based on the arguments, so if we copy the original array, it will be discarded anyway.

wyli:
Sure, there was a similar performance bug earlier https://github.com/Project-MONAI/MONAI/issues/3322 hope we can add some tests to avoid reoccurring

wyli (approved):
would be great to have some integration tests

PR #6107, opened 2023-03-06
WSI sliding window splitter

### LINKED ISSUE #5871: Implement a sliding window Splitter for whole slide images

drbeh:
Fixes #5871 

### Description
This PR:
- implements a `WSISlidingWindowSplitter ` a splitter for whole slide imaging,
- refactors `SlidingWindowSplitter` to make it inheritable for wsi version (backward incompatible arguments),
- adds new feature to `AvgMerger` to crop the output (backward compatible), and
- updates `PatchInference` to set cropping shape for new feature in `AvgMerger` (backward compatible)

### Types of changes


drbeh:
@Nic-Ma @wyli do you know why flake8 is failing here?
I cannot reproduce it locally. All checks of `./runtests.sh --build --codeformat` are being passed on my machine.

Nic-Ma:
Hi @drbeh ,

You can use `/black` or `--autofix` command to format the code, refer to:
https://github.com/Project-MONAI/MONAI/blob/dev/runtests.sh#L77

Thanks.

monai/inferers/splitter.py
```diff
@@ -58,101 +59,56 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+    """
+
     def __init__(
         self,
         patch_size: Sequence[int] | int,
         offset: Sequence[int] | int = 0,
         overlap: Sequence[float] | float = 0.0,
         filter_fn: Callable | None = None,
         device: torch.device | str | None = None,
-        pad_mode: str | None = PytorchPadMode.CONSTANT,
```
KumoLiu, line 68: May I know why remove this arg? In this case the user must set the pad mode if the pad is needed, perhaps we can keep this default value here, and keep it consistent with what in the transforms.
https://github.com/Project-MONAI/MONAI/blob/f754928bb220bda15f059779a5c45ca04a05ad5a/monai/transforms/croppad/array.py#L134-L141

drbeh: All the padding information is now captured with `pad_kwargs`, so `pad_mod` can be set as `{"mode": "constant"}`:
https://github.com/Project-MONAI/MONAI/blob/4d51cd118bad233e8255d63f41328037fa7b5268/monai/inferers/splitter.py#L72-L75

monai/inferers/splitter.py
```diff
@@ -90,18 +113,18 @@ class SlidingWindowSplitter(Splitter):
     def __init__(
         self,
         patch_size: Sequence[int] | int,
-        patch_level: int = 0,
```
KumoLiu: May I ask why remove `patch_level` here, does it mean we can only process data in level 0?

monai/inferers/splitter.py
```diff
@@ -56,144 +57,210 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
-        """
         super().__init__(patch_size=patch_size, device=device)
         self.offset = offset
-        if any(ov < 0 or ov >= 1 for ov in ensure_tuple(overlap)):
-            raise ValueError(f"Overlap must be between 0 and 1 but {overlap} is given.")
+        # check if fraction overlaps are within the range of [0, 1)
+        if isinstance(ensure_tuple(overlap)[0], float) and any(ov < 0 or ov >= 1 for ov in ensure_tuple(overlap)):
+            raise ValueError(
+                f"Fraction overlaps must be between 0.0 and 1.0 but {overlap} is given. "
+                "If you wish to use number of pixels as overlap, please provide integer numbers."
```
KumoLiu, line 131: Can we also explain it in doc-string?

monai/inferers/splitter.py
```diff
@@ -56,144 +57,210 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
-        offset = self._get_valid_offset(spatial_shape, spatial_ndim, patch_size)
+        spatial_shape: tuple
+        processed_input: Any
+        if isinstance(inputs, torch.Tensor):
+            processed_input = inputs
+            self.get_patch = self._get_patch_tensor
+            spatial_shape = inputs.shape[self.non_spatial_ndim :]
+        elif isinstance(inputs, (str, os.PathLike)):
+            self.get_patch = self._get_patch_wsi
+            processed_input = self.reader.read(inputs)
```
KumoLiu, line 240: If I understand correctly here, we can only read wsi from a path without any preprocessing(such as `ScaleIntensity`), right?

monai/inferers/splitter.py
```diff
@@ -56,144 +57,210 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+                    f"The provided callable ({filter_fn}) has {n_params} parameters."
+                )
+            elif num_pos_params > 2:
+                raise ValueError(
+                    f"`patch_filter_fn` can have at most two positional parameters (patch, location)."
+                    f"The provided callable ({filter_fn}) has {num_pos_params} positional parameters."
+                )
+        elif filter_fn is not None:
+            raise ValueError(
+                "`patch_filter_fn` should be a callable with two input parameters (patch, location). "
```
Nic-Ma, line 78: `patch_filter_fn` -> `filter_fn`.

Thanks.

monai/inferers/splitter.py
drbeh: Good point, sure!

monai/inferers/splitter.py
```diff
@@ -56,144 +57,210 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+        overlap: the amount of overlap between patches in each dimension [0, 1). Defaults to 0.0.
+        filter_fn: a callable to filter patches. It should accepts exactly two parameters (patch, location), and
+            return True for a patch to keep. Defaults to no filtering.
+        device: the device where the patches are generated. Defaults to the device of inputs.
+        non_spatial_ndim: number of non-spatial dimensions (e.g. batch, color)
+        pad_kwargs: arguments for `torch.nn.functional.pad`.
+            For padding the input images in order to capture the patches crossing the border of the image
+            (either when the offset is negative or the image is non-divisible by the patch_size),
+            "mode" need to be set, e.g., {"mode": "constant"}.
+        reader: the module to be used for loading whole slide imaging. If `reader` is
```
Nic-Ma, line 100: I personally feel this `SlidingWindowSplitter` should be a general purpose class, we should not put WSI specific logic here.
I will vote to develop a subclass as your initial idea.

Thanks.

monai/inferers/splitter.py
drbeh: You are right, the user cannot perform any transformation here, and the reason is WSI is usually not fit into memory so we don't have any transform in MONAI that acts on the entire WSI (if it doesn't fit into memory) and instead if acts upon patches, so this class should only do the splitting and the rest of transformation can be done in the `PatchInferer` where `preprocessing` argument can be any chain of transform on the patch.

monai/inferers/splitter.py
```diff
@@ -56,144 +57,210 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+        slices = (slice(None),) * self.non_spatial_ndim + tuple(
+            slice(loc, loc + ps) for loc, ps in zip(location, patch_size)
+        )
+        patch = inputs[slices]
+        # send the patch to target device
+        patch.to(self.device)
+        return patch
+
+    def _get_patch_wsi(self, inputs: Any, location: tuple[int, int], patch_size: tuple[int, int]) -> torch.Tensor:
+        patch, _ = self.reader.get_data(wsi=inputs, location=location, size=patch_size)
```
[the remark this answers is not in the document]
Nic-Ma: I see you need to pass `inputs`, `location`, `size` args to `WSIReader.get_data()`, but other readers don't have these args: https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/image_reader.py#L126
How do you handle it within SlidingWindowSlitter? Or maybe I misunderstand your idea.

Thanks.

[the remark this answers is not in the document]
drbeh: It is not implement in this PR but we can check if it is not wsi and call as following:
```python
if isinstance(self.reader, ImageReader) and not isinstance(self.reader, BaseWSIReader):
    img_obj = self.reader.read(inputs)
    processed_input, _ = reader.get_data(img_obj)
```
Then it will use `self._get_patch_tensor` instead.
https://github.com/Project-MONAI/MONAI/blob/c032036e16c196cf912827ca44513d24a337d35b/monai/inferers/splitter.py#L209

Please keep in mind that here we have two pathways, one for WSI and one for anythong else. WSI pathway is specific because the image does not fit into memory for the rest of image readers, the image fits into memory so it should go through the tensor path way after it is loaded.

monai/inferers/splitter.py
```diff
@@ -56,144 +57,210 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+    It also allows to offset the starting position and filter the patches.
+
+    Args:
+        patch_size : the size of the patches to be generated.
+        offset: the amount of offset for the patches with respect to the original input.  Defaults to 0.
+        overlap: the amount of overlap between patches in each dimension [0, 1). Defaults to 0.0.
+        filter_fn: a callable to filter patches. It should accepts exactly two parameters (patch, location), and
+            return True for a patch to keep. Defaults to no filtering.
+        device: the device where the patches are generated. Defaults to the device of inputs.
+        non_spatial_ndim: number of non-spatial dimensions (e.g. batch, color)
```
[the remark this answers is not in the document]
drbeh: Right, here we assume that by default the input image size is BCHW[D] and this argument will allow for other cases, e.g. if the input is not batched.

Since we are supporting both 2D and 3D images, I can't think of other ways to dynamically infer the spatial shape of the inputs.

monai/inferers/splitter.py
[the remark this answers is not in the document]
Nic-Ma: Thanks for your explanation.
@ericspod @wyli as WSI and the regular sliding window are 2 quite different pathways, I still feel it's better to extend the WSI feature in a subclass, what do you guys think?

Thanks.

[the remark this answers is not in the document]
Nic-Ma: Hi @drbeh ,

I am open to a discussion, and hope we can get some input from @ericspod @wyli, and others.

Thanks.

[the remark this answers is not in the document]
drbeh: Thanks @Nic-Ma, with the changes that I made, now subclassing would make sense. 
We have several features in the backlog so it's best if we can move fast from here but thanks again for your constrictive comments.

monai/inferers/splitter.py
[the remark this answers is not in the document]
Nic-Ma: I would suggest to fix it to `BCHW[D]`, do you have any concerns?
CC @wyli @ericspod .

Thanks.

monai/data/utils.py
```diff
@@ -207,7 +207,7 @@ def iter_patch_position(
     image_size: Sequence[int],
     patch_size: Sequence[int] | int | np.ndarray,
     start_pos: Sequence[int] = (),
-    overlap: Sequence[float] | float = 0.0,
+    overlap: Sequence[float] | float | Sequence[int] | int = 0.0,
```
KumoLiu, line 210: May also need to update the doc-string.

KumoLiu (approved):

monai/inferers/splitter.py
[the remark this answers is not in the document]
Nic-Ma: Cool!

drbeh:
@Nic-Ma do you have any other comments? thanks

wyli (dismissed):
in this PR of creating a new subclass `WSISlidingWindowSplitter` of `Splitter`, it almost rewrites the current component of `SlidingWindowSplitter` (changed all the code previously merged). I think we should take a pause to review a design and create a prototype for the proposed subclasses before rush into detailed implementations.

wyli:
thanks for the offline discussions @drbeh @Nic-Ma. we agreed that @drbeh will create example usages using this PR and any other proposed subclasses of `Splitter` design. Once we reviewed the working examples, we can continue consider this PR.

drbeh:
Hi @Nic-Ma @wyli @KumoLiu 

I have created a [notebook](https://github.com/Project-MONAI/tutorials/blob/75559a54d3a1f4209c3f67594d0dee2f57079a06/patch_inferer/modular_patch_inferer.ipynb) to demonstrate the use of all components of the new patch inference and the effect of various parameters (https://github.com/Project-MONAI/tutorials/pull/1316). 
It includes both the patch inference for regular images and whole slide images, and covers 
- `PatchInferer` ,
- `AvgMerger`, 
- `SlidingWindowSplitter`, and 
- `WSISlidingWindowSplitter`. 

However, Zarr merger requires more discussion in the working group to achieve the desire behavior, so will be implemented later. For now, WSI patch inference can use `AvgMerger` when memory allows (as demonstrated in the notebook).

Also I added a new feature to `AvgMerger` which is unrelated to this PR but to have everything that I addressed in the notebook in one place, I did not a create separate PR. Hope this is fine.

Please let me know if anything else is required for a better review of this PR. Otherwise, please feel free to proceed with the review. Thanks

monai/inferers/splitter.py
```diff
@@ -58,142 +91,372 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+    """
+
     def __init__(
         self,
         patch_size: Sequence[int] | int,
+        overlap: Sequence[float] | float | Sequence[int] | int = 0.0,
         offset: Sequence[int] | int = 0,
-        overlap: Sequence[float] | float = 0.0,
         filter_fn: Callable | None = None,
+        pad_mode: str | None = None,
```
KumoLiu, line 123: I would suggest to use `PytorchPadMode.CONSTANT` as the default pad_mode, otherwise, the shape of the inferred images may not be the same as the original by default. What do you think?

monai/inferers/splitter.py
```diff
@@ -58,142 +91,372 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+        self.filter_fn = self._validate_filter_fn(filter_fn)
+        # padding
         self.pad_mode = pad_mode
-        self.pad_kwargs = pad_kwargs
+        self.pad_value = pad_value
+        # check a valid padding mode is provided if there is any negative offset.
+        if not self.pad_mode and any(off < 0 for off in ensure_tuple(offset)):
+            raise ValueError(f"Negative `offset`requires a valid padding mode but `mode` is set to {self.pad_mode}.")
+        # batch and color are non-spatial dimensions
+        self.non_spatial_ndim = 2
```
KumoLiu, line 147: Do you mean "batch and channel" here? And seems we have already assume that by default the input image size is BCHW[D], maybe don't need this temporary variable anymore，what do you think?

monai/inferers/splitter.py
```diff
@@ -58,142 +91,372 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+            raise ValueError(f"The input should be a tensor. {type(inputs)} is given.")
+
+        spatial_shape = inputs.shape[self.non_spatial_ndim :]
+        spatial_ndim = len(spatial_shape)
+        patch_size, overlap, offset = self._get_valid_shape_parameters(spatial_shape)
         pad_size, is_start_padded = self._calculate_pad_size(spatial_shape, spatial_ndim, patch_size, offset, overlap)
 
-        if any(pad_size):
+        # Padding
+        if self.pad_mode and any(pad_size):
```
KumoLiu, line 268: Maybe we could move these padding part in to `get_output_shape`, seems there are some duplicated code. What do you think?

KumoLiu (commented):
Thanks for the tutorials help me understand how to use these components. Just leave some comments inline.

wyli:
Sure, I'm busy with the v1.2 release, I'll review this and the notebook after the deadline.

monai/inferers/merger.py
```diff
@@ -36,8 +36,18 @@ class Merger(ABC):
         device: the device where Merger tensors should reside.
     """
 
-    def __init__(self, output_shape: Sequence[int] | None = None, device: torch.device | str | None = None) -> None:
+    def __init__(
+        self,
+        output_shape: Sequence[int] | None = None,
+        crop_shape: Sequence[int] | None = None,
+        crop: bool = False,
```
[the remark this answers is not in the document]
drbeh: Thanks @KumoLiu for your comment. In PatchInferer the crop shape is calculated based on the actual size of the input image by default; however, I will revise it to make it clear for the user how things are working and move the cropping functionality to inferer instead of merger. Will let you know when it's done. Thanks

monai/inferers/splitter.py
drbeh: Sure, I can simplify that too. Thanks

drbeh: Sure, I can change it.

monai/inferers/splitter.py
drbeh: I know but they were required to satisfy the logic for cropping. I will check if I can change it considering the revision in the cropping. thanks

drbeh:
Hi @wyli, it has been more than three weeks since this PR is blocked by your review. I would appreciate it if you can review it in a timely manner or let the others review it. Thanks

CC: @Nic-Ma @daguangxu

monai/inferers/splitter.py
```diff
@@ -340,10 +336,12 @@ def __init__(
...
         self._set_reader(reader, reader_kwargs)
-        if self.pad_mode and self.reader.backend.lower() == "tifffile":
+        if self.reader.backend.lower() == "tifffile":
             raise ValueError(
-                "Padding is not supported with 'TiffFile' reader. "
-                "Please either set the padding to `None` or use 'OpenSlide' or 'cuCIM' reader."
+                "'TiffFile' reader is not appropriate for efficiently loading patches, so it is prohibited. "
+                "Please either use other readers such as 'OpenSlide' and 'cuCIM', or "
+                "first load the whole slide image using `monai.data.WSIReader` with `backend=tifffile` "
+                "(or `monai.data.TiffFileWSIReader`), and then use `SlidingWindowSplitter` instead."
```
KumoLiu, line 344: I see your point here, but there is no reason to disable the TiffFile backend except that it is not efficient, right? Maybe we don't need to disable it here but add more explanation in the doc strings and also add some guide in the tutorial, what do you think? Thanks!

monai/inferers/splitter.py
KumoLiu: Will this part be updated or not?

KumoLiu:
Hi @drbeh, another question here is that will `SlidingWindowInferer` be refactored by using `Splitter` and `Merger` or just remain unchanged or removed since I just noticed that you using `SlidingWindowInferer` as a title in https://app.reviewnb.com/Project-MONAI/tutorials/pull/1316/.
Thanks!

monai/inferers/splitter.py
```diff
@@ -325,9 +321,9 @@ def __init__(
         overlap: Sequence[float] | float | Sequence[int] | int = 0.0,
         offset: Sequence[int] | int = 0,
         filter_fn: Callable | None = None,
-        pad_mode: str | None = None,
+        pad_mode: str | None = "constant",
         device: torch.device | str | None = None,
-        reader: str | BaseWSIReader | type[BaseWSIReader] | None = "cuCIM",
+        reader: str | BaseWSIReader | type[BaseWSIReader] | None = "OpenSlide",
```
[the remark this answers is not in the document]
drbeh: Although cuCIM is more performant than OpenSlide, the supporting file formats are more limited than OpenSlide. We have been working with cucim team to improve the file format supports over past year but unfortunately it is unknown if more file formats will be supported. Thus, having openslide is a safer choice as the default.

drbeh:
> Hi @drbeh, another question here is that will `SlidingWindowInferer` be refactored by using `Splitter` and `Merger` or just remain unchanged or removed since I just noticed that you using `SlidingWindowInferer` as a title in https://app.reviewnb.com/Project-MONAI/tutorials/pull/1316/.
> Thanks!

Hi @KumoLiu, thanks for poitning this out. That is a typo and I'll fix it.

monai/inferers/splitter.py
drbeh: I see your point but the problem is if we allow TiffFile here, which is always a wrong thing to do (without exception as fas as I can think), it means that we are supporting it while we don't. However, I can make it less restrictive and raise a warning. How about that?

monai/inferers/splitter.py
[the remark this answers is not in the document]
KumoLiu: Got it, thanks, then may also need to update the relevant doc string.

monai/inferers/splitter.py
KumoLiu: Sure, that would be better. I also want to know if it will be possible that the other two backends are not applicable and only `TiffFile` can be used so that the user will still require this backend even if the speed is slow. Just not sure about that, it would be great if you could help me confirm it, thanks!

monai/inferers/inferer.py
```diff
@@ -244,11 +264,27 @@ def _aggregate(self, outputs, locations, batch_size, mergers, ratios):
...
+        original_spatial_shape = self.splitter.get_input_shape(inputs)
+        padded_spatial_shape = self.splitter.get_output_shape(inputs)
+
+        # output spatial shapes
+        output_spatial_shape = tuple(round(s * r) for s, r in zip(original_spatial_shape, ratio))
+        padded_output_spatial_shape = tuple(round(s * r) for s, r in zip(padded_spatial_shape, ratio))
+
+        # output shapes
+        crop_shape = out_patch.shape[:2] + output_spatial_shape
+        output_shape = out_patch.shape[:2] + padded_output_spatial_shape
```
Nic-Ma, line 285: I feel confusing about the `crop_shape` and `output_shape` here, could you please help help explain them? Maybe `self.crop=True` should not work with `output_shape` arg together?

Thanks.

monai/inferers/splitter.py
```diff
@@ -58,142 +91,370 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+            the range of [0.0, 1.0) that defines relative overlap to the patch size, or it can be a non-negative int
+            that defines number of pixels for overlap. Defaults to 0.0.
+        filter_fn: a callable to filter patches. It should accepts exactly two parameters (patch, location), and
+            return True for a patch to keep. Defaults to no filtering.
+        pad_mode: define the mode for padding. Either "constant" or None. Default to None.
+            Padding is only supported with "OpenSlide" or "cuCIM" backend, and the filling value is 256.
+        device: the device where the patches are generated. Defaults to the device of inputs.
+        reader: the module to be used for loading whole slide imaging. If `reader` is
+
+            - a string, it defines the backend of `monai.data.WSIReader`. Defaults to cuCIM.
```
Nic-Ma, line 305: Seems the default reader is not `cuCIM` in the code.
Could you please help list all the supported readers in the doc-string?

Thanks.

monai/inferers/splitter.py
```diff
@@ -58,142 +91,370 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+            raise ValueError(
+                f"The underlying wsi readers only support for constant padding. pad_mod={pad_mode} is given."
+            )
+
+        super().__init__(
+            patch_size=patch_size, overlap=overlap, offset=offset, filter_fn=filter_fn, device=device, pad_mode=pad_mode
+        )
+        # Set WSI reader
+        self._set_reader(reader, reader_kwargs)
+        if self.reader.backend.lower() == "tifffile":
```
Nic-Ma, line 339: Maybe no need to handle `tifffile` reader particularly? I feel it's a little bit hacky.
Just describing the supported readers in docstring is enough?

Thanks.

monai/inferers/inferer.py
```diff
@@ -100,6 +100,7 @@ class PatchInferer(Inferer):
         output_keys: if the network output is a dictionary, this defines the keys of
             the output dictionary to be used for merging.
             Defaults to None, where all the keys are used.
+        crop: whether to crop the output to match the input shape. Defaults to True.
```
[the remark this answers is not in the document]
drbeh: Sure, let me think about a better name.

monai/inferers/inferer.py
```diff
@@ -244,11 +264,27 @@ def _aggregate(self, outputs, locations, batch_size, mergers, ratios):
...
     def _get_output_shape(self, inputs, out_patch, ratio):
-        """Define the shape of output merged tensors"""
-        in_spatial_shape = inputs.shape[2:]
-        out_spatial_shape = tuple(round(s * r) for s, r in zip(in_spatial_shape, ratio))
-        output_shape = out_patch.shape[:2] + out_spatial_shape
-        return output_shape
+        """Define the shape of output merged tensors (non-padded and padded)"""
+        if (
+            self.splitter is None
+            or not hasattr(self.splitter, "get_input_shape")
```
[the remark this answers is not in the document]
drbeh: I agree. Will do the change.

monai/inferers/inferer.py
drbeh: Sure, `output_shape` is the shape of the output tensor (might be padded) that needs to be created to store the merged output results and it is always required to properly initialize the merger. However, `crop_shape` is an optional shape to which the output tensor is cropped to have the same shape as the input (non-padded). Thus is if there is no padding involved, both are the same.

monai/inferers/splitter.py
drbeh: The problem is the users most probably does not know about inefficiency of TiffFile when dealing with extracting patches, so I believe at least a warning is essential. What I can do to make it better is to check the reader if the reader is not one of the supported/tested ones (cucim and openslide) and give a warning for the rest.

Nic-Ma: OK, this alternative solution sounds good. Thanks.

monai/inferers/splitter.py
drbeh: default padding is "constant" now.

monai/inferers/splitter.py
drbeh: Basically, use should not use "TiffFile" backend here since it is _unnecessarily_ slow. The reason that I am saying "unnecessarily" is that, the user instead can load the entire WSI into memory first (using TiffFile load it into memory anyways) and use `SlidingWindowSplitter` instead. This simply can give the user a lot faster performance.

monai/inferers/splitter.py
drbeh: Add it in L310.

monai/inferers/splitter.py
drbeh: done!

monai/inferers/inferer.py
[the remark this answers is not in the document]
drbeh: I changed it to "match_spatial_shape". I think now it is clearer. Thanks for the feedback.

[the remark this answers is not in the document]
drbeh: I changed it to "match_spatial_shape". I hope it is clearer now. Thanks for the feedback.

monai/inferers/inferer.py
drbeh: I changed the var names and some changes to the classes to make it less confusing. Now, merger has two shapes: `merging_shape` and `final_shape`.
1. `merging_shape` is the shape of the tensor required to aggregate and merge patches. For instance, it might need to be padded to include the patches crossing the boundaries.
2. `final_shape` is the shape of the final merged output. By default it is the same as `merging_output` but if user provide it, the output tensor will be cropped to have `final_shape` spatial shape.
3. Per previous comment, `crop` is renamed to `match_spatial_shape` and it is a feature of `PatchInferer` and not `Merger`, which tells the `PatchInferer` to find the original input shape of the input data and provide it to the Merger to crop the output to that shape. Thus, I believe it should remain as a main argument of `PatchInferer` and not `meger_kwargs`

Does it make sense to you now? Thanks

monai/inferers/splitter.py
drbeh: I checked that and I think we should keep this duplicity as is. The reason for that is `get_output_shape` (now called `get_padded_shape`) is being called in a different place than `__call__`. In other words, `__call__` is being called by any user/code that uses this Splitter but `get_padded_shape` only needed for specific situations where a user/code wants to know the padded shape of the input, for instance in `PatchInferer` where we are interested to match the output merged shape with the original input shape.

monai/inferers/inferer.py
Nic-Ma: Thanks for the quick update and detailed explanation.
"final_shape" may be a little confusing because it's hard to "guess" which step is "final" here.
And not very sure the `shape` means input or output.
How about change `merging_shape` and `final_shape` to `merge_input_shape` and `merge_output_shape`?
What do you think?

Thanks.

monai/inferers/inferer.py
```diff
@@ -100,34 +100,29 @@ class PatchInferer(Inferer):
...
         preprocessing: Callable | None = None,
         postprocessing: Callable | None = None,
         output_keys: Sequence | None = None,
+        match_spatial_shape: bool = True,
         **merger_kwargs: Any,
     ) -> None:
         Inferer.__init__(self)
-
         # splitter
         if splitter is not None and not isinstance(splitter, Splitter):
```
ericspod, line 122:
```suggestion
        if not isinstance(splitter, (Splitter, type(None))):
```
and remove the enclosed `if not isinstance(splitter, Splitter):` as redundant.

monai/inferers/inferer.py
```diff
@@ -226,14 +224,24 @@ def _initialize_mergers(self, inputs, outputs, patches, batch_size):
...
-            # calculate output_shape only if it is not provided and splitter is not None.
-            if self.splitter is not None and "output_shape" not in self.merger_kwargs:
-                output_shape = self._get_output_shape(inputs, out_patch, ratio)
-                merger = self.merger_cls(output_shape=output_shape, **self.merger_kwargs)
-            else:
-                merger = self.merger_cls(**self.merger_kwargs)
+
+            # calculate merging_shape and final_shape
+            merger_kwargs = self.merger_kwargs.copy()
+            final_shape, merging_shape = self._get_merged_shapes(inputs, out_patch, ratio)
```
ericspod, line 230: if `splitter` is None then these two values will be None as well in which case they have to be provided in `merger_kwargs`. I think this relationship needs to be mentioned in the docstring somewhere, what is says is that `splitter` can be None when assuming inputs are split into patches but it's not clear what the implications of that are. Perhaps this should be checked in the constructor?

monai/inferers/splitter.py
```diff
@@ -58,142 +92,369 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+            return pad_size, False
         # set the starting pad size only if the offset is negative
         pad_size[1::2] = (-min(off, 0) for off in offset)
         # set the ending pad size only if it is not divisible by the patch size
         pad_size[::2] = (
-            0 if ps == 0 else (off - ins + ps) % round(ps * (1.0 - ov))
-            for ins, off, ps, ov in zip(spatial_shape, offset, patch_size, overlap)
+            0 if ps == 0 else (off - sh + ps) % round(ps - (ps * ov if isinstance(ov, float) else ov))
+            for sh, off, ps, ov in zip(spatial_shape, offset, patch_size, overlap)
         )
```
ericspod, line 182: This expression is so complex now it might be better as a for loop with `if-else` statements rather than expressions.

monai/inferers/splitter.py
```diff
@@ -58,142 +92,369 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
-            spatial_shape = inputs.shape[n_non_spatial_dims:]
+            spatial_shape = inputs.shape[2:]
+            # correct the offset with respect to the padded image
+            if is_start_padded:
+                offset = tuple(off + p for off, p in zip(offset, pad_size[1::2]))
+
+        # Splitting
+        for location in iter_patch_position(spatial_shape, patch_size, offset, overlap, False):
+            patch = self._get_patch(inputs, location, patch_size)  # type: ignore
+            patch = ToTensor(device=self.device)(patch)  # type: ignore
```
ericspod, line 277: I'd suggest calling `convert_to_tensor` directly here rather than using a transform. When you do need a transform I'd also recommend not instantiating it just to call but keep an instance around and reuse it to avoid creating objects.

monai/inferers/splitter.py
```diff
@@ -58,142 +92,369 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+            # correct the location with respect to original inputs (remove starting pads)
+            if is_start_padded:
+                location = tuple(loc - p for loc, p in zip(location, pad_size[1::2]))
+            # filter patch and yield
+            if self.filter_fn is None:
+                yield patch, location
+            elif self.filter_fn(patch, location):
+                yield patch, location
```
ericspod, line 285:
```suggestion
            if self.filter_fn is None or self.filter_fn(patch, location):
                yield patch, location
```

monai/inferers/splitter.py
```diff
@@ -58,142 +92,369 @@ def __call__(self, inputs: Any) -> Iterable[tuple[torch.Tensor, Sequence[int]]]:
...
+
+        Returns:
+            spatial_shape
+
+        """
+        wsi = self.reader.read(inputs)
+        level = self.reader_kwargs.get("level", 0)
+        return self.reader.get_size(wsi, level)
+
+    def get_padded_shape(self, inputs: Any) -> tuple:
```
ericspod, line 390: This method appears to be identical to the inherited one from `SlidingWindowSplitter`.

drbeh: Thanks for catching it.

monai/inferers/splitter.py
KumoLiu: Thanks, mark it as resolved now.

ericspod (approved):
Looks good!

monai/inferers/inferer.py
drbeh: @Nic-Ma I agree with you that merging and final are not very clear. On the other hand, `merge_input_shape` is also can be confusing since it is the output shape and the input to the merger is basically patches. Let me rethink it to come up with some alternatives.

drbeh: What about `merged_shape`, which is the shape of the merged tensor, and `crop_to_shape` (or `cropped_shape`), which is the shape that we *optionally* want to crop to?

drbeh: I think in this way, it is clear that we have a "merged" tensor shape and a shape to possibly "crop" the merged shape. How does it feel? Are you good with them?

Nic-Ma (approved):

PR #355, opened 2020-05-06
298-sliding-window

madil90:
Fixes #298  .

### Description
Adds overlap parameter and gaussian weighting support to sliding window algorithm.

- Previous overlap method is removed and replaced with configurable parameter.
- Added parameter to use gaussian weights when blending patches.
- Enhanced tests to include new functionality.

### Status
**Ready**

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or new feature that would cause existing functionality to change)
- [ ] New tests added to cover the changes
- [ ] Docstrings/Documentation updated

PR #6140, opened 2023-03-14
Improve GPU memory efficiency on sliding-window inferer

dongyang0122:
### Description
Improve GPU memory efficiency on sliding-window inferer.

### Types of changes

### DIFF
--- a/monai/inferers/utils.py
+++ b/monai/inferers/utils.py
@@ -142,7 +142,9 @@ def sliding_window_inference(
         diff = max(roi_size[k - 2] - inputs.shape[k], 0)
         half = diff // 2
         pad_size.extend([half, diff - half])
-    inputs = F.pad(inputs, pad=pad_size, mode=look_up_option(padding_mode, PytorchPadMode), value=cval)
+
+    if any(pad_size) > 0:
+        inputs = F.pad(inputs, pad=pad_size, mode=look_up_option(padding_mode, PytorchPadMode), value=cval)
 
     scan_interval = _get_scan_interval(image_size, roi_size, num_spatial_dims, overlap)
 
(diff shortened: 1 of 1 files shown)


monai/inferers/utils.py
```diff
@@ -142,7 +142,9 @@ def sliding_window_inference(
         diff = max(roi_size[k - 2] - inputs.shape[k], 0)
         half = diff // 2
         pad_size.extend([half, diff - half])
-    inputs = F.pad(inputs, pad=pad_size, mode=look_up_option(padding_mode, PytorchPadMode), value=cval)
+
+    if any(pad_size) > 0:
```
wyli, line 146: any returns a boolean value
```suggestion
    if any(pad_size):
```

monai/inferers/utils.py
```diff
@@ -167,8 +169,8 @@ def sliding_window_inference(
     importance_map_ = convert_data_type(importance_map_, torch.Tensor, device, compute_dtype)[0]
 
     # handle non-positive weights
-    min_non_zero = max(importance_map_[importance_map_ != 0].min().item(), 1e-3)
-    importance_map_ = torch.clamp(importance_map_.to(torch.float32), min=min_non_zero).to(compute_dtype)
+    min_non_zero = max(torch.min(importance_map_).item(), 1e-3)
```
wyli, line 172: I can't remember why there is a `importance_map_[importance_map_ != 0]` cc @Nic-Ma 
removing the logic doesn't raise any test error in this PR so I assume it's ok to remove.

Nic-Ma: Hi @wyli ,

Adil added this feature in PR: https://github.com/Project-MONAI/MONAI/pull/355 before.
The comment for this line code is: "importance_map cannot be 0, otherwise we may end up with nans!".

Thanks.

wyli: I think with the current `max` and `clamp`, the map will always be positive. now the difference is to use the smallest positive element or 1e-3 for the zero and negative weights. is there a case that the difference becomes important?

dongyang0122: the `clamping` logic does not change. and `importance_map_` would remain positive after changes. I am not sure why 1e-3 is chosen.

Issue #5312, opened 2022-10-12
Add SomeOf in monai.transforms.Compose 


visionNoob:
**Is your feature request related to a problem? Please describe.**
I think although "SomeOf" is very useful feature for data augmentation and its easy implementation, 
there is no feature for "SomeOf" in monai transforms (Only naive compose and OneOf are implemented) 
https://docs.monai.io/en/stable/_modules/monai/transforms/compose.html#OneOf

**Describe the solution you'd like**
Adding SomeOf class with reference to albumentation's (https://github.com/albumentations-team/albumentations/blob/master/albumentations/core/composition.py#L332)

**Describe alternatives you've considered**
I would like to implement SomeOf feature and contribute to MONAI.

**Additional context**
none

rachmadionl:
hi @Nic-Ma,
I'm new to the whole open-source contributions (never done this before), but I'd like to give it a shot for this issue. May I help? What should I prepare and read first? Thanks in advance.

rachmadionl:
hi @Nic-Ma,
I have read it just now, and I'll try to implement SomeOf. If I understnad it correctly, what I should do first before anything is to create a fork of MONAI repo, right?

Nic-Ma:
@rachmadionl Exactly, thanks!

tuanchien:
Hi all, I wrote up this compositor to use on another project a few months back. Didn't get around to looking at upstreaming it until today and saw this issue. It's not exactly the same as Albumentations. The TLDR of this implementation is it samples without replacement each time using uniform sampling.  You can choose to grab a fixed number of transforms, or variable number (within a given range) each time.

I believe the Albumentations version allows non uniform weighting of transforms, and sampling with replacement as a configurable option. It only allows a fixed number of samples to be drawn each time.

Am open to extending the code if you are wanting to use those missing features from Albumentations.

@rachmadionl I also don't mean to usurp your efforts if you were planning on submitting something. Happy to close my PR and let you submit yours instead if you had something cooked up already.

wyli:
thanks, it's addressed by https://github.com/Project-MONAI/MONAI/pull/6143

PR #6143, opened 2023-03-14
Add SomeOf transform composer

tuanchien:
SomeOf transform composer

### Description

Adds a new transform composer allowing the user to sample "some of" the transforms from a given list of transforms. Allows sampling a fixed number, or variable number of transforms within a configurable range.
Sampling is uniform and without replacement.

The Compose class does almost all of the heavy lifting. This class just replaces the list of transforms each time with a newly sampled list of transforms on each `__call__` invocation. The simple sampling entry point will allow for easy extension later in case there's a need to add support for other probability distributions.

This allows users to compose MONAI transforms in a non sequential way without needing to write their own compositor, or wrapping compositors from frameworks like Albumentations.

Maybe of interest to Issue #5312

### Types of changes


tests/test_some_of.py
```diff
@@ -0,0 +1,99 @@
...
+    def __call__(self, x):
+        return 3 * x
+
+
+class C(Transform):
+    def __call__(self, x):
+        return 5 * x
+
+
+set_determinism(seed=0)
```
wyli, line 37: thanks, please use this approach to avoid changing the global random states: https://github.com/Project-MONAI/MONAI/blob/a8302eca3fc31554d4bd30fae84150766ce13472/tests/test_grid_dataset.py#L32-L37

monai/transforms/compose.py
```diff
@@ -358,3 +359,61 @@ def inverse(self, data):
...
+        transforms: list of callables.
+        map_items: whether to apply transform to each item in the input `data` if `data` is a list or tuple.
+            Defaults to `True`.
+        unpack_items: whether to unpack input `data` with `*` as parameters for the callable function of transform.
+            Defaults to `False`.
+        log_stats: whether to log the detailed information of data and applied transform when error happened,
+            for NumPy array and PyTorch Tensor, log the data shape and value range,
+            for other metadata, log the values directly. Default to `False`.
+        min_samples: minimum number of transforms to sample. Defaults to `0`.
+        max_samples: maximum number of transforms to sample. Defaults to `3`.
```
KumoLiu, line 384: A little confused about these args, maybe `min_num_transforms` or `min_num`. Do you have some suggestions on the name? @wyli @Nic-Ma

tests/test_some_of.py
tuanchien: Good spotting. Made the change.

monai/transforms/compose.py
```diff
@@ -358,3 +359,61 @@ def inverse(self, data):
...
+        self.random = random.Random(get_seed())
+
+    def __call__(self, data):
+        if self.n_transforms != 0:
+            sample_size = self.max_samples if self.fixed else self.random.randint(self.min_samples, self.max_samples)
+            self.active = self.random.sample(np.arange(self.n_transforms).tolist(), sample_size)
+            self.transforms = [self.original_transforms[i] for i in self.active]
+
+        result = super().__call__(data)
+        return result
```
KumoLiu, line 415: Could you please help also add the `inverse` function in this class?

monai/transforms/compose.py
```diff
@@ -358,3 +359,61 @@ def inverse(self, data):
...
+
+    def set_random_state(self, seed: int | None = None, state: np.random.RandomState | None = None) -> Compose:
```
KumoLiu, line 417: May I know why need this function here? Seems we could use the one in the `Compose`.

tests/test_some_of.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class TestSomeOf(unittest.TestCase):
```
KumoLiu: Could you please also help add some combination tests here? Something like:
https://github.com/Project-MONAI/MONAI/blob/a8302eca3fc31554d4bd30fae84150766ce13472/tests/test_random_order.py#L43

KumoLiu (commented):
Thanks for your contribution! Just leave some comments in line.

ericspod:
@atbenmurray Is there anything we need to look at here relating to lazy transforms and/or random state?

monai/transforms/compose.py
```diff
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import random
```
ericspod, line 17: The `SomeOf` class should inherit from `Randomizable` and use our existing methods for doing randomisation, otherwise we're introducing another way of doing random state in addition that found in Numpy and Pytorch. This would make ensuring determinism much harder as well.

monai/transforms/compose.py
```diff
@@ -358,3 +359,61 @@ def inverse(self, data):
...
+    ) -> None:
+        super().__init__(transforms, map_items, unpack_items, log_stats)
+        if transforms is None:
+            transforms = []
+        self.original_transforms = transforms
+        self.n_transforms = len(transforms)
+        self.max_samples = min(self.n_transforms, max_samples)
+        self.fixed = fixed
+        self.min_samples = min_samples
+        self.random = random.Random(get_seed())
```
ericspod, line 406: I would initialise `self.active` here so that it always exists rather than popping in when `__call__` is called for the first time. It's initial value can just be `list(range(len(transforms)))`.

monai/transforms/compose.py
```diff
@@ -358,3 +359,61 @@ def inverse(self, data):
...
+
+    def __call__(self, data):
+        if self.n_transforms != 0:
+            sample_size = self.max_samples if self.fixed else self.random.randint(self.min_samples, self.max_samples)
+            self.active = self.random.sample(np.arange(self.n_transforms).tolist(), sample_size)
```
ericspod, line 411: Using the `R` random state from `Randomizable`:
```suggestion
            all_indices=np.arange(self.n_transforms)
            self.R.shuffle(all_indices)
            self.active = np.sort(all_indices[:sample_size])
```

tuanchien: Thanks for the suggestion. Switched to using self.R from Randomizable.

monai/transforms/compose.py
tuanchien: No longer needed after switching to self.R from Randomizable class.  self.transforms is set in the call to `super().__init__` which has the effect of what you're suggesting.

monai/transforms/compose.py
tuanchien: Thanks. Switched to using self.R instead of a call to the random module.  The `Compose` class already inherits `Randomizable`, and `SomeOf` inherits `Compose`

monai/transforms/compose.py
tuanchien: Thanks for the comment. I've switched the naming to `min_num_transforms` and `max_num_transforms`. Let me know if you prefer a different name. Will explain those variables in the comment below.

monai/transforms/compose.py
```diff
@@ -358,3 +359,61 @@ def inverse(self, data):
...
+        map_items: whether to apply transform to each item in the input `data` if `data` is a list or tuple.
+            Defaults to `True`.
+        unpack_items: whether to unpack input `data` with `*` as parameters for the callable function of transform.
+            Defaults to `False`.
+        log_stats: whether to log the detailed information of data and applied transform when error happened,
+            for NumPy array and PyTorch Tensor, log the data shape and value range,
+            for other metadata, log the values directly. Default to `False`.
+        min_samples: minimum number of transforms to sample. Defaults to `0`.
+        max_samples: maximum number of transforms to sample. Defaults to `3`.
+        fixed: whether to sample exactly `max_samples` transforms, or up to it. Defaults to `False`.
```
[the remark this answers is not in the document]
tuanchien: The purpose of these 3 parameters is to allow you the choice to configure the number of transforms applied with each call.
If `fixed=True`, and `max_samples=3`, then everytime you invoke `__call__`, it will always sample 3 number of transforms to apply. In this case, `min_samples` is ignored.

If `fixed=False`, and `min_samples=0`, `max_samples=3` then every `__call__` will first choose a number $x \in [0,3]$, which is the number of transforms to apply. Then it will sample $x$ transforms to apply.  In this example, all 3 parameters are relevant.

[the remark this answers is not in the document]
KumoLiu: Yes, I know that. 
But if only one arg `n_transforms`, when `n_transforms=3`, it will always sample 3 number of transforms to apply. 
And when `n_transforms=(0, 3)`, will first choose a number from 0-3, which is the number of transforms to apply.
If I misunderstand, just correct me, thanks!

monai/transforms/compose.py
tuanchien: Excuse my ignorance of the code base, but since `SomeOf` inherits `Compose` class, and `Compose` class implements an `inverse` function, what is the benefit here of specifically implementing another `inverse` function inside of `SomeOf`, and how should it differ from the implementation in `Compose`?

monai/transforms/compose.py
tuanchien: This has now been removed :) The code switched to using `self.R` from the `Randomizable` base class, which now makes this no longer necessary. Before, it was necessary because I created a different random state object for this class that `super().set_random_state()` would not change on its own.

tests/test_some_of.py
tuanchien: I've added some combination tests with SomeOf and Compose.

atbenmurray:
I'm taking a look now

tests/test_some_of.py
KumoLiu: Could you please also help add the inverse test? And also for `fixed=False`? Thanks!

monai/transforms/compose.py
atbenmurray: Ok, so I've taken a look at your implementation and how it works. Technically speaking, it looks like you don't need to implement an ```inverse```, but there is a catch. Your implementation modifies the list of transforms destructively on a call by call basis. This is different to how ```OneOf``` and ```RandomOrder``` work, which are the transforms conceptually very close to ```SomeOf```, which leave the transform lists they are initialised with static and instead use metadata to track what transforms were called or what order they were called in. The inverses for these then use the metadata to call the inverses in the right order.

The issue that arises is that any non-trivial use of compose is very likely to cause major issues, from two causes. _Firstly, if you cause ```Compose.__call__``` multiple times before calling inverse, you will and up with the wrong set of transforms when you call the inverse. Secondly, there are other parts of the codebase that interact with the list of transforms in a compose instance, and none of these expect that list to change on a call by call basis._
 
I think it will help to take a look at ```OneOf``` / ```RandomOrder``` and how it does its inverse. ```SomeOf``` and ```OneOf``` are very close conceptually, and I think if you take the same approach that ```OneOf``` does with its metadata, you can implement inverse in the same way that ```OneOf``` does.

```OneOf``` writes the index of the transform that got called into the metadata during ```__call__```. In inverse, it looks at the metadata to figure out which of its transforms was called and then calls inverse on that. Your ```SomeOf``` could write a list of indices for the methods that were called in the forward pass, and then go through that list of indices in reverse. Really the only difference is iterating over a list of indices rather than a single index.

File: monai/transforms/compose.py [hunk 10]
```diff
@@ -358,3 +358,62 @@ def inverse(self, data):
         for o in reversed(applied_order):
             data = apply_transform(self.transforms[o].inverse, data, self.map_items, self.unpack_items, self.log_stats)
         return data
+
+
+class SomeOf(Compose):
```
atbenmurray, line 363: I wrote a longer comment on this in the PR, but I think SomeOf should be implemented similarly to OneOf / RandomOrder. The list of transforms needs to stay static for this to work in the general case and with other parts of monai that assume the transform list doesn't change.

You can instead write the indices of the transforms that were called into the metadata of the tensor on the forward pass, and then iterate over those indices in reverse order during the call to __inverse__, calling the inverse of the transform at the given index.

atbenmurray (changes requested):
In general, implementations cannot dynamically modify transform lists on a call by call basis without breaking in the general case. OneOf and RandomOrder are good examples of how to implement such a function in a way that the transform list remains static.

atbenmurray:
@tuanchien I realise I have dropped a lot of complexity on you here about all this. I am happy to discuss how you might achieve what you want with this PR using a ```OneOf```/```RandomOrder``` style approach.

monai/transforms/compose.py
tuanchien: Thanks for the suggestion, I'll take a look at those examples.

tuanchien:
@atbenmurray thanks for your offer. Will take a stab at it first.

monai/transforms/compose.py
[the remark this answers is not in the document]
tuanchien: Thanks for your comments. Sorry, I should've picked a different example to try explain the goal. It could be that I'm also misunderstanding your main point so feel free to ask further questions.

Consider this setup:
```
s = SomeOf((A(), B(), C(), D(), E()), min_num_transforms=2, max_num_transforms=4, fixed=False)
```

Here, `n_transforms = 5`,  and with each call,  want to sample $x \in \\{2,3,4\\}$ transforms.

So if we call s twice, we could hypothetically have:
```
s(first_input)  # First call.  It randomly chooses x=3, and the transforms [ C(), A(), E() ] to apply
s(second_input)  # Second call. It randomly chooses x=2, and the transforms [ D(), C() ] to apply
```

So rather than just being restricted to always sampling a fixed number of transforms to apply, you can configure it to sample from a range of subsets within a specified size range.  To simplify the maths expressions, let $n$ be `n_transforms`, and $x$ the number of transforms to sample.  Let's ignore permutations for this discussion. The usual implementation of SomeOf (e.g., in Albumentations) takes a fixed $x$ and chooses from ${n \choose x}$ different subsets of transforms to apply.

Let $s$ be `min_num_transforms`, and $t$ be `max_num_transforms`. The goal of these extra parameters is to allow users to increase the sampling space of subsets of transforms to $$\sum_{x=s}^t  {n \choose x}$$

monai/transforms/compose.py
tuanchien: The latest commit is my attempt to refactor it to mimick the RandomOrder design pattern

tuanchien:
From my understanding, the differences between this implementation and the Albumentations version are:
This version:
- Uniformly samples transforms to use.
- Allowed for varying number of sampled transforms.

Albumentations:
- Takes the p's (probabilities) from the transforms, and normalises them to 1, and uses this as the sampling probability distribution.

tuanchien:
I've added a weights with a similar normalisation to `OneOf` to use for sampling. So now it should have feature parity with Albumentations too.

monai/transforms/compose.py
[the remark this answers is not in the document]
KumoLiu: Why can't `n_transforms=(2, 4)` do the things you clarified above? Maybe you misunderstood what I mean, I would suggest `n_transforms` can both accept `int` or `sequence`. The type of `n_transforms` specified by the user will determine whether the num of the transforms will be fixed or random. So it means that when `n_transforms` is `int`, it will do something just like `fixed=True`, and when `n_transforms` is a sequence like `(2, 4)`, it will do something like `fixed=False`.
If I misunderstood, correct me, thanks!

monai/transforms/compose.py
```diff
@@ -358,3 +358,124 @@ def inverse(self, data):
...
+    """
+
+    def __init__(
+        self,
+        transforms: Sequence[Callable] | Callable | None = None,
+        map_items: bool = True,
+        unpack_items: bool = False,
+        log_stats: bool = False,
+        min_num_transforms: int = 0,
+        max_num_transforms: int = 3,
```
atbenmurray, line 396: I'd suggest setting ```max_num_transforms``` to ```None```. When it is ```None```, it can internally set ```max_num_transforms``` to ```len(transforms)```

monai/transforms/compose.py
atbenmurray: This looks really good. It is create to have them all consistent!

tests/test_some_of.py
```diff
@@ -0,0 +1,191 @@
...
+TEST_INVERSES = [
+    (SomeOf((InvC(KEYS), InvD(KEYS))), True, True),
+    (Compose((SomeOf((InvC(KEYS), InvD(KEYS))), SomeOf((InvD(KEYS), InvC(KEYS))))), True, False),
+    (SomeOf((SomeOf((InvC(KEYS), InvD(KEYS))), SomeOf((InvD(KEYS), InvC(KEYS))))), True, False),
+    (SomeOf((Compose((InvC(KEYS), InvD(KEYS))), Compose((InvD(KEYS), InvC(KEYS))))), True, False),
+    (SomeOf((NonInv(KEYS), NonInv(KEYS))), False, False),
+]
+
+
+class TestSomeOf(unittest.TestCase):
```
atbenmurray, line 64: The code coverage checker is flagging a few corner cases still. Once those are resolved, I'll do a proper pass over the tests

atbenmurray (commented):
Looking much better; a few edge cases in testing still to hit, but nearly there

monai/transforms/compose.py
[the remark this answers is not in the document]
tuanchien: I misunderstood your point before, sorry about that.  I have changed the constructor to now accept a `num_transforms` which can be an int, or a tuple specifying the lower/upper bounds.

atbenmurray:
Taking a look now

wyli (approved):

atbenmurray:
@wyli @nic-ma @ericspod Is there an issue here with batches and differing numbers of transforms per-sample?

wyli:
> @wyli @Nic-Ma @ericspod Is there an issue here with batches and differing numbers of transforms per-sample?

good point, I think it should work by looking at the collate function:
https://github.com/Project-MONAI/MONAI/blob/b87375f647b5d4c40e5b70f8dabeb9bd19604719/monai/data/utils.py#L429-L438

@tuanchien have you tried this transform in an end-to-end training example, with monai's dataloader and dataset?

atbenmurray (approved):

tuanchien:
@wyli sorry for the late response.  I have not looked at the collate function.  But for what it's worth, the project I used SomeOf on was end-to-end using MONAI transforms and data loaders and dataset classes.

atbenmurray:
I'd like to offer our belated thanks for your contribution @tuanchien. I am sure people will find this very useful!

PR #6144, opened 2023-03-14
added spacing to surface distances calculations

### LINKED ISSUE #6137: Add `spacing` to surface distances calculations
The current implementation of Hausdorff distance and other surface distance metrics only works for images with isotropic spacing. 

Consider adding a new parameter `spacing` to the metric call in order to pass it as `sampling=spacing` to both `distance_transform_edt` and `distance_transform_cdt` functions inside the `get_surface_distance function` (https://github.com/Project-MONAI/MONAI/blob/a8302eca3fc31554d4bd30fae84150766ce13472/monai/metrics/utils.py#L175). This will allow users to specify the voxel spacing during distance transform computation, which can be useful for handling different image resolutions. 
The same is done in DeepMind's implementation (https://github.com/deepmind/surface-distance/blob/master/surface_distance/metrics.py#L254).

gasperpodobnik:
Fixes #6137 .

### Description

A user can now pass a `spacing` parameter to surface distance metrics (Hausdorff distance, surface distance, surface dice), so that correct results can be obtained for images with non-isotropic spacing (e.g. (0.5, 0.5, 2) mm). 
If `spacing` is a sequence, it must be of length equal to the image dimensions; if a single number, this spacing is used for all axes. If ``None``, spacing of unity is used. Defaults to ``None``so that current behaviour is preserved. 

### Types of changes


wyli:
thank you, could you please add a few test cases for example in this file: https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_surface_distance.p

gasperpodobnik:
Hey, I noticed that the typical usage of the metrics classes is to simply call them (i.e. `metric_class(y_pred, y)`). This means that a `__call__` method that is inherited from the parent `IterationMetric` class is used. However, this method does not support additional arguments, so there is no way to pass the `spacing` argument to the `_compute_tensor` method of the original `metric_class` (e.g. `SurfaceDistanceMetric`). 
I think the best solution would be to add an additional class (e.g. `CumulativeIterationDistanceMetric`) that inherits from `CumulativeIterationMetric` and supports the handling of the `spacing` argument. This could then be passed as a parent class to all user-level metrics classes (i.e. `SurfaceDistanceMetric`, `HausdorffDistanceMetric`, etc.) that require the `spacing` argument for the correct calculation of results. 
Let me know if that's ok or if there is a better way to do it :)

wyli:
thanks @gasperpodobnik, I think we can add `**kwargs` to here
https://github.com/Project-MONAI/MONAI/blob/600621d45c07f9ecc264bd4f13c859c93e435d02/monai/metrics/metric.py#L321
and 
https://github.com/Project-MONAI/MONAI/blob/dev/monai/metrics/metric.py#L52

for additional parameters, could you please help revise if you agree?

gasperpodobnik:
sure :)
Do you maybe know which tool was used to get the reference measurements of Hausdorff distance, surface distance, etc. for testing?

wyli (approved):
many thanks, it looks good to me.

PR #6147, opened 2023-03-14
6136 6146 update the default writer flag

### LINKED ISSUE #6136: Inconsistent writing of Metadata. Loading +  Saving results in different metadata.
**Describe the bug**
Using SaveImage after LoadImage results in image saved with different metadata.

**To Reproduce**
```
    original2 = LoadImage(image_only=True)(original_file)
    SaveImage(output_dir=output_folder + "/monai_modified_meta",
              output_ext = ".nrrd",
              writer= "ITKWriter",
              resample=False,
              separate_folder= False,
              squeeze_end_dims = False,
              output_dtype=itk.SI)(original2)

```
Where `original_file` points to the attached image (a dcm 2D slice):
[CT.X.1.2.276.0.7230010.3.1.4.0.6.1667732449.1809819.zip](https://github.com/Project-MONAI/MONAI/files/10960531/CT.X.1.2.276.0.7230010.3.1.4.0.6.1667732449.1809819.zip) 

**Expected behavior**
The metadata should be the same of the original image after save.

**Screenshots**
The original dcm loaded in Slicer:
image

The result after loading + saving with the above code:
image

**Environment**

```
```
```
MONAI rev id: a2ec3752f54bfc3b40e7952234fbeb5452ed63e3
```

**Additional context**
I have tested using `itk`:
```
original_itk = itk.imread(original_file)
itk.imwrite(original_itk, output_folder + "/monai_modified_meta/original_after_itk_save.nrrd")
```

and the metadata is conserved (afer opening it in Slicer).
There must be some internal Monai logic messing this up. 

The data is a bit odd, but correct, the spacing is negative. ITK corrects this at the moment of reading, flipping the sign in the direction matrix for those indices with negative spacing (and flipping the signs in spacing).

### LINKED ISSUE #6146: SaveImage `set_options` to return `self`
**Describe the bug**
follow up of #6136, `SaveImage.set_options` is not easy to use in a compose transform:
```py
saver = SaveImaged(keys="pred")
saver.set_options(init_kwargs={"affine_lps_to_ras": True})
postprocessing = Compose(
    [
        EnsureTyped(keys=["pred", "seg"]),  # ensure Tensor type after `decollate`
        Activationsd(keys="pred", sigmoid=True),
        ...
        AsDiscreted(keys="pred", threshold=0.5),
        saver,
    ]
)
```

this is a feature request to support:
```py
postprocessing = Compose(
    [
        EnsureTyped(keys=["pred", "seg"]),  # ensure Tensor type after `decollate`
        Activationsd(keys="pred", sigmoid=True),
        ...
        AsDiscreted(keys="pred", threshold=0.5),
        SaveImaged(keys="pred").set_options(init_kwargs={"affine_lps_to_ras": True}),
    ]
)
```

which is easily achievable if `SaveImage.set_options` can return `self`

wyli:
Fixes #6136 
fixes #6146 

### Types of changes


monai/data/image_writer.py
```diff
@@ -373,11 +373,11 @@ class ITKWriter(ImageWriter):
     output_dtype: DtypeLike = None
     channel_dim: int | None
 
-    def __init__(self, output_dtype: DtypeLike = np.float32, affine_lps_to_ras: bool = True, **kwargs):
+    def __init__(self, output_dtype: DtypeLike = np.float32, affine_lps_to_ras: bool | None = None, **kwargs):
```
ericspod, line 376: We should warn that the default value has changed here (with deprecated decorator?).

Nic-Ma: Hi @ericspod ,

I think should use this API: `deprecated_arg_default`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/utils/deprecate_utils.py#L230

Thanks.

monai/data/image_writer.py
```diff
@@ -373,11 +373,11 @@ class ITKWriter(ImageWriter):
...
         """
         Args:
             output_dtype: output data type.
-            affine_lps_to_ras: whether to convert the affine matrix from "LPS" to "RAS". Defaults to ``True``.
+            affine_lps_to_ras: whether to convert the affine matrix from "LPS" to "RAS". Defaults to ``None``.
                 Set to ``True`` to be consistent with ``NibabelWriter``,
```
KumoLiu, line 381: May also update the doc-string here?

monai/data/image_writer.py
```diff
@@ -489,16 +489,21 @@ def create_backend_obj(
             channel_dim: channel dimension of the data array. This is used to create a Vector Image if it is not ``None``.
             affine: affine matrix of the data array. This is used to compute `spacing`, `direction` and `origin`.
             dtype: output data type.
-            affine_lps_to_ras: whether to convert the affine matrix from "LPS" to "RAS". Defaults to ``True``.
+            affine_lps_to_ras: whether to convert the affine matrix from "LPS" to "RAS". Defaults to ``None``.
                 Set to ``True`` to be consistent with ``NibabelWriter``,
```
KumoLiu, line 493: Also here.

monai/data/image_writer.py
KumoLiu, line 376: Not sure why we need this default value change here? But I find another wired thing just using the data provided in this [issue](https://github.com/Project-MONAI/MONAI/issues/6136). The "nrrd" data after `SaveImage` will have affine with shape (5, 5), It looks like it assumes that the data becomes 4-dimensional, without excluding the channel dimension.

```
original_data = LoadImage(image_only=True)(original_file)
SaveImage(output_dir="./monai_modified_meta",
          output_ext = ".nrrd",
          writer= "ITKWriter",
          resample=False,
          separate_folder= False,
          squeeze_end_dims = False,
          output_dtype=itk.SI)(original_data)
print(original_data.meta["spatial_shape"]) # --> ([512 512   1])

save_data = LoadImage(image_only=True)(save_data_path)
print(save_data.meta["spatial_shape"]) # --> [512 512   1   1]
```

wyli: sure, I've changed to the previous default

wyli: thanks, that's a bug and this https://github.com/Project-MONAI/MONAI/commit/91d974f2e1654c391781b4204ee87635659100b3 fixes it. 

(`SaveImage` has some logic to handle the test case even if it's not channel first: https://github.com/Project-MONAI/MONAI/blob/778d0aa558b69a600e4db74e7d59490b7ca9370f/monai/transforms/io/array.py#L453-L454)

KumoLiu (approved):
Thanks for fixing, LGTM.

monai/data/image_writer.py
```diff
@@ -414,9 +415,12 @@ def set_data_array(
             spatial_ndim=kwargs.pop("spatial_ndim", 3),
             contiguous=kwargs.pop("contiguous", True),
         )
-        self.channel_dim = (
-            channel_dim if self.data_obj is not None and len(self.data_obj.shape) >= _r else None
-        )  # channel dim is at the end
+        if channel_dim is None:
+            self.channel_dim = -1  # after convert_to_channel_last the last dim is the channel
```
KumoLiu, line 419: I think there has a bug. Since the default value of `squeeze_end_dims` is `True` in `SaveImage`, the `data_obj` after `convert_to_channel_last` will always don't have a channel. 
https://github.com/Project-MONAI/MONAI/blob/678b5124a74f32859d692d142728ca1f7fa98a25/monai/transforms/io/array.py#L382

wyli: hmm with auto squeeze_end_dims, we'll lose track of the channel dim

Issue #6149, opened 2023-03-15
integration errors


wyli:
**Describe the bug**
some version dependent integration errors
```
[2023-03-15T05:40:29.871Z] ======================================================================
[2023-03-15T05:40:29.871Z] FAIL: test_bspline (tests.test_global_mutual_information_loss.TestGlobalMutualInformationLoss)
[2023-03-15T05:40:29.871Z] ----------------------------------------------------------------------
[2023-03-15T05:40:29.871Z] Traceback (most recent call last):
[2023-03-15T05:40:29.871Z]   File "/home/jenkins/agent/workspace/MONAI-postmerge/monai_github/tests/test_global_mutual_information_loss.py", line 109, in test_bspline
[2023-03-15T05:40:29.871Z]     np.testing.assert_allclose(result, expected_value, rtol=1e-3, atol=5e-3)
[2023-03-15T05:40:29.871Z]   File "/opt/conda/lib/python3.8/site-packages/numpy/testing/_private/utils.py", line 1530, in assert_allclose
[2023-03-15T05:40:29.871Z]     assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
[2023-03-15T05:40:29.871Z]   File "/opt/conda/lib/python3.8/site-packages/numpy/testing/_private/utils.py", line 844, in assert_array_compare
[2023-03-15T05:40:29.871Z]     raise AssertionError(msg)
[2023-03-15T05:40:29.871Z] AssertionError: 
[2023-03-15T05:40:29.871Z] Not equal to tolerance rtol=0.001, atol=0.005
[2023-03-15T05:40:29.871Z] 
[2023-03-15T05:40:29.871Z] Mismatched elements: 1 / 1 (100%)
[2023-03-15T05:40:29.871Z] Max absolute difference: 0.028
[2023-03-15T05:40:29.871Z] Max relative difference: 0.045
[2023-03-15T05:40:29.871Z]  x: array(-0.596109, dtype=float32)
[2023-03-15T05:40:29.871Z]  y: array(-0.624335)
[2023-03-15T05:40:29.871Z] 
[2023-03-15T05:40:29.871Z] ----------------------------------------------------------------------
[2023-03-15T05:40:29.871Z] Ran 12322 tests in 2105.197s
[2023-03-15T05:40:29.871Z] 
[2023-03-15T05:40:29.871Z] FAILED (failures=1, skipped=116)
```

```
[...]

PR #6150, opened 2023-03-15
Fix "fast_training_tutorial.ipynb"

KumoLiu:
Fixes https://github.com/Project-MONAI/tutorials/issues/1250.
Fixes https://github.com/Project-MONAI/MONAI/issues/6149

### Description
Now when `Resize` accept the same spatial_size with the original shape and `set_track_meta(False)`, it will only return meta information.
https://github.com/Project-MONAI/MONAI/blob/9fd6d4ced3efb344cf3b24c51d899d8fafc882d5/monai/transforms/spatial/functional.py#L297-L301

In fast_training_tutorial.ipynb, when in "fast" mode, it will `set_track_meta(False)` which causes error when doing `Resize`
This pr change `sliding_window_inference` to only `Resize` when the shape is different.
https://github.com/Project-MONAI/MONAI/blob/a2ec3752f54bfc3b40e7952234fbeb5452ed63e3/monai/inferers/utils.py#L253

### Types of changes


monai/inferers/utils.py
```diff
@@ -255,7 +255,11 @@ def sliding_window_inference(
                             "Tips: if overlap*roi_size*zoom_scale is an integer, it usually works."
                         )
                     original_idx_zoom[axis] = slice(int(zoomed_start), int(zoomed_end), None)
-                importance_map_zoom = resizer(importance_map.unsqueeze(0))[0].to(compute_dtype)
+                importance_map_zoom = (
+                    resizer(importance_map.unsqueeze(0))[0].to(compute_dtype)
```
wyli, line 259: thanks for looking into this issue, for the resize functional, I think we should also fix the logic when orig_size==out_size? (this is untested):
```py
    if transform_info.get(TraceKeys.LAZY_EVALUATION, False):
        if anti_aliasing and transform_info.get(TraceKeys.LAZY_EVALUATION, False):
            warnings.warn("anti-aliasing is not compatible with lazy evaluation.")
        return out.copy_meta_from(meta_info) if isinstance(out, MetaTensor) else meta_info
    if tuple(convert_to_numpy(orig_size)) == out_size:
        return out.copy_meta_from(meta_info) if isinstance(out, MetaTensor) else out
```

KumoLiu: Have changed the logic in `Resize` and added a unit test for `set_track_meta(False)`.

wyli (approved):
Thanks, it looks good to me. I'll include minor integration tests https://github.com/Project-MONAI/MONAI/compare/dev...6149-integraion

monai/transforms/spatial/functional.py
```diff
@@ -295,10 +295,12 @@ def resize(img, out_size, mode, align_corners, dtype, input_ndim, anti_aliasing,
         lazy_evaluation=transform_info.get(TraceKeys.LAZY_EVALUATION, False),
     )
     out = convert_to_tensor(img.as_tensor() if isinstance(img, MetaTensor) else img, track_meta=get_track_meta())
-    if transform_info.get(TraceKeys.LAZY_EVALUATION, False) or tuple(convert_to_numpy(orig_size)) == out_size:
+    if transform_info.get(TraceKeys.LAZY_EVALUATION, False):
         if anti_aliasing and transform_info.get(TraceKeys.LAZY_EVALUATION, False):
             warnings.warn("anti-aliasing is not compatible with lazy evaluation.")
         return out.copy_meta_from(meta_info) if isinstance(out, MetaTensor) else meta_info
+    if tuple(convert_to_numpy(orig_size)) == out_size:
+        return out.copy_meta_from(meta_info) if isinstance(out, MetaTensor) else out
```
wyli, line 303: to be consistent with line 321, this should always return float32...

also there's the same issue here https://github.com/Project-MONAI/MONAI/blob/9fd6d4ced3efb344cf3b24c51d899d8fafc882d5/monai/transforms/spatial/functional.py#L135-L138

I'll update

PR #6161, opened 2023-03-16
Add supports for AMD GPU

yaomingamd:
Fixes # .

### Description

Add supports for AMD GPU: 
~1. add docker file  "Dockerfile.amd " for creating docker image to  support AMD GPU. The file is based on "Dockerfile" for NVIDIA GPU~
2. In monai/_extensions/gmm/gmm_cuda.cu ,   replaces __shfl_down_sync and __shfl_xor_sync with __shfl_down and __shfl_xor when build for AMD GPU.

### Types of changes


wyli:
thanks for the PR, we don't currently have the ci/cd resources to verify/test the changes, is there any ci runners we can use to test and for future maintainence?

yaomingamd:
Thanks for the quick response. Currently we are considering  different options to run the ci/cd on AMD hardware.  We would like to discuss with you  at your convenience, so that we can setup ci to fit current ci pipeline of MONAI project better.

wyli:
> HI @wyli Would it be possible for us to setup a call to discuss how we can integrate this PR ? We would like to understand how we can help in setting up the CI/CD resources for validating the changes. Thank you.

sure @yaomingamd @japarada, let's have a call, please send your contact details to wenqil@nvidia.com I'll make a meeting time poll and invite colleagues.

ericspod:
As we discussed let's remove the Dockerfile from this PR for now since we'll refactor the existing file to allow for selection of image by variable which will let us choose AMD images. The changes otherwise are fine since they pass our tests and we'll have to assume for now our notional AMD tests will pass also. One thing that does have to be done is to sign your commits to get DCO to pass, clicking on details will explain how to do a remedial commit for that. Thanks!

monai/_extensions/gmm/gmm_cuda.cu
```diff
@@ -82,13 +82,19 @@ __global__ void CovarianceReductionKernel(
 
   for (int i = 0; i < MATRIX_COMPONENT_COUNT; i++) {
     float matrix_component = matrix[i];
-
+#ifdef __HIP_PLATFORM_AMD__
+    matrix_component += __shfl_down(matrix_component, 16);
```
ericspod, line 86: To reduce the amount of code, and remove near-duplication, perhaps defining macros for unsupported functions would make more sense, eg.

```cpp
#ifdef __HIP_PLATFORM_AMD__
  #define __SHFL_DOWN __shfl_down
  #define __SHFL_XOR __shfl_xor
#else
  #define __SHFL_DOWN __shfl_down_sync
  #define __SHFL_XOR __shfl_xor_sync
#endif
```
Wherever `__shfl_down_sync` or `__shfl_xor_sync` appears replace with this macro, the same can be done for the other unsupported functions. If support for these comes about in the future we can remove the macro.

yaomingamd: Thanks Eric for the suggestion! I will update and commit soon.

wyli (approved):
Thanks, it looks good to me, merging for more tests..

PR #6163, opened 2023-03-16
more efficient Dice metrics for large num_class

wyli:
- mostly from @myron's implementation
- remove `is_binary_tensor` checks, they are too expensive.
- remove the deprecated `compute_meandice` (use `compute_dice` instead)

### Types of changes

### DIFF
--- a/monai/metrics/meandice.py
+++ b/monai/metrics/meandice.py
@@ -73,7 +73,6 @@ class DiceMetric(CumulativeIterationMetric):
                 The values should be binarized.
 
         Raises:
-            ValueError: when `y` is not a binarized tensor.
             ValueError: when `y_pred` has less than three dimensions.
         """
         dims = y_pred.ndimension()
--- a/monai/metrics/meandice.py
+++ b/monai/metrics/meandice.py
@@ -28,13 +28,11 @@ class DiceMetric(CumulativeIterationMetric):
     It supports both multi-classes and multi-labels tasks.
     Input `y_pred` is compared with ground truth `y`.
     `y_pred` is expected to have binarized predictions and `y` can be single-channel class indices or in the
-    one-hot format.
-    The `include_background` parameter can be set to ``False`` to exclude
+    one-hot format. The `include_background` parameter can be set to ``False`` to exclude
     the first category (channel index 0) which is by convention assumed to be background. If the non-background
     segmentations are small compared to the total image size they can get overwhelmed by the signal from the
-    background.
-    `y_preds` and `y` can be a list of channel-first Tensor (CHW[D]) or a batch-first Tensor (BCHW[D]), `y` can
-    also be in the format of `B1HW[D]`.
+    background. `y_preds` and `y` can be a list of channel-first Tensor (CHW[D]) or a batch-first Tensor (BCHW[D]),
+    `y` can also be in the format of `B1HW[D]`.
 
     Example of the typical execution steps of this metric class follows :py:class:`monai.metrics.metric.Cumulative`.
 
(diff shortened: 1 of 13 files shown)


monai/metrics/meandice.py
```diff
@@ -114,45 +119,116 @@ def compute_dice(
...
+        reduction=MetricReduction.NONE,
+        get_not_nans=False,
+        softmax=False,
+        ignore_empty=ignore_empty,
+    )(y_pred=y_pred, y=y)
 
-    Raises:
-        ValueError: when `y_pred` and `y` have different shapes.
 
+class DiceHelper:
```
KumoLiu, line 142: I find the name "DiceHelper" a little wired. Why can't we just call it "ComputeDice"?
Feel free to ignore this comment. :)

monai/metrics/meandice.py
```diff
@@ -114,45 +119,116 @@ def compute_dice(
...
+        if y_o > 0:
+            return (2.0 * torch.sum(torch.masked_select(y, y_pred))) / (y_o + torch.sum(y_pred))
+        if self.ignore_empty:
+            return torch.tensor(float("nan"), device=y_o.device)
+        denorm = y_o + torch.sum(y_pred)
+        if denorm <= 0:
+            return torch.tensor(1.0, device=y_o.device)
+        return (2.0 * torch.sum(torch.masked_select(y, y_pred))) / denorm
+
+    def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
```
KumoLiu, line 210: Missing doc-string here.

monai/metrics/meandice.py
wyli: this is from the original implementation, and when we chat about this implementation, I find "DiceHelper" makes it easy to understand what we are referring to

monai/metrics/meandice.py
```diff
@@ -66,29 +69,29 @@ def _compute_tensor(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor
...
-        is_binary_tensor(y, "y")
-
         dims = y_pred.ndimension()
         if dims < 3:
             raise ValueError(f"y_pred should have at least 3 dimensions (batch, channel, spatial), got {dims}.")
         # compute dice (BxC) for each channel for each batch
-        return compute_dice(
-            y_pred=y_pred, y=y, include_background=self.include_background, ignore_empty=self.ignore_empty
-        )
+        return DiceHelper(  # type: ignore
```
ericspod, line 82: I would always advise against instantiating a class just to call a method like this, it's like a poltergeist. This instance of `DiceHelper` should be a member value which is used here instead.

monai/metrics/meandice.py
```diff
@@ -114,45 +117,124 @@ def compute_dice(
...
+        reduction=MetricReduction.NONE,
+        get_not_nans=False,
+        softmax=False,
+        ignore_empty=ignore_empty,
+    )(y_pred=y_pred, y=y)
 
-    Raises:
-        ValueError: when `y_pred` and `y` have different shapes.
 
+class DiceHelper:
```
ericspod, line 140: What is the rationale behind putting the dice code into a class? This class doesn't really have much state, whatever is passed to its constructor could be arguments for the function like in the original version. The code using this class creates a new instance each time as well which doesn't let us reuse the object's state in future invocations of `__call__`.

wyli: useful for configuration based workflows to prepare instances cc @myron

myron:
Thank you for adding it, looks very good. 
One small comment here
https://github.com/Project-MONAI/MONAI/blob/dev/monai/metrics/meandice.py#L206-L208

if we reach the last "return" statement, it seems to be guaranteed to be 0, so we don't need to calculate it

PR #6170, opened 2023-03-18
Add transform to handle empty box as training data

### LINKED ISSUE #5990: ConvertBoxToStandardModed cannot handle empty boxes
**Describe the bug**
When input boxes are empty [], it got error since the boxes are supposed to have shape [N,4] or [N,6].

ValueError: Currently we support only boxes with shape [N,4] or [N,6], got boxes with shape torch.Size([1, 0])

**To Reproduce**
2. Install '....'
3. Run commands '....'

**Expected behavior**

**Screenshots**

**Environment**

```
```

**Additional context**

Can-Zhao:
Fixes #5990 .

### Description

Add transforms to convert empty box with shape (0,M) or (0,) into (0,4) or (0,6).
Provide format checking inside detector so users can know how to solve the format issue with empty box input.

### Types of changes


monai/apps/detection/transforms/dictionary.py
```diff
@@ -95,6 +99,50 @@
 DEFAULT_POST_FIX = PostFix.meta()
 
 
+class StandardizeEmptyBoxd(MapTransform, InvertibleTransform):
```
wyli, line 102: I'm not familiar with the use cases, is there a case that the user don't want to use this transform... maybe this transform should be enabled by default within some existing transforms, for example in  `MaskToBoxD`?

Can-Zhao: Good point. How about running it by default inside the detector?

wyli: sure, perhaps call the util in the detector https://github.com/Can-Zhao/MONAI/blob/c13895d021b0d5aa4caccc85f742a7dac4b1cc89/monai/data/box_utils.py#L519

Can-Zhao: I let the detector automatically correct the empty box shape and label data type, while print a warning message to let the users know what has been done. Shall we remove the array and dict transforms?

wyli (approved):

monai/apps/detection/transforms/dictionary.py
wyli: looks good to me!

Issue #5821, opened 2023-01-06
Define python level abstraction for a Bundle


Nic-Ma:
**Is your feature request related to a problem? Please describe.**
Currently, we developed many bundles in the `model-zoo` repo and developed the `ConfigParser` to load configs.
But actually we don't have a clear definition about what a Bundle is in the python level.
It would be nice to define the **Bundle class** to represent a bundle and enable users to follow & define their customized bundle structure, and provide the APIs for other applications to easily load and handle a bundle, like `MONAI Label`, `MONAI FL`, etc.
Several draft steps to do:

- [x] Develop `BundleWorkflow` interface and implement config-based subclass, so bundle can be implemented with JSON/YAML config or python code, just need to follow the interface.
- [ ] How to expose the CLI entry and args for customized `BYOCBundleWorkflow` in `bundle/scripts/`.
- [ ] Update FL monai algo to use new workflow.
- [ ] Update MONAI Label to use new workflow.
- [ ] Develop `Bundle` interface for the whole bundle, mainly about folder structure, files, download, etc.

Nic-Ma:
CC @wyli @SachidanandAlle for more discussion.

Thanks.

vikashg:
Hi @Nic-Ma 
Thanks for starting this post. As I mentioned that we need to think of usability of bundles from 2 different angles
1. Ease of use from an end-user perspective
2. Ease of building a MONAI Bundle

From an end-user perspective, if the monai model weights should be available from the python script (as they are for pytorch and keras) without explicitly downloading the model. For example
```
import torchvision
from torchvision.models.detection.faster_rcnn import FastRCNNPredictor
from torchvision.models.detection.mask_rcnn import MaskRCNNPredictor
```
or for keras
```
base_model = keras.applications.Xception(
    weights="imagenet",  # Load weights pre-trained on ImageNet.
    input_shape=(150, 150, 3),
    include_top=False,
)  # Do not include the ImageNet classifier at the top.

# Freeze the base_model
base_model.trainable = False
```

It will be nice to have similar interface for MONAI models. As far as I understand and used, we can create a model architecture in the similar way in MONAI, but they are not populated with model weights. We looked at loading torchvision models and fine-tuning their model weights in [this pull request](https://github.com/Project-MONAI/MONAI/pull/4873) last year with @wyli. 

So maybe if we can reach a place where we can load the model weights as follows

```
from monai.model_zoo import brats_tumor_segmentation_model
[...]

PR #6181, opened 2023-03-20
5821 Enhance bundle CLI entry for different bundle workflows

Nic-Ma:
part of #5821 .

### Description

This PR enhanced the bundle CLI entry to support different customized bundle workflows.

### Types of changes


monai/bundle/scripts.py
```diff
@@ -580,102 +581,59 @@ def get_bundle_info(
...
-        init_id=init_id_,
-        run_id=run_id_,
-        final_id=final_id_,
-        tracking=tracking_,
-        **_args,
-    )
-    workflow.initialize()
-    workflow.run()
-    workflow.finalize()
+    (workflow_name,) = _pop_args(_args, workflow=ConfigWorkflow)
```
ericspod, line 621: It's really key for what's going here that the default value for `workflow_name` will be `ConfigWorkflow` is `None` is passed as the argument, it's worth commenting here that that's the case.

monai/bundle/scripts.py
```diff
@@ -580,102 +581,59 @@ def get_bundle_info(
...
+    if isinstance(workflow_name, str):
+        workflow_class, has_built_in = optional_import("monai.bundle", name=f"{workflow_name}")  # search built-in
+        if not has_built_in:
+            workflow_class = locate(f"{workflow_name}")  # search dotted path
+        if workflow_class is None:
+            raise ValueError(f"can not locate specified workflow class: {workflow_name}.")
```
ericspod, line 627:
```suggestion
            raise ValueError(f"Cannot locate specified workflow class: {workflow_name}.")
```

ericspod (changes requested):
I have a few concerns with backwards compatibility with this change. I see what your intent is and that it does work correctly, but if someone has a command line `python -m monai.bundle run some_expr` they would need to have `python -m monai.bundle run --run_id some_expr`? Even besides that, because the arguments are being removed from the signature of the function the help output from Fire gives less information about what the arguments are, and what they are for, since all the important values get put into `**kwargs`.

ericspod:
Also it looks like the premerge tests are failing for related reasons, that is the command line has changed.

monai/bundle/scripts.py
```diff
@@ -580,102 +581,59 @@ def get_bundle_info(
 
 
 @deprecated_arg("runner_id", since="1.1", removed="1.3", new_name="run_id", msg_suffix="please use `run_id` instead.")
-def run(
```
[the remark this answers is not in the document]
Nic-Ma: Hi @ericspod ,

I see your point, adding another CLI entry may be a good idea, let me investigate it.

Thanks.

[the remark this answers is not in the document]
Nic-Ma: Added a new entry in the latest commit.

Thanks.

monai/bundle/scripts.py
Nic-Ma: Sure, added in the latest commit.

Thanks.

monai/bundle/scripts.py
Nic-Ma: updated in the latest commit.

Thanks.

monai/bundle/scripts.py
```diff
@@ -580,102 +581,59 @@ def get_bundle_info(
...
+    if isinstance(workflow_name, str):
+        workflow_class, has_built_in = optional_import("monai.bundle", name=f"{workflow_name}")  # search built-in
+        if not has_built_in:
+            workflow_class = locate(f"{workflow_name}")  # search dotted path
+        if workflow_class is None:
+            raise ValueError(f"can not locate specified workflow class: {workflow_name}.")
+    elif issubclass(workflow_name, BundleWorkflow):
+        workflow_class = workflow_name
+    else:
+        raise ValueError(f"`workflow` must be the bundle workflow class name, but got: {workflow_name}.")
```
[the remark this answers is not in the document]
Nic-Ma: updated in the latest commit.

Thanks.

Nic-Ma:
Hi @ericspod ,

I totally updated the PR based on the comments and added tests, could you please help review it again?

Thanks in advance.

monai/bundle/scripts.py
```diff
@@ -678,6 +679,48 @@ def run(
...
+    if isinstance(workflow_name, str):
+        workflow_class, has_built_in = optional_import("monai.bundle", name=f"{workflow_name}")  # search built-in
+        if not has_built_in:
+            workflow_class = locate(f"{workflow_name}")  # search dotted path
+        if workflow_class is None:
+            raise ValueError(f"cannot locate specified workflow class: {workflow_name}.")
+    elif issubclass(workflow_name, BundleWorkflow):
+        workflow_class = workflow_name
+    else:
+        raise ValueError(f"argument `workflow` must be the bundle workflow class name or type, got: {workflow_name}.")
```
ericspod, line 716:
```suggestion
        raise ValueError(f"Argument `workflow` must be a bundle workflow class name or subclass of BundleWorkflow, got: {workflow_name}.")
```

monai/bundle/scripts.py
```diff
@@ -678,6 +679,48 @@ def run(
...
+        workflow: specified bundle workflow name, should be a string or class, default to "ConfigWorkflow".
+        args_file: a JSON or YAML file to provide default values for this API.
+            so that the command line inputs can be simplified.
+        kwargs: arguments to instantiate the workflow class.
+
+    """
+
+    _args = _update_args(args=args_file, workflow=workflow, **kwargs)
+    _log_input_summary(tag="run", args=_args)
+    (workflow_name,) = _pop_args(_args, workflow=ConfigWorkflow)  # the default workflow name is "ConfigWorkflow"
```
ericspod, line 706:
```suggestion
    workflow_name, = _pop_args(_args, workflow=ConfigWorkflow)  # the default workflow name is "ConfigWorkflow"
```

monai/bundle/scripts.py
```diff
@@ -678,6 +679,48 @@ def run(
...
+    if isinstance(workflow_name, str):
+        workflow_class, has_built_in = optional_import("monai.bundle", name=f"{workflow_name}")  # search built-in
```
ericspod, line 708:
```suggestion
        workflow_class, has_built_in = optional_import("monai.bundle", name=str(workflow_name))  # search built-in
```

monai/bundle/scripts.py
```diff
@@ -678,6 +679,48 @@ def run(
...
+        if not has_built_in:
+            workflow_class = locate(f"{workflow_name}")  # search dotted path
```
ericspod, line 710:
```suggestion
            workflow_class = locate(str(workflow_name))  # search dotted path
```

monai/bundle/scripts.py
```diff
@@ -678,6 +679,48 @@ def run(
     workflow.finalize()
 
 
+def run_workflow(workflow: str | BundleWorkflow | None = None, args_file: str | None = None, **kwargs: Any) -> None:
+    """
+    Specify `bundle workflow` to run monai bundle components and workflows.
```
ericspod, line 684: We should elaborate a little more here on what customised workflow classes would be and how they would operate. What's required here is to create a subclass of `BundleWorkflow` then pass it in or mention it by name, which you can on the command line. We should just state the user needs to provide that subclass and be available to import, the documentation for how to do that is elsewhere.

ericspod (approved):
Looks good, I have very minor style suggestions. I see you're using `pydoc.locate` here and I wonder if we should have our own function in `monai.utils.module` for doing this. This is a topic for some other time and place but I had used something like this in the past that worked well enough:

```python
def find_type_def(qualified_name: str):
    if "." not in qualified_name:
        modname, defname = "builtins", qualified_name
    else:
        modname, defname = qualified_name.rsplit(".", 1)

    mod = __import__(modname)
    return getattr(mod, defname)
```

monai/bundle/scripts.py
Nic-Ma: Updated in the latest commit, thanks.

monai/bundle/scripts.py
Nic-Ma: Updated in the latest commit, thanks.

monai/bundle/scripts.py
Nic-Ma: Sure, added doc-string in the latest commit, thanks.

Nic-Ma:
Hi @ericspod ,

Thanks for your review, I updated the PR accordingly.
And for the `locate` suggestion, I created a ticket to discuss and track: https://github.com/Project-MONAI/MONAI/issues/6275.
Because we use `locate` in several places of current codebase:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/networks/nets/flexible_unet.py#L16

https://github.com/Project-MONAI/MONAI/blob/dev/monai/utils/module.py#L244

https://github.com/Project-MONAI/MONAI/blob/dev/monai/transforms/io/array.py#L208

If we want to change them together, maybe better to do a separate PR?

Thanks.

Nic-Ma:
Hi @wyli ,

Do you know the root cause of Blossom system error here?

Thanks in advance.

PR #6203, opened 2023-03-20
Enable swinunetr-v2

### LINKED ISSUE #6183: Improve SwinUNETR pipeline by adding SwinUNETR-V2
Add swinunetr-v2 as option for swinunetr initialization.

heyufan1995:
Fixes #6183 .

### Description
Added a "use_v2" option in swinunetr initialization. Default is false will not affect the original swinunetr.
Once changed to true, will become swinunetr-v2 with 4 additional convolution block.
Tested running from auto3dseg bundles, no change needed for original swinunetr, and works for swinunetr-v2
Tested running from monai research contribution repo for swinuntr, no change needed for original swinunetr, and works for swinunetr-v2
Tested TensorRT, compiled .ts file successfully.

### Types of changes


monai/networks/nets/swin_unetr.py
```diff
@@ -65,6 +65,7 @@ def __init__(
         use_checkpoint: bool = False,
         spatial_dims: int = 3,
         downsample="merging",
+        use_v2=False
```
tangy5, line 68: can we use the flag: version="v1" instead of use_v2=False?. We can use this flag to control different versions if there are variants in the future...

tangy5:
The V2 implementation is great, one more thing to discuss here. 
If we use the branching options format, there will be more challenges to use torch.jit.script for torchscript conversion, but now we do not have the plan to support scriptable conversion. Jit.trace should be good enough. Wenqi may provide more insights. Thanks!

wyli (approved):
thanks for confirming @tangy5, I'm merging this for now, and I think for future extensions we should start new classes instead of adding more logic to this class...

monai/networks/nets/swin_unetr.py
Nic-Ma: Hi @wyli ,

I agree with @tangy5 , "use_v2" will cause breaking change if we add "v3" later.. an option "version: int = 1" may be better?
What do you think?

Thanks.

wyli: this class will get complicated if we support three or more different modes, for readability I think we should start new classes for new architectural designs, like different versions of the inception models https://pytorch.org/vision/main/models/generated/torchvision.models.inception_v3.html

PR #6213, opened 2023-03-21
#6188 Allow user to define own `FolderLayout` in `SaveImage` and `SaveImaged` transforms

MathijsdeBoer:
Improves issues discussed in #6188 .

### Description

As I found out in #6188, `SaveImage` and `SaveImaged` provide a fairly rigid way to define the output path. I've made some changes that allow the user to more easily define their own `FolderLayout` derivative, and pass it to these transforms.
Additionally, I've added a boolean that adds a key `"saved_to"` to the metadata, so the user can more easily assess where their output actually went.

Finally, and most optionally, I've prepended `"meta"` to the `MetaTensor.__repr__` and `MetaTensor.__str__` outputs in an effort to more clearly show to the user that the printed tensor has special properties. This change was made in its own commit, which can be left out if undesired.

Other small changes to documentation formatting, mainly changing:

```text
"""
    ...
    some documentation `code` more docs
    ...
"""
```

to

```text
"""
    ...
    some documentation ``code`` more docs
    ...
"""
```

To format these correctly in the docs. Seemed like some people assumed the single backticks would format as `code`, as opposed to _italics_.

---

### Test results

`./runtests.sh -f -u --net --coverage`
results: 1 failure, 2 errors
The failure was related to me changing the `MetaTensor.__repr__`, fixed the test in a separate commit

`./runtests.sh --quick --unittests  --disttests`
In progress

### Types of changes


monai/data/meta_tensor.py
```diff
@@ -572,14 +572,14 @@ def __repr__(self):
...
-        return self.as_tensor().__repr__()
+        return "meta" + self.as_tensor().__repr__()
 
     def __str__(self):
         """
         Prints a representation of the tensor identical to ``torch.Tensor.__str__``.
         Use ``print_verbose`` for associated metadata.
         """
-        return str(self.as_tensor())
+        return "meta" + str(self.as_tensor())
```
wyli, line 582: I think this is good idea, could you please make a dedicated PR for this feature?

MathijsdeBoer: I've taken these changes out of this fork, to reapply in another PR!

MathijsdeBoer:
Will do! (Though I might have accidentally clicked a wrong on this page)

MathijsdeBoer:
I hope it's okay if I didn't rerun the full test suite, and only ran the edited tests. It takes about an hour to do on my work PC, which is considerably more powerful than my personal PC that I used for these last few commits.

monai/transforms/io/array.py
```diff
@@ -478,6 +503,8 @@ def __call__(self, img: torch.Tensor | np.ndarray, meta_data: dict | None = None
                 )
             else:
                 self._data_index += 1
+                if self.savepath_in_metadict and meta_data is not None:
```
Nic-Ma, line 506: Hi @wyli ,

I don't remember clearly, why we support multiple writers here with a `for loop`?
If so, I think we should not update "self._data_index += 1" for every writer, is it a bug?

And @MathijsdeBoer , here if we save the "filename" to `meta_data`, will this information be saved into the nifti file header? If yes, I think we should only save the filename instead of the full path which may contain the user's personal information?

Thanks.

monai/transforms/io/array.py
```diff
@@ -315,11 +316,14 @@ class SaveImage(Transform):
 
     Args:
         output_dir: output image directory.
+        Handled by ``folder_layout`` instead, if ``folder_layout`` not ``None``.
```
Nic-Ma, line 319: if ``folder_layout`` **is** not ``None``.
Same for other places.

Thanks.

monai/transforms/io/dictionary.py
```diff
@@ -211,41 +216,47 @@ class SaveImaged(MapTransform):
...
+            if ``None``, use the default writer from ``monai.data.image_writer`` according to ``output_ext``.
             if it's a string, it's treated as a class name or dotted path;
             the supported built-in writer classes are ``"NibabelWriter"``, ``"ITKWriter"``, ``"PILWriter"``.
         output_name_formatter: a callable function (returning a kwargs dict) to format the output file name.
             see also: :py:func:`monai.data.folder_layout.default_name_formatter`.
-
+            If using a custom ``folder_layout``, consider providing your own formatter.
+        folder_layout: A customized ``monai.data.FolderLayoutBase`` subclass to define file naming schemes.
+            if ``None``, uses the default ``FolderLayout``
+        savepath_in_metadict: if ``True``, adds a key ``saved_to`` to the metadata, which contains the path
```
KumoLiu, line 258: May I know the use case when we need a `save_to` key in metadata? Does `print_log` not enough?
I am concerned when users pre-processing the same data twice and saved it twice, the `save_to` may be confused here?

KumoLiu (commented):
Thanks for the contributions! Few comments in line.
You could also run `./runtests.sh --autofix` for auto style fixes.

monai/transforms/io/array.py
MathijsdeBoer: I was under the impression that saving things to the `meta_data` was fine. A quick skim of the `Writer` code, it seems like the `NiBabelWriter` specifically gets certain keys, rather than taking them all. 
Additionally, I believe that the Nifti specification is quite strict with what the header can contain, i.e.: very little to no room for custom data. If this is also true in the case for, for example, a `.mhd/.mha`, which is a little more freeform, I don't know.
Finally, the key is inserted into the `meta_data` _after_ writing the file to disk, which would not include it in the resulting file either way.

monai/transforms/io/array.py
MathijsdeBoer: Will fix

monai/transforms/io/dictionary.py
MathijsdeBoer: `print_log` only, well, prints. If I want to do something with the path to where the images were saved I'd need to either parse the output logs on the fly, or rerun the `FolderLayout` code to get the path. It would be a lot more efficient and less prone to error to save the output path to a variable.

My use case was that I'm trying to set up an experiment across several datasets, which has me create `Dataset` objects. Each of these objects holds a list of paths for the raw data. After preprocessing these paths, I want to add the outputs of the `SaveImaged` to another list in the dataset which contains the locations of the saved images. Currently, there is no simple way to either get the path, or infer the path. Adding `"save_to"` to the dict would at least simplify things.

For more details, see #6188

MathijsdeBoer:
Haha, now I'm starting to get slightly embarrased about the amount of small commits I have to make because I thought I had ran every test, but forgot _that one_, I'll get to work on fixing that mypy error.

monai/transforms/io/dictionary.py
KumoLiu: Hi @MathijsdeBoer, I may get your points.
In your case, you want to prepare a dataset from several sides, but I think the output path is been standardized (such as "E:\...\Datasets\BraTS2021\Preprocessed")? And which is specified in the "output_dir"? I would like to say I didn't find the benefits from this "save_to" but may get some confusion when you want to preprocess the same data twice. 

But it's okay if you all think this arg is ok. cc @wyli @Nic-Ma

MathijsdeBoer: How else would you suggest I programmatically store the output of `SaveImage` or `SaveImaged` in a list if none of the outputs actually provide this information?

As for preprocessing twice, It could be changed to be `"saved_to_{n}"`, where `n` is the nth time this particular `MetaTensor` has been saved?

wyli: I think the current PR's implementation is ok for now, another idea is to set the path string as a property of SaveImage, then after saving we can get the path by `saver.last_saved_filename` cc @ericspod

wyli (approved):
thanks, I think the

PR #6214, opened 2023-03-21
#6213 Prepend `"meta"` to `MetaTensor.__repr__` and `MetaTensor.__str__` for easier identification

MathijsdeBoer:
Discussed in #6213

### Description

Prepends `"meta"` to the `MetaTensor.__repr__` and `MetaTensor.__str__` output so printing a MetaTensor does not look the exact same as a regular `torch.Tensor`.

I don't expect this change to cause any breaks, with me running the risk of invoking [xkcd 1172](https://xkcd.com/1172/).

---

1 failure in `./runtests.sh -f -u --net --coverage`:
```text
FAIL: test_values (tests.test_tciadataset.TestTciaDataset)
  File "D:\PythonProjects\MONAI-SaveImageFormatting\tests\test_tciadataset.py", line 72, in test_values
    self.assertTrue(
```

### Types of changes


MathijsdeBoer:
@wyli Running `pip freeze` on the venv with which I'm making my changes contains `torch==2.0.0`, is this change to the `torch.Tensor` printing in an in-dev PyTorch version? I ask because I wasn't seeing this behaviour myself.

I've tried seeing what would happen if I would comment out `MetaTensor.__repr__` and `MetaTensor.__str__`:

```python
Python 3.10.10 (tags/v3.10.10:aad5f6a, Feb  7 2023, 17:20:36) [MSC v.1929 64 bit (AMD64)] on win32
>>> import monai
>>> mt = monai.data.MetaTensor([1.])
>>> print(mt)
Metadata
        [0., 1., 0., 0.],
        [0., 0., 1., 0.],
	space: RAS
Applied operations
```

wyli:
Hi @MathijsdeBoer my example in the previous comment is not about monai, it's a pytorch demo to show subclassing...

wyli:
these are from the MetaObj... https://github.com/Project-MONAI/MONAI/blob/d57bff6fa8fd3f18b6cca27d1b35d6f296215984/monai/data/meta_obj.py#L157-L168

that's why I think this PR is a good idea..

wyli:
comment out `MetaTensor.__repr__` and `MetaTensor.__str__`, and change this:
` class MetaTensor(MetaObj, torch.Tensor):`
to
`
 class MetaTensor(torch.Tensor, MetaObj):
`
will get the `__repr__` resolved to pytorch's implementation :)
but these changes will require more thorough tests

PR #6237, opened 2023-03-26
Add onnx export with verification

liqunfu:
Fixes # Issue#TBD.

### Description
A MONAI Bundle can contain models in PyTorch, TorchScript, and ONNX formats.
Utility exists in MONAI to export a model in TorchScript format. Exported TorchScript models can be verified with Torch runtime. 
It is helpful (and complete) to provide a same utility to export ONNX models and to validate the models with PyTorch, ONNX(https://github.com/onnx/onnx), and onnxruntime (https://github.com/microsoft/onnxruntime).

With the proposed utility, ONNX model can be added to bundles in model-zoo with calling monai.bundle ckpt_export.

### Types of changes

- [x ] New tests added to cover the changes.

`./runtests.sh --quick --unittests  --disttests`
TOTAL                                                               72191  11290    84%


monai/networks/utils.py
```diff
@@ -552,6 +555,80 @@ def save_state(src: torch.nn.Module | dict, path: PathLike, **kwargs):
     save_obj(obj=ckpt, path=path, **kwargs)
 
 
+def convert_to_onnx(
```
ericspod, line 558: With `convert_to_torchscript` we have the option to save to a file, can we add that here?

tests/test_convert_to_onnx.py
```diff
@@ -0,0 +1,49 @@
...
+from monai.networks import convert_to_onnx
+from monai.networks.nets import UNet
+
+
+TESTS = list(itertools.product(["cpu", "cuda"], [True, False], ["CUDAExecutionProvider", "CPUExecutionProvider"]))
+
+class TestConvertToOnnx(unittest.TestCase):
+    @parameterized.expand(TESTS)
+    def test_value(self, device, use_ort=False, ort_provider=None):
+        model = UNet(
```
ericspod, line 30: `UNet` should be compatible with ONNX but not all models will be so we'll have to have per-model tests like we do with Torchscript. This can be left to future PRs if it's not a priority however.

binliunls:
Hi @liqunfu ,
     Do you have an ETA for this PR? We are working on this too. If you can have this PR merged this week, we can wait. Otherwise we can start a PR for this and invite you to review it.

Thanks,
Bin

liqunfu:
> Hi @liqunfu , Do you have an ETA for this PR? We are working on this too. If you can have this PR merged this week, we can wait. Otherwise we can start a PR for this and invite you to review it.
> 
> Thanks, Bin

I am starting today to update my PR with reviewers' comments and to fix CI failures. I was trying to add more test for models that are used by model-zoo but it is time consuming because I need to get evaluation data some of which also need to wait for access approval. I agree with @ericspod that it can be in another PR.

BTW thanks all for reviewing this PR. I will update it by the end of today.

liqunfu:
> > Hi @liqunfu , Do you have an ETA for this PR? We are working on this too. If you can have this PR merged this week, we can wait. Otherwise we can start a PR for this and invite you to review it.
> > Thanks, Bin
> 
> I am starting today to update my PR with reviewers' comments and to fix CI failures. I was trying to add more test for models that are used by model-zoo but it is time consuming because I need to get evaluation data some of which also need to wait for access approval. I agree with @ericspod that it can be in another PR.
> 
> BTW thanks all for reviewing this PR. I will update it by the end of today.

@binliunls Please feel free to add to this PR with what you may already have so we can merge sooner.

.github/workflows/pythonapp.yml
```diff
@@ -150,6 +150,7 @@ jobs:
         # however, "pip install monai*.tar.gz" will build cpp/cuda with an isolated
         # fresh torch installation according to pyproject.toml
         python -m pip install torch>=1.9 torchvision
+        python -m pip install onnx onnxruntime
```
Nic-Ma, line 153: @wyli I think we don't need to install here?

Thanks.

monai/bundle/scripts.py
```diff
@@ -932,6 +940,92 @@ def _export(
...
+        filepath=filepath,
+        meta_file=meta_file,
+        config_file=config_file,
+        ckpt_file=ckpt_file,
+        key_in_ckpt=key_in_ckpt,
+        use_trace=use_trace,
+        input_shape=input_shape,
+        **override,
+    )
+    _log_input_summary(tag="ckpt_export", args=_args)
```
Nic-Ma, line 993: tag should be "onnx_export"

Nic-Ma (commented):
Hi @liqunfu ,

About adding a new optional dependency package, please refer to:
https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#adding-new-optional-dependencies

Thanks.

Nic-Ma:
Mark: This PR will be targeting MONAI 1.2 P1 feature as several bundles which can't be supported by `torch-trt` so far can leverage `ONNX path` to convert to TensorRT model.

Thanks.

monai/networks/utils.py
liqunfu: updated the function so that it can be called with a filename. Converted onnx model will be saved to the file if provided.

tests/test_convert_to_onnx.py
liqunfu: I added test case for SegResNet. More test cases shall be added in subsequent PRs.

.github/workflows/pythonapp.yml
liqunfu: you are correct @Nic-Ma , I have the line removed - got to know it more.

monai/bundle/scripts.py
liqunfu: Thanks for suggesting the lazy import. I have updated to use it. Also updated to onnx-export.

liqunfu:
> Mark: This PR will be targeting MONAI 1.2 P1 feature as several bundles which can't be supported by `torch-trt` so far can leverage `ONNX path` to convert to TensorRT model.
> 
> Thanks.

I am very glad that onnx and onnxruntime can offer its value for MONAI.

monai/bundle/scripts.py
```diff
@@ -1010,17 +1011,30 @@ def ckpt_export(
...
-        parser,
-        net_id=net_id_,
-        filepath=filepath_,
-        ckpt_file=ckpt_file_,
-        config_file=config_file_,
-        key_in_ckpt=key_in_ckpt_,
-        use_trace=use_trace_,
-        inputs=inputs_,
-    )
+    export_onnx = filepath_.endswith(".onnx")
```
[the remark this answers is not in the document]
binliunls: Thanks for updating. I think the model saving part in the `_export` function blocks you from using it. Please go ahead with your way. I can make a PR to update it later.

Thanks in advance.

monai/networks/utils.py
```diff
@@ -552,6 +554,92 @@ def save_state(src: torch.nn.Module | dict, path: PathLike, **kwargs):
...
+        if use_trace:
+            script_module = torch.jit.trace(model, example_inputs=inputs)
+        else:
+            script_module = torch.jit.script(model, **kwargs)
+        if filename is None:
+            f = io.BytesIO()
+            torch.onnx.export(script_module, inputs, f=f, input_names=input_names, output_names=output_names)
+            onnx_model = onnx.load_model_from_string(f.getvalue())
+        else:
+            torch.onnx.export(script_module, inputs, f=filename, input_names=input_names, output_names=output_names)
```
binliunls, line 607: Hi @liqunfu ,
     Could you please also add the `dynamic_axes` parameter to the `convert_to_onnnx` function and the `torch.onnx_export`, in case the different batch sizes are needed like in the `trt_export` API.

Thanks,
Bin

tests/test_convert_to_onnx.py
ericspod: We can test the networks like we do for Torchscript, that is testing with random data on small versions of the networks since we're not concerned with results. Testing the network as they are in core suffices for testing the bundles that use them, you don't need to worry about testing bundles directly.

liqunfu: Thanks @ericspod , I created test_onnx_save test utility and applied it to TestRetinaNet. It does no cross validation between PyTorch and onnxruntime.

monai/bundle/scripts.py
```diff
@@ -932,6 +939,93 @@ def _export(
...
+
+    parser.read_config(f=config_file_)
+    if meta_file_ is not None:
+        parser.read_meta(f=meta_file_)
+
+    # the rest key-values in the _args are to override config content
+    for k, v in _args.items():
+        parser[k] = v
+
+    inputs_ = [torch.rand(input_shape_)]
```
Nic-Ma, line 1017: Hi @binliunls ,

Please make below logic as a private function:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/scripts.py#L1132-L1135
And enable all the export APIs to get input shape from metadata if `input_shape` is None and `use_trace=True`.
And update the doc-string to make the logic clear.

Thanks.

monai/bundle/scripts.py
```diff
@@ -932,6 +939,93 @@ def _export(
...
+        meta_file=meta_file,
+        config_file=config_file,
+        ckpt_file=ckpt_file,
+        key_in_ckpt=key_in_ckpt,
+        use_trace=use_trace,
+        input_shape=input_shape,
+        **override,
+    )
+    _log_input_summary(tag="onnx_export", args=_args)
+    filepath_, ckpt_file_, config_file_, net_id_, meta_file_, _, _, input_shape_ = _pop_args(
```
Nic-Ma, line 995: Why you ignored the `key_in_ckpt` and `use_trace` args here?

Thanks.

monai/bundle/scripts.py
```diff
@@ -932,6 +939,93 @@ def _export(
...
+        _args,
+        "filepath",
+        "ckpt_file",
+        "config_file",
+        net_id="",
+        meta_file=None,
+        key_in_ckpt="",
+        use_trace=True,
```
Nic-Ma, line 1003: the default should be `False`?

Thanks.

monai/bundle/scripts.py
```diff
@@ -932,6 +939,93 @@ def _export(
...
+
+    inputs_ = [torch.rand(input_shape_)]
+    net = parser.get_parsed_content(net_id_)
+    if has_ignite:
+        # here we use ignite Checkpoint to support nested weights and be compatible with MONAI CheckpointSaver
+        Checkpoint.load_objects(to_load={key_in_ckpt: net}, checkpoint=ckpt_file_)
+    else:
+        ckpt = torch.load(ckpt_file_)
+        copy_model_state(dst=net, src=ckpt if key_in_ckpt == "" else ckpt[key_in_ckpt])
+    onnx_model = convert_to_onnx(model=net, inputs=inputs_, use_trace=use_trace if use_trace else False)
```
Nic-Ma, line 1025: `use_trace=use_trace_` is enough as we already set default in the `pop` call above.

Thanks.

monai/bundle/scripts.py
```diff
@@ -932,6 +939,93 @@ def _export(
...
+    meta_file: str | Sequence[str] | None = None,
+    config_file: str | Sequence[str] | None = None,
+    key_in_ckpt: str | None = None,
+    use_trace: bool | None = None,
+    input_shape: Sequence[int] | None = None,
+    args_file: str | None = None,
+    **override: Any,
+) -> None:
+    """
+    Export the model checkpoint to onnx with metadata and config included as JSON files.
```
Nic-Ma, line 955: I don't think ONNX export has the logic "metadata and config included as JSON files" as described.

Thanks.

monai/bundle/scripts.py
```diff
@@ -932,6 +939,93 @@ def _export(
...
+        filepath: filepath where the onnx model is saved to.
+        ckpt_file: filepath of the model checkpoint to load.
+        meta_file: filepath of the metadata file, if it is a list of file paths, the content of them will be merged.
+        config_file: filepath of the config file that contains extract network information,
+        key_in_ckpt: for nested checkpoint like `{"model": XXX, "optimizer": XXX, ...}`, specify the key of model
+            weights. if not nested checkpoint, no need to set.
+        use_trace: whether using `torch.jit.trace` to convert the pytorch model to torchscript model.
+        input_shape: must specify the `input_shape` of the network to convert the model when `use_trace` is True.
+            Should be a list like [N, C, H, W] or [N, C, H, W, D].
+        args_file: a JSON or YAML file to provide default values for `meta_file`, `config_file`,
```
Nic-Ma, line 974: provide default values for `meta_file`, `config_file`net_id` and override pairs` -> "provide default values for all the parameters of this function".
Please also update all the export functions for this arg.

Thanks.

monai/bundle/scripts.py
```diff
@@ -932,6 +939,93 @@ def _export(
...
+            `net_id` and override pairs. so that the command line inputs can be simplified.
+        override: id-value pairs to override or add the corresponding config content.
+            e.g. ``--_meta#network_data_format#inputs#image#num_channels 3``.
+
+    """
+    onnx, _ = optional_import("onnx")
```
Nic-Ma, line 980: Please import ONNX in the `convert_to_onnx` instead.

Thanks.

monai/bundle/scripts.py
```diff
@@ -932,6 +939,93 @@ def _export(
...
+def onnx_export(
+    net_id: str | None = None,
+    filepath: PathLike | None = None,
+    ckpt_file: str | None = None,
+    meta_file: str | Sequence[str] | None = None,
+    config_file: str | Sequence[str] | None = None,
+    key_in_ckpt: str | None = None,
+    use_trace: bool | None = None,
+    input_shape: Sequence[int] | None = None,
+    args_file: str | None = None,
```
Nic-Ma, line 951: Please add `converter_kwargs` parameter to pass all the other args for the `convert_to_onnx()` except for `model, inputs, use_trace`.
Please also add it to `ckpt_export()` and `trt_export()`.

Thanks.

Nic-Ma (commented):
Hi @wyli ,

Do we need to add
```
onnx
onnxruntime; python_version <= '3.10'
```
to all the required places by: https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#adding-new-optional-dependencies?

Thanks.

Nic-Ma:
Hi @binliunls ,

Please help address the comments.

Thanks in advance.

docs/requirements.txt
```diff
@@ -35,3 +35,5 @@ h5py
 nni
 optuna
 opencv-python-headless
+onnx
+onnxruntime; python_version <= '3.10'
```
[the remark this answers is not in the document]
liqunfu: onnxruntime does not have release package for python 3.11. https://pypi.org/project/onnxruntime/#files

monai/bundle/scripts.py
liqunfu: PyTorch trace is in general more reliable than scripting. scripting is useful for loop and if which is no comment in computer vision. But if it is a convention in MONAI, we can change default to False.

monai/bundle/scripts.py
liqunfu: > Hi @binliunls ,
> 
> Please help address the comments.
> 
> Thanks in advance.

Thank you @Nic-Ma  and @binliunls !

monai/bundle/scripts.py
Nic-Ma: I would suggest to use `jit.script()` if possible, especially for dynamic input shapes.
@ericspod Do you have any comments?

Thanks.

monai/bundle/scripts.py
binliunls: Hi @Nic-Ma ,
     The last line in this API using the onnx to save models.  I think it is necessary here, unless we have a saver to replace this one.

Thanks,
Bin

Nic-Ma: OK, how about moving it to the top of this file?
I think the reason why we import `torch-trt` in the function is just a workaround for the import issue that changes CUDA GPU index. @wyli .

Thanks.

tests/test_convert_to_onnx.py
binliunls: Hi @liqunfu ,
     Why do you add a `test_onnx_save` and apply it to TestRetinaNet. I think the `test_convert_to_onnx` and `test_bundle_onnx_export` is enough for this PR. We may add the bundle onnx test in MONAI model-zoo to test more models. Could you please remove these tests, if there are no other reasons?

Thanks,
Bin

liqunfu: As @ericspod commented, we need a way to test MONAI's nets with onnx in the same way nets are tested for TorchScript. For this purpose, RetinaNet is tested in test_retinanet via test_onnx (similar to test_script) by calling test_onnx_save. Eventually, all test_xxxnet.py (or similar) will have a test_onnx call to validate the net's definition with onnx export. It is for validation of individual torch nnmodules in MONAI.

On the other hand, test_convert_to_onnx is to test convert_to_onnx utility in case it is broken.

test_bundle_onnx_export (thank to you for adding it!) is to ensure the utility added to script (onnx_export) is working as we expected. so that we can use it to generate model bundles.

monai/bundle/scripts.py
ericspod: We should use `jit.script` where we can but there's times that we would want to use `jit.trace`. Not all networks are compatible with `jit.script` (or at least weren't in the past with previous versions of Pytorch) so we need that option, I'm not aware of any other reason to use one function or the other but they must exist. Networks created with `jit.script` appear compatible with ONNX from my testing.

binliunls:
Hi @wyli ,
    The windows-latest ci/cd test reports OOM error. Could you please give me some advice on why this happened and how to fix it?

error:
```
ERROR: test_vnet_shape_5 (tests.test_vnet.TestVNet)
  File "C:\hostedtoolcache\windows\Python\3.8.10\x64\lib\site-packages\parameterized\parameterized.py", line 620, in standalone_func
  File "D:\a\MONAI\MONAI\tests\test_vnet.py", line 69, in test_vnet_shape
    net = VNet(**input_param).to(device)
  File "D:\a\MONAI\MONAI\monai\networks\nets\vnet.py", line 241, in __init__
    self.up_tr256 = UpTransition(spatial_dims, 256, 256, 2, act, dropout_prob=dropout_prob)
  File "D:\a\MONAI\MONAI\monai\networks\nets\vnet.py", line 151, in __init__
    self.ops = _make_nconv(spatial_dims, out_channels, nconvs, act)
  File "D:\a\MONAI\MONAI\monai\networks\nets\vnet.py", line 55, in _make_nconv
    layers.append(LUConv(spatial_dims, nchan, act, bias))
  File "D:\a\MONAI\MONAI\monai\networks\nets\vnet.py", line 36, in __init__
    self.conv_block = Convolution(
  File "D:\a\MONAI\MONAI\monai\networks\blocks\convolutions.py", line 143, in __init__
    conv = conv_type(
  File "C:\hostedtoolcache\windows\Python\3.8.10\x64\lib\site-packages\torch\nn\modules\conv.py", line 591, in __init__
    super(Conv3d, self).__init__(
  File "C:\hostedtoolcache\windows\Python\3.8.10\x64\lib\site-packages\torch\nn\modules\conv.py", line 137, in __init__
    self.weight = Parameter(torch.empty(
RuntimeError: [enforce fail at C:\actions-runner\_work\pytorch\pytorch\builder\windows\pytorch\c10\core\impl\alloc_cpu.cpp:72] data. DefaultCPUAllocator: not enough memory: you tried to allocate 32768000 bytes.

FAIL: test_cachedataset (tests.test_sampler_dist.DistributedSamplerTest)
  File "D:\a\MONAI\MONAI\tests\utils.py", line 539, in _wrapper
    assert results.get(), "Distributed call failed."
AssertionError: Distributed call failed.
```

Thanks,
Bin

monai/bundle/scripts.py
Nic-Ma: Yes, that's why I would suggest to set the default to `False`, and I agree we need this parameter.

Thanks.

setup.cfg
```diff
@@ -77,6 +77,8 @@ all =
     h5py
     nni
     optuna
+    onnx>=1.13.0
+    onnxruntime
```
Nic-Ma, line 81: Hi @binliunls ,

Please add below things in this postion:
https://github.com/Project-MONAI/MONAI/blob/dev/setup.cfg#L135
```
nni =
    nni
optuna =
    optuna
onnx =
    onnx>=1.13.0
    onnxruntime
```

Nic-Ma (approved):
Looks good to me, put some minor comments inline.
@wyli Do you think we need to add all the optional packages in this function or only selected?
https://github.com/Project-MONAI/MONAI/blob/dev/monai/config/deviceconfig.py#L64

Thanks.

monai/bundle/scripts.py
```diff
@@ -48,6 +55,7 @@
 ValidationError, _ = optional_import("jsonschema.exceptions", name="ValidationError")
 Checkpoint, has_ignite = optional_import("ignite.handlers", IgniteInfo.OPT_IMPORT_VERSION, min_version, "Checkpoint")
 requests_get, has_requests = optional_import("requests", name="get")
+onnx, has_onnx = optional_import("onnx")
```
Nic-Ma, line 58: No need this `has_onnx` var?

Thanks.

docs/requirements.txt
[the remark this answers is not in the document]
binliunls: Hi @Nic-Ma @wyli ,
     What should I do with this `python_version`?

Thanks,
Bin

setup.cfg
Nic-Ma, line 81: Hi @wyli , should we add " python_version <= '3.10'" here?

Thanks.

monai/bundle/scripts.py
```diff
@@ -980,6 +1001,114 @@ def _export(
...
+        net_id: ID name of the network component in the config, it must be `torch.nn.Module`.
+        filepath: filepath where the onnx model is saved to.
+        ckpt_file: filepath of the model checkpoint to load.
+        meta_file: filepath of the metadata file, if it is a list of file paths, the content of them will be merged.
+        config_file: filepath of the config file that contains extract network information,
+        key_in_ckpt: for nested checkpoint like `{"model": XXX, "optimizer": XXX, ...}`, specify the key of model
+            weights. if not nested checkpoint, no need to set.
+        use_trace: whether using `torch.jit.trace` to convert the pytorch model to torchscript model.
+        input_shape: a shape used to generate the random input of the network, when converting the model to an
+            onnx model.Should be a list like [N, C, H, W] or [N, C, H, W, D]. If not given, will try to parse from
```
Nic-Ma, line 1036: Missed a blank: "to an onnx model.Should be a list like"

tests/test_bundle_onnx_export.py
```diff
@@ -0,0 +1,69 @@
...
+
+from monai.bundle import ConfigParser
+from monai.networks import save_state
+from tests.utils import SkipIfBeforePyTorchVersion, SkipIfNoModule, command_line_tests, skip_if_windows
+
+TEST_CASE_1 = ["True"]
+TEST_CASE_2 = ["False"]
+
+
+@skip_if_windows
```
Nic-Ma, line 28: Hi @liqunfu ,

May I know why we skip tests on windows here?
I also see some CI failure about the RetinaNet tests on windows, could you please help take a look?

Thanks.

Nic-Ma (approved):
Thanks for the update, looks good to me.
Please fix the failed CI tests on windows, if it's expected error, please add `skip_if_windows` decorator.

Thanks.

tests/test_bundle_onnx_export.py
liqunfu: I see test_bundle_trt_export and test_bundle_ckpt_export are also skipped on Windows. I think the common reason they are skipped is because the way configuration files are parsed by subprocess.run. It fails with this argument, f"['{config_file}','{def_args_file}']"].
I did not fix this because it is a pre-existing issue, I followed the way with which trt and ckpt bundle tests were treated to annotate the test with @skip_if_windows. 

Let me check the CI failure.

binliunls: Hi @liqunfu ,
    I have fixed the CI bug by add `atol=0.1` to the unit test. Currently the ci failure is more like a github issue.
    @wyli  Could you please take a look at the window-latest test? It blocked this PR from merging.

Thanks in advance.

binliunls: Hi @liqunfu ,
     You are the contributor to onnx right? Do you have some knowledge about whether it's possible that the import of onnx and onnxruntime lead to the memory overhead issue on windows?

Thanks,
Bin

liqunfu: Hi @binliunls @wyli @Nic-Ma , 

It turned out that the CI failure was caused by a bug in pythonapp.yml (https://github.com/Project-MONAI/MONAI/pull/6237/commits/3a80a9810aeaca84bbe4af1c7e86061af2eb4f96). I made a correction and the CI pipeline is passing.
I see all required pipelines have passed. Hope the rest will complete without any issue.
Please let me know if there is anything I need to do to have this PR merged.

Thank you,
Liqun

wyli: thanks for looking into this @liqunfu, I didn't pay attention to the page file config issue https://github.com/al-cheb/configure-pagefile-action/issues/5 it's been there for a while, interesting that we didn't notice this issue earlier...I'll merge this for now and verify it with all the postmerge tests.

Nic-Ma: @liqunfu Thanks very much for your help here!

binliunls: @liqunfu , thank you for your your help!

PR #6239, opened 2023-03-26
Enable algorithm skip in bundle_gen

### LINKED ISSUE #6184: Skip 2D/3D training if the data slice thickness are highly anisotropic

heyufan1995:
Fixes #6184  .

### Description

Enable bundlle_gen to skip an algorithm if the BundleAlgo pre_check_skip_algo() function sets skip_bundlegen=True.  SegResNet2D overrided pre_check_skip_algo() and will be skipped if data is not highly anisotropic.

### Types of changes


mingxin-zheng:
Thank you @heyufan1995 for the PR. 

I think it would be better to set the `skip_bundlegen` 's default value in the __init__ function too.

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -80,6 +80,15 @@ def __init__(self, template_path: str):
         # track records when filling template config: {"<config name>": {"<placeholder key>": value, ...}, ...}
         self.fill_records: dict = {}
 
+    def pre_check_skip_algo(self):
```
mingxin-zheng, line 83: Maybe we can make the function inputs more general? For example `pre_check_skip_algo(self, skip_bundlegen: bool = False, skip_info:str = "")

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -511,15 +520,20 @@ def generate(
...
-                if gpu_customization:
-                    gen_algo.export_to_disk(
-                        output_folder,
-                        name,
-                        fold=f_id,
-                        gpu_customization=True,
-                        gpu_customization_specs=gpu_customization_specs,
-                    )
+                gen_algo.pre_check_skip_algo()
+                if not gen_algo.skip_bundlegen:
```
mingxin-zheng, line 524: Maybe we can commit less line changes by:
```
if gen_algo.skip_bundlegen:
	continue

if gpu_customization:
	...
```

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -511,15 +520,20 @@ def generate(
...
+                    else:
+                        gen_algo.export_to_disk(output_folder, name, fold=f_id)
+
+                    algo_to_pickle(gen_algo, template_path=algo.template_path)
+                    self.history.append({name: gen_algo})  # track the previous, may create a persistent history
                 else:
-                    gen_algo.export_to_disk(output_folder, name, fold=f_id)
-                algo_to_pickle(gen_algo, template_path=algo.template_path)
-                self.history.append({name: gen_algo})  # track the previous, may create a persistent history
+                    logger.info(f"{name} is skipped! Msg:{gen_algo.skip_info}")
```
mingxin-zheng, line 539: The Msg is empty by default and the message `Msg: ` will look confusing.

monai/apps/auto3dseg/bundle_gen.py
myron: I recommend this function to return a tuple (bool, str), E.g.
return (False, '') - default, and similar from each "algo" that supports it

In this case, you don't need to set class variables self.skip_bundlegen, and self.skip_info
since the only place that information is used is in generate() function

monai/apps/auto3dseg/bundle_gen.py
heyufan1995: removed msg:

monai/apps/auto3dseg/bundle_gen.py
heyufan1995: thanks, updated

monai/apps/auto3dseg/bundle_gen.py
heyufan1995: @myron Thanks Andriy, I've changed the function to return a tuple.

wyli (approved):

PR #6241, opened 2023-03-27
update docstring and eval with overrides

wyli:
- docstring updates
- Compose execute with start=None

### Types of changes

### DIFF
--- a/monai/transforms/compose.py
+++ b/monai/transforms/compose.py
@@ -262,8 +262,8 @@ class Compose(Randomizable, InvertibleTransform):
                 )
 
     def get_index_of_first(self, predicate):
-        for i in range(len(self.transforms)):
-            if predicate(self.transforms[i]):
+        for i, xform in enumerate(self.transforms):
+            if predicate(xform):
                 return i
         return None
 
--- a/monai/transforms/compose.py
+++ b/monai/transforms/compose.py
@@ -294,11 +294,11 @@ class Compose(Randomizable, InvertibleTransform):
         transforms: Sequence[Any],
         map_items: bool = True,
         unpack_items: bool = False,
-        start: int = 0,
+        start: int | None = 0,
         end: int | None = None,
         lazy_evaluation: bool = False,
-        overrides: dict = None,
-        override_keys: tuple = None,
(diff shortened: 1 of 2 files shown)


ericspod:
@atbenmurray is planning on making some changes to a number of aspects this PR touches so perhaps we wait on this one a bit.

atbenmurray:
Still need a pause on this one for the moment

monai/transforms/compose.py
```diff
@@ -309,27 +309,48 @@ def execute(
...
+                of transforms that have the ``RandomizedTrait`` interface.
+            log_stats: whether to log detailed information about the application of ``transforms``
+                to ``input_``. For NumPy ndarrays and PyTorch tensors, log only the data shape and
+                value range. Defaults to False if not set.
+            verbose: whether to print debugging info when lazy_evaluation=True.
         """
         end_ = len(transforms) if end is None else end
         if start is None:
-            raise ValueError(f"'start' ({start}) cannot be None")
+            input_ = evaluate_with_overrides(
```
atbenmurray, line 345: I am not sure this is the correct way to handle this issue. I know you are trying to isolate dataset from changes, but I don't understand why evaluate_with_overrides needs to be called

atbenmurray: In any case, please hold off on this for now until I have assembled the PR that refactors some of this code for simplicity.

wyli: mainly to replicate this line on the dev branch: https://github.com/Project-MONAI/MONAI/blob/be3d13869d9e0060d17a794d97d528d4e4dcc1fc/monai/data/dataset.py#L355 
so that the final transform always gets the pending operations evaluated

atbenmurray: If ```first_random``` is None, that means your entire compose was run before the item was cached, which would include the final evaluation. When the item is loaded from the cache, it has no pending transforms, so you have no way for pending transforms to be added that would then need to be evaluated

atbenmurray:
As per the review on #6244, if we agree that we aren't supporting caching of pending transforms, we can close this PR without a merge.

Issue #6189, opened 2023-03-20
Optimize log printing to be more informative in showing the progress 


heyufan1995:

mingxin-zheng:
Reopen this as the request is not fully completed yet.

wyli:
addressed by algo template updates

PR #6243, opened 2023-03-27
[WIP] Add RankFilter to skip logging when the rank is not meeting criteria

### LINKED ISSUE #6230: optionally skip logging when rank >0
it's useful to remove the logging handlers when rank > 0, example code:

```py
import logging
import torch.distributed as dist
from monai.apps.auto3dseg.auto_runner import logger


def mute_logger(_logger):
    if _logger.hasHandlers():
        for h in list(_logger.handlers):
            _logger.removeHandler(h)
    _logger.addHandler(logging.NullHandler())


logger.warning("logging start")  # example of logging (all ranks)
if dist.is_available() and dist.is_initialized() and dist.get_rank() > 0:
    mute_logger(logger)
logger.warning("logging end")  # example of logging (only rank 0)
```

would be great to make it a generic utility in monai:
- adds the capability of 'unmute'
- adds the support of modifying any monai logger -- `logging.getLogger("name")`
- when starting a logger, provides an option of rank0 only?

mingxin-zheng:
Fixes #6230 and a part of #6189 

### Description

The RankFilter class is a convenient filter that extends the Filter class in the Python logging module.
The purpose is to control which log records are processed based on the rank in a distributed environment.

### Types of changes

### DIFF
--- a/monai/utils/dist.py
+++ b/monai/utils/dist.py
@@ -12,6 +12,8 @@
 from __future__ import annotations
 
 import sys
+from collections.abc import Callable
+from logging import Filter
 
 if sys.version_info >= (3, 8):
     from typing import Literal
@@ -26,7 +28,7 @@ from monai.utils.module import min_version, optional_import
 
 idist, has_ignite = optional_import("ignite", IgniteInfo.OPT_IMPORT_VERSION, min_version, "distributed")
 
-__all__ = ["get_dist_device", "evenly_divisible_all_gather", "string_list_all_gather"]
+__all__ = ["get_dist_device", "evenly_divisible_all_gather", "string_list_all_gather", "RankFilter"]
 
 
 def get_dist_device():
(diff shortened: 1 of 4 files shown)


monai/utils/dist.py
```diff
@@ -174,3 +176,29 @@ def string_list_all_gather(strings: list[str], delimiter: str = "\t") -> list[st
...
+            but the user can define their own function to implement custom filtering logic.
+    """
+
+    def __init__(self, filter_fn: Callable = lambda rank: rank == 0):
+        super().__init__()
+        self.filter_fn: Callable = filter_fn
+        if dist.is_available() and dist.is_initialized():
+            self.rank: int = dist.get_rank()
+        else:
+            raise ValueError("The torch.distributed is either unavailable and uninitiated.")
```
wyli, line 198: this may be too restrictive... for non-distributed environment

mingxin-zheng: Fixed on b35230e. Thanks!

mingxin-zheng: Added it in a452056

monai/utils/dist.py
```diff
@@ -174,3 +177,31 @@ def string_list_all_gather(strings: list[str], delimiter: str = "\t") -> list[st
...
+        filter_fn: an optional lambda function used as the filtering criteria.
+            The default function logs only if the rank of the process is 0,
+            but the user can define their own function to implement custom filtering logic.
+    """
+
+    def __init__(self, rank: int | None = None, filter_fn: Callable = lambda rank: rank == 0):
+        super().__init__()
+        self.filter_fn: Callable = filter_fn
+        if dist.is_available() and dist.is_initialized():
+            self.rank: int = rank if rank else dist.get_rank()
```
wyli, line 198:
```suggestion
            self.rank: int = rank if rank is not None else dist.get_rank()
```

Issue #6249, opened 2023-03-27
SSIM returning wrong values and limited to uniform kernel only


Warvito:
**Describe the bug**
The SSIMMetric is reshaping the values with `view(ssim_value.shape[1], -1)` instead `view(ssim_value.shape[0], -1)` returning a vector of (channels, 1) shape instead (batch, 1).

https://github.com/Project-MONAI/MONAI/blob/8eceabf281ab31ea4bda0ab8a6d2c8da06027e82/monai/metrics/regression.py#L385

Besides that, in most packages we compute the SSIM using an gaussian kernel, instead a uniform kernel. The option to be able to choose the type of kernel should be added.

Finally, in most package, there is no cov_norm parameter, producing a significantly different value when compared to the [torchmetrics](https://torchmetrics.readthedocs.io/en/stable/) package with an uniform kernel, and compared to torchmetrics, [piq](https://github.com/photosynthesis-team/piq) and[ pytorch-msssim](https://github.com/VainF/pytorch-msssim) when using an gaussian kernel

PR #6250, opened 2023-03-27
Add new version of the SSIM metric

Warvito:
fixes https://github.com/Project-MONAI/MONAI/issues/6249

### Types of changes


monai/losses/ssim_loss.py
```diff
@@ -32,29 +32,67 @@ class SSIMLoss(_Loss):
...
-    def __init__(self, win_size: int = 7, k1: float = 0.01, k2: float = 0.03, spatial_dims: int = 2):
+    def __init__(
+        self,
+        spatial_dims: int,
+        data_range: float = 1.0,
+        kernel_type: KernelType | str = KernelType.GAUSSIAN,
+        kernel_size: int | Sequence[int] = 11,
+        kernel_sigma: float | Sequence[float] = 1.5,
+        k1: float = 0.01,
+        k2: float = 0.03,
```
wyli, line 43: I think we can match the API with the existing version and make it backwards compatible here

Warvito: I replaced `kernel_size` back to `win_size`

wyli: thanks, this will still probably change the training pipeline defined here https://github.com/Project-MONAI/tutorials/blob/main/reconstruction/MRI_reconstruction/varnet_demo/train.py#L141  and we don't currently have automated tests to avoid regression. have you tested this code change in an end-to-end pipeline?

Warvito: I only evaluated it as a metric (comparing its results agains other packages listed in #6249 ). The results were consistent with the other metrics in both gaussian and uniform kernel. However, it is slight different from the previous implementation based on Facebook code (https://github.com/facebookresearch/fastMRI/blob/main/banding_removal/fastmri/ssim_loss_mixin.py). The main difference is the `conv_norm` component

https://github.com/Project-MONAI/MONAI/blob/aa326bdf069c076bd8a02ca2c799f5c6e74c2981/monai/metrics/regression.py#L280

This is not used in other implementations and I could not find reference about it. 
I will try re-run the training pipeline, to see if the pattern changes

Warvito: I just saw that the dataset is +1TB, it might take some time to download it yet

wyli: sure, alternatively we can add some tests like this https://github.com/Project-MONAI/MONAI/blob/57f05bafa09e7e9e6d4e9c66be6b6f1f0e01077f/tests/test_seg_loss_integration.py. It can still be included in v1.2 (code freeze ~13th/April) if the tests are included or the end-to-end example work fine.

jak0bw: @wyli I wanted to add to the discussion, that it's probably not necessary to worry to much about regression/changed behavior for SSIMLoss and SSIMMetric as MONAI 1.1 only works with channel_size=1 and batch_size<=2. The commit fixing batch_size > 2 (ee38882)  was only merged end of January but is not in MONAI 1.1.

I therefore do not think there is a  heavy reliance on the current implementation. This also means there is a heavy need for the new, in my opinion much more functional implementation and it should be merged as soon as possible.

Just in case there is no absolute code freeze for MONAI 1.2 already, otherwise there is of course a lot of time left.

wyli: Ok, good point, it'll be very unlikely to make it worse than 1.1, let me try to merge it now.

wyli (approved):

PR #6251, opened 2023-03-27
SlidingWindowInfererAdapt

### LINKED ISSUE #6340: lazy.utils.resample output dtype
**Describe the bug**
the output of grid resampling is always float32

https://github.com/Project-MONAI/MONAI/blob/6a7f35b7e271cdf3b2c763ae35acf4ab25d85d23/monai/transforms/lazy/utils.py#L226-L227

https://github.com/Project-MONAI/MONAI/blob/6a7f35b7e271cdf3b2c763ae35acf4ab25d85d23/monai/transforms/spatial/functional.py#L182

the crop pad 'resampling' should also output float32

https://github.com/Project-MONAI/MONAI/blob/6a7f35b7e271cdf3b2c763ae35acf4ab25d85d23/monai/transforms/lazy/utils.py#L220-L222

otherwise collating random transforms of different data types can run into exceptions.

cc @dongyang0122

myron:
SlidingWindowInfererAdapt extends SlidingWindowInferer to automatically switch to buffered and then to CPU stitching, when OOM on GPU. It also records a size of such large images to automatically try CPU stitching for the next large image of a similar size.  If the stitching 'device' input parameter is provided,
    automatic adaptation won't be attempted, please keep the default option device = None for adaptive behavior.
    Note: the output might be on CPU (even if the input was on GPU), if the GPU memory was not sufficient.

---
also fixes #6340 by adding one line to the resampling

### DIFF
--- a/monai/inferers/inferer.py
+++ b/monai/inferers/inferer.py
@@ -531,7 +531,7 @@ class SlidingWindowInfererAdapt(SlidingWindowInferer):
 
                     if gpu_stitching:  # if failed on gpu
                         gpu_stitching = False
-                        self.cpu_thresh = inputs.shape[2:].numel()  # update thresh
+                        self.cpu_thresh = inputs.shape[2:].numel() - 1  # update thresh
 
                         if not skip_buffer:
                             buffered_stitching = True
--- a/monai/inferers/inferer.py
+++ b/monai/inferers/inferer.py
@@ -448,7 +448,9 @@ class SlidingWindowInferer(Inferer):
 
         """
 
-        device = self.device
+        device = kwargs.pop("device", self.device)
+        buffer_steps = kwargs.pop("buffer_steps", self.buffer_steps)
+
         if device is None and self.cpu_thresh is not None and inputs.shape[2:].numel() > self.cpu_thresh:
             device = "cpu"  # stitch in cpu memory if image is too large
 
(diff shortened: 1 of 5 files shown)


myron:
I'm not sure why it fails premerge/flake test. It's fine for me.  can someone help plz @mingxin-zheng @wyli

mingxin-zheng:
Hi @myron , as I replied in the slack channel, this could be caused by using a different version of `isort`.

I ran `./runtests --autofix` and it shows the line 623 and 630 should be removed.

monai/inferers/inferer.py
```diff
@@ -588,3 +588,56 @@ def network_wrapper(
             return out
 
         return tuple(out_i.unsqueeze(dim=self.spatial_dim + 2) for out_i in out)
+
+
+class SlidingWindowInfererAdapt(SlidingWindowInferer):
```
wyli, line 593: this could be useful for all the `Inferer` subclasses, how about:
if you agree I can update the PR to include this

```py
class AutoDeviceInferer(Inferer):
    def __init__(self, inferer=None):
        self.inferer = inferer
        self.cpu_thresh = getattr(self.inferer, "cpu_thresh", None)

    def __call__(
        self,
        inputs: torch.Tensor,
        network: Callable[..., torch.Tensor | Sequence[torch.Tensor] | dict[Any, torch.Tensor]],
        *args: Any,
        **kwargs: Any,
    ) -> torch.Tensor | tuple[torch.Tensor, ...] | dict[Any, torch.Tensor]:
        try:
            return self.inferer(inputs=inputs, network=network, *args, **kwargs)
        except RuntimeError as e:
            if (
                inputs.is_cuda
                and not torch.device(self.inferer.device).type == "cuda"
                and not (self.cpu_thresh is not None and inputs.shape[2:].numel() > self.cpu_thresh)
                and "OutOfMemoryError" in str(type(e).__name__)
            ):
                warnings.warn(f"GPU stitching failed {e}, attempting CPU stitching, image dim {inputs.shape}")
                self.cpu_thresh = inputs.shape[2:].numel() - 1  # remember large image size
                # CPU stitching (trigger cpu_thresh)
                return self.inferer(inputs=inputs, network=network, *args, **kwargs)
            raise e
```

wyli: I think this logic could be moved to this class as well:
https://github.com/Project-MONAI/MONAI/blob/0ce91ecc4d40e16773f1c73554be01911c0832b9/monai/inferers/inferer.py#L439-L441

myron: Okay, feel free to update the PR if you think the wrapper class is better. 

1) "self.cpu_thresh = ..."  may fail for cases when that field doesn't exist  ?!
2) perhaps it doesn't need to be subclassed from (Inferer)
3)  "and torch.device(self.inferer.device).type == "cuda""
we don't need this check, for one device could be None and it may crash (not sure).  But also  "cpu_thresh" is only used when device is None

wyli: sure, I've updated the PR, please have a look, if you agree with this change, I'll add some tests and finalise it.

mingxin-zheng: Is it useful to have an option to put the output image back to gpu？

wyli: good point... at the moment if the module has the `device` property, this class sets `mod.device='cpu'`, hopefully `mod` will make sure the output is on `cpu` (that's usually what torch.nn.modules do).

monai/networks/utils.py
```diff
@@ -837,3 +838,67 @@ def replace_modules_temp(
...
+                    warnings.warn(
+                        f"self.mod.cpu_thresh is set to {self.mod.cpu_thresh}, "
+                        f" and input size: {inputs.shape[2:].numel()}"
+                    )
+                raise e  # self.mod currently runs on cpu according to self.mod.cpu_thresh
+            warnings.warn(
+                f"{type(self.mod)} failed {e} device={self.mod.device}, "
+                f" attempting CPU, image dim {inputs.shape}. The error handling is expensive, "
+                f" please review {type(self.mod)} settings if this message appears frequently."
+            )
```
myron, line 890: I think this will crash if self.mod does not have "device" property

monai/networks/utils.py
```diff
@@ -837,3 +838,67 @@ def replace_modules_temp(
...
+            if not self.has_cpu_thresh:  # mod doesn't support dynamic input size based device allocation
+                if isinstance(inputs, torch.Tensor):
+                    inputs = inputs.to("cpu")
```
myron, line 893: I don't think we should do it,  if "model" weights are on GPU, it won't accept inputs on CPU, and will crash, so then we need to worry where the model weights are.

monai/inferers/inferer.py
myron: I've made a few inline code comments, but mainly I have a mixed feeling about a few things

1) we make this wrapper general for any self.mod, but mainly it's designed for Inference (which is our actual usecase). 
I've looked at our current inferers inside https://github.com/Project-MONAI/MONAI/blob/dev/monai/inferers/inferer.py

and the only classes that have "self.device" property are: SlidingWindowInferer and SliceInferer (because it's a subclass of   SlidingWindowInferer).  So at the moment this wrapper class will only work for those 2 cases, because other classes have no "device property" and no "cpu_thresh". 

so may be, it's easier if just have "self.mod: SlidingWindowInferer = None" to always be of SlidingWindowInferer type, then we don't need hasattr checks, and it'll look simpler. 

PS: I understand that we want to support some arbitrary hypothetical scenarios of self.mod, but so far we don't have such usecases, and It may lead to future bugs.  I'm open to discussion here.

2) I saw the PR for buffered mode support https://github.com/Project-MONAI/MONAI/pull/6254
when that PR is merged, then we'll have an additional stitch_mode (it won't be only device==cpu or device==gpu), so we'll need to do 3 steps (or more steps) here: 1) on GPU stitching 2) on GPU-CPU hybrid larger buffer,, 3) on GPU-CPU hybrid for smaller buffer 4)finally on CPU

given that using self.device='cpu'; self.mod(); self.device='original' will be complicated.   So may be it's better if we wait for PR  https://github.com/Project-MONAI/MONAI/pull/6254 to be merged , and then  finish this PR

I'm guessing we need to add an enum field to SlidingWindowInferer (stitch_mode:str=['gpu', 'cpu', 'buffer', ...]
this will be better then using "device" property

3) minor comment: "class AutoDeviceAdapt' name is very vague. May something like "AutoSlidingWindowInferenceAdapt" or something more explicit.

wyli: > and the only classes that have "self.device" property are: SlidingWindowInferer and SliceInferer. So at the moment this wrapper class will only work for those 2 cases, because other classes have no "device property"

sure, how about we try to call `.to(device)` if the module supports this? e.g.:
```py
>>> net = UNet(2, 3, 3, (1,2),(1,1))
>>> net.to(torch.device("cpu"))
```
otherwise I think we can revert to your previous solution as there's no use case for a generic implementation.

myron: do you mean call net.to('cpu') inside of slidingwindowinferer somewhere? I don't think we should do it, it's better for a caller/user to manage it.

And I vote to go back to sub-classing SlidingWindowInferer logic, but it's up to you.   And the merged PR https://github.com/Project-MONAI/MONAI/pull/6254 looks great, and is working, so that logic needs to be added here to: with something like: a) try GPU b) try buffered (buffer_size=large) c) try buffered (buffer_size//2), until buffer size is 1 d) then finally try CPU.   And keep track of cpu_thresh and buffer_size  (not to repeat it again). 

it might be useful to also update the main class SlidingWindowInferer _call_() to accept a "stitch_mode" enum parameter, so that we can call it with that parameter. 

let me know if you want to take on it, or if you need help

thank you

wyli: thanks, sounds good, I've reverted and please take over from here, I can help address minor issues before merging

myron:
@wyli I've updated this PR to include buffered case handling, please check and merge

also mypy complained on the super()._call__() lines with error:
 ```
mypy
mypy 1.2.0 (compiled: yes)
monai/inferers/inferer.py:509:20: error: "__call__" of "SlidingWindowInferer" gets multiple values for keyword argument "inputs"  [misc]
monai/inferers/inferer.py:509:20: error: "__call__" of "SlidingWindowInferer" gets multiple values for keyword argument "network"  [misc]
monai/inferers/inferer.py:519:23: error: "__call__" of "SlidingWindowInferer" gets multiple values for keyword argument "inputs"  [misc]
monai/inferers/inferer.py:519:23: error: "__call__" of "SlidingWindowInferer" gets multiple values for keyword argument "network"  [misc]
```
even though, the code looks correct.
I added "type:ignore" , because I couldn't figure it out, may be there is better way.

**Review by wyli — APPROVED** [review]
thanks, I'll test and merge it soon

File: monai/inferers/inferer.py [hunk 4]
```diff
@@ -467,13 +480,96 @@ def __call__(
...
+        if self.device is not None:
+            return super().__call__(inputs, network, *args, **kwargs)
+
+        skip_buffer = self.buffer_steps is not None and self.buffer_steps <= 0
+        cpu_cond = self.cpu_thresh is not None and inputs.shape[2:].numel() > self.cpu_thresh
+        gpu_stitching = inputs.is_cuda and not cpu_cond
+        buffered_stitching = inputs.is_cuda and cpu_cond and not skip_buffer
+        buffer_steps = max(1, self.buffer_steps) if self.buffer_steps is not None else 1
+
+        for _ in range(10):  # at most 10 trials
```
wyli, line 528: I was not sure about the `while True` loop, so changed it to be sure at most 10 iterations

Issue #5471, opened 2022-11-04
config component shouldn't allow duplicated arguments


wyli:
**Describe the bug**
```json
"dataset": {
    "_target_": "CacheDataset",
    "num_workers": 0,
    "num_workers": 4
},
```

in this configuration example, the duplicated `num_workers` is currently quietly allowed, which is error-prone in my understanding

cc @Nic-Ma

PR #6252, opened 2023-03-28
warn or raise ValueError on duplicated key in json/yaml config

kretes:
Fixes https://github.com/Project-MONAI/MONAI/issues/5471 .

### Description

[WIP]
Monai will now raise an error whenever there is a duplicated key in any of the config json files.

TBD:
- support similiar thing when reading from yaml

Things to discuss:
- do we want to raise an Error or do we just want a warning message?
- This will raise with the first duplicate encountered, not with a full list, but I think that's a fair & simple approach.

### Types of changes

- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`. (I couldn't run them locally due to OOM)

~~Documentation updated, tested `make html` command in the `docs/` folder.~~ I think that's a safeguard that doesn't require explicit documentation

### DIFF
--- a/monai/bundle/config_parser.py
+++ b/monai/bundle/config_parser.py
@@ -12,9 +12,7 @@
 from __future__ import annotations
 
 import json
-import os
 import re
-import warnings
 from collections.abc import Sequence
 from copy import deepcopy
 from pathlib import Path
@@ -25,7 +23,7 @@ from monai.bundle.reference_resolver import ReferenceResolver
 from monai.bundle.utils import ID_REF_KEY, ID_SEP_KEY, MACRO_KEY
 from monai.config import PathLike
 from monai.utils import ensure_tuple, look_up_option, optional_import
-from monai.utils.misc import check_key_duplicates
+from monai.utils.misc import check_key_duplicates, CheckKeyDuplicatesYamlLoader
 
 if TYPE_CHECKING:
     import yaml
@@ -405,7 +403,7 @@ class ConfigParser:
             if _filepath.lower().endswith(cls.suffixes[0]):
                 return json.load(f, object_pairs_hook=check_key_duplicates, **kwargs)  # type: ignore[no-any-return]
             if _filepath.lower().endswith(cls.suffixes[1:]):
-                return yaml.safe_load(f, **kwargs)  # type: ignore[no-any-return]
+                return yaml.load(f, CheckKeyDuplicatesYamlLoader, **kwargs)  # type: ignore[no-any-return]
             raise ValueError(f"only support JSON or YAML config file so far, got name {_filepath}.")
 
     @classmethod
--- a/monai/bundle/config_parser.py
+++ b/monai/bundle/config_parser.py
@@ -16,7 +16,7 @@ import re
 from collections.abc import Sequence
 from copy import deepcopy
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Tuple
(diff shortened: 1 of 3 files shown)


wyli:
Thanks, I think we can make it a warning be default, and provide an option to raise an exception. Cc @Nic-Ma

kretes:
I've made it a warning by default and error only when a specific env variable is set. LMK if this is ok and then I can extend it to yaml.

monai/bundle/config_parser.py
```diff
@@ -383,6 +387,29 @@ def _do_parse(self, config: Any, id: str = "") -> None:
         else:
             self.ref_resolver.add_item(ConfigItem(config=item_conf, id=id))
 
+    @staticmethod
+    def check_key_duplicates(ordered_pairs: Sequence[tuple[Any, Any]]) -> dict[Any, Any]:
```
Nic-Ma, line 391: @wyli Should we put this function in the `utils`? I feel it's not part of `ConfigParser` signature.

Thanks.

tests/test_config_parser.py
```diff
@@ -303,6 +307,33 @@ def test_substring_reference(self, config, expected):
         parser = ConfigParser(config=config)
         self.assertEqual(parser.get_parsed_content("total"), expected)
 
+    @parameterized.expand([TEST_CASE_DUPLICATED_KEY])
+    @mock.patch.dict(os.environ, {"MONAI_FAIL_ON_DUPLICATE_CONFIG": "1"})
+    def test_parse_json_raise(self, config_string, _, __):
```
Nic-Ma, line 312: Can we combine these 2 test functions `test_parse_json_raise` and `test_parse_json_warn`?
Seems most of the logic are similar.

Thanks.

Nic-Ma (approved):
Looks good to me, put some minor comments inline.

Thanks.

monai/bundle/config_parser.py
```diff
@@ -383,6 +387,29 @@ def _do_parse(self, config: Any, id: str = "") -> None:
...
+        Args:
+            ordered_pairs: sequence of (key, value)
+        """
+        keys = set()
+        for k, _ in ordered_pairs:
+            if k in keys:
+                if os.environ.get("MONAI_FAIL_ON_DUPLICATE_CONFIG", "0") == "1":
+                    raise ValueError(f"Duplicate key: `{k}`")
+                else:
+                    logger.warning(f"Duplicate key: `{k}`")
```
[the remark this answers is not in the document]
kretes: done

monai/bundle/config_parser.py
kretes: done

tests/test_config_parser.py
kretes: I think it's gonna reduce the readability a lot. 
1. We use `@mock.patch.dict(os.environ, {"MONAI_FAIL_ON_DUPLICATE_CONFIG": "1"})` which as an annotation nicely handles changing and restoring os.environ
2. The two scenarios differ by:
- the context manager in which the tested method is applied in
- how the expected reaction is verified
- additional assertions

Introducing all this as parameters would make the main test method abstract and diluted.

At the same time - I've introduced this functionality for yaml without introducing new test methods, since this nicely fits as a parameter

Nic-Ma: OK, that makes sense. thanks.

PR #6254, opened 2023-03-29
6193 add a buffer_step option to the sliding windows util/speed up

wyli:
Fixes https://github.com/Project-MONAI/MONAI/issues/6193

### Description

- adds `buffer_step` (code adapted from @myron's idea/implementation)
- clean up multioutput resizing logic
- perf optimize 
- `overlap` to support different values along the spatial dimensions

### Types of changes


myron:
I'm going to test it

myron:
@wyli this looks great! thanks for writing.

a few comments
1) perhaps the default for buffer_dim should be -1 (as this is the most common)
2) SlidingWindowInferer parameter type for "overlap" should be updated to overlap: float | Sequence (since it accepts a list now)
3) setting buffer_step option seems only appropriate together with setting device='cpu". It wasn't obvious. I tried using buffer_step=1, and GPU memory increased (compared to buffer_step==None).  and it took me some time to figure out that device parameter needs to be manually specified to cpu.  a user usually leaves that option as a default None, so it won't be obvious. may be we can give a warning, or just automatically set device to 'cpu' when the buffer_step is not None. 

thanks!

wyli:
thanks @myron, I'll revise it

PR #6257, opened 2023-03-30
Lazy eval refactor

atbenmurray:
# Lazy evaluation refactor

This PR contains a reworking of the lazy evaluation mechanism on dev to provide the following:

1. There are now three lazy execution options:
    1. the `options=None` (default) mode guarantees that all pending operations are evaluated before evaluating non-lazy _apply_transforms
       or transforms with lazy set to False. *This means that all pipelines should be able to execute lazily without breaking (the first thing that users will try)*
    1. there is an `options={'reorder': 'lazy_last'}` mode that moves all lazy operations past non-lazy ones, (bounded by use of 
       ApplyPending/ApplyPendingd) and guarantees invertibility
    1. there is an `options={'reorder': 'lazy_last_nosync'}` mode that relies on ApplyPending/ApplyPendingd to cause application of
       pending transforms, and precisely mimics the behaviour of the lazy execution on dev.
1. Lazy now has three states on Compose:
    1. `lazy=True` overrides the lazy flag on lazy transforms, making them all execute lazily
    1. `lazy=None` uses the lazy flags set on transforms, and can execute them lazily or non-lazily depending
    1. `lazy=False` (default) causes lazy processing to be disabled and all lazy transforms will execute non-lazily
1. Lazy transforms have a lazy override that they can optionally set during `__call__`. This means that lazy behaviour can be overridden without mutating the transform instance
1. Logging has been fixed so that the first transform is included in the logging messages. Logging activities are clearer and contain addition important information about lazy modes
1. `ApplyPending` and `ApplyPendingd` replace the use of `Identity` and `Identityd` for causing pending ops to be evaluated when using the `lazy_last[_nosync]` options
1. Compose API changes:
    1. `overrides` and `override_keys` have been consolidated into overrides
        1. on dev: `override_keys=('a', 'b'), overrides={'mode': ('bilinear', None), 'padding_mode: (None, 'zeros')'}`
        1. on PR: `overrides={'a': {'mode': 'bilinear'}, 'b': {'padding_mode': 'zeros'}}`
    1. `log_stats` and `verbose` have been replaced with `logger_name`, consistent with other parts of MONAI
1. additional lazy resampling documentation has / is being added
1. additional lazy tests have been added

### Types of changes

 * Breaking changes:
   * `lazy=True/None` and `options=None` is new functionality relative to `dev`
   * `lazy=True/None` and `options={"reorder": "lazy_last"}` is new functionality relative to `dev`
   * `lazy=None` and setting `lazy=True` on a transform is new functionality relative to `dev`
   * `Identity[d]` doesn't cause pending operations to be evaluated; they must be replaced with `ApplyPending[d]`
 * Non-breaking changes:
   * `lazy=True` and `options={"reorder": "lazy_last_nosync"} (with `Identity[d]` being replaced by `ApplyPending[d]`) should be identical in behaviour to `dev`

- [wip] New tests added to cover the changes.
- [wip] Documentation updated, tested `make html` command in the `docs/` folder.


wyli:
hi @atbenmurray sorry, earlier I didn't notice this PR is pending review. please address the merging conflicts I'll test and merge it. `ApplyPending` is better than implicit resampling calls in compose, and allows for flexible overriding.

atbenmurray:
Gone through the merge from dev. I still need to test things before we merge!

atbenmurray:
@wyli, @Nic-Ma, @ericspod I think that all of the tests should now be passing on this. I'm updating docs and adding a few additional unit-tests but I would say that, with a final review, this should be merge-ready once I've resolved conflicts

atbenmurray:
@wyli, @ericspod I think this is about as good as it gets on passing tests, yes? Please can we try to get this in tomorrow?

monai/transforms/compose.py
```diff
@@ -148,26 +72,22 @@ def execute_compose(
...
             {``"mode"``, ``"padding_mode"``, ``"dtype"``, ``"align_corners"``, ``"resample_mode"``, ``device``},
             please see also :py:func:`monai.transforms.lazy.apply_transforms` for more details.
         override_keys: this optional parameter specifies the keys to which ``overrides`` are to be applied. If
             ``overrides`` is set, ``override_keys`` must also be set.
         threading: whether executing is happening in a threaded environment. If set, copies are made
             of transforms that have the ``RandomizedTrait`` interface.
-        log_stats: whether to log the detailed information of data and applied transform when error happened,
-            for NumPy array and PyTorch Tensor, log the data shape and value range,
-            for other metadata, log the values directly. default to `False`.
-        verbose: whether to print debugging info when lazy_evaluation=True.
```
wyli, line 170: why remove these two flags?

atbenmurray: They didn't match with the rest of the API. @ericspod and I were discussing it and he suggested that these parameters should be replaced with 'logger_name', which allows you to override the default logger, as this is how it is done in the rest of the code, then everything goes to the logger with either level 'info' or 'error', depending on what bits are being logged. Looks like I have forgotten to put it in so I need to do it still

monai/transforms/compose.py
```diff
@@ -148,26 +72,22 @@ def execute_compose(
         start: the index of the first transform to be executed. If not set, this defaults to 0
         end: the index after the last transform to be exectued. If set, the transform at index-1
             is the last transform that is executed. If this is not set, it defaults to len(transforms)
-        lazy_evaluation: whether to enable lazy evaluation for lazy transforms. If False, transforms will be
+        lazy: whether to enable lazy evaluation for lazy transforms. If False, transforms will be
             carried out on a transform by transform basis. If True, all lazy transforms will
             be executed by accumulating changes and resampling as few times as possible.
             A `monai.transforms.Identity[D]` transform in the pipeline will trigger the evaluation of
             the pending operations and make the primary data up-to-date.
         overrides: this optional parameter allows you to specify a dictionary of parameters that should be overridden
```
KumoLiu, line 80: Do we need to specify here we could accept a dict or a nested dict here?

monai/utils/enums.py
```diff
@@ -647,6 +648,37 @@ class LazyAttr(StrEnum):
...
+    'ON' indicates that all transforms capable of being executed lazily will be executed lazily
+    See: :py:class: monai.transforms.compose.Compose for more details.
+    """
+
+    OFF = "off"
+    ENABLED = "enabled"
+    ON = "on"
+
+    @staticmethod
+    def as_bool(lazy_mode):
```
wyli, line 667: this `as_bool` doesn't actually return a boolean value, not sure what the intention is. maybe override `self.__bool__` instead of creating this method.

monai/apps/detection/transforms/dictionary.py
```diff
@@ -1307,6 +1308,7 @@ class RandRotateBox90d(RandRotate90d):
         spatial_axes: 2 int numbers, defines the plane to rotate with 2 spatial axes.
             Default: (0, 1), this is the first two axis in spatial dimensions.
         allow_missing_keys: don't raise exception if key is missing.
+        lazy:
```
wyli, line 1311: looks like this is work in progress?

monai/transforms/compose.py
```diff
@@ -308,54 +238,39 @@ class Compose(Randomizable, InvertibleTransform):
...
     """
 
     def __init__(
         self,
         transforms: Sequence[Callable] | Callable | None = None,
         map_items: bool = True,
         unpack_items: bool = False,
-        log_stats: bool = False,
-        lazy_evaluation: bool | None = None,
+        lazy: str | bool | None = LazyMode.OFF,
```
wyli, line 261: logically this `Compose.lazy` which as type `str|bool|None` is different from `LazyTransform.lazy` (type `bool|None`) https://github.com/Project-MONAI/MONAI/blob/b16a87fce2ecf51a10efe28ca9eb9bfbf2aa89ba/monai/transforms/transform.py#L281-L284

but they share the same property name `lazy`,  I found it's confusing.

monai/transforms/transform.py
```diff
@@ -254,17 +278,22 @@ class LazyTransform(Transform, LazyTrait):
     dictionary transforms to simplify implementation of new lazy transforms.
     """
 
-    _lazy_evaluation: bool = False
+    def __init__(self, lazy: bool | None = False):
```
wyli, line 281: this constructor is introduced by this PR, all the subclasses should have `super().__init__` calls to ensure this is executed during initialisation.

monai/transforms/compose.py
atbenmurray: This is a consequence of nesting Composes. Compose instances can force non-lazy execution, lazy execution or can leave the choice to the transforms that it contains. bool | none makes sense for containers like Compose, but not for transforms. However, lazy still means lazy. It is just that the semantics are one step more complicated for Compose due to its container nature

monai/apps/detection/transforms/dictionary.py
```diff
@@ -1326,7 +1328,11 @@ def __init__(
         super().__init__(self.image_keys + self.box_keys, prob, max_k, spatial_axes, allow_missing_keys)
         self.box_ref_image_keys = ensure_tuple_rep(box_ref_image_keys, len(self.box_keys))
 
-    def __call__(self, data: Mapping[Hashable, torch.Tensor]) -> Mapping[Hashable, torch.Tensor]:
+    def __call__(
+        self, data: Mapping[Hashable, torch.Tensor], lazy: bool | None = None
+    ) -> Mapping[Hashable, torch.Tensor]:
+        if lazy is True:
```
wyli, line 1334:
```suggestion
        if lazy:
```

monai/apps/reconstruction/transforms/dictionary.py
```diff
@@ -229,6 +230,9 @@ def __call__(self, data: Mapping[Hashable, Tensor]) -> dict[Hashable, Tensor]:
         Returns:
             the new data dictionary
         """
+        if lazy is True:
```
wyli, line 233:
```suggestion
        if lazy:
```

monai/utils/enums.py
atbenmurray: To be honest, I thought there would be more value to having LazyMode as a enum than there turns out to be. I think we could drop it and stay with bool | None. What do you think?

atbenmurray: It is to be able to map between lazy modes and an optional bool. As such __bool__ won't work, so I did as_bool instead

monai/transforms/compose.py
wyli: I don't think it's the same, `Compose` doesn't inherent `LazyTransform` and the outcome of `Compose` always has a `execute_pending_transforms`, which is different from the other subclasses of `LazyTransform`

monai/apps/detection/transforms/dictionary.py
atbenmurray: No just a docstring tidy up required. Initially I was adding in the lazy parameter but decided it would be fine without it, but didn't remove the docstring entry

monai/apps/detection/transforms/dictionary.py
atbenmurray: I prefer `lazy is True` because lazy could be passed some arbitrary value by the user such as '123'

monai/utils/enums.py
wyli: I feel the possible values of input argument `lazy` could be `True` `False` `None` `LazyMode.OFF`, `LazyMode.ON`, `LazyMode.ENABLED`, it's not easy to use

monai/apps/detection/transforms/dictionary.py
wyli: and in that case, there's no warning at all, I don't think that's the expected behaviour of this input check.

monai/apps/reconstruction/transforms/dictionary.py
atbenmurray: Same thing:

```
foo = 123
if foo:
  print("foo")
if foo is True:
  print("foo")
else:
  print("bar")
```
This is why I try to avoid using if x: unless I want those semantics

monai/utils/enums.py
atbenmurray: So, just True, False, None then?

monai/apps/detection/transforms/dictionary.py
atbenmurray: The difference is that if you accidentally pass a bad type to lazy, it is going to perform lazy evaluation. With `lazy is True`, you are only going to perform lazy evaluation if you pass it True. Given that this is an experimental feature, that feels safer to me

monai/transforms/lazy/array.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    ApplyPending can be inserted into a pipeline that is being executed lazily in order to ensure
+    resampling happens before the next transform. If passed a ``MetaTensor`` that has pending
+    transforms, it executes those pending transforms and returns the resampled ``MetaTensor`` instance.
+
+    See ``Compose`` for a detailed explanation of the lazy resampling feature.
+    """
+
+    def __init__(self):
+        super().__init__()
```
wyli, line 32:
```suggestion


```

also what's the exepcted behaviour when working with the cache-based datasets?

monai/transforms/lazy/dictionary.py
```diff
@@ -0,0 +1,49 @@
...
+    at that key location.
+
+    See ``Compose`` for a detailed explanation of the lazy resampling feature.
+    """
+
+    def __init__(self, keys):
+        super().__init__()
+        self.keys = keys
+
+    def __call__(self, data, **kwargs):
```
wyli, line 35: all the dictionary based transform don't currently accept `kwargs` in this at this call

monai/transforms/lazy/functional.py
```diff
@@ -25,14 +25,33 @@
...
-__all__ = ["apply_transforms"]
+__all__ = ["apply_pending", "execute_pending_transforms"]
 
 __override_keywords = {"mode", "padding_mode", "dtype", "align_corners", "resample_mode", "device"}
 
 
-def apply_transforms(
-    data: torch.Tensor | MetaTensor, pending: list | None = None, overrides: dict | None = None, **kwargs: Any
-):
+def execute_pending_transforms(data, overrides: dict | None = None):
```
wyli, line 33: please add docstring, also `execute_pending_transforms` and `apply_pending` have very similar name and signature, it's unclear what to use without reading the source code

atbenmurray: Will do

atbenmurray: @wyli done

monai/transforms/compose.py
atbenmurray: @wyli LazyMode has been replaced by True/False/None

monai/transforms/compose.py
atbenmurray: @wyli are you fine with the `logger_name`? It is consistent with other parts of the API

monai/transforms/compose.py
atbenmurray: Yes, I'll create an example

monai/transforms/compose.py
wyli: currently `verbose=True` prints pending operation counts:
```
2023-04-26 10:59:48,011 - INFO - Lazy - 'image' - upcoming: 'EnsureTyped'- pending 0
2023-04-26 10:59:48,012 - INFO - Lazy - 'label' - upcoming: 'EnsureTyped'- pending 0
2023-04-26 10:59:48,709 - INFO - Lazy - 'image' - upcoming: 'EnsureChannelFirstd'- pending 0
2023-04-26 10:59:48,709 - INFO - Lazy - 'label' - upcoming: 'EnsureChannelFirstd'- pending 0
2023-04-26 10:59:48,709 - INFO - Lazy - 'image' - upcoming: 'Orientationd'- pending 0
2023-04-26 10:59:48,709 - INFO - Lazy - 'label' - upcoming: 'Orientationd'- pending 0
2023-04-26 10:59:48,712 - INFO - Lazy - 'image' - upcoming: 'Spacingd'- pending 1
2023-04-26 10:59:48,712 - INFO - Lazy - 'label' - upcoming: 'Spacingd'- pending 1
2023-04-26 10:59:48,721 - INFO - Lazy - 'image' - upcoming: 'CastToTyped'- pending 2
2023-04-26 10:59:48,721 - INFO - Lazy - 'label' - upcoming: 'CastToTyped'- pending 2
2023-04-26 10:59:48,722 - INFO - Lazy - 'image' - upcoming: 'CropForegroundd'- pending 2
2023-04-26 10:59:48,722 - INFO - Lazy - 'label' - upcoming: 'CropForegroundd'- pending 2
```
it's useful for debugging the transforms, what's the equivalent usage in this PR?

monai/transforms/compose.py
```diff
@@ -148,26 +72,22 @@ def execute_compose(
...
         overrides: this optional parameter allows you to specify a dictionary of parameters that should be overridden
             when executing a pipeline. These each parameter that is compatible with a given transform is then applied
-            to that transform before it is executed. Note that overrides are currently only applied when lazy_evaluation
-            is True. If lazy_evaluation is False they are ignored.
+            to that transform before it is executed. Note that overrides are currently only applied when lazy
+            is True. If lazy is False they are ignored.
             currently supported args are:
             {``"mode"``, ``"padding_mode"``, ``"dtype"``, ``"align_corners"``, ``"resample_mode"``, ``device``},
             please see also :py:func:`monai.transforms.lazy.apply_transforms` for more details.
         override_keys: this optional parameter specifies the keys to which ``overrides`` are to be applied. If
```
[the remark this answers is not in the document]
atbenmurray: Done

monai/utils/enums.py
wyli: sure, sounds good

monai/transforms/lazy/dictionary.py
wyli: ok... so what's the current recommended usage of `ApplyPending` and `ApplyPendingd`, we don't support overriding? I thought we'll have the flexibility of different overrides:
````
[Orientation(),  ApplyPending(overrides_kwargs_1), Flip(), Spacing(), ApplyPending(overrides_kwargs_2)] -> [RandRotate()]
```

monai/transforms/compose.py
atbenmurray: I can put something roughly equivalent in.

atbenmurray: @wyli logging of pending is back in

monai/transforms/lazy/dictionary.py
wyli: sure... mark this as resolved. (but is there any difficulty we were unaware of in implementing this? I thought it'd be simply invoking `transforms.lazy.functional.apply_pending_transforms`)

monai/transforms/compose.py
wyli: thanks, it's almost there, I think we need logs for each key such as `image` and `label`  to debug them separately...also now the `_apply_transform` in transforms/transform.py need some refactoring to decouple the concepts, 

 The current dev branch has a much cleaner version https://github.com/Project-MONAI/MONAI/blob/4f291723c4ce1a1baf7efbd499495b777dfb1869/monai/transforms/transform.py#L64-L67

monai/transforms/lazy/dictionary.py
atbenmurray: The Law of Unintended Consequences
I wouldn't want to put a feature out there that we don't understand and that bites us in the medium term. It's a design choice with potentially significant consequences.

monai/transforms/compose.py
atbenmurray: It already handles dictionary based transforms by key. The test doesn't reflect that but is a quick addendum.

You can't really separate the various different methods:
On dev we have:
`evaluate_with_overrides`, `apply_transforms`, `apply_transform` and `_apply_transform`
On this branch we have:
`apply_pending_transforms`, `apply_pending`, `apply_transform` and `_apply_transform`

In both cases, these four methods are all quite interdependent (they rely on each other for certain pre and post conditions to be the case). I've spread the functionality across them differently so that none of them has overwhelming complexity, and they do more on the branch than on dev (for example the lambda problem is fixed and we have the enabled mode (lazy = None on compose) that will execute transforms according the the flags they have set on them, allowing for more fine tuning).

atbenmurray: I specifically moved a bit of that responsibility into _apply_transforms as it reduced the need to have a separate call-stack doing the evaluate_with_overrides bit. That is where so much of the complexity in that function comes from. Moving the functionality there simplified everything hugely while also allowing for a more general solution

wyli: Disagree...I think the apply_transform logic should be improved, also using logger name to trigger logging looks like a workaround, what if in the future we want different levels of verbosity? The core elements must be simple and readable

atbenmurray: That is a different matter. It would be a good idea to have a discussion about logging in general across the app and determine what meets the needs of MONAI for all its scenarios. I followed Eric's advice about making it at least consistent with other places that use logging. I'd target a reworked logging mechanism as a 1.3 task

atbenmurray: `evaluate_with_overrides, apply_transforms, apply_transform and _apply_transform`

 is far more complex than what I've replaced it with. It also fixed issues like the issues that you were having with Lambda and it introduced a third lazy mode.

`apply_pending_transforms, apply_pending, apply_transform and _apply_transform`

wyli: What is the lambda issue? You mentioned multiple times and don't actually get it. Could you please file a ticket about it?

atbenmurray: The issue where you end up with a Lambda function in the wrong place in the applied operations sequence (it should be at the end but it ended up at the start because it got applied while a tensor had pending operations (the lazy integration test). It doesn't need a ticket because you guys are aware of it and have done several PRs to modify code and tests to account for it

monai/transforms/lazy/dictionary.py
wyli: What's are the potential consequences could you please make a few concrete examples? (Asking because previously this functionality was one of the main motivations for having this PR.)

monai/transforms/compose.py
atbenmurray: FYI - dictionary based transform output
```
            "Accumulate pending transforms - lazy mode: True, key: 'a', pending: 0, upcoming 'Flipd', transform.lazy: False (overridden)\n"
            "Accumulate pending transforms - lazy mode: True, key: 'b', pending: 0, upcoming 'Flipd', transform.lazy: False (overridden)\n"
Accumulate pending transforms - lazy mode: True, key: 'a', pending: 1, upcoming 'Spacingd', transform.lazy: False (overridden)
Accumulate pending transforms - lazy mode: True, key: 'b', pending: 1, upcoming 'Spacingd', transform.lazy: False (overridden)
Accumulate pending transforms - lazy mode: True, key: 'a', pending: 2, upcoming 'Rotate90d', transform.lazy: False (overridden)
Accumulate pending transforms - lazy mode: True, key: 'b', pending: 2, upcoming 'Rotate90d', transform.lazy: False (overridden)
Apply pending transforms - lazy mode: True, key: 'a', pending: 3, upcoming 'NormalizeIntensityd', transform is not lazy
```

wyli: This PR description says it resolves an issue with lambda, so it would be great to report the issue first. it's unclear to me what's not working and what's the new behaviour after this fix.

atbenmurray: tests\test_integration_lazy_samples.py
The pipeline has a LambdaD at the end of it. That LambdaD gets put out of place in the applied transforms, which blows up the invert. Hence the code that you added at https://github.com/Project-MONAI/MONAI/blob/4f291723c4ce1a1baf7efbd499495b777dfb1869/tests/test_integration_lazy_samples.py#L152

atbenmurray: This isn't needed because the LambdaD can't get out of place in the applied operations with the modified lazy resampling execution

atbenmurray: #6439

monai/transforms/transform.py
```diff
@@ -44,7 +45,11 @@
...
 def _apply_transform(
-    transform: Callable[..., ReturnType], parameters: Any, unpack_parameters: bool = False
+    transform: Callable[..., ReturnType],
+    data: Any,
+    unpack_parameters: bool = False,
+    lazy: str | bool | None = LazyMode.OFF,
+    overrides: dict | None = None,
 ) -> ReturnType:
     """
     Perform transformation `transform` with the provided parameters `parameters`.
```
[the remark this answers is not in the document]
atbenmurray: Done!

monai/transforms/compose.py
wyli: I think we spend enough time on this PR and try to merge it before v1.2 release, still don't agree on a few points discussed here, and it's also risky to merge this fairly large PR as we are already in codefreeze for v1.2. Let's revisit this after the release. For now I don't have time to discuss/improve it further.

by the way the logging still doesn't work, to reproduce:

```
import torch
import monai.transforms as mt

data = torch.zeros((1, 40, 40))
xform = mt.Compose([mt.Flip(0), mt.Rotate90(1), mt.Zoom(0.8)], lazy=True, logger_name="log")
xform(data)

```

doesn't output anything.

monai/transforms/transform.py
atbenmurray: Done

monai/transforms/compose.py
atbenmurray: Probably this is an issue of logging configuration. The logging tests as passing.

This is ready to go (pending :
. It fixes issues that exist (or existed) on dev
. It passes all tests
. It is more comprehensively documented

mjorgecardoso: Hi All,

This PR has several important features for 1.2, as it is critical for several stuff we are working on. Can we please try to merge this before 1.2 is released? I still see many PRs with new features being merged, so I don't quite get what the issue is.

wyli (changes requested):
- needs more design discussions for `monai/transforms/transform.py` https://github.com/Project-MONAI/MONAI/pull/6257#discussion_r1178405854
- although the pre-merge quick tests work fine, all the relevant integration tests are failing, because of the API changes introduced.

monai/transforms/lazy/functional.py
```diff
@@ -26,17 +26,83 @@
...
+
+    """
+    if isinstance(data, list):
+        return [apply_pending_transforms(d, logger_name=logger_name) for d in data]
+
+    if isinstance(data, tuple):
+        return tuple(apply_pending_transforms(d, logger_name=logger_name) for d in data)
+
+    if isinstance(data, dict):
+        d = data
```
ericspod, line 64: Is it needed to make a fresh dict here and not modify the input?
```suggestion
        d = dict(data)
```

atbenmurray: Ahh, this was actually a fix from
```
d = dict(data)
```
specifically to 
```
d = data
```
because tests were expecting id compatibility. it is true that `data` could just be used throughout instead

monai/transforms/compose.py
wyli: > @mjorgecardoso  I still see many PRs with new features being merged, so I don't quite get what the issue is.
> 

we are currently at the 2d quality assurance drop, 5th release candidate of v1.2, we shouldn't merge new feature like this into the release candidate. please could you help me understand 'many PRs with new features'?

wyli: @atbenmurray logging -- expected behaviour:
Screenshot 2023-04-27 at 13 35 46

this branch removes this feature, and in my opinion, replaced it with a worse design ...

monai/transforms/lazy/functional.py
ericspod: Should the tests be changed then or is this a property that we want? It needs to be commented here and (if not already) documented that this mutates the input dictionary.

atbenmurray: Ahh, I just need to move the dict copy inside the conditional

monai/transforms/compose.py
atbenmurray: @wyli I am taking a look at this

atbenmurray: New commit should resolve that. The only issue was that I needed to use `monai.app.utils.get_logger` to ensure that the named logger was properly configured

atbenmurray:
```
2023-04-27 17:56:05,617 - INFO - Accumulate pending transforms - lazy: True, pending: 0, upcoming 'Flip', transform.lazy: False (overridden)
2023-04-27 17:56:05,622 - INFO - Accumulate pending transforms - lazy: True, pending: 1, upcoming 'Rotate90', transform.lazy: False (overridden)
2023-04-27 17:56:05,625 - INFO - Accumulate pending transforms - lazy: True, pending: 2, upcoming 'Zoom', transform.lazy: False (overridden)
torch.Size([1, 40, 40, 40])
```

wyli: ok sure, marking this as resolved...

atbenmurray: Additional modification

```
import torch
import monai.transforms as mt

data = torch.zeros((1, 40, 40, 40))
xform = mt.Compose([mt.Flip(0), mt.Rotate90(1), mt.Zoom(0.8), mt.ApplyPending(), mt.Spacing(1.2)],
                   lazy=True, logger_name="log")
xform(data)
```

gives

```
2023-04-27 18:31:14,550 - INFO - Accumulate pending transforms - lazy: True, pending: 0, upcoming 'Flip', transform.lazy: False (overridden)
2023-04-27 18:31:14,552 - INFO - Accumulate pending transforms - lazy: True, pending: 1, upcoming 'Rotate90', transform.lazy: False (overridden)
2023-04-27 18:31:14,554 - INFO - Accumulate pending transforms - lazy: True, pending: 2, upcoming 'Zoom', transform.lazy: False (overridden)
2023-04-27 18:31:14,557 - INFO - Apply pending transforms - lazy: True, pending: 3, upcoming 'ApplyPending', transform is not lazy
2023-04-27 18:31:14,567 - INFO - Accumulate pending transforms - lazy: True, pending: 0, upcoming 'Spacing', transform.lazy: False (overridden)
2023-04-27 18:31:14,575 - INFO - Pending transforms applied: applied_operations: 4
```

atbenmurray: I think this should be sufficient:
```
import torch
import monai.transforms as mt

data = torch.zeros((1, 40, 40, 40))
xform = mt.Compose([mt.Flip(0), mt.Rotate90(1), mt.Zoom(0.8), mt.ApplyPending(), mt.Spacing(1.2)],
                   lazy=True, logger_name="log")
xform(data)
```

gives

```
2023-04-27 18:35:32,310 - INFO - Accumulate pending transforms - lazy: True, pending: 0, upcoming 'Flip', transform.lazy: False (overridden)
2023-04-27 18:35:32,313 - INFO - Accumulate pending transforms - lazy: True, pending: 1, upcoming 'Rotate90', transform.lazy: False (overridden)
2023-04-27 18:35:32,314 - INFO - Accumulate pending transforms - lazy: True, pending: 2, upcoming 'Zoom', transform.lazy: False (overridden)
2023-04-27 18:35:32,317 - INFO - Apply pending transforms - lazy: True, pending: 3, upcoming 'ApplyPending', transform is not lazy
2023-04-27 18:35:32,326 - INFO - Pending transforms applied: applied_operations: 3
2023-04-27 18:35:32,327 - INFO - Accumulate pending transforms - lazy: True, pending: 0, upcoming 'Spacing', transform.lazy: False (overridden)
2023-04-27 18:35:32,335 - INFO - Pending transforms applied: applied_operations: 4
```

atbenmurray: I need to modify the logging test to remove the timestamp. This will probably go in first thing tomorrow morning

monai/transforms/transform.py
```diff
@@ -61,18 +106,32 @@ def _apply_transform(
...
+    lazy_tx = isinstance(transform, LazyTrait)
+
+    if lazy_tx is False or lazy is False:
+        _log_pending_info(data, transform, "Apply pending transforms", lazy, logger_name)
+        data = apply_pending_transforms(data, overrides)
+    elif lazy is None and transform.lazy is False:  # type: ignore[attr-defined]
+        _log_pending_info(data, transform, "Apply pending transforms", lazy, logger_name)
+        data = apply_pending_transforms(data, overrides)
+    else:
+        _log_pending_info(data, transform, "Accumulate pending transforms", lazy, logger_name)
```
wyli, line 119: so now I think the only remaining debate (assuming it doesn't introduce any other debates) is here about 'correctness' and 'performance'. we can discuss it tomorrow at the dev meeting. cc @ericspod @atbenmurray @Nic-Ma

monai/transforms/transform.py
```diff
@@ -43,15 +45,58 @@
...
+        return
+    logger = get_logger(logger_name)
+
+    if isinstance(transform, LazyTrait):
+        if lazy is not None and lazy != transform.lazy:
+            tlazy = f", transform.lazy: {transform.lazy} (overridden)"
+        else:
+            tlazy = f", transform.lazy: {transform.lazy}"
+    else:
+        tlazy = ", transform is not lazy"
```
wyli, line 61: this kind of transform handling logic in `_log_pending_info` is duplicated in `_apply_transform`, it's not easy to read and maintain, let's think about a better way.

monai/transforms/compose.py
wyli: I ran the example using this PR, got:
```
  File "my_test.py", line 5, in <module>
    xform = mt.Compose([mt.Flip(0), mt.Rotate90(1), mt.Zoom(0.8), mt.ApplyPending(), mt.Spacing(1.2)],
AttributeError: module 'monai.transforms' has no attribute 'ApplyPending'

```

could you confirm @atbenmurray ?

monai/transforms/transform.py
atbenmurray: It is better out of the main logic than wrapped up with it

atbenmurray: evaluate_with_overrides actually mixes up logging and logic. I think the refactored code is more decoupled than that

wyli: It'll be difficult to maintain, cause mismatch of the actual behaviour and the logs. That'll be a new issue introduced by this change.

atbenmurray: I'm not sure what you mean. What mismatch of the actual behaviour and the logs? This code doesn't change any state on anything. It just logs what is happening, so I don't see how the logs and the behaviour will be different, unless there is a bug here that I am not seeing

atbenmurray: Are you referring to the fact that there are three lazy modes, False (Force Non lazy), True (Force Lazy) and None (do what the transforms themselves say)? If so, that is the intended design

wyli: for example, this logging says "(overriden)" but the actual behaviour is not in this function and when reading this code, we never know the actual overriding behaviour, and we don't know where to check the logic.

or another example, `_apply_transform` supports recursively checking a nested structure of dictionaries/tuples, but the `_log_pending_info` doesn't have this logic.
so this already creates various inconsistencies. My original implementation doesn't have this issue.

monai/transforms/lazy/functional.py
```diff
@@ -26,17 +26,83 @@
...
+        for k, v in d.items():
+            if isinstance(v, MetaTensor) and v.has_pending_operations:
+                overrides_ = None if overrides is None else overrides[k]
+                d[k], _ = apply_pending(v, overrides=overrides_, logger_name=logger_name)
+        return d
```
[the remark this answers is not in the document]
atbenmurray: Yep, caught that while updating the compose logging tests. Its fixed in the upcoming commit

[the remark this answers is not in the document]
atbenmurray: Can you be more specific? Do you have a case that is failing?

monai/transforms/transform.py
atbenmurray: overridden means that the lazy mode of the transform is being overridden by the Compose lazy mode

atbenmurray: The behaviour is not in the logging function because it is a logging function

monai/transforms/lazy/functional.py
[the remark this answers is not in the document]
wyli: I mean this `tlazy = f", transform.lazy: {transform.lazy} (overridden)"` in `_log_pending_info `

monai/transforms/transform.py
wyli: I mean this `tlazy = f", transform.lazy: {transform.lazy} (overridden)"` in `_log_pending_info `

monai/transforms/lazy/functional.py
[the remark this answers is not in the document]
atbenmurray: I think I need an example here. Are you talking about overrides or something else?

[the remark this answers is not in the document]
atbenmurray: If you are talking about overrides:
```
override_keys = {'a', 'b'}
overrides = {'mode': ('nearest', None), 'padding_mode': (None, 'edge')}
```
becomes
```
overrides = {'a': {'mode': 'nearest'}, 'b': {'padding_mode': 'edge'}}
```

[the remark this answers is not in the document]
wyli: a simple case is

```
import torch

import monai.transforms as mt

d = {"image": torch.zeros((1, 40, 40, 40))}

xform = mt.Compose(
    [
        mt.FlipD(keys="image", spatial_axis=0),
        mt.Rotate90D(keys="label", k=1, allow_missing_keys=True),
        mt.ZoomD(keys=("image", "label"), zoom=0.8, allow_missing_keys=True),
        mt.SpacingD(keys="image", pixdim=1.2),
    ],
    lazy=True,
    logger_name="log"
)
xform(d)
```
your implementation gives:
```
    pcount = len(data[k].pending_operations) if isinstance(data[k], MetaTensor) else 0
KeyError: 'label'

```

there are for sure other cases, please be extra careful when refactoring. I didn't put all the test cases in the current version.

monai/transforms/transform.py
atbenmurray: I'm not following. This is a substring that gets formatted and then put into the main logging string.

```python
    if isinstance(transform, LazyTrait):
        if lazy is not None and lazy != transform.lazy:
            tlazy = f", transform.lazy: {transform.lazy} (overridden)"
        else:
            tlazy = f", transform.lazy: {transform.lazy}"
    else:
        tlazy = ", transform is not lazy"

    if isinstance(transform, MapTransform):
        for k in transform.keys:
            pcount = len(data[k].pending_operations) if isinstance(data[k], MetaTensor) else 0
            logger.info(
                f"{activity} - lazy mode: {lazy}, key: '{k}', "
                f"pending: {pcount}, upcoming '{transform.__class__.__name__}'{tlazy}"
            )
    else:
        pcount = len(data.pending_operations) if isinstance(data, MetaTensor) else 0
        logger.info(
            f"{activity} - lazy: {lazy}, " f"pending: {pcount}, upcoming '{transform.__class__.__name__}'{tlazy}"
        )
```

If it is a lazy transform, we log whether the transforms.lazy is different from the compose's lazy mode. If it is, we note that the transform's .lazy is being overridden on the call.
That tlazy string then gets inserted into the full logging string.

monai/transforms/lazy/functional.py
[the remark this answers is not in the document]
atbenmurray: Ok, I have a fix for that. Added to the upcoming commit.

monai/transforms/transform.py
wyli: that's just an example that the substring should be agnostic to the actual lazy resampling logic, 

if you don't understand this example, then please think about the other example I provided, it may help you understand the design issue here.

monai/transforms/lazy/functional.py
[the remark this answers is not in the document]
wyli: there's no rush to fix it quickly, I'd recommend reading the existing implementations and the core logic carefully first.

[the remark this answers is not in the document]
atbenmurray: It is a simple and obvious fix. I check whether the key from transform.keys is on the data.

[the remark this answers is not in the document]
atbenmurray: As you know, I've been over that core logic very carefully. I did a number of refactors for clarity, but at the end of the day it is still extremely complex in its implementation and needed refactoring in order to be able to provide the lazy mode where compose would honour the lazy values set on the transforms.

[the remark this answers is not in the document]
atbenmurray: I agree.
I do have to point out that https://github.com/Project-MONAI/MONAI/pull/5860 was merged to dev without my signoff. It contained significant issues and still contains significant issues that in my opinion should have been resolved on the branch before it was merged.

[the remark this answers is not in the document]
wyli: let's focus on this PR. The previous PR received at least one of [the core reviewers](https://github.com/orgs/Project-MONAI/teams/core-reviewers)'s approval for merging.

monai/transforms/transform.py
atbenmurray: This is actually informing the reader that although the transform has 'lazy' True it is being overridden by the pipeline executing with 'lazy' False

wyli: for example, if we have a list of image patches, the logging is incorrect...
```
import torch

import monai.transforms as mt

d = [{"image": torch.zeros((1, 40, 40, 40))}, {"image": torch.zeros((1, 40, 40, 40))}, {"image": torch.zeros((1, 40, 40, 40))}]

xform = mt.Compose(
    [
        mt.FlipD(keys="image", spatial_axis=0),
        mt.Rotate90D(keys="label", k=1, allow_missing_keys=True),
        mt.ZoomD(keys=("image", "label"), zoom=0.8, allow_missing_keys=True),
        mt.SpacingD(keys="image", pixdim=1.2),
    ],
    lazy=True,
    logger_name="log"
)
xform(d)
```

monai/transforms/compose.py
wyli: this is resolved..

atbenmurray:
https://github.com/Project-MONAI/MONAI/issues/6450 discusses compiler options. I think it this point it is our best bet.

atbenmurray:
@wyli @ericspod
Ok, so I've implemented the changes discussed as a result of Friday's meeting and subsequent conversations

Lazy resampling now has three modes, controlled by an `options` parameter on compose:
1. "in_order": this preserves the ordering while executing during lazy mode: `options = None` and means that users pipelines cannot be broken by setting `lazy=True`
2. "reorder: lazy_last": this allows some reordering but treats ApplyPending/ApplyPendingd as a barrier, so reordering can only happen across ApplyPending/ApplyPendingd. It supports `inverse` and works the same way as `OneOf`, `RandomOrder`, etc. To use it you supply `options={'reorder': 'lazy_last'}`
3. "reorder: lazy_last_nosync": this _should_ mimic the dev mode behaviour precisely. It allows effective execution order to be different for different keys in a dictionary and does not allow `inverse`. To use it, you supply `options={'reorder': 'lazy_last_nosync'}

I've also tidied up some of the validation so that inverse behaviour is clearer to the user

Tests need augmenting still

wyli:
> 3. "reorder: lazy_last_nosync": this _should_ mimic the dev mode behaviour precisely. It allows effective execution order to be different for different keys in a dictionary and does not allow `inverse`.
> 

the current dev branch supports invert, which is used in benchmarking the CT segmentation workflow https://github.com/Project-MONAI/model-zoo/pull/397/ please see the train.py file in that PR. without the inverse, we may not be able to achieve the same validation scores. please double-check.

monai/transforms/compose.py
```diff
@@ -308,54 +369,44 @@ class Compose(Randomizable, InvertibleTransform):
...
         transforms: Sequence[Callable] | Callable | None = None,
         map_items: bool = True,
         unpack_items: bool = False,
-        log_stats: bool = False,
-        lazy_evaluation: bool | None = None,
+        lazy: bool | None = False,
         overrides: dict | None = None,
-        override_keys: Sequence[str] | None = None,
-        verbose: bool = False,
+        options: dict | None = None,
```
wyli, line 397: please add documentation for this `options` then I guess we can start reviewing the PR?

atbenmurray:
When does it support it. In the case where there are no non lazy transforms?

atbenmurray:
@wyli Please confirm for me the following:

As I understand it, you can invert with `lazy_last_nosync` if and only if your pipeline contains no transforms that are both invertible and not lazy (ie Lambda)

wyli:
> @wyli Please confirm for me the following:
> 
> As I understand it, you can invert with `lazy_last_nosync` if and only if your pipeline contains no transforms that are both invertible and not lazy (ie Lambda)

yeah, I'm not sure about the "if and only if" statement. "pipeline contains no transforms that are both invertible and not lazy" is sufficient but not necessary.

wyli:
> > > @wyli Please confirm for me the following:
> > > As I understand it, you can invert with `lazy_last_nosync` if and only if your pipeline contains no transforms that are both invertible and not lazy (ie Lambda)
> > 
> > 
> > yeah, I'm not sure about the "if and only if" statement. "pipeline contains no transforms that are both invertible and not lazy" is sufficient but not necessary.
> 
> I'll rephrase: . If you have transforms that are not invertible, they aren't added to the applied operations, so are not touched during inverse . If you have only lazy transforms and ApplyPendings, again you are fine because the lazy transforms are ordered correctly . If you have transforms that are invertible but not lazy, they mean that the applied order is not compatible with the transform order, and so inverse cannot be run
> 
> Hence if and only if contains no non-lazy, invertible transforms

Still inaccurate... if there's `Identity[D]` transforms placed accordingly before the non-lazy invertible transforms, the current dev branch can do the inverse.

If you are looking for the precise condition of not allowing inverse, it's been described in this PR in the warning message https://github.com/Project-MONAI/MONAI/pull/6372/files

inversing of a sequence of user-defined transforms applied to a metatensor is not possible if, when applying the transforms, `applied_operations` of that metatensor is modified while its `pending_operations` is not empty.

atbenmurray:
> > > > @wyli Please confirm for me the following:
> > > > As I understand it, you can invert with `lazy_last_nosync` if and only if your pipeline contains no transforms that are both invertible and not lazy (ie Lambda)
> > > 
> > > 
> > > yeah, I'm not sure about the "if and only if" statement. "pipeline contains no transforms that are both invertible and not lazy" is sufficient but not necessary.
> > 
> > 
> > I'll rephrase: . If you have transforms that are not invertible, they aren't added to the applied operations, so are not touched during inverse . If you have only lazy transforms and ApplyPendings, again you are fine because the lazy transforms are ordered correctly . If you have transforms that are invertible but not lazy, they mean that the applied order is not compatible with the transform order, and so inverse cannot be run
> > Hence if and only if contains no non-lazy, invertible transforms
> 
> Still inaccurate... if there's `Identity[D]` transforms placed accordingly before the non-lazy invertible transforms, the current dev branch can do the inverse.
> 
> If you are looking for the precise condition of not allowing inverse, it's been described in this PR in the warning message https://github.com/Project-MONAI/MONAI/pull/6372/files
> 
> inversing of a sequence of user-defined transforms applied to a metatensor is not possible if, when applying the transforms, `applied_operations` of that metatensor is modified while its `pending_operations` is not empty.

Those will need to be replaced with ApplyPending/ApplyPendingd, but fine, yes, they are also not added to the applied operations so don't break invertibility

wyli:
currently the runtime error will be raised with an additional warning of the details when accessing the applied_operations in such cases:
https://github.com/Project-MONAI/MONAI/blob/4bfc8e97661d44cc13553721e5025005757bad91/tests/test_invert.py#L100-L102

atbenmurray:
I think it is better to have inverse throw the reasons why data isn't convertible. Those warnings tend to get lost in the noise. Given that we capture the reason why inverse will fail during the forward pass, we really ought to make use of it if we know inverse will fail. Anyway, that should avoid the use of a can invert flag on the execution policy

monai/transforms/inverse.py
```diff
@@ -227,17 +223,16 @@ def track_transform_meta(
...
+                if key is not None:
+                    msg += f" for key {key}"
+
                 pend = out_obj.pending_operations[-1]
-                if not isinstance(pend.get(TraceKeys.EXTRA_INFO), dict):
-                    pend[TraceKeys.EXTRA_INFO] = dict(pend.get(TraceKeys.EXTRA_INFO, {}))
-                if not isinstance(info.get(TraceKeys.EXTRA_INFO), dict):
-                    info[TraceKeys.EXTRA_INFO] = dict(info.get(TraceKeys.EXTRA_INFO, {}))
-                info[TraceKeys.EXTRA_INFO]["warn"] = pend[TraceKeys.EXTRA_INFO]["warn"] = msg
+                reasons = pend.get(TraceKeys.INVERT_DISABLED, list())
```
wyli, line 233: I don't think the flag name `INVERT_DISABLED` is appropriate here. at this point this is not just an 'invert' related issue, maybe some user-specified transforms accidentally run into this logic, and the result is also not just about 'inverse', it potentially affects any operations that rely on the ordering of applied_operations.

atbenmurray: What are the other circumstances?

wyli: these `TraceableTransform` and `applied_operations` are generic APIs in monai, the user may call `track_transform_meta` in their own transform, may read `applied_operations` in their applications and process their business logic.

atbenmurray: That doesn't matter. For whatever reason that has happened, if someone pushes an `applied_operation` while there are `pending_operations`, the order has changed and the pipeline can't be inverted by `Compose.inverse`.

wyli: it matters because invertible is just one of the possible applications of traceable

atbenmurray: In that case, the logic shouldn't be part of that function, but should be handled outside of the function.

atbenmurray: @wyli There is now a STATUS TraceKey and a new enum TraceStatusKeys that currently contains PENDING_DURING_APPLY.

atbenmurray:
`test_integration_lazy_samples.py` now passes for `option=None` and `option={"reorder": "lazy_last_nosync"}`

wyli:
> `test_integration_lazy_samples.py` now passes for `option=None` and `option={"reorder": "lazy_last_nosync"}`

Hi @atbenmurray there are min-tests failing... please let us know when this is ready for review. (We'll also need a few days for providing comments and addressing comment about the code changes and repeating some basic benchmarks to confirm.)

monai/transforms/transform.py
wyli: this is still an issue as I run it with the latest commit (db1486bfc7be170f1a6517b3194b84e448e696ba), the expected logging (from the dev branch):

```
2023-05-04 10:01:10,000 - INFO - Lazy - 'image' - upcoming: 'Zoomd'- pending 1
2023-05-04 10:01:10,000 - INFO - Lazy - 'image' - upcoming: 'Zoomd'- pending 1
2023-05-04 10:01:10,000 - INFO - Lazy - 'image' - upcoming: 'Zoomd'- pending 1
2023-05-04 10:01:10,005 - INFO - Lazy - 'image' - upcoming: 'Spacingd'- pending 2
2023-05-04 10:01:10,006 - INFO - Lazy - 'image' - upcoming: 'Spacingd'- pending 2
2023-05-04 10:01:10,006 - INFO - Lazy - 'image' - upcoming: 'Spacingd'- pending 2
2023-05-04 10:01:10,019 - INFO - Evaluated - 'image' - up-to-date for - final output
2023-05-04 10:01:10,027 - INFO - Evaluated - 'image' - up-to-date for - final output
2023-05-04 10:01:10,035 - INFO - Evaluated - 'image' - up-to-date for - final output
```

wyli: the logging results are incorrect for the use case I provided above. that's also a sign of inflexible API design. now I'm worried about this refactoring, at this speed of development, it can't be finished as promised in the previous meeting.

atbenmurray: How are they wrong?

atbenmurray:
```
2023-05-04 12:36:35,087 - INFO - Accumulate pending transforms - lazy mode: True, key: 'image', pending: 0, upcoming 'Flipd', transform.lazy: False (overridden)
2023-05-04 12:36:35,090 - INFO - Accumulate pending transforms - lazy mode: True, key: 'image', pending: 0, upcoming 'Flipd', transform.lazy: False (overridden)
2023-05-04 12:36:35,090 - INFO - Accumulate pending transforms - lazy mode: True, key: 'image', pending: 0, upcoming 'Flipd', transform.lazy: False (overridden)
2023-05-04 12:36:35,091 - INFO - Accumulate pending transforms - lazy mode: True, key: 'image', pending: 1, upcoming 'Zoomd', transform.lazy: False (overridden)
2023-05-04 12:36:35,094 - INFO - Accumulate pending transforms - lazy mode: True, key: 'image', pending: 1, upcoming 'Zoomd', transform.lazy: False (overridden)
2023-05-04 12:36:35,096 - INFO - Accumulate pending transforms - lazy mode: True, key: 'image', pending: 1, upcoming 'Zoomd', transform.lazy: False (overridden)
2023-05-04 12:36:35,098 - INFO - Accumulate pending transforms - lazy mode: True, key: 'image', pending: 2, upcoming 'Spacingd', transform.lazy: False (overridden)
2023-05-04 12:36:35,101 - INFO - Accumulate pending transforms - lazy mode: True, key: 'image', pending: 2, upcoming 'Spacingd', transform.lazy: False (overridden)
2023-05-04 12:36:35,103 - INFO - Accumulate pending transforms - lazy mode: True, key: 'image', pending: 2, upcoming 'Spacingd', transform.lazy: False (overridden)
2023-05-04 12:36:35,112 - INFO - Pending transforms applied: key: 'image', applied_operations: 3
2023-05-04 12:36:35,118 - INFO - Pending transforms applied: key: 'image', applied_operations: 3
2023-05-04 12:36:35,123 - INFO - Pending transforms applied: key: 'image', applied_operations: 3
```

atbenmurray: I'd say these aren't missing the first transform (missing out the first transform is a bug, frankly) and provide additional information. Furthermore, given that logging must occur for two separate execution strategies, repeating the logging functionality across multiple paths would be less than ideal. It isn't inflexible design to have a function that handles logging of an object. Its more inflexible to embed that logging in with the logic as opposed to decoupling it. That's how the first transform ended up being missed from the logging

atbenmurray: Furthermore, if we now want different levels of verbosity (say instead of logging the number of pending transforms we log a list of the names) we can add that to the logging code. If you put that kind of logic in and around the actual logic that does the work, no-one will be able to make sense of what is actual logic and what is logging

tests/test_integration_lazy_samples.py
```diff
@@ -193,27 +197,34 @@ def train_and_infer(self, idx=0):
...
-        for a, b in zip(sorted(lazy_files), sorted(regular_files)):
-            img_lazy = mt.LoadImage(image_only=True)(a)
-            img_regular = mt.LoadImage(image_only=True)(b)
-            diff = np.size(img_lazy) - np.sum(np.isclose(img_lazy, img_regular, atol=1e-4))
-            diff_rate = diff / np.size(img_lazy)
-            diffs.append(diff_rate)
-            np.testing.assert_allclose(diff_rate, 0.0, atol=0.03)
-        print("volume diff:", diffs)
-        return results
+        for options in ({"reorder": "lazy_last_nosync"},):
```
ericspod, line 204: Why the for-loop here?

tests/test_compose.py
```diff
@@ -283,6 +305,308 @@ def test_compose_execute_equivalence(self, keys, pipeline):
...
+        ),
+    ],
+]
+
+
+class TestComposeExecuteWithLogging(unittest.TestCase):
+    @staticmethod
+    def data_from_keys(keys):
+        if keys is None:
+            data = torch.unsqueeze(torch.tensor(np.arange(12 * 16).reshape(12, 16)), dim=0)
```
ericspod, line 583:
```suggestion
            data = torch.arange(12 * 16).reshape(1, 12, 16)
```

wyli:
> PR text is updated to list the changes

can you highlight for each item whether they are breaking changes or non-breaking new features?

monai/transforms/transform.py
wyli: the code has been changed after my previous comment, and the logging output is now different... just another example showing the logging mechanism in this PR is error prone...

atbenmurray:
Having an issue with `tests/test_integration_lazy_samples.py`
It does inversion by pulling out the transforms from the Compose object (this is something that we should really not be doing. Those transforms are reordered by `options={"reorder": "lazy_last"}`, obviously the test then fails. I think the test should be updated to use the `compose.inverse` instead. Is there a specific reason that it isn't?

monai/transforms/transform.py
atbenmurray: Wenqi. I moved the function to a different file...

atbenmurray: I think it would be most productive to focus on ensuring that the actual behaviour is correct for the pipelines that require it, to avoid having to re-benchmark

atbenmurray:
`ComposeCompiler` is a design that I've tested out on the full branch. You've seen it before. In this case, it is a very small class that, given a flag, produces a policy object. It is also docstringed as not being part of the public API.

The lazy execution code has been separated out into 'in_order' and 'out_of_order' execution. They are no less tested (actually more tested) than what is on `dev`.

Reordering of the transforms uses the same mechanism as is used by `OneOf`, `SomeOf` and `RandomOrder`.

All of this was necessary in order to provide an optional lazy mode that matched `dev` for the benchmarks and a default mode that doesn't break pipelines in the general case.

atbenmurray:
@wyli
At this point, I think that the best thing to do is to try to propose the most constructive way forward that I can think of.

I think that the PR is very close to being ready. @ericspod has gone over the PR several times and also believes that the solution is close to merge ready.

The primary outstanding task from a correctness standpoint is to confirm that `lazy=True, options={"reorder": "lazy_last_nosync"}` does what `dev` does. That is something that you are best placed to do.

The remaining tasks are making sure we haven't missed anything from a test standpoint and documenting.

If you have design concerns, I am more than happy to meet with you and discuss them properly. We know that there are documentation holes, but documentation is low risk and, in any case, is something that I have been working on today. Similarly, we can ensure that we haven't left any testing holes, either left over from `dev` or newly introduced in the PR.

`ComposeCompiler` in particular is an implementation detail. I've discussed the design with you and presented it at several dev meetings, but it is documented as being an implementation detail that users are not supposed to interact with directly, so it can be subsequently changed without impact.

the different lazy executors have been added by necessity. The "in order" executor implements what was supposed to be the default behavour that has been discussed and presented throughout. The "out of order" executor carefully mimics the behavour on `dev` and I believe it now does that completely. If you can find any issues with it, we can certainly deal with them. I separated them out from each other because I found the cyclomatic complexity of `evaluate_with_overrides` to be prohibitive even as it was, and trying to have those two different execution strategies bound up in the same code would have caused many problems, especially if we then required a third. This design separates them out and makes the logic of when one needs to apply pending transforms and when one doesn't to be very clear.

We can discuss any of these matters directly if you prefer. I am happy to meet, as always.

wyli:
sure @atbenmurray, our discussions haven't been very productive/constructive... and I don't think I have enough energy to follow up or to write more comments about this PR. (but as one of the core maintainers of this codebase, I wholeheartedly hope this PR is not merged in the current form.)

Nic-Ma:
Hi @atbenmurray @ericspod @wyli ,

Sorry I didn't follow carefully in this PR, currently, I think the biggest gap of design discussion is the transform order?
And this PR introduced many new concepts to solve the order problem in new ways, to help quickly move forward, how about let's do a simplified step first: We only do lazy resampling for **continuous spatial / crop pad transforms**, then we can guarantee the order of other transforms? I think we don't need much code change to achieve it in MONAI 1.2.
When this first step merged, we can enhance it with more fancy methods to adjust the transform order in the next version?
What do you think?

Thanks.

atbenmurray:
> Hi @atbenmurray @ericspod @wyli ,
> 
> Sorry I didn't follow carefully in this PR, currently, I think the biggest gap of design discussion is the transform order? And this PR introduced many new concepts to solve the order problem in new ways, to help quickly move forward, how about let's do a simplified step first: We only do lazy resampling for **continuous spatial / crop pad transforms**, then we can guarantee the order of other transforms? I think we don't need much code change to achieve it in MONAI 1.2. When this first step merged, we can enhance it with more fancy methods to adjust the transform order in the next version? What do you think? Please feel free to correct me if I misunderstood anything.
> 
> Thanks in advance.


Hi Nic, the problem that we would have there is that we would be back to the issue that we need to replicate the `dev` behaviour in order to avoid the new QA cycle. If we change the behaviour now, we are back to the same issues that we had before:

1. The pipelines will break / output wrong results in the general case: `[Flip, Spacing, Rand3DElastic, Rotate]` becomes `[Rand3DElastic, Flip, Spacing, Rotate]`. This will upset a lot of users, I think
2. Not having 'benchmark' mode {'rotate': 'lazy_last_nosync'} will invalidate the benchmarks and impact their performance, but it is absolutely only for expert users as they have to synchronise with `ApplyPending[d]` correctly.

I've had a week with the `dev` behaviour and I believe that I have replicated it successfully, barring issues that arise during the integration tests. I've also provided the standard lazy resampling mode as per the design that does no reordering and should be completely safe with all pipelines. Nothing should break when the user sets `lazy=True`.

We agreed last Friday that this was probably the best way forward and the discussion is reflected in #6450. Providing the safe (no-out of order execution) mode as per the design means users can set lazy=True safely. Providing the `dev` behaviour means no benchmark invalidation.

I went over the design and implementation with @ericspod today; I'm happy to go over the design and implementation with you and or @wenqi tomorrow.

atbenmurray:
> sure @atbenmurray, our discussions haven't been very productive/constructive... and I don't think I have enough energy to follow up or to write more comments about this PR. (but as one of the core maintainers of this codebase, I wholeheartedly hope this PR is not merged in the current form.)

I'm sorry that you feel our discussions have not been constructive. It hasn't been my intention, and the last week's work has been entirely towards trying to meet the resolution that we agreed last Friday. I honestly feel that this work is closer than you think and, as I've said, I am happy to sit with you and go over every line of it if that will help

monai/transforms/compose.py
```diff
@@ -188,21 +115,164 @@ def execute_compose(
...
-    data = evaluate_with_overrides(
-        data, None, lazy_evaluation=lazy_evaluation, overrides=overrides, override_keys=override_keys, verbose=verbose
-    )
+    if isinstance(data, (list, tuple)) and map_items:
+        data = [apply_pending_transforms(d, None, overrides, logger_name=logger_name) for d in data]
+    data = apply_pending_transforms(data, None, overrides, logger_name=logger_name)
     return data
 
 
+class ComposeCompiler:
```
ericspod, line 134: As we've discussed today at the meeting I'd suggest something like `LazyPolicy`  given the current state of this class. If the intent is to evolve this to do more complicated work to "compile" in some manner we might want to put that into a new class anyway and keep something to do policy like this separate.

monai/transforms/compose.py
```diff
@@ -432,51 +509,105 @@ def __len__(self):
...
+    @staticmethod
+    def _raise_if_tensor_is_not_invertible(data: Any):
+        invertible, reasons = is_tensor_invertible(data)
+
+        if invertible is False:
+            if reasons is not None:
+                reason_text = "\n".join(reasons)
+                raise RuntimeError(f"Unable to run inverse on 'data' for the following reasons:\n{reason_text}")
+            else:
+                raise RuntimeError("Unable to run inverse on 'data'; no reason logged in trace data")
```
ericspod, line 610:
```suggestion
        if invertible is False:
            reason_text = "No reaon logged in trace data" if reasons is None else "\n".join([""]+list(reasons))
            raise RuntimeError(f"Unable to run inverse on 'data': {reason_text}")
```

monai/transforms/compose.py
```diff
@@ -188,21 +115,164 @@ def execute_compose(
...
+    be considered an implementation detail that should not be interacted with directly by users of MONAI, although that
+    may change in subsequent releases. Its job is to parse options provided to `Compose.__call__`_ to set execution
+    modes for executing the pipeline.
+
+    See `Compose`_ for a detailed explanation of lazy resampling.
+    """
+
+    def __init__(self):
+        # construct the list of options
+        options = {"reorder": {"lazy_last": self.reorder_lazy_last, "lazy_last_nosync": self.reorder_lazy_last_nosync}}
```
ericspod, line 144: I would avoid having nested structures like this, instead:

```suggestion
        options = {("reorder","lazy_last"): self.reorder_lazy_last, ("reorder","lazy_last_nosync"): self.reorder_lazy_last_nosync}
```

Also the string keywords here should be in enums and the structure that's expected should be documented in the docstring.

monai/transforms/compose.py
```diff
@@ -188,21 +115,164 @@ def execute_compose(
...
+        self.options = options
+
+    def __call__(self, transforms, lazy: bool | None, options: dict | None = None):
```
ericspod, line 147: `options` is only ever used with one key-value pair so maybe just have two variables here. Otherwise `eval_options` or something, with more docstring on what it means.

monai/transforms/compose.py
```diff
@@ -188,21 +115,164 @@ def execute_compose(
...
+            a dictionary specifying the execution policy
+
+        """
+        if lazy is False or options is None:
+            return ExecutionOptions.generate_policy()
+
+        if len(options.keys()) > 1:
+            raise ValueError("Only one option can currently be set")
+
+        for k, v in options.items():
```
ericspod, line 168: `k, v = monai.utils.first(options.items())`?

monai/transforms/compose.py
```diff
@@ -308,54 +393,45 @@ class Compose(Randomizable, InvertibleTransform):
...
         unpack_items: bool = False,
-        log_stats: bool = False,
-        lazy_evaluation: bool | None = None,
+        lazy: bool | None = False,
         overrides: dict | None = None,
-        override_keys: Sequence[str] | None = None,
-        verbose: bool = False,
+        options: dict | None = None,
+        logger_name: str | None = None,
     ) -> None:
```
ericspod, line 423:
```suggestion
        logger_name: str | None = None,
        exec_options: ExecutionOptions | None = None
    ) -> None:
```

monai/transforms/compose.py
```diff
@@ -308,54 +393,45 @@ class Compose(Randomizable, InvertibleTransform):
...
-        self.override_keys = override_keys
-        self.verbose = verbose
-
-        if self.lazy_evaluation is not None:
-            for t in self.flatten().transforms:  # TODO: test Compose of Compose/OneOf
-                if isinstance(t, LazyTransform):
-                    t.lazy_evaluation = self.lazy_evaluation
+        self.options = options
+        self.logger_name = logger_name
+        self.execution_options = ExecutionOptions()
```
ericspod, line 434:
```suggestion
        self.exec_options = ExecutionOptions() if exec_options is None else exec_options
```

monai/transforms/traits.py
```diff
@@ -27,23 +29,44 @@ class LazyTrait:
...
         """
-        Set whether lazy_evaluation is enabled for this transform instance.
+        Set whether lazy evaluation is enabled for this transform instance.
         Args:
             enabled: True if the transform should operate in a lazy fashion, False if not.
         """
         raise NotImplementedError()
 
+    @property
+    def partially_lazy(self):
```
ericspod, line 50:
```suggestion
    def partially_lazy(self) -> bool:
```

atbenmurray:
This is out of date. We'll revisit further lazy resampling functionality in new PRs. See #7151 for details

PR #6269, opened 2023-04-02
6268 enhance hovernet load pretrained function

### LINKED ISSUE #6268: Enhance Hovernet load pretrained weights functionality
Currently, `_remap_standard_resnet_model` is designed according to the network structure shown in https://pytorch.org/vision/main/models/generated/torchvision.models.resnet50.html .
Therefore, we assume that there is not extra key (see also [here](https://github.com/Project-MONAI/MONAI/blob/1b0808db48b2a3c0eb4287c6290bd754e70fb88b/monai/networks/nets/hovernet.py#L655) ) to extract the expected state dict.
It would be nice to make the network be compatible with the weights that contain some other keys. For example, users may also save other information rather than state dict within the weights:
```
dict_keys(['epoch', 'arch', 'state_dict', 'optimizer', 'version', 'args', 'amp_scaler', 'metric'])
```
Therefore, I suggest we can add an arg to support this kind of case. In addition, we can also add some notifications within the load function to help users double check that if the load method is successful.

cc @KumoLiu

yiheng-wang-nv:
Fixes #6268 .

### Description

This PR enhances Hovernet's load pretrained function.

### Types of changes


Nic-Ma:
Hi @yiheng-wang-nv @KumoLiu ,

I don't quite understand the use case of this PR, could you please help share some examples here?

Thanks in advance.

yiheng-wang-nv:
> Hi @yiheng-wang-nv @KumoLiu ,
> 
> I don't quite understand the use case of this PR, could you please help share some examples here? I mainly want to ensure the bundle user experience is not affected, especially configs.
> 
> Thanks in advance.

Hi Nic, some users are used to save not only weights, but also other information such as learning rate, epoch ... into the whole weight file, if print the keys of the file we will see something like:
```
dict_keys(['epoch', 'arch', 'state_dict', 'optimizer', 'version', 'args', 'amp_scaler', 'metric'])
```
Where, only the `state_dict` is the place we need.

Currently, Hovernet is designed to load standard pytorch resnet50 pretrained weights which only contain the weights information, and print the keys of the file we will see:
```
odict_keys(['conv1.weight', 'bn1.running_mean', 'bn1.running_var', 'bn1.weight', 'bn1.bias', 'layer1.0.conv1.weight', 'layer1.0.bn1.running_mean', 'layer1.0.bn1.running_var', 'layer1.0.bn1.weight', 'layer1.0.bn1.bias', 'layer1.0.conv2.weight', 'layer1.0.bn2.running_mean', 'layer1.0.bn2.running_var', 'layer1.0.bn2.weight', 'layer1.0.bn2.bias', 'layer1.0.conv3.weight', 'layer1.0.bn3.running_mean', 'layer1.0.bn3.running_var', 'layer1.0.bn3.weight', 'layer1.0.bn3.bias', 'layer1.0.downsample.0.weight' ...
```

Nic-Ma:
I mean an example to show how to use this arg in a python program and a bundle config.

yiheng-wang-nv:
I see, in a bundle config we can do:
```
    "network_def": {
        "_target_": "HoVerNet",
        "mode": "@hovernet_mode",
        "in_channels": 3,
        "out_classes": 5,
        "adapt_standard_resnet": true,
        "pretrained_url": "$None",
        "freeze_encoder": true,
        "pretrained_state_dict_key": "state_dict"
    },
```
Only add this arg: `"pretrained_state_dict_key"`

Nic-Ma (approved):

Nic-Ma:
@yiheng-wang-nv OK, looks good to me, please fix the CICD errors.

Thanks.

PR #6270, opened 2023-04-02
Autorunner clear gpu mem allocatoins of dataanalyzer

myron:
clears gpu mem allocatoins of dataanalyzer

For some reason, after data analyzer finishes, GPU memory was not fully released, and about 1GB persisted, 
this fix helps.

PS: about 300MB still persists on CUDA:0 device even after this fix, but unclear why yet

### DIFF
--- a/monai/apps/auto3dseg/auto_runner.py
+++ b/monai/apps/auto3dseg/auto_runner.py
@@ -719,7 +719,7 @@ class AutoRunner:
             )
             da.get_all_case_stats()
 
-            da = None
+            da = None  # type: ignore
             torch.cuda.empty_cache()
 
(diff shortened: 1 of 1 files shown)


wyli:
the 300MB is used by the cuda driver and won't be released unless quitting the process https://github.com/pytorch/pytorch/issues/20532

wyli (approved):

myron:
> the 300MB is used by the cuda driver and won't be released unless quitting the process [pytorch/pytorch#20532](https://github.com/pytorch/pytorch/issues/20532)

interesting, thank you for letting me know! and thank you for merging quickly

PR #6272, opened 2023-04-03
Initial version for multinode auto_runner and ensembler

### LINKED ISSUE #6191: Enable Multi-node training

### LINKED ISSUE #6259: autorunner support multinode multigpu devices
Update the autorunner to support multinode training and ensembling.
Requires some refactoring of auto-runner.
Currently, autorunner uses set_training_params to define  a variable called "CUDA_VISIBLE_DEVICES", which is not a global env variable for autorunner, only for subprocess train.py. This creates a problem for storing and sharing device settings across functions in autorunner. 
Plans:
1. Create autorunner.set_device() function to set self.device_setting->Dict to  store "NUM_NODE", "CUDA_VISIBLE_DEVICE", "MULTI_NODE_START_METHOD" e.t.c.. Separate those values from self.train_params. If user did not pass any parameters, use os.environ.get("NUM_NODES", 1) e.t.c.

2. Data analyzer: currently use torch.cuda.device_count for multi-gpu processing. The autorunner should pass self.device_setting['CUDA_VISIBLE_DEVICE'] to data analyzer for which GPU to use. Not supporting multinode due to nested-dict gathering problem

3.  BundleAlgo.train: currently autorunner passes self.train_params to bundleAlgo subprocess train. Change it to self.device_setting, and update all the logic in subprocess cmd creation. Launch the multinode or single node subprocess 
4. Ensemble: currently running in sequential, single gpu. Rewrite it to support multinode,multigpu as device defined in self.device_setting.

heyufan1995:
Fixes #6191 
fixes #6259 .

### Description
Big changes over autorunner to enable multinode training and multinode-multiGPU ensembler
Multiple changes:
1. Add set_device_info() to create a self.device_dict to define device information (CUDA_VISIBLE_DEVICES, NUM_NODE, e.t.c.) for all parts in autorunner, including data analyzer, trainer, ensembler. No global env variable is set, all device info is from self.device_dict. Changes to bundlegen is made.
2. To enable multi-gpu/multi-node training for ensembler (call from subprocess), we need to separate the ensembler from autorunner (for subprocess to run from autorunner). Created a new EnsembleRunner class (similar to bundleGen), and moved all ensemble related function from autorunner to this class. Local multi-GPU ensembling passed.

Passed some quick local testing. Needs to fix details and do test. Created PR to do a initial design pattern discussion. Slack me if there is any major concern of the change.
@mingxin-zheng @wyli


monai/apps/auto3dseg/auto_runner.py
```diff
@@ -285,17 +277,13 @@ def __init__(
...
-        self.save_image = self.set_image_save_transform(kwargs)
-
-        self.ensemble_method: AlgoEnsemble
-        self.ensemble_method_name: str | None = None
-
+        self.ensemble_method_name = 'AlgoEnsembleBestByFold'
         self.set_num_fold(num_fold=num_fold)
-        self.set_ensemble_method("AlgoEnsembleBestByFold")
-
+        self.kwargs = kwargs
```
mingxin-zheng, line 286: At the first sight, `kwargs` seems too general as a property in the `__init__` function. Usually we may need some logics to process the possible kwargs.

monai/apps/auto3dseg/auto_runner.py
```diff
@@ -569,58 +575,6 @@ def set_nni_search_space(self, search_space):
         self.search_space = search_space
         self.hpo_tasks = value_combinations
 
-    def set_image_save_transform(self, kwargs):
```
mingxin-zheng, line 572: The deletion of the methods:
- `set_image_save_transform`
- `set_ensemble_method`
mean breaking changes.

monai/apps/auto3dseg/auto_runner.py
```diff
@@ -489,6 +470,31 @@ def set_training_params(self, params: dict[str, Any] | None = None) -> None:
         """
         self.train_params = deepcopy(params) if params is not None else {}
 
+    def set_device_info(self, cuda_visible_devices: list[int] | str=os.environ.get('CUDA_VISIBLE_DEVICES', None),
+                        num_nodes: int=os.environ.get('NUM_NODES', 1),
+                        mn_start_method: str=os.environ.get('MN_START_METHOD', 'bcprun')):
+        """ Set the device related info
```
mingxin-zheng, line 476: Would be good to add a line return after the `"""` here.

monai/apps/auto3dseg/auto_runner.py
```diff
@@ -489,6 +470,31 @@ def set_training_params(self, params: dict[str, Any] | None = None) -> None:
...
+        Args:
+            cuda_visible_device: define GPU ids for data analyzer, training, and ensembling. List of GPU ids [0,1,2,3] or a string "0,1,2,3"
```
mingxin-zheng, line 478: Please add the missing `.` in the end of the sentence. Thanks!

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -141,13 +143,12 @@ def export_to_disk(self, output_path: str, algo_name: str, **kwargs: Any) -> Non
             self.fill_records = self.fill_template_config(self.data_stats_files, self.output_path, **kwargs)
         logger.info(self.output_path)
 
-    def _create_cmd(self, train_params=None):
+    def _create_cmd(self, train_params: dict={}):
```
mingxin-zheng, line 146: Better not use an empty dict as the default because of [this](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python)

How about adding one line:
```
train_params = {} if train_params is None else train_params
```

mingxin-zheng:
Hi @heyufan1995 , I have a question that confuses me. What's the benefits expectation of using multi-node to run the ensemble, as the ensemble execution is basically for-loops as below:

```
for file in files:
    for algo in algos:
        pred = algo[some_key].predict(...)

```

monai/apps/auto3dseg/bundle_gen.py
heyufan1995: Thank you. Will change it

heyufan1995:
> Hi @heyufan1995 , I have a question that confuses me quite a bit. What's the benefits expectation of using multi-node to run the ensemble, as the ensemble execution is basically for-loops as below:
> 
> ```
> for file in files:
>     for algo in algos:
>         pred = algo[some_key].predict(...)
> ```

@mingxin-zheng The files are partitioned to all nodes and gpus. So if you have 16 files, 2 nodes 16 GPUs, then each GPU will sequentially run 5 fold models on 1 file. I also think the for loop should be changed to 
> ```
> for algo in algos:
>     for file in files:
>         pred = algo[some_key].predict(...)
> ```
Since loading model weights can be slow.

mingxin-zheng:
Hi @heyufan1995 , thanks for the explanation. Your proposal also make sense. One possible concern I had was that the memory needs to hold (n_algos x n_files) before the ensemble takes place, which was the reason that I went one file by another. We may need to be careful and do something when there are lots of files to infer.

By the way, `infer_instance.predict` supports multiple files in one batch in the `predict_files` argument.

monai/apps/auto3dseg/auto_runner.py
heyufan1995: Yes I agree kwargs is too general. kwargs was originally in __init__, I did not add it. I only added self.kwargs = kwargs because set_image_transform function needs kwargs as input, which is moved to EnsembleRunner.

monai/apps/auto3dseg/bundle_gen.py
heyufan1995: > Hi @heyufan1995 , thanks for the explanation. Your proposal also makes sense. One possible concern I had was that the memory needs to hold (n_algos x n_files) before the ensemble takes place, which was the reason that I went one file by another. We may need to be careful and do something when there are lots of files to infer.
> 
> By the way, `infer_instance.predict` supports multiple files in one batch in the `predict_files` argument.

Thanks. I think we  just keep the current for loop.

heyufan1995: Thanks for pointing out. Will update

heyufan1995:
> Hi @heyufan1995 , thanks for the explanation. Your proposal also makes sense. One possible concern I had was that the memory needs to hold (n_algos x n_files) before the ensemble takes place, which was the reason that I went one file by another. We may need to be careful and do something when there are lots of files to infer.
> 
> By the way, `infer_instance.predict` supports multiple files in one batch in the `predict_files` argument.

Thanks, we can keep the current for loop. It's safer.

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -165,45 +168,62 @@ def _create_cmd(self, train_params=None):
...
+            else:
+                raise NotImplementedError(f"{self.device_setting['MN_START_METHOD']} is not supported yet. \
+                                             Try modify BundleAlgo._create_cmd for your cluster.")
         else:
-            cmd = "python "  # TODO: which system python?
+            # if using single node
+            if self.device_setting['n_devices'] > 1:
+                cmd = f"torchrun --nnodes={1:d} --nproc_per_node={self.device_setting['n_devices']:d} "
+            else:
+                cmd = "python "  # TODO: which system python?
```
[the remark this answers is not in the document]
heyufan1995: @wyli Thanks, what do you suggest? BundleAllgo and EnsembleRunner all have this similar logic to create cmd and run using subprocess. Should I move this function to monai.apps.auto3dseg.utils.py?

monai/apps/auto3dseg/auto_runner.py
mingxin-zheng: Can we keep the `set_image_save_transform` method in AutoRunner and pass the save transform to the `__init__` of `EnsembleRunner`?

The disadvantage of passing `self.kwargs` to another module `EnsembleRunner` : kwargs is a `dict` and the methods in  `EnsembleRunner` can change the values in `kwargs`, which will change `self.kwargs` and make it unsafe when more changes are built into Auto3DSeg.

Here is an example of how bug can be created:

```
class Bar:
    def __init__(self, **kwargs):
        self.kwargs = kwargs
    
    def buggy_method(self):
        self.kwargs["input"]["a"] = 1

class Foo:
    def __init__(self, **kwargs):
        self.kwargs = kwargs
    
    def update(self):
        self.bar = Bar(**self.kwargs)
        self.bar.buggy_method()

c = {"a": 0}
A = Foo(input=c)
print("before: ", c)
A.update()
print("bug occurred: ", c)
```

The output:
```
before:  {'a': 0}
bug occurred:  {'a': 1}
```

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -327,3 +335,225 @@ def get_ensemble(self):
...
+                                             **pred_params)
+            ensemble_runner.run(device_setting)
+
+    """
+    def __init__(self, data_src_cfg_name: str='./work_dir/input.yaml',
+                 work_dir: str='./work_dir',
+                 num_fold: int=5,
+                 ensemble_method_name: str='AlgoEnsembleBestByFold',
+                 mgpu: bool=True,
+                 **kwargs):
```
mingxin-zheng, line 369: We can add an optional argument `transform=None` here, and accept the save transform as one of the post-processing step.

monai/apps/auto3dseg/auto_runner.py
mingxin-zheng: Also I think `set_ensemble_method` is necessary, or otherwise, the user cannot change ensemble method in `AutoRunenr`.

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -175,46 +176,71 @@ def _create_cmd(self, train_params=None):
...
                 base_cmd += f"'{config_yaml}'"
-
-        if "CUDA_VISIBLE_DEVICES" in params:
-            devices = params.pop("CUDA_VISIBLE_DEVICES")
-            n_devices, devices_info = len(devices), ",".join([str(x) for x in devices])
-        else:
-            n_devices, devices_info = torch.cuda.device_count(), ""
-        if n_devices > 1:
-            cmd = f"torchrun --nnodes={1:d} --nproc_per_node={n_devices:d} "
+        cmd = self.device_setting['CMD_PREFIX']
```
mingxin-zheng, line 179: The property `device_setting` is not defined.

heyufan1995: it's defined in self.train. self._create_cmd is only callable by self.train so device_setting should always be defined. But I can move it to init

heyufan1995: This device_setting should be passed by autorunner, since autorunner decides which device bundle algo train on

mingxin-zheng: I think all property used needs to be defined into the __init__ with proper annotations, or otherwise it will fail some format requirements

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -175,46 +176,71 @@ def _create_cmd(self, train_params=None):
...
+                                             str(self.device_setting['n_devices']),'-c', cmd], env=ps_environ, check=True)
+            else:
+                raise NotImplementedError(f"{self.device_setting['MN_START_METHOD']} is not supported yet.\
+                                             Try modify BundleAlgo._run_cmd for your cluster.")
+        else:
+            normal_out = subprocess.run(cmd.split(), env=ps_environ, check=True)
         return normal_out
 
-    def train(self, train_params=None):
+    def train(self, train_params: None | dict = None, device_setting: None | dict = None):
```
mingxin-zheng, line 222: The train function no longer takes "CUDA_VISIBLE_DEVICES" as an effective param in the "train_params", which seems another breaking change.

mingxin-zheng: I would suggest to raise deprecation warnings if user puts "CUDA_VISIBLE_DEVICES" in train_params and let them know it can be written in `device_setting`.

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -175,46 +176,71 @@ def _create_cmd(self, train_params=None):
...
+            device_settings: device related settings, should follow the device_setting in auto_runner.set_device_info.
+            'CUDA_VISIBLE_DEVICES' should be a string e.g. '0,1,2,3'
         """
-        cmd, devices_info = self._create_cmd(train_params)
-        return self._run_cmd(cmd, devices_info)
+        # device_setting set default value and sanity check, in case device_setting not from autorunner
+        self.device_setting = {'CUDA_VISIBLE_DEVICES': ','.join([str(x) for x in range(torch.cuda.device_count())]),
+                               'n_devices': torch.cuda.device_count(), 'NUM_NODES': int(os.environ.get('NUM_NODES', 1)),
+                               'MN_START_METHOD': os.environ.get('MN_START_METHOD', 'bcprun'),
+                               'CMD_PREFIX':os.environ.get('CMD_PREFIX', None),
```
mingxin-zheng, line 236: `CMD_PREFIX` seems a broad concept. I feel confused what may be put there.

monai/apps/auto3dseg/ensemble_builder.py
mingxin-zheng: Hi @heyufan1995 ! Although you mentioned that we can only pass strings, instead of an instance such as SaveImage Transforms, I have a new idea and it is open to discussion.

- Python fire API supports dictionary, e.g. `python example.py {"name":'"David Bieber"'}`
- MONAI ConfigParser can resolve a dictionary and instantiate a class

So it may be possible combine both, to construct a dictionary in `set_image_save_transform`, and pass that dictionary to EnsembleRunner. The instantiation of SaveImage Transform can be inside the EnsembleRunner as you want, and we don't need to discontinue the API of `set_image_save_transform`.

monai/apps/auto3dseg/bundle_gen.py
mingxin-zheng: After some thoughts, I think it is reasonable to drop the support `CUDA_VISIBLE_DEVICES` in `train_params`, because we don't want two different ways to change the GPU device settings. But it would be good to raise warning here to let user know it is no longer supported.

monai/apps/auto3dseg/bundle_gen.py
heyufan1995: changed to init

monai/apps/auto3dseg/bundle_gen.py
heyufan1995: wenqi suggested user defined runner
export CMD_PREFIX="python "
export CMD_PREFIX="/opt/conda/bin/python3.8 "
export CMD_PREFIX="torchrun --nnodes=3 --nproc_per_node=2 "

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -375,9 +401,7 @@ class BundleGen(AlgoGen):
             zip url will be downloaded and extracted into the algo_path. The current default options are released at:
             https://github.com/Project-MONAI/research-contributions/tree/main/auto3dseg
         data_stats_filename: the path to the data stats file (generated by DataAnalyzer)
-        data_src_cfg_name: the path to the data source config YAML file. The config will be in a form of
-            {"modality": "ct", "datalist": "path_to_json_datalist", "dataroot": "path_dir_data"}
-
+        data_src_cfg_name: the path to the data kwargs
```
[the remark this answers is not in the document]
heyufan1995: updated

monai/apps/auto3dseg/bundle_gen.py
heyufan1995: I added this to set_train_params:
```
if 'CUDA_VISIBLE_DEVICES' is self.train_params.keys():
        warnings.warn("CUDA_VISIBLE_DEVICES definition is deprecated from `set_training_params`. Use `set_device_info` intead.", warnings.DeprecationWarning)
```

monai/apps/auto3dseg/bundle_gen.py
mingxin-zheng: The examples are helpful for me. Maybe we can put that in the doc-string?

monai/apps/auto3dseg/auto_runner.py
```diff
@@ -488,6 +471,77 @@ def set_training_params(self, params: dict[str, Any] | None = None) -> None:
 
         """
         self.train_params = deepcopy(params) if params is not None else {}
+        if 'CUDA_VISIBLE_DEVICES' == self.train_params.keys():
```
mingxin-zheng, line 474:
```
if "CUDA_VISIBLE_DEVICES" in self.train_params:
```

mingxin-zheng:
Thanks @heyufan1995 for the PR. The design looks good to me as we minimize the number of breaking changes. There are some format issues and test failures to fix. CC @wyli for viz.

monai/apps/auto3dseg/ensemble_builder.py
mingxin-zheng: Feel free to close this one as we addressed the API breaking change.

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -155,16 +164,13 @@ def __call__(self, pred_param: dict[str, Any] | None = None) -> list[torch.Tenso
...
-            print(i)
+        for i, file in enumerate(tqdm(files, desc='Ensembling (rank 0)...')) if pred_param.get('rank',0)==0 else enumerate(files):
             preds = []
             for algo in self.algo_ensemble:
                 infer_instance = algo[AlgoKeys.ALGO]
                 pred = infer_instance.predict(predict_files=[file], predict_params=param)
                 preds.append(pred[0])
-            outputs.append(self.ensemble_pred(preds, sigmoid=sigmoid))
-        return outputs
+            pred_param['image_save_func'](self.ensemble_pred(preds, sigmoid=sigmoid))
```
wyli, line 173: what happens when `pred_param` is None`?

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -168,53 +177,86 @@ def _create_cmd(self, train_params=None):
...
         else:
-            cmd = "python "  # TODO: which system python?
+            if cmd is None:
+                cmd = f"torchrun --nnodes={1:d} --nproc_per_node={self.device_setting['n_devices']:d} "
         cmd += base_cmd
         if params and isinstance(params, Mapping):
             for k, v in params.items():
                 cmd += f" --{k}={v}"
-        return cmd, devices_info
+        return cmd, ""
```
wyli, line 205: kept this in order to be compatible with the existing algos  https://github.com/Project-MONAI/research-contributions/blob/8b10ad5b04d60835e1a2bd16a49abc378045b3f8/auto3dseg/algorithm_templates/dints/scripts/algo.py#L426-L428

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -327,3 +339,244 @@ def get_ensemble(self):
...
+            os.makedirs(output_dir)
+            if self.rank == 0:
+                logger.info(f"Directory {output_dir} is created to save ensemble predictions")
+
+        self.output_dir = output_dir
+        output_postfix = kwargs.pop("output_postfix", "ensemble")
+        output_dtype = kwargs.pop("output_dtype", "$np.uint8")
+        resample = kwargs.pop("resample", False)
+
+        self.save_image = {
```
wyli, line 444: cannot use SaveImage here because it's not compatible with `yaml.dump(config)` https://github.com/Project-MONAI/research-contributions/blob/8b10ad5b04d60835e1a2bd16a49abc378045b3f8/auto3dseg/algorithm_templates/segresnet/scripts/segmenter.py#L523

wyli (approved):
integration verified https://github.com/Project-MONAI/MONAI/actions/runs/4699028053/jobs/8332016913 I'm merging this to unblock benchmarking tasks.

PR #6281, opened 2023-04-03
Remove usage of deprecated `AsChannelFirst`, `AddChannel `, and `SplitChannel`

### LINKED ISSUE #6280: Ensure usage of anything deprecated since 0.8 is removed
There are several components that have been deprecated since v0.8 and many of them has been removed including all pathology related classes (addressed in PR #6262); however, there are still classes that have been used vastly throughout test cases, tutorials and bundles, so removing them requires some coordination with other monai repos and modifying the code that are utilizing them before removing them.

[UPDATE] The usage of below classes will be removed but they will be kept as a wrapper with an associated remove version of 1.3.

- [ ] AsChannelFirst and AsChannelFirstd
- [ ] AddChannel and AddChanneld
- [ ] SplitChannel and SplitChannels

drbeh:
Fixes #6280 

### Description
This PR removes the usage of `AsChannelFirst`, `AddChannel`, and `SplitChannel` (which are deprecated since v0.8) and add removed version to v1.3.

### Types of changes


Nic-Ma:
Hi @ericspod @drbeh ,

May I know the latest status and plan for this PR?

Thanks.

ericspod:
> Hi @ericspod @drbeh ,
> 
> May I know the latest status and plan for this PR?
> 
> Thanks.

I think @drbeh still wanted to work on this one to get the other old classes removed. I'm fine removing them so long as they are all gone from our tutorials and bootcamp materials. We could keep them as wrappers around the correct classes to use and set a deletion version as 1.3 if people aren't sure about removing now.

drbeh:
> > Hi @ericspod @drbeh ,
> > May I know the latest status and plan for this PR?
> > Thanks.
> 
> I think @drbeh still wanted to work on this one to get the other old classes removed. I'm fine removing them so long as they are all gone from our tutorials and bootcamp materials. We could keep them as wrappers around the correct classes to use and set a deletion version as 1.3 if people aren't sure about removing now.

That's right! We should keep them as aliases but my question is do we really need to deprecate these aliases? Since they are aliases, the maintenance cost is very low. I am just wondering if we want to keep them to avoid unnecessary breaking changes. What do you think? @ericspod @Nic-Ma @wyli

ericspod:
> we really need to deprecate these aliases?

I'd say we should to push people towards the better practice of using the real transforms and not these aliases. Once we've had these in place for a version of two we can remove them safely, they are lightweight but less stagnant code is still better.

ericspod:
Are we removing the classes entirely and no aliases?

drbeh:
> Are we removing the classes entirely and no aliases?

@ericspod I am a bit unsure about the best way to do that. Can you help to find the best way to move forward?
Here are three options:
1. remove them as they have already deprecated since 0.8, which follows the existing monai guideline.
2. create an alias for `AsChannelFirst` to `EnsureChannelFirst`, which would easily work and I have updated the PR with this.
3. make `AsChannelFirst` a wrapper around `EnsureChannelFirst` and set another deprecation version for it, which is fine too but seems to be a bit arbitrary.

drbeh:
> Are we removing the classes entirely and no aliases?

@ericspod I am a bit unsure about the best way to do that. Can you help to find the best way to move forward?
Here are three options:
1. remove them entirely since they have already deprecated since 0.8, which follows the existing MONAI guideline.
2. create an alias as `AsChannelFirst = EnsureChannelFirst`, which would work but we should add it to the documentation too.
3. make `AsChannelFirst` a wrapper around `EnsureChannelFirst` and set another deprecation version for it, which is fine too but seems to be a bit arbitrary what version to put and also we should keep CI/CD for this wrapper.

ericspod:
Thanks @drbeh, I'd say option 3 is the solution since we deprecated the class but didn't set a removal version. We can create a wrapper and set the removal for 1.3, meanwhile we can remove the documentation and tests as you've done and rely on the tests for `EnsureChannelFirst`. 

If instead no one's worried about people using these classes we can go with option 1.

drbeh:
@ericspod I made aliases (wrapper classes) for these deprecated classes, set `removed` to `v1.3`, and replaced all their usage in the test cases with the correct classes. However, I kept their specific unit tests to make sure that they continue working before removing them. It will give some time to the users to consider replacing these transforms and also to us to clean up their usage in the tutorials, bundles, and MONAILabel. Please take another look and let me know if you have any other comment. Thanks

ericspod (approved):
Looks good to me. Thanks! We should see if anyone else has any comments.

drbeh:
@wyli @Nic-Ma do you have any comment here?

wyli (approved):

Nic-Ma (approved):

Issue #6177, opened 2023-03-20
Add Text Encoder with CLIP Embedding


tangy5:
add text encoder with CLIP Embedding

PR #6282, opened 2023-04-04
Add text to vision embedding

tangy5:
As part of the text to vision encoder for medical image analysis.

Support CLIP pre-trained embedding and random text embedding. 

Linked to the issue: https://github.com/Project-MONAI/MONAI/issues/6177

### DIFF
--- a/monai/networks/blocks/text_enbedding.py
+++ b/monai/networks/blocks/text_enbedding.py
@@ -1,0 +1,60 @@
+# Copyright (c) MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import torch
+import torch.nn as nn
+
+class TextEncoder(nn.Module):
+    """
+    Text to vision encoding by Contrastive Language-Image Pre-training (CLIP) or random embedding.
+    The text to vision encoder loads the pre-trained or random initialized weights with connection to 2D/3D vision models.
+
+    Contrastive Language-Image Pre-training (CLIP), based on: "Radford et al.,
+    Learning Transferable Visual Models From Natural Language Supervision <https://arxiv.org/abs/2103.00020>"
+
+    Connecting text and medical 3D image, based on: "Liu et al.,
+    CLIP-Driven Universal Model for Organ Segmentation and Tumor Detection <https://arxiv.org/pdf/2301.00785.pdf>"
+    """    
+    def __init__(
+        self,
+        out_channels: int,
+        text_dim: int = 512,
+        hidden_size: int = 256,
+        encoding: str = "clip_embedding",
+    ) -> None:
+        """
+        Args:
+            out_channels: number of output channels, to control text-baesd embedding for classes.
+            text_dim: dimension of text embeddings.
+            hidden_size: dimension of hidden features, compatible to different vision feature dimensions.
+            encoding: the text embedding type, default to use clip text pretrained weights
+        """
+        self.encoding = encoding
+
+        if self.encoding == 'rand_embedding':
+            self.organ_embedding = nn.Embedding(out_channels, hidden_size)
+        elif self.encoding == 'clip_embedding':
+            self.register_buffer('text_embedding', torch.randn(out_channels, text_dim))
+            self.text_to_vision = nn.Linear(text_dim, hidden_size)
+
+    def forward(self):
+        if self.encoding == 'clip_embedding':
+            task_encoding = nn.function.relu(self.text_to_vision(self.organ_embedding))
+            task_encoding = task_encoding.unsqueeze(2).unsqueeze(2).unsqueeze(2)
+        else:
+            # text embedding as random initialized 'rand_embedding'
+            task_encoding = self.text_embedding.weight.unsqueeze(2).unsqueeze(2).unsqueeze(2)
+        return task_encoding
+
+
+


tangy5:
Hi @wyli , it would be great you can take a look at this PR when you get time. The PR is ready for review, but I struggled to get the flake8 check pass. It seems not this new script's format issue. Can you help to give a look?

Thank you so much, let me know if there are any comments.

wyli:
thanks @tangy5, would be great to first have a prototype of https://github.com/Project-MONAI/MONAI/milestone/98 with all the basic functionality (it's fine to not have detailed docstrings/unit tests), then we identify possible reusable modules and create smaller PRs. e.g it's interesting to see the module's API is `def forward(self)` which doesn't take any input additional input, with a prototype it'll be easier to see how this could be used as part of a larger network.

monai/networks/blocks/text_embedding.py
```diff
@@ -0,0 +1,59 @@
...
+            out_channels: number of output channels, to control text-baesd embedding for classes.
+            text_dim: dimension of text embeddings.
+            hidden_size: dimension of hidden features, compatible to different vision feature dimensions.
+            encoding: the text embedding type, default to use clip text pretrained weights
+        """
+        super().__init__()
+        self.encoding = encoding
+
+        if self.encoding == 'rand_embedding':
+            self.organ_embedding = nn.Embedding(out_channels, hidden_size)
```
wyli, line 47:
```suggestion
            self.text_embedding = nn.Embedding(out_channels, hidden_size)
```

tangy5:
> thanks @tangy5, would be great to first have a prototype of https://github.com/Project-MONAI/MONAI/milestone/98 with all the basic functionality (it's fine to not have detailed docstrings/unit tests), then we identify possible reusable modules and create smaller PRs. e.g it's interesting to see the module's API is `def forward(self)` which doesn't take any input additional input, with a prototype it'll be easier to see how this could be used as part of a larger network.

Great, how can I create this "prototype"? A unit test workflow? Basically, the two PRs should be together, this PR's (https://github.com/Project-MONAI/MONAI/pull/6283) network uses this PR's module. 

I plan to make these two classes reusable for most network backbones, e.g. unet, swinunetr, that if users want to use the "text embedding" for their network, it could be safely concatenated to the vision feature predicted by CNN/Transformers backbones.

I feel a complete unit test or an integration test would be better to show the prototype here? 

Thank you.

monai/networks/blocks/text_embedding.py
tangy5: Thanks, updated.

tangy5:
Hi @wyli , thank you for the suggestion. For the prototype thing you mentioned, a complete workflow and application is here: https://github.com/ljwztc/CLIP-Driven-Universal-Model

I'd like to make another pipeline for a partially supervised learning workflow later. 

I'm thinking this as two parts:

1: The text_embedding class that can load pre-trained or to add any text embedding to any network modules. 
2. A dynamic segmentor class that can be used (reusable) as a separate module attached to segmentation network backbones (e.g., SwinUNETR, UNET, segResNet, etc)

These two modules are reusable designs. 
Let me know if we need a discussion. Thank you.

tests/test_text_encoding.py
```diff
@@ -0,0 +1,43 @@
...
+import torch
+from monai.networks.blocks.text_embedding import TextEncoder
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
+
+
+class TestTextEncoder(unittest.TestCase):
+    def test_test_encoding_shape(self):
+        # test 2D encoder
+        text_encoder = TextEncoder(spatial_dims=2, out_channels=32, pretrained=True).to(device)
```
wyli, line 25: please add the context manager to avoid reporting downloading issues, like this https://github.com/Project-MONAI/MONAI/blob/1a55ba5423d04d2ef7ac19356ccabc4c7906f577/tests/test_flexible_unet.py#L371

and remove the `print` commands as they are not needed. Then I can test and merge this soon, thanks!

tangy5: Thanks, removed.

tangy5:
@wyli , thank you so much for the suggestions, it's very helpful. 

Some changes according to your suggestions have been made:
1. add pretrained option
2. add url_map for released text embedding
3. add 2D embedding support
4. add test cases 3D, 2D.

could you have a look if the `skip_if_downloading_fails` is used the correct way..

Thanks.

wyli:
thanks @tangy5, looks good to me, there are issues with the CPU only tests, could you please revise? (@yiheng-wang-nv knows more if you need help)

monai/networks/blocks/text_embedding.py
```diff
@@ -0,0 +1,85 @@
...
+            test_encoding = self.text_embedding.weight
+        else:
+            test_encoding = nn.functional.relu(self.text_to_vision(self.text_embedding))
+
+        if self.spatial_dims == 3:
+            test_encoding = test_encoding.unsqueeze(2).unsqueeze(2).unsqueeze(2)
+        elif self.spatial_dims == 2:
+            test_encoding = test_encoding.unsqueeze(2).unsqueeze(2)
+
+        return test_encoding
```
wyli, line 85: is this a typo in the variable name? `test_encoding` -> `text_encoding`?

monai/networks/blocks/text_embedding.py
```diff
@@ -0,0 +1,85 @@
...
+    CLIP-Driven Universal Model for Organ Segmentation and Tumor Detection <https://arxiv.org/pdf/2301.00785.pdf>"
+    """
+    def __init__(
+        self,
+        out_channels: int,
+        spatial_dims: int = 3,
+        text_dim: int = 512,
+        hidden_size: int = 256,
+        encoding: str = "rand_embedding",
+        pretrained: bool = False
```
wyli, line 41: I think most users will prefer `pretrained=True` and `encoding="clip_embedding"`? and when `pretrained=True`, hiddne_size and text_dim are not tunable, right?

tangy5:
> thanks @tangy5, looks good to me, there are issues with the CPU only tests, could you please revise? (@yiheng-wang-nv knows more if you need help)

Thanks. The test fails should because the loaded pre-trained weights are mapped to GPU, set to "CPU". 
Other changes are modified according to your comments, thanks. 
I set the `pretrained` to `True` by default and to load the release clip embedding.  In the future, if there are more released weights, users can load it, and modify `text_dim` accordingly.

wyli (approved):

PR #6283, opened 2023-04-04
Add text driven segmentor for promptable outputs

tangy5:
This PR adds a head controller and Universal model for promptable outputs. 

Users can choose and control the number and name of output channels from a multi-class segmentation model. This can enable incremental learning by adding new classes to an existing pre-trained model without catastrophic forgetting.


tangy5:
Thanks. I will reformat this PR, and we might need further discussion.

PR #6284, opened 2023-04-04
ClassesToIndicesd limit the cache memory

myron:
to solve: https://github.com/Project-MONAI/MONAI/issues/6285

Currently ClassesToIndicesd can be used to pre-compute and cache the locations of all background/foreground classes. 
This significantly speedups the cropping transform RandCropByLabelClassesd, since the coordinates of class voxels do not need to be computed every time. 

Unfortunately for an average dataset it requires ~80gb of extra RAM, with an average cache (per image) is on the order of image size itself. (most of it due to background coordinates, or by some large segmentation classes). 

This PR clips (optionally) the number of coordinates in each class to a parameter  max_indices_per_class. (e.g. 5000), which greatly reduces cache size requirements only to a few MB.  Generally these coordinates are used to random choose a cropping center at each epoch, e.g. for 1000 epochs -> 1000 centers per image, so we don't need to store ALL the foreground coordinates, it's sufficient to store just some large random subset. 

PS: I wasn't sure if it's better to also change ClassesToIndices (not ClassesToIndicesD), feel free edit this PR. 

thank you

### Description

### Types of changes


monai/transforms/utility/dictionary.py
```diff
@@ -1320,18 +1321,28 @@ def __init__(
...
         self.image_key = image_key
+        self.max_indices_per_class = max_indices_per_class
         self.converter = ClassesToIndices(num_classes, image_threshold, output_shape)
 
     def __call__(self, data: Mapping[Hashable, Any]):
         d = dict(data)
         image = d[self.image_key] if self.image_key else None
         for key in self.key_iterator(d):
-            d[str(key) + self.indices_postfix] = self.converter(d[key], image)
+            out = self.converter(d[key], image)
```
wyli, line 1337: if it's feasible in general, perhaps we consider subsampling when generating the indices?

https://github.com/Project-MONAI/MONAI/blob/9e33ff25f34ebd11e026c49634743d9389564765/monai/transforms/utils.py#L332-L341

myron: I'm open to any option, I just need to limit the amount of cache here.  feel free to edit, or how would I like me to proceed? thanks

wyli: sure I'll try to update `map_classes_to_indices` for the same purpose.

myron: thank you!

monai/transforms/utils.py
```diff
@@ -352,28 +353,38 @@ def map_classes_to_indices(
...
+        else:
+            label_flat = ravel(label == c)
+        if img_flat is not None:
+            label_flat = img_flat & label_flat
         # no need to save the indices in GPU, otherwise, still need to move to CPU at runtime when crop by indices
         cls_indices: NdarrayOrTensor = convert_data_type(nonzero(label_flat), device=torch.device("cpu"))[0]
-        indices.append(cls_indices)
+        if max_samples_per_class and len(cls_indices) > max_samples_per_class:
+            sample_id = np.round(np.linspace(0, len(cls_indices) - 1, max_samples_per_class)).astype(int)
+            indices.append(cls_indices[sample_id])
```
wyli, line 385: I updated this utility to choose evenly spaced samples @myron please let me know what you think, I couldn't find any benefit of randomizing but I might miss something here

wyli (approved):
updated based on your comments @myron, please have a look, let's merge this soon.

myron:
looks great! let's merge. thank you

PR #6290, opened 2023-04-05
Auto3DSeg skip trained algos

myron:
Issues: https://github.com/Project-MONAI/MONAI/issues/6291

Allows to skip the already trained algos, and continue training only for the non-trained ones. 

after this PR, the default option AutoRunner(train=None) will have this behavior, whereas manually setting AutoRunner(train=True/False) will always train all or skip all training.  Previously we can only train all or skip all (without any option to resume)


mingxin-zheng (approved):

mingxin-zheng:
Hi @myron , thanks for the PR. Please run `./runtests --autofix` for the format fix, and other looks good to me.

myron:
I ran it before, and I just ran it again, it sees to issues on my machine (macos)

black-fix
python -m black, 22.6.0 (compiled: no)
Python (CPython) 3.9.7
All done! ✨ 🍰 ✨
1065 files left unchanged.
passed!

where can i see what the issue is here on github?

monai/apps/auto3dseg/auto_runner.py
```diff
@@ -758,7 +758,8 @@ def run(self):
             logger.info("Skipping algorithm generation...")
 
         # step 3: algo training
-        if self.train:
+        auto_train_choice = self.train is None
+        if self.train == True or (auto_train_choice and not self.cache["train"]):
```
wyli, line 762:
```suggestion
        if self.train or (auto_train_choice and not self.cache["train"]):
```

PR #6295, opened 2023-04-05
Revert "Auto3DSeg skip trained algos"

wyli:
Reverts Project-MONAI/MONAI#6290
fixes https://github.com/Project-MONAI/MONAI/issues/6294

this commit is not compatible with the integration tests

PR #6310, opened 2023-04-05
Auto3DSeg continue training (skip trained algos)

### LINKED ISSUE #6291: Auto3DSeg skip trained algos
Allows to skip the already trained algos, and continue training only for the non-trained ones.

myron:
Second PR for issue https://github.com/Project-MONAI/MONAI/issues/6291
fixes #6291

Since the previous PR https://github.com/Project-MONAI/MONAI/pull/6290 was reverted https://github.com/Project-MONAI/MONAI/pull/6295

Allows to skip the already trained algos, and continue training only for the non-trained ones.

after this PR, the default option AutoRunner(train=None) will have this behavior, whereas manually setting AutoRunner(train=True/False) will always train all or skip all training. Previously we can only train all or skip all (without any option to resume)

I changed  import_bundle_algo_history() to return a better algo_dict 

previously it returned "list[dict(name: algo)]" - a list of dict, but each dict must have a single key name "name => algo". Not it returns a list of dicts, each with several keys dict(AlgoEnsembleKeys.ID: name, AlgoEnsembleKeys.ALGO, algo, "is_trained": bool, etc).
this allows to put additional metadata inside of each algo_dict, and it's easier to read it back.

previously, to get a name we had to use "name = history[0].keys()[0]", now it's more elegant "name = history[0][AlgoEnsembleKeys.ID]".

this however required to change many files, everywhere where import_bundle_algo_history and export_bundle_algo_history was used.

All the tests have passed, except for "integration GPU utilization tests" , but those errors seems unrelated

After this PR, tutorials need to be updated too https://github.com/Project-MONAI/tutorials/pull/1288


monai/apps/auto3dseg/utils.py
```diff
@@ -63,6 +72,6 @@ def export_bundle_algo_history(history: list[dict[str, BundleAlgo]]) -> None:
     Args:
         history: a List of Bundle. Typically, the history can be obtained from BundleGen get_history method
     """
-    for task in history:
-        for _, algo in task.items():
-            algo_to_pickle(algo, template_path=algo.template_path)
+    for algo_dict in history:
+        algo = algo_dict[AlgoEnsembleKeys.ALGO]
```
mingxin-zheng, line 76: I guess after this PR, the user will see this enum "AlgoEnsembleKeys" even if they don't do ensemble, because that's the way to iterate over the Auto3DSeg-templated algorithms. 
Originally the keys are for ensemble but the scope will be expanded. Maybe we can consider another name, such as "Auto3DSegAlgoKeys" and deprecate `AlgoEnsembleKeys` to avoid confusion.
The deprecation doesn't have to be in this PR, and I can submit a PR for that after this one is merged.

monai/apps/auto3dseg/auto_runner.py
```diff
@@ -767,20 +771,34 @@ def run(self):
                     "Possibly the required algorithms generation step was not completed."
                 )
 
-            if not self.hpo:
-                self._train_algo_in_sequence(history)
-            else:
-                self._train_algo_in_nni(history)
+            if auto_train_choice:
+                history = [h for h in history if not h["is_trained"]]  # skip trained
```
mingxin-zheng, line 775: Shall we have a logging at the debug level for the algorithms picked for training, or the algorithms skipped because `is_trained` is true?

monai/apps/auto3dseg/auto_runner.py
```diff
@@ -767,20 +771,34 @@ def run(self):
...
         # step 4: model ensemble and write the prediction to disks.
         if self.ensemble:
-            history = import_bundle_algo_history(self.work_dir, only_trained=True)
+            history = import_bundle_algo_history(self.work_dir, only_trained=False)
+
+            history_untrained = [h for h in history if not h["is_trained"]]
+            if len(history_untrained) > 0:
+                warnings.warn(
+                    f"Ensembling step will skip {[h['name'] for h in history_untrained]} untrained algos"
+                    "Generally it means these algos did not complete training"
```
mingxin-zheng, line 795: Would be good to add a period `.` in the end for both sentences. Thanks!

monai/apps/auto3dseg/utils.py
myron: sure, may be something even shorter e.g. renaming AlgoEnsembleKeys -> AlgoKeys (after this PR)

monai/apps/auto3dseg/auto_runner.py
myron: added

monai/apps/auto3dseg/auto_runner.py
myron: added logging

mingxin-zheng (approved):
Looks good to me. Thanks!

wyli (approved):
thanks, verified that the integration tests work fine with this PR

Issue #6314, opened 2023-04-06
Make spatial transform inverse compatible for multiple passes


liamchalcroft:
Using SpacingD as a pre-processing transform in a MONAILabel app, and the Deepedit click transforms often require multiple iterations. The default inverse transform runs into 2 errors when trying to inverse the SpacingD transform for the second time:

1. 'TypeError: Cannot interpret 'torch.float64' as a data type' from 'get_numpy_dtype_from_string' in 'monai/utils/type_conversion.py' . I am able to fix this locally by changing the function to remove any numpy/torch prefix: 'dtype=str(dtype).split('.')[-1]'
2. ' KeyError: 'src_affine' ' from 'monai/transforms/spatial/array.py", line 226, in inverse: kw_args["dst_affine"] = kw_args.pop("src_affine")' . I am able to fix this locally by adding this line before the inverse fn return: 'kw_args["src_affine"] = kw_args["dst_affine"]' in order to restore the src_affine.

wyli:
thanks for reporting these, if you plan to contribute, 
for 1, please add tests to https://github.com/Project-MONAI/MONAI/blob/06defb7644585c462aba126325dd1fdda5d98db6/tests/test_get_equivalent_dtype.py#L45-L50

for 2, please change it to `kw_args["src_affine"] = kw_wargs.get("dst_affine")` to allow for empty `src_affine` before calling the inverse fn

PR #6315, opened 2023-04-06
Fix MONAILabel-related bugs in dtype conversion and inverse of Spacing

liamchalcroft:
Fixes issues relating to dtype and src_affine in inverse of spatial transforms. Issue found to occur when inverse for a given function is required many times, e.g. for iterative updates in interactive segmentation.  Related issue Project-MONAI/MONAI#6314


wyli (approved):
thanks, looks good to me, I'll add the unit tests before merging into dev

Issue #6361, opened 2023-04-14
Auto3DSeg ensemble needs to take images from different devices


mingxin-zheng:
**Describe the bug**
Referece link:
https://github.com/Project-MONAI/research-contributions/pull/211?notification_referrer_id=NT_kwDOARtBabM1OTU2NDAwNjk5OjE4NTYzNDMz#issuecomment-1507906702

Different algorithms may return results on different devices (CPU or GPU) , e.g. if algo didn't have enough gpu memory it may do CPU based slidingwindow inference. So than when we do Ensembling, some internal function concatinate the predictions and it fails, since the predictions are on different devices.

wyli:
Addressed by #6368

PR #6368, opened 2023-04-14
Ensure ensemble preds are on the same device

myron:
Fixes # https://github.com/Project-MONAI/MONAI/issues/6361

Ensures ensemble preds are on the same device


mingxin-zheng (approved):
Thanks!

PR #6401, opened 2023-04-19
Remove deprecated arg in torchvision models

drbeh:
### Description

For torchvisno models, the parameter `pretrained` is deprecated since 0.13 and may be removed in the future. Our `TorchVisionFC` model has been using this parameter. This PR updated `TorchVisionFC` to remove calling pretrained and instead to call `weights` with appropriate defaults without any change in the functionality (backward compatible).

### Types of changes


Nic-Ma (approved):
Looks good to me.
Please ensure it's compatible with PyTorch 1.7 - 2.0.

Thanks.

PR #6437, opened 2023-04-26
Fix for zoom used in a lazy pipeline

atbenmurray:
If you run the following pipeline with lazy resampling it breaks as the pending info contains a 3x3 affine transform and there are 4x4 affine transforms already in the list:

```
[Flip(0), Flip(1), Rotate90(1), Zoom(0.8), NormalizeIntensity()]
```

Although it could be fixed by ensuring that Zoom always puts out a 4x4 affine, the more flexible and pragmatic solution is to promote 2d affine transforms to 3d affine transforms as a matter of course. This way we never have the issue that a given transform (particularly potential 3rd party transforms modified for lazy execution) can subsequently cause this problem.

### Description

### Types of changes

### DIFF
--- a/monai/transforms/lazy/functional.py
+++ b/monai/transforms/lazy/functional.py
@@ -15,6 +15,7 @@ from typing import Any
 
 import torch
 
+from monai.data.utils import to_affine_nd
 from monai.data.meta_tensor import MetaTensor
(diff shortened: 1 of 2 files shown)


atbenmurray:
Incidental bug-fix noticed while prepping #6257

PR #6464, opened 2023-05-04
6463 Optimized the `deepcopy` logic in `ConfigParser`

### LINKED ISSUE #6463: `ConfigParser` doesn't need `deepcopy`
In the current implementation, `ConfigParser` executes `deepcopy` for every config item when recursively parse the whole config:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/config_parser.py#L379
But actually, all the parsing logic just read the config item content and generate a new config object:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/reference_resolver.py#L295
So I would suggest to remove this `deepcopy` logic, it can bring 3 benefits:
1. If use python object in the config dict, `deepcopy` it may cause `Invertd` failure due to mismatch object ID.
2. If the config structure contains very big python object (like CacheDataset), it will increase memory usage, especially considering it's a recursive operation.
3. Some python objects may not support pickle operation and deepcopy will fail.

Nic-Ma:
Fixes #6463 .

### Description

This PR optimized the deepcopy logic in `ConfigParser` to avoid the problems in #6463 .

### Types of changes


wyli (approved):
works fine with the premerge tests, let's merge to have more tests...

PR #6467, opened 2023-05-04
6439 adds a compose `mode` for lazy evaluation 

### LINKED ISSUE #6439: Lazy resampling: MetaTensor.applied_operations ends up out of order
**Describe the bug**
When executing pipelines with lazy resampling, the applied_operations list on a MetaTensor instance can end up out of order.

**To Reproduce**
1. Run tests/test_integration_lazy_samples.py
2. Add a breakpoint at https://github.com/Project-MONAI/MONAI/blob/4f291723c4ce1a1baf7efbd499495b777dfb1869/tests/test_integration_lazy_samples.py#L155
3. Examine the state of batch_data and you'll see that the applied operations are out of order. Lambdad is the last transform in the pipeline, but appears as the first element in the pipeline

Correct order:
```
[SpatialResample, Orientation, RandRotate90d, RandCropByPosNegLabel, RandRotated, RandZoomd, ResizeWithPadOrCrop, Rotate, Lambdad]
```
Observed order:
```
[Lambdad, SpatialResample, Orientation, RandRotate90d, RandCropByPosNegLabel, RandRotated, RandZoomd, ResizeWithPadOrCrop, Rotate]
```

Note that this issue is fixed in #6257

wyli:
Fixes #6439

### Description
adds a `mode="strict"` to `Compose` so that only consecutive lazy transforms are combined together for one-step resampling (when `lazy_evaluation=True`).

### Types of changes

### DIFF
--- a/monai/transforms/compose.py
+++ b/monai/transforms/compose.py
@@ -51,6 +51,7 @@ def evaluate_with_overrides(
     overrides: dict | None = None,
     override_keys: Sequence[str] | None = None,
     verbose: bool = False,
+    mode: str = "default",
 ):
     """
     The previously applied transform may have been lazily applied to MetaTensor `data` and
@@ -73,16 +74,19 @@ def evaluate_with_overrides(
         override: keyword arguments to apply transforms.
         override_keys: to which the override arguments are used when apply transforms.
         verbose: whether to print debugging info when evaluate MetaTensor with pending operations.
+        mode: "default" or "strict".
 
     """
     if not lazy_evaluation:
         return data  # eager evaluation
     overrides = (overrides or {}).copy()
     if isinstance(data, monai.data.MetaTensor):
+        if isinstance(upcoming, mt.MapTransform) and override_keys not in upcoming.keys:
+            return data
         if data.has_pending_operations and (
             (upcoming is None)
-            or (isinstance(upcoming, mt.Identity))
-            or (isinstance(upcoming, mt.Identityd) and override_keys in upcoming.keys)
+            or (isinstance(upcoming, (mt.Identity, mt.Identityd)))
+            or (mode == "strict" and not isinstance(upcoming, mt.LazyTransform))
         ):
             data, _ = mt.apply_transforms(data, None, overrides=overrides)
             if verbose:
@@ -109,12 +113,14 @@ def evaluate_with_overrides(
         for k in data:
             if k in keys_to_override:
                 dict_for_key = dict_overrides[override_keys.index(k)]
-                data[k] = evaluate_with_overrides(data[k], upcoming, lazy_evaluation, dict_for_key, k, verbose)
+                data[k] = evaluate_with_overrides(data[k], upcoming, lazy_evaluation, dict_for_key, k, verbose, mode)
             else:
-                data[k] = evaluate_with_overrides(data[k], upcoming, lazy_evaluation, None, k, verbose)
+                data[k] = evaluate_with_overrides(data[k], upcoming, lazy_evaluation, None, k, verbose, mode)
 
     if isinstance(data, (list, tuple)):
-        return [evaluate_with_overrides(v, upcoming, lazy_evaluation, overrides, override_keys, verbose) for v in data]
+        return [
+            evaluate_with_overrides(v, upcoming, lazy_evaluation, overrides, override_keys, verbose, mode) for v in data
+        ]
     return data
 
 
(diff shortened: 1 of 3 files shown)


monai/transforms/compose.py
```diff
@@ -73,16 +74,19 @@ def evaluate_with_overrides(
...
     overrides = (overrides or {}).copy()
     if isinstance(data, monai.data.MetaTensor):
+        if isinstance(upcoming, mt.MapTransform) and override_keys not in upcoming.keys:
+            return data
         if data.has_pending_operations and (
             (upcoming is None)
-            or (isinstance(upcoming, mt.Identity))
-            or (isinstance(upcoming, mt.Identityd) and override_keys in upcoming.keys)
+            or (isinstance(upcoming, (mt.Identity, mt.Identityd)))
+            or (mode == "strict" and not isinstance(upcoming, mt.LazyTransform))
```
atbenmurray, line 89: This change fixes precisely one aspect of the issues raised. `mode` has no consideration of extensibility. "strict" should be the default mode; the other mode should be called "benchmark" or "relaxed". There is no discussion in any documentation about the modes and their execution behaviour. This function is still far too complex and only maintainable by the person who wrote it. The Compose API still has all of the issues raised in #6257.

atbenmurray (changes requested):
I suggest that you continue to engage with the process of accepting #PR6257 instead of presenting a separate PR which is missing nearly all of the benefits of it.

monai/transforms/compose.py
```diff
@@ -338,6 +350,7 @@ def __init__(
         overrides: dict | None = None,
         override_keys: Sequence[str] | None = None,
         verbose: bool = False,
```
atbenmurray, line 352: This should be replaced with a logger_name to be consistent with the rest of MONAI

monai/transforms/compose.py
```diff
@@ -326,6 +335,9 @@ class Compose(Randomizable, InvertibleTransform):
         override_keys: this optional parameter specifies the keys to which ``overrides`` are to be applied. If
             ``overrides`` is set, ``override_keys`` must also be set.
         verbose: whether to print debugging info when lazy_evaluation=True.
+        mode: "default" or "strict"; only available when `lazy_evaluation=True`.
+            "default": run transforms lazily as much as possible for preprocessing performance.
+            "strict": only run consecutive LazyTransform classes lazily.
     """
 
     def __init__(
```
atbenmurray, line 343: log_stats and verbose should be replaced and logger_name put in their place
overrides and override_keys should be merged as per #PR6257. It is more complex than necessary and we shouldn't have to change it in the next release. `lazy_evaluation` should be renamed to `lazy` as discussed and agreed by everyone. None is not a valid lazy mode in this PR and is only used to prevent a stack overflow occurring during construction. The default value for it should be `False`

monai/transforms/compose.py
```diff
@@ -131,6 +137,7 @@ def execute_compose(
     threading: bool = False,
     log_stats: bool = False,
     verbose: bool = False,
+    mode: str = "default",
 ) -> NdarrayOrTensor | Sequence[NdarrayOrTensor] | Mapping[Any, NdarrayOrTensor]:
```
atbenmurray, line 141: Logging still has a bug whereby the first transform is not logged.

monai/transforms/compose.py
```diff
@@ -73,16 +74,19 @@ def evaluate_with_overrides(
...
         return data  # eager evaluation
     overrides = (overrides or {}).copy()
     if isinstance(data, monai.data.MetaTensor):
+        if isinstance(upcoming, mt.MapTransform) and override_keys not in upcoming.keys:
+            return data
         if data.has_pending_operations and (
             (upcoming is None)
-            or (isinstance(upcoming, mt.Identity))
-            or (isinstance(upcoming, mt.Identityd) and override_keys in upcoming.keys)
+            or (isinstance(upcoming, (mt.Identity, mt.Identityd)))
```
atbenmurray, line 88: Identity and Identityd should not be used as a synchronisation mechanism. There should be proper transforms that have the specific responsibility of causing this to occur.

atbenmurray (changes requested):

monai/transforms/compose.py
mjorgecardoso: Hi all, as agreed on Friday on the team meeting, we should be working on merging Ben's codebase and not on changing the previous implementation.

wyli: Hi @mjorgecardoso, we agreed to have a week's time. I've been actively helping Ben review the design and providing advice, unfortunately, the codebase still has large technical gaps as I commented on the PR. On the other hand, we don't find any issue on the current dev branch.

atbenmurray:
This is out of date. We'll revisit further lazy resampling functionality in new PRs. See https://github.com/Project-MONAI/MONAI/discussions/7151 for details

PR #6468, opened 2023-05-05
map_classes_to_indices() ClassesToIndices() update

myron:
map_classes_to_indices utility (and ClassesToIndices transforms) currently returns a list of coordinates for each class, but the type is List[ of MetaTensors], where each class coordinate sub-list is its own MetaTensor.  This PR changes it to return a list of torch.Tensors (or ndarray), since we don't need a MetaTensor here.

I ran into an issue with current MetaTensor list, where it would just freeze without any errors, when trying to save the cached indices (returned from ClassesToIndices to cache, which is ListProxy shared mem).  It randomly happens, but much more frequently when the number of classes is large (e.g. 105 output classes, so ClassesToIndices returns a list of 105 MetaTensors).  I'm not sure what the cause of the freeze is, but my guess is that ListProxy tries to pickle each element of this list (and struggles with MetaTensors). Disabling MetaTensor return type here, solves the issue. 

since we don't really need meta tensor return type for each class (coordinates), this seems like harmless fix. the original image/label always will have meta info.

@wyli plz check.

### Description

### Types of changes

### DIFF
--- a/monai/transforms/utils.py
+++ b/monai/transforms/utils.py
@@ -409,8 +409,12 @@ def map_classes_to_indices(
         if img_flat is not None:
             label_flat = img_flat & label_flat
         # no need to save the indices in GPU, otherwise, still need to move to CPU at runtime when crop by indices
-        output_type = torch.Tensor if isinstance(label, monai.data.MetaTensor) else type(label) # output in tensor or ndarray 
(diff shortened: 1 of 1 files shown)


wyli (approved):
Looks good to me, I'll also look into the metatensor issue

PR #6481, opened 2023-05-06
Auto3dseg ensemble path fix

myron:
Auto3dseg ensembling saves output results to the same filename if input images had the same name (but in different folders). this PR maintains the folder structure, so that the outputs are in different folders.  

### Description

A few sentences describing the changes proposed in this pull request.

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick --unittests  --disttests`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

PR #6485, opened 2023-05-07
Redo Auto3dseg ensemble path fix

mingxin-zheng:
redo #6481 .

### Description

- Take the code from #6481 
- Deprecate `data_src_cfg_filename` and replace it with `data_src_cfg_name` so that the same thing can be consistent across AutoRunner, BundleGen, and Ensemble classes
- Make `data_src_cfg_name` a required argument in the ensemble runner because it is prone to error to use a default name.
- Improves test structure, but nothing major is changed.

What is not done:
- Add tests to check the output structure, because EnsembleRunner doesn't support changing the pred_param for individual algo seperately, and then I can't really run the `ensemble` method in EnsembleRunner.

### Types of changes


wyli (approved):
Thanks!

myron:
@mingxin-zheng  since you've standardized it to use "data_src_cfg_name" , do  you want to replace "data_src_cfg_filename" in other places? I still see that name in BundleGen vars ? thanks

mingxin-zheng:
Thanks @myron! I forgot to do so but I will for the `BundleGen`.

PR #6529, opened 2023-05-18
Set autoscale false for auto3dseg test

heyufan1995:
Fixes # .
Disabled autoscale for auto3dseg template testing. 
### Description

### Types of changes


wyli (approved):
@mingxin-zheng please let me know if any concerns

mingxin-zheng:
The changes are only in HPO, which I think it should be fine.
My question to @heyufan1995 would be whether it is necessary to keep the auto scaling off in all unit tests, as we strived to use the minimal resources in the unit test and avoid overloading the testing systems.

PR #6530, opened 2023-05-18
Fix infer files < world size bug

heyufan1995:
Fixes # .

### Description
Fix the ensembler bug where infer_files are empty or smaller than GPU number

### Types of changes


monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -582,9 +582,18 @@ def ensemble(self):
...
+            infer_files = partition_dataset(data=infer_files, shuffle=False,
+                                            num_partitions=len(infer_files))[self.rank] \
+                                            if self.rank < len(infer_files) else []
+        else:
+            infer_files = partition_dataset(
+                data=infer_files, shuffle=False, num_partitions=self.world_size, even_divisible=True
+            )[self.rank]
+
         # TO DO: Add some function in ensembler for infer_files update?
         self.ensembler.infer_files = infer_files
```
myron, line 598: do we need the logic below (including self.ensembler(pred_param=self.kwargs) ) when infer_files is empty list?

are you sure self.ensembler() will not crash with an empty list? may be it's saver to write, 

```
if len(infer_files) >0:
  self.ensembler(...)
  
```

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -582,9 +582,18 @@ def ensemble(self):
         builder.set_ensemble_method(self.ensemble_method)
         self.ensembler = builder.get_ensemble()
         infer_files = self.ensembler.infer_files
-        infer_files = partition_dataset(
-            data=infer_files, shuffle=False, num_partitions=self.world_size, even_divisible=True
-        )[self.rank]
+        if len(infer_files) < self.world_size:
+            if len(infer_files) == 0:
+                logger.info("No existing infer files. Ensembler ending.")
```
myron, line 587: consider changing test "No testing files for inference is provided, ensure testing key exists in the datalist json file"

monai/apps/auto3dseg/ensemble_builder.py
heyufan1995: if infer_files is empty, it's returned. I will test if it will crash
```
        if len(infer_files) < self.world_size:
            if len(infer_files) == 0:
                logger.info(f"No existing infer files. Ensembler ending.")
                return 
```

monai/apps/auto3dseg/ensemble_builder.py
heyufan1995: ok

monai/apps/auto3dseg/ensemble_builder.py
```diff
@@ -582,9 +582,18 @@ def ensemble(self):
...
+                logger.info("No testing files for inference is provided. Ensembler ending.")
+                return
+            infer_files = partition_dataset(data=infer_files, shuffle=False,
+                                            num_partitions=len(infer_files))[self.rank] \
+                                            if self.rank < len(infer_files) else []
+        else:
+            infer_files = partition_dataset(
+                data=infer_files, shuffle=False, num_partitions=self.world_size, even_divisible=True
+            )[self.rank]
+
```
myron, line 596: shouldn't it be simply:
`infer_files = infer_files[self.rank] if  self.rank < len(infer_files) else []`
why do we  need the partition dataset here?

wyli (approved):

PR #6538, opened 2023-05-22
Match type hint with `ResidualUnit` and `Convolution`

MathijsdeBoer:
### Description

I noticed the type hint for `dropout` in `ConvDenseBlock` did not match the ones further down the file, when `dropout` is passed on to `ResidualUnit` or `Convolution`.
This pull request changes this type hint to match the downstream ones.

I don't think this will cause issues with the unit tests, but they're currently running while I prepare this pull request.

### Types of changes


wyli (approved):
Thanks, it looks good to me.

PR #6544, opened 2023-05-23
add softmax version to focal_loss

### LINKED ISSUE #6510: implement a softmax version of focal loss
```
The softmax version of focal loss is:

  FL(p_t) = -alpha * (1 - p_t)**gamma * log(p_t),

where p_i = exp(s_i) / sum_j exp(s_j), t is the target (ground truth) class, and
s_j is the unnormalized score for class j.
```

cf https://github.com/pytorch/pytorch/blob/main/modules/detectron/softmax_focal_loss_op.cc#L38-L43

the current implementation is hard-coded to use sigmoid https://github.com/Project-MONAI/MONAI/blob/e5cf74aac60cb3512d92f9abb3dd48b5ed36e8d0/monai/losses/focal_loss.py#L26

qingpeng9802:
Fixes #6510 .

### Description

Add softmax version to Focal loss

### Types of changes

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.9"]
+        python-version: ["3.9", "3.10"]
     runs-on: ${{ matrix.os }}
     env:
       QUICKTEST: True
@@ -36,15 +36,17 @@ jobs:
       with:
         auto-update-conda: true
         python-version: ${{ matrix.python-version }}
-    - name: Install env (CPU ${{ runner.os }})
-      shell: bash -l {0}
+        auto-activate-base: false
+        environment-file: environment-dev.yml
+        activate-environment: monai
+    - name: Env info (CPU ${{ runner.os }})
+      shell: bash -el {0}
       run: |
         conda info
         conda list
-        conda env create --file environment-dev.yml
     - if: runner.os == 'windows'
       name: Windows only install
-      shell: bash -l {0}
+      shell: bash -el {0}
       run: |
         conda activate monai
         # this `cpuonly` and -c conda-forge is needed to reduce the paging file size on a github instance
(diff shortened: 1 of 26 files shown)


qingpeng9802:
For further reference, the implementation reference (pytorch/detectron) has been visualized above.
It is worth to notice that current MONAI's sigmoid focal loss is kind of different from detectron's. Detectron's sigmoid implementation does not include background class by default. See, https://github.com/pytorch/pytorch/blob/6e0c741105b4022fc7c3f8c2919aad9f971beb62/modules/detectron/sigmoid_focal_loss_op.cc#L35-L36
Other implementations also include background class by default like MONAI so this might not be a big issue.
https://github.com/kornia/kornia/blob/edbcae696909bca677c71615a71cb89ba87321e1/kornia/losses/focal.py#L144 and 
https://github.com/qubvel/segmentation_models.pytorch/blob/e7cbea0bbd546167b9d1757b2cfeca3fbe26a33e/segmentation_models_pytorch/losses/_functional.py#L36 

@wyli Please help review the current implementation. Some unit tests will be added after the review.

wyli:
thanks @qingpeng9802, it looks good to me, one minor thing is that if we assume `target` is a one-hot binary tensor, then I think `- (1 - input_ls.exp()).pow(gamma) * input_ls * target` could be simplified to compute for `target==1` only.

qingpeng9802:
> thanks @qingpeng9802, it looks good to me, one minor thing is that if we assume `target` is a one-hot binary tensor, then I think `- (1 - input_ls.exp()).pow(gamma) * input_ls * target` could be simplified to compute for `target==1` only.

```python
loss: torch.Tensor = torch.where(target>0.5, - (1 - input_ls.exp()).pow(gamma) * input_ls, 0)
```
and 
```python
input_ls[target>0.5]=0.
loss: torch.Tensor = - (1 - input_ls.exp()).pow(gamma) * input_ls
```
are benchmarked. Both are significantly slower than the original one. It looks like `torch.jit.script` can fuse element-wise operators efficiently so `torch.where` or indexing might break the optimization. Therefore, the original one is kept.

wyli (approved):

PR #6550, opened 2023-05-24
Removing L2-norm in contrastive loss (L2-norm already present in cosi…

Lucas-rbnt:
…ne-similarity computation)

Fixes # .

### Description

The `forward` method of the `ContrastiveLoss` performs L2-normalization before computing cosine similarity. The [`torch.nn.functional.cosine_similarity`](https://pytorch.org/docs/stable/generated/torch.nn.functional.cosine_similarity.html)method) already handles this pre-processing to make sure that `input` and `target` lie on the surface of the unit hypersphere. This step involves an unnecessary cost and, thus, can be removed.

### Types of changes


wyli:
> Hi @Lucas-rbnt, when the L2 norms are large values (e.g. for high dimensional embeddings) do you think this might be less stable numerically, have you tested this PR in end-to-end trainings?
> 
> cc @finalelement

seems to be addressed in pytorch 1.2 for the same https://github.com/pytorch/pytorch/commit/9e137ee583c4fdb2dd3aa0c425dc9c289454cbf2

wyli (approved):
Thanks, it looks good to me. (probably less stable for early versions of pytorch but more efficient for the recent versions)

PR #6578, opened 2023-06-04
[WIP]6577-Support-torch-trt-dynamo

### LINKED ISSUE #6577: Support TensorRT export with dynamo
PyTorch introduced the [dynamo](https://pytorch.org/docs/stable/dynamo/index.html) as a part to compile PyTorch models to different backends. And it also works for the Torch-TensorRT compiler and the Torch-TensorRT team is working on supporting it. Therefore, it would be better to have MONAI `trt_export` API also support this feature.  

Add the `use_torchscript` parameter to the `trt_export` API to choose whether to use the torchscript as IR. Because the `dynamo` supports to directly export models from pytorch models instead of the torchscript models.

binliunls:
Fixes #6577  .

### Description
PyTorch introduced the [dynamo](https://pytorch.org/docs/stable/dynamo/index.html) as a part to compile PyTorch models to different backends. And it also works for the Torch-TensorRT compiler and the Torch-TensorRT team is working on supporting it. Therefore, it would be better to have MONAI `trt_export` API also support this feature.

Currently the `trt_export` API only supports exporting models via torchscript models. Therefore a `use_torchscript` parameter will be added in this PR to the trt_export API to choose whether to use the torchscript as IR. Because the dynamo mode only supports to directly export models from pytorch models instead of the torchscript models.

### Types of changes


binliunls:
After discussing with the Torch-TensorRT team, currently there is not a way to serialize/export/save a dynamo model. So if we want to support the dynamo model, the best method as I know is to convert the model at the beginning of the inference. Since the conversion may take some overhead, the total inference latency may not necessary be optimized with this way.
On the other hand, we can create a feature request for the torch-tensorrt team to add the support for the serialization.

PR #6616, opened 2023-06-14
6612-support-mlflow-data-tracking

### LINKED ISSUE #6612: Support the MLFlfow data tracking in MONAI
The experiment management tool MLFlow released [version 2.4.0](https://github.com/mlflow/mlflow/releases/tag/v2.4.0) last week, which introduced data tracking APIs. With these APIs, users can record some info about the dataset used during training, inference and evalution. This is a very useful feature to reproduce experiments using different subsets of the original dataset.  And it would be better if MONAI can also support these APIs in the experiment management part.

Add the data tracking APIs to the MLFlowHandler API.

**Additional context**
Please note these warnings:
1) Most data tracking APIs are experimental and may change in the future.
2) The [API web page](https://mlflow.org/docs/latest/python_api/mlflow.data.html#) doesn't show any support for PyTorch dataset.

binliunls:
Fixes #6612  .

### Description
Add the dataset tracking support to MLFlow handler.

### Types of changes


monai/bundle/utils.py
```diff
@@ -114,6 +114,7 @@
         "run_name": None,
         # may fill it at runtime
         "execute_config": None,
+        "dataset_log": True,
```
Nic-Ma, line 117: I personally feel no need to add this arg in the default MLFlow settings.
What do you think?

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -68,10 +71,14 @@ class MLFlowHandler:
         epoch_log: whether to log data to MLFlow when epoch completed, default to `True`.
             ``epoch_log`` can be also a function and it will be interpreted as an event filter.
             See ``iteration_log`` argument for more details.
+        dataset_log: whether to log information about the dataset at the beginning. This arg
```
Nic-Ma, line 74: I think the arg name should be "log_dataset".

tests/test_handler_mlflow.py
```diff
@@ -230,6 +235,52 @@ def test_multi_thread(self):
                 self.tmpdir_list.append(res)
                 self.assertTrue(len(glob.glob(res)) > 0)
 
+    @skip_if_quick
+    @unittest.skipUnless(has_dataset_tracking, reason="Requires mlflow version >= 2.4.0.")
+    def test_dataset_tracking(self):
+        test_bundle_name = "endoscopic_tool_segmentation"
+        with tempfile.TemporaryDirectory() as tempdir:
```
wyli, line 242: please add a context to skip the test: https://github.com/Project-MONAI/MONAI/blob/18daff1d8b26002b02de526cc19b92a9a4907f89/tests/test_bundle_get_data.py#L45

monai/handlers/mlflow_handler.py
```diff
@@ -352,3 +402,46 @@ def _default_iteration_log(self, engine: Engine) -> None:
...
+        sample_dict["images"] = []
+        for sample in dataset:
+            if isinstance(sample, dict):
+                image_name = sample["image_meta_dict"]["filename_or_obj"] if "image_meta_dict" in sample else None
+            elif isinstance(sample, list):
+                # When using a transform like `RandCropByPosNegLabel`, a sample will be a list containing image slices.
+                image_name = sample[0]["image_meta_dict"]["filename_or_obj"] if "image_meta_dict" in sample[0] else None
+            else:
+                image_name = None
+                warnings.warn(f"Don't support {type(sample)} type samples when recording the dataset with MLFlow.")
```
wyli, line 437: this could be a generic utility function, please also consider metatensor case, for example https://github.com/Project-MONAI/research-contributions/blob/689d595503822a0622a1e905352dd692d50a90d6/auto3dseg/algorithm_templates/swinunetr/scripts/train.py#L541-L544 and put it in monai.utils

monai/handlers/mlflow_handler.py
```diff
@@ -352,3 +402,46 @@ def _default_iteration_log(self, engine: Engine) -> None:
...
+        https://mlflow.org/docs/latest/python_api/mlflow.data.html#mlflow.data.pandas_dataset.PandasDataset
+
+        Please note that it may take a while to record the dataset if it has too many samples.
+
+        Args:
+            engine: Ignite Engine, it can be a trainer, validator or evaluator.
+
+        """
+        dataloader = getattr(engine, "data_loader", None)
+        dataset = getattr(dataloader, "dataset", None) if dataloader else None
```
Nic-Ma, line 423: Can we pass the dataset object to `self.log_dataset` during initialization?
if dataset is not None, then use it to log the information.
Similar to: https://github.com/Project-MONAI/MONAI/blob/dev/monai/handlers/lr_schedule_handler.py#L39.
What do you think?

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -352,3 +402,46 @@ def _default_iteration_log(self, engine: Engine) -> None:
...
+        if not dataset:
+            raise AttributeError("The dataset of the engine is None. Cannot record it by MLFlow.")
+
+        sample_dict: dict[str, list[str]] = {}
+        sample_dict["images"] = []
+        for sample in dataset:
```
Nic-Ma, line 429: Will this line run transforms for all the images? Maybe it's very slow, you can add a TQDM progress bar?
@wyli What do you think?

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -352,3 +402,46 @@ def _default_iteration_log(self, engine: Engine) -> None:
...
+            if isinstance(sample, dict):
+                image_name = sample["image_meta_dict"]["filename_or_obj"] if "image_meta_dict" in sample else None
```
Nic-Ma, line 431: Maybe better to use `MetaTensor` feature instead?
CC @KumoLiu 

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -352,3 +402,46 @@ def _default_iteration_log(self, engine: Engine) -> None:
...
+                warnings.warn(f"Don't support {type(sample)} type samples when recording the dataset with MLFlow.")
+
+            if not isinstance(image_name, str):
+                warnings.warn(
+                    f"Expected type string, got type {type(image_name)} of the image name."
+                    "May log an empty dataset in MLFlow"
+                )
+            else:
+                sample_dict["images"].append(image_name)
+        dataset_type = "train" if isinstance(engine, Trainer) else "nontrain"
```
Nic-Ma, line 446: "train" , "nontrain" look weird to me.. @ericspod What do you think?

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -222,6 +239,39 @@ def _set_experiment(self):
...
+            f"{dataset_name}_samples": pandas_dataset.profile["num_rows"],
+        }
+
+    def _log_dataset(self, sample_dict: dict[str, Any], context: str = "nontrain") -> None:
+        if not self.cur_run:
+            raise ValueError("Current Run is not Active to log the dataset")
+
+        # Need to update the self.cur_run to sync the dataset log, otherwise the `inputs` info will be out-of-date.
+        self.cur_run = self.client.get_run(self.cur_run.info.run_id)
+        logged_train_set = [x for x in self.cur_run.inputs.dataset_inputs if "train" == x.dataset.name[: len("train")]]
```
Nic-Ma, line 256: `if "train" == x.dataset.name[: len("train")]]` looks very confusing...
Maybe `x.dataset.name.startswith("train")`?

Thanks.

tests/test_handler_mlflow.py
```diff
@@ -230,6 +235,52 @@ def test_multi_thread(self):
...
+            workflow.bundle_root = bundle_root
+            workflow.dataset_dir = data_dir
+            workflow.initialize()
+            mlflow_handler.attach(workflow.evaluator)
+            workflow.run()
+            workflow.finalize()
+
+            cur_run = mlflow_handler.client.get_run(mlflow_handler.cur_run.info.run_id)
+            logged_nontrain_set = [
+                x for x in cur_run.inputs.dataset_inputs if "nontrain" == x.dataset.name[: len("nontrain")]
```
Nic-Ma, line 279: And I didn't get where we set the name `nontrain` for `x.dataset`?

Thanks.

monai/handlers/mlflow_handler.py
```diff
@@ -222,6 +239,39 @@ def _set_experiment(self):
             raise ValueError(f"Cannot set a deleted experiment '{self.experiment_name}' as the active experiment")
         self.experiment = experiment
 
+    @staticmethod
```
Nic-Ma, line 242: I don't get why make this function as static?

Thanks.

monai/handlers/mlflow_handler.py
wyli: yes, if it's tracking a static dataset (doesn't change during workflow execution), perhaps this should be done before running the workflow, for example with a command `python -m monai.bundle log_dataset ...`

monai/handlers/mlflow_handler.py
binliunls: Hi @Nic-Ma ,
    I just named `dataset_log` arg this way to keep consistence with the `epoch_log` and `iteration_log` parameters. Would you suggest to change all these parameters to `log_xxx` or just the `log_dataset`?

Thanks,
Bin

monai/handlers/mlflow_handler.py
binliunls: Hi @Nic-Ma ,
    I thought this function didn't relate to any varible or method of this class and more like a utility function of MLFlow hander, so I made it static.

Thanks,
Bin

monai/handlers/mlflow_handler.py
binliunls: Hi @Nic-Ma ,
    This is where the 'train' or 'nontrain' dataset was set for [this comment](https://github.com/Project-MONAI/MONAI/pull/6616/files#r1252028840). 

Thanks,
Bin

monai/handlers/mlflow_handler.py
binliunls: Thanks, I will take a look at this.

monai/handlers/mlflow_handler.py
binliunls: Hi @KumoLiu ,
    Actually, the `nontrain` parameter is to record validation dataset or test dataset, which are used during the training or inference. So I think `unused` may not be a decent for it.

Thanks,
Bin

monai/handlers/mlflow_handler.py
```diff
@@ -222,6 +239,39 @@ def _set_experiment(self):
...
+            raise ValueError("Current Run is not Active to log the dataset")
+
+        # Need to update the self.cur_run to sync the dataset log, otherwise the `inputs` info will be out-of-date.
+        self.cur_run = self.client.get_run(self.cur_run.info.run_id)
+        logged_train_set = [x for x in self.cur_run.inputs.dataset_inputs if "train" == x.dataset.name[: len("train")]]
+        logged_nontrain_set = [
+            x for x in self.cur_run.inputs.dataset_inputs if "nontrain" == x.dataset.name[: len("nontrain")]
+        ]
+        # In case there are more datasets.
+        dataset_cnt = str(len(logged_nontrain_set if context == "nontrain" else logged_train_set))
```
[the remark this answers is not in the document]
binliunls: Oh, it means `dataset count` to record sample numbers in the dataset. I can update it to `sample_count` if it looks ambiguours to you.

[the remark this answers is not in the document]
KumoLiu: Got it, thanks!

monai/handlers/mlflow_handler.py
```diff
@@ -222,6 +239,39 @@ def _set_experiment(self):
...
+        dataset_name = f"{context}_dataset_{dataset_cnt}"
+        sample_df = pandas.DataFrame(sample_dict)
+        dataset = mlflow.data.from_pandas(sample_df, name=dataset_name)
+        exist_dataset_list = list(
+            filter(lambda x: x.dataset.digest == dataset.digest, self.cur_run.inputs.dataset_inputs)
+        )
+
+        if not len(exist_dataset_list):
+            datasets = [mlflow.entities.DatasetInput(dataset._to_mlflow_entity())]
+            self.client.log_inputs(run_id=self.cur_run.info.run_id, datasets=datasets)
```
KumoLiu, line 271: I am concerned about whether we need to support this new feature for now. Seems this feature is still experimental and may change or be removed in a future release without warning, then we will be easily influenced by their release. Or maybe we should fix the version of mlflow to 2.4.0?
https://mlflow.org/docs/latest/python_api/mlflow.client.html?highlight=log_inputs#mlflow.client.MlflowClient.log_inputs

binliunls: Yes, I added this concern in the [issue](https://github.com/Project-MONAI/MONAI/issues/6612#issue-1755085439) when I created it and made the default value of `dataset_log` to be `False`.  I will also update the parameter in the default tracking set to be false. If this is not enough, I think we can pending this PR for now and keep on updating it until mlflow make it a long term API.

monai/handlers/mlflow_handler.py
KumoLiu: ok, thanks. I just misunderstood here.

monai/handlers/mlflow_handler.py
KumoLiu: `log_dataset` at first makes sense, but what if the dataset is too large if may also take a long time to do transforms on all the data? Seems here we only need to track the file name and number of the dataset here, actually, we don't need to do any transforms on them, right? Is it possible to use a JSON file here instead of looping the dataset? Seems we only set `dataset_dir` as `REQUIRED` in the bundle property.

binliunls: It would be better if there is a more efficient way to get all sample information without doing any transform on them.

KumoLiu: Can we use `dataset.data` here? However, this does not guarantee that the inputs are already loaded, but in this case, there is also no file_name.

binliunls: Thanks @KumoLiu , I think it's a more efficient way to get the sample information and will update this part accordingly.

monai/handlers/mlflow_handler.py
```diff
@@ -222,6 +249,36 @@ def _set_experiment(self):
...
 
+    @staticmethod
+    def _get_pandas_dataset_info(pandas_dataset):
+        dataset_name = pandas_dataset.name
+        return {
+            f"{dataset_name}_digest": pandas_dataset.digest,
+            f"{dataset_name}_samples": pandas_dataset.profile["num_rows"],
+        }
+
+    def _log_dataset(self, sample_dict: dict[str, Any], context: str | KeysCollection = "train") -> None:
```
KumoLiu, line 260: I don't think the typehint here for context can be `KeysCollection`, what do you think?

binliunls: This type hints was added to solve the mypy error, since the input parameter in line 466 is `KeysCollection`. While currently MLFlow only supports string type key, I can update all the relative parameters to `str` and remove the `KeysCollection`.

Thanks,
Bin

monai/handlers/mlflow_handler.py
```diff
@@ -68,10 +72,19 @@ class MLFlowHandler:
...
         epoch_logger: customized callable logger for epoch level logging with MLFlow.
             Must accept parameter "engine", use default logger if None.
         iteration_logger: customized callable logger for iteration level logging with MLFlow.
             Must accept parameter "engine", use default logger if None.
+        dataset_logger: customized callable logger to log the dataset information with MLFlow.
+            Must accept parameter "dataset_dict", use default logger if None.
+        dataset_dict: a dictionary in which the key is the name of the dataset and the value is a PyTorch
+            dataset, that needs to be recorded.
+        dataset_keys: a key or a collection of keys to indicate contents in the dataset that
+            need to be stored by MLFlow.
```
Nic-Ma, line 87: I feel these `dataset_logger`, `dataset_dict`, `dataset_keys` are too complicated to understand and use..
Can we simplify them? Maybe only 1 arg?
CC @KumoLiu and @wyli 

Thanks.

binliunls: Thanks Nic, I think the `dataset_log` and `dataset_dict` can be simplified to 1 arg.

tests/test_handler_mlflow.py
binliunls: Added it to line 247.

monai/handlers/mlflow_handler.py
binliunls: Updated these parameters.

monai/bundle/utils.py
binliunls: Deleted these part from the `utils` file, since the mlflow data tracking is an experiment API and may change in the future.

monai/handlers/mlflow_handler.py
binliunls: Deleted this arg.

KumoLiu: So after removing this `dataset_log`, by default, it will always log the dataset, am I right?

monai/handlers/mlflow_handler.py
```diff
@@ -352,3 +405,61 @@ def _default_iteration_log(self, engine: Engine) -> None:
...
+                raise AttributeError(f"The {dataset_type} dataset of is None. Cannot record it by MLFlow.")
+
+            sample_dict: dict[str, list[str]] = {}
+            dataset_samples = getattr(dataset, "data", [])
+            for sample in tqdm(dataset_samples, f"Recording the {dataset_type} dataset"):
+                for key in self.dataset_keys:
+                    if key not in sample_dict:
+                        sample_dict[key] = []
+
+                    if key in sample:
```
KumoLiu, line 453: Another concern here is that array format input data can not be logged here, right? If like this, do we also need to update the docstring for the `dataset_dict` to clarify this one?

KumoLiu (approved):
Other parts LGTM, do we need to pending this one since the experimental features? cc @wyli, @Nic-Ma, @ericspod.

monai/handlers/mlflow_handler.py
KumoLiu: Ok, thanks. I missed Line 422.

wyli (approved):
Thanks, it looks good to me, could you please confirm and resolve the comments?

tests/test_handler_mlflow.py
binliunls: Updated this part.

Issue #5873, opened 2023-01-18
Implement a min/max Merger for overlapping patches


drbeh:
One strategy to merge patches that are the outputs of the inference model (e.g., a probability map) is to take the minimum or maximum of the prediction at the overlapping region between patches.

Ref: #5837

PR #6633, opened 2023-06-20
Zarr Merger

### LINKED ISSUE #6006: Implement Zarr Merger
The main challenge with whole slide images (WSI) is that they do not commonly fit into CPU/GPU memory. Thus requires patching when performing inference. Although our WSIReader efficiently handle loading this patches without loading the entire WSI, the outputs also will be patches that usually cannot be merged on the memory so it requires Mergers that can effectively handle these outputs and merge them on the disk. To that end, Zarr is being used in digital pathology, which is a format for the storage of chunked, compressed, N-dimensional arrays.

drbeh:
Fixes #6006 

### Description

This PR implements `ZarrAvgMerger` which can be used for patch inference. Also a use case is demonstrated [here](https://github.com/Project-MONAI/tutorials/pull/1433).

### Types of changes


monai/inferers/merger.py
```diff
@@ -170,6 +186,177 @@ def get_counts(self) -> torch.Tensor:
...
         Returns:
-            torch.Tensor: Number of accumulated samples at each location.
+            torch.Tensor: number of accumulated samples at each location.
+        """
+        return self.counts
+
+
+@require_pkg(pkg_name="zarr")
+class ZarrAvgMerger(Merger):
+    """Merge patches by taking average of the overlapping area and store the results in zarr array.
```
ericspod, line 196: I think you really need to explain here what zarr is, it's the first that comes to mind reading this. It would be helpful to explain what it's used for and why.

ericspod:
We do also need tests to cover this code. Zarr does look interesting, I wonder if it would have other application MONAI, eg. data handling in transforms.

monai/inferers/merger.py
```diff
@@ -152,12 +159,21 @@ def finalize(self) -> torch.Tensor:
...
+        Get the final merged output.
+
+        Returns:
+            torch.Tensor: merged output.
+        """
+        return self.finalize()
+
     def get_values(self) -> torch.Tensor:
         """
         Get the accumulated values during aggregation or final averaged values after it is finalized.
```
JHancox, line 173: Are the final values always averaged? Are there not other aggregation options?

monai/inferers/merger.py
drbeh: Sure thing!

monai/inferers/merger.py
drbeh: Yes, it is an average merger similar to `AvgMerger` but we can implement any other aggregation operations too, like min, max, etc. Are there anything in particular that you think would be commonly used? If we can have a list with priority we can plan to implement them after this.

monai/inferers/merger.py
```diff
@@ -170,6 +186,177 @@ def get_counts(self) -> torch.Tensor:
...
+        """
+        if self.is_finalized:
+            raise ValueError("`ZarrAvgMerger` is already finalized. Please instantiate a new object to aggregate.")
+        patch_size = values.shape[2:]
+        map_slice = tuple(slice(loc, loc + size) for loc, size in zip(location, patch_size))
+        map_slice = ensure_tuple_size(map_slice, values.ndim, pad_val=slice(None), pad_from_start=True)
+        self.values[map_slice] += values.numpy()
+        self.counts[map_slice] += 1
+
+    def finalize(self) -> zarr.Array:
```
JHancox, line 292: Is having separate `aggregate` and `finalize `methods so that you can gradually accumulate results as patches are inferred and then do a reduction when complete? If so, perhaps the nomenclature is not quite right.

monai/inferers/merger.py
JHancox: You might want to only take the maximum class prediction value (from C channel)

drbeh: If there are multiple values in the C channel and want to take the maximum, it can be done in the [patch] post-processing part by providing a function/transform. Also for the overlapping region we would need a `MaxMerger` which has yet to be implemented. https://github.com/Project-MONAI/MONAI/issues/5873

monai/inferers/merger.py
drbeh: Added more info about zarr in the docstring.

monai/inferers/merger.py
```diff
@@ -170,6 +186,185 @@ def get_counts(self) -> torch.Tensor:
...
+        Raises:
+            NotImplementedError: When the subclass does not override this method.
+
+        """
+        if self.is_finalized:
+            raise ValueError("`ZarrAvgMerger` is already finalized. Please instantiate a new object to aggregate.")
+        patch_size = values.shape[2:]
+        map_slice = tuple(slice(loc, loc + size) for loc, size in zip(location, patch_size))
+        map_slice = ensure_tuple_size(map_slice, values.ndim, pad_val=slice(None), pad_from_start=True)
+        self.values[map_slice] += values.numpy()
```
JHancox, line 289: I think you may need to add some locking here to prevent possible data races, if two processes are working on different overlapping patches.
[](https://zarr.readthedocs.io/en/stable/tutorial.html#parallel-computing-and-synchronization)

monai/inferers/merger.py
```diff
@@ -170,6 +186,185 @@ def get_counts(self) -> torch.Tensor:
...
+        self,
+        merged_shape: Sequence[int],
+        cropped_shape: Sequence[int] | None = None,
+        output_dtype: np.dtype | str = "float32",
+        value_dtype: np.dtype | str = "float32",
+        count_dtype: np.dtype | str = "uint8",
+        store: zarr.storage.Store | str = "merged.zarr",
+        value_store: zarr.storage.Store | str | None = None,
+        count_store: zarr.storage.Store | str | None = None,
+        compressor: str = "default",
```
JHancox, line 233: If the primary use-case is for temporary storage, did you try setting `compressor=None`. This might reduce latency, but probably depends on a few factors.

monai/inferers/merger.py
JHancox: I wonder whether an option to weight the aggregation (say, linearly) according to how far each pixel is from its patch border might help reduce border artefacts.

monai/inferers/merger.py
drbeh: That's a good point. I will check that.

monai/inferers/merger.py
JHancox: In response to @ericspod's question about where else Zarr might apply: it could have a role in patch-based training pipelines. i.e. the first time, you load the patch directly from the study/WSI into Zarr. Thereafter, you just use the Zarr array.

monai/inferers/merger.py
```diff
@@ -257,15 +258,15 @@ def __init__(
             shape=self.merged_shape,
             chunks=self.chunks,
             dtype=self.value_dtype,
-            compressor=self.compressor,
+            compressor=None,
```
dzenanz, line 261: Having a light compression is usually advantageous even from speed point of view.
https://news.ycombinator.com/item?id=25926758

drbeh:
Hi @ericspod, please take another look and let me know if you have any other comments. Thanks

monai/inferers/merger.py
KumoLiu: Hi @drbeh, sorry for interrupting here. Could you please help explain more here? Do you mean that if you we one merger such as `AvgMerger` and we can use another finalize function in another merger such as `MaxMerger`?
Thanks!

drbeh: Hi @KumoLiu, I meant that some of the mergers do not need to do anything in the finalize method, like MaxMerger. I did not mean to combine mergers. They are independent and only one can be used in this framework.

KumoLiu: Thanks for clarifying!

monai/inferers/merger.py
```diff
@@ -170,6 +187,195 @@ def get_counts(self) -> torch.Tensor:
...
+        self.counts = zarr.zeros(
+            shape=self.merged_shape,
+            chunks=self.chunks,
+            dtype=self.count_dtype,
+            compressor=self.count_compressor,
+            store=self.count_store,
+            overwrite=True,
+        )
+        self.lock = threading.Lock()
+        self.lock_finalize = threading.Lock()
```
wyli, line 279: these two locks don't seem to handle the case of running aggregate and finalize at the same time in different threads, I think they should somehow be merged into one lock.

also for multi-thread aggregating, it would be great to allow for concurrent writing when the threads are dealing with non-overlapping in-place modification of `self.values[map_slice]` and `self.counts[map_slice]`. the `self.lock` might be a bit restrictive.

JHancox: It would make sense to have a barrier prior to the Finalize method. Need to be mindful of the deadlock potential.
@Wenqi - are you thinking that a 'no_lock' option would be manually set? Should default to 'False', but agree that it could be useful.

wyli: yes, agree that making the self.lock optional would be flexible enough and it could default to using a lock

drbeh:
@wyli @ericspod @JHancox please take another look to see if all comments are addressed. If so, it is ready to be merged.

monai/inferers/merger.py
```diff
@@ -170,6 +188,199 @@ def get_counts(self) -> torch.Tensor:
...
+class ZarrAvgMerger(Merger):
+    """Merge patches by taking average of the overlapping area and store the results in zarr array.
+
+    Zarr is a format for the storage of chunked, compressed, N-dimensional arrays.
+    Zarr data can be stored in any storage system that can be represented as a key-value store,
+    like POSIX file systems, cloud object storage, zip files, and relational and document databases.
+    See https://zarr.readthedocs.io/en/stable/ for more details.
+    It is particularly useful for storing N-dimensional arrays too large to fit into memory.
+    One specific use case of this class is to merge patches extracted from whole slide images (WSI),
+    where the merged results does not fit into memory and need to be stored on a file system.
```
ericspod, line 206:
```suggestion
    where the merged results do not fit into memory and need to be stored on a file system.
```

monai/inferers/merger.py
```diff
@@ -170,6 +188,199 @@ def get_counts(self) -> torch.Tensor:
...
+        count_dtype: the dtype for sample counting tensor. Default is `uint8`.
+        store: the zarr store to save the final results. Default is "merged.zarr".
+        value_store: the zarr store to save the value aggregating tensor. Default is a temporary store.
+        count_store: the zarr store to save the sample counting tensor. Default is a temporary store.
+        compressor: the compressor for final merged zarr array. Default is "default".
+        value_compressor: the compressor for value aggregating zarr array. Default is None.
+        count_compressor: the compressor for sample counting zarr array. Default is None.
+        chunks : int or tuple of ints that defines the chunk shape, or boolean. Default is True.
+            If True, chunk shape will be guessed from `shape` and `dtype`.
+            If False, ir will be set to `shape`, i.e., single chunk for the whole array.
```
ericspod, line 223:
```suggestion
            If False, it will be set to `shape`, i.e., single chunk for the whole array.
```

monai/inferers/merger.py
```diff
@@ -170,6 +188,199 @@ def get_counts(self) -> torch.Tensor:
...
+        """
+        Aggregate values for merging.
+
+        Args:
+            values: a tensor of shape BCHW[D], representing the values of inference output.
+            location: a tuple/list giving the top left location of the patch in the original image.
+
+        Raises:
+            NotImplementedError: When the subclass does not override this method.
+
```
ericspod, line 298:
```suggestion
```

monai/inferers/merger.py
```diff
@@ -170,6 +188,199 @@ def get_counts(self) -> torch.Tensor:
...
+        """
+        return self.values
+
+    def get_counts(self) -> zarr.Array:
+        """
+        Get the aggregator tensor for number of samples.
+
+        Returns:
+            zarr.Array: Number of accumulated samples at each location.
```
ericspod, line 357:
```suggestion
        Get the accumulated values during aggregation

        Returns:
            zarr.Array: accumulated values.

        """
        return self.values

    def get_counts(self) -> zarr.Array:
        """
        Get the accumulation tensor for number of samples.

        Returns:
            zarr.Array: Number of accumulated samples at each location.
```

ericspod (approved):
Looks good, I had just very minor changes to docs.

monai/inferers/merger.py
drbeh: Thanks for your comments. I applied the rest but would like to keep this one as is to match `aggregate` method.

JHancox (approved):

PR #6635, opened 2023-06-21
Fix md5 hashing with FIPS mode

MattTheCuber:
### Description
MD5 hashing is not allowed in FIPS enabled machines for security. A simple fix is to use the `usedforsecurity=False` flag in the `md5()` calls.

### Types of changes

### DIFF
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -1370,7 +1370,11 @@ def json_hashing(item) -> bytes:
 
     """
     # TODO: Find way to hash transforms content as part of the cache
-    cache_key = hashlib.md5(json.dumps(item, sort_keys=True).encode("utf-8"), usedforsecurity=False).hexdigest()
+    cache_key = ""
+    if (sys.version_info.minor < 9):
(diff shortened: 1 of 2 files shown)


monai/data/utils.py
```diff
@@ -1370,7 +1370,7 @@ def json_hashing(item) -> bytes:
 
     """
     # TODO: Find way to hash transforms content as part of the cache
-    cache_key = hashlib.md5(json.dumps(item, sort_keys=True).encode("utf-8")).hexdigest()
+    cache_key = hashlib.md5(json.dumps(item, sort_keys=True).encode("utf-8"), usedforsecurity=False).hexdigest()
```
wyli, line 1373: Thanks, this requires python 3.9+, so maybe add a switch here based on sys.version_info.minor for 3.8 compatibility

MattTheCuber: Added, lmk if you want a different format

wyli (approved):
Thanks, it looks good to me.

Issue #6655, opened 2023-06-26
Installing the dev environment fails with AttributeError: install_layout


matt3o:
**Describe the bug**
I wanted to do a short Pull request, however ran into error when trying to run the dev environment. I'll paste the output below. I forked the code from Github, pulled the repo, created my own branch (from dev) and then tried to run tests. When that failed I went back to the MONAI installation site and tried the official building steps, but they return the same error:

```bash
(monai_dev) matteo@matteo-ubuntu-server:~/code/MONAI_matteo$ python setup.py develop
setup.py with torch 2.0.0+cu117
Please set environment variable `BUILD_MONAI=1` to enable Cpp/CUDA extension build.
BUILD_MONAI_CPP=False, BUILD_MONAI_CUDA=False, TORCH_VERSION=20000.
running develop
/usr/lib/python3/dist-packages/setuptools/command/easy_install.py:158: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
/usr/lib/python3/dist-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
Traceback (most recent call last):
  File "/home/matteo/code/MONAI_matteo/setup.py", line 144, in <module>
    setup(
  File "/usr/lib/python3/dist-packages/setuptools/__init__.py", line 153, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/core.py", line 148, in setup
    return run_commands(dist)
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/core.py", line 163, in run_commands
    dist.run_commands()
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/dist.py", line 967, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/dist.py", line 985, in run_command
    cmd_obj.ensure_finalized()
  File "/usr/lib/python3/dist-packages/setuptools/_distutils/cmd.py", line 107, in ensure_finalized
    self.finalize_options()
  File "/usr/lib/python3/dist-packages/setuptools/command/develop.py", line 52, in finalize_options
    easy_install.finalize_options(self)
  File "/usr/lib/python3/dist-packages/setuptools/command/easy_install.py", line 299, in finalize_options
[...]

PR #6656, opened 2023-06-26
Workflow profiler changes

matt3o:
Fixes # .

### Description

When debugging issues with the WorkflowProfiler I saw that it creates 100% CPU usage on one core since it loops relentlessy. This commit fixes that by querying only every 0.1 seconds using ` self.queue_timeout` which I guess was inteded for that because it is unused so far in the code. If it is used in a parent class or has a different function an other variable should be added for this behaviour.

I tried to run the tests, however I ran into the error I just opened a report for: https://github.com/Project-MONAI/MONAI/issues/6655

### Types of changes

### DIFF
--- a/monai/utils/profiling.py
+++ b/monai/utils/profiling.py
@@ -21,7 +21,8 @@ from collections import defaultdict, namedtuple
 from contextlib import contextmanager
 from functools import wraps
 from inspect import getframeinfo, stack
-from time import perf_counter, perf_counter_ns, sleep
+from queue import Empty
+from time import perf_counter, perf_counter_ns
 from typing import TYPE_CHECKING, Any, cast
 
 import numpy as np
@@ -203,8 +204,8 @@ class WorkflowProfiler:
         self.parent_pid = os.getpid()
         self.read_thread: threading.Thread | None = None
         self.lock = threading.RLock()
-        self.queue: multiprocessing.SimpleQueue = multiprocessing.SimpleQueue()
-        self.queue_timeout = 0.1
+        self.queue: multiprocessing.Queue = multiprocessing.Queue()
+        self.queue_timeout = 0.01
         self.call_selector = call_selector
 
     def _is_parent(self):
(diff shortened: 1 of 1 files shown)


wyli (approved):
Thanks, it looks good to me. @ericspod please help confirm this one..

monai/utils/profiling.py
```diff
@@ -219,15 +218,14 @@ def _is_thread_active(self):
     def _read_thread_func(self):
         """Read results from the queue and add to self.results in a thread stared by `__enter__`."""
         while self._is_parent() and self._is_thread_active():
-            try:
+            while not self.queue.empty():
                 result = self.queue.get()
```
ericspod, line 222: I'd suggest instead modifying the `get` method only and not have a loop-in-loop, so keep the `try/except` and add a `timeout` argument:

```suggestion
                result = self.queue.get(timeout=self.queue_timeout)
```

ericspod: Also a sleep value of 0.01 would be sufficient as well, it's still a long time as far as the CPU is concerned.

matt3o: @ericspod At least on my machine this does not work, raises an error. According to the Python docs your solution should work, however in my Python 3.9.16 implementation get() has no timeout:

```python
class SimpleQueue(object):
[...]
    def get(self):
        with self._rlock:
            res = self._reader.recv_bytes()
        # unserialize the data after having released the lock
        return _ForkingPickler.loads(res)
```
Switching to Queue instead of SimpleQueue would fix this, or we use the double loop as I proposed above.

ericspod: Let's use `Queue` then and take advantage of its own timeout mechanism.

wyli:
Hi @matt3o if this is still an issue please feel free to create a bug report for the same.

matt3o:
@wyli I realized I missunderstood the code. The self.queue.get blocks infinitely, so it does not create a CPU usage problem, which I initially thought. One might discuss the permant blocking of the thread since it disables Ctrl+c but it works as intended, which is why I closed this PR.

wyli:
I see, thank you! (the unused `self.queue_timeout` looks strange though)

matt3o:
That's true, I think I'll try to remove it and rerun the tests. My best guess would be someone had a similar idea but found this solution in the end and forgot to remove the variable.
I have initially run into problems with WorkflowProfiler not stopping correctly, raising an OSError when the code gets interrupted in any way, which might be linked to this blocking thread, I'm not sure.

ericspod:
If it's not actually causing the CPU issues then we should not change it much, though having a timeout will give the thread a chance to complete cleanly. We could instead make the thread a daemon thread so that it doesn't prevent termination.

matt3o:
I just checked those two: Removing `self.queue_timeout` and setting `daemon=True` passes all tests. Regarding the timeout: At least the switch from SimpleQueue to Queue to activate the timeout did not work for me. It resulted in a lot of unit tests for the profiler failing and from quick glance I had no idea why that is happening (also see CI output above out that - the initial solution with time.sleep I posted worked).

Issue #1332, opened 2020-12-08
Versatile distance transforms (Euclidean and Geodesic)


tvercaut:
**Is your feature request related to a problem? Please describe.**
Several segmentation pipelines, and in particular interactive segementation ones, rely on some form of distance transform (e.g. Euclidean, Chamfer or Geodesic Distance Transform).

At the moment there is no off-the-shelf pytorch function to compute these.

**Describe the solution you'd like**
A MONAI-based implementation of distance transforms (ideally differentiable) would be fantastic. It woul accelerate uptake of such methods.

**Describe alternatives you've considered**
- Relying on scipy functions such as [distance_transform_edt](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.distance_transform_edt.html#scipy.ndimage.distance_transform_edt) and [distance_transform_cdt](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.distance_transform_cdt.html)
- Rely on external code such as [GeodisTK](https://github.com/taigw/GeodisTK)

**Additional context**
Example use cases of distance transforms in segmentaiton:
- https://github.com/Project-MONAI/MONAI/pull/1329
- https://github.com/HiLab-git/UGIR
- https://arxiv.org/abs/1710.04043
- https://arxiv.org/abs/1707.00652

wyli:
also used in monai metrics computation https://github.com/Project-MONAI/MONAI/blob/f01994d47403f920b5a51cfeae75418383d10178/monai/metrics/utils.py#L23

ddrobny:
There is a recent paper using a convolution-based and differential approximation for a distance transform:
https://link.springer.com/chapter/10.1007%2F978-3-030-71278-5_31
There doesn't seem to be related code published yet but it might be worth looking into.

tvercaut:
Thanks @ddrobny.

On a related note, tensorflow has a GPU implementation of the Felzenszwalb & Huttenlocher Distance transform:
- https://www.tensorflow.org/addons/api_docs/python/tfa/image/euclidean_dist_transform
- https://github.com/tensorflow/addons/blob/master/tensorflow_addons/custom_ops/image/cc/kernels/euclidean_distance_transform_op_gpu.cu.cc
Note sure if it's differentiable but it would be neat to at least feature match it in MONAI

For Euclidean distance, we can compare the performance (and read the interesting comments) with an established (but GPL so don't look at the code) implementation:
[...]

PR #6681, opened 2023-07-01
4103 enhances surface Dice to use subvoxel borders

### LINKED ISSUE #4103: extensions of the surface-based metric
Follow up of https://github.com/Project-MONAI/MONAI/pull/4050,
this is a feature request to extend the surface-based metric module for 3D, with the support of more detailed quantizations when computing local surface lengths/areas (such as the computation proposed in https://github.com/deepmind/surface-distance).

https://github.com/Project-MONAI/MONAI/pull/4050#discussion_r842152042
> the way 3D surfaces & surface distances are handled in deep mind is different from MONAI: Instead of defining surfaces at surface voxel boundaries, surface points are located on a grid that is shifted by half of the grid spacing along each axis (cf. Fig. 10 in their [paper](https://arxiv.org/abs/1809.04430)):
> 
> grafik
> 
> I think it should be discussed more broadly whether to keep the surface computation currently used in MONAI, or to switch to the deep mind computation, or to establish a switch such that the user can decide on the computation. This choice should be made consistently for all boundary-overlap metrics.

cc @SilviaSeidlitz

wyli:
Fixes #4103

### Description
- considers spacing and subvoxel borders when computing surface Dice
- reimplemented http://medicaldecathlon.com/files/Surface_distance_based_measures.ipynb using pytorch API
- there's possibility of speed up once https://github.com/Project-MONAI/MONAI/issues/1332 is addressed

### Types of changes


tests/test_surface_dice.py
```diff
@@ -330,21 +331,8 @@ def test_asserts(self):
...
 
-        with self.assertRaises(ValueError) as context:
-            SurfaceDiceMetric(class_thresholds=[1, 1], include_background=True)(predictions_no_hot, predictions_hot)
-        self.assertEqual("y_pred and y should be one-hot encoded.", str(context.exception))
-        with self.assertRaises(ValueError) as context:
-            SurfaceDiceMetric(class_thresholds=[1, 1], include_background=True)(predictions_hot, predictions_no_hot)
-        self.assertEqual("y_pred and y should be one-hot encoded.", str(context.exception))
-
         predictions_no_hot = predictions_no_hot.float()
         predictions_no_hot[0, :, 0, 0] = torch.tensor([0.5, 0])
```
yiheng-wang-nv, line 335: If these tests are remove, `predictions_no_hot` is not used and can also be remove.

monai/metrics/utils.py
```diff
@@ -292,45 +323,410 @@ def prepare_spacing(
...
+        device: device to put the table on.
+
+    Returns:
+        An array of size 256, mapping neighbourhood code to the surface area.
+        ENCODING_KERNEL[3] which is the kernel used to compute the neighbourhood code.
+    """
+    # compute the area for all 256 possible surface elements given a 2x2x2 neighbourhood according to the spacing_mm
+    c = _get_neighbour_code_to_normals_table(device)
+    s = torch.as_tensor(
+        [[[spacing_mm[1] * spacing_mm[2], spacing_mm[0] * spacing_mm[2], spacing_mm[0] * spacing_mm[1]]]],
```
yiheng-wang-nv, line 670: could we also has a simple check for `spacing_mm` ? I think it should has length 3.

monai/metrics/utils.py
```diff
@@ -292,45 +323,410 @@ def prepare_spacing(
...
+
+    Args:
+        spacing_mm: 2-element list-like structure. Pixel spacing along the 1st and 2nd spatial axes.
+        device: device to put the table on.
+
+    Returns:
+        A 16-element array mapping neighbourhood code to the contour length.
+        ENCODING_KERNEL[2] which is the kernel used to compute the neighbourhood code.
+    """
+    first, second = spacing_mm  # spacing along the first and second spatial dimension respectively
```
yiheng-wang-nv, line 698: similar here

yiheng-wang-nv (commented):
Thanks @wyli . I checked the changes and this PR looks good to me. Could you address a few  minor issues?

wyli:
> Thanks @wyli . I checked the changes and this PR looks good to me. Could you address a few minor issues?

thanks, the comments are addressed, I also update to use the pytorch APIs as much as possible instead of converting everything into numpy arrays.

yiheng-wang-nv (approved):

PR #6686, opened 2023-07-02
Fix the use of `allow_smaller` in `CropForeground`

### LINKED ISSUE #6685: cropforeground with margin>0
**Describe the bug**
```py
import numpy as np
import monai.transforms as mt

mask_gt[0, 0:50] = 1

matshow3d(output, show=True)
```

Figure_1

Figure_2

KumoLiu:
Fixes #6685 .

### Description
When `allow_smaller=True`, it allows the image size to be smaller than the box size and will pad it.
But in `generate_spatial_bounding_box`, it generates the wrong bounding box.
https://github.com/Project-MONAI/MONAI/blob/e7fb74f32b5371bb54bff7a47447df31d06d8edf/monai/transforms/utils.py#L1001-L1003

Change the default value of `allow_smaller` in `CropForeground` to `False` to be consistent with the previous behavior.

### Types of changes

### DIFF
--- a/monai/transforms/croppad/array.py
+++ b/monai/transforms/croppad/array.py
@@ -837,9 +837,9 @@ class CropForeground(Crop):
             channel_indices: if defined, select foreground only on the specified channels
                 of image. if None, select foreground on the whole image.
             margin: add margin value to spatial dims of the bounding box, if only 1 value provided, use it for all dims.
-            allow_smaller: when computing box size with `margin`, whether allow the image size to be smaller
-                than box size, default to `False`. If `True`, the output will pad with specified `mode`
-                when the margined size is larger than image size.
+            allow_smaller: when computing box size with `margin`, whether to allow the final box edges to be outside of
+                the image edges (the image is smaller than the box). If `True`, part of a padded output box might be outside
+                of the original image, if `False`, the image edges will be used as the box edges.
             return_coords: whether return the coordinates of spatial bounding box for foreground.
             k_divisible: make each spatial dimension to be divisible by k, default to 1.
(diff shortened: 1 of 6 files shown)


monai/transforms/croppad/array.py
```diff
@@ -838,8 +838,8 @@ def __init__(
                 of image. if None, select foreground on the whole image.
             margin: add margin value to spatial dims of the bounding box, if only 1 value provided, use it for all dims.
             allow_smaller: when computing box size with `margin`, whether allow the image size to be smaller
```
wyli, line 840: thanks, now I understand the logic, maybe the docstring could be enhanced:
```suggestion
            allow_smaller: when computing box size with `margin`, whether to allow the final box edges to be outside of the image edges (the image is smaller than the box). If `True`, part of a padded output box might be outside of the original image, if `False`, the image edges will be used as the box edges.
```

KumoLiu: Hi Wenqi, seems the meaning of your enhanced docstring is opposite.
If True, means the margined size can be larger than the image size, so it will pad the output.
If False, means the margined size can not be larger than the image size, so it will not pad the output.
Correct me if I'm wrong.

wyli: Yes I didn't see your change in the transform utils when commenting... The code logic looks good to me, perhaps the docstring still needs to clarify the current default behaviour when allow_smaller is False.

wyli (approved):

monai/transforms/croppad/array.py
KumoLiu: Thanks! Have updated the docstring based on your comments!

Issue #6693, opened 2023-07-03
Add DiceCELoss as a metric


matt3o:
**Is your feature request related to a problem? Please describe.**
Since DiceCELoss is available as a loss, it would make sense for it to be available as a metric too.

**Describe the solution you'd like**
Either a dedicated DiceCELoss metric like e.g. MeanDice for the DiceLoss OR a pluggable DiceCELossMetric which can be used as input to IgniteMetric.

**Describe alternatives you've considered**
Passing the DiceCELoss into IgniteMetric results in an error since it does not have a reset button. Which kinda makes sense, this is no metric, it does something different.

**Additional context**
I am also curious about why the dice got implemented at two places, namely MeanDice and the DiceLoss. From a quick glance I believe they do the same computation - so is there any reason for the code to duplicated here? The calculation logic for both should be the same.

wyli:
This function actually wraps any loss as part of cumulative metric https://github.com/Project-MONAI/MONAI/blob/dev/monai/metrics/loss_metric.py perhaps we can focus on this class.

Meandice is not differentiable and (thus) is much faster than dice loss.

PR #6695, opened 2023-07-03
Add loss_fn to IgniteMetric and rename to IgniteMetricHandler

matt3o:
### Description

As explained in #6693 I would like to use the DiceCELoss as a train metric as well. This branch adds a very crude but working version of that.
The added tests, which I copied from the MeanDice metric, do still fail. It would be cool if someone more experienced could check, what needs to be done there.
I ran the code with my full DeepEdit setup and it appears to be working just fine there.

No formatting checks done so far - I want to find out first if this code is useful for others. Docs not updated yet either.

### Types of changes


monai/handlers/dice_ce_metric.py
```diff
@@ -0,0 +1,63 @@
...
+            save_details: whether to save metric computation details per image, for example: mean dice of every image.
+                default to True, will save to `engine.state.metric_details` dict with the metric name as key.
+            args: Arguments for the DiceCELoss
+
+        See also:
+            :py:meth:`monai.metrics.meandice.compute_dice`
+        """
+        # metric_fn = DiceMetric(include_background=include_background, reduction=reduction, num_classes=num_classes)
+        loss_function = DiceCELoss(*args, **kwargs)
+        metric_fn = LossMetric(loss_fn=loss_function, reduction=reduction, get_not_nans=False)
```
wyli, line 62: thanks, do you think we can make a generic `LossMetricHandler` here, instead of hard-coding `DiceCELoss`, this class can take a user-initialized `loss_function` as input? 

currently it's achievable with the `IgniteMetric` https://github.com/Project-MONAI/model-zoo/blob/b245833e22544126f23925eac89da471562467a1/models/mednist_reg/configs/train.yaml#L180-L185 but I think `LossMetricHandler` would make the module logic easier to understand.

matt3o: @wyli Sure, I just added the code for that. The tests will be even more broken then. But in theory it works just fine, I am calling it with

```python
    all_val_metrics["val_mean_dice"] = IgniteLossMetric(
        loss_fn=loss_function_metric,
        output_transform=from_engine(["pred", "label"]),
    )
```

I called it IgniteLossMetric, but I can change it to LossMetricHandler if you prefer that.

Anyways I'm not even sure it we need this class, since it only has 3 actual lines of code. The same results can simply be achieved with the current code by doing it like this:

```python
    loss_function_metric = DiceCELoss(softmax=True, squared_pred=True)
    metric_fn = LossMetric(loss_fn=loss_function_metric, reduction="mean", get_not_nans=False)
    ignite_metric = IgniteMetric(metric_fn=metric_fn, output_transform=from_engine(["pred", "label"]), save_details=True)

    all_val_metrics = dict()
    all_val_metrics["val_mean_dice"] = ignite_metric

```
Only issue I see here is that this super non intuitive for a beginner wanting to create a new metric.. Maybe overloading the IgniteMetric to accept either metric_fn or loss_fn would be another idea?

wyli: sure, IgniteLossMetric looks good to me.. the related metrics handlers are also simple https://github.com/Project-MONAI/MONAI/blob/dev/monai/handlers/mean_dice.py cc @Nic-Ma @ericspod

(please let me know if you need any help creating the unit tests.)

ericspod: Hi @matt3o. It's only 3 lines as you show to implement this but it would be good to reduce confusion. I like the idea of adding a `loss_fn` argument to `IgniteMetric` to wrap whatever is passed by it within a `LossMetric` object. 

One issue that would be worth fixing as well is that `IgniteMetric` is in fact a handler whereas `LossMetric` is an actual metric class. It's out of scope here but I'd suggest renaming `IgniteMetric` to `IgniteMetricHandler` or something that makes it clearer what the class is not a metric but a handler _for_ metrics.

monai/handlers/dice_ce_metric.py
KumoLiu, line 1: Seems we can remove this file.

monai/handlers/ignite_loss_metric.py
```diff
@@ -0,0 +1,65 @@
...
+                https://github.com/Project-MONAI/tutorials/blob/master/modules/batch_output_transform.ipynb.
+            save_details: whether to save metric computation details per image, for example: mean dice of every image.
+                default to True, will save to `engine.state.metric_details` dict with the metric name as key.
+            args: Arguments for the DiceCELoss
+
+        See also:
+            :py:meth:`monai.metrics.meandice.compute_dice`
+        """
+        self.loss_fn = loss_fn
+        # loss_function = DiceCELoss(*args, **kwargs)
```
KumoLiu, line 63: Also for this line.

monai/handlers/ignite_loss_metric.py
```diff
@@ -0,0 +1,65 @@
...
+
+from collections.abc import Callable
+
+from monai.handlers.ignite_metric import IgniteMetric
+from monai.utils import MetricReduction
+from monai.metrics import LossMetric
+
+from torch.nn.modules.loss import _Loss
+
+class IgniteLossMetric(IgniteMetric):
```
KumoLiu, line 22: Agree with Wenqi, we can update this class name to `LossMetricHandler` which may make it clearer.

monai/handlers/ignite_loss_metric.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Computes DiceCE score metric from full size Tensor and collects average over batch, class-channels, iterations.
```
KumoLiu, line 24: May update the docstring here.

monai/handlers/ignite_loss_metric.py
```diff
@@ -0,0 +1,65 @@
...
+                lists of `channel-first` Tensors. the form of `(y_pred, y)` is required by the `update()`.
+                `engine.state` and `output_transform` inherit from the ignite concept:
+                https://pytorch.org/ignite/concepts.html#state, explanation and usage example are in the tutorial:
+                https://github.com/Project-MONAI/tutorials/blob/master/modules/batch_output_transform.ipynb.
+            save_details: whether to save metric computation details per image, for example: mean dice of every image.
+                default to True, will save to `engine.state.metric_details` dict with the metric name as key.
+            args: Arguments for the DiceCELoss
+
+        See also:
+            :py:meth:`monai.metrics.meandice.compute_dice`
```
KumoLiu, line 60:
```suggestion
            :py:meth:`monai.metrics.loss_metric.LossMetric`
```

monai/handlers/ignite_loss_metric.py
```diff
@@ -0,0 +1,65 @@
...
+                single-channel class indices and the number of classes is not automatically inferred from data.
+            output_transform: callable to extract `y_pred` and `y` from `ignite.engine.state.output` then
+                construct `(y_pred, y)` pair, where `y_pred` and `y` can be `batch-first` Tensors or
+                lists of `channel-first` Tensors. the form of `(y_pred, y)` is required by the `update()`.
+                `engine.state` and `output_transform` inherit from the ignite concept:
+                https://pytorch.org/ignite/concepts.html#state, explanation and usage example are in the tutorial:
+                https://github.com/Project-MONAI/tutorials/blob/master/modules/batch_output_transform.ipynb.
+            save_details: whether to save metric computation details per image, for example: mean dice of every image.
+                default to True, will save to `engine.state.metric_details` dict with the metric name as key.
+            args: Arguments for the DiceCELoss
```
KumoLiu, line 57: Seems can remove `include_background`, `num_classes` and `args` these unused args here.

monai/handlers/ignite_loss_metric.py
```diff
@@ -0,0 +1,65 @@
...
+
+        See also:
+            :py:meth:`monai.metrics.meandice.compute_dice`
+        """
+        self.loss_fn = loss_fn
+        # loss_function = DiceCELoss(*args, **kwargs)
+        metric_fn = LossMetric(loss_fn=self.loss_fn, reduction=reduction, get_not_nans=False)
```
KumoLiu, line 64: Can we make this `get_not_nans` not be hard-coded here?

matt3o:
Okay, I will implement the changes this weekend. Just to make sure I am understanding you @wyli and @ericspod correctly:
1) I will delete all the code added so far (so no new class `IgniteLossMetric`) and
2) move it to `IgniteMetric` where it will become the new `loss_fn` parameter
3) I'll add a few new tests for that. If anyone has any ideas what would be could possibly even with the expected outcome, that would be nice.
4) (optional) I'll check out if I can change the name of `IgniteMetric` to `IgniteMetricHandler` easily

Sounds good?

wyli:
> Okay, I will implement the changes this weekend. Just to make sure I am understanding you @wyli and @ericspod correctly:
> 
> 1. I will delete all the code added so far (so no new class `IgniteLossMetric`) and
> 2. move it to `IgniteMetric` where it will become the new `loss_fn` parameter
> 3. I'll add a few new tests for that. If anyone has any ideas what would be could possibly even with the expected outcome, that would be nice.
> 4. (optional) I'll check out if I can change the name of `IgniteMetric` to `IgniteMetricHandler` easily
> 
> Sounds good?

thanks @matt3o, the plan looks good to me.

matt3o:
I started the implementation as discussed. Also I added a first test and copied it over from `MeanDice` to `TestHandlerIgniteMetricHandler`.  I am using the `DiceLoss` as the training loss.
However the MeanDice appears to return the inverse value of the loss, so `1 - DiceLoss`, i.e. 0.75 (MeanDice) instead of 0.25 (DiceLoss). This means the copied tests fail.
Now my question would be how to handle that. I can of course of simply change the tests to check for the loss result and thus 0.25. 
Imho however LossMetric should convert the loss to a metric and thus return 0.75 instead of 0.25. However this might break quite a lot of code. 
Shall I just leave this is it is and fix only the expected test values?

wyli:
Sure, please simply change the expected results, there would be some discrepancies between MeanDice and DiceLoss…

matt3o:
Finally I need instructions how to generate the alias and how to set the rename warning for IgniteMetricHandler. Apart from that the code is already running and all tests should pass. I updated the class name in the docs, tell me if there is anything more to do.

monai/handlers/ignite_metric.py
```diff
@@ -35,7 +36,7 @@
     )
 
 
-class IgniteMetric(Metric):
```
ericspod, line 38: To provide a deprecation warning about the old name I would create a new subclass of `IgniteMetricHandler` called `IgniteMetric` and decorate this with [deprecated](https://github.com/Project-MONAI/MONAI/blob/17c1e3acff583e241066da08dae4b75554b52dda/monai/utils/deprecate_utils.py#L41). The `since` value should be "1.3" and `removed` should be "1.4". The `msg_suffix` should mention the new name. You can search the repo for examples of using this decorator with functions and classes.

wyli:
> Finally I need instructions how to generate the alias and how to set the rename warning for IgniteMetricHandler. Apart from that the code is already running and all tests should pass. I updated the class name in the docs, tell me if there is anything more to do.

looks good to me, please add the unit tests to this list https://github.com/Project-MONAI/MONAI/blob/dev/tests/min_tests.py#L75 so that it's skipped when running in a minimal environment

wyli (approved):
thanks, it looks good to me, premerge tests all good

monai/handlers/ignite_metric.py
```diff
@@ -52,18 +59,37 @@ class IgniteMetric(Metric):
...
         self._engine: Engine | None = None
         self._name: str | None = None
+
+        if self.metric_fn is None and self.loss_fn is None:
+            raise ValueError("Either metric_fn or loss_fn have to be passed.")
+        if self.metric_fn is not None and self.loss_fn is not None:
+            raise ValueError("Either metric_fn or loss_fn have to be passed, but not both.")
+        if self.loss_fn:
+            self.metric_fn = LossMetric(loss_fn=self.loss_fn, reduction=reduction, get_not_nans=get_not_nans)
+        # else:
```
KumoLiu, line 90: Seem we can remove these two comment lines?

matt3o: Yes, I forgot to remove those

KumoLiu (approved):

PR #6701, opened 2023-07-07
6699 random sort grid patch

jmnolte:
### Description

The PR introduces a random sorting option to the `GridPatch, RandGridPatch, GridPatchd, RandGridPatchd` transforms. So far, the transforms allowed patches to be sorted according to the sum of their intensity or in their default order (i.e., sliding-window manner). With this PR, random extraction of patches as described in [Myronenko et al.](https://arxiv.org/abs/2111.01556) is enabled.

### Types of changes


monai/transforms/spatial/array.py
```diff
@@ -3262,7 +3262,7 @@ def filter_threshold(self, image_np: NdarrayOrTensor, locations: np.ndarray) ->
 
     def filter_count(self, image_np: NdarrayOrTensor, locations: np.ndarray) -> tuple[NdarrayOrTensor, np.ndarray]:
         """
-        Sort the patches based on the sum of their intensity, and just keep `self.num_patches` of them.
+        Sort the patches based on the sum of their intensity or in random order, and just keep `self.num_patches` of them.
```
KumoLiu, line 3265: Could you please help also update the docstring for the `sort_fn` in Line3199?
And do we also need update `RandGridPatch` since it's inherit from the `GridPatch`.

monai/transforms/spatial/array.py
```diff
@@ -3277,8 +3277,10 @@ def filter_count(self, image_np: NdarrayOrTensor, locations: np.ndarray) -> tupl
                 idx = argsort(image_np.sum(tuple(range(1, n_dims))))
             elif self.sort_fn == GridPatchSort.MAX:
                 idx = argsort(-image_np.sum(tuple(range(1, n_dims))))
+            elif self.sort_fn == GridPatchSort.RANDOM:
+                idx = np.random.permutation(image_np.shape[0])
```
wyli, line 3281: thanks for the PR, this randomised operation should be part of RandGridPatch and use self.R, similar to https://github.com/Project-MONAI/MONAI/blob/6f8e63f6cb2932492610515cd9b62f581fcb6d3e/monai/transforms/spatial/array.py#L3437
to ensure the pseudo random states are reproducible in different use cases. But I'm not sure how to implement it with minor changes, we may need to refactor both classes for this feature.

drbeh (changes requested):
@jmnolte Thanks for your contribution. The only downside of implementing the randomness is `GridPatch` is that this transform is supposed to be non-random while relying on its random counter part `RandGridPatch` to include randomness. Thus, I would suggest to only modify `RandGridPatch` for this purpose by overriding its `filter_count` method, handling situation with `sort_fn="random"`, and delegating the rest to the its parent class.

```python
def filter_count(self, image_np: NdarrayOrTensor, locations: np.ndarray) -> tuple[NdarrayOrTensor, np.ndarray]:
    if self.sort_fn == GridPatchSort.RANDOM:
        pass # do your logic for random sorting here
        return image_np, locations
    return super().filter_count(image_np, locations)
```

drbeh (commented):
Also it would be great if you can update docstring for `RandGridPatchd` (the dictionary version of the transform) for this new option (`"random"`) too. Thanks

jmnolte:
Thanks to everyone for the reviews. As per popular request from @wyli and @drbeh, I have now adapted the random sorting option to only be implemented in the class' randomised version `RandGridPatch`.

monai/transforms/spatial/array.py
```diff
@@ -3277,8 +3278,10 @@ def filter_count(self, image_np: NdarrayOrTensor, locations: np.ndarray) -> tupl
                 idx = argsort(image_np.sum(tuple(range(1, n_dims))))
             elif self.sort_fn == GridPatchSort.MAX:
                 idx = argsort(-image_np.sum(tuple(range(1, n_dims))))
+            elif self.sort_fn == GridPatchSort.RANDOM:
```
KumoLiu, line 3281: Hi @jmnolte, thanks for your quick update!
If you rewrite the `filter_count` method here, it might be a little weird. Theoretically we don't want the `GridPatch` to support random operations, so based on @drbeh  [comment](https://github.com/Project-MONAI/MONAI/pull/6701#pullrequestreview-1522897921), you should probably override this method inside `RandGridPatch`.
Free feel to ask if you have any question.
Thanks!

jmnolte:
Thanks for the feedback! I have now created a new `filter_count` method in `RandGridPatch` that overrides the parent's method if `sort_fn=random`. Let me know what you think.

wyli (approved):
thanks, it looks good to me -- please help fix the typing error

KumoLiu (approved):
Thanks for the quick update, LGTM. Could you please help also update the docstring in the `RandGridPatchd`?

jmnolte:
I am not quite sure, what's causing the typing error. It appears that the index is of an invalid type but I don't see how the index type returned from `GridPatchSort.RANDOM` is different from the type returned from the other sorting options. Could anyone take a look?

wyli:
sure, I'll have a look and trigger more tests @jmnolte.

drbeh (approved):
Thanks for addressing the comments. LGTM!

PR #6736, opened 2023-07-18
Revert #6686

### LINKED ISSUE #6685: cropforeground with margin>0

KumoLiu:
Fixes #6685.

- Revert #6686 without changing the meaning of `allow_smaller` in `generate_spatial_bounding_box`.
- Update the docstring of the `allow_smaller` to make it more clear.
- Add `deprecated_arg_default`, then after 1.3, `CropForeground` with pad by default even if the image edges are smaller than the final box edges.

### Types of changes

### DIFF
--- a/monai/metrics/utils.py
+++ b/monai/metrics/utils.py
@@ -176,7 +176,7 @@ def get_mask_edges(
         else:  # pytorch subvoxel, maybe on gpu, but croppad boolean values on GPU is not supported
             seg_pred, seg_gt, or_vol = convert_to_tensor(channel_first, dtype=torch.float16)
         cropper = CropForegroundD(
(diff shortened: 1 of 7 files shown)


wyli:
thanks, please also
- update https://github.com/Project-MONAI/MONAI/blob/a5dff5b7e4dfe3fa711e7b77b615d6b8177a0868/monai/metrics/utils.py#L178-L180
- review the docstring and add descriptions for the behaviour `allow_smaller=False` https://github.com/Project-MONAI/MONAI/blob/2008c0e28d670592e12596634ffbfbf1e29c6329/monai/transforms/utils.py#L974-L975

wyli (approved):
Thanks for the quick updates

PR #6737, opened 2023-07-18
Add str type annotation to device parameter in ToTensorD

function2-llx:
### Description

A simple fix adding `str` type annotation to the `device` parameter in `ToTensorD`, making it consistent with `ToTensor`. As this fix is very simple, I hope it's okay I don't create an issue in advance.


wyli (approved):
Thanks, it looks good to me.

PR #6768, opened 2023-07-25
feat(vnet): convert `dropout_prob` to a tuple

### LINKED ISSUE #6116: A Dropout rate of 50% is systematically applied in the V-Net even when dropout_prob is set to 0
First of all thanks for this great repo ! 

I noticed that the segmentation performance of the V-Net was a step below other architectures (DynUNet, BasicUnetPlusPlus ...). By looking at the implementation, I noticed that a dropout rate of 50% is systematically applied (hard-coded) in the UpTransition layers of the V-Net, even the user sets ```dropout_prob```  to 0. 

In monai.networks.nets.vnet : 
`self.dropout = dropout_type(dropout_prob) if dropout_prob is not None else None` (l147)
-->  `self.dropout2 = dropout_type(0.5)` **(l148)**

**Expected behavior**
It is expected that dropout is set to 0 when ```dropout_prob``` is set to 0 in the VNet class.

[https://github.com/Project-MONAI/MONAI/blob/dev/monai/networks/nets/vnet.py](url)

SauravMaheshkar:
Fixes #6116 

### Description

Converts dropout probability from a Optional float to a Optional tuple of floats.

### Types of changes


monai/networks/nets/vnet.py
```diff
@@ -224,7 +224,8 @@ def __init__(
         in_channels: int = 1,
         out_channels: int = 1,
         act: tuple[str, dict] | str = ("elu", {"inplace": True}),
-        dropout_prob: float = 0.5,
```
wyli, line 227: thanks for the PR, would be great to keep this argument backward compatible for a while -- when the user specifies `dropout_prob=` or relies on the default value, it can still work but generates warnings about the code changes. (docstring should be updated as well)

please see an example of handling the deprecated args here:
https://github.com/Project-MONAI/MONAI/blob/af0e0e9f757558d144b655c63afcea3a4e0a06f5/monai/transforms/post/array.py#L157-L178

SauravMaheshkar: Addressed in 0e9240716839f4d8c3d71e76ffecd431aef37c7f

wyli (approved):
Thanks, it looks good to me.

PR #6774, opened 2023-07-25
Fix lazy rand affine

### LINKED ISSUE #6773: lazy `RandAffine` has no effect
**Describe the bug**
When using `RandAffine` with `lazy=True`, it does not perform any transform at all.

**To Reproduce**
```python
import torch

from monai import transforms as mt

def main():
    x = torch.randn(1, 32, 32, 32)
    trans = mt.Compose(
        [mt.RandAffine(1., translate_range=[(0, 0), (1, 1), (2, 2)])],
        lazy=False
    )
    print(trans(x).affine)
    lazy_trans = mt.Compose(
        [mt.RandAffine(1., translate_range=[(0, 0), (1, 1), (2, 2)])],
        lazy=True,
    )
    print(lazy_trans(x).affine)

if __name__ == '__main__':
    main()
```

**Actual Result**
```plain
tensor([[1., 0., 0., 0.],
        [0., 1., 0., 1.],
        [0., 0., 1., 2.],
tensor([[1., 0., 0., 0.],
        [0., 1., 0., 0.],
        [0., 0., 1., 0.],
```

function2-llx:
Fixes #6773.

### Description

Call `rand_affine_grid()` once before call `rand_affine_grid.get_transformation_matrix()`, since its documented as "Get the most recently applied transformation matrix", or the `.affine` attribute will not be set.
Also, set `randomize=False` here since randomization if performed in the beginning of the function.

### Types of changes


function2-llx:
<del>very sad, find this bug after training the model for 5 days</del>

ericspod:
@atbenmurray could you please have a look (I'm away still)? Thanks!

atbenmurray:
Hi there @function2-llx , I'm sorry you've had a problem with this. I'll take a look at the proposed fix.

function2-llx:
Thank you for your concern, @atbenmurray. I understand that this is an open-source project and issues are part of the process. Let's continue to work together to improve this project!

atbenmurray:
Please forgive the delay. I've been looking into the code this morning. I think we might want to go a little further than the current PR, but I'll make suggestions in the review once I've come to a conclusion on that.

atbenmurray:
@function2-llx Ok, I had a look at a limited refactor of RandAffine and the classes that it uses, but I think it isn't possible to make tweaks without making a larger refactor. As such, I think we should go ahead with the change that you are proposing.

Do you want to write some unit tests to cover the change to RandAffine? We certainly need additional tests to cover this.

function2-llx:
@atbenmurray Hi, I'd like to help, but I have to rush my conference submission recently (Due on Aug. 15th), so I'm afraid that this PR is the most I can help recently. I can help more after the submission is finished.

atbenmurray:
@function2-llx Don't worry about it. It has been a very useful spot! I'll write up some additional unit tests to go with it.

monai/transforms/spatial/array.py
```diff
@@ -2523,6 +2523,7 @@ def __call__(
         img = convert_to_tensor(img, track_meta=get_track_meta())
         if lazy_:
             if self._do_transform:
+                self.rand_affine_grid(sp_size, randomize=randomize, lazy=True)
```
atbenmurray, line 2526: @wyli I just noticed your commits. I'm going to expand the unit tests further for this PR.
I noticed that you've reintroduced randomize=randomize. Strictly speaking, that randomize always gets called twice if it is called, but I'm presuming this is for backwards compatibility for the RNG outputs

wyli (approved):
integration verified, please feel free to create follow up PRs, thanks

Issue #5466, opened 2022-11-03
Generalized Dice Loss - Reduction="none" + batch=True


feliksh:
**Describe the bug**
`GeneralizedDiceLoss(..., reduction="none", batch=True)` does not return the expected shape.

**To Reproduce**

```
Given tensors of shape:
  network_prediction -> B x C x H x W x D
  labels -> B x 1 x H x W x D
  with B=Batch_size, C=Channels/Classes, Height, Width, Depth
```
Correct behaviour, since I am asking for no reduction and batch=True:
```
loss_fun1 = DiceLoss(to_onehot_y=True, softmax=True, reduction="none", batch=True)
loss1 = loss_fun1(network_prediction, labels)

loss1 -> [C x 1 x 1 x 1]
```

While if you do the same with GeneralizedDiceLoss:
```
loss_fun2 = GeneralizedDiceLoss(to_onehot_y=True, softmax=True, reduction="none", batch=True)
loss2 = loss_fun1(network_prediction, labels)

loss2 -> 1 x 1 x 1 x 1
```
which is not correct, since again you are asking for no reduction and batch=True.

The relevant piece of code which seems to lose the correct shape is in 
https://github.com/Project-MONAI/MONAI/blob/dev/monai/losses/dice.py#L360
```
final_reduce_dim = 0 if self.batch else 1
numer = 2.0 * (intersection * w).sum(final_reduce_dim, keepdim=True) + self.smooth_nr
denom = (denominator * w).sum(final_reduce_dim, keepdim=True) + self.smooth_dr
f: torch.Tensor = 1.0 - (numer / denom)
```
where `numer` and `denom` are collapsed to shape (1,) since both `intersection,denominator` and `w` are of shape (C,), 
thus `.sum(final_reduce_dim, keepdim=True)` does collapse the shape (C,) into (1,).

I believe the solution would be to not let those tensors be of singleton shape (C,) but instead keep them at (1, C) or (C, 1)

**Environment**

Python 3.7.9, 
Monai 0.9.1, 
PyTorch 1.11.0

KumoLiu:
Hi @feliksh, thanks for pointing it out, I just create a PR #5468 to fix it.

feliksh:
That's fantastic, impressed by the quickness. Well done, thank you!

PR #6775, opened 2023-07-25
6765 update `GeneralizedDiceLoss`

### LINKED ISSUE #6765: Bug of GeneralizedDiceLoss
**Describe the bug**
Code:
numer = 2.0 * (intersection * w) + self.smooth_nr
denom = (denominator * w) + self.smooth_dr
f: torch.Tensor = 1.0 - (numer / denom)

If self.batch is True, the shape of intersection, denominator and w are [C]. In the code, both the intersection and denominator are multiplied by w. If smooth item is ignored, then w doesn't work.

NiftyNet code:
generalised_dice_numerator = 2 * tf.reduce_sum(tf.multiply(weights, intersect))
generalised_dice_denominator = tf.reduce_sum(tf.multiply(weights, tf.maximum(seg_vol + ref_vol, 1)))

In niftynet code, they will first sum and then divide.

wyli:
Fixes #6765

### Description
as discussed in #6765, when `batch=True` the loss should still return 1 aggregated value instead of C channels. https://github.com/Project-MONAI/MONAI/issues/5466 is not actually achievable with this formulation.

### Types of changes


KumoLiu (approved):
Thanks for the update, LGTM.

PR #6788, opened 2023-07-27
feat: set `data_range` as a property

### LINKED ISSUE #6441: SSIMLoss supports set_data_range
follow-up of https://github.com/Project-MONAI/tutorials/pull/1351

should we set the `data_range` as a property that could be set and reset anytime, so that we don't need to initialize a new object every time we need to change the `data_range` parameter?

_Originally posted by @binliunls in https://github.com/Project-MONAI/tutorials/pull/1351#discussion_r1178759252_

SauravMaheshkar:
Fixes #6441 

### Description

Creates a setter method for `data_range` by setting it as a property using the **`@property`** decorator.

### Types of changes


wyli (approved):
Thanks, it looks good to me.

PR #6806, opened 2023-07-31
Added RandSimulateLowResolution(d) array and dictionary transforms and corresponding unit tests

### LINKED ISSUE #3781: Downsampling Augmentation Transform
Working with augmentation for DL pipelines, it would be nice to have a transform that downsamples the image by a factor (as opposed to a specified tensor shape or voxel size) while preserving the shape of the input to simulate images acquired at a low resolution that have been resampled to a larger shape.

An ideal solution would downsample the image by a specified factor, then upsample back to the original shape. Similar to `RandomAnisotropy` from TorchIO (https://torchio.readthedocs.io/transforms/augmentation.html) except with the capability to do isotropic downsampling.

I can get close to the desired behavior by placing two instances of a Zoom transform one after another, with reciprocal zoom factors and `keep_size=False` (e.g., `Compose([Zoom(zoom=0.5, keep_size=False), Zoom(zoom=2.0, keep_size=False)])`); however, this solution doesn't guarantee that the output shape will be the same as the input (although it will likely be close). For example, if the input shape is `[192, 256, 256]` and the desired downsampling factor is 3, the output shape will be `[192, 255, 255]`.

There may be an alternative solution that I haven't considered. Otherwise, if you think this is worth adding as a new transform (or an adaptation to an existing transform), I'd be more than happy to try to submit a PR.

aaronkujawa:
Fixes #3781.

### Description
Random simulation of low resolution corresponding to nnU-Net's (https://github.com/MIC-DKFZ/batchgenerators/blob/7651ece69faf55263dd582a9f5cbd149ed9c3ad0/batchgenerators/transforms/resample_transforms.py#L23). 
First, the array/tensor is resampled at lower resolution as determined by the zoom_factor which is uniformly sampled from the `zoom_range`. Then, the array/tensor is resampled at the original resolution. MONAI's `Resize` transform is used for the resampling operations.

### Types of changes


wyli (approved):
it looks good to me, thanks!

monai/transforms/spatial/array.py
```diff
@@ -3456,3 +3457,95 @@ def __call__(self, array: NdarrayOrTensor, randomize: bool = True):
...
+        upsample_mode: InterpolateMode | str = InterpolateMode.TRILINEAR,
+        zoom_range: Sequence[float] = (0.5, 1.0),
+        align_corners=False,
+        device: torch.device | None = None,
+    ) -> None:
+        """
+        Args:
+            prob: probability of performing this augmentation
+            downsample_mode: interpolation mode for downsampling operation
+            upsample_mode: interpolation mode for downsampling operation
```
KumoLiu, line 3485:
```suggestion
            upsample_mode: interpolation mode for upsampling operation
```

monai/transforms/spatial/dictionary.py
```diff
@@ -2518,6 +2522,94 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> dict[Hashable, N
...
+        align_corners=False,
+        allow_missing_keys: bool = False,
+        device: torch.device | None = None,
+    ) -> None:
+        """
+        Args:
+            keys: keys of the corresponding items to be transformed.
+            prob: probability of performing this augmentation
+            downsample_mode: interpolation mode for downsampling operation
+            upsample_mode: interpolation mode for downsampling operation
```
KumoLiu, line 2552:
```suggestion
            upsample_mode: interpolation mode for upsampling operation
```

KumoLiu (approved):
LGTM, thanks!

Issue #6754, opened 2023-07-21
warning msg/documentation on the tf32 related system flags and usage


wyli:
(follow up of https://github.com/Project-MONAI/MONAI/issues/6525) My larger concern is that other operations in monai will be also affected by the tf32 issue (since all operations uses `cuda.matmul` are affected). This may lead to significant reproducibility issues.

My proposal is adding something like 
https://github.com/Lightning-AI/lightning/pull/16037/files#diff-909e246d6c36514f952ae5023bd9fbcc3e8f2c6a0837ebf81d7dc96790b5f938R190-R210
to related classes/functions in monai. Then, monai will print warnings when the flag is True. Not sure when it is better to print warnings, maybe during import? Maybe warnings can be suppressed when the flage is explicitly set by users, but it seems technically challenging.
&
adding a part in the documentation to educate users how to use tf32 properly.

_Originally posted by @qingpeng9802 in https://github.com/Project-MONAI/MONAI/issues/6525#issuecomment-1643896567_

qingpeng9802:
Also, found a related part in the repo, fyi
https://github.com/Project-MONAI/MONAI/blob/2800a76764bc2f58402718fcef8b38f3a60776d0/tests/utils.py#L173-L198

myron:
@qingpeng9802 @wyli 

Guys, I understand what you're trying to do, but I train on multi-gpu and the screens starts full of Warnings, which is a bit overwhelming

a) is there a way to disable these warnings? ( I do know that TF32 is enabled)
b) does this new check introduce some overhead? seems like every process in DDP ran it separately


```
/usr/local/lib/python3.10/dist-packages/monai/utils/tf32.py:76: UserWarning: Environment variable `TORCH_ALLOW_TF32_CUBLAS_OVERRIDE = 1` is set.                
  This environment variable may enable TF32 mode accidentally and affect precision.                                                                             
  See https://docs.monai.io/en/latest/precision_accelerating.html#precision-and-accelerating                                                                    
  warnings.warn(                                                                                                                                                
/usr/local/lib/python3.10/dist-packages/monai/utils/tf32.py:76: UserWarning: Environment variable `TORCH_ALLOW_TF32_CUBLAS_OVERRIDE = 1` is set.                
  This environment variable may enable TF32 mode accidentally and affect precision.                                                                             
  See https://docs.monai.io/en/latest/precision_accelerating.html#precision-and-accelerating                                                                    
[...]

Issue #2357, opened 2021-06-11
Tests failing on A100 due to tolerance


madil90:
**Describe the bug**
14 tests are failing on A100 due to issue with tolerance. The details of all tests and relative error are given below

**To Reproduce**
On a A100 GPU:
- Setup and install MONAI
- Run tests

** Details **
Failing tests and relative error on A100 are:

- `test_rand_3d_elastic_3 (tests.test_rand_elastic_3d.TestRand3DElastic)
Mismatched elements: 1 / 8 (12.5%)
Max absolute difference: 0.00179242
Max relative differenc: 0.00010684`
- `test_value_cuda_0 (tests.test_lltm.TestLLTM)
AssertionError: With rtol=0.0001 and atol=0.0001, found 2 element(s) (out of 8) whose difference(s) exceeded the margin of error (including 0 nan comparisons). The greatest difference was 0.00022590160369873047 (0.616874098777771 vs. 0.6171000003814697), which occurred at index (2, 1).`
- `test_shape_5 (tests.test_local_normalized_cross_correlation_loss.TestLocalNormalizedCrossCorrelationLoss)
Not equal to tolerance rtol=1e-05, atol=0
Mismatched elements: 1 / 1 (100%)
Max absolute difference: 0.
Max relative dfference: 0.
 ​x: array(-0.918834, dype=float32)
 ​y: array(-0.918672)`
- `test_shape_0 (tests.test_global_mutual_information_loss.TestGlobalMutualInformationLoss)
Not equal to tolerance rtol=0.0001, atol=0
Mismatched elements: 1 / 1 (100%)
Max absolute difference: 0.
Max relative dfference: 0.
 ​x: array(-1.098231, dype=float32)
 ​y: array(-1.098602)`
- `test_shape_1 (tests.test_global_mutual_information_loss.TestGlobalMutualInformationLoss)
AssertionError: 
Not equal to tolerance rtol=0.0001, atol=0
Mismatched elements: 1 / 1 (100%)
Max absolute difference: 0.
Max relative dfference: 0.
 ​x: array(-1.083798, dype=float32)
 ​y: array(-1.083999)`
- `test_shape_2 (tests.test_global_mutual_information_loss.TestGlobalMutualInformationLoss)
AssertionError: 
Not equal to tolerance rtol=0.0001, atol=0
Mismatched elements: 1 / 1 (100%)
Max absolute difference: 0.
Max relative dfference: 0.
[...]

PR #6816, opened 2023-08-03
Tf32 warnings

qingpeng9802:
about #6754 .

### Description

show a warning if any thing may enable tf32 is detected

### Types of changes


monai/losses/__init__.py
```diff
@@ -11,6 +11,13 @@
 
 from __future__ import annotations
 
+try:
+    from ..utils import detect_default_tf32
+
+    detect_default_tf32()
```
wyli, line 17: actually this load modules will always trigger all the submodules import https://github.com/Project-MONAI/MONAI/blob/d6bafc9e9ec09b47e87c4d57f3c31232ee13686f/monai/__init__.py#L57-L61 so probably we just need one `detect_default_tf32` in `monai/__init__.py`

wyli (commented):
thanks! please see some minor inline comments

tests/utils.py
```diff
@@ -179,11 +180,8 @@ def is_tf32_env():
     global _tf32_enabled
```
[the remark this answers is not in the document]
wyli: Hi @qingpeng9802 This is not actually part of the monai API.
it's a utility function used only in the unit test pipelines to detect tf32 enabled environments. It works fine but it's probably not using the latest pytorch APIs.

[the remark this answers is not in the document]
qingpeng9802: After checking https://github.com/Project-MONAI/MONAI/issues/2357, I think this function is to relax the rtol practically if tf32 reduced the precision. That is, we would not like to relax the rtol in all tf32 situations, but only relax the rtol when we find the delta value >0.001. 

If so, the condition does not seem to reliably include all possible tf32 situations. Thus, the condition may be better changed to
```python
if detect_default_tf32() or torch.backends.cuda.matmul.allow_tf32:
```
Any thought on this?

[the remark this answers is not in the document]
wyli: Thanks for looking into this, looks good to me, please feel free to update, I'll trigger more tests to confirm. Cc @yiheng-wang-nv

wyli (approved):
Thanks, it looks good to me, testing with more environments

qingpeng9802:
The failed test is caused by https://github.com/Project-MONAI/MONAI/issues/2161. Pytorch only initializes `torch.cuda` once if `torch.cuda` is called, and there is no way to revert the initialized state (https://github.com/pytorch/pytorch/issues/28829). The only solution is letting users to adjust the position of `os.environ["CUDA_VISIBLE_DEVICES"]`, which is [recommend](https://discuss.pytorch.org/t/os-environ-cuda-visible-devices-not-functioning/105545/4) by a pytorch maintainer.

wyli:
> The failed test is caused by #2161 (also see [kornia/kornia#1951](https://github.com/kornia/kornia/issues/1951)). Pytorch only initializes `torch.cuda` once if `torch.cuda` is called, and there is no way to revert the initialized state ([pytorch/pytorch#28829](https://github.com/pytorch/pytorch/issues/28829)). The only solution is letting users to adjust the position of `os.environ["CUDA_VISIBLE_DEVICES"]`, which is [recommend](https://discuss.pytorch.org/t/os-environ-cuda-visible-devices-not-functioning/105545/4) by a pytorch maintainer.

ok... in this case how about only calling `detect_default_tf32` in the config printing ?

https://github.com/Project-MONAI/MONAI/blob/65cf5fe1234d0d01cc52c9ab48395a3efbad6f45/monai/config/deviceconfig.py#L195

currently the `import monai` is a bit slow already.

qingpeng9802:
Per my test, the function `reset_torch_cuda_after_run` in the commit 348f089 can resolve the `torch.cuda` issue. @wyli Could you trigger the GPU CIs to confirm?
Also, this new added warning increases time <0.005% so it should be fine.

qingpeng9802:
This failed case is kind of interesting🤔 
Since this issue https://github.com/pytorch/pytorch/issues/80876 is resolved in PyTorch 1.12.1, @wyli Could you trigger the GPU CI PT113 and PT210?

My test env is `Pytorch version: 2.0.0+cu117  NVIDIA-SMI 515.65.01 Driver Version: 516.94 CUDA Version: 11.7`, and this env can pass `tests.test_set_visible_devices`.

wyli:
Sure, I’ll try to rerun them. I think the utility function in the PR looks good, but adding it and the workaround to `monai/__init__.py` is a bit risky. (I read the relevant GitHub discussions, most of them are trying to avoid unnecessary early calls to `torch.cuda`)  @ericspod @Nic-Ma

qingpeng9802:
The key of `torch.cuda` issue is calling to https://github.com/pytorch/pytorch/blob/v2.0.1/torch/cuda/__init__.py#L219. IMO, `importlib.reload` should be relatively safe on PyTorch(Python) side. The main risk is that it is unclear whether PyTorch's initialization operation on CUDA is idempotent. That is, we would expect get the same result on the CUDA state after twice initialization.

monai/utils/tf32.py
```diff
@@ -0,0 +1,79 @@
...
+    import torch
+
+    from monai.utils.module import version_geq
+
+    if not version_geq(f"{torch.version.cuda}", "11.0"):
+        return False
+    for i in range(torch.cuda.device_count()):
+        major, _ = torch.cuda.get_device_capability(i)
+        if major >= 8:  # Ampere and later
+            return True
```
qingpeng9802, line 38: Two alternatives that do not use `torch.cuda` for this:
1.
`=compute_cap` flag is available after 11.6(not sure)
```py
try:
    capabilities_str = subprocess.run(
        ["nvidia-smi", "--query-gpu=compute_cap", "--format=csv,noheader"],
        check=True,
        capture_output=True,
        text=True
    ).stdout
    capabilities_list = capabilities_str.split('\n')
    if any([float(c) >= 8 for c in capabilities_list if c]):
        return True
except BaseException:
    pass
```
2.
[pyNVML](https://pypi.org/project/nvidia-ml-py/) 36KB
```py
from pynvml import nvmlInit, nvmlShutdown, nvmlDeviceGetCount, nvmlDeviceGetHandleByIndex, nvmlDeviceGetCudaComputeCapability
try:
    nvmlInit()
    for i in range(nvmlDeviceGetCount()):
        handle = nvmlDeviceGetHandleByIndex(i)
        major, _ = nvmlDeviceGetCudaComputeCapability(handle)
        if major >= 8:
            return True
except BaseException:
    pass
finally:
    nvmlShutdown()
```

ericspod: I'm trying to get back up to speed on MONAI stuff, but here I would avoid external programs since this can be slow and will break if the program isn't found through PATH, eg. under Windows in some cases. The pyNVML solution would be better and would be minor dependency to add to MONAI.

qingpeng9802: > I'm trying to get back up to speed on MONAI stuff, but here I would avoid external programs since this can be slow and will break if the program isn't found through PATH, eg. under Windows in some cases. The pyNVML solution would be better and would be minor dependency to add to MONAI.

Thanks for looking into this. Could you also check the `run_command()` of `_version.py`? It uses `subprocess.Popen`, and per my understanding of https://stackoverflow.com/a/39187984/10973766, it will not block the `import monai` process.

qingpeng9802: However, the PATH on windows of alt 1. is indeed unreliable.

ericspod: > > I'm trying to get back up to speed on MONAI stuff, but here I would avoid external programs since this can be slow and will break if the program isn't found through PATH, eg. under Windows in some cases. The pyNVML solution would be better and would be minor dependency to add to MONAI.
> 
> Thanks for looking into this. Could you also check the `run_command()` of `_version.py`? It uses `subprocess.Popen`, and per my understanding of https://stackoverflow.com/a/39187984/10973766, it will not block the `import monai` process.

Yeah I had seen before how this file works and it's a similar issue. [Here](https://github.com/Project-MONAI/MONAI/blob/65cf5fe1234d0d01cc52c9ab48395a3efbad6f45/monai/_version.py#L232) Popen is used to call git and parse out information about version and where the HEAD currently is. This could be parsed from data in the .git directory directly, which is what git does, but would be more complex and possibly error prone if the formats of these files change. It's arguably more portable to rely on `git/git.cmd/git.exe` to do the parsing for you and simple to use but still can break if the executable isn't found. On a regular desktop environment the main walltime cost when importing MONAI is still initializing CUDA in Pytorch so that's not a concern but something I'd prefer to redo sometime.

qingpeng9802:
As the disscussion above, we prefer to use alt 2. Feel free to add any comment on alt 2. If alt 2. is okay, I can push an alt 2. commit.

wyli:
sure, I think we can go for option 2 if it works fine across the test platforms and the import timing doesn't increase too much `python -X importtime -c "import monai"`

requirements.txt
```diff
@@ -1,2 +1,3 @@
 torch>=1.9
 numpy>=1.20
+nvidia-ml-py
```
wyli, line 3: I think this should be an optional dependency. if it's not installed properly, we can simply print the warning message assuming `major >= 8` is true..

PR #6828, opened 2023-08-07
Update `monai.bundle` to support model zoo NGC hosting

### LINKED ISSUE #6761: Update `monai.bundle` to support model zoo NGC hosting
Related changes in model zoo:
https://github.com/Project-MONAI/model-zoo/pull/463

I will add more information later

KumoLiu:
Fixes #6761 .

### Description

- update `download` to support monaihosting download.
- update relevant unit tests.

### Types of changes

### DIFF
--- a/tests/test_bundle_download.py
+++ b/tests/test_bundle_download.py
@@ -116,7 +116,7 @@ class TestDownload(unittest.TestCase):
                 parser = ConfigParser()
                 parser.export_config_file(config=def_args, filepath=def_args_file)
                 cmd = ["coverage", "run", "-m", "monai.bundle", "download", "--args_file", def_args_file]
-                cmd += ["--url", url, "--progress", "False"]
+                cmd += ["--url", url, "--progress", "False", "--source", "monaihosting"]
                 command_line_tests(cmd)
                 for file in bundle_files:
                     file_path = os.path.join(tempdir, bundle_name, file)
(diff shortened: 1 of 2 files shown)


KumoLiu:
In this PR, I mainly update the `download` method.
It appears that there are still several other methods that need to be updated, such as `get_all_bundles_list`, `get_bundle_versions`, and `get_bundle_info`. 

The[ NGC API doc](https://docs.ngc.nvidia.com/api/) does not, however, contain the API for the current monaihosting link. I can't currently get access to all of the bundles that exist; I can only get access to all of the version information given the name of a specific bundle. Do you think I can refer to the [model_info.json](https://github.com/Project-MONAI/model-zoo/blob/dev/models/model_info.json) method to update these methods or do you have any other suggestions?
cc @wyli @ericspod @Nic-Ma @yiheng-wang-nv

tests/test_bundle_download.py
```diff
@@ -58,6 +58,12 @@
     "model.ts",
 ]
 
+TEST_CASE_6 = [
+    ["models/model.pt", "models/model.ts", "configs/train.json"],
+    "brats_mri_segmentation",
```
KumoLiu, line 63: Is it necessary to upload a fake model to monaihosting because the real one here may download more slowly?

wyli: may include a minimal example from https://github.com/ericspod/tutorials/blob/bundles_update/bundle/bundle_intro.ipynb once that's completed

yiheng-wang-nv:
cc @zephyrie , model zoo web page may also need to be changed due to updated NGC hosting

PR #6829, opened 2023-08-07
Fix docs code for ScaleIntensityRangePercentiles

matt3o:
When running the docs code for `ScaleIntensityRangePercentiles`, it failed with the error (MONAI 1.2.dev2316):
```
  File "/home/mhadlich/anaconda3/envs/monai/lib/python3.8/site-packages/monai/transforms/intensity/array.py", line 998, in __call__
    img_t = self._normalize(img=img_t)
  File "/home/mhadlich/anaconda3/envs/monai/lib/python3.8/site-packages/monai/transforms/intensity/array.py", line 971, in _normalize
    a_min: float = percentile(img, self.lower)  # type: ignore
  File "/home/mhadlich/anaconda3/envs/monai/lib/python3.8/site-packages/monai/transforms/utils_pytorch_numpy_unification.py", line 121, in percentile
    result = torch.quantile(x, q, dim=dim, keepdim=keepdim)
RuntimeError: quantile() input tensor must be either float or double dtype
```

This commit updates the docs for the function to use `torch.Tensor` instead.

### Types of changes

### DIFF
--- a/monai/transforms/intensity/array.py
+++ b/monai/transforms/intensity/array.py
@@ -1132,7 +1132,7 @@ class ScaleIntensityRangePercentiles(Transform):
     .. code-block:: python
         :emphasize-lines: 11, 22
 
-        image = np.array(
+        image = torch.Tensor(
             [[[1, 2, 3, 4, 5],
               [1, 2, 3, 4, 5],
               [1, 2, 3, 4, 5],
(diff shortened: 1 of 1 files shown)


monai/transforms/intensity/array.py
```diff
@@ -1132,7 +1132,7 @@ class ScaleIntensityRangePercentiles(Transform):
     .. code-block:: python
         :emphasize-lines: 11, 22
 
-        image = np.array(
+        image = torch.Tensor(
```
KumoLiu, line 1135: The type of the `image` may not be the key point, the input dtype is the reason for the error message. It's possible that torch and numpy produce different default dtype.

matt3o: Hi @KumoLiu! Yes definitely, if I use `np.array` with `dtype=np.float32`, it works. Numpy choses np.int64 by default for this list. Nevertheless the return type is now a metatensor, so the prints are incorrect anyways.

KumoLiu: Yes, the update should be fine and just wanted to bring that up.

matt3o: But you are correct that this is not a bug then, it's just the wrong input dtype :) Just checked, typing is fine too, not sure what I saw there..

wyli (approved):

PR #6830, opened 2023-08-07
Add warning in a multiprocessing special case

matt3o:
Related to discussion https://github.com/Project-MONAI/MONAI/discussions/6657

This code at least adds a warning if `set_track_meta(False)` and `multiprocessing_context='spawn'` are used in the same code. However this warning only triggers if `set_track_meta(False)` has been called before the DataLoader has been initialized. I will append some example code where this is not True, still the bug is triggered.
Imo this is still a MONAI bug even though in the discussion it was claimed otherwise. The multiprocessing_context='spawn' has unintended consequences and this is only true for MONAI and not for torch. (I believe the problem is that with 'spawn' Python and all the libs get reinitialized and thus _TRACK_META is reset to being True).

### Types of changes

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -45,6 +45,7 @@ from monai.networks import (
 from monai.utils import (
     check_parent_dir,
     deprecated_arg,
+    deprecated_arg_default,
     ensure_tuple,
     get_equivalent_dtype,
     min_version,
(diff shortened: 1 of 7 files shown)


wyli (approved):
thanks for looking into this issue! the `logger` is not used in the PR, please remove it or change `warnings.warn` to `logger.warn`.

matt3o:
> thanks for looking into this issue! the `logger` is not used in the PR, please remove it or change `warnings.warn` to `logger.warn`.

Oh yeah, I totally forgot about it. If you think warnings.warn is fine, I will simply remove the import of the logger.

PR #6835, opened 2023-08-08
6834 `get` API: support get properties and metadata from bundle

### LINKED ISSUE #6834: Add `MetaProterties` and update `load` API
Part of #6726

- Add `create_workflow`, which can create a `BundleWorkflow` by specifying config_file or workflow_name (which can support Python-based bundle)
- Add `MetaProterties`, now users can directly call meta information from an initialized `BundleWorkflow`
- Update `load` API

KumoLiu:
Fixes #6834.

### Description
This PR introduces the idea of `BundleManager`, with `get` as one of the methods, to avoid repeatedly instantiating `ConfigWorkflow`.

**usage**
```
bundle = BundleManager("spleen_ct_segmentation")
bundle.get("network_data_format")

or

bundle.get(property="train_preprocessing")
```

### Types of changes


KumoLiu:
After all the reviews, I'll add the relevant unittests.

monai/bundle/scripts.py
```diff
@@ -1539,3 +1539,115 @@ def init_bundle(
...
+            download_args = {
+                "version": kwargs.pop("version", None),
+                "args_file": kwargs.pop("args_file", None),
+                "source": kwargs.pop("source", download_source),
+            }
+            download(bundle_name, bundle_dir=bundle_dir, **download_args)
+            bundle_dir = _process_bundle_dir(bundle_dir)
+            config_root_path = bundle_dir / bundle_name / "configs"  # type: ignore
+            if len(configs) > 0:
+                config_file = [str(config_root_path / f"{_config}.json") for _config in configs]
```
yiheng-wang-nv, line 1612: here should be flexible, the postfix could be `.json`, `.yaml`, `.yml`

KumoLiu: Thanks, have updated in this commit https://github.com/Project-MONAI/MONAI/pull/6835/commits/8fb5076713f5c4a7be6124c037f563420b8fb9c2.

monai/bundle/scripts.py
```diff
@@ -1539,3 +1539,126 @@ def init_bundle(
         copyfile(str(ckpt_file), str(models_dir / "model.pt"))
     elif network is not None:
         save_state(network, str(models_dir / "model.pt"))
+
+
+def _find_bundle_file(root_dir: Path, file_name: str, suffix: Sequence[str] = ("json", "yaml", "yml")) -> str | None:
+    # find bundle file with possible suffix
+    for _suffix in suffix:
+        full_name = f"{file_name}.{_suffix}"
+        if full_name in os.listdir(root_dir):
```
wyli, line 1548: this could be implemented with `os.path.isfile`, the function name could be `_find_config_file`

monai/bundle/scripts.py
```diff
@@ -1539,3 +1539,126 @@ def init_bundle(
...
+
+        # Create a BundleManager instance for the 'spleen_ct_segmentation' bundle
+        bundle = BundleManager("spleen_ct_segmentation")
+
+        # Get properties defined in `TrainProperties` or `InferProperties`
+        train_preprocessing = bundle.get("train_preprocessing")
+        print(train_preprocessing)
+
+        # Also support to retrieve meta information from the "metadata.json" file
+        version = bundle.get(meta="version")
```
wyli, line 1572: this requires the user to remember whether a property is a `meta` or a regular property, but in my understanding it's not necessary from a user's perspective. how about supporting:

```suggestion
        version = bundle.get("version")
```

monai/bundle/scripts.py
```diff
@@ -1539,3 +1539,126 @@ def init_bundle(
...
+            config_root_path = bundle_dir / bundle_name / "configs"  # type: ignore
+            if len(configs) > 0:
+                config_file = [
+                    str(config_root_path / _find_bundle_file(config_root_path, _config)) for _config in configs  # type: ignore
+                ]
+            else:
+                config_file = str(config_root_path / _find_bundle_file(config_root_path, configs[0]))  # type: ignore
+
+        logging_file = config_root_path / "logging.conf"
+        self.meta_file = config_root_path / "metadata.json"
```
wyli, line 1628: I guess these files are optional? some bundles may not have these files

monai/bundle/scripts.py
```diff
@@ -1539,3 +1539,126 @@ def init_bundle(
...
+        self.workflow = ConfigWorkflow(
+            config_file=config_file,
+            meta_file=str(self.meta_file),
+            logging_file=str(logging_file),
+            workflow=workflow,
+            **kwargs,
+        )
+        self.workflow.initialize()
+
+    def get(self, property: str | None = None, meta: str | None = None) -> Any:
```
wyli, line 1639: would be great to support a default value, such as https://github.com/Project-MONAI/MONAI/blob/a86c0e01555727bd848275c7eabc5b9dc26da73d/monai/bundle/config_parser.py#L191

monai/bundle/scripts.py
```diff
@@ -1539,3 +1539,126 @@ def init_bundle(
...
+    for _suffix in suffix:
+        full_name = f"{file_name}.{_suffix}"
+        if full_name in os.listdir(root_dir):
+            return full_name
+    return None
+
+
+class BundleManager:
+    """
+    The `BundleManager` class facilitates the automatic downloading and instantiation of bundles.
```
wyli, line 1555: will this support the python script based workflow? https://github.com/Project-MONAI/tutorials/tree/main/bundle/python_bundle_workflow cc @Nic-Ma

wyli (commented):
thanks, please see some comments inline..

monai/bundle/scripts.py
KumoLiu: Thanks, fixed it in this [commit](https://github.com/Project-MONAI/MONAI/pull/6835/commits/dc8858b53ca5d33b423f7a6d74644362d1a085f5).

monai/bundle/scripts.py
KumoLiu: Good points, have updated it in this [commit](https://github.com/Project-MONAI/MONAI/pull/6835/commits/dc8858b53ca5d33b423f7a6d74644362d1a085f5). Thanks.

monai/bundle/scripts.py
KumoLiu: Yes, you are right, updated it in this [commit](https://github.com/Project-MONAI/MONAI/pull/6835/commits/dc8858b53ca5d33b423f7a6d74644362d1a085f5). Thanks.

monai/bundle/scripts.py
KumoLiu: Supported now, thanks.

monai/bundle/scripts.py
KumoLiu: I take a look at this one, it doesn't support it now, I will add some support for this one. Thanks.

monai/bundle/scripts.py
```diff
@@ -1539,3 +1539,123 @@ def init_bundle(
...
+            if _config_path.is_file():
+                config_file = config_path
+                config_root_path = _config_path.parent
+            else:
+                raise FileNotFoundError(f"Cannot find the config file: {config_path}.")
+        else:
+            download_args = {
+                "version": kwargs.pop("version", None),
+                "args_file": kwargs.pop("args_file", None),
+                "source": kwargs.pop("source", download_source),
```
ericspod, line 1615: I'd suggest making "version", "args_file", "source" actual arguments of the constructor rather than using kwargs. For "source" the default value can be None which is taken to mean "download_source". Separately I'd also suggest renaming "download_source" to "DEFAULT_DOWNLOAD_SOURCE" or something like that since it's a constant and meant to be a default value.

monai/bundle/scripts.py
```diff
@@ -1539,3 +1539,123 @@ def init_bundle(
...
+            }
+            download(bundle_name, bundle_dir=bundle_dir, **download_args)
```
ericspod, line 1617: If the argument `bundle_dir` is None at this point the next line will change it to point somewhere `torch.hub.get_dir` states. Should line 1618 be above this line?

monai/bundle/scripts.py
KumoLiu: Thanks, Eric, I have updated this based on your comments in this [commit](https://github.com/Project-MONAI/MONAI/pull/6835/commits/5eec914f3f98f20414d10ae9f3090b1296567061).

monai/bundle/scripts.py
KumoLiu: Good points, otherwise it may need to process the dir twice, I have updated it in this [commit](https://github.com/Project-MONAI/MONAI/pull/6835/commits/5eec914f3f98f20414d10ae9f3090b1296567061). Thanks!

KumoLiu:
> There's a few lines of the new code not covered by the tests but otherwise looks good. Were definitions for `train` and `predict` planned for this PR?

Hi @ericspod, after discussing with @Nic-Ma  and @wyli offline, I will remove the `BundleManager` and `get` API and simplify the logic of the `load` and `ConfigWorkflow` since `get` can be directly replaced by getattr in `BundleWorkflow`.  
I may need to:
- Add `MetaProperties` in properties.py which may be used in the downstream app such as getting definitions for the label.
- enhance the `load` API to support loading the state dict more flexibly.
- add `create_workflow` to return workflow instance which could also support "python script based workflow".

I'm still debating whether or not to directly support download in `ConfigWorkflow` because the download bundle doesn't seem to fit there. Any suggestions would be appreciated.

vikashg (approved):
Looks good to me.

KumoLiu:
Hi @vikashg, from the error message, seems you didn't download the requests package, could you please `pip install requests` and try again? Thanks!

vikashg:
ok I will do so

KumoLiu:
Hi @ericspod, @wyli, @vikashg, @Nic-Ma and @yiheng-wang-nv,
I have finished this PR according to your comments which mainly contains several parts listed below:
- Add `create_workflow`, which can create a `BundleWorkflow` by specifying config_file or workflow_name (which can support python-based bundle)
typical usage:
```
# Specify config_file path to create workflow: 
workflow = create_workflow(config_file="/workspace/spleen_ct_segmentation/configs/train.json", workflow="train") 

# Set the workflow to other customized BundleWorkflow subclass to create workflow: 
workflow = create_workflow(workflow_name=CustomizedWorkflow)
```

- Add `MetaProterties`, now users can directly call meta information from an initialized `BundleWorkflow`
typical usage:
```
# Specify config_file path to create workflow: 
workflow = create_workflow(config_file="/workspace/spleen_ct_segmentation/configs/train.json", workflow="train") 
version = workflow.version
channel_def = workflow.channel_def
```

- update `load` API, remove `net_name` and `net_kwargs`, add `model`.
typical usage:
```
# directly update the parameters from the bundle.
model = load(name=bundle_name, bundle_dir=tempdir,source="monaihosting", spatial_dims=2, out_channels=1)

# using `exclude_vars` to filter loading weights, perhaps resolve this issue #5629
model = load(name=bundle_name, bundle_dir=tempdir,source="monaihosting", exclude_vars="model.0.conv")

# pass model
my_model = monai.networks.net.SwinUNETR(...)
model = load(name=bundle_name, bundle_dir=tempdir, source="monaihosting", model=my_model)
```

I have also added the unit tests, could you please kindly help review it again?
Thanks in advance!

monai/bundle/scripts.py
```diff
@@ -748,35 +787,17 @@ def run_workflow(workflow: str | BundleWorkflow | None = None, args_file: str |
         python -m monai.bundle run_workflow --meta_file <meta path> --config_file <config path>
 
         # Set the workflow to other customized BundleWorkflow subclass:
-        python -m monai.bundle run_workflow --workflow CustomizedWorkflow ...
+        python -m monai.bundle run_workflow --workflow_name CustomizedWorkflow ...
 
     Args:
-        workflow: specified bundle workflow name, should be a string or class, default to "ConfigWorkflow".
+        workflow_name: specified bundle workflow name, should be a string or class, default to "ConfigWorkflow".
```
KumoLiu, line 793: This argument's name was changed because it was duplicated here and may cause confusion.
https://github.com/Project-MONAI/MONAI/blob/2229c4d9cb02b093054a04ede442953497df673e/monai/bundle/workflows.py#L42

cc @Nic-Ma

monai/bundle/properties.py
```diff
@@ -48,6 +48,11 @@
         BundleProperty.REQUIRED: True,
         BundlePropertyConfig.ID: f"train{ID_SEP_KEY}trainer",
     },
+    "network_def": {
```
Nic-Ma, line 51: May I know why add this property? We may need to update the CI test of model-zoo for it, CC @yiheng-wang-nv .

Thanks.

monai/bundle/properties.py
```diff
@@ -216,3 +221,42 @@
         BundlePropertyConfig.REF_ID: f"evaluator{ID_SEP_KEY}key_val_metric",
     },
 }
+
+MetaProterties = {
+    "version": {
+        BundleProperty.DESC: "bundle version",
+        BundleProperty.REQUIRED: False,
```
Nic-Ma, line 228: I think verison should be required property, what do you think?

Thanks.

monai/bundle/properties.py
```diff
@@ -216,3 +221,42 @@
...
+        BundlePropertyConfig.ID: f"_meta_{ID_SEP_KEY}version",
+    },
+    "monai_version": {
+        BundleProperty.DESC: "required monai version used for bundle",
+        BundleProperty.REQUIRED: False,
```
Nic-Ma, line 233: MONAI version may be required?

monai/bundle/properties.py
```diff
@@ -216,3 +221,42 @@
...
+        BundlePropertyConfig.ID: f"_meta_{ID_SEP_KEY}monai_version",
+    },
+    "pytorch_version": {
+        BundleProperty.DESC: "required pytorch version used for bundle",
+        BundleProperty.REQUIRED: False,
```
Nic-Ma, line 238: PyTorch version may be required?

monai/bundle/properties.py
```diff
@@ -216,3 +221,42 @@
...
+        BundleProperty.DESC: "required numpy version used for bundle",
+        BundleProperty.REQUIRED: False,
+        BundlePropertyConfig.ID: f"_meta_{ID_SEP_KEY}numpy_version",
+    },
+    "description": {
+        BundleProperty.DESC: "description for bundle",
+        BundleProperty.REQUIRED: False,
+        BundlePropertyConfig.ID: f"_meta_{ID_SEP_KEY}description",
+    },
+    "spatial_shape": {
```
Nic-Ma, line 251: Some bundle may have multiple input data? Generative AI bundle may not have input Tensor?

Thanks.

monai/bundle/scripts.py
```diff
@@ -417,37 +430,44 @@ def load(
...
-        net_name: if not `None`, a corresponding network will be instantiated and load the achieved weights.
-            This argument only works when loading weights.
-        net_kwargs: other arguments that are used to instantiate the network class defined by `net_name`.
+        dst_prefix: used in `copy_model_state`, `dst` key prefix, so that `dst[dst_prefix + src_key]`
+            will be assigned to the value of `src[src_key]`.
+        mapping: used in `copy_model_state`, a `{"src_key": "dst_key"}` dict, indicating that `dst[dst_prefix + dst_key]`
+            to be assigned to the value of `src[src_key]`.
+        exclude_vars: used in `copy_model_state`, a regular expression to match the `dst` variable names,
+            so that their values are not overwritten by `src`.
+        inplace: used in `copy_model_state`, whether to set the `dst` module with the updated `state_dict` via `load_state_dict`.
```
Nic-Ma, line 447: dst_prefix, mapping, exclude_vars, inplace are used in `copy_model_state`, I would suggest to avoid repeat args here, please use `copy_model_args: dict` instead.

Thanks.

monai/bundle/scripts.py
Nic-Ma: Please don't change it, it's aligned with:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/workflows.py#L41

Thanks.

monai/bundle/workflows.py
```diff
@@ -50,14 +50,16 @@ class BundleWorkflow(ABC):
 
     def __init__(self, workflow: str | None = None):
         if workflow is None:
-            self.properties = None
+            self.properties = copy(MetaProterties)
             self.workflow = None
             return
         if workflow.lower() in self.supported_train_type:
             self.properties = copy(TrainProperties)
+            self.properties.update(copy(MetaProterties))
```
Nic-Ma, line 58: No need `copy()` for the update function I think? CC @wyli .

Thanks.

monai/bundle/workflows.py
```diff
@@ -50,14 +50,16 @@ class BundleWorkflow(ABC):
...
             self.workflow = "train"
         elif workflow.lower() in self.supported_infer_type:
             self.properties = copy(InferProperties)
+            self.properties.update(copy(MetaProterties))
```
Nic-Ma, line 62: No need `copy()` for the update function I think? CC @wyli .

Thanks.

monai/bundle/workflows.py
```diff
@@ -194,11 +196,12 @@ class ConfigWorkflow(BundleWorkflow):
 
     """
 
+    @deprecated_arg_default("workflow", None, "train", since="1.3", replaced="1.4")
```
Nic-Ma, line 199: I don't see you changed any thing for the arg "workflow".

Thanks.

monai/bundle/scripts.py
```diff
@@ -417,37 +430,44 @@ def load(
...
-            maintain the consistency between these two sources, remove prefix is necessary.
+            maintain the consistency between these three sources, remove prefix is necessary.
             Therefore, if specified, downloaded folder name will remove the prefix.
         progress: whether to display a progress bar when downloading.
         device: target device of returned weights or module, if `None`, prefer to "cuda" if existing.
         key_in_ckpt: for nested checkpoint like `{"model": XXX, "optimizer": XXX, ...}`, specify the key of model
             weights. if not nested checkpoint, no need to set.
         config_files: extra filenames would be loaded. The argument only works when loading a TorchScript module,
             see `_extra_files` in `torch.jit.load` for more details.
-        net_name: if not `None`, a corresponding network will be instantiated and load the achieved weights.
```
Nic-Ma, line 428: May I know why remove this arg and net_kwargs?

Thanks.

tests/test_bundle_download.py
```diff
@@ -159,20 +166,58 @@ def test_load_weights(self, bundle_files, bundle_name, repo, device, model_file)
 
                 # load instantiated model directly and test, since the bundle has been downloaded,
                 # there is no need to input `repo`
+                _model_2 = nets.__dict__[model_name](**net_args)
```
ericspod, line 169:
```suggestion
                _model_2 = getattr(nets, model_name)(**net_args)
```

Very nitpicky but I think this would work.

monai/bundle/workflows.py
```diff
@@ -166,18 +168,18 @@ class ConfigWorkflow(BundleWorkflow):
     For more information: https://docs.monai.io/en/latest/mb_specification.html.
 
     Args:
-        run_id: ID name of the expected config expression to run, default to "run".
-            to run the config, the target config must contain this ID.
+        config_file: filepath of the config file, if it is a list of file paths, the content of them will be merged.
+        meta_file: filepath of the metadata file, if it is a list of file paths, the content of them will be merged.
+            If None, default to "configs/metadata.json", which is commonly used for bundles in MONAI model zoo.
```
ericspod, line 173:
```suggestion
        config_file: filepath of the config file, if this is a list of file paths, their contents will be merged in order.
        meta_file: filepath of the metadata file, if this is a list of file paths, their contents will be merged in order.
            If None, default to "configs/metadata.json", which is commonly used for bundles in MONAI model zoo.
```

I think the ordering is important to highlight here.

monai/bundle/workflows.py
```diff
@@ -207,10 +210,21 @@ def __init__(
         **override: Any,
     ) -> None:
         super().__init__(workflow=workflow)
+        if config_file is not None:
+            _config_path = Path(ensure_tuple(config_file)[0])
+            if _config_path.is_file():
+                config_file = config_file
```
ericspod, line 216: This appears not to do anything. Also we're not considering the case when `config_file` is a sequence of file paths since onyl the first is taken here?

vikashg:
Hi @KumoLiu, 
As you are adding `MetaProperties` I think its good to have Bundle Version as @Nic-Ma mentioned. However, it is also important to add a property on data provenance. It is essential to include information about which dataset was used to train the model. So, there should be following provisions in the `MetaProvisions`
1. it should be possible to have two different weights file for model trained on two different datasets, so a possibility to add a dataset name. Think of parallels between PyTorch when you are downloading weights for InceptionV3 you need to specify if the dataset is ImageNet or ILSVRC etc.
2. Also, datasets can have versions too, think ImageNet_v1 ImageNet_v2 as in hospitals we are continuously acquiring more data and re-training the model. So there should be a model version tied to the dataset version.

monai/bundle/scripts.py
KumoLiu: Hi @Nic-Ma,  the one I changed is not the `workflow` in BundleWorkflow, it's the workflow name you send to the `run_workflow`.
https://github.com/Project-MONAI/MONAI/blob/2daabf917a1e6afbc738d22db602f4859672a90d/monai/bundle/scripts.py#L754

Do you still think the change does not make sense?

monai/bundle/properties.py
KumoLiu: Hi @Nic-Ma, I added this property, so that in the `load` API, it can directly use the network in the bundle.
https://github.com/Project-MONAI/MONAI/blob/5da4ea6c8e05f123c5d53913037950a05daa8985/monai/bundle/scripts.py#L505

Thanks!

monai/bundle/properties.py
KumoLiu: Hi @Nic-Ma, I don't have these two properties set to required given these issues. Not sure whether I need to do that.
https://github.com/Project-MONAI/tutorials/issues/1455
https://github.com/Project-MONAI/tutorials/issues/1454
https://github.com/Project-MONAI/tutorials/issues/1453
https://github.com/Project-MONAI/tutorials/issues/1452

monai/bundle/workflows.py
KumoLiu: Thanks, @Nic-Ma and @ericspod, have updated it according to Eric's suggestion in https://github.com/Project-MONAI/MONAI/pull/6835/commits/63d3a845b92a643773e08a05482b939bbb88e20d.

monai/bundle/workflows.py
KumoLiu: Updated in https://github.com/Project-MONAI/MONAI/pull/6835/commits/63d3a845b92a643773e08a05482b939bbb88e20d.

monai/bundle/scripts.py
KumoLiu: Thanks for the suggestion, have updated it in this commit https://github.com/Project-MONAI/MONAI/pull/6835/commits/63d3a845b92a643773e08a05482b939bbb88e20d.

monai/bundle/properties.py
KumoLiu: Have updated it in https://github.com/Project-MONAI/MONAI/pull/6835/commits/63d3a845b92a643773e08a05482b939bbb88e20d.

monai/bundle/workflows.py
KumoLiu: My intention was that "train" is more commonly used, so I changed the default values for workflow; I'm not sure if you have any other concerns here.

monai/bundle/scripts.py
KumoLiu: Hi @Nic-Ma, I replaced it with a "model" arg, so that the user doesn't have to pass the name and args of the net separately, do you think it makes sense?
https://github.com/Project-MONAI/MONAI/blob/6938718b757e0366b8fc6344a1e7dbafcf588bbd/monai/bundle/scripts.py#L413

tests/test_bundle_download.py
KumoLiu: Thanks for the suggestion, updated it in https://github.com/Project-MONAI/MONAI/pull/6835/commits/63d3a845b92a643773e08a05482b939bbb88e20d.

monai/bundle/workflows.py
KumoLiu: Thanks for pointing out, there's a mistake. I deleted the line 216. 
And `_config_path` is only used to get the `config_root_path`. A sequence of `config_file` is still under consideration.
https://github.com/Project-MONAI/MONAI/blob/6938718b757e0366b8fc6344a1e7dbafcf588bbd/monai/bundle/workflows.py#L235

monai/bundle/properties.py
KumoLiu: Hi @Nic-Ma, “meta_file” is not required, do we also have to make the bundle version required?
https://github.com/Project-MONAI/MONAI/blob/2daabf917a1e6afbc738d22db602f4859672a90d/monai/bundle/workflows.py#L222

It may affect `check_properties` since `MetaProterties` is automatically added to the property when initializing a bundleworflow.
https://github.com/Project-MONAI/MONAI/blob/6938718b757e0366b8fc6344a1e7dbafcf588bbd/monai/bundle/workflows.py#L161

KumoLiu:
Hi @vikashg, thank you for the suggestion; it's a good idea if we can specify the dataset information. However, we are currently obtaining meta information based on the existing meta_data.json. Perhaps we should update what information needs to be added inside the meta_data.json to enable this.

monai/bundle/properties.py
KumoLiu: Does this cause problems if required is False? My main reference is that MONAILabel needs to get this parameter from the bundle.
https://github.com/Project-MONAI/MONAILabel/blob/43e9e2a982ecac32a34d36b5b9c620b658bb68de/monailabel/tasks/infer/bundle.py#L137

KumoLiu:
Hi @ericspod, @wyli, and @vikashg, after discussing with @Nic-Ma offline, I set "meta_file" as required in `ConfigWorkflow` and also set "version", "monai_version", "pytorch_version" and "numpy_version" as required in `MetaProperty`. For situations in these issues(https://github.com/Project-MONAI/tutorials/issues/1455; https://github.com/Project-MONAI/tutorials/issues/1454; https://github.com/Project-MONAI/tutorials/issues/1453; https://github.com/Project-MONAI/tutorials/issues/1452), let's discuss it later since now almost all the bundles have these requirements.

Do you have any other comments on this PR? Thanks!

monai/bundle/scripts.py
wyli: we may change all the `workflow: specifies the workflow type` arg to `workflow_type`, `workflow` as an input argument name seems a bit vague..

monai/bundle/scripts.py
```diff
@@ -712,11 +734,11 @@ def run(
...
-        meta_file="configs/metadata.json",
+        meta_file=None,
         init_id="initialize",
         run_id="run",
         final_id="finalize",
-        logging_file="configs/logging.conf",
+        logging_file=None,
         tracking=None,
     )
     workflow = ConfigWorkflow(
```
wyli, line 744: I guess this could now be implemented with `create_workflow`?

wyli (approved):
looks good to me, it seems `monai.bundle.scripts` module now has too many functions, perhaps it could be refactored into multiple modules/files..

monai/bundle/scripts.py
KumoLiu: Yes, have updated it. Thanks!

Nic-Ma:
I will review this PR tonight.

Thanks.

monai/bundle/properties.py
Nic-Ma: OK, I see.

monai/bundle/properties.py
Nic-Ma: OK, it can not exist for generative AI bundles.

Thanks.

monai/bundle/scripts.py
Nic-Ma: OK, so it's a breaking change, please add a `deprecated` decorator.

Thanks.

monai/bundle/scripts.py
```diff
@@ -456,10 +468,21 @@ def load(
             repo=repo,
             remove_prefix=remove_prefix,
             progress=progress,
+            args_file=args_file,
         )
+        train_config_file = bundle_dir_ / name / "configs" / f"{workflow}.json"
```
Nic-Ma, line 473: Please use `os.path.join()` to concat the path.

monai/bundle/scripts.py
Nic-Ma: Hi @wyli , OK, looks reasonable to me.

Thanks.

monai/bundle/workflows.py
Nic-Ma: In which line you changed it? I didn't any change in current code:
image

monai/bundle/workflows.py
```diff
@@ -207,10 +208,20 @@ def __init__(
         **override: Any,
     ) -> None:
         super().__init__(workflow=workflow)
+        if config_file is not None:
+            _config_path = Path(ensure_tuple(config_file)[0])
+            if _config_path.is_file():
+                config_root_path = _config_path.parent
+            else:
+                raise FileNotFoundError(f"Cannot find the config file: {config_file}.")
```
ericspod, line 216:
```suggestion
                raise FileNotFoundError(f"Cannot find the config file: {_config_path}.")
```

I see that this section only checks the first file so that it can get `config_root_path` but I think it should check that all config files exist or the class relies on the parser to do these checks which will make for less understandable errors. This way of getting `config_root_path` also assumes that all the config files are in the `configs` directory which isn't enforced elsewhere. They *should* be so I'd just note that we'd have to say any command line that refers to files outside this directory is technically wrong.

ericspod:
> looks good to me, it seems `monai.bundle.scripts` module now has too many functions, perhaps it could be refactored into multiple modules/files..

Yes maybe we should turn this into a subdirectory now and have separate files for each script. I would leave that to a later PR however and get this one completed sooner as I want to align my bundle tutorials with this.

monai/bundle/scripts.py
KumoLiu: May I ask why? "bundle_dir_" is a Path object here.

monai/bundle/scripts.py
```diff
@@ -1539,3 +1521,61 @@ def init_bundle(
...
+            "Argument `workflow_name` must be a bundle workflow class name"
+            f"or subclass of BundleWorkflow, got: {workflow_name}."
+        )
+
+    if config_file is not None:
+        workflow_ = workflow_class(config_file=config_file, **_args)
+    else:
+        workflow_ = workflow_class(**_args)
+
+    workflow_.initialize()
```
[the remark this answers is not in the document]
KumoLiu: Hi @Nic-Ma, In most cases I've seen, workflow always comes after initialize. For example, if I want to get the property, I can skip this step; I'm not sure if including it would cause issues in any way.

monai/bundle/scripts.py
KumoLiu: Thanks, have added it in this [commit](https://github.com/Project-MONAI/MONAI/pull/6835/commits/561a68e362026de630bc247f542eb8bccf852c1a).

monai/bundle/scripts.py
KumoLiu: Hi @wyli and @Nic-Ma, I have changed `worklfow` to `workflow_type` in this [commit](https://github.com/Project-MONAI/MONAI/pull/6835/commits/0039860de8b02e0e62f6aa935d7bdb88dd5c545e), could you please help check? Thanks!

monai/bundle/scripts.py
Nic-Ma: OK, I just don't want to concat path with raw "/", maybe Path has some function.

Thanks.

monai/bundle/scripts.py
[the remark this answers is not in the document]
Nic-Ma: If no obvious requirement, we should try to decouple the logic. I would suggest to only create the workflow here.
@wyli What do you think?

Thanks.

Nic-Ma (approved):
Put some minor comments inline, others look good to me.

Thanks.

monai/bundle/workflows.py
KumoLiu: Hi @ericspod, thanks for the suggestion, I just added some warnings and also checked whether all config files exist in this [commit](https://github.com/Project-MONAI/MONAI/pull/6835/commits/edb8768b9fb1ad8f0d39e4419f99373d3eaaf4b1). Could you please help review it, thanks!

monai/bundle/scripts.py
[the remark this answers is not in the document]
wyli: it seems `initialize()` is needed in almost all the use cases https://github.com/Project-MONAI/MONAI/blob/6f5005fddfaf935cc9b723be823f446a09136d11/monai/bundle/workflows.py#L316-L317

I think we can hard-code this and we can make it optional if we need the flexibility.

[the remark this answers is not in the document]
Nic-Ma: OK, sure.

monai/bundle/workflows.py
ericspod: I think that looks good thanks!

Issue #6676, opened 2023-06-30
Move MONAI Generative into core


marksgraham:
As discussed in the core meeting, [MONAI Generative](https://github.com/Project-MONAI/GenerativeModels) is going to be integrated into the core MONAI repository. The plan is to do this in stages:

1. Port the main parts of the generative code (utils, networks, metrics, losses, inferers,  engines, and associated tests). This might lead to some redundancy, but we can focus on the port for now and aim to reduce redundancies further on.
2. Once these features are available in dev, we can move the [tutorials](https://github.com/Project-MONAI/GenerativeModels/tree/main/tutorials) to the [tutorial repo](https://github.com/Project-MONAI/tutorials) - this should be easy to do, mostly involving updates imports.
3. We can then move the [model zoo](https://github.com/Project-MONAI/GenerativeModels/tree/main/model-zoo) to the [model zoo repo](https://github.com/Project-MONAI/model-zoo). 

I can start working on stage 1 soon. Any thoughts/comments?

@ericspod @Warvito @wyli 

## Progress
- [x] Losses [(PR link)](https://github.com/Project-MONAI/MONAI/pull/6729)
- [x] Metrics
- [x] Utils
- [x] Networks
- [x] DIffusion schedulers
- [x] Inferers
- [x] Engines

mingxin-zheng:
I am looking at the unit test structures [here](https://github.com/Project-MONAI/GenerativeModels/tree/main/tests). Is it reasonable to further break down 1 into smaller tasks in the testing order?

wyli:
There are also some components share the same name but have some implementation differences, for example:

https://github.com/Project-MONAI/MONAI/blob/e7fb74f32b5371bb54bff7a47447df31d06d8edf/monai/metrics/regression.py#L240

https://github.com/Project-MONAI/GenerativeModels/blob/c1ec4ed4381de90ef18061c98624fa931c42e9b6/generative/metrics/ssim.py#L28

we may take a bit more time to understand how to merge them. if the differences are significant perhaps we can just keep both versions as 'Component' and 'ComponentV1' (this may introduce breaking changes though).

marksgraham:
> I am looking at the unit test structures [here](https://github.com/Project-MONAI/GenerativeModels/tree/main/tests). It seems the unit tests are well organized and test each component "orthogonally". Is it reasonable to decompose stage 1 into smaller tasks in a similar order?

[...]

PR #6836, opened 2023-08-08
6676 port generative metrics

marksgraham:
Work towards addressing #6676. Ports all metrics to MONAI.

### Description

### Types of changes

### DIFF
--- a/monai/metrics/fid.py
+++ b/monai/metrics/fid.py
@@ -25,7 +25,7 @@ class FIDMetric(Metric):
     """
     Frechet Inception Distance (FID). The FID calculates the distance between two distributions of feature vectors.
     Based on: Heusel M. et al. "Gans trained by a two time-scale update rule converge to a local nash equilibrium."
-    https://arxiv.org/abs/1706.08500#. The inputs for this metric should be two groups of feature vectors (with format
+    https://arxiv.org/abs/1706.08500. The inputs for this metric should be two groups of feature vectors (with format
     (number images, number of features)) extracted from the a pretrained network.
 
     Originally, it was proposed to use the activations of the pool_3 layer of an Inception v3 pretrained with Imagenet.
--- a/monai/metrics/fid.py
+++ b/monai/metrics/fid.py
@@ -26,12 +26,12 @@ class FIDMetric(Metric):
     Frechet Inception Distance (FID). The FID calculates the distance between two distributions of feature vectors.
     Based on: Heusel M. et al. "Gans trained by a two time-scale update rule converge to a local nash equilibrium."
     https://arxiv.org/abs/1706.08500. The inputs for this metric should be two groups of feature vectors (with format
-    (number images, number of features)) extracted from the a pretrained network.
+    (number images, number of features)) extracted from a pretrained network.
 
     Originally, it was proposed to use the activations of the pool_3 layer of an Inception v3 pretrained with Imagenet.
     However, others networks pretrained on medical datasets can be used as well (for example, RadImageNwt for 2D and
-    MedicalNet for 3D images). If the chosen model output is not a scalar, usually it is used a global spatial
-    average pooling.
+    MedicalNet for 3D images). If the chosen model output is not a scalar, a global spatia average pooling should be
+    used.
     """
 
     def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
@@ -39,6 +39,12 @@ class FIDMetric(Metric):
 
 
 def get_fid_score(y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+    """Computes the FID score metric on a batch of feature vectors.
+
+    Args:
+        y_pred: feature vectors extracted from a pretrained network run on generated images.
+        y: feature vectors extracted from a pretrained network run on images from the real data distribution.
+    """
     y = y.double()
     y_pred = y_pred.double()
 
--- a/monai/metrics/fid.py
+++ b/monai/metrics/fid.py
@@ -34,9 +34,6 @@ class FIDMetric(Metric):
     average pooling.
     """
 
-    def __init__(self) -> None:
-        super().__init__()
-
     def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
         return get_fid_score(y_pred, y)
 
--- a/monai/metrics/fid.py
+++ b/monai/metrics/fid.py
@@ -1,0 +1,106 @@
+# Copyright (c) MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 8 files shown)


wyli (approved):
Thanks, it looks good to me, just a minor comment inline, cc @mingxin-zheng please help confirm the PR

monai/metrics/fid.py
```diff
@@ -0,0 +1,105 @@
...
+    However, others networks pretrained on medical datasets can be used as well (for example, RadImageNwt for 2D and
+    MedicalNet for 3D images). If the chosen model output is not a scalar, usually it is used a global spatial
+    average pooling.
+    """
+
+    def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+        return get_fid_score(y_pred, y)
+
+
+def get_fid_score(y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
```
mingxin-zheng, line 41: Hi @marksgraham , thanks for the PR!

Can you add some docstrings here for `get_fid_score`? Thanks!

monai/metrics/mmd.py
```diff
@@ -0,0 +1,87 @@
...
+
+    y_y = torch.mm(y, y.t())
+    y_pred_y_pred = torch.mm(y_pred, y_pred.t())
+    y_pred_y = torch.mm(y_pred, y.t())
+
+    y_y = y_y / y.shape[1]
+    y_pred_y_pred = y_pred_y_pred / y.shape[1]
+    y_pred_y = y_pred_y / y.shape[1]
+
+    # Ref. 1 Eq. 3 (found under Lemma 6)
```
mingxin-zheng, line 86: The formula used here is different than the Lemma 6 in Ref. 1
```math
uMMD^2 = \beta (\text{mean}(y \cdot y) + \text{mean}(y_{pred} \cdot y_{pred})) - \gamma \text{mean}(y_{pred} \cdot y)
```
Would it be possible to explain the cause of the discrepancy? Thanks

monai/metrics/regression.py
```diff
@@ -441,3 +441,168 @@ def compute_ssim_and_cs(
...
+    k2: float = 0.03,
+    weights: Sequence[float] = (0.0448, 0.2856, 0.3001, 0.2363, 0.1333),
+) -> torch.Tensor:
+    """
+    Args:
+        y_pred: Predicted image.
+            It must be a 2D or 3D batch-first tensor [B,C,H,W] or [B,C,H,W,D].
+        y: Reference image.
+            It must be a 2D or 3D batch-first tensor [B,C,H,W] or [B,C,H,W,D].
+                spatial_dims: number of spatial dimensions of the input images.
```
mingxin-zheng, line 534: There is an extra indentation before `spatial_dims` in this line.

monai/metrics/mmd.py
```diff
@@ -0,0 +1,87 @@
...
+    distributions. It is a non-negative metric where a smaller value indicates a closer match between the two
+    distributions.
+
+    Gretton, A., et al,, 2012.  A kernel two-sample test. The Journal of Machine Learning Research, 13(1), pp.723-773.
+
+    Args:
+        y_transform: Callable to transform the y tensor before computing the metric. It is usually a Gaussian or Laplace
+            filter, but it can be any function that takes a tensor as input and returns a tensor as output such as a
+            feature extractor or an Identity function.
+        y_pred_transform: Callable to transform the y_pred tensor before computing the metric.
```
mingxin-zheng, line 33: Can we have an example how to write the transform? Also I am wondering if the term `transform` is appropiate as we have `monai.transforms`. Thanks!

monai/metrics/mmd.py
marksgraham: Hmm you're right, comparing to equation 3:

image

Here $k(x_i,x_j)$ gives the same result as $torch.mm(y, y.t())$ - a matrix we can index using $i,j$, if the kernel function $k$ takes the inner product between $x_i,x_j$. Then given $\beta$ and $\gamma$ are fixed to 1 and 2 in the code, I'm not sure what lines 82-84 are doing. We also need to exclude the diagonal components of e.g. $y_y$ from the calculations.

I didn't write this code myself, going to contact the author and ask why it is different.

mingxin-zheng: Thank you for contacting the author @marksgraham . The formula looks like some approximation to me. Feel free to skip and close this one when you see fit, as I don't want this to block your PR.

danieltudosiu: @marksgraham @mingxin-zheng When I implemented this I based it on an open-source code available [here](https://github.com/cyclomon/3dbraingen/blob/master/Test.ipynb).

From what I remember lines 82-84 are there to account for different dimensionalities `(number of columns)` of the kernel `(y_transform/y_pred_transform)` before meaning which should be more a mean of the population `(number of rows)`. 

I remember discussing this at some point with @mjorgecardoso when I was getting peculiarly high numbers.

@marksgraham you are right that I did not get the right of the main diagonal. This could quickly be done similarly with [this one](https://jejjohnson.github.io/research_journal/appendix/similarity/mmd/). By grabbing the main diagonal and then creating a diagonal matrix that we subtract from the matrix of interest `(y_y, y_pred_y_pred, y_pred_y)`.

Also, I think the `torch.mean()` is wrong as it would mean dividing by `m*m` instead of `m*(m-1)` as shown in equation 3.   

Please let me know if I can be of more help.

monai/metrics/mmd.py
danieltudosiu: @mingxin-zheng I agree that the term transform might be ambiguous, especially in torch nomenclature. We could use `y_kernel` or `y_kernel_transform`. 

I used that nomenclature for speed when I implemented it.

marksgraham: I've renamed this `y_mapping` to better reflect that we're suppling $\varphi$ rather than the kernel $k$ as detailed [here](https://en.wikipedia.org/wiki/Kernel_method#Mathematics:_the_kernel_trick). I have also modified so only a single mapping can be passed - I don't think it makes sense to map `y` and `y_pred` differently within the MMD calculation.

monai/metrics/mmd.py
marksgraham: I've rewritten to fully match the equation being referenced. I think we should stay faithful to the original MMD implementation. I've broken it down into parts that more clearly match the calculation so it should be easier to follow.

> From what I remember lines 82-84 are there to account for different dimensionalities (number of columns) of the kernel (y_transform/y_pred_transform) before meaning which should be more a mean of the population (number of rows).

I suggest we revisit providing this as an option later, if needs be.

marksgraham: If @mingxin-zheng / @danieltudosiu could can a look to make sure you're in agreement with the changes would be great!

mingxin-zheng: Hi @marksgraham , in the update, I believe `m` equals to batch_size of `y` tensors. When the batch_size is 1, it will be a zero-division error. Is this expected?

PR #6862, opened 2023-08-13
fix for #6841 lazy argument has no effect in Compose.__call__

### LINKED ISSUE #6841: `lazy` argument has no effect in `Compose.__call__`
According to the code, only `self.lazy` is used:
https://github.com/Project-MONAI/MONAI/blob/8e99af5f96df0746d6cbcaed88feaea0e51abd56/monai/transforms/compose.py#L321-L335

There's the same issue in `OneOf`. I'm not sure if this is intended. Also, the typing of `lazy` strangely has `str`.

https://github.com/Project-MONAI/MONAI/blob/8e99af5f96df0746d6cbcaed88feaea0e51abd56/monai/transforms/compose.py#L450-L473

elitap:
Fixes #6841

### Description

added check if lazy flag is set, then use it!

### Types of changes

No tests added yet, will do asp. However unsure where. I think something similar as TestComposeExecuteWithLogging would work as the results of the flag could be made visible there. I dont like the string compares, but applying the flag is not changing the output, so thats probalby the best option anyway.

### DIFF
--- a/monai/transforms/compose.py
+++ b/monai/transforms/compose.py
@@ -319,6 +319,7 @@ class Compose(Randomizable, InvertibleTransform):
         return len(self.flatten().transforms)
 
     def __call__(self, input_, start=0, end=None, threading=False, lazy: bool | None = None):
+        _lazy = self.lazy if lazy is None else lazy
         result = execute_compose(
             input_,
             transforms=self.transforms,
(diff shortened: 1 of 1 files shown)


wyli (approved):
Thanks, it looks good to me.

PR #6871, opened 2023-08-15
Update Checkpoint Loader docs with an example

matt3o:
Related to #6687, this adds some documentation how to resume a network. I did not find this documented anywhere in MONAI and it's rather important imo. Please shortly check doc the syntax, I don't have any experience with it. The code itself is from my own repository and working.

What is potentially still missing here is some information on when to use the Checkpoint Loader. As far I understood the docs, this shall mostly be used to resume interrupted training runs.
But what is 1) with pure inference runs where the state of the trainer does not matter, only the evaluator 2) Resuming the training at epoch 200 but with a learning rate reset (e.g. DeepEdit train without clicks first for 200 epochs, then 200 epochs with clicks on top).
1 works well in my experience, 2 as well if you modify the state_dict to exclude e.g. the learning rate scheduler.

### Description

### Types of changes


wyli (approved):
Thanks, it looks good to me.

PR #6878, opened 2023-08-16
6841 follow up tests and upgrades

elitap:
Followup to #6841 

### Description

Added testcases to catch bug found in #6841. For consistency Compose implements LazyTransform.
Probably there is a better option to test...

### Types of changes

- [x] Fix tests failing due to compose beeing of type LazyTransform, mocked compose did not consider lazy option


wyli (approved):
Thanks, it looks good to me.

Issue #6804, opened 2023-07-31
enhances `Compose` input checks


wyli:
**Is your feature request related to a problem? Please describe.**
came across a similar issue as https://github.com/Project-MONAI/MONAI/discussions/6075:
```py
import monai.transforms as mt

xform = mt.Compose(
    mt.LoadImage(image_only=True),
    mt.Orientation("RAS"),
)
xform("image.nii.gz")
```

no error message but the outcomes are different from the expected usage (with the `[]` in the `Compose`)
```py
xform = mt.Compose([
    mt.LoadImage(image_only=True),
    mt.Orientation("RAS"),
])
```

this issue can be difficult to debug..

PR #6882, opened 2023-08-17
6804 enhance compose input check

elitap:
Fixes case mentioned in #6804 by checking type of second argument

### Types of changes


wyli (approved):
Thanks, it looks good to me.

PR #6899, opened 2023-08-23
WIP: Move to numpy.Generators

john-zielke-snkeos:
Fixes 6854 .

## Description

In order to move from the deprecated numpy.random.RandomState to the faster and supported numpy.random.Generator, in a first step the generation of randomness in transform and related functions is migrated (i.e. everything with self.R/ implementing Randomizable). The idea of this PR is to change the API for randomness to that of Generators. In order to preserve reproducibility, the old Random Generator should still be used by default while providing an option to switch over to the new implementation.

### Breaking changes
- The API for generating randomness inside transforms uses the new [numpy.random.Generator API](https://numpy.org/doc/stable/reference/random/new-or-different.html), custom transforms would therefore have to adapt the new functionality as well

### Types of changes

### DIFF
--- a/monai/transforms/transform.py
+++ b/monai/transforms/transform.py
@@ -26,9 +26,10 @@ from monai import config, transforms
 from monai.config import KeysCollection
 from monai.data.meta_tensor import MetaTensor
 from monai.transforms.traits import LazyTrait, RandomizableTrait, ThreadUnsafe
-from monai.utils import MAX_SEED, ensure_tuple, first
+from monai.utils import MAX_SEED, deprecated_arg, ensure_tuple, first
 from monai.utils.enums import TransformBackends
 from monai.utils.misc import MONAIEnvVars
+from monai.utils.utils_random_generator_adaptor import LegacyRandomStateAdaptor, SupportsRandomGeneration
 
 __all__ = [
     "ThreadUnsafe",
@@ -171,6 +172,9 @@ def apply_transform(
         raise RuntimeError(f"applying transform {transform}") from e
 
 
+_default_random_generator_factory: Callable[[Any | None], SupportsRandomGeneration] = LegacyRandomStateAdaptor
+
+
 class Randomizable(ThreadUnsafe, RandomizableTrait):
     """
     An interface for handling random state locally, currently based on a class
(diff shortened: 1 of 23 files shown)


wyli:
thanks for the PR, I remember that the integer random seeds are flexible and could be used for coordinating the randomised behaviour across worker processes (it seems that it's not possible to pass a random state object instead), we may want to keep the integer type random seeds settings. please see usage in dataloader: https://github.com/Project-MONAI/MONAI/blob/c564deda2c33e487d92ef0f513e89f7133b1cafe/monai/data/utils.py#L702-L705

john-zielke-snkeos:
> thanks for the PR, I remember that the integer random seeds are flexible and could be used for coordinating the randomised behaviour across worker processes (it seems that it's not possible to pass a random state object instead), we may want to keep the integer type random seeds settings. please see usage in dataloader:
@wyli Sorry I am not sure what this is referencing. What is changing that is breaking this existing behavior?

wyli:
right, I thought the proposal was about replacing both `seed` and `state` with a `generator`, I had another look, it's about replacing `state` with `generator`, in that case I don't have any concern, thanks!

monai/transforms/transform.py
```diff
@@ -183,9 +187,17 @@ class Randomizable(ThreadUnsafe, RandomizableTrait):
     the random states will be duplicated.
     """
 
-    R: np.random.RandomState = np.random.RandomState()
-
-    def set_random_state(self, seed: int | None = None, state: np.random.RandomState | None = None) -> Randomizable:
+    R: SupportsRandomGeneration = LegacyRandomStateAdaptor()  # FIXME Why is this initialized here?
```
KumoLiu, line 190: Seems we also need to update the docstring in L181 to tell users we may change to Generator near future.

monai/utils/utils_random_generator_adaptor.py
```diff
@@ -0,0 +1,144 @@
...
+        return self.random_state.normal(loc, scale, size)
+
+    def uniform(self, low: float = 0.0, high: float = 1.0, size: ShapeLike | None = None) -> NdarrayTensor:
+        return self.random_state.uniform(low, high, size)
+
+
+def handle_legacy_random_state(
+    rand_state: np.random.RandomState | None = None,
+    generator: SupportsRandomGeneration | None = None,
+    return_legacy_default_random: bool = False,
```
KumoLiu, line 130: Can the user directly change to the Generator by setting the flag?

KumoLiu (commented):
Thanks for the PR, overall looks good to me. Several minor comments inline, feel free to ignore them.

monai/utils/utils_random_generator_adaptor.py
john-zielke-snkeos: This function is more of a internal utility function (should probably be prefixed with _) during the migration to be used in the utility functions that accept a random_state (e.g. get_random_patch). The return_legacy_default_random is to return the default value as it is currently done (If there ends up being no use case where the default is not required, the parameter could be removed as well).

To change the default generator for transforms the place would be to change the [_default_random_generator_factory](https://github.com/john-zielke-snkeos/MONAI/blob/c564deda2c33e487d92ef0f513e89f7133b1cafe/monai/transforms/transform.py#L175) variable to numpy.random.default_rng

PR #6908, opened 2023-08-27
Update utils.py

bf2harven:
In some edge cases the inferer outputs:
expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!

The following change ensures the buffer and w_t are on the same device.

Fixes # .

### Description

### Types of changes

### DIFF
--- a/monai/inferers/utils.py
+++ b/monai/inferers/utils.py
@@ -278,8 +278,8 @@ def sliding_window_inference(
                 else:
                     output_image_list[0][o_slice] += sw_device_buffer[0].to(device=device)
             else:
-                sw_device_buffer[ss] *= w_t
                 sw_device_buffer[ss] = sw_device_buffer[ss].to(device)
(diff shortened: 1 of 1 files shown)


monai/inferers/utils.py
```diff
@@ -278,8 +278,8 @@ def sliding_window_inference(
                 else:
                     output_image_list[0][o_slice] += sw_device_buffer[0].to(device=device)
             else:
-                sw_device_buffer[ss] *= w_t
                 sw_device_buffer[ss] = sw_device_buffer[ss].to(device)
+                sw_device_buffer[ss] *= w_t
```
wyli, line 282: this `sw_device_buffer[ss] *= w_t` should happen on `sw_device`... I'm curious how you run into issues, maybe somehow the `predictor` returned a `device` tensor by mistake? it should return an output on `sw_device`.

PR #6910, opened 2023-08-29
Add bundle root directory to Python search directories automatically

### LINKED ISSUE #6722: Add scripts directory to Python search directories automatically and avoid using PYTHONPATH

KumoLiu:
Fixes #6722 .

### Description
Add scripts directory to Python search directories automatically in the `run` function in `ConfigWorkflow`.

### Types of changes


KumoLiu:
Hi @ericspod, could you please help review whether this one can help fix #6722? I added the scripts folder in the sys.path before running. This one should be merged after #6835.
cc @wyli @Nic-Ma

ericspod:
This looks in the right direction. We should have a test for it which creates a bundle with a scripts directory containing `__init__.py` with a single function definition. The bundle would call this function which would just print something to stdout and we'd check that value appears in the output. This test should use `subprocess` to run the bundle as a command line program to ensure a new Python runtime is created which has its own isolated environment.

KumoLiu:
> This looks in the right direction. We should have a test for it which creates a bundle with a scripts directory containing `__init__.py` with a single function definition. The bundle would call this function which would just print something to stdout and we'd check that value appears in the output. This test should use `subprocess` to run the bundle as a command line program to ensure a new Python runtime is created which has its own isolated environment.

Agreed, will add the test ASAP, thanks!

monai/bundle/workflows.py
```diff
@@ -285,6 +286,10 @@ def run(self) -> Any:
         Run the bundle workflow, it can be a training, evaluation or inference.
 
         """
+        _bundle_root_path = (
+            self.config_root_path.parent if self.config_root_path.name == "configs" else self.config_root_path
+        )
+        sys.path.append(str(_bundle_root_path))
```
wyli, line 292: we may consider `sys.path.insert(0, path)` to make the newly added path higher priority. please also update the docstring about this sys path change in this function.

ericspod: When setting PYTHONPATH these values get placed in `sys.path` after an initial empty string but before everything else, so perhaps `sys.path.insert(1, path)` would make more sense. Unless names clash anywhere it wouldn't actually matter. let's leave this for now since users can always set PYTHONPATH to prepend things, but yes we should also document th behaviour in the docstring.

wyli (approved):
Thanks, it looks good to me.

PR #6913, opened 2023-08-29
Mask usernames in print_config function (#6471)

### LINKED ISSUE #6471: print_config to hide user info
https://github.com/Project-MONAI/MONAI/blob/25c9c3951858bd60b4d02bcaca7fff9c4236bff7/monai/config/deviceconfig.py#L101

`print_config` is extensively used in the notebooks, and would be great to make sure the output not leaking any user-specific information such as `/home/username`

anudeepadi:
Fixes #6471 

### Description

### Types of changes


wyli (approved):

ericspod:
This will work almost always but if home directories are in other locations, eg. networked directories, the regex patterns won't match. I'd suggest a more thorough approach is to get the username with `getpass.getuser()` or `os.path.basename(os.path.expanduser("~"))` then replace all instances of that string in the output with "<username>". This will catch more places but will miss paths that might contain other identifying names, but these should be very rare.

Issue #6552, opened 2023-05-24
Network Arch Code refactoring to allow replacing the last FC and add a freeze core parameter 


AHarouni:
**Describe the solution you'd like**
Once there is a good model out there (either we trained it, from model zoo, or other sources), we would like to easily replace the last FC layer with 104 outputs for total segmentor for example with a different number of outputs. user would get errors when loading the checkpoint and need to do some network surgery.

Second request is usually user would want to load the check point then freeze the core parameters and just train the newly added FC. 

**Describe alternatives you've considered**
Code should be refactored to allow to give the last FC a different name. with that the check point would be loaded without any errors 
as in https://github.com/Project-MONAI/MONAILabel/issues/1298 I had to write my own segresnet below to change the name of the last layer 
```
class SegResNetNewFC(SegResNet):
    def __init__(self,spatial_dims: int = 3,
            init_filters: int = 8,out_channels: int = 2,
            **kwargs,):
        super().__init__(spatial_dims=spatial_dims , init_filters=init_filters
                         ,use_conv_final=False,**kwargs)
        self.conv_final=None # delete this layer
        self.conv_final_xx = self._make_final_conv(out_channels)

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        beforeFC = super().forward(x) ## will build everything without the last FC since we init with use_conv_final =False
        x = self.conv_final_xx(beforeFC)
        return x
```

**Additional context**
We do have code to frezze the layers but it needs to be a utility that calls into a function of the network to get FC layer name. For this we should have an interface for network models to implement so this util function can be called 
I have it hacked as 
```
  def optimizer(self, context: Context):
      freezeLayers(self._network, ["conv_final_xx"], True)
      context.network = self._network
      return torch.optim.AdamW(context.network.parameters(), lr=1e-4, weight_decay=1e-5) ## too small

def freezeLayers(network, exclude_vars, exclude_include=True):
    src_dict = get_state_dict(network)
    to_skip={''} # initalize single string dict 
    for exclude_var in exclude_vars:
        s = {s_key for s_key in src_dict if exclude_var and re.compile(exclude_var).search(s_key)}
        to_skip.update(s)
    to_skip.remove('')  ## remove the empty string
    logger.error(f"--------------------- layer freeze with {exclude_include=}")
    for name, para in network.named_parameters():
        if exclude_include:
            if name not in to_skip:
                logger.info(f"------------ freezing {name} size {para.shape}")
[...]

PR #6917, opened 2023-08-31
Support regular expression  in the mapping arg of `copy_model_state`

KumoLiu:
Part of #6552.

### Description
After PR #6835, we have added `copy_model_args` in the `load` API which can help us update the state_dict flexibly.
https://github.com/KumoLiu/MONAI/blob/93a149a611b66153cf804b31a7b36a939e2e593a/monai/bundle/scripts.py#L397

Given this [issue](https://github.com/Project-MONAI/MONAI/issues/6552), we need to be able to filter the model's weights flexibly.
In `copy_model_state`, we already have a "mapping" arg, the filter will be more flexible if we can support regular expression in the mapping. This PR mainly added the support for regular expression for "mapping" arg. 

In the [example](https://github.com/Project-MONAI/MONAI/issues/6552#issuecomment-1587582734) in this [issue](https://github.com/Project-MONAI/MONAI/issues/6552), after this PR, we can do something like:
```
exclude_vars = "encoder.mask_token|encoder.norm.weight|encoder.norm.bias|out.conv.conv.weight|out.conv.conv.bias"
mapping={"encoder.layers(.*).0.0.": "swinViT.layers(.*).0."}
dst_dict, updated_keys, unchanged_keys = copy_model_state(
       model, ssl_weights, exclude_vars=exclude_vars, mapping=mapping
)
```

Additionally, based on the comments of Eric [here](https://github.com/Project-MONAI/MONAI/issues/6552#issuecomment-1697404590), I totally agree, we could add a handler to make the pipeline easier to implement, but perhaps this task is no need to set as a "BundleTodo" for MONAIv1.3 but as an enhancement for MONAI near future.
What do you think? @ericspod @wyli @Nic-Ma 

### Types of changes

### DIFF
--- a/monai/networks/utils.py
+++ b/monai/networks/utils.py
@@ -531,11 +531,13 @@ def copy_model_state(
             updated_keys.append(dst_key)
     for s in mapping if mapping else {}:
         dst_key = f"{dst_prefix}{mapping[s]}"
-        if dst_key in dst_dict and dst_key not in to_skip:
-            if dst_dict[dst_key].shape != src_dict[s].shape:
-                warnings.warn(f"Param. shape changed from {dst_dict[dst_key].shape} to {src_dict[s].shape}.")
-            dst_dict[dst_key] = src_dict[s]
-            updated_keys.append(dst_key)
+        src_keys = sorted(set(s_key for s_key in src_dict if s_key not in to_skip and re.compile(s).search(s_key)))
+        dst_keys = sorted(set(d_key for d_key in dst_dict if d_key not in to_skip and re.compile(dst_key).search(d_key)))
+        for _src_key, _dst_key in zip(src_keys, dst_keys):
+            if dst_dict[_dst_key].shape != src_dict[_src_key].shape:
+                warnings.warn(f"Param. shape changed from {dst_dict[_dst_key].shape} to {src_dict[_src_key].shape}.")
+            dst_dict[_dst_key] = src_dict[_src_key]
+            updated_keys.append(_dst_key)
 
     updated_keys = sorted(set(updated_keys))
     unchanged_keys = sorted(set(all_keys).difference(updated_keys))
--- a/tests/test_copy_model_state.py
+++ b/tests/test_copy_model_state.py
@@ -134,7 +134,7 @@ class TestModuleState(unittest.TestCase):
         model_two.to(device_1)
         # test weight map
         model_dict, ch, unch = copy_model_state(
-            model_one, model_two, mapping={"layer_1.weight": "layer.weight", "layer_1.bias": "layer_1.weight"}
+            model_one, model_two, mapping={"layer_1.weight": "^layer.weight", "layer_1.bias": "layer_1.weight"}
         )
         model_one.load_state_dict(model_dict)
         x = np.random.randn(4, 10)


monai/networks/utils.py
```diff
@@ -531,11 +531,13 @@ def copy_model_state(
...
     for s in mapping if mapping else {}:
         dst_key = f"{dst_prefix}{mapping[s]}"
-        if dst_key in dst_dict and dst_key not in to_skip:
-            if dst_dict[dst_key].shape != src_dict[s].shape:
-                warnings.warn(f"Param. shape changed from {dst_dict[dst_key].shape} to {src_dict[s].shape}.")
-            dst_dict[dst_key] = src_dict[s]
-            updated_keys.append(dst_key)
+        src_keys = sorted({s_key for s_key in src_dict if s_key not in to_skip and re.compile(s).search(s_key)})
+        dst_keys = sorted({d_key for d_key in dst_dict if d_key not in to_skip and re.compile(dst_key).search(d_key)})
+        for _src_key, _dst_key in zip(src_keys, dst_keys):
```
wyli, line 536: this `zip` assumes 1 to 1 mapping for the first N keys, I can see in practice there could be missing keys in the middle of `dst_keys` or `src_keys` and will cause mismatches... in my opinion defining this kind of search pattern is complicated, but if there's a use case I'm fine with this new feature.

KumoLiu: Thanks, Wenqi, I know your concern. I didn't find a good way to do such a thing when we have some complex rules to match weights such as the https://github.com/Project-MONAI/MONAI/issues/6552#issuecomment-1587582734,  
Maybe we should leave these filter logic to users by just exposing a `filter` arg.
Or maybe I can add some warning here when the length of `dst_keys` and `src_keys` mismatch.

Anyway, let's wait and see what others say.

KumoLiu:
According to Wenqi's comments and https://github.com/Project-MONAI/MONAI/issues/6552#issuecomment-1593318890, we can leave `mapping` doesn't support regular expression here, but add a `filter_func` in `copy_model_state` which defaults to `None`. 

Since `copy_model_state` is also used in `CheckpointLoader`, we can also easily pass our `filter_func` into it to help filter weights, researchers can also provide some common-used filters for users to use.
https://github.com/Project-MONAI/MONAI/blob/ef7debe47603f882513b255e449b26659766b64f/monai/handlers/checkpoint_loader.py#L138

```
def filter_swinunetr(k, v):
    if k in [
        "out.conv.conv.weight",
        "out.conv.conv.bias",
    ]:
        return None

    if k[:8] == "encoder.":
        if k[8:19] == "patch_embed":
            new_key = "swinViT." + k[8:]
        else:
            new_key = "swinViT." + k[8:18] + k[20:]

        return new_key, v
    else:
        return None

dst_dict, updated_keys, unchanged_keys = copy_model_state(
       model, ssl_weights, filter_func=filter_swinunetr
)
```
If you think this works, I can continue with this PR. Thanks.

monai/networks/nets/swin_unetr.py
```diff
@@ -1055,3 +1055,34 @@ def forward(self, x, normalize=True):
         x4 = self.layers4[0](x3.contiguous())
         x4_out = self.proj_out(x4, normalize)
         return [x0_out, x1_out, x2_out, x3_out, x4_out]
+
+
+def filter_swinunetr(key, value):
```
wyli, line 1060: perhaps in the docstring we can mention that this function is typically used with `copy_model_state`

wyli (approved):
looks good to me, the test case show the usage,cc @ericspod @Nic-Ma @vikashg  please share any further comments

monai/networks/nets/swin_unetr.py
```diff
@@ -1055,3 +1055,35 @@ def forward(self, x, normalize=True):
...
```
ericspod, line 1060: We should have tests with this function. If possible a code example of using this with a default network would help as well since just mentioning `copy_model_state` doesn't really say enough about use case.

ericspod: Looks good, thanks!

ericspod (approved):

monai/networks/nets/swin_unetr.py
KumoLiu: Thanks Wenqi, have mentioned it in the docstring.

PR #6922, opened 2023-08-31
Robust Username Masking in print_config

### LINKED ISSUE #6914: robust username masking in print_config
https://github.com/Project-MONAI/MONAI/blob/6f5005fddfaf935cc9b723be823f446a09136d11/monai/config/deviceconfig.py#L103-L110
if home directories are in other locations, eg. networked directories, the regex patterns won't match. I'd suggest a more thorough approach is to get the username with `getpass.getuser()` or `os.path.basename(os.path.expanduser("~"))` then replace all instances of that string in the output with "<username>". This will catch more places but will miss paths that might contain other identifying names, but these should be very rare.

_Originally posted by @ericspod in https://github.com/Project-MONAI/MONAI/issues/6913#issuecomment-1699093111_

anishrajan25:
Fixes #6914.

### Description

### Types of changes

### DIFF
--- a/monai/config/deviceconfig.py
+++ b/monai/config/deviceconfig.py
@@ -11,6 +11,7 @@
 
 from __future__ import annotations
 
+import getpass
 import os
 import platform
 import re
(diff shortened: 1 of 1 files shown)


wyli (approved):
thanks! please sign the commit https://github.com/Project-MONAI/MONAI/pull/6922/checks

PR #6934, opened 2023-09-04
fix #6685

### LINKED ISSUE #6685: cropforeground with margin>0

KumoLiu:
Fixes # .

### Description
Change the default value of `allow_smaller` to `False` where `CropForeground` is used to avoid FutureWarning.

### Types of changes


monai/metrics/utils.py
```diff
@@ -176,7 +176,7 @@ def get_mask_edges(
         else:  # pytorch subvoxel, maybe on gpu, but croppad boolean values on GPU is not supported
             seg_pred, seg_gt, or_vol = convert_to_tensor(channel_first, dtype=torch.float16)
         cropper = CropForegroundD(
-            ["pred", "gt"], source_key="src", margin=1, allow_smaller=False, start_coord_key=None, end_coord_key=None
+            ["pred", "gt"], source_key="src", margin=1, allow_smaller=True, start_coord_key=None, end_coord_key=None
```
KumoLiu, line 179: Hi @wyli, I just changed this value back to `True` since it was `True` before. And the meaning of `allow_smaller` doesn't change so if we changed it here, the meaning will be opposite.
https://github.com/Project-MONAI/MONAI/pull/6736#issuecomment-1639896622

wyli (approved):
Thanks, it looks good to me.

PR #6950, opened 2023-09-06
Avoid breaking change in creating `BundleWorkflow`

KumoLiu:
Fixes # .

### Description
Avoid breaking changes introduced by https://github.com/Project-MONAI/MONAI/pull/6835
- when creating `BundleWorkflow`
- when using `load` API, add `return_state_dict` when `model` and `net_name` are both `None`.

### Types of changes

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -533,7 +533,11 @@ def load(
 
     model.to(device)  # type: ignore
 
-    copy_model_state(dst=model, src=model_dict if key_in_ckpt is None else model_dict[key_in_ckpt], **copy_model_args)  # type: ignore
+    copy_model_state(  # type: ignore
+        dst=model,
+        src=model_dict if key_in_ckpt is None else model_dict[key_in_ckpt],
+        **copy_model_args
+    )
 
     return model
 
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -533,10 +533,8 @@ def load(
 
     model.to(device)  # type: ignore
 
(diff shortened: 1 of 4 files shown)


wyli (approved):
Thanks, it looks good to me.

monai/bundle/scripts.py
```diff
@@ -477,34 +491,52 @@ def load(
...
+                args_file=args_file,
+                config_file=str(bundle_config_file),
+                workflow_type=workflow_type,
+                **_net_override,
+            )
+        else:
+            warnings.warn(f"Cannot find the config file: {bundle_config_file}, return state dict instead.")
+            return model_dict
+        if _workflow is not None:
+            if getattr(_workflow, "network_def") is None:
```
ericspod, line 524:
```suggestion
            if not hasattr(_workflow, "network_def"):
```
Using `getattr` with no default value will raise an exception, I don't think that was intended here?

KumoLiu: Yeah, thanks!

monai/bundle/scripts.py
```diff
@@ -395,8 +397,10 @@ def load(
     workflow_name: str | BundleWorkflow | None = None,
     args_file: str | None = None,
     copy_model_args: dict | None = None,
+    return_state_dict: bool = True,
+    net_override: dict = {},
```
wyli, line 401: there's an error issue with this mutable default, I think it could default to `None` and in the function add something like `if net_override is None: net_override = {}`

KumoLiu: Thanks, have updated it to `None`.

Issue #6721, opened 2023-07-14
Improve Syntax of Sigils


wyli:
Using the # character as a sigil or separator is slightly cumbersome because it's also the comment character in Python and Bash. Values mentioned on the command line need to be put in quotes, eg. `'--trainer#max_epochs' 5`. Perhaps deprecating this and using : or :: instead?

https://github.com/Project-MONAI/MONAI/blob/3b6b11a815b2de9e2b1bc1f1dbbca4e66189458f/monai/bundle/utils.py#L26-L29

KumoLiu:
How about `^^`? I didn't find it being used [here](https://www.w3schools.com/python/python_operators.asp).

wyli:
Follow-up of our previous discussions, the syntax change from `#` to `::` seems to be technically feasible, but it'll affect many related codebases, FL, auto3dseg, model-zoo...

One way to make it backward compatible is to provide a system environment variable https://github.com/Project-MONAI/MONAI/pull/6952 (by default it's non-breaking)

I think we should have another discussion about how to proceed.
cc @ericspod @Nic-Ma @KumoLiu 


Please see a list of related code to this syntax change:
<details>

```
monai/auto3dseg/analyzer.py:100:            nested_key: str that has format of 'key1#0#key2'.
monai/auto3dseg/analyzer.py:105:            raise ValueError("Nested_key input format is wrong. Please ensure it is like key1#0#key2")
monai/fl/client/monai_algo.py:78:    if "validate#handlers" in parser:
monai/fl/client/monai_algo.py:79:        for h in parser["validate#handlers"]:
monai/bundle/config_parser.py:144:            id: id of the ``ConfigItem``, ``"#"`` in id are interpreted as special characters to
monai/bundle/config_parser.py:147:                For example: ``"xform#5"``, ``"net#channels"``. ``""`` indicates the entire ``self.config``.
monai/bundle/config_parser.py:170:            id: id of the ``ConfigItem``, ``"#"`` in id are interpreted as special characters to
monai/bundle/config_parser.py:173:                For example: ``"xform#5"``, ``"net#channels"``. ``""`` indicates the entire ``self.config``.
monai/bundle/config_parser.py:230:        For example, ``parser.update({"train#epoch": 100, "train#lr": 0.02})``
monai/bundle/config_parser.py:275:            id: id of the ``ConfigItem``, ``"#"`` in id are interpreted as special characters to
[...]

PR #6955, opened 2023-09-06
6721 bundle syntax # as alias of ::

wyli:
for #6721

### Description

compatible syntax by normalising any `#` in ids to `::`

### Types of changes


ericspod:
This is a simple solution but it appears to work just fine, we can assume there won't be comments ever so there's no issue removing #. An alternative is to use `re.split(f"#|{ID_SEP_KEY}", str(id))` or something like this but it doesn't seem necessary.

monai/bundle/reference_resolver.py
```diff
@@ -31,7 +31,7 @@ class ReferenceResolver:
     The IDs must be unique within this set. A string in ``ConfigItem``
     starting with ``@`` will be treated as a reference to other ``ConfigItem`` objects by ID.
     Since ``ConfigItem`` may have a nested dictionary or list structure,
-    the reference string may also contain a ``#`` character to refer to a substructure by
```
KumoLiu, line 34: May also update L55.

KumoLiu (approved):
Thanks, looks good to me.

PR #6958, opened 2023-09-07
Add `download_large_files` in `monai/bundle/scripts.py`

### LINKED ISSUE #6362: Bundle Download Large Files
Many bundles have the `large_files.yml` file specifying large files (specifically pre-trained weights). The bundles in the GenerativeModels repo are kept separately currently from those in the model zoo so the bundle download program doesn't work quite as expected so users would want to use bundles from a checkout of the repo. These bundles will be moved over but other users may want their own separate hosting for bundles so I feel there is still a need to make it easier to start using a bundle from a checkout. 

One step for this is to make it easier to download these large files on the command line or programmatically. This would help people working with checkout code and make it easier to get bundles and their data into Colab and other environments. 

I would also suggest changes to the `download` command we have for bundles to work with a URL pointing to a directory in an arbitrary repo. This is much more complicated now but would allow a user to provide [this](https://github.com/Project-MONAI/GenerativeModels/tree/main/model-zoo/models/brain_image_synthesis_latent_diffusion_model) as the URL and get that whole bundle large files and all. Again this would be helpful in Colab and elsewhere.

I would like the code for downloading the large files [here](https://github.com/Project-MONAI/model-zoo/blob/131b0da3f540ed13a8a1c02d8395e92a682d47cf/ci/utils.py#L93) to be put into a command in the `monai/bundle/scripts.py` file so that users can run this and get the files they need to use the bundle.

One has to manually download weights and other files currently which isn't impossible but adds an extra layer of complexity to tell people to look into the yaml file. 

**Additional context**
This was from experience trying to get going quickly with a bundle. We do also need a way of programmatically getting a bundle's network with pre-trained weights like `torch.hub.load` though this will require some added specification in the bundle.

KumoLiu:
Fixes #6362 .

### Description

Add `download_large_files` in `monai/bundle/scripts.py`.
Refer to the one in https://github.com/Project-MONAI/model-zoo/blob/131b0da3f540ed13a8a1c02d8395e92a682d47cf/ci/utils.py#L93.

### Types of changes

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -1641,14 +1641,13 @@ def download_large_files(bundle_path: str, large_file_name: str | None = None) -
 
     """
     if large_file_name is None:
-        for large_file_type in [".yml", ".yaml", ".json"]:
-            large_file_name = "large_files" + large_file_type
-            large_file_path = os.path.join(bundle_path, large_file_name)
-            if os.path.exists(large_file_path):
-                break
+        large_file_path = list(Path(bundle_path).glob("large_files*"))
+        large_file_path = list(filter(lambda x: x.suffix in [".yml", ".yaml", ".json"], large_file_path))
+        if len(large_file_path) == 0:
+            raise FileNotFoundError("Cannot find the large files.")
 
     parser = ConfigParser()
-    parser.read_config(os.path.join(bundle_path, large_file_name))
+    parser.read_config(large_file_path)
     large_files_list = parser.get()["large_files"]
     for lf_data in large_files_list:
         lf_data["fuzzy"] = True
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -1586,3 +1586,43 @@ def create_workflow(
     workflow_.initialize()
 
     return workflow_
+
+
+def download_large_files(bundle_path: str, large_file_name: str | None = None) -> None:
+    """
+    This utility allows you to download large files from a bundle. It supports file suffixes like ".yml", ".yaml", and ".json".
+    If you don't specify a `large_file_name`, it will automatically search for large files among the supported suffixes.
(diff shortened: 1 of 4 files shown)


monai/bundle/scripts.py
```diff
@@ -1619,3 +1619,42 @@ def create_workflow(
     workflow_.initialize()
 
     return workflow_
+
+
+def download_large_files(bundle_path: str, large_file_name: str | None = None) -> None:
```
wyli, line 1624: I think the bundle path could be 'os.getcwd()' if it's not specified?

monai/bundle/scripts.py
```diff
@@ -1619,3 +1619,42 @@ def create_workflow(
...
+    Args:
+        bundle_path: The path to the bundle where the files are located.
+        large_file_name: (Optional) The name of the large file to be downloaded.
+
+    """
+    if large_file_name is None:
+        large_file_path = list(Path(bundle_path).glob("large_files*"))
+        large_file_path = list(filter(lambda x: x.suffix in [".yml", ".yaml", ".json"], large_file_path))
+        if len(large_file_path) == 0:
+            raise FileNotFoundError("Cannot find the large files.")
```
wyli, line 1647: The error message could be more specific, display the folder searched and no large_files.yml/JSON found. The user may ignore the error if the files exist or add the missing large_files config

wyli (approved):
Thanks, it looks good to me.

monai/bundle/scripts.py
KumoLiu: Thanks, have updated it in the latest commit.

monai/bundle/scripts.py
KumoLiu: Good point, have updated it in the latest commit. Thanks!

PR #6965, opened 2023-09-08
Improve `ckpt_export `

### LINKED ISSUE #6953: Improve Converting to Torchscript
Given a bundle with saved weights in the `models` directory and a network that's compatible with Torchscript, the command line to use `ckpt_export` gets rather long:

```bash
python -m monai.bundle ckpt_export --net_id network_def --bundle_root . --meta_file configs/metadata.json --config_file configs/inference.json --filepath models/model.ts  --ckpt_file models/model.pt --key_in_ckpt model
```

Would it be possible to simplify this by providing defaults for `meta_file` and others, or try to guess, if the loaded checkpoint is a dictionary, what `key_in_ckpt` would be? 

If the command line would be:

```bash
python -m monai.bundle ckpt_export --bundle_root . --config_file configs/inference.json --filepath models/model.ts
```
with everything omitted from the previous version replaced with defaults or guesses. If `key_in_ckpt` is needed that's fine but trying to guess and saying in the output what was guessed would often work. Similarly a defaults of `configs/metadata.json` for the meta file would usually work.

Keep using the old command line, it does work so this is about usability.

**Additional context**
The commands need more tutorial materials which is a work in progress.

KumoLiu:
Fixes #6953

### Types of changes

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -1287,7 +1287,7 @@ def ckpt_export(
     filepath_ = os.path.join(bundle_root_, "models", "model.ts") if filepath_ is None else filepath_
     ckpt_file_ = os.path.join(bundle_root_, "models", "model.pt") if ckpt_file_ is None else ckpt_file_
     if not os.path.exists(ckpt_file_):
-        raise FileNotFoundError(f"ckpt_file in {ckpt_file_} does not exist, please specify it.")
+        raise FileNotFoundError(f'Checkpoint file "{ckpt_file_}" not found, please specify it in argument "ckpt_file".')
     if os.path.exists(meta_file_):
         parser.read_meta(f=meta_file_)
 
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -1291,7 +1291,7 @@ def ckpt_export(
         try:
             parser.get_parsed_content(net_id_)
         except ValueError as e:
-            raise ValueError(f"Default net_id: network_def in {config_file_} does not exist. {e}")
+            raise ValueError(f"Default net_id: network_def in {config_file_} does not exist.") from e
 
     # the rest key-values in the _args are to override config content
     for k, v in _args.items():
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -1214,9 +1214,13 @@ def ckpt_export(
 
     Args:
         net_id: ID name of the network component in the config, it must be `torch.nn.Module`.
+            Default to "network_def".
         filepath: filepath to export, if filename has no extension it becomes `.ts`.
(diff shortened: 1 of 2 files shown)


monai/bundle/scripts.py
```diff
@@ -1191,10 +1191,10 @@ def onnx_export(
 
 
 def ckpt_export(
-    net_id: str | None = None,
-    filepath: PathLike | None = None,
-    ckpt_file: str | None = None,
-    meta_file: str | Sequence[str] | None = None,
+    net_id: str | None = "network_def",
```
KumoLiu, line 1194: I didn't add `deprecated_arg_default` here since no breaking change will be introduced, what do you think?
cc @wyli @ericspod

wyli: sure, please update the docstring to clarify the default value. also, for os compatibility, the path should be `os.path.join('models', 'model.ts')` instead of `models/model.ts`?

KumoLiu: Good points, have updated docstring and the path. Thanks.

monai/bundle/scripts.py
```diff
@@ -1191,10 +1191,10 @@ def onnx_export(
...
+    filepath: PathLike | None = "models/model.ts",
```
wyli, line 1195:
```suggestion
    filepath: PathLike | None = None,
```
is it possible to still use `None` here? and then the default logic could be:
```
meta_file_ = os.path.join(bundle_root_, "configs", "metadata.json") if meta_file_ is None " else meta_file_
```

if None doesn't work we can consider `""`

KumoLiu: Sure, updated it in the latest commit.

wyli (approved):
Thanks, it looks good to me.

monai/bundle/scripts.py
```diff
@@ -1275,9 +1281,22 @@ def ckpt_export(
...
+    if not os.path.exists(ckpt_file_):
+        raise FileNotFoundError(f"ckpt_file in {ckpt_file_} does not exist, please specify it.")
+    if os.path.exists(meta_file_):
         parser.read_meta(f=meta_file_)
 
+    net_id_ = "network_def" if net_id_ is None else net_id_
+    try:
+        parser.get_parsed_content(net_id_)
+    except ValueError as e:
+        raise ValueError(f"net_id: network_def in {config_file_} does not exist.") from e
```
ericspod, line 1298:
```suggestion
        raise ValueError(f'Network definition "{net_id_}" cannot be found in "{config_file_}", specify name with argument "net_id".') from e
```

ericspod (approved):
Looks good, I had a few minor changes for error messages but otherwise this looks like a good solution.

PR #6981, opened 2023-09-13
Implement distance_transform_edt and the DistanceTransformEDT transform

matt3o:
Related to https://github.com/Project-MONAI/MONAI/discussions/6845, this commits adds an EDT distance transform to MONAI.
Most importantly this enables GPU based distance transforms which lead to a huge speedup.

### Description

### Types of changes


matt3o:
If that PR is accepted I would change the code in following functions to use the new code:
`compute_multi_instance_mask() `in `monai/apps/pathology`
`get_surface_distance()` in `monai/metrics`. It also offers cdt, which I can't change since there is no GPU implementation available

That should speed up at least the SurfaceDice a lot, currently it is rather slow on big volumes.

Please give feedback if that would be fine as well, thanks!

wyli:
looks great, please add the unit test file names to this list to skip them in the minimal testing environments https://github.com/Project-MONAI/MONAI/blob/c22a2bda776d625009a8589c890767aefedcf5c2/tests/min_tests.py#L129

matt3o:
I just added the addtional `DistanceTransformEDTd` transform. Also @wyli shall I implement the two changes above that I mentioned? Or shall we keep it separate and I create a second PR?

wyli:
Please keep them separate, modifying existing algorithm will require more checks to prevent regression...

matt3o:
@wyli How can I fix the optional import error? 

`monai.utils.module.OptionalImportError: import cupy (No module named 'cupy').`

wyli:
yes, I tried running it locally, the parameter expand decorator should be put before the skipping decorator:
```
    @parameterized.expand(SAMPLING_TEST_CASES)
    @skip_if_no_cuda
    @unittest.skipUnless(HAS_CUPY, "CuPy is required.")
    @unittest.skipUnless(momorphology, "cuCIM transforms are required.")
```
also `test_sampling` might need some decorators to skip as well.

monai/transforms/post/array.py
```diff
@@ -936,3 +938,30 @@ def __call__(self, image: NdarrayOrTensor) -> torch.Tensor:
...
+
+    def __init__(self, sampling: None | float | list[float] = None, force_scipy: bool = False) -> None:
+        super().__init__()
+        self.force_scipy = force_scipy
+        self.sampling = sampling
+
+    def __call__(self, img: NdarrayOrTensor) -> NdarrayOrTensor:
+        """
+        Args:
+            img: shape must 2D or 3D for cupy, otherwise no restrictions
```
wyli, line 962: please make sure it can work with this input shape assumption (channel first and the distances are computed within each channel): https://github.com/Project-MONAI/MONAI/blob/f76957eb22c924302666eb9c60f0ac3e5eb4edc6/monai/transforms/post/array.py#L320-L324

matt3o: Sadly CuPy cannot handle 4D inputs, see the tests about that. I made the docs a bit more clear about that and added some text there.
CuPy only accepts 2D and 3D inputs, whereas scipy can handle any input. The transform however is configurable to force the scipy implementation so I guess people can choose scipy if they have a different shaped volume.
Is that good enough?

wyli: Ideally these should be applied channel wise, that's the shape format most of the monai transform follow. please see the logic here using a for loop and then stack the outputs: https://github.com/Project-MONAI/MONAI/blob/56ca2249541bf18c7d3b4f0218db18767ee8bf65/monai/transforms/intensity/array.py#L685

john-zielke-snkeos:
Hey, I just noticed this PR and we seem to be working on similar functionality regarding the euclidean distance transform. So before any merging, we should make sure that this PR and #7008 don't duplicate functionality and make the changes in a consistent implementation.

wyli:
it seems #7008 could be refactored to use the utility function defined in this PR... I'd suggest @john-zielke-snkeos please also help review this PR and we merge this first, what do you think?

monai/transforms/post/array.py
wyli: thanks, looks good to me,  running more tests

john-zielke-snkeos:
My only concern is that the utility function here works a little bit different to the one that I have since created. In my worktree my function looks like this:
``` python
def distance_transform_edt_on_cucim_if_possible(
    image: NdarrayOrTensor,
    sampling: float | Sequence[float] | None = None,
    return_distances: bool = True,
    return_indices: bool = False,
    *,
    block_params=None,
    float64_distances=False,
    allow_cucim: bool = True,
) -> NdarrayOrTensor | tuple[NdarrayOrTensor, NdarrayOrTensor]:
    use_cucim = (
        allow_cucim
        and has_cucim_distance_transform_edt
        and isinstance(image, torch.Tensor)
        and image.device != torch.device("cpu")
    )
    # use_cucim = False
    if use_cucim:
        if image.dim() > 3:
            raise NotImplementedError("only images with 3 (HW[D]) or fewer dimensions are supported")

        return_value = cucim_distance_transform_edt(
            convert_to_cupy(image),
            sampling=sampling,
            return_distances=return_distances,
            return_indices=return_indices,
            block_params=block_params,
            float64_distances=float64_distances,
        )
        return convert_to_tensor(return_value)
    else:
        return_value = distance_transform_edt(
            convert_to_numpy(image), sampling=sampling, return_distances=return_distances, return_indices=return_indices
        )
        if isinstance(image, torch.Tensor):
            return convert_to_tensor(return_value)
        else:
            return return_value
```
Note: During this implementation I also noticed that convert_to_cupy() does not work for boolean cuda tensors.

Comments on the utility function:
1. There is a pattern in some places (e.g. https://github.com/Project-MONAI/MONAI/blob/e5f933781b1a5f4eb7d278e06389d531bff38dca/monai/transforms/utils.py#L1031C22-L1031C22 ) where whether to use scipy or cucim is decided by the location of the cuda tensor. Should we keep this functioniality? I am not sure here, but with this pattern the user can decide where to execute a transform based on what input tensor they pass. Otherwise we might need to expose `force_scipy` in every metric etc that uses this.
2. I am not sure how we should handle dims > 3. I did not test this, but I am assuming that the scipy functionality does not treat the dimensions > 3 different to the other ones, so would probably jump across them for the shortest distance. Whereas I am assuming that we want to have a maximum of 3 spatial dimensions and the other ones would always be batch or channel. Therefore if we implement the support for more than 3 dims, it should probable use the loop and stack approach anyway. But while it might be a common pattern to have dims > 3, it might lead to confusion if some dimensions are implicitly handled differently (for example how do I differentiate between CHW and HWD and not create very hard to spot bugs?)
3. I think it might be nice to add support for `return_indices` and `return_distances` as well

monai/transforms/utils.py
```diff
@@ -2051,5 +2058,46 @@ def has_status_keys(data: torch.Tensor, status_key: Any, default_message: str =
...
+            if a single number, this is used for all axes. If not specified, a grid spacing of unity is implied.
+        force_scipy: Force the CPU based scipy implementation of the euclidean distance transform
+    """
+    distance_transform_edt, has_cucim = optional_import(
+        "cucim.core.operations.morphology", name="distance_transform_edt"
+    )
+
+    if has_cp and has_cucim and not force_scipy:
+        img_ = convert_to_cupy(img)
+        if img_.ndim == 4:
```
john-zielke-snkeos, line 2086: For a 4D Tensor, this will lead to different results depending on whether you use force_scipy is True or False. Or even more important depending on whether you have CUDA available or not. I think we either need to use this logic for both cases or not support any input with dims > 3 here at all (see https://github.com/Project-MONAI/MONAI/pull/6981#issuecomment-1725969539) .

john-zielke-snkeos (changes requested):

tests/test_distance_transform_edt.py
```diff
@@ -0,0 +1,141 @@
...
+
+from monai.transforms import DistanceTransformEDT, DistanceTransformEDTd
+from tests.utils import HAS_CUPY, assert_allclose, optional_import, skip_if_no_cuda
+
+momorphology, has_cucim = optional_import("cucim.core.operations.morphology")
+ndimage, has_ndimage = optional_import("scipy.ndimage")
+cp, _ = optional_import("cupy")
+
+TEST_CASES = [
+    [
```
john-zielke-snkeos, line 27: Maybe we should also have some test cases with C and B > 1. Just repeating the axis might be fine for this.

john-zielke-snkeos (changes requested):

john-zielke-snkeos:
But since I did not mention that in the previous comment: Yes I agree it makes sense to use the utility function this PR creates in my PR as well. We should also consider which utils.py file this should go in, since it will be used from transforms/*, losses/* and probably metrics as well.

monai/transforms/utils.py
matt3o: Hi @john-zielke-snkeos,

first of all thanks for the comments :)

Yeah that is definitely a concern.
However by default the choice by default (without the force_scipy) only dependent on the import and will default to cupy which will simply crash if the input does not have the correct format.
We could at least make it more clear by enhancing the comment

`Note that the calculations of CuPy and Scipy can deviate, for details look into the cuCIM about distance_transform_edt(). Also for 4D input the CuPy version will calculate the distance transform channel-wise which returns different results compared to scipy.  If you need a full 4D distance transform, use scipy instead. We recommend to stick with one of the two libraries for the entire program.`

Do you think that would be enough? I don't see many other solutions right now apart from adding an actual working 4D implementation.

matt3o: Also to be on the save side, I did a trial with a 4D tensor, the results actually are different.

```python
(monai_dev) matteo@matteo-ubuntu-server:~/code/MONAI_matteo$ python
Python 3.10.11 (main, May 16 2023, 00:28:57) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from monai.transforms.utils import distance_transform_edt
>>> import numpy as np
>>> t = np.array([[[[1,1,1], [1,1,0], [0,0,0]]], [[[1,1,1], [1,1,0], [1,1,0]]],])
>>> t.ndim
4
>>> distance_transform_edt(t)
array([[[[2.       , 1.4142135, 1.       ],
         [1.       , 1.       , 0.       ],
         [0.       , 0.       , 0.       ]]],


       [[[2.236068 , 1.4142135, 1.       ],
         [2.       , 1.       , 0.       ],
         [2.       , 1.       , 0.       ]]]], dtype=float32)
>>> distance_transform_edt(t, force_scipy=True)
array([[[[2.        , 1.41421356, 1.        ],
         [1.        , 1.        , 0.        ],
         [0.        , 0.        , 0.        ]]],


       [[[2.23606798, 1.41421356, 1.        ],
         [1.41421356, 1.        , 0.        ],
         [1.        , 1.        , 0.        ]]]])

```

matt3o:
> My only concern is that the utility function here works a little bit different to the one that I have since created. In my worktree my function looks like this:
> 
> ```python
> def distance_transform_edt_on_cucim_if_possible(
>     image: NdarrayOrTensor,
>     sampling: float | Sequence[float] | None = None,
>     return_distances: bool = True,
>     return_indices: bool = False,
>     *,
>     block_params=None,
>     float64_distances=False,
>     allow_cucim: bool = True,
> ) -> NdarrayOrTensor | tuple[NdarrayOrTensor, NdarrayOrTensor]:
>     use_cucim = (
>         allow_cucim
>         and has_cucim_distance_transform_edt
>         and isinstance(image, torch.Tensor)
>         and image.device != torch.device("cpu")
>     )
>     # use_cucim = False
>     if use_cucim:
>         if image.dim() > 3:
>             raise NotImplementedError("only images with 3 (HW[D]) or fewer dimensions are supported")
> 
>         return_value = cucim_distance_transform_edt(
>             convert_to_cupy(image),
>             sampling=sampling,
>             return_distances=return_distances,
>             return_indices=return_indices,
>             block_params=block_params,
>             float64_distances=float64_distances,
>         )
>         return convert_to_tensor(return_value)
>     else:
>         return_value = distance_transform_edt(
>             convert_to_numpy(image), sampling=sampling, return_distances=return_distances, return_indices=return_indices
>         )
>         if isinstance(image, torch.Tensor):
>             return convert_to_tensor(return_value)
>         else:
>             return return_value
> ```
> 
> Note: During this implementation I also noticed that convert_to_cupy() does not work for boolean cuda tensors.
> 
> Comments on the utility function:
> 
> 1. There is a pattern in some places (e.g. https://github.com/Project-MONAI/MONAI/blob/e5f933781b1a5f4eb7d278e06389d531bff38dca/monai/transforms/utils.py#L1031C22-L1031C22 ) where whether to use scipy or cucim is decided by the device of the cuda tensor. Should we keep this functioniality? I am not sure here, but with this pattern the user can decide where to execute a transform based on what input tensor they pass. Otherwise we might need to expose `force_scipy` in every metric etc that uses this.
> 2. I am not sure how we should handle dims > 3. I did not test this, but I am assuming that the scipy functionality does not treat the dimensions > 3 different to the other ones, so would probably jump across them for the shortest distance. Whereas I am assuming that we want to have a maximum of 3 spatial dimensions and the other ones would always be batch or channel. Therefore if we implement the support for more than 3 dims, it should probable use the loop and stack approach anyway. But while it might be a common pattern to have dims > 3, it might lead to confusion if some dimensions are implicitly handled differently (for example how do I differentiate between CHW and HWD and not create very hard to spot bugs?)
> 3. I think it might be nice to add support for `return_indices` and `return_distances` as well

Hey,

1) I think why the `force_scipy` here makes sense is because CuPy does not offer the same functionality, so there are use cases where you actually have to force scipy. For other functions I would except it covers all the use cases, then you can just use the GPU functionality where available. Here it enables people who want to run a distance transforms on a 5D tensor for some, to do with the scipy implementation.
Imo that does not justify having to add it everywhere.
2) Yeah as I wrote above this is definitely an issue. I think @wyli has to answer this as one of maintainers, I have no clue what the default here should be.
3) That is no problem. Do we only need that for the `utils.py` function or shall I add that to the `DistanceTransformEDT` and `DistanceTransformEDTd`  as well? 
Also do you need the cucim flags? I can add those, however I'm not if it is desired to expose those in a simplified MONAI version of distance_transform_edt. 
@wyli can you add some thoughts here?

Thanks guys!

Ps: Kind of funny you had the same idea. I wanted to touch those places next, see the comments above. The slow SurfaceDice was also annoying for me. My use case comes from the DeepEdit guidance stuff however.

tests/test_distance_transform_edt.py
matt3o: You mean for the shape of the input? I can add those, that's not much work now. I would just add (3,5,5) then

monai/transforms/utils.py
KumoLiu, line 2086: Thanks for the PR, I would suggest all the input should be channel-first since most of the monai transform follows.
https://github.com/Project-MONAI/MONAI/blob/e5f933781b1a5f4eb7d278e06389d531bff38dca/monai/transforms/utils.py#L969
Thanks!

matt3o: So if I understand it correctly, you propose to always assume the first channel is the dimension. That means we `assert img_.ndim > 1` and then always apply transform channel-wise, even for scipy?
Thanks!

```
out = []
            for channel in img_:
                out.append(distance_transform_edt(channel, sampling=sampling))
            distance = cp.stack(out)
```

tests/test_distance_transform_edt.py
john-zielke-snkeos: Yes with batch and channel dimensions > 1

monai/transforms/utils.py
john-zielke-snkeos: How about we use the spatial_dims=2/3 as a first mandatory parameter like we do here: https://docs.monai.io/en/stable/transforms.html#probnms Right now we can still change that interface the way we want. This way we can have an arbitrary number of extra dimensions in front

KumoLiu: Sorry, my bad. I checked the code again. I think we can just keep spatial dim here.
https://github.com/Project-MONAI/MONAI/blob/e5f933781b1a5f4eb7d278e06389d531bff38dca/monai/metrics/surface_distance.py#L181
https://github.com/Project-MONAI/MONAI/blob/e5f933781b1a5f4eb7d278e06389d531bff38dca/monai/metrics/hausdorff_distance.py#L188

monai/transforms/post/array.py
```diff
@@ -936,3 +938,33 @@ def __call__(self, image: NdarrayOrTensor) -> torch.Tensor:
...
+    def __init__(self, sampling: None | float | list[float] = None, force_scipy: bool = False) -> None:
+        super().__init__()
+        self.force_scipy = force_scipy
+        self.sampling = sampling
+
+    def __call__(self, img: NdarrayOrTensor) -> NdarrayOrTensor:
+        """
+        Args:
+            img: Input image on which the distance transform shall be run.
+                For CuPy shape must be ([channel, ]spatial_dim1, spatial_dim2 [, spatial_dim3]), so 2D, 3D and 4D input only.
```
KumoLiu, line 963: We should assume the img here has a channel-first shape just like most MONAI transforms do.

monai/transforms/utils.py
```diff
@@ -2051,5 +2058,46 @@ def has_status_keys(data: torch.Tensor, status_key: Any, default_message: str =
...
+            4D input gets passed channel-wise to cupy.
+        sampling: Spacing of elements along each dimension. If a sequence, must be of length equal to the input rank;
+            if a single number, this is used for all axes. If not specified, a grid spacing of unity is implied.
+        force_scipy: Force the CPU based scipy implementation of the euclidean distance transform
+    """
+    distance_transform_edt, has_cucim = optional_import(
+        "cucim.core.operations.morphology", name="distance_transform_edt"
+    )
+
+    if has_cp and has_cucim and not force_scipy:
```
john-zielke-snkeos, line 2084: When we implement non-spatial dimensions and we care about performance here, I think we should use the option to pass in a preallocated torch tensor/numpyarray with the final dimension using the [distances and indices parameter](https://docs.rapids.ai/api/cucim/nightly/api/#cucim.core.operations.morphology.distance_transform_edt) (applies to [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.distance_transform_edt.html#scipy-ndimage-distance-transform-edt) as well). We then also need to make sure to match the proper datatypes. Keep in mind that we should support the float64_distances parameter for cucim here as well, since the float32 should be faster on GPU and the precision is probably not necessary in most cases. I am not 100% sure, but I think in case of cucim we should allocate a torch tensor and then use convert_to_cupy() to create a temporary view on this tensor cupy to pass to the transform. This way the result should get allocated using the pytorch memory pool

tests/test_distance_transform_edt.py
john-zielke-snkeos: We should also test different datatypes.

monai/transforms/utils.py
```diff
@@ -2051,5 +2058,46 @@ def has_status_keys(data: torch.Tensor, status_key: Any, default_message: str =
     return True, None
 
 
+def distance_transform_edt(
+    img: NdarrayOrTensor, sampling: None | float | list[float] = None, force_scipy: bool = False
+) -> NdarrayOrTensor:
+    """
+    Euclidean distance transform, either GPU based with CuPy / cuCIM
+    or CPU based with scipy.ndimage.
+    Choice depends on cuCIM being available or scipy can be forced with the ``force_scipy`` flag.
```
john-zielke-snkeos, line 2067: @wyli @KumoLiu What is your opinion on how to choose between cpu and gpu? Should it just depend on Cucim being available or on whether the tensor is on GPU already?

KumoLiu: I didn't get the point here why do we need a `force_scipy` flag? Seems we can directly check the dependency and the device here.
https://github.com/Project-MONAI/MONAI/blob/e5f933781b1a5f4eb7d278e06389d531bff38dca/monai/transforms/utils.py#L1031C22-L1031C22

john-zielke-snkeos: I agree, since the callee that would have to pass the flag has control over the input tensor as well so they could just move it to CPU if scipy is desired.

monai/transforms/utils.py
john-zielke-snkeos: But since it's such a common pattern, should we add support for non_spatial dims as well? Then not every user of this function needs to implement this pattern.

wyli:
thanks for the discussions here, just to clarify the transform shape convention that we've been using:
the first dimension is assumed the channel dimension, so
- single channel spatial 2D shape is `1` x `spatial_dim_1` x `spatial_dim_2`
- single channel spatial 3D shape is `1` x `spatial_dim_1` x `spatial_dim_2` x `spatial_dim_3`

shape such as `A`x`B`x`C` should be treated as `A`-channel spatially `B`x`C`, and when computing the distance transform in this case, each `B`x`C` frame should be computed independently. if the user wants to compute the transform for spatially 3D `A`x`B`x`C`, the input should be reshaped to `1`x`A`x`B`x`C`.

john-zielke-snkeos:
> thanks for the discussions here, just to clarify the transform shape convention that we've been using: the first dimension is assumed the channel dimension, so
> 
> * single channel spatial 2D shape is `1` x `spatial_dim_1` x `spatial_dim_2`
> * single channel spatial 3D shape is `1` x `spatial_dim_1` x `spatial_dim_2` x `spatial_dim_3`
> 
> shape such as `A`x`B`x`C` should be treated as `A`-channel spatially `B`x`C`, and when computing the distance transform in this case, each `B`x`C` frame should be computed independently. if the user wants to compute the transform for spatially 3D `A`x`B`x`C`, the input should be reshaped to `1`x`A`x`B`x`C`.

But isn't that quite suspectible for a hard to spot error if the shape you pass like this is implicit? Since it would accept both the `N` x `spatial_dim_1` x `spatial_dim_2` case and the `spatial_dim_1` x `spatial_dim_2` x `spatial_dim_3` case, and only differ in the values in some locations? I think passing an explicit parameter here would prevent this.

wyli:
> > thanks for the discussions here, just to clarify the transform shape convention that we've been using: the first dimension is assumed the channel dimension, so
> > 
> > * single channel spatial 2D shape is `1` x `spatial_dim_1` x `spatial_dim_2`
> > * single channel spatial 3D shape is `1` x `spatial_dim_1` x `spatial_dim_2` x `spatial_dim_3`
> > 
> > shape such as `A`x`B`x`C` should be treated as `A`-channel spatially `B`x`C`, and when computing the distance transform in this case, each `B`x`C` frame should be computed independently. if the user wants to compute the transform for spatially 3D `A`x`B`x`C`, the input should be reshaped to `1`x`A`x`B`x`C`.
> 
> But isn't that quite suspectible for a hard to spot error if the number of non-spatial dims you pass like this is implicit? Since it would accept both the `N` x `spatial_dim_1` x `spatial_dim_2` case and the `spatial_dim_1` x `spatial_dim_2` x `spatial_dim_3` case, and only differ in the values in some locations? I think passing an explicit parameter here would prevent this. Especially since this is not a transform, where this expectation is kind of well-known, but a utility function, that would then have a different behavior to the scipy implementation

the shape assumption has been used for a while now, it seems to simplify the user input validations in many cases and I don't see any issues. but we are open to discuss/revise the design, perhaps drafting a new proposal is a good starting point (please let me know if you need any help on this topic). But for this PR for now I think we can just follow the current assumption.

matt3o:
So to summarize the discussion and what I understood so far:

For `distance_transform_edt()` in `utils.py`: The input shape is always channel first. The follow up question here would be, do we then always want to extract the channels and run the transform on a loop over the channels or do we say scipy and cupy work differently always? Otherwise at least for CHWD, CHW, CH we would have the same results from cupy and scipy, however the results would be for channel times H(W(D)) accordingly.

Also I don't really follow on the selection of the algorithm, so scipy vs cupy. This is completely down to what MONAI usually does and I don't know if there is any convention. If the convention is that cupy get chosen if the device is GPU, I am completely fine with that. I am however not sure how well this approach generalizes. The example code cited before [get_largest_connected_component_mask](https://github.com/Project-MONAI/MONAI/blob/e5f933781b1a5f4eb7d278e06389d531bff38dca/monai/transforms/utils.py#L1031C22-L1031C22) actually relies on the tensor type being a torch Tensor. If for any reason you pass in a cupy tensor, it will choose the cpu implementation instead.. So it is kind of true it accepts NdarrayTensor, but for GPU support you actually need to pass a torch tensor, which is not stated in the docs there.
I don't find that a much better solution.
We could add code checking for the device on torch and cupy and if that evaluates to True, run on cupy, that would alleviate the problem I guess.

For the transforms `DistanceTransformEDT` `DistanceTransformEDTd` I am not sure. Is the convention there also input in CHWD or is it always BCHWD? Then we'd need to add some code for the batch wise handling of the input.

KumoLiu:
> For the transforms DistanceTransformEDT DistanceTransformEDTd I am not sure. Is the convention there also input in CHWD or is it always BCHWD? Then we'd need to add some code for the batch wise handling of the input.

Yes, for transforms `DistanceTransformEDT` and `DistanceTransformEDTd`, the input should also follow the convention of being channel-first (CHWD).

matt3o:
However that still did not answer my questions.
1) Do we leave the selection of the cupy versus scipy as it is or shall I change it to the device of the input? Shall I check for torch and cupy device then?
2) I will update the `distance_transform_edt()` to always assume channel-first data, so that cupy and scipy return the same results for 4D data, correct? This will however mean that `distance_transform_edt()` returns different data for 3D data compared to the cuCIM implementation - which I will mention in the docs. For the same behaviour people have to a add a single empty channel / dimension.
3) Since the transforms are also channel-first, there is nothing to change here, correct?

@wyli @KumoLiu

KumoLiu:
Hi @matt3o,

> 1. Do we leave the selection of the cupy versus scipy as it is or shall I change it to the device of the input? Shall I check for torch and cupy device then?

I would suggest doing something like this one instead using `force_scipy`.
https://github.com/Project-MONAI/MONAI/blob/e5f933781b1a5f4eb7d278e06389d531bff38dca/monai/transforms/utils.py#L1031C22-L1031C22

> 2. I will update the `distance_transform_edt()` to always assume channel-first data, so that cupy and scipy return the same results for 4D data, correct? This will however mean that `distance_transform_edt()` returns different data for 3D data compared to the cuCIM implementation - which I will mention in the docs. For the same behaviour people have to a add a single empty channel / dimension.

I think we can only support 2d(CHW) and 3d(CHWD) for now. The result can be concatenated in batch dim if needed. The result for cupy and scipy must be the same.

> 3. Since the transforms are also channel-first, there is nothing to change here, correct?
Yes.

cc @wyli in case you have any other comments.

monai/transforms/post/array.py
KumoLiu, line 963:
```suggestion
                channel first array, must have shape: (num_channels, H[, W, ..., ])
```

monai/transforms/utils.py
```diff
@@ -2012,7 +2019,7 @@ def has_status_keys(data: torch.Tensor, status_key: Any, default_message: str =
 
     Status keys are defined in :class:`TraceStatusKeys<monai.utils.enums.TraceStatusKeys>`.
 
-    This function also accepts:
+    This fun    ction also accepts:
```
KumoLiu, line 2022:
```suggestion
    This function also accepts:
```

monai/transforms/utils.py
```diff
@@ -2051,5 +2058,46 @@ def has_status_keys(data: torch.Tensor, status_key: Any, default_message: str =
...
+    """
+    Euclidean distance transform, either GPU based with CuPy / cuCIM
+    or CPU based with scipy.ndimage.
+    Choice depends on cuCIM being available or scipy can be forced with the ``force_scipy`` flag.
+
+    Note that the runtime running on the CPU may be really depending on the inputs size.
+
+    Args:
+        img: Input image on which the distance transform shall be run.
+            For CuPy shape must be ([channel, ]spatial_dim1, spatial_dim2 [, spatial_dim3]), so 2D, 3D and 4D input only.
```
KumoLiu, line 2073:
```suggestion
            channel first array, must have shape: (num_channels, H[, W, ..., ])
```

matt3o:
Hopefully final questions:
Both scipy and cupy have the common parameters `return_distances=True, return_indices=False, distances=None, indices=None`, as @john-zielke-snkeos mentioned. Do we want to expose them too?
At least `return_distances=True`, `return_indices=False` might make sense for the `utils.py` function, `distances` and `indices` might speed up code - I have not used them so far. The last two I would probably leave away for now, they can be added later on if that is necessary. `return_distances=True`, `return_indices=False` however do change the API, so we should decide on that now.
The transforms I would leave as they are until that the indices are actually required.

Those two flags do complicate the code a lot and I'm not sure how to solve it without duplication then:

```python
    distance_transform_edt, has_cucim = optional_import(
        "cucim.core.operations.morphology", name="distance_transform_edt"
    )
    use_cp = has_cp and has_cucim and isinstance(img, torch.Tensor) and img.device != torch.device("cpu")

    if not return_distances and not return_indices:
        raise RuntimeError("Neither return_distances nor return_indices True")

    distances, indices = [], []
    if use_cp:
        img_ = convert_to_cupy(img)
        for channel in img_:
            outputs = distance_transform_edt(channel, sampling=sampling, return_distances=return_distances, return_indices=return_indices)

            if return_distances and return_indices:
                distances.append(outputs[0])
                indices.append(outputs[1])
            elif return_distances:
                distances.append(outputs)
            elif return_indices:
                indices.append(outputs)
    else:
        if not has_ndimage:
            raise RuntimeError("scipy.ndimage required if cupy is not available")
        img_ = convert_to_numpy(img)
        for channel in img_:
            outputs =  ndimage.distance_transform_edt(channel, sampling=sampling, return_distances=return_distances, return_indices=return_indices)

            if return_distances and return_indices:
                distances.append(outputs[0])
                indices.append(outputs[1])
            elif return_distances:
                distances.append(outputs)
            elif return_indices:
                indices.append(outputs)

    d_out, i_out = None, None
    if use_cp:
        if return_distances and return_indices:
            d_out = cp.stack(distances)
        if return_indices:
            i_out = cp.stack(indices)
    else:
        if return_distances and return_indices:
            d_out = np.stack(distances)
        if return_indices:
            i_out = np.stack(indices)

    return d_out, i_out
```

KumoLiu:
Hi @matt3o, both are ok for me.
You can simplify the code like this:
https://github.com/Project-MONAI/MONAI/blob/3bea5cf6c4005c686d69a2e3ec6502c3d28bafba/monai/transforms/utils_pytorch_numpy_unification.py#L238
Thanks!

john-zielke-snkeos:
How about sth like this:
```py
def distance_transform_edt(
    img: NdarrayOrTensor,
    sampling: None | float | list[float] = None,
    return_distances: bool = True,
    return_indices: bool = False,
    distances=None,
    indices=None,
    *,
    block_params=None,
    float64_distances=False,
) -> NdarrayOrTensor:
    """
    Euclidean distance transform, either GPU based with CuPy / cuCIM
    or CPU based with scipy.ndimage.
    Choice depends on cuCIM being available or scipy can be forced with the ``force_scipy`` flag.

    Note that the runtime running on the CPU may be really depending on the inputs size.

    Args:
        ...
    """
    distance_transform_edt, has_cucim = optional_import(
        "cucim.core.operations.morphology", name="distance_transform_edt"
    )
    use_cp = has_cp and has_cucim and isinstance(img, torch.Tensor) and img.device != torch.device("cpu")

    if not return_distances and not return_indices:
        raise RuntimeError("Neither return_distances nor return_indices True")
    distances_original, indices_original = distances, indices
    distances, indices = None, None
    if use_cp:
        distances_, indices_ = None, None
        if return_distances:
            dtype = torch.float64 if float64_distances else torch.float32
            if distances is None:
                distances = torch.zeros_like(img, dtype=dtype)
            else:
                if not isinstance(distances, torch.Tensor) and distances.device != img.device:
                    raise TypeError("distances must be a torch.Tensor on the same device as img")
                if not distances.dtype == dtype:
                    raise TypeError("distances must be a torch.Tensor of dtype float32 or float64")
            distances_ = convert_to_cupy(distances)
        if return_indices:
            dtype = torch.int32
            if indices is None:
                indices = torch.zeros((img.dim(),) + img.shape, dtype=dtype)
            else:
                if not isinstance(indices, torch.Tensor) and indices.device != img.device:
                    raise TypeError("indices must be a torch.Tensor on the same device as img")
                if not indices.dtype == dtype:
                    raise TypeError("indices must be a torch.Tensor of dtype int32")
            indices_ = convert_to_cupy(indices)
        img_ = convert_to_cupy(img)
        for channel_idx in range(img_.shape[0]):
            distance_transform_edt(
                img_[channel_idx],
                sampling=sampling,
                return_distances=return_distances,
                return_indices=return_indices,
                distances=distances_[channel_idx] if distances_ is not None else None,
                indices=indices_[channel_idx] if indices_ is not None else None,
                block_params=block_params,
                float64_distances=float64_distances,
            )
    else:
        if not has_ndimage:
            raise RuntimeError("scipy.ndimage required if cupy is not available")
        img_ = convert_to_numpy(img)
        if return_distances:
            if distances is None:
                distances = np.zeros_like(img_, dtype=np.float64)
            else:
                if not isinstance(distances, np.ndarray):
                    raise TypeError("distances must be a numpy.ndarray")
                if not distances.dtype == np.float64:
                    raise TypeError("distances must be a numpy.ndarray of dtype float64")
        if return_indices:
            if indices is None:
                indices = np.zeros((img_.ndim,) + img_.shape, dtype=np.int32)
            else:
                if not isinstance(indices, np.ndarray):
                    raise TypeError("indices must be a numpy.ndarray")
                if not indices.dtype == np.int32:
                    raise TypeError("indices must be a numpy.ndarray of dtype int32")

        for channel_idx in range(img_.shape[0]):
            ndimage.distance_transform_edt(
                img_[channel_idx],
                sampling=sampling,
                return_distances=return_distances,
                return_indices=return_indices,
                distances=distances[channel_idx] if distances is not None else None,
                indices=indices[channel_idx] if indices is not None else None,
            )

    r_vals = []
    if return_distances and distances_original is None:
        r_vals.append(distances)
    if return_indices and indices_original is None:
        r_vals.append(indices)
    if not r_vals:
        return None
    if len(r_vals) == 1:
        return r_vals[0]
    return tuple(r_vals)

```

matt3o:
Looking good @john-zielke-snkeos, thanks! My only concern is: Does MONAI  want to expose the parameters of just cupy? Those flags don't do anything for scipy. Is that important to your use case to have `block_params` and `float64_distances`?

john-zielke-snkeos:
I think that's fine. We should mention that these are cucim specific just like they do in the cucim docs. And since they are simply ignored in the other case, I think that is fine. With the proper docs and them being keyword-only arguments, normal users will not really stumble upon these and users that need the options (especially the precision one) still have the possibility to use them.

john-zielke-snkeos:
@matt3o @wyli What I just wondered is if it's not better to check for `use_cp = has_cp and has_cucim and isinstance(img, torch.Tensor) and img.device.type == "cuda"` instead of `use_cp = has_cp and has_cucim and isinstance(img, torch.Tensor) and img.device != torch.device("cpu")`. Otherwise amd GPUs and other accelerators like the AWS one would fail in this case. And if you agree, should we also change this in https://github.com/Project-MONAI/MONAI/blob/e5f933781b1a5f4eb7d278e06389d531bff38dca/monai/transforms/utils.py#L1031C22-L1031C22 ?

wyli:
> @matt3o @wyli What I just wondered is if it's not better to check for `use_cp = has_cp and has_cucim and isinstance(img, torch.Tensor) and img.device.type == "cuda"` instead of `use_cp = has_cp and has_cucim and isinstance(img, torch.Tensor) and img.device != torch.device("cpu")`. Otherwise amd GPUs and other accelerators like the AWS one would fail in this case. And if you agree, should we also change this in https://github.com/Project-MONAI/MONAI/blob/e5f933781b1a5f4eb7d278e06389d531bff38dca/monai/transforms/utils.py#L1031C22-L1031C22 ?

thanks, I agree, we should be more specific in the if condition.

matt3o:
Sure, I'll match that right now. I will have to change the check to 
`use_cp = has_cp and has_cucim and isinstance(img, torch.Tensor) and img.device.type == torch.device("cuda").type`
however, otherwise it won't work with "cuda:0" etc devices

monai/transforms/utils.py
```diff
@@ -2051,5 +2058,136 @@ def has_status_keys(data: torch.Tensor, status_key: Any, default_message: str =
...
+            It will have the same shape as image. For cuCIM: Will have dtype torch.float64 if float64_distances is True,
+            otherwise it will have dtype torch.float32. For scipy: Will have dtype np.float64.
+        indices: The calculated feature transform. It has an image-shaped array for each dimension of the image.
+            Returned only when `return_indices` is True and `indices` is not supplied. dtype np.float64.
+
+    """
+    distance_transform_edt, has_cucim = optional_import(
+        "cucim.core.operations.morphology", name="distance_transform_edt"
+    )
+    use_cp = has_cp and has_cucim and isinstance(img, torch.Tensor) and img.device.type == torch.device("cuda").type
```
KumoLiu, line 2109: ` img.device.type == "cuda"` should be ok.
```
import torch

a = torch.tensor([1,2,3]).to("cuda:1")
print(a.device.type) --> "cuda"
```

wyli (approved):
thank you, it looks good to me, please help address the type annotation issues https://github.com/Project-MONAI/MONAI/actions/runs/6316993047/job/17169932102?pr=6981 and the docstring issues https://github.com/Project-MONAI/MONAI/actions/runs/6316993047/job/17169930462?pr=6981 then I'll trigger more tests and merged the PR.

monai/transforms/utils.py
```diff
@@ -2051,5 +2058,136 @@ def has_status_keys(data: torch.Tensor, status_key: Any, default_message: str =
...
+    For details about the implementation, check out the
+    `SciPy<https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.distance_transform_edt.html/>`_
+    and `cuCIM<https://docs.rapids.ai/api/cucim/nightly/api/#cucim.core.operations.morphology.distance_transform_edt/>`_ docs.
+
+    Args:
+        img: Input image on which the distance transform shall be run.
+            channel first array, must have shape: (num_channels, H[, W, ..., ])
+            Input gets passed channel-wise to the distance-transform, thus results from this function may differ
+            from directly calling ``distance_transform_edt()`` in CuPy or scipy.
+        sampling: Spacing of elements along each dimension. If a sequence, must be of length equal to the input rank;
```
john-zielke-snkeos, line 2085: Should this not be "length equal to input rank -1"? Since the first is the channel dimension?

monai/transforms/utils.py
```diff
@@ -2051,5 +2058,136 @@ def has_status_keys(data: torch.Tensor, status_key: Any, default_message: str =
...
+    or CPU based with scipy.
+    To use the GPU implementation, make sure cuCIM is available and that the data is a `torch.tensor` on a GPU device.
+
+    For details about the implementation, check out the
+    `SciPy<https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.distance_transform_edt.html/>`_
+    and `cuCIM<https://docs.rapids.ai/api/cucim/nightly/api/#cucim.core.operations.morphology.distance_transform_edt/>`_ docs.
+
+    Args:
+        img: Input image on which the distance transform shall be run.
+            channel first array, must have shape: (num_channels, H[, W, ..., ])
```
john-zielke-snkeos, line 2082: And this should be (num_channels, H[, W [, D ]]) right? Since we only support up to 3D images.

monai/transforms/utils.py
```diff
@@ -2051,5 +2058,136 @@ def has_status_keys(data: torch.Tensor, status_key: Any, default_message: str =
...
+    """
+    distance_transform_edt, has_cucim = optional_import(
+        "cucim.core.operations.morphology", name="distance_transform_edt"
+    )
+    use_cp = has_cp and has_cucim and isinstance(img, torch.Tensor) and img.device.type == torch.device("cuda").type
+
+    if not return_distances and not return_indices:
+        raise RuntimeError("Neither return_distances nor return_indices True")
+    distances_original, indices_original = distances, indices
+    distances, indices = None, None
```
john-zielke-snkeos, line 2114: Maybe we should also check if ndims is in the expected range (so I guess between 2 and 4 dims for 1D to 3D)

monai/transforms/utils.py
```diff
@@ -2051,5 +2058,136 @@ def has_status_keys(data: torch.Tensor, status_key: Any, default_message: str =
...
+    float64_distances: bool = False,
+) -> tuple[NdarrayOrTensor, NdarrayOrTensor]:
+    """
+    Euclidean distance transform, either GPU based with CuPy / cuCIM
+    or CPU based with scipy.
+    To use the GPU implementation, make sure cuCIM is available and that the data is a `torch.tensor` on a GPU device.
+
+    For details about the implementation, check out the
+    `SciPy<https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.distance_transform_edt.html/>`_
+    and `cuCIM<https://docs.rapids.ai/api/cucim/nightly/api/#cucim.core.operations.morphology.distance_transform_edt/>`_ docs.
```
john-zielke-snkeos, line 2078: We should maybe add the info about the results between cucim and scipy potentially differing here as well.

monai/transforms/utils.py
```diff
@@ -2051,5 +2058,136 @@ def has_status_keys(data: torch.Tensor, status_key: Any, default_message: str =
...
 
 
+def distance_transform_edt(
+    img: NdarrayOrTensor,
+    sampling: None | float | list[float] = None,
+    return_distances: bool = True,
+    return_indices: bool = False,
+    distances: NdarrayOrTensor | None = None,
+    indices: NdarrayOrTensor | None = None,
+    block_params: tuple[int, int, int] | None = None,
```
john-zielke-snkeos, line 2068: I think we should make block_params and float64_distances keyword only arguments just like the cucim implementation does.

john-zielke-snkeos (changes requested):

monai/transforms/utils.py
matt3o: Ah sweet, I did not yet know that functionality. Readded :)

monai/transforms/utils.py
matt3o: I have added that now. As above, cupy does not support 1D data, so it has to be `if not (img.ndim >= 3 and img.ndim <= 4):`
As we discussed scipy and cupy shall behave in the same way so I guess it makes sense to raise an error there

wyli:
many thanks @matt3o for the PR, and @john-zielke-snkeos @KumoLiu for the detailed comments, all premerge tests passed, and I'm merging this PR to include it in the upcoming monai v1.3 , please feel free to create follow-ups..

john-zielke-snkeos:
That's great news, thank you very much for the support! I'll adjust my Pull Request ASAP, would be nice if we could include the faster Metrics in 1.3 as well.

PR #7011, opened 2023-09-19
Add SignalFillEmptyd tranform

matt3o:
As mentioned here https://github.com/Project-MONAI/MONAI/discussions/2637#discussioncomment-7038045 , this transform helps to fix NaN values after the input transforms.
It is basically only a wrapper around `SignalFillEmpty`. 

### Types of changes


wyli (approved):
looks good to me, please add a unit test following the existing one https://github.com/Project-MONAI/MONAI/blob/dev/tests/test_signal_fillempty.py

not sure if `replacement` should allow for different inputs for different items in the `keys`, I guess we can expand the logic later if needed.

wyli:
thanks @matt3o please include the test case fix in this PR..

PR #7014, opened 2023-09-20
Add LICENSE requirement in spec

### LINKED ISSUE #6964: init_bundle Should Put License File in `docs`
When `init_bundle` is used to create a new bundle directory it should put the temp license file in `docs` instead of the root.

KumoLiu:
Fixes #6964.

### Description
Add LICENSE requirement in specification.
https://github.com/Project-MONAI/model-zoo/blob/bdd183ebc113a70d50db4e4fb2ebd9a7c8c7532b/ci/verify_bundle.py#L28

### Types of changes


docs/source/mb_specification.rst
```diff
@@ -13,11 +13,12 @@ This specification defines the directory structure a bundle must have and the ne
 Directory Structure
 ===================
 
-A MONAI Bundle is defined primarily as a directory with a set of specifically named subdirectories containing the model and metadata files. The root directory should be named for the model, given as "ModelName" in this example, and should contain the following structure:
+A MONAI Bundle is defined primarily as a directory with a set of specifically named subdirectories containing the model, metadata files and license. The root directory should be named for the model, given as "ModelName" in this example, and should contain the following structure:
 
 ::
 
   ModelName
+  ┣━ LICENSE
```
ericspod, line 21: The optional file `license.txt` is already mentioned in the docs directory, I would leave it there and make it mandatory if you like (and rename it `LICENSE`). If we want a separate license file for the data that should be `DATA_LICENSE` or something, having `LICENSE` in one place and `license.txt` in another seems confusing.

KumoLiu: Hi @ericspod, the `LICENSE` is for the entire bundle, and it appears to be a must-have file for a bundle. 
https://github.com/Project-MONAI/model-zoo/blob/bdd183ebc113a70d50db4e4fb2ebd9a7c8c7532b/ci/verify_bundle.py#L28
The `*license.txt` file, which appears to be an optional license, for a specific component of the bundle, such as data.
Does it makes sense?

ericspod: Yes I see what's been done so the verification hasn't matched the bundle's specification for a while then. The specification had said `license.txt` was for the model, we definitely should have separate licenses for model and data. We've let it get out of sync. I would say then we should change the verification as well but that's going to require modifying existing stored bundles. I still feel having all files in subdirectories is better organisation.

KumoLiu: If there is no particular reason to place the license in the root directory, I can assist in later modifying the existing bundle.
Could you please help confirm here? cc @yiheng-wang-nv @wyli @Nic-Ma

wyli: Maybe we were thinking about the GitHub default license location for the bundles https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#determining-the-location-of-your-license

If we release individual bundle as GitHub repo, the current location is consistent with that default..

ericspod: It's weird that Github will find the `README.md` file if it's in a subdirectory but not `LICENSE`, see this test: https://github.com/ericspod/BundleTestRepo

wyli: Found some explanations about it https://github.com/licensee/licensee/issues/335 makes sense to me

ericspod: I understand the rationale somewhat but I still don't see justification for forcing people to do things their way. Make it an option like others have requested, look for the file in the root first then in docs. We can't change it though so we'll have to live with it, I don't know if symlinking to the file in docs would work so if it's critical Github display the license we have to leave it in the root.

wyli (approved):
approve according to https://github.com/licensee/licensee/issues/335, feel free to revisit this if there are further concerns.

PR #7032, opened 2023-09-22
Allow for defining reference grid on non-integer coordinates

brudfors:
This is a non-breaking change that is disabled by default. It allows for defining the reference identity grid on non-integer values. This can be beneficial for registration applications as shown in: B. Likar and F. Pernus. A heirarchical approach to elastic registration based on mutual information. Image and Vision Computing, 19:33-44, 2001.

cc: @nvahmadi


wyli (approved):
Thanks, it looks good to me.

PR #7068, opened 2023-09-30
Adding padding option to autoencoder

### LINKED ISSUE #7045: Add Padding as an Argument to AutoEncoder
In the current implementation of the AutoEncoder class in the `monai.networks.nets` module, the padding parameter is not customizable. This limits the model's flexibility, as different padding strategies might suit different data-sizes. 
https://github.com/Project-MONAI/MONAI/blob/18a671a7f12389dd8263a4c2232196dcca814bb9/monai/networks/nets/autoencoder.py#L90

I would like to propose adding a padding argument to the AutoEncoder class constructor. This would allow users to specify the padding strategy when instantiating the model. 

An alternative solution could be to create a subclass of AutoEncoder (like CustomAE) whenever a different padding strategy is needed. See below,
```{python}
class CustomAE(AutoEncoder):
    def __init__(self, padding, **kwargs):
        self.padding = padding
        super().__init__(**kwargs)
    
    def _get_encode_layer(self, in_channels: int, out_channels: int, strides: int, is_last: bool) -> nn.Module:
        # ...
        mod = Convolution(
            spatial_dims=self.dimensions,
            in_channels=in_channels,
            out_channels=out_channels,
            strides=strides,
            kernel_size=self.kernel_size,
            padding=self.padding,
            # ...
        )
        return mod

```

JupiLogy:
Fixes #7045  .

### Description

Added "padding" option to `monai/network/nets/autoencoder.py` such that the conv and residual units will be passed the padding option.

### Types of changes

### Notes
I haven't been able to run the `./runtests.sh`, even tried running `./runtests.sh -h` and got no output (no error or anything). I guess I don't have permissions to run `.sh` files on this machine. However the changes are very small and default to previous functionality so unless somebody passes a padding argument, this should not break existing usage of the function.

### DIFF
--- a/monai/networks/nets/autoencoder.py
+++ b/monai/networks/nets/autoencoder.py
@@ -58,6 +58,8 @@ class AutoEncoder(nn.Module):
         bias: whether to have a bias term in convolution blocks. Defaults to True.
             According to `Performance Tuning Guide <https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html>`_,
             if a conv layer is directly followed by a batch norm layer, bias should be False.
+        padding: controls the amount of implicit zero-paddings on both sides for padding number of points
+            for each dimension in convolution blocks. Defaults to None.
 
     Examples::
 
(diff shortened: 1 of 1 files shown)


wyli (approved):
Thanks, it looks good to me, merging once premerge tests pass

PR #7089, opened 2023-10-05
Signalfillempty set track_meta to True

matt3o:
Related to #7011. I was using the `Signalfillemptyd` (based on `Signalfillempty`) transform in monailabel and found out, it currently allows no inversion. 
When digging deeper I realized that SignalFillEmpty just throws away the meta information.
With the simple addition of `track_meta=True` it works as expected.
I hope it has no other impact, but I honestly don't know.

@wyli would be cool if we can add this to MONAI 1.3.0, thanks! I can also rework `Signalfillempty` to just accept any datatype, if that would the more appropriate approach.

### Types of changes


wyli (approved):
Thanks, it looks good to me.

PR #7095, opened 2023-10-06
7047 simplify resnet pretrained

### LINKED ISSUE #7047: Simplify existing `resnet` API for `pretrained` flag
The `pretrained` flag in the `resnet50` function currently only accepts a boolean value. If it's set to True, the function throws an error pointing the user to download MedicalNet weights. Then, the user has to download these and manually load the `state_dict` into the model. 

This process can be simplified largely by allowing the `pretrained` flag to take `str` values that point to a path and automatically loading `state_dict` from these paths.

```python
model = resnet50_monai(pretrained=False, n_input_channels=n_input_channels, widen_factor=widen_factor, conv1_t_stride=conv1_t_stride, feed_forward=feed_forward, bias_downsample=bias_downsample)
model = model.to(device)
if pretrained:
    if Path(pretrained).exists():
        logger.info(f"Loading weights from {weights_path}...")
        checkpoint = torch.load(pretrained, map_location=device)
    else:
         ### Throw error

     if "state_dict" in checkpoint:
        model_state_dict = checkpoint["state_dict"]
        model_state_dict = {key.replace("module.", ""): value for key, value in model_state_dict.items()}
        
    model.load_state_dict(model_state_dict, strict=True)
```

This can help with loading the MedicalNet weights more easily and potentially open up this API to loading other models that have been trained with monai as well. If we don't want to support duck typing, then we could also add another flag to the function. 

**Additional context**

I do understand that monai is moving in a different direction with loading bundles etc. but since this `pretrained` flag still exists, it might be useful to simplify it.

vgrau98:
Fixes #7047

### Description

Resnet did not support `True` value (not implemented ) for its pretrained flag.
2 implemented behavior: 
- When pretrained is True, download weights from https://huggingface.co/TencentMedicalNet
- When pretrained is a string, loads weights from the path defined by the string

### Types of changes


vgrau98:
Not sure about good practices in added test

vgrau98:
Documentation should be updated too

monai/networks/nets/resnet.py
```diff
@@ -14,6 +14,10 @@
 from collections.abc import Callable
 from functools import partial
 from typing import Any
+from pathlib import Path
+import logging
+import re
+from huggingface_hub import hf_hub_download
```
KumoLiu, line 20: May use `optional_import` here and also in the unit test.
https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#adding-new-optional-dependencies

monai/networks/nets/resnet.py
```diff
@@ -329,21 +339,40 @@ def _resnet(
...
+                filename=f"{MEDICALNET_HUGGINGFACE_FILES_BASENAME}{resnet_depth}.pth")
+
+            checkpoint = torch.load(pretrained_path, map_location=device)
+
+        if "state_dict" in checkpoint:
+            model_state_dict = checkpoint["state_dict"]
+            model_state_dict = {key.replace("module.", ""): value for key, value in model_state_dict.items()}
+        else:
+            ### Throw error
+            raise KeyError(
```
KumoLiu, line 370: May I know why must have this `state_dict` key here? What if users use their own pre-trained weights?

vgrau98: This was to be consistent with the way weights are saved in MedicalNet (checkpoint containing the state dict). But it's probably best to pass the state dict directly when users pass their own pretrained weights ? 

```suggestion
            model_state_dict = checkpoint
```

monai/networks/nets/resnet.py
vgrau98: Thanks for your feedback. I'll fix it.

monai/networks/nets/resnet.py
```diff
@@ -14,14 +14,23 @@
...
 
 from monai.networks.layers.factories import Conv, Norm, Pool
 from monai.networks.layers.utils import get_pool_layer
-from monai.utils import ensure_tuple_rep
+from monai.utils import ensure_tuple_rep, optional_import
 from monai.utils.module import look_up_option
+from monai.networks.utils import get_pretrained_resnet_medicalnet
+
+Path, _ = optional_import("pathlib", name = "Path")
+re, _ = optional_import("re")
```
KumoLiu, line 30: pathlib and re are both Python built-in libraries and do not require an optional import.

monai/networks/nets/resnet.py
KumoLiu: Yes, sure. You can demonstrate it in the docstring.

vgrau98:
`test_resnet`  gives errors, all the tests i wrote don't seem to be unitary, there's an integration part with the huggingface api. I'll try to fix this quickly by creating integration tests in the dedicated `test_integration_resnet.py` file. Maybe the test environment is not the same between unit and integration tests ?

wyli:
> `test_resnet` gives errors, all the tests i wrote don't seem to be unitary, there's an integration part with the huggingface api. I'll try to fix this quickly by creating integration tests in the dedicated `test_integration_resnet.py` file. Maybe the test environment is not the same between unit and integration tests ?

thanks, please follow this guide for adding new dep https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#adding-new-optional-dependencies

I think the current unit test looks great, we just need to somehow pin down a specific commit hash of the model version, then I don't see much risk of regression.

vgrau98:
> > `test_resnet` gives errors, all the tests i wrote don't seem to be unitary, there's an integration part with the huggingface api. I'll try to fix this quickly by creating integration tests in the dedicated `test_integration_resnet.py` file. Maybe the test environment is not the same between unit and integration tests ?
> 
> thanks, please follow this guide for adding new dep https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#adding-new-optional-dependencies
> 
> I think the current unit test looks great, we just need to somehow pin down a specific commit hash of the model version, then I don't see much risk of regression.

Yes sorry my bad, didn't read the doc carefully enough, should be fixed now

monai/networks/utils.py
```diff
@@ -1164,3 +1166,60 @@ def freeze_layers(model: nn.Module, freeze_vars=None, exclude_vars=None):
                 warnings.warn(f"The exclude_vars includes {param}, but requires_grad is False, change it to True.")
 
     logger.info(f"{len(frozen_keys)} of {len(src_dict)} variables frozen.")
+
+
+def get_pretrained_resnet_medicalnet(
```
KumoLiu, line 1171: If this function is only used for resnet, I would suggest moving it to `nets.resnet`.

monai/networks/utils.py
```diff
@@ -1164,3 +1166,60 @@ def freeze_layers(model: nn.Module, freeze_vars=None, exclude_vars=None):
...
+
+    Returns:
+        Pretrained state dict
+
+    Raises:
+        huggingface_hub.utils._errors.EntryNotFoundError: if pretrained weights are not found on huggingface hub
+        NotImplementedError: if `resnet_depth` is not supported
+    """
+
+    medicalnet_huggingface_repo_basename = "TencentMedicalNet/MedicalNet-Resnet"
```
KumoLiu, line 1191: I think we can directly use the one in the `nets.resnet` in order to keep it from the same source.

monai/networks/nets/resnet.py
```diff
@@ -329,21 +340,62 @@ def _resnet(
...
+                # Throw error
+                raise FileNotFoundError("The pretrained checkpoint file is not found")
+        else:
+            # Also check bias downsample and shortcut.
+            if kwargs.get("spatial_dims", 3) == 3:
+                if kwargs.get("n_input_channels", 3) == 1 and kwargs.get("feed_forward", True) is False:
+                    resnet_depth = int(re.search(r"resnet(\d+)", arch).group(1))
+                    # get shortcut_type and bias_downsample.
+
+                    def get_medicalnet_pretrained_resnet_args(resnet_depth: int):
```
KumoLiu, line 363: I would suggest moving it out from the `_resnet` as a separate function then the unittest can also import this one without duplicating another one.

KumoLiu (approved):
Looks good to me, several minor comments inline.
Could you please also fix the format ci error?
https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#checking-the-coding-style

wyli (approved):
thanks! trying to merge this pr for more integration tests

PR #7098, opened 2023-10-07
Add weight in `DiceLoss`

### LINKED ISSUE #7065: DiceLoss add weight per class parameter
Currently DiceFocalLoss(), has "focal_weight", which reweights class contributions (but only for the focal loss part). 
Similarly DiceCELoss(), has "ce_weight" parameter for the same thing of CE loss part. 

1) Let's standardize the naming of this argument to be the same for all losses, e.g. simply "weight"
2) Let's add this "weight" parameters to DiceLoss() class, if provided to re-weights contributions of individual per-component dices.  And modify DiceFocalLoss() and DiceCELoss() to propagate this "weight" to the Dice loss initialization.

This can be achieved by modifying DiceLoss() after this line
https://github.com/Project-MONAI/MONAI/blob/14fcf72a6b733d7b65888d67181fd66d8ebcf755/monai/losses/dice.py#L182

with an additional code snippet (or similar)

```
        if self.weight is not None:
            # make sure the lengths of weights are equal to the number of classes
            class_weight: Optional[torch.Tensor] = None
            num_of_classes = target.shape[1]
            if isinstance(self.weight, (float, int)):
                class_weight = torch.as_tensor([self.weight] * num_of_classes)
            else:

                if class_weight.shape[0] != num_of_classes:
                    raise ValueError(
                        """the length of the `weight` sequence should be the same as the number of classes.
                        If `include_background=False`, the weight should not include
                        the background category class 0."""
                    )
            if class_weight.min() < 0:
                raise ValueError("the value/values of the `weight` should be no less than 0.")

            f =  f * class_weight.to(f)


```

thank you

KumoLiu:
Fixes #7065.

### Description
- standardize the naming to be simply "weight".
- add this "weight" parameter to `DiceLoss`.
### Types of changes

### DIFF
--- a/monai/losses/dice.py
+++ b/monai/losses/dice.py
@@ -112,6 +112,7 @@ class DiceLoss(_Loss):
         self.smooth_dr = float(smooth_dr)
         self.batch = batch
         self.weight = weight
+        self.register_buffer('class_weight', None)
 
     def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
         """
@@ -190,23 +191,21 @@ class DiceLoss(_Loss):
 
         if self.weight is not None and target.shape[1] != 1:
             # make sure the lengths of weights are equal to the number of classes
-            class_weight: Optional[torch.Tensor] = None
-            self.register_buffer('class_weight', class_weight)
             num_of_classes = target.shape[1]
             if isinstance(self.weight, (float, int)):
-                class_weight = torch.as_tensor([self.weight] * num_of_classes)
+                self.class_weight = torch.as_tensor([self.weight] * num_of_classes)
             else:
-                class_weight = torch.as_tensor(self.weight)
-                if class_weight.shape[0] != num_of_classes:
+                self.class_weight = torch.as_tensor(self.weight)
+                if self.class_weight.shape[0] != num_of_classes:
                     raise ValueError(
                         """the length of the `weight` sequence should be the same as the number of classes.
                         If `include_background=False`, the weight should not include
                         the background category class 0."""
                     )
-            if class_weight.min() < 0:
+            if self.class_weight.min() < 0:
                 raise ValueError("the value/values of the `weight` should be no less than 0.")
             # apply class_weight to loss
-            f = f * class_weight.to(f)
+            f = f * self.class_weight.to(f)
 
         if self.reduction == LossReduction.MEAN.value:
             f = torch.mean(f)  # the batch and channel average
@@ -920,7 +919,7 @@ class DiceFocalLoss(_Loss):
         return total_loss
 
 
-class GeneralizedDiceFocalLoss(torch.nn.modules.loss._Loss):
+class GeneralizedDiceFocalLoss(_Loss):
     """Compute both Generalized Dice Loss and Focal Loss, and return their weighted average. The details of Generalized Dice Loss
     and Focal Loss are available at ``monai.losses.GeneralizedDiceLoss`` and ``monai.losses.FocalLoss``.
 
--- a/monai/losses/dice.py
+++ b/monai/losses/dice.py
@@ -188,7 +188,7 @@ class DiceLoss(_Loss):
 
         f: torch.Tensor = 1.0 - (2.0 * intersection + self.smooth_nr) / (denominator + self.smooth_dr)
 
-        if self.weight is not None:
+        if self.weight is not None and target.shape[1] != 1:
             # make sure the lengths of weights are equal to the number of classes
             class_weight: Optional[torch.Tensor] = None
             num_of_classes = target.shape[1]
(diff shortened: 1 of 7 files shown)


tests/test_masked_loss.py
```diff
@@ -54,14 +54,12 @@ def test_shape(self, input_param, expected_val):
         pred = torch.randn(size)
         result = MaskedLoss(**input_param)(pred, label, None)
         out = result.detach().cpu().numpy()
-        checked = np.allclose(out, expected_val[0][0]) or np.allclose(out, expected_val[0][1])
```
KumoLiu, line 57: I didn't get the point of why needed to check two values here so removed another one.

monai/losses/dice.py
```diff
@@ -181,6 +188,25 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
 
         f: torch.Tensor = 1.0 - (2.0 * intersection + self.smooth_nr) / (denominator + self.smooth_dr)
 
+        if self.weight is not None and target.shape[1] != 1:
+            # make sure the lengths of weights are equal to the number of classes
+            class_weight: Optional[torch.Tensor] = None
```
wyli, line 193: we may define the class_weight as a buffer https://github.com/pytorch/pytorch/blob/fa8e4ea2122f45d59b6a7e748ecd3c157e504024/torch/nn/modules/loss.py#L31C14-L32

KumoLiu: Updated it in the latest commit.

wyli: thanks, please define it in the constructor, or do you see any issues?

tests/test_masked_loss.py
wyli: There was no deterministic results with torch.argmax

wyli (approved):
thanks, it looks good to me, cc @myron feel free to create further request for any concerns

monai/losses/dice.py
```diff
@@ -105,6 +111,8 @@ def __init__(
         self.smooth_nr = float(smooth_nr)
         self.smooth_dr = float(smooth_dr)
         self.batch = batch
+        self.weight = weight
+        self.register_buffer("class_weight", torch.zeros(1))
```
wyli, line 115: the default is None in the pytorch impl,  does that work? otherwise I think torch.ones is better than torch.zeros as a default value.

myron:
@KumoLiu 

Thanks for adding. But why is "self.class_weight" a property and registered as buffer?  We recompute it every time on the fly from self.weight.  I think "class_weight" can be just an intermediate variable,  to avoid confusion. thanks

KumoLiu:
Hi @myron, I think the main reason for registering the class_weight as a buffer is that "class_weight" doesn't require training optimization.

myron:
@KumoLiu,  "self.class_weight" does not require optimization, also it is fully defined by "self.weight" and is recomputed every time in your implementation.  As far as I can tell, there is no reason to define "self.class_weight" to be a class property of DiceLoss() at all. We never update that property directly.

Unless there is a reason, it should not be a property of DiceLoss.  Please remove it as class property, you can simply have an intermediate local variable called  "class_weight" during loss calculation. And in general, if it a simple local variable, do not declare new class properties, it will be very confusing for users. CC @wyli @Nic-Ma 

https://github.com/KumoLiu/MONAI/blob/55416891b778afa94d495c538acc16e47a445f02/monai/losses/dice.py#L192-L208

wyli:
Thanks @myron, this is to be consistent with the weighted loss interface and with the benefit of saving the weights as part of training states: https://github.com/pytorch/pytorch/blob/798efab53274ff44d0b5bbd2de59299b529e757c/torch/nn/modules/loss.py#L28

If there's significant computional overheads we can look into refactoring.

myron:
There is no compute overhead, but it looks like a bad coding practice to me.  We have self.weight and self.class_weight now, which both represent the same thing, and furthermore self.class_weight is recomputed every time from self.weight in the forward() pass. 

1) If we are to be consistent with PyT weighted loss, then we already have "self.weight" property.  What is the benefit of declaring another one "self.class_weight".
2) do we really save self.weight as part of training stats somewhere? (we just introduced self.weight) I'm interested to know, where do we save it and what do we save: "self.weight" or "self.class_weight". 

Currently, it looks confusing, and a user may attempt to change dice_loss.class_weight=.. property directly, which will not accomplish anything. We should try to simplify our code, unless I'm missing the benefit of this extra property. thanks

wyli:
perhaps the two variables `self.weight` and `self.class_weight` could be merged into one registered buffer, or could we remove `self.weight`? cc @KumoLiu 

(also this discussion has more details about using the usage https://discuss.pytorch.org/t/what-is-the-difference-between-register-buffer-and-register-parameter-of-nn-module/32723)

KumoLiu:
Thanks @wyli and @myron for the suggestions. I have created another [PR](https://github.com/Project-MONAI/MONAI/pull/7158) to address it. I removed `self.weight` and now we only have one registered buffer. Please check whether it makes sense to you.

PR #7106, opened 2023-10-09
Code formatting for user specified directories

### LINKED ISSUE #7099: making `./runtests.sh --autofix` available for user specified directories other than `pwd`
`./runtests.sh --autofix --precommit` is helpful for ensuring a consistent coding style for monai and monai related projects (bundle scripts and research contributions), but it currently is hard-coded for this codebase `$(pwd)`. 

would be great to make some of the minimal code formating standard available for example via `./runtests.sh --format-fix --path /my/bundle/scripts`

KumoLiu:
Fixes #7099

### Types of changes

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -63,7 +63,7 @@ function print_usage {
     echo "runtests.sh [--codeformat] [--autofix] [--black] [--isort] [--flake8] [--pylint] [--ruff]"
     echo "            [--clangformat] [--precommit] [--pytype] [-j number] [--mypy]"
     echo "            [--unittests] [--disttests] [--coverage] [--quick] [--min] [--net] [--build] [--list_tests]"
-    echo "            [--dryrun] [--copyright] [--clean] [--help] [--version]"
+    echo "            [--dryrun] [--copyright] [--clean] [--help] [--version] [--path] [--formatfix]"
     echo ""
     echo "MONAI unit testing utilities."
     echo ""
@@ -107,6 +107,8 @@ function print_usage {
     echo "    -c, --clean       : clean temporary files from tests and exit"
     echo "    -h, --help        : show this help message and exit"
     echo "    -v, --version     : show MONAI and system version information and exit"
+    echo "    -p, --path     : specify the path used for formatting"
+    echo "    --formatfix     : format code using \"isort\" and \"black\" for user specified directories"
     echo ""
     echo "${separator}For bug reports and feature requests, please file an issue at:"
     echo "    https://github.com/Project-MONAI/MONAI/issues/new/choose"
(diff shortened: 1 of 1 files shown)


runtests.sh
```diff
@@ -582,9 +602,9 @@ then
 
     if [ $doRuffFix = true ]
     then
-        ruff check --fix "$(pwd)"
+        ruff check --fix "$homedir"
```
wyli, line 605: it seems the ruff config is ignored https://github.com/Project-MONAI/MONAI/blob/825021d46d6c1a97d6c38bcbfbb0e5e0e148995e/pyproject.toml#L39 if it's not current directory

runtests.sh
```diff
@@ -280,6 +283,14 @@ do
             doRuffFormat=true
             doCopyRight=true
         ;;
+        --formatfix)
+            doIsortFix=true
+            doBlackFix=true
+            doRuffFix=true
+            doIsortFormat=true
+            doBlackFormat=true
+            doRuffFormat=true
```
wyli, line 292: the `print_style_fail_msg` doesn't give a correct instruction for non default directory, I think we can ignore

wyli (commented):
thanks! I have some minor comments inline please help test different use cases

runtests.sh
KumoLiu: Yes, updated in this [commit](https://github.com/Project-MONAI/MONAI/pull/7106/commits/571b62bfa8c5c494b52369557bed5a0043cedc72). Thanks!

runtests.sh
KumoLiu: I copied "pyproject.toml" and "setup.cfg" to the specified dir. Does it look good?

wyli: those files contain specific info about this project codebase, I don't think it should overwrite the specified dir. if it's not easy to fix it, perhaps let's not support ruff for the external dir for now.

KumoLiu: But there are also other configs, such as flake8, mypy contained in setup.cfg.
https://github.com/Project-MONAI/MONAI/blob/825021d46d6c1a97d6c38bcbfbb0e5e0e148995e/setup.cfg#L163

wyli: this doesn't seem to related to flake8 or mypy in my understanding, the currently enabled autofix tools are
```
        --formatfix)
            doIsortFix=true
            doBlackFix=true
            doRuffFix=true
            doIsortFormat=true
            doBlackFormat=true
            doRuffFormat=true
```

KumoLiu: Yes, my original thought is that we can also do something like `./runtests.sh --flake8 --path /my/bundle/scripts`. So if we only support formatfix with path, that makes sense. I will remove the ruff format for now.

wyli (approved):

PR #7109, opened 2023-10-10
6387 update_kwargs for merging multiple configs

### LINKED ISSUE #5899: Merging/overriding configs with same format in monai bundle run
Currently, overriding config can be done through an `args_file` argument pointing to a json/yaml provided to the `run` script. However, it only merges configs at the base level, and to override nested structures, the `#` format must be used. 

Example,
````
monai.bundle.run("", **{
    "args_file": "./config2.json",
    "config_file": "./config1.json", 
    
})
````

````
```
config1.json
```
{
    "trainer":
    {
        "max_epochs": 5

    }
}
````

````
```
config2.json
```
{
    "trainer#devices": 1
}
````

This gives the output,
````
[{'_meta_': {}, 'trainer': {'max_epochs': 5, 'devices': 1}}]
````

However, if the config provided to  `args_file` is in the same format (shown below) as the `config_file`, it does not do the merge,
````
```
config2.json
```
{
    "trainer":
    {
        "devices": 1

    }
}
````
It would be beneficial to have the ability to merge these two configs when they are in the same format to allow for use cases where a default config is set to have good parameter definitions, and the user only needs to define/override specific fields.

I tried specifying two of these configs as a list in this form as well,
````
monai.bundle.run("", **{
    "config_file": ["./config1.json", "./config2.json"], 
    
})
````
But it doesn't seem to handle overriding keys, although it merges non-conflicting keys. 

**My temporary fix**
To achieve the functionality I needed, I've changed the following,
https://github.com/Project-MONAI/MONAI/blob/86c7ecdc1225529a379e2cd58c23fe53aa79b0d0/monai/bundle/config_parser.py#L223

to 
````{python}
    def update(self, pairs: dict[str, Any]):
        """
        Set the ``id`` and the corresponding config content in pairs, see also :py:meth:`__setitem__`.
        For example, ``parser.update({"train#epoch": 100, "train#lr": 0.02})``

        Args:
            pairs: dictionary of `id` and config pairs.

        """
        for k, v in pairs.items():
            if isinstance(v, dict):
                self.update({f'{k}#{_k}': _v for _k, _v in v.items()})
                return 
                
            self[k] = v
````

**Additional context**
Ideally, it would be nice to have the ability to merge configs in the `config_file` list. Is there a way to do this, and am potentially missing something?

### LINKED ISSUE #6387: APIs for combining file-based and dict-based configs

the current API is private

https://github.com/Project-MONAI/MONAI/blob/e18097d0a8ba323ee66aeb7830d296bf54f44bdd/monai/bundle/scripts.py#L68-L71

should be made available properly for reading multiple config files and dictionary-based overrides.

### PR DESCRIPTION
**Pull request by wyli**
Fixes #6387
Fixes #5899

### Description
- add api for update_kwargs
- add support of merging multiple configs files and dictionaries
- remove warning message of directory in `runtests.sh`

### Types of changes

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -103,7 +103,7 @@ def update_kwargs(args: str | dict | None = None, ignore_none: bool = True, **kw
     return args_
 
 
-_update_kwargs = update_kwargs
+_update_args = update_kwargs  # backward compatibility
 
 
 def _pop_args(src: dict, *args: Any, **kwargs: Any) -> tuple:
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -93,7 +93,6 @@ def update_kwargs(args: str | dict | None = None, ignore_none: bool = True, **kw
     for k, v in kwargs.items():
         if ignore_none and v is None:
             continue
-        print(f"{k}={v}")
         if isinstance(v, dict) and isinstance(args_.get(k), dict):
             args_[k] = update_kwargs(args_[k], ignore_none, **v)
         else:
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -87,6 +87,9 @@ def update_kwargs(args: str | dict | None = None, ignore_none: bool = True, **kw
     if isinstance(args, str):
         # args are defined in a structured file
         args_ = ConfigParser.load_config_file(args)
+    if isinstance(args, (tuple, list)) and all(isinstance(x, str) for x in args):
+        primary, overrides = args
+        args_ = update_kwargs(primary, ignore_none, **update_kwargs(overrides, ignore_none, **kwargs))
     if not isinstance(args_, dict):
         return args_
     # recursively update the default args with new args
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: monai/bundle/scripts.py [hunk 1]
```diff
@@ -66,28 +66,35 @@
 PPRINT_CONFIG_N = 5
 
 
-def _update_args(args: str | dict | None = None, ignore_none: bool = True, **kwargs: Any) -> dict:
+def update_kwargs(args: str | dict | None = None, ignore_none: bool = True, **kwargs: Any) -> dict:
```
KumoLiu, line 69: Do we need to support sequence str here and use `load_config_files` in L89?

ericspod (approved):
LGTM

monai/bundle/scripts.py
wyli: good point, that logic in in the config parser https://github.com/Project-MONAI/MONAI/blob/fc1350aae1de702404189f334ac17fd6f51321d0/monai/bundle/config_parser.py#L318-L320

but that reminds me there's a case of comma separated filenames https://github.com/Project-MONAI/research-contributions/blob/b2809b5adfa49b7e9220f1e2d1ff26a16f23dc1b/auto3dseg/algorithm_templates/swinunetr/scripts/train.py#L142-L143 I'll update the load config to fix it

wyli: had another look and add support for sequence str, which can close https://github.com/Project-MONAI/MONAI/issues/5899 at the CLI level.

PR #7116, opened 2023-10-11
Starting updates to docs

ericspod:
### Description

This updates the docs with a minor fix to the bundle metadata example and adds diagrams to help explain MONAI, and bundles vs. MAPs.

### Types of changes


docs/images/MONAI_arch.png
JHancox, line 1: This is useful. Any reason why MAPs are not included in the top diagram?

ericspod: That was meant to be integrated into the "MONAI Deploy" but I can make a separate one for MAPs specifically.

docs/source/bundle_intro.rst
vikashg, line 1: Model Serving is still not supported by MAPs.

docs/source/bundle_intro.rst
```diff
@@ -13,3 +18,20 @@ Bundle
 Detailed bundle examples and get started tutorial: https://github.com/Project-MONAI/tutorials/tree/main/bundle
 
 A collection of medical imaging models in the MONAI Bundle format: https://github.com/Project-MONAI/model-zoo
+
+
+Bundle vs. MAPs
```
vikashg, line 23: <html>
<body>
<google-sheets-html-origin><style type="text/css"></style>

  |   | Bundle | MAP
-- | -- | -- | --
  |   |   |  
  | When to use | To share a model with enough information about the pre and post transforms that someone can use to write their application | To share an application package which can be used with minimal effort to deploy an inference in an institutions. It contains piece of code with pre transforms. Although a model might be trained on Nifti JPEG image, the MAPs are geared towards dicom i/o
  | Challenges | To ensure that right amount of information is present in the model schema contained in the bundle so that the users can write their code to retrain the model or use it for inferenceing/evaluating | Have enough flexibility to account for different dicom IO and communication scenario. Some of them were heavily discussed on github for example dicom series selector, dicom sr writer etc.
  | Research | Mostly research use for building new models and sharing | The research is about building new DICOM pipelines. and writing self contained turnkey applications
  | Clinical | Learning from clinical use case need to write code (transforms) to incorporate new datatypes (stl, geometric annotations) etc | Needs research for incorporating messaging, formats like FHIR, HL7 and integrate it in the clinical operations and MONAI
  | Training | More or protential research in caching/ better GPU utilization and data pipelining | Not applicable
  | Validation | Need to incorporate any non standard metric someone may desire to compute | Not applicable
  | Inference | A bundle can be potentially used for inference.  Bundle defines the requirements for execution | Main Application of a MAP
  | DICOM I/O | Although a dicom file can be used for training it slows down the training process and GPU utilization by a factor of 3. So maybe tools need to be written which account for a proper conversion of a DICOM file to a JPEG or numpy array while maintaining a proper orientation/geomtery spacing etc. There needs to be tools which will incorporate more visual means of verifying that the conversion is correct. It is not easy as dicom formats vary a lot but something to be thought of. | Thre are lot of dicom tools in MAP Thanks to @Ming and @Gigon. However it is an evolving thing where new dicom formats specially SR Seg and RTs need to be incorporated in a systematic way. The problem is that there are more disagreements how data should be written and displayed across different viewers and different vendors. This part is challenging.
  |   |   |  


</body>
</html>

vikashg: maybe we can merge this information

vikashg (changes requested):
I hope this is useful. 
Thanks

docs/source/bundle_intro.rst
ericspod: Things which use MAPs do provide model serving in some form, eg. AIDE. I put this here because it's the eventual goal to integrating MAPs into these sorts of use cases even if it doesn't provide the functionality directly.

docs/source/bundle_intro.rst
ericspod: Thanks, I added a bit more description from this to what should be a brief discussion here about the differences. I don't think this should be the place for a longer discussion, just enough to dispel initial confusion about which is which.

wyli (approved):
Thanks, it looks good to me.

JHancox (approved):

PR #7131, opened 2023-10-13
Compose caching

### LINKED ISSUE #7130: Datasets that cache in some form should be able to handle nested Compose
The issue is as described in the Discussion #6172. Datasets that perform caching in some form can examine a top level Compose object but, if presented with nested Compose objects, treats those as being both randomizable and indivisible.

Given the following pipeline
```
Compose[Compose[A, B], Compose[C, D]]
```
where A, B and C are deterministic, and D is random
Dataset caching mechanisms should cache the output of C and execute D for each operation

This is identical behaviour to the pipeline:
```
Compose[A, B, C, D]
```

In order for this to be possible, we either have to:
 . flatten the heirarchy of all vanilla Compose objects (although this will cause issues)
 . execute datasets on a flattened heirarchy of Compose objects (although this will cause issues)
 . make the Caching Dataset iteration behaviour robust to nested Compose objects (my preferred option)

atbenmurray:
Fixes #7130

### Description

This is still a work in progress and a focus point for the discussion on #7130. At present, it provides an initial solution that only
solves the Compose always being treated as random part of the solution, but doesn't yet allow nested Composes to be iterated into.

### Types of changes


dzenanz (approved):
Code looks good on a glance. Are the preceding commits related to this fix?

atbenmurray:
> Code looks good on a glance. Are the preceding commits related to this fix?

Thanks for taking a look!
I'm out at the moment but I'll double check. I still want to modify the solution so it has the nested compose and flat compose equivalent behaviour anyway.

monai/transforms/compose.py
```diff
@@ -334,6 +334,18 @@ def __call__(self, input_, start=0, end=None, threading=False, lazy: bool | None
 
         return result
 
+    def is_random(self):
```
KumoLiu, line 337: Thanks for the PR!
Do we have another special reason to add this `is_random` function?
Looks like we could directly use this `flatten` function in the `CacheDataset`? Do you have concerns about this easy solution?
https://github.com/Project-MONAI/MONAI/blob/8a70678baa976a01274f55837e339f2c2975dec7/monai/transforms/compose.py#L313
https://github.com/Project-MONAI/MONAI/blob/8a70678baa976a01274f55837e339f2c2975dec7/monai/data/dataset.py#L808

atbenmurray: ```is_random``` is an initial attempt to solve the problem. I've thought about it over the weekend and if I can avoid introducing ```is_random``` then I'll get rid of it. In particular, if we always iterate into Compose objects, we don't need a check to determine whether the Compose object contains only provably deterministic transforms.

```flatten``` has its own problems in that it isn't a complete mechanism and it produces a physically flattened set of transforms. I'm prepping another commit that iterates over transforms without flattening them and returns an index instead.

atbenmurray:
Ok. I think I have a clean solution. I'm just testing it, then I'll either wrap up this PR and create a new one or I'll continue with this one.

Essentially, the iterative approach can replace the physical flattening quite nicely. This is good because we don't flatten the transforms when we execute the Compose pipeline but we do effective execute in a nested way. The solution is to modify execute_compose to be able to perform the iteration itself without flattening the data. The solution doesn't need is_random, although we might want to keep it (or something like it) so we don't have to keep a list of subclasses of Compose that we don't iterate into.

PR #7134, opened 2023-10-16
[WIP] Port MONAI Generative utils

marksgraham:
Towards completing #6676 .

### Description

### Types of changes

### DIFF
--- a/monai/utils/enums.py
+++ b/monai/utils/enums.py
@@ -13,8 +13,11 @@ from __future__ import annotations
 
 import random
 from enum import Enum
+from typing import TYPE_CHECKING
 
+from monai.config import IgniteInfo
 from monai.utils import deprecated
+from monai.utils.module import min_version, optional_import
 
 __all__ = [
     "StrEnum",
@@ -86,6 +89,14 @@ class StrEnum(str, Enum):
 
     def __repr__(self):
         return self.value
+
+
+if TYPE_CHECKING:
+    from ignite.engine import EventEnum
+else:
+    EventEnum, _ = optional_import(
+        "ignite.engine", IgniteInfo.OPT_IMPORT_VERSION, min_version, "EventEnum", as_type="base"
+    )
 
 
 class NumpyPadMode(StrEnum):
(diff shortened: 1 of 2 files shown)


marksgraham:
@ericspod I'm getting mypy errors from `unsqueeze_left` and `unsqueeze_right`:

```
mypy 1.4.1 (compiled: yes)
monai/utils/misc.py:895:5: error: Returning Any from function declared to return "T"  [no-any-return]
monai/utils/misc.py:895:12: error: Value of type "object" is not indexable  [index]
monai/utils/misc.py:895:43: error: "T" has no attribute "ndim"  [attr-defined]
monai/utils/misc.py:900:5: error: Returning Any from function declared to return "T"  [no-any-return]
monai/utils/misc.py:900:12: error: Value of type "object" is not indexable  [index]
monai/utils/misc.py:900:34: error: "T" has no attribute "ndim"  [attr-defined]
```

so far they only operate on tensors - am I ok to type them specifally for tensors/np arrays to satisfy mypy?

ericspod:
> am I ok to type them specifally for tensors/np arrays to satisfy mypy?

If you need to yes do that, but we do have [NdarrayOrTensor](https://github.com/Project-MONAI/MONAI/blob/0a3cdd9f1403cbad057b96230840ee17cf430dbf/monai/config/type_definitions.py#L69).

monai/utils/enums.py
```diff
@@ -692,3 +703,39 @@ class AlgoKeys(StrEnum):
     ALGO = "algo_instance"
     IS_TRAINED = "is_trained"
     SCORE = "best_metric"
+
+
+class AdversarialKeys(StrEnum):
```
KumoLiu, line 708: Would you mind adding a brief description here and also for `AdversarialIterationEvents`?

tests/test_squeeze_unsqueeze.py
```diff
@@ -0,0 +1,47 @@
...
+
+import unittest
+
+import numpy as np
+import torch
+from parameterized import parameterized
+
+from monai.utils import unsqueeze_left, unsqueeze_right
+
+RIGHT_CASES = [(np.random.rand(3, 4), 5, (3, 4, 1, 1, 1)), (torch.rand(3, 4), 5, (3, 4, 1, 1, 1))]
```
KumoLiu, line 22: Could you please also test other data types here?
An example here:
https://github.com/Project-MONAI/MONAI/blob/0a3cdd9f1403cbad057b96230840ee17cf430dbf/tests/test_normalize_intensity.py#L24

KumoLiu (approved):
Thanks for the PR, few minor comments inline.

monai/utils/component_store.py
```diff
@@ -0,0 +1,117 @@
...
+
+T = TypeVar("T")
+
+
+def is_variable(name):
+    """Returns True if `name` is a valid Python variable name and also not a keyword."""
+    return name.isidentifier() and not iskeyword(name)
+
+
+class ComponentStore:
```
wyli, line 27: we might be able to use/enhance the existing `LayerFactory` to achieve the same capability https://github.com/Project-MONAI/MONAI/blob/0a3cdd9f1403cbad057b96230840ee17cf430dbf/monai/networks/layers/factories.py#L76 

cc @ericspod I think the usage is https://github.com/Project-MONAI/GenerativeModels/blob/f28afa003f9b5be0ebf9f31270abc06e7957a607/generative/networks/schedulers/scheduler.py#L44

ericspod: > we might be able to use/enhance the existing `LayerFactory` to achieve the same capability

I thought about this before when writing this initially but concluded that there were too many differences between the two classes. We could refactor things such that they inherit from a common implementation, we need two classes because the usage here isn't to generate layers but be a factory for functions.

ericspod: But again these aren't network layers. The purpose of LayerFactory is to be a factory for the components of a network, so at the least we need to rename classes a bit.

wyli: indeed, the `LayerFactory` doesn't assume network components/blocks, the class could be renamed to a more generic one and deprecating the existing one. or have a refactoring into two subclasses as @ericspod mentioned. would you be interested in implementing it? @marksgraham

marksgraham: so a `Factory` class which `LayerFactory` and `ComponentStore` inherit from? That to me seems preferable to deprecating `LayerFactory`.

wyli: sounds good, thanks @marksgraham!

wyli (approved):
Thanks, it looks good to me.

PR #7137, opened 2023-10-17
Add `in_mm3` in `remove_small_objects`

### LINKED ISSUE #7034: Add min_size_in_mm to RemoveSmallObjects transform
RemoveSmallObjects is a great transform that is very much needed. However its arg is  
min_size (int) – objects smaller than this size (in pixel) are removed.
Tumors and organs in medical are measured in mm^3 or CC^3. As radiologist same remove tumors that are smaller than 3 mm 

I had to use a transform to resample to 1x1x1 so the pixel count is 1mm^3 remove the tumors then resample back to org size. Instead this should be a simple multiplication as we have the dims actual measurements in the metadata

KumoLiu:
Fixes #7034.

### Description
Add `in_mm3` in `remove_small_objects`, if `in_mm3=True`, `min_size` will be divided by pixdim when the image is MetaTensor.

### Types of changes

### DIFF
--- a/monai/transforms/post/array.py
+++ b/monai/transforms/post/array.py
@@ -361,7 +361,7 @@ class RemoveSmallObjects(Transform):
     Data should be one-hotted.
 
     Args:
-        min_size: objects smaller than this size (in pixel or in physical scale if physical_scale is True) are removed.
+        min_size: objects smaller than this size (in number of voxels; or volume in mm^3 if in_mm3 is True) are removed.
         connectivity: Maximum number of orthogonal hops to consider a pixel/voxel as a neighbor.
             Accepted values are ranging from  1 to input.ndim. If ``None``, a full
             connectivity of ``input.ndim`` is used. For more details refer to linked scikit-image
--- a/monai/transforms/post/array.py
+++ b/monai/transforms/post/array.py
@@ -361,7 +361,8 @@ class RemoveSmallObjects(Transform):
     Data should be one-hotted.
 
     Args:
-        min_size: objects smaller than this size (in number of voxels; or volume in mm^3 if in_mm3 is True) are removed.
+        min_size: objects smaller than this size (in number of voxels; or surface area/volume value
+            in whatever units your image is if by_measure is True) are removed.
         connectivity: Maximum number of orthogonal hops to consider a pixel/voxel as a neighbor.
             Accepted values are ranging from  1 to input.ndim. If ``None``, a full
             connectivity of ``input.ndim`` is used. For more details refer to linked scikit-image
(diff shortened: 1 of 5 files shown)


KumoLiu:
In this PR, I added the `resample` and `physical_scale` options. They can't both be true.
- If "resample=True," the image will resize to unit scale, remove objects smaller than "min_size," and then rescale back. This will result in a very slow transformation.
- If "physical_scale=True," "min_size" will be represented by physical scale, which means, if "min_size" is 2, multiply "physical_scale" by "pixdim" to represent the number of pixels in 2mm^3. Then this transform will remove the objects smaller than 2mm^3.

@AHarouni, could you please check whether this one can help address your issue?

monai/transforms/post/array.py
```diff
@@ -361,22 +361,68 @@ class RemoveSmallObjects(Transform):
     Data should be one-hotted.
 
     Args:
-        min_size: objects smaller than this size (in pixel) are removed.
+        min_size: objects smaller than this size (in pixel or in physical scale if physical_scale is True) are removed.
```
wyli, line 364: how about change `physical_scale` -> `in_mm3: whether the specified `min_size` is in number of voxels or volume in mm^3. otherwise we'll still have ambiguity in terms of millimeter/micrometer. it's fine to just assume the spacing is defined in millimeter

```suggestion
        min_size: objects smaller than this size (in number of voxels; or volume in mm^3 if in_mm3 is True) are removed.
```

monai/transforms/utils.py
```diff
@@ -1093,6 +1102,17 @@ def remove_small_objects(
...
+    if physical_scale:
+        sr = len(img.shape[1:])
+        if isinstance(img, monai.data.MetaTensor):
+            _pixdim = img.pixdim
+        elif pixdim is not None:
+            _pixdim = ensure_tuple_rep(pixdim, sr)
+        else:
+            warnings.warn("`img` is not of type MetaTensor and `pixdim` is None, assuming affine to be identity.")
+            _pixdim = (1.0,) * sr
+        min_size = int(np.prod(np.array(_pixdim)) * min_size)
```
wyli, line 1114:
```suggestion
        min_size = int(np.prod(np.array(_pixdim)) * min_size)
    elif pixdim is not None:
        warnings.warn(`pixdim is specified by not in use when computing the volume.`)
```

wyli (approved):
Thanks, it looks good to me.

ericspod:
This assumes the images are 3D, what about 2D? mm^3 isn't the unit being used in that case. Also we haven't specified what units we're working in, I would just state that the measure is in unit area or volume and not assume units are mm. Typically we resample in terms of square or cubic mm but not always and we don't encode in types or anywhere else what units we're dealing with, they've elsewhere always been implicit.

monai/transforms/post/array.py
```diff
@@ -361,22 +361,68 @@ class RemoveSmallObjects(Transform):
...
-        min_size: objects smaller than this size (in pixel) are removed.
+        min_size: objects smaller than this size (in number of voxels; or volume in mm^3 if in_mm3 is True) are removed.
         connectivity: Maximum number of orthogonal hops to consider a pixel/voxel as a neighbor.
             Accepted values are ranging from  1 to input.ndim. If ``None``, a full
             connectivity of ``input.ndim`` is used. For more details refer to linked scikit-image
             documentation.
         independent_channels: Whether or not to consider channels as independent. If true, then
             conjoining islands from different labels will be removed if they are below the threshold.
             If false, the overall size islands made from all non-background voxels will be used.
+        in_mm3: Whether the specified min_size is in number of voxels or volume in mm^3, default is false.
```
ericspod, line 372: So I would call this instead `by_measure` and say if this is True then `min_size` is a surface area or volume value of whatever units your image is in (mm^3, cm^2, etc.).

ericspod (approved):

PR #7140, opened 2023-10-18
Nested Compose iteration

### LINKED ISSUE #7130: Datasets that cache in some form should be able to handle nested Compose

atbenmurray:
Fixes #7130. This also replaces PR #7131 as it is a better solution with less impact.

### Description

Introduced compose_iterator, a function that knows how to iterate through the transforms of a Compose object, recursively iterating through nested Compose objects as necessary. This is used internally by flatten, execute_compose and __init__.

### Types of changes

### DIFF
--- a/monai/transforms/compose.py
+++ b/monai/transforms/compose.py
@@ -41,7 +41,85 @@ from monai.utils import MAX_SEED, TraceKeys, TraceStatusKeys, ensure_tuple, get_
 
 logger = get_logger(__name__)
 
-__all__ = ["Compose", "OneOf", "RandomOrder", "SomeOf", "execute_compose"]
+__all__ = [
+    "Compose",
+    "OneOf",
+    "RandomOrder",
+    "SomeOf",
+    "compose_iterator",
+    "ranged_compose_iterator",
+    "execute_compose",
+]
+
+
+def compose_iterator(compose, step_into_all=False):
+    """
+    ``compose_iterator`` is a function that returns an iterator over the transforms in a ``Compose``
+    instance.
+
+    It works by iterating over transforms, and stepping into nested ``Compose`` instances in a
+    recursive fashion. Given the following example:
+
+    ..code-block:: python
+
+    ::
+        t1, t2, t3, t4 = # some transforms
+        c = Compose([t1, Compose([t2, t3]), t4])
+        print([tx for tx in compose_iterator(c)])
+
+        > [t1, t2, t3, t4]
+
+    By default, ``compose_iterator`` doesn't step into ``OneOf``, ``SomeOf``, ``RandomOrder`` or
+    other classes that inherit ``Compose``. If ``step_into_all`` is passed True, it will step into
+    any Compose-like object as if it were just a container. As such, if you are using ``compose_iterator``
+    to step over objects as if executing them, ``step_into_all`` should be False. If you are using
+    ``compose_iterator`` to step over the static structure of the pipeline (such as if you are setting
+    or clearing flags on every transform), you should set ``step_into_all`` to True
+
+    Args:
+        compose: A `Compose` instance that contains the transforms to be iterated over
+        step_into_all: A boolean flag that indicates whether to step into randomised `Compose`
+            sub-classes such as `OneOf`, `SomeOf`, and `RandomOrder`
+    """
+
+    if not isinstance(compose, Compose):
+        raise ValueError(f"must be of type {(type.Compose)} but is type {type(compose)}")
+
+    for i in range(len(compose.transforms)):
+        tx = compose.transforms[i]
+        if type(tx) == Compose or (step_into_all is True and isinstance(tx, Compose)):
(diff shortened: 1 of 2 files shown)


monai/transforms/compose.py
```diff
@@ -41,7 +41,84 @@
...
+
+    for i in range(len(compose.transforms)):
+        tx = compose.transforms[i]
+        if type(tx) == Compose or (step_into_all is True and isinstance(tx, Compose)):
+            yield from compose_iterator(tx)
+        else:
+            yield tx
+
+
+def ranged_compose_iterator(compose, start=None, end=None, step_into_all=False):
```
ericspod, line 96: Would [islice](https://docs.python.org/3/library/itertools.html#itertools.islice) do the same?

atbenmurray: No, because we have the special semantic that there are containers that we don't step into.

monai/transforms/compose.py
```diff
@@ -41,7 +41,84 @@
...
+    recursive fashion. Given the following example:
+
+    ..code-block:: python
+
+    ::
+        t1, t2, t3, t4 = # some transforms
+        c = Compose([t1, Compose([t2, t3]), t4])
+        print([tx for tx in compose_iterator(c)])
+
+        > [t1, t2, t3, t4]
```
KumoLiu, line 70: It looks like the `compose_iterator` returns the flattened transforms, which may ignore the `lazy` arg in the nested `Compose`?

atbenmurray:
> I think we could send a proper `predicate` to `get_index_of_first` to get the right index without modifying the existing function. In this way, we still consider the nested `Compose` as a single transform.
> 
> https://github.com/Project-MONAI/MONAI/blob/9f1168f43c4cb22cf4ef4ce72caeffd7209fa772/monai/data/dataset.py#L889
> 
> Something like:
> 
> ```
> def is_random(t):
>     if type(t) is Compose:
>         for _t in t.transforms:
>             if is_random(_t):
>                 return True
>         return False
>     elif isinstance(t, RandomizableTrait) or not isinstance(t, Transform):
>         return True
>     return False
> ```
> 
> cc @wyli @ericspod @Nic-Ma

`is_random` isn't sufficient to solve the problem though. With `is_random` you get the ability to determine whether a Compose object has only deterministic transforms, but you can't decide to run the pipeline up to part-way through a nested Compose.

```
# c is a compose object
# dn is a deterministic transform
# rn is a random transform

c1 = [d1, d2]
c2 = [d3, r4]
c = [c1, c2]
```
we want `[[d1, d2], [d3]]` to be run deterministically and cached
then we want `[[r4]]` to be run each time we call the pipeline

monai/transforms/compose.py
atbenmurray: It's also true for `overrides` and anything else that might be set on a Compose. I'm looking into a nice way to handle this

dzenanz (commented):
I am fine with this approach too, assuming it works 😄

atbenmurray:
> I am fine with this approach too, assuming it works 😄

@dzenanz @KumoLiu I think this is closing in on a fully working solution. Please take a look when you have a moment.

monai/transforms/compose.py
```diff
@@ -41,7 +41,116 @@
...
+
+    Args:
+        compose: A `Compose` instance that contains the transforms to be iterated over
+        step_into_all: A boolean flag that indicates whether to step into randomised `Compose`
+            sub-classes such as `OneOf`, `SomeOf`, and `RandomOrder`
+    """
+
+    if isinstance(compose, (list, tuple)):
+        transforms = compose
+    elif type(compose) is Compose or (step_into_all is True and isinstance(compose, Compose)):
```
wyli, line 87: this is not a pythonic way of handling the compose and its other potential subclasses. let's not merge this into core.

atbenmurray: It's pretty much how flatten operates, so its no less pythonic than what is already in the repo

atbenmurray: https://github.com/Project-MONAI/MONAI/blob/fbbb64b4cb2facffccbf4fdf6c44e20f8d31ca10/monai/transforms/compose.py#L322

PR #7141, opened 2023-10-18
refacto generate_spatial_bounding_box more general

vgrau98:
Fixes # 7033

### Description

generate_spatial_bounding_box returns only one BB. In several cases we want to compute bb for multiple components (tumors for example).

The use of the function could be clarified. Do we want a single bounding box encompassing the entire foreground, or a function returning as many bounding boxes as there are components (tumors)? 

In the first case, the function should not be modified, and a new one should be developed to respond to the issue presented in 7033. For the moment, the second case is being considered and the function modified directly. 

### Types of changes

### DIFF
--- a/monai/transforms/utils.py
+++ b/monai/transforms/utils.py
@@ -954,7 +954,7 @@ def _create_translate(
 
 
 @deprecated_arg_default("allow_smaller", old_default=True, new_default=False, since="1.2", replaced="1.5")
-def generate_spatial_bounding_box(
+def generate_one_spatial_bounding_box(
     img: NdarrayOrTensor,
     select_fn: Callable = is_positive,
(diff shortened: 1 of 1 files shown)


wyli:
thanks for the PR, please develop a new function based on the existing `generate_spatial_bounding_box`, because the connected components might have a relatively large memory footprint which is different from the current lightweight utility function.

KumoLiu:
Hi @vgrau98,  is there any more progress on this PR so far, do you plan to finish this one?
Thanks!

vgrau98:
> Hi @vgrau98, is there any more progress on this PR so far, do you plan to finish this one? Feel free to ask if you have any issues. Thanks!

Yes, it'll be finished by this weekend at the latest, I've been pretty busy these last few weeks :)

PR #7148, opened 2023-10-20
Using keys in .npz files for `NumpyReader`

### LINKED ISSUE #7147: Bug: NumpyReader for .npz files with variable names.
https://github.com/Project-MONAI/MONAI/blob/6e5fdc0066d9b21a2b50153f5ba1951d0d82f8e4/monai/data/image_reader.py#L1071

If npz_keys is not given, `NumpyReader` assumes variables in .npz files to be named arr_0, arr_1,...
This may or may not be true. It would be better to use the variable names from the file like so:
```python
                npz_keys = self.npz_keys if self.npz_keys else list( img.keys() ) # or img.files
```

KumoLiu:
Fixes #7147.

### Types of changes

### DIFF
--- a/monai/data/image_reader.py
+++ b/monai/data/image_reader.py
@@ -1068,7 +1068,7 @@ class NumpyReader(ImageReader):
             img = np.load(name, allow_pickle=True, **kwargs_)
             if Path(name).name.endswith(".npz"):
                 # load expected items from NPZ file
-                npz_keys = [f"arr_{i}" for i in range(len(img))] if self.npz_keys is None else self.npz_keys
+                npz_keys = list(img.keys()) if self.npz_keys is None else self.npz_keys
                 for k in npz_keys:
                     img_.append(img[k])
             else:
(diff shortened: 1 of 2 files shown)


monai/data/image_reader.py
```diff
@@ -1068,7 +1068,7 @@ def read(self, data: Sequence[PathLike] | PathLike, **kwargs):
             img = np.load(name, allow_pickle=True, **kwargs_)
             if Path(name).name.endswith(".npz"):
                 # load expected items from NPZ file
-                npz_keys = [f"arr_{i}" for i in range(len(img))] if self.npz_keys is None else self.npz_keys
+                npz_keys = list(img.keys()) if self.npz_keys is None else self.npz_keys
```
wyli, line 1071: The existing implementation is still good for array inputs, perhaps we use the keys only if 'img' is a dictionary?

KumoLiu: It also works for array inputs, the keys are only used in the npz file.
https://github.com/numpy/numpy/blob/d35cd07ea997f033b2d89d349734c61f5de54b0d/numpy/lib/npyio.py#L725

wyli (approved):
Thanks, it looks good to me.

PR #7158, opened 2023-10-24
Refactor `self.weight` in `DiceLoss` and `FocalLoss`

### LINKED ISSUE #7065: DiceLoss add weight per class parameter

KumoLiu:
Fixes #7065

### Description
Remove `self.weight` in `DiceLoss` and `FocalLoss`

### Types of changes

### DIFF
--- a/monai/losses/dice.py
+++ b/monai/losses/dice.py
@@ -113,6 +113,7 @@ class DiceLoss(_Loss):
         self.batch = batch
         weight = torch.as_tensor(weight) if weight is not None else None
         self.register_buffer("class_weight", weight)
+        self.class_weight: None | torch.Tensor
 
     def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
         """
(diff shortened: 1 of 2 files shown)


wyli (approved):
Thanks, it looks good to me.

PR #7159, opened 2023-10-24
[WIP] 7145 common factory class

marksgraham:
Fixes # .

### Description

### Types of changes


monai/networks/layers/factories.py
```diff
@@ -82,15 +82,7 @@ class LayerFactory:
...
 
-    @property
-    def names(self) -> tuple[str, ...]:
-        """
-        Produces all factory names.
-        """
-
-        return tuple(self.factories)
-
-    def add_factory_callable(self, name: str, func: Callable) -> None:
```
ericspod, line 93: I would keep this name if even just as an alias so that it's clearer what we're adding to the factory.

monai/networks/layers/factories.py
```diff
@@ -103,17 +95,6 @@ def add_factory_callable(self, name: str, func: Callable) -> None:
             + ".\nPlease see :py:class:`monai.networks.layers.split_args` for additional args parsing."
         )
 
-    def factory_function(self, name: str) -> Callable:
```
ericspod, line 106: I would also leave this method here if the class inherits from `ComponentStore` and leave it's name so that it's clearer what it is decorating.

monai/networks/layers/factories.py
```diff
@@ -274,91 +259,91 @@ def instance_nvfuser_factory(dim):
...
-Act.add_factory_callable("prelu", lambda: nn.modules.PReLU)
-Act.add_factory_callable("relu6", lambda: nn.modules.ReLU6)
-Act.add_factory_callable("selu", lambda: nn.modules.SELU)
-Act.add_factory_callable("celu", lambda: nn.modules.CELU)
-Act.add_factory_callable("gelu", lambda: nn.modules.GELU)
-Act.add_factory_callable("sigmoid", lambda: nn.modules.Sigmoid)
-Act.add_factory_callable("tanh", lambda: nn.modules.Tanh)
-Act.add_factory_callable("softmax", lambda: nn.modules.Softmax)
-Act.add_factory_callable("logsoftmax", lambda: nn.modules.LogSoftmax)
+Act.add("elu", lambda: nn.modules.ELU)
```
wyli, line 262: thanks for the refactoring! with the new API, I suspect the `lambda: nn.modules.ELU` could be further simplified as `nn.modules.ELU`...

ericspod: Remember though that what's being added here is a callable which the factory will call to return the layer that you want. We could add a `add_type` method which takes the type and does this lambda thing.

marksgraham: I added an `add_factory_class` method that does the lambda internally, and simplified the activations you pointed out @wyli. It looks like a lot more could be further simplified now, actually - all the factory additions which only supply a single class. Shall I go ahead and do that?

monai/networks/layers/factories.py
```diff
@@ -68,34 +68,36 @@ def use_factory(fact_args):
...
     Factory object for creating layers, this uses given factory functions to actually produce the types or constructing
     callables. These functions are referred to by name and can be added at any time.
     """
 
-    def __init__(self) -> None:
-        self.factories: dict[str, Callable] = {}
-
-    @property
-    def names(self) -> tuple[str, ...]:
+    def add_factory_callable(self, name: str, func: Callable, desc: str | None = None) -> None:
```
ericspod, line 82: Just a mild suggestion, if `name` is None this could be taken from the `__name__` member if present. 

```suggestion
    def add_factory_callable(self, func: Callable, name: str | None = None, desc: str | None = None) -> None:
        name = name if name is not None else getattr(func, "__name__", "???")
```

monai/networks/layers/factories.py
```diff
@@ -68,34 +68,36 @@ def use_factory(fact_args):
...
+            "The supported member"
+            + ("s are: " if len(self.names) > 1 else " is: ")
+            + ", ".join(f"``{name}``" for name in self.names)
+            + ".\nPlease see :py:class:`monai.networks.layers.split_args` for additional args parsing."
+        )
 
-        return tuple(self.factories)
-
-    def add_factory_callable(self, name: str, func: Callable) -> None:
+    def add_factory_class(self, name: str, cls: type, desc: str | None = None) -> None:
```
ericspod, line 95: This could call `add_factory_callable` with `lambda x=None: cls` passed as `func`?

monai/networks/layers/factories.py
```diff
@@ -68,34 +68,36 @@ def use_factory(fact_args):
...
-    @property
-    def names(self) -> tuple[str, ...]:
+    def add_factory_callable(self, name: str, func: Callable, desc: str | None = None) -> None:
         """
-        Produces all factory names.
+        Add the factory function to this object under the given name, with optional description.
         """
+        description: str = desc or func.__doc__ or ""
+        self.add(name.upper(), description, func)
+        self.__doc__ = (
```
ericspod, line 88: This is going to overwrite the docstring you've already written, maybe the last line of the docstring should be "Please see :py:class:`monai.networks.layers.split_args` for additional args parsing.\n\nThe supported member(s) are:" then append new  lines when something is added with this method.

marksgraham: adressed so it now appends. unfortunately needs an assert before the append to stop mypy throwing errors!

PR #7165, opened 2023-10-26
simplify `list_data_collate` and `collate_meta_tensor`

### LINKED ISSUE #5917: simplify `list_data_collate` and `collate_meta_tensor ` using `collate`
the generic `collate` API is [implemented in pytorch ](https://github.com/pytorch/pytorch/blob/d7a3f2128fb4457dd60fd5d23e77d2c66a8b0f02/torch/utils/data/_utils/collate.py#L88-L92)

https://github.com/pytorch/pytorch/pull/85748

```
def collate(batch, *, collate_fn_map: Optional[Dict[Union[Type, Tuple[Type, ...]], Callable]] = None):
    r"""
        General collate function that handles collection type of element within each batch
        and opens function registry to deal with specific element types. `default_collate_fn_map`
        provides default collate functions for tensors, numpy arrays, numbers and strings.

```

probably the monai helper functions could be simplified:
https://github.com/Project-MONAI/MONAI/blob/50c3f328cb98e4f0c4e6a38a04a98badf0961c18/monai/data/utils.py#L448-L451

https://github.com/Project-MONAI/MONAI/blob/50c3f328cb98e4f0c4e6a38a04a98badf0961c18/monai/data/utils.py#L427-L430

KumoLiu:
Fixes #5917

### Types of changes

### DIFF
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -449,7 +449,8 @@ def pickle_operations(data, key=PICKLE_KEY_SUFFIX, is_encode: bool = True):
 
 def collate_meta_tensor_fn(batch, *, collate_fn_map=None):
     """
-    Collate a sequence of meta tensor into a single batched metatensor. This is called by `collage_meta_tensor` and so should not be used as a collate function directly in dataloaders.
+    Collate a sequence of meta tensor into a single batched metatensor. This is called by `collage_meta_tensor`
+    and so should not be used as a collate function directly in dataloaders.
     """
     collate_fn = collate_tensor_fn if pytorch_after(1, 13) else default_collate
     collated = collate_fn(batch)  # type: ignore
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -54,7 +54,7 @@ from monai.utils import (
 )
 
 if pytorch_after(1, 13):
-    # import private code for reuse purposes
+    # import private code for reuse purposes, comment in case things break in the future
     from torch.utils.data._utils.collate import collate_tensor_fn, default_collate_fn_map
 pd, _ = optional_import("pandas")
 DataFrame, _ = optional_import("pandas", name="DataFrame")
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -448,10 +448,9 @@ def pickle_operations(data, key=PICKLE_KEY_SUFFIX, is_encode: bool = True):
 
 
 def collate_meta_tensor_fn(batch, *, collate_fn_map=None):
-    """collate a sequence of meta tensor sequences/dictionaries into
-    a single batched metatensor or a dictionary of batched metatensor"""
+    """collate a sequence of meta tensor into a single batched metatensor"""
     collate_fn = collate_tensor_fn if pytorch_after(1, 13) else default_collate
-    collated = collate_fn(batch)
+    collated = collate_fn(batch)  # type: ignore
     meta_dicts = [i.meta or TraceKeys.NONE for i in batch]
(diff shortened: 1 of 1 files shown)


wyli (approved):
thanks, this is a minor refactoring to use the latest pytorch API, please let me know any concerns @ericspod @Nic-Ma.

monai/data/utils.py
```diff
@@ -444,22 +447,28 @@ def pickle_operations(data, key=PICKLE_KEY_SUFFIX, is_encode: bool = True):
     return data
 
 
+def collate_meta_tensor_fn(batch, *, collate_fn_map=None):
+    """collate a sequence of meta tensor into a single batched metatensor"""
```
ericspod, line 451:
```suggestion
    """
    Collate a sequence of meta tensor into a single batched metatensor. This is called by `collage_meta_tensor` and so should not be used as a collate function directly in dataloaders.
    """
```

I think we should say a little more to make it clear what this function is and to use `collage_meta_tensor` instead.

monai/data/utils.py
```diff
@@ -50,8 +50,11 @@
     issequenceiterable,
     look_up_option,
     optional_import,
+    pytorch_after,
 )
 
+if pytorch_after(1, 13):
+    from torch.utils.data._utils.collate import collate_tensor_fn, default_collate_fn_map
```
ericspod, line 57: Since the module is `_utils` I wonder if this is meant to be private code we shouldn't rely on. What is the advantage of using these functions over the ones we've been using so far?

monai/data/utils.py
```diff
@@ -479,9 +490,15 @@ def list_data_collate(batch: Sequence):
         Need to use this collate if apply some transforms that can generate batch data.
 
     """
+
+    if pytorch_after(1, 13):
```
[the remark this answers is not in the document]
KumoLiu: Yes, you are right. But it needs to go here to avoid circular import.

monai/data/utils.py
KumoLiu: The benefits may be we could simplify L463-L466 after we add a new collate map for `MetaTensor`.
https://github.com/Project-MONAI/MONAI/blob/85243f52085b0946870fc7808c484c8b639eaf1b/monai/data/utils.py#L463-L466

ericspod: I would just be careful using `_utils` directly. [Here](https://github.com/pytorch/pytorch/blob/d7a3f2128fb4457dd60fd5d23e77d2c66a8b0f02/torch/utils/data/dataloader.py#L64C37-L64C37) the function is imported into a public module. It seems `torch.utils.data.dataloader.default_collate` is meant to be the public member which just happens to be implemented as an alias to a private one, but this could change without breaking the public interface.

KumoLiu: Yes, we should always be careful with the private code. I didn't find an alias for these two. 
But it seems they tend to extend collate functions that can register collate functions to handle specific types.
https://github.com/pytorch/pytorch/pull/85748

Do you have some suggestions for using these or we could just leave them here?

ericspod: It seems that `default_collate_fn_map` is meant to be internal and used only by `default_collate`, but `collate` is also private so shouldn't be used either? We have no way of tweaking collate behaviour if so. I wouldn't myself play around with Pytorch internals like this but there's no other way it seems to get your behaviour and reuse things. Let's put in comments where you do the imports stating you're importing private code for reuse purposes so we keep track in case things break in the future.

KumoLiu: Sure, thanks! I added comments in the latest commit.

PR #7168, opened 2023-10-27
Enable Chaining in Auto3DSeg CLI 

### LINKED ISSUE #7167: Auto3DSeg AutoRunner Fire CLI Chaining Fails
**Describe the bug**
Auto3DSeg `AutoRunner` uses Fire to generate CLI. Fire supports chaining of multiple methods in a single command. However, it fails for `AutoRunner` and they cannot chained. I realized that the `AutoRunner` methods does not return `self`, which is a requirements for Fire CLI chaining to work.

drbeh:
Fixes #7167

### Description
Make all the setting methods in `AutoRunner` to return `self` to enable chaining in cli.

### Types of changes

### DIFF
--- a/monai/apps/auto3dseg/auto_runner.py
+++ b/monai/apps/auto3dseg/auto_runner.py
@@ -442,6 +442,8 @@ class AutoRunner:
         if gpu_customization_specs is not None:
             self.gpu_customization_specs = gpu_customization_specs
 
+        return self
+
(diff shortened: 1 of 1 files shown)


mingxin-zheng (approved):
Thanks. According to the [usage guide](https://google.github.io/python-fire/guide/#chaining-function-calls), this should enable the chaining function call of AutoRunner in command line. Maybe it would be good to add some docstring there too.

PR #7176, opened 2023-10-31
[Auto3DSeg] Add mlflow support in autorunner.

dongyang0122:
Add MLflow support in AutoRunner Class.

### Types of changes


monai/apps/auto3dseg/auto_runner.py
```diff
@@ -209,6 +211,7 @@ def __init__(
         not_use_cache: bool = False,
         templates_path_or_url: str | None = None,
         allow_skip: bool = True,
+        mlflow_tracking_uri: str | None = None,
```
drbeh, line 214: I think having a method `set_mlflow_params(mlflow_tracking_uri=None, **kwargs)` might be better as we may need to modify more parameters of mlflow?

monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -129,6 +130,17 @@ def set_data_source(self, data_src_cfg: str) -> None:
         """
         self.data_list_file = data_src_cfg
 
+    def set_mlflow_tracking_uri(self, mlflow_tracking_uri: str | None = None) -> None:
```
drbeh, line 133: The `__init__` is already accepting `mlflow_tracking=_uri`, do we really need this method here? I believe we need either this or that (the same for AutoRunner), what do you think?

myron:
I think this should be a part of kwargs of Autorunner, instead of adding more names parameters for very specific use case. 

Then you check for it as 
If kwargs.get('mlflow_url', None) is not None:

drbeh:
> I think this should be a part of kwargs of Autorunner, instead of adding more names parameters for very specific use case.
> 
> Then you check for it as If kwargs.get('mlflow_url', None) is not None:

@myron @dongyang0122 As the zen of python says "explicit is better than implicit". In my opinion, having more names means better documentation. Deferring to kwargs usually makes sense when dealing with a third party parameters that might change over time without out notice, but mlflow_tracking_uri is our parameter and we decide how to use it (although we are passing it to mlflow).

monai/apps/auto3dseg/bundle_gen.py
dongyang0122: I think that the modification of more parameters of mlflow would be better.

monai/apps/auto3dseg/auto_runner.py
dongyang0122: Can you provide a few examples about which params can be modified?

monai/apps/auto3dseg/bundle_gen.py
dongyang0122: But we need to define the scope of the parameters to cover in the first place.

dongyang0122: Having this function follows conventions as other similar functions in `BundleAlgo` and `BundleGen`, for example, `set_data_stats`.

drbeh: I agree with you and I like this method but I believe we don't need the redundancy of having both this method and a parameter in the init method. If we keep this, we may need to remove `mlflow_tracking_uri` from __init__. What do you think?

monai/apps/auto3dseg/auto_runner.py
dongyang0122: Okay.

monai/apps/auto3dseg/bundle_gen.py
dongyang0122: I follow the design of `data_stats_filename` and `data_src_cfg_name` in the same class. @mingxin-zheng could you please share some thoughts here? Thanks!

dongyang0122:
> > I think this should be a part of kwargs of Autorunner, instead of adding more names parameters for very specific use case.
> > Then you check for it as If kwargs.get('mlflow_url', None) is not None:
> 
> @myron @dongyang0122 As the zen of python says "explicit is better than implicit". In my opinion, having more names means better documentation. Deferring to kwargs usually makes sense when dealing with a third party parameters that might change over time without our notice, but mlflow_tracking_uri is our parameter and we decide how to use it (although we are passing it to mlflow).

I acknowledge the perspectives presented from both sides. Incorporating an input variable `mlflow_tracking_uri` would enhance parameter visibility and documentation, whereas including it under ‘kwargs’ could conserve script space. Personally, I favor the first approach, though both methods are viable. Input from individuals @Nic-Ma and @daguangxu on this matter would be beneficial to finalize the decision.

Nic-Ma:
Hi @drbeh , @dongyang0122 ,

Do we have any usecase to allow users to set `experiment_name`? If no, I would vote for current implementation with `mlflow_tracking_uri` arg directly.

Thanks.

Nic-Ma (approved):

PR #7178, opened 2023-11-01
Implemented VoxelMorph

### LINKED ISSUE #5484: Adding voxelmorph in MONAI?
Adding 3d image registration model called voxelmorph https://arxiv.org/pdf/1809.05231.pdf.

Trainable voxelmorph model.

kvttt:
Fixes #5484.

### Description

Implemented VoxelMorph and added some docstrings. Checked coding style locally.

### Types of changes


tests/test_voxelmorph.py
```diff
@@ -0,0 +1,316 @@
...
+    def test_script(self):
+        net = VoxelMorph(
+            spatial_dims=3,
+            in_channels=2,
+            unet_out_channels=32,
+            channels=(16, 32, 32, 32, 32, 32),
+            final_conv_channels=(16, 16),
+        ).net
+        test_data = torch.randn(1, 2, 160, 192, 224)
+        test_script_save(net, test_data)
```
kvttt, line 307: Hi @wyli, I wonder if this would be the appropriate way to test saving the script. I tried to save the entire model (including the `DVF2DDF` for velocity field integration and `Warp` for applying the deformation). But it looks like jit doesn't like the way `Warp` is implemented:

```
RuntimeError: 
cannot statically infer the expected size of a list in this context:
  File "/home/runner/work/MONAI/MONAI/monai/networks/blocks/warp.py", line 125
        if spatial_dims not in (2, 3):
            raise NotImplementedError(f"got unsupported spatial_dims={spatial_dims}, currently support 2 or 3.")
        ddf_shape = (image.shape[0], spatial_dims) + tuple(image.shape[2:])
                                                     ~~~~~~~~~~~~~~~~~~~~ <--- HERE
```

So, for the sake of passing the tests, I am trying to save only the network part. Is there a more elegant way of passing the tests without having to leave these two modules behind? Or would it be possible to just rewrite this line in a way that jit likes?

https://github.com/Project-MONAI/MONAI/blob/cf886e7dc41361b61df3d612d897963d4043c61d/monai/networks/blocks/warp.py#L125

wyli: thanks for looking into this, it seems that the `ddf_shape` variable is just evaluated for a minor shape checking https://github.com/Project-MONAI/MONAI/blob/cf886e7dc41361b61df3d612d897963d4043c61d/monai/networks/blocks/warp.py#L125-L130 I feel we can just remove the check in favour of jit.

monai/networks/nets/voxelmorph.py
```diff
@@ -0,0 +1,392 @@
...
+from monai.utils import alias, export
+
+__all__ = ["VoxelMorph", "voxelmorph"]
+
+
+@export("monai.networks.nets")
+@alias("voxelmorph")
+class VoxelMorph(nn.Module):
+    """
+    VoxelMorph network for medical image registration as described in https://arxiv.org/pdf/1809.05231.pdf.
```
wyli, line 33:
```suggestion
    A re-implementation of VoxelMorph network for medical image registration as described in https://arxiv.org/pdf/1809.05231.pdf.
https://github.com/voxelmorph/voxelmorph
```

wyli:
CCing @ebrahimebrahim @brudfors @nvahmadi in case you have comments/suggestions here..

wyli (approved):
thanks, it looks good to me in general, have you used this to train some baseline successfully? also please help make the test cases smaller if possible (like input size (2, 1, 96, 96, 48)?), so they can run with minimal system resources of the github runner.

nvahmadi:
Thanks for tagging me, @wyli and congrats @kvttt - this looks like a valuable contribution, awesome in fact! Thanks a lot! :) Some thoughts on this PR:

- I like the input parameter for integration steps and that a number `int_steps>0` automatically switches on diffeomorphic registration. To be more explicit, I would recommend to rename the parameter to "integration_steps", because the prefix "int" reminds me of "integer" and the suffix "steps" alone does not really explain what this parameter is about. 
- Also, I saw that you implemented the construction of a U-Net all from scratch. I wonder whether there was an explicit reason to reimplement this? We have the U-Net class, the DynUNet class, both are nicely parametrizable, and the VoxelMorph backbone can be mimicked with these classes (e.g. by setting an appropriate number of channels etc). An advantage over the reimplementation would be that the existing MONAI (Dyn)UNet implementations offer a lot of extra functionality (e.g. in-layer residuals, anisotropic kernels and strides etc.), which all would need to be reimplemented in this class, if we wanted to cover them.
- Thinking this a bit further, the VoxelMorph approach does not even require a U-Net backbone, it can be any fully convolutional architecture that maps the input to a same-sized 3-channel output (either DVF or DDF). For example, in a recent [tutorial](https://github.com/Project-MONAI/tutorials/blob/main/3d_registration/learn2reg_nlst_paired_lung_ct.ipynb), we used a SegResNet as the mapping backbone. One could even consider letting the user choose the backbone at constructor level (either as a string, or as a class definition), and letting appropriate network params be passed as an additional constructor parameter (a dictionary, in this case). The default class would then be a UNet without in-layer residuals and the params set up in this PR (to mimic the original VoxelMorph paper). Overall, this would allow much more flexibility for the user to explore different backbones, e.g. [UNet](https://docs.monai.io/en/stable/networks.html#unet), [DynUNet](https://docs.monai.io/en/stable/networks.html#dynunet), [AttentionUNet](https://docs.monai.io/en/stable/networks.html#attentionunet), [SegResNet](https://docs.monai.io/en/stable/networks.html#segresnet) and event 3D-ViTs like ([Swin-](https://docs.monai.io/en/stable/networks.html#swinunetr))[UNETR](https://docs.monai.io/en/stable/networks.html#unetr) etc.

Apart form that though, as said, I think this is a very welcome addition to MONAI Core. I would recommend to also create an accompanying tutorial. We could use the OASIS subtask from the Learn2Reg challenge as a dataset. Feel free to DM me to sync on this, we are working on this as well, would be great to collaborate on this to avoid duplicate efforts - ideally we could even use your new VoxelMorph class. :)

brudfors:
Thanks @kvttt! 

I agree with @nvahmadi that it would be preferable to make use of the networks that are already in MONAI, so that there are not multiple implementations of a UNet let's say. Looking at the [Paired Lung CT 3d Registration tutorial](https://github.com/Project-MONAI/tutorials/blob/main/3d_registration/paired_lung_ct.ipynb); another way of including a VoxelMorph model in MONAI could be to simply add a new tutorial to MONAI, similar to the Lung CT 3d Registration, but replacing the `LocalNet` with a `UNet` (and whatever hyper-parameters that were used in the VoxelMorph paper) and switching its `forward` function with the one implemented in @kvttt's PR. Perhaps with the OASIS brain scans used in the learn2reg challenge, as @nvahmadi proposed.

wyli:
in the current implementation there's already `self.net` decoupling the backbone from the rest of the network, perhaps we can further make `self.net` configurable by providing an argument `backbone`
```py
my_convnet = MyConvNet(...)
model = VoxelMorph(backbone=my_convnet, ...)
```

nvahmadi:
Thanks @kvttt, good point, the UNets in MONAI are all based on convolution for down-/upsampling, whereas VoxelMorph was still relying on pooling. It's a good idea to keep your implementation as the "original VoxelMorph", but offering to pass a different backbone, as @wyli suggested. Great to hear that you're supportive of this idea. :) This will be a very nice and flexible addition.

ebrahimebrahim:
This looks amazing! And I agree with earlier comments about have an "original voxelmorph mode" and a "provide your own backbone" mode, including decoupling the current construction of the `self.net` into its own `VoxelMorphConvNet` (or whatever name makes sense). Continuing from the comment in https://github.com/Project-MONAI/MONAI/pull/7178#issuecomment-1790781540 one could then build the original voxelmorph with  
```py
my_convnet = VoxelMorphConvNet(in_channels, ...)
model = VoxelMorph(backbone=my_convnet, ...`) 
```

kvttt:
In response to @wyli's comment

> in the current implementation there's already `self.net` decoupling the backbone from the rest of the network, perhaps we can further make `self.net` configurable by providing an argument `backbone`
> 
> ```python
> my_convnet = MyConvNet(...)
> model = VoxelMorph(backbone=my_convnet, ...)
> ```

and @ebrahimebrahim's comment,

> This looks amazing! And I agree with earlier comments about have an "original voxelmorph mode" and a "provide your own backbone" mode, including decoupling the current construction of the `self.net` into its own `VoxelMorphConvNet` (or whatever name makes sense). Continuing from the comment in [#7178 (comment)](https://github.com/Project-MONAI/MONAI/pull/7178#issuecomment-1790781540) one could then build the original voxelmorph with
> 
> ```python
> my_convnet = VoxelMorphConvNet(in_channels, ...)
> model = VoxelMorph(backbone=my_convnet, ...`) 
> ```

I made the distinction clear in my code between `VoxelMorph` (referring to the framework) versus `VoxelMorphUNet` (the backbone network) and added the option to specify a backbone network.

Issue #6904, opened 2023-08-24
Generalize GridPatchDataset and PatchDataset. Potentially introduce Persistent and/or Cache IterableDatasets.


ibro45:
**Is your feature request related to a problem? Please describe.**

- `GridPatchDataset`'s `data` argument must be a sequence of already loaded data. However, it would be consistent with other datasets to allow loading and manipulating other types of data through `transform`. For example, you might want to provide a dict of image and label paths and load them through transforms using `LoadImaged()`.
- `PatchDataset` should follow the same. Moreover, it has a `transform` argument that is actually the same as `patch_transform` in `GridPatchDataset`, so it should follow that naming.
- Finally, it would be useful to have Persistent or Cache versions of these (or just any `IterableDataset`) - is that something that you would want to have?

**Describe the solution you'd like**

- `GridPatchDataset` currently has `patch_transform`, but should have two transforms: `transform` and `patch_transform`. The `transform` is to be used the same as in the IterableDataset and is applied prior to patching, while `patch_transform` is applied to the patch.
- `PatchDataset` should follow the same.

**Describe alternatives you've considered**
None

**Additional context**
I started looking into these patch-based datasets because I want to perform validation on large images during my training. Specifically, images are sometimes too big to perform SlidingWindowInferer on GPU, and using `cpu_thresh` just results in a very long computation. I hope to use this as a replacement for it, and it should be way faster if Persistent/Cache mode is available too. It basically emulates sliding window inference - still allows you to split into patches with some overlap and pass them through the network to get the prediction, but with a caveat that the predicted patches are never merged and instead evaluated against the corresponding label patches individually. 

I am happy to make a PR for the first two points. I am also interested in implementing the third point if you agree with it and if we discuss how we should approach it.

KumoLiu:
Hi, @ibro45, thanks for your proposal!
- Is it possible to use `GridPatchDataset` right after a `Dataset`? Something similar in this [tutorial](https://github.com/Project-MONAI/tutorials/blob/87792475f4476dcc63f92e47e109dde18cdfb30d/modules/nifti_read_example.ipynb).
- It seems to be `transform` in both `GridPatchDataset` and `PatchDataset`
https://github.com/Project-MONAI/MONAI/blob/2daabf917a1e6afbc738d22db602f4859672a90d/monai/data/grid_dataset.py#L185
[...]

PR #7180, opened 2023-11-01
Add cache option in `GridPatchDataset`

KumoLiu:
Part of #6904

### Description
- Fix inefficient patching in `PatchDataset`
- Add cache option in `GridPatchDataset`

### Types of changes


KumoLiu:
Hi @wyli @ericspod, could you please help review the PR, I haven't updated the docstring and unit tests yet.
In this PR, I mainly added:
- add cache support in `GridPatchDataset`, if set `cache=True`, then the patches from `data * cache_num` will be cached
- Update `PatchDataset` to inherit from `IterableDataset` to avoid duplicated patching
- `GridPatchDataset` and `PatchDataset` can also accept file names instead of only accepting already loaded data.

monai/data/grid_dataset.py
```diff
@@ -193,25 +210,160 @@ def __init__(
...
+                            out_patch = apply_transform(self.patch_transform, patch, map_items=False)
+                        if (
+                            self.with_coordinates and len(others) > 0
+                        ):  # patch_iter to yield at least 2 items: patch, coords
+                            yield out_patch, others[0]
+                        else:
+                            yield out_patch
+
+                if self._cache is None:
+                    raise RuntimeError("cache buffer is not initialized, please call `set_data()` first.")
```
ericspod, line 332:
```suggestion
                    raise RuntimeError("Cache buffer is not initialized, call `set_data()` before epoch begins.")
```

Does it need to be called before or just after an epoch? It should be stated here when.

monai/data/grid_dataset.py
```diff
@@ -193,25 +210,160 @@ def __init__(
...
+                data = self._cache[cache_index]  # type: ignore
+                other = self._cache_other[cache_index]  # type: ignore
+
+                # load data from cache and execute from the first random transform
+                if not isinstance(self.patch_transform, Compose):
```
ericspod, line 337: This check should be in the constructor?

monai/data/grid_dataset.py
```diff
@@ -193,25 +210,160 @@ def __init__(
...
+                    raise ValueError("transform must be an instance of monai.transforms.Compose.")
+
+                first_random = self.patch_transform.get_index_of_first(
+                    lambda t: isinstance(t, RandomizableTrait) or not isinstance(t, Transform)
+                )
+                if first_random is not None:
+                    data = deepcopy(data) if self.copy_cache is True else data
```
ericspod, line 344:
```suggestion
                    data = deepcopy(data) if self.copy_cache else data
```

If `copy_cache` is only ever True or False and not None you don't need to check identity. We've done that pattern elsewhere when `None` means do some default thing.

monai/data/grid_dataset.py
```diff
@@ -193,25 +210,160 @@ def __init__(
...
+                if self._cache is None:
+                    raise RuntimeError("cache buffer is not initialized, please call `set_data()` first.")
+                data = self._cache[cache_index]  # type: ignore
+                other = self._cache_other[cache_index]  # type: ignore
+
+                # load data from cache and execute from the first random transform
+                if not isinstance(self.patch_transform, Compose):
+                    raise ValueError("transform must be an instance of monai.transforms.Compose.")
+
+                first_random = self.patch_transform.get_index_of_first(
```
ericspod, line 340: Could this index be computed in the constructor instead? It's used in multiple places.

monai/data/grid_dataset.py
```diff
@@ -193,25 +210,160 @@ def __init__(
...
+        else:
+            for image in super().__iter__():
+                for patch, *others in self.patch_iter(image):
+                    out_patch = patch
+                    if self.patch_transform is not None:
+                        out_patch = apply_transform(self.patch_transform, patch, map_items=False)
+                    if self.with_coordinates and len(others) > 0:  # patch_iter to yield at least 2 items: patch, coords
+                        yield out_patch, others[0]
+                    else:
+                        yield out_patch
```
ericspod, line 363: This loop appears multiple times in essentially the same form. We could define the following private method to be used here and wherever else we're applying transforms to generate patches:

```python
def _generate_patches(self, src, **apply_args):
    for patch, *others in src:
        out_patch = patch
        if self.patch_transform is not None:
            out_patch = apply_transform(self.patch_transform, patch, **apply_args)
        if self.with_coordinates and len(others) > 0:  # patch_iter to yield at least 2 items: patch, coords
            yield out_patch, others[0]
        else:
            yield out_patch
```

I think this captures the commonality in all the places where we're yield patches after transforms are applied. You'd have to double check my logic but I think we can reduce repeating the same pattern and make it something possibly to use later in subclasses.

monai/data/grid_dataset.py
KumoLiu: Yes. I just followed the CacheDataset to add this check here, perhaps we can remove this check since `self.patch_transform` has been hard-coded to be Compose in the constructor. What do you think?
https://github.com/Project-MONAI/MONAI/blob/cf40cbc833aa076706d9ecacae3eb4e20736f66a/monai/data/grid_dataset.py#L224

ericspod: It can go then yes if we're sure the value of `patch_transform` is expected to change ever.

monai/data/grid_dataset.py
```diff
@@ -11,18 +11,35 @@
...
+    from tqdm import tqdm
+
+    has_tqdm = True
+else:
+    tqdm, has_tqdm = optional_import("tqdm", "4.47.0", min_version, "tqdm")
+
+cp, _ = optional_import("cupy")
+lmdb, _ = optional_import("lmdb")
+pd, _ = optional_import("pandas")
+kvikio_numpy, _ = optional_import("kvikio.numpy")
```
ericspod, line 42: This doesn't appear used?

ericspod: Actually none of these optional imports appear used.

monai/data/grid_dataset.py
```diff
@@ -193,25 +210,141 @@ def __init__(
...
+                    yield from self._generate_patches(self.patch_iter(image))
+
+                if self._cache is None:
+                    raise RuntimeError("Cache buffer is not initialized, please call `set_data()` before epoch begins.")
+                data = self._cache[cache_index]  # type: ignore
+                other = self._cache_other[cache_index]  # type: ignore
+
+                # load data from cache and execute from the first random transform
+                data = deepcopy(data) if self.copy_cache else data
+                yield from self._generate_patches(zip(data, other), start=self.first_random)
```
ericspod, line 341:
```suggestion
                else:
                    if self._cache is None:
                        raise RuntimeError("Cache buffer is not initialized, please call `set_data()` before epoch begins.")
                    data = self._cache[cache_index]  # type: ignore
                    other = self._cache_other[cache_index]  # type: ignore

                    # load data from cache and execute from the first random transform
                    data = deepcopy(data) if self.copy_cache else data
                    yield from self._generate_patches(zip(data, other), start=self.first_random)
```

I think you don't want to execute the rest of the code in this block so need an `else`?

ericspod (commented):
Looks better! I can review again when the PR is ready.

monai/data/grid_dataset.py
```diff
@@ -193,27 +224,153 @@ def __init__(
...
+            self.cache_num = min(int(self.set_num), int(data_len * self.set_rate), data_len)
+
+        # only compute cache for the unique items of dataset, and record the last index for duplicated items
+        mapping = {self.hash_func(v): i for i, v in enumerate(self.data)}
+        _compute_cache_num(len(mapping))
+        self._hash_keys = list(mapping)[: self.cache_num]
+        indices = list(mapping.values())[: self.cache_num]
+
+        self._cache = self._fill_cache(indices)
+        return
```
ericspod, line 285:
```suggestion
        # only compute cache for the unique items of dataset, and record the last index for duplicated items
        mapping = {self.hash_func(v): i for i, v in enumerate(self.data)}
        self.cache_num = min(int(self.set_num), int(len(mapping)* self.set_rate), len(mapping))
        self._hash_keys = list(mapping)[: self.cache_num]
        indices = list(mapping.values())[: self.cache_num]
        self._cache = self._fill_cache(indices)
```

monai/data/grid_dataset.py
```diff
@@ -193,27 +224,153 @@ def __init__(
...
+        for patch, *others in self.patch_iter(item):
+            if self.first_random is not None:
+                patch = self.patch_transform(patch, end=self.first_random, threading=True)  # type: ignore
+
+            if self.as_contiguous:
+                patch = convert_to_contiguous(patch, memory_format=torch.contiguous_format)
+            if self.with_coordinates and len(others) > 0:  # patch_iter to yield at least 2 items: patch, coords
+                other_cache.append(others[0])
+            patch_cache.append(patch)
+        self._cache_other.append(other_cache)
```
ericspod, line 323: This isn't thread-safe since `_cache_other` is just a `list` object and so there's no locking. There's also no guarantee of ordering if the threads in the `ThreadPool` operate at different timings than what you've tested with. I think you need to be accumulating a `cache_other` list at the same time you're accumulating `patch_cache` and also return it, then in `set_data` have something like `self._cache, self._cache_other = zip(*self._fill_cache(indices))` to produce the two lists that you need.

monai/data/grid_dataset.py
```diff
@@ -193,27 +224,153 @@ def __init__(
...
+        if self.cache_num <= 0:
+            return []
+        if indices is None:
+            indices = list(range(self.cache_num))
+        if self.progress and not has_tqdm:
+            warnings.warn("tqdm is not installed, will not show the caching progress bar.")
+        with ThreadPool(self.num_workers) as p:
+            if self.progress and has_tqdm:
+                return list(tqdm(p.imap(self._load_cache_item, indices), total=len(indices), desc="Loading dataset"))
+            return list(p.imap(self._load_cache_item, indices))
```
ericspod, line 305:
```suggestion
        if self.progress and not has_tqdm:
            warnings.warn("tqdm is not installed, will not show the caching progress bar.")
        
        pfunc = tqdm if self.progress and has_tqdm else (lambda v, **_: v)
        with ThreadPool(self.num_workers) as p:
            return list(pfunc(p.imap(self._load_cache_item, indices), total=len(indices), desc="Loading dataset"))
```

Very minor reduction in code.

ericspod:
Hi @KumoLiu I have a few suggestions that aren't critical but more importantly there's a thread-safety issue we need to fix, otherwise I think we're good. Thanks!

monai/data/grid_dataset.py
KumoLiu: Thanks for pointing it out, addressed it in the latest commit, could you please help review it again?

ericspod (approved):
Looks good to me now, thanks!

PR #7181, opened 2023-11-01
6627 reading dcm folder filtering filenames

### LINKED ISSUE #6627: Filter only ".dcm" files when reading DICOM series
When I try to load a DICOM folder, the loading will fail if there is some irrelevant file in the folder (or I have to set `force=True`). E.g., when I load the DICOM series from the [LIDC-IDRI](https://wiki.cancerimagingarchive.net/pages/viewpage.action?pageId=1966254) database, there's a `.xml` annotation file in the folder, the loader will try to load it as well and cause failure.

Change this line 
https://github.com/Project-MONAI/MONAI/blob/2cbed6cfa7a007fa8853a7bd8cf09303172686c9/monai/data/image_reader.py#L469
to something like
```python
series_slcs = Path(name).glob("*.dcm")
```
I'm not sure if there are other suffixes to be considered.

wyli:
Fixes #6627

### Description
adding a `fname_regex` option to the pydicom reader

### Types of changes

### DIFF
--- a/monai/data/image_reader.py
+++ b/monai/data/image_reader.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 import glob
 import os
+import re
 import warnings
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable, Iterator, Sequence
@@ -403,8 +404,12 @@ class PydicomReader(ImageReader):
         label_dict: label of the dicom data. If provided, it will be used when loading segmentation data.
             Keys of the dict are the classes, and values are the corresponding class number. For example:
             for TCIA collection "C4KC-KiTS", it can be: {"Kidney": 0, "Renal Tumor": 1}.
+        fname_regex: a regular expression to match the file names when the input is a folder.
+            If provided, only the matched files will be included.
+            For example, if the file name is "image_0001.dcm", and the regular expression is `".*image_(\\d+).dcm"`.
+            Default to `"^(?!.*LICENSE).*"`, ignoring any file name containing `"LICENSE"`.
         kwargs: additional args for `pydicom.dcmread` API. more details about available args:
-            https://pydicom.github.io/pydicom/stable/reference/generated/pydicom.filereader.dcmread.html#pydicom.filereader.dcmread
+            https://pydicom.github.io/pydicom/stable/reference/generated/pydicom.filereader.dcmread.html
             If the `get_data` function will be called
             (for example, when using this reader with `monai.transforms.LoadImage`), please ensure that the argument
             `stop_before_pixels` is `True`, and `specific_tags` covers all necessary tags, such as `PixelSpacing`,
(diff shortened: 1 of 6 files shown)


monai/data/image_reader.py
```diff
@@ -418,6 +423,7 @@ def __init__(
         swap_ij: bool = True,
         prune_metadata: bool = True,
         label_dict: dict | None = None,
+        fname_regex: str = r"^(?!.*LICENSE).*",
```
ericspod, line 426: Is this excluding enough? What about files starting with "." which are meant to be hidden, anything with README in the title, DICOMDIR files, or other known file types? Was there a specific reason why LICENSE was important to skip? 

I would also put this default value into a constant at the top of this file to make it easier to spot and reuse elsewhere.

ericspod:
Should we test specifically for passing a different pattern, or a pattern which matches nothing?

monai/data/image_reader.py
wyli: good question, this is mainly to be consistent with the existing logic https://github.com/Project-MONAI/MONAI/blob/3239db8152cad7e3997a0e09147863c9b34be14b/monai/data/image_reader.py#L470-L471 @yiheng-wang-nv are these from the tcia dataset?

KumoLiu: Perhaps we could directly set the rules for inclusion, something like r".*.dcm|.*.DCM"?

wyli: Our test cases have no extension names.. https://github.com/Project-MONAI/MONAI/tree/dev/tests/testing_data/CT_DICOM

yiheng-wang-nv: >@yiheng-wang-nv are these from the tcia dataset?
Hi @wyli , yes, for TCIA dataset, a LICENSE file will be included in the folder, thus I wrote this logic.

wyli: Ok thanks, i'll move the filter to monai.apps.datasets.

ericspod (approved):
Looks good here. We've encountered the issue here of exceptions being raised when trying to read a non-DICOM file such as the LICENSE file. The approach I've used in the past with pydicom is to try to read everything and skip over any file raising InvalidDicomError, this has unintented consequences of course. What we can do instead is use [is_dicom](https://github.com/pydicom/pydicom/blob/335934d5afdd655d9b9c73f8470fb6da3cec26c5/src/pydicom/misc.py#L40) from the library to check any file first if this doesn't have too much of a performance penalty.

wyli:
> Looks good here. We've encountered the issue here of exceptions being raised when trying to read a non-DICOM file such as the LICENSE file. The approach I've used in the past with pydicom is to try to read everything and skip over any file raising InvalidDicomError, this has unintented consequences of course. What we can do instead is use [is_dicom](https://github.com/pydicom/pydicom/blob/335934d5afdd655d9b9c73f8470fb6da3cec26c5/src/pydicom/misc.py#L40) from the library to check any file first if this doesn't have too much of a performance penalty.

thanks! I'm adding some exception handling for InvalidDicomError, and user can set `fname_regex=None` to use the is_dicom checker.

PR #7198, opened 2023-11-03
2872 implementation of mixup, cutmix and cutout

### LINKED ISSUE #2872: Support Mixup and Cutmix transformation
HI, the more strong transformation  "RandCoarseDropout" has been released in the latest version.
It's possible to support Mixup or Cutmix in MonAI? 
image

juampatronics:
Fixes #2872 

### Description

Implementation of mixup, cutmix and cutout as described in the original papers.
Current implementation support both, the dictionary-based batches and tuples of tensors.

### Types of changes


juampatronics:
I do not fully understand the reason for the error:

 def __call__(self, data: torch.Tensor, labels: torch.Tensor | None = None):
TypeError: unsupported operand type(s) for |: 'torch._C._TensorMeta' and 'NoneType'

Is it because of Python versions < 3.10?. What confuses me is that that construct it is actually used in MONAIs code a lot already.

wyli (commented):
Thanks for the PR! These are useful data preprocessing/augmentation tools, how about moving these into the monai.transforms module? Please also use the randomizable API to introduce random factor into the transforms https://github.com/Project-MONAI/MONAI/blob/fbbb64b4cb2facffccbf4fdf6c44e20f8d31ca10/monai/transforms/transform.py#L174

juampatronics:
Happy to contribute again @wyli. I have been enjoying MONAI for quite some time already.

Sure, I will move it to the transforms modules.  Should I keep it as a submodule called "regularization"?
My motivation was twofold: on the one side they expect a batch, as opposed to all other sort of transforms and therefore not "composable" with them, and their use is purely for regularization during training.

wyli:
> Happy to contribute again @wyli. I have been enjoying MONAI for quite some time already.
> 
> Sure, I will move it to the transforms modules. Should I keep it as a submodule called "regularization"? My motivation was twofold: on the one side they expect a batch, as opposed to all other sort of transforms and therefore not "composable" with them, and their use is purely for regularization during training.

Cool, thanks @juampatronics, the plan looks good to me.

juampatronics:
Done

wyli (approved):
thanks, it looks good to me, please fix the format errors, and update the docstring about the input shapes.

monai/transforms/regularization/dictionary.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __call__(self, data):
+        self.mixup.randomize()
+        result = dict(data)
+        for k in self.keys:
+            result[k] = self.mixup.apply(data[k])
+        return result
+
+
+MixUpD = MixUpDict = MixUpd
```
KumoLiu, line 51: This line and Line 87 can be moved to the end of the file.

KumoLiu (approved):
Thanks for the pr.
LGTM, It would be great if you could include a little usage example in the docstring.

juampatronics:
> thanks, it looks good to me, please fix the format errors, and update the docstring about the input shapes.

Regretfully I did not manage to run  ./runtests.sh --autofix under WSL. Running isort under Windows gives no error.

ericspod:
> Regretfully I did not manage to run ./runtests.sh --autofix under WSL. Running isort under Windows gives no error.

This might be related to versions of isort, please double check you have the most recent version. I wouldn't think WSL would be the issue here, which is a full Linux system really, but more likely what libraries you're using.

monai/transforms/regularization/array.py
```diff
@@ -0,0 +1,141 @@
...
+# limitations under the License.
+
+from __future__ import annotations
+
+from abc import abstractmethod
+from math import ceil, sqrt
+
+import torch
+
+from monai.transforms import Randomizable, Transform
```
KumoLiu, line 19:
```suggestion
from monai.transforms.transform import Randomizable, Transform
```

PR #7205, opened 2023-11-06
Expose EnesembleRunner in AutoRunner

### LINKED ISSUE #7204: Ensemble Outputs for AutoRunner
Currently, after AutoRunner finished the job, the information about the ensemble methods and metrics are only printed in the log and the user doesn't have any access to the results of the ensemble method. The AutoRunner needs to provide this information to user so she can extract the ensembled models for further inference or other utilizations.

drbeh:
Fixes #7204 

### Description

This PR makes `EnesembleRunner` instance be accessible via `AutoRunner` instance, so the use can get information about the ensembling after the models are trained and ensembled.

### Types of changes

### DIFF
--- a/monai/apps/auto3dseg/auto_runner.py
+++ b/monai/apps/auto3dseg/auto_runner.py
@@ -863,7 +863,7 @@ class AutoRunner:
 
         # step 4: model ensemble and write the prediction to disks.
         if self.ensemble:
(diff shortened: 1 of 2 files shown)


drbeh:
This only covers single GPU, we'll check if we can do the same when using multiple gpu (which is command line based).

PR #7217, opened 2023-11-09
fix useless error msg in nnunetv2runner

elitap:
Fixes # fix useless error msg in nnunetv2runner

### Types of changes


ericspod (approved):
Thanks for spotting this!

PR #7249, opened 2023-11-21
add class label option to write metric report to improve readability …

elitap:
add class label option to write metric report to improve readability, without that option in case of many classes the resulting report is very hard to interpret.

### Types of changes

### DIFF
--- a/monai/handlers/utils.py
+++ b/monai/handlers/utils.py
@@ -124,8 +124,8 @@ def write_metrics_reports(
             if class_labels is None:
                 class_labels = ["class" + str(i) for i in range(v.shape[1])]
             else:
-                class_labels=[str(i) for i in class_labels]  # ensure to have a list of str 
-                
+                class_labels = [str(i) for i in class_labels]  # ensure to have a list of str
+
             class_labels += ["mean"]
             v = np.concatenate([v, np.nanmean(v, axis=1, keepdims=True)], axis=1)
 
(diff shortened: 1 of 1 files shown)


KumoLiu (approved):
LGTM, could you please address the mypy error?

monai/handlers/utils.py
```diff
@@ -59,6 +59,7 @@ def write_metrics_reports(
     metrics: dict[str, torch.Tensor | np.ndarray] | None,
     metric_details: dict[str, torch.Tensor | np.ndarray] | None,
     summary_ops: str | Sequence[str] | None,
+    class_labels: list[str] | None = None,
```
ericspod, line 62: This will  have to be made the last argument in the constructor otherwise any invocation using only positional arguments will break.

monai/handlers/utils.py
```diff
@@ -118,7 +121,9 @@ class    mean    median    max    5percentile 95percentile  notnans
                 v = v.reshape((-1, 1))
 
             # add the average value of all classes to v
-            class_labels = ["class" + str(i) for i in range(v.shape[1])] + ["mean"]
+            if class_labels is None:
+                class_labels = ["class" + str(i) for i in range(v.shape[1])]
```
ericspod, line 125:
```suggestion
            if class_labels is None:
                class_labels = ["class" + str(i) for i in range(v.shape[1])]
            else:
                class_labels=[str(i) for i in class_labels]  # ensure we have a list of string values
```

monai/handlers/utils.py
elitap: Thanks for the suggestion, totally agree :-)

monai/handlers/utils.py
elitap: agree as well :-)

Issue #7254, opened 2023-11-23
Invalid shape sample from RandCropByLabelClasses


ltetrel:
**Describe the bug**
I was running `Auto3dseg` and the train transform failed when creating the batch (collate the samples), because few did not match the patch size I defined (one had `(96, 95, 96)` when others had `(96, 96, 96)`).

It seems that the culprit was actually `RandCropByLabelClasses`. When generating its [random sample centers](https://github.com/Project-MONAI/MONAI/blob/8e134b8cb92e3c624b23d4d10c5d4596bb5b9d9b/monai/transforms/croppad/array.py#L1347C24-L1347C59), it will [correct them](https://github.com/Project-MONAI/MONAI/blob/8e134b8cb92e3c624b23d4d10c5d4596bb5b9d9b/monai/transforms/utils.py#L640) if they do not fall within the boundary of the label. However, if the image and label have different shape, the center are not correctly corrected because they use the size of the label, when it should use the size of the image (which we are actually cropping). As a result, the resulting sample are not correctly cropped and don't have the shape defined by [`spatial_size` parameter](https://github.com/Project-MONAI/MONAI/blob/8e134b8cb92e3c624b23d4d10c5d4596bb5b9d9b/monai/transforms/croppad/array.py#L1269).

Having different shapes can happen because of truncating error when using a nearest neighbor (for label) VS bilinear (for image) resampler.

I suppose the same issue could happen with `RandCropByPosNegLabel` ?

**To Reproduce**
Here is a code snippet to reproduce the issue

```
import numpy as np

from monai.config import print_config
from monai.utils import set_determinism
from monai.transforms import RandCropByLabelClasses
from monai.data.synthetic import create_test_image_3d


def main():
    seed = 9
    set_determinism(seed=seed)
    roi_size = (24, 24, 24)

    test_img, test_label = create_test_image_3d(height=100, width=100, depth=100, channel_dim=0)
    test_img = test_img[..., :-1, :-1, :]
    rand_cropper_by_label = RandCropByLabelClasses(num_classes=5, spatial_size=roi_size, warn=False, lazy=False)
    rand_cropper_by_label.set_random_state(seed=seed)
    rand_cropped_img = rand_cropper_by_label(img=test_img, label=test_label)

    np.testing.assert_equal(rand_cropped_img[0].shape[1:], roi_size)
    print(f"{rand_cropped_img[0].shape}")


if __name__ == '__main__':
    main()

```

**Expected behavior**
[...]

PR #7255, opened 2023-11-23
7254 fix crop center correction

ltetrel:
Fixes [#7254](https://github.com/Project-MONAI/MONAI/issues/7254).

### Description

### Types of changes

### DIFF
--- a/monai/transforms/croppad/array.py
+++ b/monai/transforms/croppad/array.py
@@ -1329,6 +1329,7 @@ class RandCropByLabelClasses(Randomizable, TraceableTransform, LazyTransform, Mu
         label: torch.Tensor | None = None,
         indices: list[NdarrayOrTensor] | None = None,
         image: torch.Tensor | None = None,
+        img: torch.Tensor | None = None,
     ) -> None:
         indices_ = self.indices if indices is None else indices
         if indices_ is None:
@@ -1344,8 +1345,9 @@ class RandCropByLabelClasses(Randomizable, TraceableTransform, LazyTransform, Mu
             _shape = image.peek_pending_shape() if isinstance(image, MetaTensor) else image.shape[1:]
         if _shape is None:
             raise ValueError("label or image must be provided to infer the output spatial shape.")
+        img_shape = img.peek_pending_shape() if isinstance(image, MetaTensor) else img.shape[1:]
         self.centers = generate_label_classes_crop_centers(
-            self.spatial_size, self.num_samples, _shape, indices_, self.ratios, self.R, self.allow_smaller, self.warn
+            self.spatial_size, self.num_samples, _shape, indices_, img_shape, self.ratios, self.R, self.allow_smaller, self.warn
         )
 
     @LazyTransform.lazy.setter  # type: ignore
@@ -1381,7 +1383,7 @@ class RandCropByLabelClasses(Randomizable, TraceableTransform, LazyTransform, Mu
         if randomize:
             if label is None:
                 label = self.label
-            self.randomize(label, indices, image)
+            self.randomize(label, indices, image, img)
         results: list[torch.Tensor] = []
         if self.centers is not None:
             img_shape = img.peek_pending_shape() if isinstance(img, MetaTensor) else img.shape[1:]
--- a/monai/transforms/croppad/array.py
+++ b/monai/transforms/croppad/array.py
@@ -1347,7 +1347,15 @@ class RandCropByLabelClasses(Randomizable, TraceableTransform, LazyTransform, Mu
             raise ValueError("label or image must be provided to infer the output spatial shape.")
         img_shape = img.peek_pending_shape() if isinstance(image, MetaTensor) else img.shape[1:]
         self.centers = generate_label_classes_crop_centers(
-            self.spatial_size, self.num_samples, _shape, indices_, img_shape, self.ratios, self.R, self.allow_smaller, self.warn
+            self.spatial_size,
+            self.num_samples,
+            _shape,
+            indices_,
+            img_shape,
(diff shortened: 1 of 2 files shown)


monai/transforms/croppad/array.py
```diff
@@ -1344,8 +1345,17 @@ def randomize(
             _shape = image.peek_pending_shape() if isinstance(image, MetaTensor) else image.shape[1:]
         if _shape is None:
             raise ValueError("label or image must be provided to infer the output spatial shape.")
+        img_shape = img.peek_pending_shape() if isinstance(image, MetaTensor) else img.shape[1:]
```
ericspod, line 1348:
```suggestion
        img_shape = img.peek_pending_shape() if isinstance(img, MetaTensor) else img.shape[1:]
```

I think this is what you meant, you may have to raise an exception if `img_shape` is None as well.

ericspod:
The assumption with this class I believe was that the image and label tensors have to be the same size in the spatial dimensions anyway. Is your use case different from that?

monai/transforms/croppad/array.py
ltetrel: Thanks,
I added a condition but if I raise an exception I am afraid it could break existing functionnalities, that is why I added this warning [here](https://github.com/Project-MONAI/MONAI/pull/7255/files/932e495d8d9103cd5bbeef67c30439e1a14e3249#diff-96cd15307195a8d855e3fb33c35abaa18afaecc182874751af4eb3144726d158R621), wdyt ?

ericspod: You can leave it in which case `generate_label_classes_crop_centers` will issue a warning every time when doing it's own check. That may be fine though I'm not sure under what conditions `img_shape` would be None and result in an exception further down stream and when it would just be a warning.

ltetrel:
`test_shape_2` fails
```
FAIL: test_shape_2 (tests.test_multi_scale.TestMultiScale)
  File "/home/vision/.venvs/moosex/lib/python3.9/site-packages/parameterized/parameterized.py", line 620, in standalone_func
  File "/home/vision/.venvs/moosex/src/MONAI/tests/test_multi_scale.py", line 58, in test_shape
    np.testing.assert_allclose(result.detach().cpu().numpy(), expected_val, rtol=1e-5)
  File "/home/vision/.venvs/moosex/lib/python3.9/site-packages/numpy/testing/_private/utils.py", line 1530, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/home/vision/.venvs/moosex/lib/python3.9/site-packages/numpy/testing/_private/utils.py", line 844, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=1e-05, atol=0

Mismatched elements: 1 / 1 (100%)
Max absolute difference: 1.62509232e-05
Max relative difference: 2.27213185e-05
 x: array(0.715212, dtype=float32)
 y: array(0.715228)

Ran 14100 tests in 684.794s

FAILED (failures=1, skipped=750)
```

monai/transforms/croppad/array.py
```diff
@@ -1344,8 +1345,19 @@ def randomize(
             _shape = image.peek_pending_shape() if isinstance(image, MetaTensor) else image.shape[1:]
         if _shape is None:
             raise ValueError("label or image must be provided to infer the output spatial shape.")
+        img_shape = None
```
KumoLiu, line 1348: Thanks for the PR!
But I think we could have a check for `img_shape` and `label_shape` if they are not equal, then we may raise a warning in the call.

monai/transforms/croppad/dictionary.py
```diff
@@ -1150,7 +1153,7 @@ def requires_current_data(self):
 
     def __call__(self, data: Mapping[Hashable, Any], lazy: bool | None = None) -> list[dict[Hashable, torch.Tensor]]:
         d = dict(data)
-        self.randomize(d.get(self.label_key), d.pop(self.indices_key, None), d.get(self.image_key))  # type: ignore
+        self.randomize(d.get(self.label_key), d.pop(self.indices_key, None), d.get(self.image_key), d.get(self.keys[0]))  # type: ignore
```
ericspod, line 1156: What is `d.get(self.keys[0])` meant to be here? The value of `self.keys[0]` is going to vary depending on what the user has provided when constructing the transform.

monai/transforms/utils.py
```diff
@@ -613,6 +615,11 @@ def generate_label_classes_crop_centers(
     if rand_state is None:
         rand_state = np.random.random.__self__  # type: ignore
 
+    if img_spatial_shape is None:
+        img_spatial_shape = label_spatial_shape
+        if warn:
+            warnings.warn("img_spatial_shape not defined, expect invalid shape for samples.")
```
ericspod, line 621:
```suggestion
            warnings.warn("'img_spatial_shape' not defined,  samples may have an invalid shape if image and label shapes differ.")
```

monai/transforms/croppad/array.py
ericspod: There should be a warning about this yes.

monai/transforms/utils.py
```diff
@@ -587,6 +587,7 @@ def generate_label_classes_crop_centers(
     num_samples: int,
     label_spatial_shape: Sequence[int],
     indices: Sequence[NdarrayOrTensor],
+    img_spatial_shape: Sequence[int] | None = None,
```
KumoLiu, line 590: I still don't think we should add `img_spatial_shape` here. We should assume image and label have the same shape, so a warning should be enough.

PR #7318, opened 2023-12-13
Support specified filenames in `Saveimage`

### LINKED ISSUE #7317: Support specified filenames in `Saveimage`
Now `SaveImage` only supports specified the `output_name_formatter` which makes it a little complicated for users.
https://github.com/Project-MONAI/MONAI/blob/b1c8b42d03bf3c1daabd6c39a35fcbdadb57e8b0/monai/transforms/io/array.py#L393C14-L393C14

Add support specified `filename` for users to save like `nibabel.save`.

_Originally posted by @heyufan1995_

KumoLiu:
Fixes #7317 

### Description

Add support specified filename for users to save like nibabel.save.

### Types of changes

### DIFF
--- a/monai/transforms/io/array.py
+++ b/monai/transforms/io/array.py
@@ -458,14 +458,17 @@ class SaveImage(Transform):
             self.write_kwargs.update(write_kwargs)
         return self
 
-    def __call__(self, img: torch.Tensor | np.ndarray, meta_data: dict | None = None, filename: str | None = None):
+    def __call__(
+        self, img: torch.Tensor | np.ndarray, meta_data: dict | None = None, filename: str | PathLike | None = None
+    ):
         """
         Args:
             img: target data content that save into file. The image should be channel-first, shape: `[C,H,W,[D]]`.
             meta_data: key-value pairs of metadata corresponding to the data.
+            filename: str or file-like object which to save img.
+                If specified, will ignore `self.output_name_formatter` and `self.folder_layout`.
         """
         meta_data = img.meta if isinstance(img, MetaTensor) else meta_data
-        kw = self.fname_formatter(meta_data, self)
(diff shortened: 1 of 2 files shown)


monai/transforms/io/array.py
```diff
@@ -458,15 +458,27 @@ def set_options(self, init_kwargs=None, data_kwargs=None, meta_kwargs=None, writ
...
         """
         meta_data = img.meta if isinstance(img, MetaTensor) else meta_data
-        kw = self.fname_formatter(meta_data, self)
-        filename = self.folder_layout.filename(**kw)
+        if filename is not None:
+            filename = str(filename) + (
+                f".{self.output_ext}"
+                if self.output_ext and not self.output_ext.startswith(".")
+                else f"{self.output_ext}"
+            )
```
ericspod, line 477:
```suggestion
            filename = f"{filename}{self.output_ext.split('.', 1)[-1]}"
```
I think this is equivalent.

ericspod (approved):
I had a minor change suggestion but otherwise looks good.

monai/transforms/io/array.py
KumoLiu: Thanks for the suggestion, may not work for "nii.gz"?

ericspod: Actually `filename = f"{filename}{('.'+self.output_ext.lstrip('.')) if self.output_ext else ''}"` would work for that case instead. Maybe `self.output_ext` should be cleaned up in the constructor to ensure it begins with "." if not empty?

KumoLiu: Sure, updated it in the latest commit! Could you please review it?

ericspod (approved):
LGTM

Issue #7227, opened 2023-11-13
Refactor additions from Generative module


marksgraham:
After merging MONAI Generative into core [issue](https://github.com/Project-MONAI/MONAI/issues/6676) some refactoring is needed to reduce repeat code, see discussion [here](https://github.com/Project-MONAI/MONAI/pull/7196#issuecomment-1808304928).

Until this is done any blocks from Generative that are likely to be removed will be marked as private.

Will update the list of items needing refactoring here as I go along:

- Upsample/Downsample/Resblock/AttentionBlock in the autoencoderkl network (see comment [here](https://github.com/Project-MONAI/MONAI/pull/7196#discussion_r1400623383) for replacing upsample/downsample with convs)
- Upsample/Downsample/BasicTransformerBlock/ResnetBlock/AttentionBlock in the diffusion_model_unet
- SABlock and TransformerBlock used by DecoderOnlyTransformer
- ControlNet refactor suggested by @ericspod [here](https://github.com/Project-MONAI/MONAI/pull/7312#discussion_r1424243496)
- Neater init on the patchgan discriminator suggested by @ericspod [here](https://github.com/Project-MONAI/MONAI/pull/7319#discussion_r1425531107)
- SpadeNorm block - use [get_norm_layer](https://github.com/Project-MONAI/MONAI/pull/7320#discussion_r1427335512) and [here](https://github.com/Project-MONAI/MONAI/pull/7320#discussion_r1427342468)
- SPADEAutoencoder - merge with the autoencoder KL as the only difference is in the decoder.
- Schedulers - refactor some calculations into the base class as suggested by @KumoLiu [here](https://github.com/Project-MONAI/MONAI/pull/7332#discussion_r1437355600)

vikashg:
Closing @marksgraham as it seems to be completed and merged.

marksgraham:
Hi @vikashg afraid this is still pending so re-opening!

KumoLiu:
Add here as a reminder.
https://github.com/Project-MONAI/MONAI/pull/7346#discussion_r1444749763
https://github.com/Project-MONAI/MONAI/pull/7346#discussion_r1444207936

vgrau98:
I can start working on `SABlock`, integrating the private block and making refacto if needed if this is ok for you @marksgraham

marksgraham:
Hi @vgrau98 yes that would be great! 
[...]

PR #7320, opened 2023-12-14
6676 port generative networks spade

marksgraham:
Towards #6676  .

### Description

This adds SPADE-enabled autoencoder and diffusion_model_unet architectures. They are new implementations for each network, rather than options in the existing network, because @virginiafdez and I felt that adding additional options to the existing networks to enable spade compatibility significantly reduced the readability of them for users who were not interested in SPADE functionality.

These are the last networks to be ported over.

### Types of changes

### DIFF
--- a/docs/source/networks.rst
+++ b/docs/source/networks.rst
@@ -588,6 +588,11 @@ Nets
 .. autoclass:: DiffusionModelUNet
   :members:
 
+`ControlNet`
+~~~~~~~~~~~~
+.. autoclass:: ControlNet
+  :members:
+
 `RegUNet`
 ~~~~~~~~~
 .. autoclass:: RegUNet
--- a/docs/source/networks.rst
+++ b/docs/source/networks.rst
@@ -248,6 +248,7 @@ Blocks
 .. autoclass:: monai.apps.reconstruction.networks.blocks.varnetblock.VarNetBlock
    :members:
 
+
 N-Dim Fourier Transform
 ~~~~~~~~~~~~~~~~~~~~~~~~
 .. automodule:: monai.networks.blocks.fft_utils_t
@@ -258,6 +259,10 @@ N-Dim Fourier Transform
 .. autofunction:: monai.networks.blocks.fft_utils_t.fftshift
 .. autofunction:: monai.networks.blocks.fft_utils_t.ifftshift
 
+`SPADE`
+~~~~~~~
+.. autoclass:: monai.networks.blocks.spade_norm.SPADE
+    :members:
 
 Layers
 ------
@@ -588,6 +593,11 @@ Nets
 .. autoclass:: DiffusionModelUNet
   :members:
 
+`SPADEDiffusionModelUNet`
+~~~~~~~~~~~~~~~~~~~~~~~~~
+.. autoclass:: SPADEDiffusionModelUNet
+  :members:
+
 `ControlNet`
 ~~~~~~~~~~~~
 .. autoclass:: ControlNet
@@ -616,6 +626,11 @@ Nets
 `AutoEncoderKL`
 ~~~~~~~~~~~~~~~
 .. autoclass:: AutoencoderKL
+  :members:
+
+`SPADEAutoencoderKL`
+~~~~~~~~~~~~~~~~~~~~
+.. autoclass:: SPADEAutoencoderKL
   :members:
 
 `VarAutoEncoder`
--- a/docs/source/networks.rst
+++ b/docs/source/networks.rst
@@ -248,7 +248,6 @@ Blocks
 .. autoclass:: monai.apps.reconstruction.networks.blocks.varnetblock.VarNetBlock
    :members:
 
-
 N-Dim Fourier Transform
 ~~~~~~~~~~~~~~~~~~~~~~~~
 .. automodule:: monai.networks.blocks.fft_utils_t
--- a/docs/source/networks.rst
+++ b/docs/source/networks.rst
@@ -770,6 +770,14 @@ Nets
 .. autoclass:: VQVAE
    :members:
 
+`PatchGANDiscriminator`
+~~~~~~~~~~~~~~~~~~~~~~~
+.. autoclass:: PatchDiscriminator
+   :members:
+
+.. autoclass:: MultiScalePatchDiscriminator
+   :members:
+
 Utilities
 ---------
 .. automodule:: monai.networks.utils
--- a/monai/networks/blocks/spade_norm.py
+++ b/monai/networks/blocks/spade_norm.py
@@ -23,12 +23,11 @@ class SPADE(nn.Module):
     Spatially Adaptive Normalization (SPADE) block, allowing for normalization of activations conditioned on a
     semantic map. This block is used in SPADE-based image-to-image translation models, as described in
     Semantic Image Synthesis with Spatially-Adaptive Normalization (https://arxiv.org/abs/1903.07291).
-    
+
     Args:
         label_nc: number of semantic labels
         norm_nc: number of output channels
         kernel_size: kernel size
-        padding: padding size for convolutions
         spatial_dims: number of spatial dimensions
         hidden_channels: number of channels in the intermediate gamma and beta layers
         norm: type of base normalisation used before applying the SPADE normalisation
@@ -40,7 +39,6 @@ class SPADE(nn.Module):
         label_nc: int,
         norm_nc: int,
         kernel_size: int = 3,
-        padding: int = 1,
         spatial_dims: int = 2,
         hidden_channels: int = 64,
         norm: str | tuple = "INSTANCE",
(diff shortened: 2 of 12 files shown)


monai/networks/blocks/spade_norm.py
```diff
@@ -0,0 +1,97 @@
...
+        norm: str | tuple = "INSTANCE",
+        norm_params: dict | None = None,
+    ) -> None:
+        super().__init__()
+
+        if norm_params is None:
+            norm_params = {}
+        if len(norm_params) != 0:
+            norm = (norm, norm_params)
+        self.param_free_norm = ADN(
```
ericspod, line 51: We could just use `get_norm_layer` instead of `ADN`.

monai/networks/blocks/spade_norm.py
```diff
@@ -0,0 +1,97 @@
...
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from monai.networks.blocks import ADN, Convolution
+
+
+class SPADE(nn.Module):
+    """
+    SPADE normalisation block based on the 2019 paper by Park et al. (doi: https://doi.org/10.48550/arXiv.1903.07291)
```
ericspod, line 23: Could we add a little more description about what the block is meant for and what it does?

monai/networks/blocks/spade_norm.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __init__(
+        self,
+        label_nc: int,
+        norm_nc: int,
+        kernel_size: int = 3,
+        spatial_dims: int = 2,
+        hidden_channels: int = 64,
+        norm: str | tuple = "INSTANCE",
+        norm_params: dict | None = None,
```
ericspod, line 43: As with the `norm` argument [here](https://github.com/marksgraham/MONAI/blob/035d2d2764c8db2836f5d2af4418a4d19fd954d6/monai/networks/blocks/convolutions.py#L107C10-L107C10) this separate parameters shouldn't be needed I don't think. `norm` is either just the name or a tuple containing the name and then whatever `norm_params` would have so I don't think the separate argument is needed.

monai/networks/blocks/spade_norm.py
```diff
@@ -0,0 +1,97 @@
...
+        self.param_free_norm = ADN(
+            act=None, dropout=0.0, norm=norm, norm_dim=spatial_dims, ordering="N", in_channels=norm_nc
+        )
+        self.mlp_shared = Convolution(
+            spatial_dims=spatial_dims,
+            in_channels=label_nc,
+            out_channels=hidden_channels,
+            kernel_size=kernel_size,
+            norm=None,
+            padding=kernel_size // 2,
```
ericspod, line 60: I don't think we need this argument since `same_padding` will be used if we provide nothing. This should give the same results for odd-sized kernels, for even ones the behaviour would be different but I don't think we'd do that?

ericspod (approved):
Looks good for merging into the branch. We can just keep track of the things I mentioned.

docs/source/networks.rst
```diff
@@ -248,6 +248,7 @@ Blocks
 .. autoclass:: monai.apps.reconstruction.networks.blocks.varnetblock.VarNetBlock
    :members:
 
+
```
KumoLiu, line 251: Add this line by accident?

monai/networks/blocks/spade_norm.py
```diff
@@ -0,0 +1,97 @@
...
+            kernel_size=kernel_size,
+            padding=kernel_size // 2,
+            act=None,
+        )
+
+    def forward(self, x: torch.Tensor, segmap: torch.Tensor) -> torch.Tensor:
+        """
+        Args:
+            x: input tensor
+            segmap: input segmentation map (bxcx[spatial-dimensions]) where c is the number of semantic channels.
```
KumoLiu, line 84:
```suggestion
            segmap: input segmentation map (B, C, [spatial-dimensions]) where C is the number of semantic channels.
```

monai/networks/blocks/spade_norm.py
```diff
@@ -0,0 +1,97 @@
...
+            act=None, dropout=0.0, norm=norm, norm_dim=spatial_dims, ordering="N", in_channels=norm_nc
+        )
+        self.mlp_shared = Convolution(
+            spatial_dims=spatial_dims,
+            in_channels=label_nc,
+            out_channels=hidden_channels,
+            kernel_size=kernel_size,
+            norm=None,
+            padding=kernel_size // 2,
+            act="LEAKYRELU",
```
KumoLiu, line 61: Is it hardcoded because of the original implementation? Maybe we should make it a default? What do you think?

monai/networks/blocks/spade_norm.py
```diff
@@ -0,0 +1,97 @@
...
+            out_channels=norm_nc,
+            kernel_size=kernel_size,
+            padding=kernel_size // 2,
+            act=None,
+        )
+
+    def forward(self, x: torch.Tensor, segmap: torch.Tensor) -> torch.Tensor:
+        """
+        Args:
+            x: input tensor
```
KumoLiu, line 83:
```suggestion
            x: input tensor with shape (B, C, [spatial-dimensions]) where C is the number of semantic channels.
```

monai/networks/nets/spade_autoencoderkl.py
```diff
@@ -0,0 +1,473 @@
...
+            channels=channels,
+            out_channels=latent_channels,
+            num_res_blocks=num_res_blocks,
+            norm_num_groups=norm_num_groups,
+            norm_eps=norm_eps,
+            attention_levels=attention_levels,
+            with_nonlocal_attn=with_encoder_nonlocal_attn,
+            use_flash_attention=use_flash_attention,
+        )
+        self.decoder = SPADEDecoder(
```
KumoLiu, line 356: If I understand correctly, the only difference is this decoder module. Or if I misunderstand, feel free to correct.
I personally feel we could combine this network and `AutoencoderKL` into one, looks there are many duplicated code. I think we can simply add a check here, my feeling is that dividing them into two networks will require users to distinguish between them on their own, which is not straightforward.

cc @ericspod, @Nic-Ma for more discussion here.

monai/networks/nets/spade_diffusion_model_unet.py
```diff
@@ -0,0 +1,909 @@
...
+            spade_intermediate_channels=spade_intermediate_channels,
+        )
+
+
+class SPADEDiffusionModelUNet(nn.Module):
+    """
+    Unet network with timestep embedding and attention mechanisms for conditioning based on
+    Rombach et al. "High-Resolution Image Synthesis with Latent Diffusion Models" https://arxiv.org/abs/2112.10752
+    and Pinaya et al. "Brain Imaging Generation with Latent Diffusion Models" https://arxiv.org/abs/2209.07162
+    Enables SPADE normalisation for semantic conditioning (Park et. al (2019): https://github.com/NVlabs/SPADE)
```
KumoLiu, line 614: Do we have some corresponding tutorial about these networks? If yes, would be better to mention the link here which may help users get started.

KumoLiu (commented):
Thanks for the PR!
I want to confirmed that the refactor will also be done in this branch, as previously discussed, right?

monai/networks/blocks/spade_norm.py
ericspod: This could be a parameter yes.

monai/networks/nets/spade_autoencoderkl.py
ericspod: Yes that'll be something to look into when refactoring the feature branch.

ericspod:
> Thanks for the PR and hard work!
> I want to confirmed that the refactor will also be done in this branch, as previously discussed, right?

We definitely will be refactoring a lot of things in this feature branch before merging into dev. There's a lot of structural overlap between networks and redundant block definitions we can consolidate. We'll also do a pass over documentation to make sure it's consistent and complete.

monai/networks/nets/spade_diffusion_model_unet.py
marksgraham: Sure, added a link

monai/networks/blocks/spade_norm.py
marksgraham: added to refactor issue

monai/networks/blocks/spade_norm.py
marksgraham: added to refactor list

monai/networks/nets/spade_autoencoderkl.py
marksgraham: I agree, it isn't all that different. Its the diffusion_model_unet where the real differences between the SPADE/non-SPADE networks are. I have added merging the normal and SPADE autoencoders to the list of items to be done in the refactor

marksgraham:
Hi @KumoLiu 
Thanks for the feedback - I've addressed the minor bits and added the others to the refactoring issue, [https://github.com/Project-MONAI/MONAI/issues/7227](https://github.com/Project-MONAI/MONAI/issues/7227). 

Just a few more components to add (noise schedulers for diffusion models, some Inferers and Engines) and then we can begin the refactor

PR #7326, opened 2023-12-18
Give more useful exception when batch is considered during matrix multiplication

### LINKED ISSUE #7323: RuntimeError with SpatialPad when input is MetaTensor and padding size is the same as tensor size
Hello,

When dealing with a MetaTensor (track_meta = True) that has the same shape you are padding to with SpatialPad, the following error occurs:

```
  File "/home/vf19/.local/share/JetBrains/Toolbox/apps/PyCharm-P/ch-0/202.7660.27/plugins/python/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
  File "/home/vf19/PycharmProjects/semantic_conditioning/venv/lib/python3.9/site-packages/monai/transforms/croppad/array.py", line 168, in __call__
    return pad_func(img_t, to_pad_, self.get_transform_info(), mode_, lazy_, **kwargs_)
  File "/home/vf19/PycharmProjects/semantic_conditioning/venv/lib/python3.9/site-packages/monai/transforms/croppad/functional.py", line 196, in pad_func
    meta_info = TraceableTransform.track_transform_meta(
  File "/home/vf19/PycharmProjects/semantic_conditioning/venv/lib/python3.9/site-packages/monai/transforms/inverse.py", line 188, in track_transform_meta
    affine = orig_affine @ to_affine_nd(len(orig_affine) - 1, affine, dtype=torch.float64)
RuntimeError: mat1 and mat2 shapes cannot be multiplied (72x3 and 24x24)
```

This doesn't happen if input is a tensor, or if I set track_meta to False with a transform prior to padding. 

**To Reproduce**

1. Use SpatialPad, not SpatialPadd
2. Input a MetaTensor with track_meta = True
3. input tensor has shape: [24, 8, 24, 32], SpatialPad gets passed spatial_size = [-1, 24, 32]

**Expected behavior**

If this specific scenario, where the shape should be different, is not supported in case of MetaTensors, the error should be clearer. In any other scenario (dealing with normal tensors) the shape is left the same and no error occurs. 

**Environment**

python 3.9
monai 1.3.0
pytorch 2.1.0, 1.13.1 (both have the same problem)

KumoLiu:
Fixes #7323

### Description

Give more useful exception when batch is considered during matrix multiplication.

### Types of changes

### DIFF
--- a/monai/transforms/inverse.py
+++ b/monai/transforms/inverse.py
@@ -189,7 +189,7 @@ class TraceableTransform(Transform):
                 affine = orig_affine @ to_affine_nd(len(orig_affine) - 1, affine, dtype=torch.float64)
             except RuntimeError as e:
                 raise RuntimeError(
-                    f"mismatch affine matrix, ensured that the batch dimension is not included in the calculation."
+                    "mismatch affine matrix, ensured that the batch dimension is not included in the calculation."
                 ) from e
             out_obj.meta[MetaKeys.AFFINE] = convert_to_tensor(affine, device=torch.device("cpu"), dtype=torch.float64)
 
--- a/monai/transforms/inverse.py
+++ b/monai/transforms/inverse.py
@@ -185,7 +185,12 @@ class TraceableTransform(Transform):
             # not lazy evaluation, directly update the metatensor affine (don't push to the stack)
             orig_affine = data_t.peek_pending_affine()
             orig_affine = convert_to_dst_type(orig_affine, affine, dtype=torch.float64)[0]
-            affine = orig_affine @ to_affine_nd(len(orig_affine) - 1, affine, dtype=torch.float64)
+            try:
+                affine = orig_affine @ to_affine_nd(len(orig_affine) - 1, affine, dtype=torch.float64)
+            except RuntimeError as e:
+                raise RuntimeError(
+                    f"mismatch affine matrix, ensured that the batch dimension is not included in the calculation."
+                ) from e
             out_obj.meta[MetaKeys.AFFINE] = convert_to_tensor(affine, device=torch.device("cpu"), dtype=torch.float64)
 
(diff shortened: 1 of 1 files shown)


monai/transforms/inverse.py
```diff
@@ -185,7 +185,12 @@ def track_transform_meta(
             # not lazy evaluation, directly update the metatensor affine (don't push to the stack)
             orig_affine = data_t.peek_pending_affine()
             orig_affine = convert_to_dst_type(orig_affine, affine, dtype=torch.float64)[0]
-            affine = orig_affine @ to_affine_nd(len(orig_affine) - 1, affine, dtype=torch.float64)
+            try:
+                affine = orig_affine @ to_affine_nd(len(orig_affine) - 1, affine, dtype=torch.float64)
+            except RuntimeError as e:
+                raise RuntimeError(
+                    "mismatch affine matrix, ensured that the batch dimension is not included in the calculation."
+                ) from e
```
ericspod, line 193:
```suggestion
            except RuntimeError as e:
                if orig_affine.ndim > 2:
                    if data_t.is_batch:
                        msg = "Transform applied to batched tensor, should be applied to instances only"
                    else:
                        msg = "Mismatch affine matrix, ensured that the batch dimension is not included in the calculation."
                    raise RuntimeError(msg) from e
                else:
                    raise
```

We should only say something about this if we know the affine matrix has an odd dimensionality (and maybe should check of it's square or not). We can check a MetaTensor to see if it's a batched tensor this way and raise the exception like this in this case. For vanilla tensors the error message may be a bit confusing but this is still better. I haven't tried this code works so please double check if you want to use it.

ericspod (approved):
Looks good, hopefully this will help others diagnose things.

Issue #6357, opened 2023-04-13
Segment Anything from google: Network implementation. Does someone started already?


vgonzalezd:
**Is your feature request related to a problem? Please describe.**
Is it someone implementing in MONAI the segment anything model in MONAI? 
[https://github.com/facebookresearch/segment-anything](url) 
My concern is the weight matching. We can help with it if someone starts already. 

**Describe the solution you'd like**
This will make more students download and use MONAI
For sure it will be an useful network for MONAILABEL

**Describe alternatives you've considered**
N/A

**Additional context**

yashika-git:
@wyli, @ericspod,
Is the implementation of Segment Anything still required?
If yes, can I try working on this?

KumoLiu:
Hi @yashika-git, yes, welcome to your contribution.
Please refer to the [CONTRIBUTION GUIDE](https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md) as well.

yashika-git:
Thanks, I've started to look how it can be implemented.

yashika-git:
Could someone please provide a little more details about the expected outcome of the implementation?
Should the users be able to infer and/or fine-tune the SAM model on custom datasets?
Thank you

KumoLiu:
Hi @yashika-git, yes first I think we should enable users to load the existing pre-trained weights, also I think they could infer and fine-tune the model on their custom datasets.

yashika-git:
Hi @KumoLiu, I have a very naive question: What exactly needs to be done to integrate SAM with MONAI? Does it involve modifying SAM's codebase to utilize MONAI's existing code, such as MONAI's networks and layers? 
I apologize for the naive question; I'm a beginner in open source and am eager to learn through contributions.

KumoLiu:
Hi @yashika-git, For existing parts that can be replaced by the MONAI, you should use the components of MONAI as much as possible.

PR #7346, opened 2023-12-29
[Attention block] relative positional embedding

### LINKED ISSUE #7356: Self attention with relative positional embedding
Enable relative positional embedding in attention block

A constructor argument specifying which type (or None) of relative embedding to use

NA

**Additional context**
Should be useful for some network implementation like SAM (https://arxiv.org/abs/2304.02643, #6357 )
PR opened #7346 with suggestion for decomposed relative positional embedding (https://arxiv.org/abs/2112.01526)

vgrau98:
Fixes #7356

### Description

Add relative positinoal embedding in attention block as described in https://arxiv.org/pdf/2112.01526.pdf
Largely inspired by https://github.com/facebookresearch/segment-anything/blob/main/segment_anything/modeling/image_encoder.py

Can be useful for #6357

### Types of changes


marksgraham:
I wonder if this is an example of the point we should start splitting out different types of attention block into their own blocks? Thoughts @KumoLiu  @ericspod ?

Note this also will have some merge conflicts as it stands with the [local window pr](https://github.com/Project-MONAI/MONAI/pull/7348)

ericspod:
@marksgraham @KumoLiu I feel with the generative merging in progress we should think about refactoring how these blocks work in general to try and separate these differing features into separate classes. We can perhaps have a single attention block class that is customisable with features, but if not we can do multiple classes but doing this cleanly will probably involve refactoring the existing things that have been merged from GenerativeModels. This PR adds something good however so we should keep this open but postpone until we have the other things merged perhaps.

vgrau98:
Following @ericspod  suggestions:
Use factory pattern to implement relative positional embedding in self attention block (use `LayerFactory` for now but open to any suggestion to better organise code). Anyone can extend the  `RelPosEmbedding` feature, or add new transformer/attention features

monai/networks/blocks/rel_pos_embedding.py
```diff
@@ -0,0 +1,51 @@
...
+
+from typing import Tuple
+
+import torch
+from torch import nn
+
+from monai.networks.blocks.attention_utils import add_decomposed_rel_pos
+
+
+class DecomposedRelativePosEmbedding(nn.Module):
```
KumoLiu, line 20: Decoupled the posembedding outside looks cleaner to me. We could also consider moving other attention components as a block such as 
https://github.com/Project-MONAI/MONAI/blob/c61c6ac2d56af08fbbfb955324b8639e266a25db/monai/networks/nets/transformer.py#L116
https://github.com/Project-MONAI/MONAI/blob/c61c6ac2d56af08fbbfb955324b8639e266a25db/monai/networks/nets/transformer.py#L138

monai/networks/blocks/selfattention.py
```diff
@@ -62,11 +72,30 @@ def __init__(
...
+        Return:
+            torch.Tensor: B x (s_dim_1 * ... * s_dim_n) x C
+        """
         output = self.input_rearrange(self.qkv(x))
         q, k, v = output[0], output[1], output[2]
-        att_mat = (torch.einsum("blxd,blyd->blxy", q, k) * self.scale).softmax(dim=-1)
+        att_mat = torch.einsum("blxd,blyd->blxy", q, k) * self.scale
+
+        # apply relative positional embedding if defined
+        att_mat = self.rel_positional_embedding(x, att_mat, q) if self.rel_positional_embedding is not None else att_mat
```
KumoLiu, line 95: Agreed with Eric, we should consider how to refactor these blocks. Not sure whether these parts can be used together, if yes, perhaps we should add them in one block otherwise we may add a base class to always avoid large classes.
https://github.com/Project-MONAI/MONAI/blob/c61c6ac2d56af08fbbfb955324b8639e266a25db/monai/networks/nets/transformer.py#L116

Do you have some suggestions here?
cc @ericspod @marksgraham @vgrau98

KumoLiu (commented):
Thanks for the quick update, looks cleaner now. Only have a few concern inline.

monai/networks/blocks/selfattention.py
marksgraham: the use of `flash attention` does not look compatible with relative positional embeddings to me, as RPE requires access to the attention matrix, whilst flash attention calculates this behind the scenes. but I think we could just check if the user has selected both and throw a warning and/or deactive flash attention if that is the case. torch 2.0 now has a native implementation of flash attention [here](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html) so we should move to that when we can, too, rather than rely on an external library.

i believe the `causal` option should be compatible with RPE - we just set some of the attention matrix elements to `-inf` after adding the RPE to it. do you agree @vgrau98?

monai/networks/layers/utils.py
```diff
@@ -124,3 +126,18 @@ def get_pool_layer(name: tuple | str, spatial_dims: int | None = 1):
...
+    embedding_type = RelPosEmbedding[embedding_name]
+
+    if has_option(embedding_type, "s_input_dims") and "s_input_dims" not in kw_args:
+        embedding_args["s_input_dims"] = s_input_dims
+    if has_option(embedding_type, "c_dim") and "c_dim" not in kw_args:
+        embedding_args["c_dim"] = c_dim
+    if has_option(embedding_type, "num_heads") and "num_heads" not in kw_args:
+        embedding_args["num_heads"] = num_heads
+
+    return embedding_type(**embedding_args)
```
ericspod, line 143:
```suggestion
def get_rel_pos_embedding_layer(name: tuple | str, s_input_dims: Optional[tuple], c_dim: int, num_heads: int):
    embedding_name, embedding_args = split_args(name)
    embedding_type = RelPosEmbedding[embedding_name]
    # create a dictionary with the default values which can be overridden by embedding_args
    kw_args = {"s_input_dims": s_input_dims, "c_dim": c_dim, "num_heads": num_heads, **embedding_args}
    # filter out unused argument names
    kw_args = {k: v for k, v in kw_args.items() if has_option(embedding_type, k)}
            
    return embedding_type(**kw_args)
```
I think this does what you wanted your implementation to do, unless I'm mistaken there was some issue with your logic when you're checking if a value is *not* in `kw_args` and adding it to `embedding_args` if that was true. In this case you wouldn't use the value `kw_args` if it was present.

monai/networks/blocks/attention_utils.py
```diff
@@ -0,0 +1,101 @@
...
+    relative_coords = (q_coords - k_coords) + (k_size - 1) * max(q_size / k_size, 1.0)
+
+    return rel_pos_resized[relative_coords.long()]
+
+
+def add_decomposed_rel_pos(
+    attn: torch.Tensor, q: torch.Tensor, rel_pos_lst: nn.ParameterList, q_size: Tuple, k_size: Tuple
+) -> torch.Tensor:
+    """
+    Calculate decomposed Relative Positional Embeddings from :paper:`mvitv2`.
```
ericspod, line 55: It's explained in the paper but a short explanation here of what this is would help.

monai/networks/blocks/rel_pos_embedding.py
```diff
@@ -0,0 +1,51 @@
...
+        self.c_dim = c_dim
+        self.num_heads = num_heads
+        self.rel_pos_arr = nn.ParameterList(
+            [nn.Parameter(torch.zeros(2 * dim_input_size - 1, c_dim)) for dim_input_size in s_input_dims]
+        )
+
+    def forward(self, x: torch.Tensor, att_mat: torch.Tensor, q: torch.Tensor) -> torch.Tensor:
+        """"""
+        batch = x.shape[0]
+        h, w = self.s_input_dims[:2] if self.s_input_dims is not None else (0, 0)
```
ericspod, line 39: When would `self.s_input_dims` ever be `None`? If never then this should be enforced in the constructor with a check that raises `ValueError`.

monai/networks/blocks/rel_pos_embedding.py
```diff
@@ -0,0 +1,51 @@
...
+        d = self.s_input_dims[2] if self.s_input_dims is not None and len(self.s_input_dims) > 2 else 1
+
```
ericspod, line 41: You could instead do something like this which will set `d` to 1 if `self.input_dims` has 2 values :
```
h, w, d = ensure_tuple_size(self.s_input_dims, 3, 1)
```

monai/networks/layers/factories.py
```diff
@@ -468,3 +472,10 @@ def constant_pad_factory(dim: int) -> type[nn.ConstantPad1d | nn.ConstantPad2d |
     """
     types = (nn.ConstantPad1d, nn.ConstantPad2d, nn.ConstantPad3d)
     return types[dim - 1]
+
+
+@RelPosEmbedding.factory_function("decomposed")
+def decomposed_rel_pos_embedding() -> type[Any]:
```
ericspod, line 478:
```suggestion
def decomposed_rel_pos_embedding() -> type[nn.Module}:
```

monai/networks/blocks/selfattention.py
ericspod: I agree, we can have a check for Pytorch version and use the included function if version 2.0 or higher.

ericspod:
I think this is looking much better. Would it be possible to merge into `gen-ai-dev` instead of `dev` so we can continue the generative refactoring? @KumoLiu @marksgraham any other feedback?

monai/networks/blocks/selfattention.py
vgrau98: i agree @marksgraham

monai/networks/layers/utils.py
vgrau98: Yes, you're absolutely right. I commit your suggestion

monai/networks/blocks/rel_pos_embedding.py
vgrau98: Never, check added in https://github.com/Project-MONAI/MONAI/pull/7346/commits/2d76c8c0926272a0711e62054bd781235ff5bf4a

monai/networks/blocks/rel_pos_embedding.py
vgrau98: https://github.com/Project-MONAI/MONAI/pull/7346/commits/ff18e81c5ff610c139383fce3cd415ca39b314b0

monai/networks/blocks/attention_utils.py
vgrau98: https://github.com/Project-MONAI/MONAI/pull/7346/commits/033ebf82873c5e67dd2798e74468272665725967

vgrau98: A simple comment, looking in more detail, if I'm not mistaken the implementation is a little different from the paper definition. 
From https://github.com/facebookresearch/segment-anything/blob/6fdee8f2727f4506cfbbe553e23b895e27956588/segment_anything/modeling/image_encoder.py#L347-L361
It seems we get 
$Attn_{logits} = QK^{T}*scale + E^{(rel)}$
instead of (from paper https://arxiv.org/abs/2112.01526):
$Attn_{logits} = (QK^{T} + E^{(rel)})*scale$

In the implementation, RPE does not seem to be scaled

vgrau98:
> ther feedback

Waiting for @marksgraham confirmation

vgrau98:
> I think this is looking much better. Would it be possible to merge into `gen-ai-dev` instead of `dev` so we can continue the generative refactoring? @KumoLiu @marksgraham any other feedback?

Waiting for @marksgraham confirmation

vgrau98:
> I think this is looking much better. Would it be possible to merge into `gen-ai-dev` instead of `dev` so we can continue the generative refactoring? @KumoLiu @marksgraham any other feedback?

target branch updated to `gen-ai-dev`

KumoLiu:
Hi @vgrau98, could you please try `git rebase --onto  <newparent> <oldparent> <until>` to rebase your commit onto the gen-ai-dev, which may avoid merging the commit added from the dev branch (causing DCO error)?
https://womanonrails.com/git-rebase-onto

vgrau98:
> Hi @vgrau98, could you please try `git rebase --onto <newparent> <oldparent> <until>` to rebase your commit onto the gen-ai-dev, which may avoid merging the commit added from the dev branch (causing DCO error)? https://womanonrails.com/git-rebase-onto

should be ok now

KumoLiu (approved):

vgrau98:
> Hi @vgrau98, could you please check-pick the commit [here](https://github.com/Project-MONAI/MONAI/pull/7391) after it merged which caused a CI error and could not merge? Thanks.

Is it still relevant? The bug seems to have been resolved according to your comments on PR #7391

KumoLiu:
Yes, I'm creating another PR to fix the error caused by the upgrade of flake8-bugbear
https://github.com/Project-MONAI/MONAI/actions/runs/7544331654/job/20537411217#step:2:2601

You can either cherry-pick that one or make the same change in this PR.

PR #7348, opened 2023-12-30
transformer block local window attention

### LINKED ISSUE #7349: Transformer block local window attention
Add local window attention in transformer block as used in transformers based networkds (SAM for instance, https://arxiv.org/abs/2304.02643)

int arg specifying window size in TransformerBlock constructor https://github.com/Project-MONAI/MONAI/blob/b3d7a48afb15f6590e02302d3b048a4f62d1cdee/monai/networks/blocks/transformerblock.py#L26C1-L34C15. If 0, global attention is used instead (no change). 

**Additional context**
Should help to add monai flexibility to implement more easily ViT with slightly different architectures. Could help for #6357

vgrau98:
Fixes #7349
Towards #6357 

### Description

Allow basic local window attention as used in Segment Anything network (image encoder) https://arxiv.org/abs/2304.02643, https://github.com/facebookresearch/segment-anything/blob/main/segment_anything/modeling/image_encoder.py

For now window attention is only supported for 2D inputs

### Types of changes


monai/networks/blocks/transformerblock.py
```diff
@@ -40,6 +45,10 @@ def __init__(
             dropout_rate (float, optional): fraction of the input units to drop. Defaults to 0.0.
             qkv_bias (bool, optional): apply bias term for the qkv linear layer. Defaults to False.
             save_attn (bool, optional): to make accessible the attention matrix. Defaults to False.
+            window_size (int): Window size for local attention as used in Segment Anything https://arxiv.org/abs/2304.02643.
+                If 0, global attention used. Only 2D inputs are supported for local attention (window_size > 0).
```
KumoLiu, line 49: Could also add support for 3d?

KumoLiu (dismissed):
Thanks for the PR. Could you please also add unit tests to cover the code?
We may also need to consider integrate all transformer block related components into one.
https://github.com/Project-MONAI/MONAI/commit/c61c6ac2d56af08fbbfb955324b8639e266a25db#diff-003ce93240016c5323702f64ea7b9f24d401ef98a8caf03763765cc958d176f6

monai/networks/blocks/transformerblock.py
vgrau98: 3d added in [70dc36c](https://github.com/Project-MONAI/MONAI/pull/7348/commits/70dc36c1c27de2609e34d6bb5645f2c55a60cd60)

vgrau98:
> Thanks for the PR. Could you please also add unit tests to cover the code? We may also need to consider integrate all transformer block related components into one. [c61c6ac#diff-003ce93240016c5323702f64ea7b9f24d401ef98a8caf03763765cc958d176f6](https://github.com/Project-MONAI/MONAI/commit/c61c6ac2d56af08fbbfb955324b8639e266a25db#diff-003ce93240016c5323702f64ea7b9f24d401ef98a8caf03763765cc958d176f6)

Unit tests added for 2d and 3d

monai/networks/blocks/transformerblock.py
```diff
@@ -51,12 +64,161 @@ def __init__(
...
+        x = self.norm1(x)
+        # Window partition
+        if self.window_size > 0:
+            if x.shape[1] != int(torch.prod(torch.tensor(self.input_size))):
+                raise ValueError(
+                    f"Input tensor spatial dimension {x.shape[1]} should be equal to {self.input_size} product"
+                )
+
+            if len(self.input_size) == 2:
+                x = rearrange(x, "b (h w) c -> b h w c", h=self.input_size[0], w=self.input_size[1])
```
KumoLiu, line 94: I would suggest moving this rearrange part to `window_partition` and 2d/3d into one `window_partition` func to make the code cleaner. After this PR, we may also consider integrating this [one](https://github.com/Project-MONAI/MONAI/commit/c61c6ac2d56af08fbbfb955324b8639e266a25db#diff-003ce93240016c5323702f64ea7b9f24d401ef98a8caf03763765cc958d176f6), which may make this `TransformerBlock` too complicated. What do you think?

marksgraham: The main integration in that PR will be added the option of cross attention for conditioning. I'm not sure how the use of local window attention + cross attention should interact. Perhaps we would use the local windowing in the self-attention block, then not use it with the second, cross attention block? 

I think it should be OK to integrate the two, but I think the integration will be much easier/cleaner if we have a single `window_partition` func as @KumoLiu suggests.

vgrau98: Thx for you comments @KumoLiu  and @marksgraham 
1. Yes you're right, i will clean the code following you suggestion
2. I also agree that the `TransformerBlock` can quickly become complex if we want to provide an exhaustive block to implement the networks presented in the literature (and those to come too). Regarding this what immediately comes to mind is defining different transformer blocks according to their subtleties (cross attention, self attention local window, and others...) rather than implement every variations in one block
3. Now considering the specifing issue of merging cross attention i don't see real issue.. local attention window was designed to be used in self-attention only (from my understanding). So cross attention should not interfere with the use of local windows for self attention. Also, i see that from https://github.com/Project-MONAI/MONAI/commit/c61c6ac2d56af08fbbfb955324b8639e266a25db#diff-003ce93240016c5323702f64ea7b9f24d401ef98a8caf03763765cc958d176f6 cross attention is defined using a self attention block, i would suggest to implement a cross attention block instead

marksgraham: Agreed with point 2, at some point we should start splitting out into different transformer blocks that inherit from a base TransformerBlock class rather than trying to implement every possible variation in one.

And also agreed with 3 - I think this specific merge should be OK, either into one block or creating a CrossAttention block for the one above linked to.

vgrau98:
- move local window utility function to utils file 
- move the logic from `TransformerBlock` to  `SABlock`

PR #7375, opened 2024-01-08
Add `compile` support in `SupervisedTrainer` and `SupervisedEvaluator`

KumoLiu:
Fixes # .

### Description

Add `compile` support in `SupervisedTrainer` and `SupervisedEvaluator`.
Convert to `torch.Tensor` internally.

### Types of changes

### DIFF
--- a/monai/engines/evaluator.py
+++ b/monai/engines/evaluator.py
@@ -17,6 +17,7 @@ import torch
 from torch.utils.data import DataLoader
 
 from monai.config import IgniteInfo, KeysCollection
+from monai.data import MetaTensor
 from monai.engines.utils import IterationEvents, default_metric_cmp_fn, default_prepare_batch
 from monai.engines.workflow import Workflow
 from monai.inferers import Inferer, SimpleInferer
@@ -25,7 +26,7 @@ from monai.transforms import Transform
 from monai.utils import ForwardMode, ensure_tuple, min_version, optional_import
 from monai.utils.enums import CommonKeys as Keys
 from monai.utils.enums import EngineStatsKeys as ESKeys
-from monai.utils.module import look_up_option
+from monai.utils.module import look_up_option, pytorch_after
 
 if TYPE_CHECKING:
     from ignite.engine import Engine, EventEnum
@@ -213,6 +214,10 @@ class SupervisedEvaluator(Evaluator):
             `device`, `non_blocking`.
         amp_kwargs: dict of the args for `torch.cuda.amp.autocast()` API, for more details:
             https://pytorch.org/docs/stable/amp.html#torch.cuda.amp.autocast.
+        compile: whether to use compile, default is False.
+            If set to True, the inputs will be converted to `torch.Tensor` internally.
+        compile_kwargs: dict of the args for `torch.compile()` API, for more details:
+            https://pytorch.org/docs/stable/generated/torch.compile.html#torch-compile.
 
     """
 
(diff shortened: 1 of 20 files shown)


monai/engines/evaluator.py
```diff
@@ -259,8 +266,12 @@ def __init__(
             to_kwargs=to_kwargs,
             amp_kwargs=amp_kwargs,
         )
-
-        self.network = network
+        if compile:
+            assert pytorch_after(2, 1)
+            self.network = torch.compile(network, **compile_kwargs)
+        else:
+            self.network = network
```
ericspod, line 273:
```suggestion
        if compile:
            if pytorch_after(2, 1):
                self.network = torch.compile(network, **compile_kwargs)
            else:
                warnings.warn("Network compilation (compile=True) not supported for Pytorch versions before 2.2, no compilation done")
```

I'd suggest warning if version is too old rather than getting an assert fail.

monai/engines/evaluator.py
```diff
@@ -288,7 +299,9 @@ def _iteration(self, engine: SupervisedEvaluator, batchdata: dict[str, torch.Ten
             kwargs: dict = {}
         else:
             inputs, targets, args, kwargs = batch
-
+        if self.compile:
+            inputs = torch.Tensor(inputs) if isinstance(inputs, MetaTensor) else inputs
+            targets = torch.Tensor(targets) if isinstance(targets, MetaTensor) else targets
```
ericspod, line 304: Is there a specific reason MetaTensor isn't compatible with compilation? I haven't tried it myself and a bit out of the loop on this. If we cast to regular tensor like this the outputs from networks lose their relationship with the original inputs, eg. filenames and such in the metadata.

KumoLiu: Will create a simple repro and raise a ticket in the torch repo, send you the link later.

KumoLiu: https://github.com/pytorch/pytorch/issues/117026

KumoLiu:
From the suggestion [here](https://github.com/pytorch/pytorch/issues/117026#issuecomment-1891774889), I tried torch-nightly. I can run the pipeline without converting `MetaTensor` to tensor after updating `traceable_tensor_subclasses`.

But as mentioned here, this feature doesn't really work and will not preserve the meta information.
https://github.com/pytorch/pytorch/blob/13473df0d770e67f92cb4bb1b728f3746f160b7c/torch/_dynamo/config.py#L101-L112

KumoLiu:
> I'm not sure that the support in Pytorch will be stable soon to be compatible with MetaTensor. If we want to integrate this we'll have to put a warning into the docstring stating that MetaTensor will be cast away. We could consider casting back to MetaTensor after the forward pass and copying the metadata from the original but this is different behaviour again that may not make things better.

Thanks, Eric for your suggestions, I have added a warning and also mentioned it in the docstring, I cast back to the `MetaTensor` after the forward pass. Could you please help review it again? Then I will add some tests.

monai/engines/evaluator.py
```diff
@@ -213,6 +215,10 @@ class SupervisedEvaluator(Evaluator):
             `device`, `non_blocking`.
         amp_kwargs: dict of the args for `torch.cuda.amp.autocast()` API, for more details:
             https://pytorch.org/docs/stable/amp.html#torch.cuda.amp.autocast.
+        compile: whether to use compile, default is False. If set to True, the inputs will be converted to `torch.Tensor`
+            and copy back the meta information after forward pass internally.
```
ericspod, line 219:
```suggestion
        compile: whether to use `torch.compile`, default is False. If True, MetaTensor inputs will be converted to 
            `torch.Tensor` before forward pass,  then converted back afterward with copied meta information.
```

ericspod (approved):
Looks good to me now, I had just a minor suggestion for the comment that should be in the Evaluator and Trainer classes.

KumoLiu:
Known issue list here as a reminder. https://github.com/pytorch/pytorch/issues/117026

PR #7379, opened 2024-01-09
6676 port generative inferers

marksgraham:
Part of #6676  .

### Description

Adds Inferers to assist with training and sampling from diffusion models and controllers.

Also takes the opportunity to  make two changes which slipped through the previous PRs:
- rename the `num_channels` arg in the spade diffusion unet to `channels` to be consistent with all the other models added from Generative - this slipped through in the networks PR.
- add the `Ordering` class to `__init__.py` for easier import

### Types of changes


monai/inferers/inferer.py
```diff
@@ -752,3 +771,1281 @@ def network_wrapper(
...
+        """
+        with torch.no_grad():
+            latent = autoencoder_model.encode_stage_2_inputs(inputs) * self.scale_factor
+
+        if self.ldm_latent_shape is not None:
+            latent = torch.stack([self.ldm_resizer(i) for i in decollate_batch(latent)], 0)
+
+        call = super().__call__
+        if isinstance(diffusion_model, SPADEDiffusionModelUNet):
+            call = partial(super().__call__, seg=seg)
```
KumoLiu, line 1139: Seems we don't need this check here, the check has already been done in the parent class?

monai/inferers/inferer.py
```diff
@@ -752,3 +771,1281 @@ def network_wrapper(
...
+            and autoencoder_model.decoder.label_nc != diffusion_model.label_nc
+        ):
+            raise ValueError(
+                "If both autoencoder_model and diffusion_model implement SPADE, the number of semantic"
+                "labels for each must be compatible. "
+            )
+
+        sample = super().sample
+        if isinstance(diffusion_model, SPADEDiffusionModelUNet):
+            sample = partial(super().sample, seg=seg)
```
KumoLiu, line 1192: Also here?

monai/inferers/inferer.py
```diff
@@ -752,3 +771,1281 @@ def network_wrapper(
             return out
 
         return tuple(out_i.unsqueeze(dim=self.spatial_dim + 2) for out_i in out)
+
+
+class DiffusionInferer(Inferer):
+    """
+    DiffusionInferer takes a trained diffusion model and a scheduler and can be used to perform a signal forward pass
+    for a training iteration, and sample from the model.
```
KumoLiu, line 779: It may be confused here.
Looks like the inferer class added in this PR can be used for training as well. Do we really need these classes to inherit `Inferer`, or should we re-add a trainer for diffusion, or more detailed in the docstring?
https://github.com/Project-MONAI/GenerativeModels/blob/bdca1bd1ea8bda4960156285f805082fff7e8686/tutorials/generative/2d_controlnet/2d_controlnet.py#L246

cc @ericspod @Nic-Ma

monai/inferers/inferer.py
```diff
@@ -752,3 +771,1281 @@ def network_wrapper(
...
+            partial(diffusion_model, seg=seg)
+            if isinstance(diffusion_model, SPADEDiffusionModelUNet)
+            else diffusion_model
+        )
+        prediction: torch.Tensor = diffusion_model(x=noisy_image, timesteps=timesteps, context=condition)
+
+        return prediction
+
+    @torch.no_grad()
+    def sample(
```
KumoLiu, line 833: The sample method is only used for visualize during training right?

KumoLiu (commented):
Thanks for the PR.
Looks like we introduce one inferer for one model, can we refactor them such as make `DiffusionInferer` more general if comes into other model. Perhaps accept pre-process or post-process or same thing others.

monai/inferers/inferer.py
marksgraham: The aim was to simplify both the trainer and the evaluation of DDPMs. Doing sampling from a DDPM is more complicated than with other models, as it involves a combination of the model, a noise scheduler, and these have to be applied many times to produce outputs. So the initial motivation was to provide a [sampling method](https://github.com/Project-MONAI/MONAI/blob/a1e1bdad9f5b0e08fc48eb6f565ddc2a8a87142f/monai/inferers/inferer.py#L833) to make it easier to sample during validation/testing. 

We then added the `__call__` method to take care of some of the steps done during training the models (adding noise to the inputs and then evaluating the model). This becomes particularly helpful when training Latent DIffusion models, as the class also takes care of transforming to and from latent space for you, significantly simplifying your training loop.

So yes, these classes can be used during both training and inference, and break the mould of MONAI's `Infererer` class, in which `__call__` is used for inference, not trianing. The structure is also quite specific to the evaluation of diffusion models. Perhaps it is better they don't inherit from the `Inferer` class? They could either inherit from `nn.Module`  or we could create a new base class for generative model training/inference? I also agree the documentation needs to be changed to make their function clearer.

monai/inferers/inferer.py
```diff
@@ -752,3 +769,1279 @@ def network_wrapper(
...
+             is instance of SPADEAutoencoderKL, segmentation must be provided.
+        """
+
+        if (
+            isinstance(autoencoder_model, SPADEAutoencoderKL)
+            and isinstance(diffusion_model, SPADEDiffusionModelUNet)
+            and autoencoder_model.decoder.label_nc != diffusion_model.label_nc
+        ):
+            raise ValueError(
+                "If both autoencoder_model and diffusion_model implement SPADE, the number of semantic"
```
KumoLiu, line 1183: `label_nc` would be better to be mentioned in the error message.

monai/inferers/inferer.py
```diff
@@ -752,3 +769,1279 @@ def network_wrapper(
...
+                "labels for each must be compatible. "
+            )
+
+        sample = super().sample
+
+        outputs = sample(
```
KumoLiu, line 1189:
```suggestion
        outputs = super().sample(
```

monai/inferers/inferer.py
```diff
@@ -752,3 +769,1279 @@ def network_wrapper(
...
+        latents = autoencoder_model.encode_stage_2_inputs(inputs) * self.scale_factor
+
+        if self.ldm_latent_shape is not None:
+            latents = torch.stack([self.ldm_resizer(i) for i in decollate_batch(latents)], 0)
+
+        get_likelihood = super().get_likelihood
+        if isinstance(diffusion_model, SPADEDiffusionModelUNet):
+            get_likelihood = partial(super().get_likelihood, seg=seg)
+
+        outputs = get_likelihood(
```
KumoLiu, line 1280:
```suggestion
        outputs = super().get_likelihood(
```

monai/inferers/inferer.py
```diff
@@ -752,3 +769,1279 @@ def network_wrapper(
...
+            inputs=latents,
+            diffusion_model=diffusion_model,
+            scheduler=scheduler,
+            save_intermediates=save_intermediates,
+            conditioning=conditioning,
+            mode=mode,
+            verbose=verbose,
```
KumoLiu, line 1287:
```suggestion
            verbose=verbose,
            seg=seg
```

monai/inferers/inferer.py
```diff
@@ -752,3 +769,1279 @@ def network_wrapper(
...
+            latent = torch.stack([self.ldm_resizer(i) for i in decollate_batch(latent)], 0)
+
+        if cn_cond.shape[2:] != latent.shape[2:]:
+            cn_cond = F.interpolate(cn_cond, latent.shape[2:])
+
+        call = super().__call__
+        if isinstance(diffusion_model, SPADEDiffusionModelUNet):
+            call = partial(super().__call__, seg=seg)
+
+        prediction = call(
```
KumoLiu, line 1659:
```suggestion
        prediction = super().__call__(
```

monai/inferers/inferer.py
```diff
@@ -752,3 +769,1279 @@ def network_wrapper(
...
+            inputs=latent,
+            diffusion_model=diffusion_model,
+            controlnet=controlnet,
+            noise=noise,
+            timesteps=timesteps,
+            cn_cond=cn_cond,
+            condition=condition,
+            mode=mode,
```
KumoLiu, line 1667:
```suggestion
            mode=mode,
            seg=seg
```

monai/inferers/inferer.py
```diff
@@ -752,3 +769,1279 @@ def network_wrapper(
...
+        """
+
+        if (
+            isinstance(autoencoder_model, SPADEAutoencoderKL)
+            and isinstance(diffusion_model, SPADEDiffusionModelUNet)
+            and autoencoder_model.decoder.label_nc != diffusion_model.label_nc
+        ):
+            raise ValueError(
+                "If both autoencoder_model and diffusion_model implement SPADE, the number of semantic"
+                "labels for each must be compatible. "
```
KumoLiu, line 1712:
```suggestion
                "labels for each must be compatible. Got {autoencoder_model.decoder.label_nc} and {diffusion_model.label_nc}"
```

monai/inferers/inferer.py
```diff
@@ -752,3 +769,1279 @@ def network_wrapper(
...
+            )
+
+        if cn_cond.shape[2:] != input_noise.shape[2:]:
+            cn_cond = F.interpolate(cn_cond, input_noise.shape[2:])
+
+        sample = super().sample
+        if isinstance(diffusion_model, SPADEDiffusionModelUNet):
+            sample = partial(super().sample, seg=seg)
+
+        outputs = sample(
```
KumoLiu, line 1722:
```suggestion
        outputs = super().sample(
```

monai/inferers/inferer.py
```diff
@@ -752,3 +769,1279 @@ def network_wrapper(
...
+            input_noise=input_noise,
+            diffusion_model=diffusion_model,
+            controlnet=controlnet,
+            cn_cond=cn_cond,
+            scheduler=scheduler,
+            save_intermediates=save_intermediates,
+            intermediate_steps=intermediate_steps,
+            conditioning=conditioning,
+            mode=mode,
+            verbose=verbose,
```
KumoLiu, line 1732:
```suggestion
            verbose=verbose,
            seg=seg
```

monai/inferers/inferer.py
```diff
@@ -752,3 +769,1279 @@ def network_wrapper(
...
+            cn_cond = F.interpolate(cn_cond, latents.shape[2:])
+
+        if self.ldm_latent_shape is not None:
+            latents = torch.stack([self.ldm_resizer(i) for i in decollate_batch(latents)], 0)
+
+        get_likelihood = super().get_likelihood
+        if isinstance(diffusion_model, SPADEDiffusionModelUNet):
+            get_likelihood = partial(super().get_likelihood, seg=seg)
+
+        outputs = get_likelihood(
```
KumoLiu, line 1823:
```suggestion
        outputs = super().get_likelihood(
```

monai/inferers/inferer.py
```diff
@@ -752,3 +769,1279 @@ def network_wrapper(
...
+            inputs=latents,
+            diffusion_model=diffusion_model,
+            controlnet=controlnet,
+            cn_cond=cn_cond,
+            scheduler=scheduler,
+            save_intermediates=save_intermediates,
+            conditioning=conditioning,
+            mode=mode,
+            verbose=verbose,
```
KumoLiu, line 1832:
```suggestion
            verbose=verbose,
            seg=seg
```

monai/inferers/inferer.py
```diff
@@ -752,3 +769,1279 @@ def network_wrapper(
...
+
+        Args:
+            input: the target images. It is assumed that this was uint8 values,
+                      rescaled to the range [-1, 1].
+            means: the Gaussian mean Tensor.
+            log_scales: the Gaussian log stddev Tensor.
+            original_input_range: the [min,max] intensity range of the input data before any scaling was applied.
+            scaled_input_range: the [min,max] intensity range of the input data after scaling.
+        """
+        assert inputs.shape == means.shape
```
KumoLiu, line 1052: Would better check the shape and raise a warning here.
Same here.
https://github.com/Project-MONAI/MONAI/blob/ccc31100e0eb23bc1e03ef6f60db28512e4a9a13/monai/inferers/inferer.py#L1070

KumoLiu (commented):
@ericspod, could you please also help review this PR?
Thanks!

monai/inferers/inferer.py
marksgraham: done - the second one was a debugging check left in so I have now deleted

test_spade_diffusion_model_unet.py
ericspod, line 1: This file needs to be in `tests`?

ericspod: There's also `test_spade_autoencoderkl.py` in the root which should be in tests.

monai/inferers/inferer.py
KumoLiu: I suggest not inheriting from the `Inferer` class and creating a new base class. What do you think? cc @marksgraham @ericspod @Nic-Ma 
But, of course, we could merge this one first.

marksgraham: sure i can do a new base class. I've added it to the refactor issue. Would be great if we could merge this if everything else is OK - getting close to having everything moved over so soon we can start on the refactor!

KumoLiu (approved):

KumoLiu:
Hi @marksgraham, could you please check-pick the [commit](https://github.com/Project-MONAI/MONAI/pull/7391) here after it merged which caused a CI error and could not merge?
Thanks.

marksgraham:
> Hi @marksgraham, could you please check-pick the [commit](https://github.com/Project-MONAI/MONAI/pull/7391) here after it merged which caused a CI error and could not merge? Thanks.

Hi, 
There is a test in `test_ordering` that sometimes fails as it has random behaviour. I don't think the test is useful and should be removed. Shall I remove it on this branch before we merge?

KumoLiu:
I think the error is caused by the update from flake8-bugbear instead of test_ordering. But anyway it is also not related to the PR, you could pin the version to avoid the error first, I will try to create a PR to fix it in the dev branch.
https://github.com/Project-MONAI/MONAI/actions/runs/7544331654/job/20537411217#step:2:2601

KumoLiu:
Hi @marksgraham, now the error seems related to the [`test_ordering.py`](https://github.com/Project-MONAI/MONAI/actions/runs/7564464286/job/20598607571#step:2:15435). As it is random, can we add `set_determinism` to ensure the test passes?
Something like:
https://github.com/Project-MONAI/MONAI/blob/25f190148c7c3e1326e55e296ef7c374903b3c52/tests/test_rand_cucim_dict_transform.py#L26

https://github.com/Project-MONAI/MONAI/blob/25f190148c7c3e1326e55e296ef7c374903b3c52/monai/utils/misc.py#L318

marksgraham:
hi @KumoLiu 

I have removed the test. As far as I can tell, the test does not make sense and does not seem to be usefully testing any part of the ordering code.

PR #7385, opened 2024-01-12
Geometric transforms proposals - Flip

### LINKED ISSUE #4027: Applying Transform Chain to Geometric Objects
Sometimes we need to apply transforms to geometric objects (point clouds, lines, curves, and meshes), particularly in radiotherapy planning and object detection.

A good solution would be a branch core library that can handle all the geometric transforms while accounting for the image metadata like image spacing and size. A geometric transform chain should be callable in a similar way as we call the regular image transforms. Something like 
```
from monai.transforms.geometric import LoadPoints, RotatePoints
```
In an ideal case when we have an image annotation pair. We apply the same set of transformation to both objects. But for geometric objects there should be some way of passing messages to the annotations transform chain. For example you are rotating images and annotations. You can apply the rotation to the annotation, but after rotation the size of the image changes in different axes and sometimes it is important that the geometric transform should know about these changes for accurately applying the transforms. So there should be a message passing mechanism between two transform chains. More relevance in discussion #4024
 
I have written my own transforms to do all these things but it is a bit hacky and not standardized. I have my geometric labels from open-source tools like [labelme](https://github.com/wkentaro/labelme). My code handles labelme annotations (in a json)  but going in I think we will need to have support for STL, DICOM RT annotations. 

**Additional context**
If there are some people working on it, please let me know as I have been working on it and will like to contribute to this branch.

KumoLiu:
Fixes #4027

### Description
In this draft PR, I have two proposals about how to add geometric transforms.
One is that assume the operation on different objects as different operators. Users can also specify their operators.
The Other one separate the flip images and flip points to Two transforms.

### Types of changes


monai/transforms/spatial/array.py
```diff
@@ -703,6 +706,131 @@ def inverse(self, data: torch.Tensor) -> torch.Tensor:
...
+        lazy: bool = False,
+        operator=None,
+        *args,
+        **kwargs,
+    ) -> None:
+        LazyTransform.__init__(self, lazy=lazy)
+        self.kwargs = kwargs
+        self.spatial_axis = spatial_axis
+        self.operators: list[Operator] = []
+        for r in SUPPORTED_OPERATORS:  # set predefined operators as default
```
Nic-Ma, line 729: Do you mean to put all kinds of operators (flip, rotate, crop, etc.) in this `SUPPORTED_OPERATORS`? `for-loop` all the operators in every transform?

Thanks.

monai/transforms/spatial/operator.py
```diff
@@ -0,0 +1,92 @@
...
+        Note that it returns a data object or a sequence of data objects.
+
+        Args:
+            data: file name or a list of file names to read.
+            kwargs: additional args for actual `read` API of 3rd party libs.
+
+        """
+        raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
+
+class FlipImageOp(Operator):
```
Nic-Ma, line 58: I don't see how to use these operators in this PR.

Thanks.

PR #7406, opened 2024-01-19
6676 port generative engines

marksgraham:
Part of #6676  .

### Description

### Types of changes


monai/engines/trainer.py
```diff
@@ -423,3 +423,282 @@ def _iteration(
             GanKeys.GLOSS: g_loss.item(),
             GanKeys.DLOSS: d_total_loss.item(),
         }
+
+
+class AdversarialTrainer(Trainer):
```
KumoLiu, line 428: I took a quick look and thought this class could be merged with `GanTrainer`. Alternatively, refactor `GanTrainer` into a more general base class. It appears that they all need to train a generator and discriminator.
What do you think?
cc @ericspod @Nic-Ma

ericspod: It would be a good idea to refactor GanTrainer, I don't think it's used much so there's room for improvement.

ericspod: Thinking about how we'd do adversarial training in general, I wonder if some approach where we have separate engines for each network, with training for the discriminator triggered by a handler when an iteration for the generator completes, and the double forward pass of the discriminator is done using a specialised inferer. It's possible we could eliminate specialised trainers for adversarial training completely, or at least make something more general that can handle things like CycleGAN or other more exotic setups.

marksgraham: Makes sense to refactor - I've added to the refactor list. 

I seem to be having a codeformat error, I think caused by an upgrade to black?

KumoLiu: Yes, have created a PR to fix it.
https://github.com/Project-MONAI/MONAI/pull/7422

marksgraham: ok merged in. not sure how to deal with DCO errors for other users, caused by merging in dev?

KumoLiu: Can the method "Rebase the branch" list here solve the issue?
https://github.com/Project-MONAI/MONAI/pull/7406/checks?check_run_id=21013930168

marksgraham:
i think this is ready to go unless there are any other comments to address?

KumoLiu:
> i think this is ready to go unless there are any other comments to address?

Yes, I will merge this one after [this one](https://github.com/Project-MONAI/MONAI/pull/7428).
@ericspod, Do you have any other comments here?

ericspod (approved):
I'm fine with this since the large number of modified files is mostly integrated changes. We'll be doing other code reviews to catch anything here later.

Issue #6959, opened 2023-09-07
Bundle Specification Needs Required Packages


ericspod:
**Is your feature request related to a problem? Please describe.**
Bundles can be specified to depend on optional packages through the `optional_packages_version` in `metadata.json`. Some bundles may require certain packages so a `required_packages_version` may be needed, or a change in name for `optional_packages_version`.

**Describe the solution you'd like**
Some way of specifying what packages other Pytorch, Numpy, and MONAI are needed by a bundle.

**Describe alternatives you've considered**
I've used `optional_packages_version` to specify needed packages but this doesn't seem adequate.

ericspod:
I don't think this is a really high priority item, but we should update the metadata schema some time anyway.

KumoLiu:
Hi @ericspod, for now, if `optional_packages_version` is in "metadata.json", it will add these packages to requirements.txt. Doesn't this currently meet the request, or do you think it would be better to change the name to `required_packages_version`?
https://github.com/Project-MONAI/model-zoo/blob/8845f3a04766e254b156c23fc4d2f91659c056d4/ci/get_bundle_requirements.py#L39

ericspod:
The issue is that there are required packages that a bundle absolutely needs and packages that would nice to have but not strictly needed. These two aren't differentiated in the json file and I felt there should be, or we change the name to `required_packages_version` to make it clear these packages are needed and not worry about having things optional.

KumoLiu:
Hi @ericspo, I think we also need to create a new schema and update all bundles. What do you think?
https://github.com/Project-MONAI/model-zoo/blob/f20f894f2c0993c94184c7002529ea303a2f8c74/models/brats_mri_segmentation/configs/metadata.json#L2C16-L2C120

ericspod:
Yes we do need a new schema. I wanted to update it to try to take into account non-tensor arguments as well as arbitrary argument names. It currently isn't quite correct without so I need to do some looking into JSON schema definitions and get to that.

KumoLiu:
What non-tensor arguments and arbitrary argument names did you refer to?
I also considering updating the output definition in the schema. For the raw output of the network, it is possible that there is more than just a single tensor, do you think we need to consider them all? Look like now we are defining the raw output of the network. But sometimes the output can be more complicated(such as detection).
https://github.com/Project-MONAI/MONAI/issues/6724
[...]

Issue #4048, opened 2022-04-01
easier access to the bundle schema


wyli:
**Is your feature request related to a problem? Please describe.**
currently the schema file is at https://github.com/Project-MONAI/MONAI-extra-test-data/releases/tag/0.8.1, would be great to have 
an alias of them under the domain `https://monai.io`, such as `https://monai.io/bundle/schema_draft/1.0.json`

e-mny:
Hi @ericspod , would this issue be addressed by #8510 ?

ericspod:
> Hi [@ericspod](https://github.com/ericspod) , would this issue be addressed by [#8510](https://github.com/Project-MONAI/MONAI/issues/8510) ?

It wouldn't I'm afraid, we want the schema to be somewhere that can be referenced by script files. I had hoped this would be the domain but it likely will have to be on Github somewhere more central. I need to discuss with internal collaborators about this one as well.

ericspod:
Schemas will not be available through the current site URL, eg. `https://project-monai.github.io/schema/meta_schema_20240725.json`.

PR #7409, opened 2024-01-22
Adding metadata schema to the code base itself

### LINKED ISSUE #7303: Update the schema for MONAI Bundle
There are several things we want to update in the schema.
https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220324.json

- Rename `optional_packages_version` to `required_packages_version`, https://github.com/Project-MONAI/MONAI/pull/7253
- Differentiate between model output and postprocessing output, https://github.com/Project-MONAI/MONAI/issues/6724
- Take into account other types in network's forward pass, https://github.com/Project-MONAI/MONAI/issues/6959#issuecomment-1825965246

ericspod:
Fixes #7303 #6959.

### Description

This adds the schema file into the code base (but this maybe should be elsewhere). The changes implement a number of new things:

* Moved definitions into a `$defs` section per the JSON schema standard
* Permits multiple input arguments and return results from networks with arbitrary names using the `patternProperties` mechanism
* Allows the types of inputs and outputs to be, additional to just tensors, numbers, booleans, or strings
* Outputs after post processing can be specified with the `post_processed_outputs` section if they are significantly changed with the post-process transforms defined in scripts
* Multiple network IO formats can be specified in addition to `network_data_format`, these must follow the pattern `<name>_data_format`
* `required_packages_version` added in addition to `optional_packages_version`

#7253 depends on this schema change.

### Types of changes


ericspod:
Should this file exist here or put elsewhere? Currently the schema is stored in `extra-test-data` within a release, but it has to be somewhere accessible through a URL in the `metadata.json` files. #4048 references the issue of making this URL accessible.

KumoLiu:
> Should this file exist here or put elsewhere? Currently the schema is stored in `extra-test-data` within a release, but it has to be somewhere accessible through a URL in the `metadata.json` files. #4048 references the issue of making this URL accessible.

Yes, just make this draft PR to help review and leave comments. After finished we can move it in `extra-test-data` or somewhere else.

yiheng-wang-nv:
Can we prepare a test case, like using 1) this schema and 2) prepare a fake metadata.json file to use `verify_metadata` in `monai.bundle.scripts` to verify that it works?

monai/bundle/meta_schema.json
```diff
@@ -0,0 +1,218 @@
+{
+    "$schema": "https://json-schema.org/draft/2019-09/schema",
```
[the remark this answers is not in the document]
ericspod: I ran the schema for existing bundles, the ones that it failed on were the new generative bundles which didn't have a `network_data_format` and so are not compliant with the current schema anyway. The open question about requiring this field or not needs resolving here, if we don't make it required then all metadata.json files pass.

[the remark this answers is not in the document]
KumoLiu: My opinion on this one is that instead of renaming and replying on the pattern property, I think these two generative models missing the `network_data_format`, so we should add them. `network_data_format` should still be a required field.
https://github.com/Project-MONAI/model-zoo/blob/e99e61aeb6fd21fc8613fd6099463ca8d7420aca/models/brats_mri_axial_slices_generative_diffusion/configs/train_diffusion.json#L18C6-L18C17
What do you think?

monai/bundle/meta_schema.json
```diff
@@ -0,0 +1,218 @@
...
+        }
+    },
+    "required": [
+        "schema",
+        "version",
+        "monai_version",
+        "pytorch_version",
+        "numpy_version",
+        "required_packages_version",
+        "optional_packages_version",
```
KumoLiu, line 211: Since we already know all of the packages we need after the bundle was created, we could remove 'optional_packages_version', which may be confusing, and instead, directly use 'required_packages_version'. What do you think? 
cc @Nic-Ma @yiheng-wang-nv

monai/bundle/meta_schema.json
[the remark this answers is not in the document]
ericspod: I think it should stay required so there's an assumption about which network is the "main" network for a bundle and can always be found. It should be a minor tweak to the generative bundles to make them conform.

monai/bundle/meta_schema.json
ericspod: My idea was that `required_packages_version` is for additional packages this bundle absolutely required which wouldn't be present with a basic MONAI install, eg. `nibabel` which would be present only if the option was specified. `optional_packages_version` are ones which the bundle uses but doesn't strictly need.

KumoLiu: Yes, I know your points. But after bundle created, the creator needs to determine which packages are optional and then finished metadata.json. For users, even if the user knows that a package is optional, and they want to use the cheapest version, they still need to modify the config file themselves, so does an optional package still make sense for the bundle? Just wondering if it's necessary to complicate this, I'm open to any opinions.

ericspod: It's not necessary that the creator or the user has to modify anything. We have networks, transforms, and other components which check optional dependencies and will work fine if one isn't present. As an example, I'm working on a bundle which, in code in the `scripts` directory, optionally uses cupy but will resort to numpy if that isn't present.

ericspod:
> Can we prepare a test case, like using 1) this schema and 2) prepare a fake metadata.json file to use `verify_metadata` in `monai.bundle.scripts` to verify that it works?

Yes I can put that together shortly.

monai/bundle/meta_schema.json
Nic-Ma: We may need to clearly describe what the "optional package" means here to avoid confusion. I was also confused at the first look..

Thanks.

ericspod:
> Can we prepare a test case, like using 1) this schema and 2) prepare a fake metadata.json file to use `verify_metadata` in `monai.bundle.scripts` to verify that it works?

@yiheng-wang-nv  I've added a notebook demonstrating this now. We won't want to merge this into dev, it's just a demo as the schema file is.

monai/bundle/meta_schema.json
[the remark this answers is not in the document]
ericspod: One thing that is different that we can easily add is that `required_packages_version` and `optional_packages_version` are both required. I think existing bundles should rename `optional_packages_version` to  `required_packages_version` in their schemas if we want these properties to be required.

monai/bundle/meta_schema.json
ericspod: I've updated the schema description line to be a bit clearer. We would also document in docs and wherever else relevant that the required packages are absolutely needed by the bundle, but the optional packages enable features or capabilities not strictly needed but may allow broader behaviours or faster performance. Eg. if an operation somewhere can use cupy to accelerate something but will default to numpy if this isn't present then cupy would be an optional package, or nibabel would be optional if the bundle can work with just numpy array loading and not need nifti loading.

ericspod:
Outstanding discussion points:
* Should `required_packages_version` and/or `optional_packages_version` be required at all? If `required_packages_version` is required then model zoo bundles will need updating.
* Should we allow multiple network formats with added `[NAME]_data_format` blocks but require there to always be a `network_data_format` block? The diffusion models in the zoo will need fixing if so.

ericspod:
I've made `required_packages_version` mandatory but not `optional_packages_version`. We'll keep the `[NAME]_data_format` content of the schema. I've added documentation about other details as well.

yiheng-wang-nv:
Hi @ericspod @KumoLiu @Nic-Ma , I rechecked this draft PR's changes.

I agree to use `required_packages_version` and enable multiple network formats. After this PR is merged, we can start to update MONAI model zoo bundles to use this schema.
Hi @ericspod , are there any other blockers for this PR? (we can remove the notebook before merge)

ericspod:
@yiheng-wang-nv I don't have any blockers for this PR. The jupyter notebook and schema file should be removed and the schema file instead uploaded [here](https://github.com/Project-MONAI/MONAI-extra-test-data/releases/tag/0.8.1) where the other versions live.

yiheng-wang-nv:
> @yiheng-wang-nv I don't have any blockers for this PR. The jupyter notebook and schema file should be removed and the schema file instead uploaded [here](https://github.com/Project-MONAI/MONAI-extra-test-data/releases/tag/0.8.1) where the other versions live.

yiheng-wang-nv:
> @yiheng-wang-nv I don't have any blockers for this PR. The jupyter notebook and schema file should be removed and the schema file instead uploaded [here](https://github.com/Project-MONAI/MONAI-extra-test-data/releases/tag/0.8.1) where the other versions live.

I see. Thanks for the response. Do you think we should add `supported_apps` into this schema directly? (sorry for mis-clicked the close button and reopen button)

KumoLiu:
I agree we can directly add `supported_apps` to the schema and make it optional. In this PR, we'll only merge the rst file and add the new schema [here](https://github.com/Project-MONAI/MONAI-extra-test-data/releases/tag/0.8.1). After that, we need to merge this PR: https://github.com/Project-MONAI/MONAI/pull/7253. Subsequently, we will update all bundles for consistency.
So The first step is to add `supported_apps` and upload the new schema. What do you think?

ericspod:
> agree we can directly add `supported_apps` to the schema and make it optional. In this PR, we'll only merge the rst file and add the new schema [here](https://github.com/Project-MONAI/MONAI-extra-test-data/releases/tag/0.8.1). After that, we need to merge this PR: #7253. Subsequently, we will update all bundles for consistency.
> So The first step is to add `supported_apps` and upload the new schema. What do you think?

I've added `supported_apps` to the schema, please check the wording of the description and documentation matches the intent.

monai/bundle/meta_schema.json
```diff
@@ -0,0 +1,227 @@
...
+        "references": {
+            "description": "List of published referenced relating to the bundle.",
+            "type": "array",
+            "items": {
+                "type": "string"
+            }
+        },
+        "supported_apps": {
+            "description": "List of supported applications, eg. 'monai-label'",
+            "type": "array",
```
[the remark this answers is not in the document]
ericspod: That would be the alternative, I don't think we had consensus on that yet. The type would just become `"object"` in that case, yes.

monai/bundle/meta_schema.json
```diff
@@ -0,0 +1,227 @@
...
+        },
+        "numpy_version": {
+            "description": "Version of NumPy the bundlewas generated with.",
+            "type": "string"
+        },
+        "required_packages_version": {
+            "description": "Dictionary relating required package names to their versions. The bundle requires these packages to operate which are additional to the base requirements for MONAI.",
+            "type": "object"
+        },
+        "optional_packages_version": {
```
KumoLiu, line 142: BTW, I recall that during our dev meeting a few weeks ago, we agreed to remove the optional_packages_version from this schema. Did you forget to remove it here?

If everyone else agrees, leaving this area optional in the schema also looks fine to me.

cc @yiheng-wang-nv @Nic-Ma

monai/bundle/meta_schema.json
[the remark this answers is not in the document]
KumoLiu: Okay, let's listen to others as well. cc @yiheng-wang-nv @Nic-Ma @mingxueg-nv

monai/bundle/meta_schema.json
KumoLiu: I suspect there may have been a misunderstanding between us. #7253 was implemented to ensure that if users are still using `optional_packages_version`, it will be converted to `required_packages_version` and a warning message will be added.

This also reminds me of another issue: if a user specifies both areas, #7253 will still force the rename of optional to required if both areas exist in the metadata.json.

ericspod: I think I understand correctly here. It makes sense to have this to convert existing usage of `optional_packages_version` to `required_packages_version` and make the warning. I have removed the reference in the schema and documentation here.

monai/bundle/meta_schema.json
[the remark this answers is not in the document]
mingxueg-nv: support_apps should to be defined in metadata.json like  `"support_apps": {"NIM": "0.0.4", "MONAI-LABEL": >=0.5.0 }`, so define it as object in schema better.
There's nothing else, looks fine to me.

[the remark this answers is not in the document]
yiheng-wang-nv: agree to define as an object.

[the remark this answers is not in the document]
ericspod: That is changed now in the proposed schema.

monai/bundle/meta_schema.json
Nic-Ma: Looks good to me.

Thanks.

Nic-Ma (approved):

monai/bundle/meta_schema.json
```diff
@@ -0,0 +1,220 @@
...
+                    "$ref": "#/$defs/tensor"
+                }
+            ]
+        },
+        "result": {
+            "$comment": "Return value from a model's forward pass, same as an argument for now.",
+            "$ref": "#/$defs/argument"
+        },
+        "network_io": {
+            "description": "Defines the format, shape, and meaning of inputs and outputs to the model.",
```
mingxin-zheng, line 76: Do we need to mention `optionally "post_processed_outputs` here, which is mentioned in **network_data_format** section of the mb_spec rst?

monai/bundle/meta_schema.json
```diff
@@ -0,0 +1,220 @@
...
+        "copyright": {
+            "description": "State copyright of the bundle.",
+            "type": "string"
+        },
+        "data_source": {
+            "description": "Where to download or prepare the data used in this bundle.",
+            "type": "string"
+        },
+        "data_type": {
+            "description": "Type of the data, like: `dicom`, `nibabel`, etc.",
```
mingxin-zheng, line 163: To resolve ambiguity of either user should put `nibabel` vs. `nii`, would it be helpful to say?
```
"Type of the data, like: `dicom`, `nibabel`, etc. (e.g., 'dicom', 'nii', 'nii.gz')",
```

docs/source/mb_specification.rst
```diff
@@ -89,6 +90,8 @@ Optional keys:
 * **data_source**: description of where training/validation can be sourced.
 * **data_type**: type of source data used for training/validation.
 * **references**: list of published referenced relating to the model.
+* **supported_apps**: list of supported applications which use bundles, eg. 'monai-label' would be present if the bundle is compatible with MONAI Label applications.
+* **required_packages_version**: dictionary relating required package names to their versions. These packages are needed to be installed with this stated version.
```
KumoLiu, line 94: don't need this "required_packages_version" here?

mingxin-zheng:
My comments are minor and please feel free to skip/close them as needed for merge. Thanks!

docs/source/mb_specification.rst
```diff
@@ -63,12 +63,13 @@ This file contains the metadata information relating to the model, including wha
...
-* **network_data_format**: defines the format, shape, and meaning of inputs and outputs to the model, contains keys "inputs" and "outputs" relating named inputs/outputs to their format specifiers (defined below).
+* **network_data_format**: defines the format, shape, and meaning of inputs and outputs to the (primary) model, contains keys "inputs" and "outputs" relating named inputs/outputs to their format specifiers (defined below). There is also an optional "post_processed_outputs" key stating the format of "outputs" after postprocessing transforms are applied, this is used to describe the final output from the bundle if it varies from the raw network output. These keys can also relate to primitive values (number, string, boolean), instead of the tensor format specified below.
+* **\*_data_format**: defines the format, shape, and meaning of inputs and outputs to additional models which are secondary to the main model. This contains the same sort of information as **network_data_format** which describes networks providing secondary functionality, eg. a localisation network used to identify ROI in an image for cropping before data is sent to the primary network of this bundle.
```
KumoLiu, line 72: And this one should be in the "Optional keys:"?

monai/bundle/meta_schema.json
ericspod: I'll add something for this.

ericspod:
I've addressed comments and remove the schema file now.

ericspod:
The schema is now uploaded as [meta_schema_20240725.json](https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20240725.json).

monai/bundle/meta_schema.json
mingxin-zheng: Sounds good

yiheng-wang-nv:
Thanks @ericspod for the contribution. I will use the new schema in: https://github.com/Project-MONAI/model-zoo/pull/605

PR #7413, opened 2024-01-24
Instantiation mode `"partial"` to `"callable"`. Return the `_target_` component as-is when in `_mode_="callable"` and no kwargs are specified

ibro45:
### Description

A `_target_` component with `_mode_="partial"` will still be wrapped in `functools.partial` even when no kwargs are passed: `functool.partial(component)`. In such cases, the component can just be returned as-is.

If you agree with this, I will add tests for it. Thank you!

### Types of changes


KumoLiu:
Hi @ibro45, thanks for your PR, could you please mention the case when it will be an issue? Perhaps a small snippets of code. Thanks.

ibro45:
I was trying to achieve something similar to this:

```yaml
    inferer:
        _target_: monai.inferers.PatchInferer
        batch_size: 8
        splitter:
            _target_: monai.inferers.SlidingWindowSplitter
            patch_size: [96, 96, 96]
            overlap: 0.5
        merger_cls:
            _target_: monai.inferers.AvgMerger  # I had a custom merger here that I would have to import if I wanted to use $ 
            _mode_: partial
```

but it would fail to instantiate because it complained that the `merger_cls` is not a subclass of `Merger`, but `a functools.partial` instead. 

I know this could be solved with `$`, but it may require additional importing which `_target_` does not need. It simply seems like a neater option to use `_target_` there.

Thank you!

KumoLiu:
> but it would fail to instantiate because the `PatchInferer` complained that the `merger_cls` is not a subclass of `Merger`, but `a functools.partial` instead.

I get the point now, so instead of return the component directly when no kwargs, I would suggest add a mode here or make it clear in the docstring.
https://github.com/Project-MONAI/MONAI/blob/6c9e49ee045c345ecd8d3518cb0d503ed02c61b7/monai/utils/module.py#L235-L239
What do you think?
cc @ericspod @Nic-Ma

ericspod:
> > but it would fail to instantiate because the `PatchInferer` complained that the `merger_cls` is not a subclass of `Merger`, but `a functools.partial` instead.
> 
> I get the point now, so instead of return the component directly when no kwargs, I would suggest add a mode here or make it clear in the docstring.

I'd suggest a mode name change too. Perhaps "partial" should become "callable" or something to indicate that you could get a partial object back or just whatever the callable thing was if you gave no arguments.

ibro45:
I think renaming `"partial"` to `"callable"` works with appropriate docstrings.

I may be overcomplicating it here, but what about changing the mode names as follows:
 
 - `"default"` -> `"invoke"` mode - Executes a function or instantiates a class immediately, returning the result or instance.

- `"partial"` -> `"callable"` mode -  Directly returns the function or class for future invoking. If kwargs are provided, it creates a `functools.partial` combining these kwargs with the function or class, enabling delayed invocation with pre-set parameters.

- `"debug"` stays as-is.

KumoLiu:
> I would leave "default" alone but otherwise this sounds sensible. @KumoLiu thoughts?

Agreed.

ericspod:
> > I would leave "default" alone but otherwise this sounds sensible. @KumoLiu thoughts?
> 
> Agreed.

Let's fix the minor flake issues but otherwise it looks good.

ericspod (approved):

KumoLiu:
Hi @ibro45, could you please fix the flake8 issue and DCO issue, then I will help merge this one. Thanks.
https://github.com/Project-MONAI/MONAI/actions/runs/7635293460/job/20800473351?pr=7413#step:7:387

ibro45:
@KumoLiu should be good now, also renamed "partial" to "callable"

PR #7462, opened 2024-02-10
#7437 issue: Raise ModuleNotFoundError for missing dependencies in Lo…

### LINKED ISSUE #7437: Raise the exception when LoadImage has a reader specified but it is not installed
**Describe the bug**

**Expected behavior**

MartinKalema:
Fixes #7437 .

### Description

This commit modifies the LoadImage constructor to raise a ModuleNotFoundError
when encountering missing dependencies for specified readers. Previously, the
constructor logged debug messages or raised warnings, which could lead to
unexpected behavior when required packages were not installed.

### Changes Made:
- Modified the constructor to raise a ModuleNotFoundError when a required package
  for a specified reader is not installed. 
These changes ensure consistent error handling and provide clearer feedback to
users about missing dependencies in the LoadImage constructor.

### Types of changes

Screen shot showing the changes
changes

Passed the linting tests
passed_linting_testspng

### DIFF
--- a/monai/transforms/io/array.py
+++ b/monai/transforms/io/array.py
@@ -180,10 +180,10 @@ class LoadImage(Transform):
         for r in SUPPORTED_READERS:  # set predefined readers as default
             try:
                 self.register(SUPPORTED_READERS[r](*args, **kwargs))
-            except OptionalImportError:
-                logging.getLogger(self.__class__.__name__).debug(
-                    f"required package for reader {r} is not installed, or the version doesn't match requirement."
-                )
+            except OptionalImportError as err:
+                raise ModuleNotFoundError(
+                    f"The required package for reader {r} is not installed, or the version doesn't match requirement."
+                ) from err
             except TypeError:  # the reader doesn't have the corresponding args/kwargs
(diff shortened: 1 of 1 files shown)


MartinKalema:
During the checks, we got the following error "ModuleNotFoundError: The required package for reader pydicomreader is not installed, or the version doesn't match requirement.
" which is a result of the code changes made. This shows that the code raises an exception rather than logging it.

error_during_checks

monai/transforms/io/array.py
```diff
@@ -180,10 +180,10 @@ def __init__(
         for r in SUPPORTED_READERS:  # set predefined readers as default
             try:
                 self.register(SUPPORTED_READERS[r](*args, **kwargs))
-            except OptionalImportError:
-                logging.getLogger(self.__class__.__name__).debug(
-                    f"required package for reader {r} is not installed, or the version doesn't match requirement."
-                )
+            except OptionalImportError as err:
```
KumoLiu, line 183: If the user does not specify the reader, when the first reader package he comes across doesn't exist, he will then report the error directly, which goes against our initial plan.

Issue #7445, opened 2024-02-07
Blossom CI crashing


KumoLiu:
```
11:20:39  Starting test: test_consistent_resize (tests.test_resized.TestResized)...
11:28:16  monai_pip.sh: line 69:  1413 Killed                  PYTHONPATH="$tmp_dir":$PYTHONPATH BUILD_MONAI=1 python ./tests/runner.py -p 'test_((?!integration).)'
```
After Pytorch 2.2, blossom ci terminating unexpectedly.

MartinKalema:
Ensure that the environment where the tests are running has sufficient memory.

KumoLiu:
https://github.com/pytorch/pytorch/issues/120406

PR #7474, opened 2024-02-19
Skip "test_gaussian_filter" as a workaround for blossom killed

KumoLiu:
workaround for #7445

### Types of changes

### DIFF
--- a/tests/min_tests.py
+++ b/tests/min_tests.py
@@ -208,6 +208,7 @@ def run_testsuit():
         "test_zarr_avg_merger",
         "test_perceptual_loss",
         "test_ultrasound_confidence_map_transform",
+        "test_gaussian_filter", # https://github.com/Project-MONAI/MONAI/issues/7445
     ]
     assert sorted(exclude_cases) == sorted(set(exclude_cases)), f"Duplicated items in {exclude_cases}"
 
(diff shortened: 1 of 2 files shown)


ericspod:
Do we know if one test or another is causing the issue? In test_gaussian_filter.py there's tests doing backprop and others testing other things, can we keep any of these instead of ignoring the file?

KumoLiu:
> Do we know if one test or another is causing the issue? In test_gaussian_filter.py there's tests doing backprop and others testing other things, can we keep any of these instead of ignoring the file?

Hi @ericspod, I still haven't found the root problem, I have discovered that this unit test is the cause and only occurs on v100. Additionally, it comes after PyTorch2.2, so I added a check only skip the test after 2.2, what do you think?

ericspod:
> Hi @ericspod, I still haven't found the root problem, I have discovered that this unit test is the cause and only occurs on v100. Additionally, it comes after PyTorch2.2, so I added a check only skip the test after 2.2, what do you think?

Thanks @KumoLiu this looks good now if it works. Please add a comment with the skip decorators so that we are aware of why we're skipping if we want to come back to it in the future. We know there's a number of other inexplicable issues with tests combined with specific environments that we should circle back later.

ericspod (approved):

PR #7487, opened 2024-02-22
Add __repr__ and __str__ to Metrics baseclass

MathijsdeBoer:
### Description

When training a model using MONAI metrics for experiment tracking, I tend to log which metrics I am using. Unfortunately, just sending the metrics objects to Tensorboard will result in a list like [CustomMetric1, CustomMetric2, <monai.metrics.Metric ...>, etc.]

Adding `__repr__` and `__str__` methods to the base class will solve this small annoyance. The current implementation will only return the class name, but if a certain metric would wish to report more data for its `__repr__` string, this can be easily overridden in any subclass.

### Types of changes

### DIFF
--- a/monai/metrics/metric.py
+++ b/monai/metrics/metric.py
@@ -37,9 +37,6 @@ class Metric(ABC):
         """
         raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
 
-    def __repr__(self):
-        return self.__class__.__name__
-
     def __str__(self):
         return self.__class__.__name__
 
--- a/monai/metrics/metric.py
+++ b/monai/metrics/metric.py
@@ -37,6 +37,12 @@ class Metric(ABC):
         """
         raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
 
+    def __repr__(self):
+        return self.__class__.__name__
+
+    def __str__(self):
+        return self.__class__.__name__
+
 
 class IterationMetric(Metric):
     """
--- a/monai/apps/nnunet/nnunetv2_runner.py
+++ b/monai/apps/nnunet/nnunetv2_runner.py
@@ -37,6 +37,7 @@ class nnUNetV2Runner:  # noqa: N801
     """
     ``nnUNetV2Runner`` provides an interface in MONAI to use `nnU-Net` V2 library to analyze, train, and evaluate
     neural networks for medical image segmentation tasks.
+    A version of nnunetv2 higher than 2.2 is needed for this class.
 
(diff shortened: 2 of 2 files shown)


MathijsdeBoer:
Ideally one would exist for the Loss classes, too. But the baseclass for them is `torch.nn.modules.loss._Loss`, so either one could make a subclass of that class, add the `__repr__` and `__str__` methods to it, and point all existing loss classes to the new one instead, or these methods could be added to the baseclass in PyTorch with a pull request in that project.

monai/metrics/metric.py
```diff
@@ -37,6 +37,12 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any:
         """
         raise NotImplementedError(f"Subclass {self.__class__.__name__} must implement this method.")
 
+    def __repr__(self):
+        return self.__class__.__name__
```
KumoLiu, line 41: Thanks for the PR!
It would be beneficial to have something akin to <monai.metrics.meandice.DiceMetric object at 0x7f7b820c08e0>, as it will yield more detailed information pertaining to DiceMetric. Despite this, I would still recommend utilizing the original implementation, or alternatively, consider the addition of the __str__ method for simplification.

What do you think? cc @ericspod @Nic-Ma

MathijsdeBoer: What additional important information does the text `<monai.metrics.meandice.DiceMetric object at 0x7f7b820c08e0>` give over `DiceMetric`?

ericspod: The purpose of `__repr__` is to provide an unambiguous representation of an object, rather than `__str__` which is a more readable and descriptive representation. With the original `__repr__` we get the object ID which allows us to tell multiple instances of an object apart, this is important during debugging or testing. I would say that returning the class name only is ambiguous since it lacks any way of certainly telling instances apart.

monai/metrics/metric.py
```diff
@@ -37,6 +37,12 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any:
...
+
+    def __str__(self):
```
ericspod, line 43: Here `__str__` is producing something readable and concise which is fine, in other contexts it might make sense to include some state information if this wasn't too lengthy.

MathijsdeBoer: I've removed `__repr__`, just having the `__str__` will reduce a bit of code for me either way!

monai/metrics/metric.py
MathijsdeBoer: `__repr__` removed, makes sense that one would like to have more detailed information from that call.

KumoLiu (approved):

PR #7490, opened 2024-02-23
Geom api

### LINKED ISSUE #7486: Geometric transforms proposal
#### 1. Integration of 'kind' Property to MetaTensor:
Propose to incorporate 'kind' property in [MetaTensor](https://github.com/Project-MONAI/MONAI/blob/f4103c5fd9d8b511041ff0e20d9db5d6f03dd039/monai/data/meta_tensor.py#L51). The property 'kind' will enable efficient identification and appropriate handling of different data types. The value of 'kind' can be conveniently retrieved using data.kind.
#### 2. Data Input/Output Enhancements:
Introduce `LoadPoint` and `LoadPointd` with properties `refer` and `refer_key`. These properties will ascertain if the loaded point corresponds to a certain coordinate system and subsequently facilitate retrieval of information such as affine information from the reference.
Usage Examples:
`LoadPointd(key="point", refer_key="image")` and `LoadPointd(data=point, refer=image)`
_**Subject for Discussion:**_ What data formats should we aim to support?
#### 3. Improvements to Transform API:
The core idea is to house the computational logic within the associated operator and register it to the transform. This modification will minimize changes to the transform API. To accommodate a new data type in MONAI, current user-facing API logic would remain unaltered. New operators will simply be added as required.
Example:
```
class Flip():
    def __init__(self) -> None:
        self.operators = [flip_image, flip_point]

    def __call__(self, data, *args: Any, **kwds: Any) -> Any:
        for _operator in self.operators:
            ret = _operator(data)
            if ret is not None:
                return ret
    
    def register():
        pass

def flip_image(data):
    if data.kind != "pixel":
        return None
    else:
        ...
        return data

def flip_point(data):
    if data.kind != "point":
        return None
    else:
        ...
        return data

```
#### 4. User Experience Enhancements:
The user experience can be improved by making the data operations more intuitive and user-friendly.
Code example:
```
from monai.transform as mt

data = [
    "image": image_path,
    "point": point_path
]

trans = mt.Compose([
    mt.LoadImaged(keys="image"),
    mt.LoadPointd(keys="point", refer_key="image"),
    mt.Flipd(keys=["image", "point"]),
    mt.Rotated(keys=["image", "point"]),
])
```

atbenmurray:
Fixes #7486.

This PR is an investigation of the API changes required to implement #7486. As it stands, it isn't yet intended to be _the_ PR for the API changes, but it could turn into it if everyone is happy with it.

### Description

API changes required to support raster and geometry tensors. So far this means:
 - MapTransform minor API extension
 - Spatial dictionary transforms (which may need no API extension)
 - Spatial array transforms (which will likely need API extensions)

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -268,6 +268,7 @@ class ResampleToMatch(SpatialResample):
         align_corners: bool | None = None,
         dtype: DtypeLike = None,
         lazy: bool | None = None,
+        update_to_match: Sequence[NdarrayOrTensor] | None = None,
     ) -> torch.Tensor:
         """
         Args:
@@ -321,6 +322,8 @@ class ResampleToMatch(SpatialResample):
                     original_fname = img.meta.get(Key.FILENAME_OR_OBJ, "resample_to_match_source")
                     img.meta = deepcopy(img_dst.meta)
                     img.meta[Key.FILENAME_OR_OBJ] = original_fname  # keep the original name, the others are overwritten
+                for t in update_to_match or []:
+                    t.apply_transform(dst_affine, lazy=False) # PROBLEM: this is in-place; we'd have to return the transforms
         else:
             if isinstance(img, MetaTensor) and isinstance(img_dst, MetaTensor):
                 original_fname = img.meta.get(Key.FILENAME_OR_OBJ, "resample_to_match_source")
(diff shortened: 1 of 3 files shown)


atbenmurray:
@KumoLiu @Nic-Ma @ericspod I think we can close this PR. I have a branch on my fork of MONAI that significantly updates this work.

KumoLiu:
@atbenmurray Sure, close for now. And you can create any further pr later. Thanks

Issue #7425, opened 2024-01-30
Add sample_std parameter to RandGaussianNoise


tim-the-baker:
**Is your feature request related to a problem? Please describe.**
The documentation for `RandGaussianNoise` does not make clear that the standard deviation is sampled uniformly from 0 to `std`. (at least in my opinion; the doc-string of `std` reads "Standard deviation (spread) of distribution'). Additionally, it may be useful to have the option to not sample the standard deviation, and just use `std` as the spread every time.

**Describe the solution you'd like**
Add a `sample_std` parameter to `RandGaussianNoise`. When True, the noise's standard deviation is sampled uniformly from 0 to `std` (i.e., what is currently done). When False, the noise's standard deviation is non-random and set to `std`. The default for `sample_std` would be True for backwards compatibility. 

**Describe alternatives you've considered**
Updating the documentation of `RandGaussianNoise` to make more clear that the `std` is being sampled.

**Additional context**
- Possible use case of `sample_std=False`: Adding different levels of random Gaussian noise to images to test the robustness of a registration algorithm to varying levels of Gaussian noise.
- `RandRicianNoise` is similar to `RandGaussianNoise` in concept and implementation. `RandRicianNoise` already has the proposed `sample_std` parameter.
- I'd like to implement this feature request and think it would be a good first issue. I would base the implementation off of `RandRicianNoise`'s implementation.

PR #7492, opened 2024-02-26
Add sample_std parameter to RandGaussianNoise.

tim-the-baker:
Fixes issue #7425

### Description

Add a `sample_std` parameter to `RandGaussianNoise` and `RandGaussianNoised`. When True, the Gaussian's standard deviation is sampled uniformly from 0 to std (i.e., what is currently done). When False, the noise's standard deviation is non-random and set to std. The default for sample_std would be True for backwards compatibility.

Changes were based on RandRicianNoise which already has a `sample_std` parameter and is similar to RandGaussianNoise in concept and implementation.

### Types of changes

### DIFF
--- a/monai/transforms/intensity/array.py
+++ b/monai/transforms/intensity/array.py
@@ -98,11 +98,12 @@ class RandGaussianNoise(RandomizableTransform):
     backend = [TransformBackends.TORCH, TransformBackends.NUMPY]
 
     def __init__(
-            self, prob: float = 0.1,
-            mean: float = 0.0,
-            std: float = 0.1,
-            dtype: DtypeLike = np.float32,
-            sample_std: bool = True,
+        self,
+        prob: float = 0.1,
+        mean: float = 0.0,
+        std: float = 0.1,
+        dtype: DtypeLike = np.float32,
+        sample_std: bool = True,
     ) -> None:
         RandomizableTransform.__init__(self, prob)
         self.mean = mean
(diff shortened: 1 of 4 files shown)


tim-the-baker:
**Question:** I added the new `sample_std` parameter to the end of the `RandGaussianNoise` and `RandGaussianNoised` `__init__` functions as to not break existing code that passed parameters by position. However, it seems weird that `allow_missing_keys` is not the last parameter of `RandGaussianNoised`. What is the correct trade-off between code consistency and backwards compatibility?

KumoLiu:
>  I added the new sample_std parameter to the end of the RandGaussianNoise and RandGaussianNoised __init__ functions as to not break existing code that passed parameters by position

This way should be ok.

tim-the-baker:
Cool thanks. Also, I am running tests locally and the `tests/test_integration_classification_2d.py` test seems to be running slow. I think it's because I am running on Mac silicon - is there anyway around this?

Not sure the integration tests are necessary for such a small change, but I may work on other MONAI features in the future.

KumoLiu:
> Not sure the integration tests are necessary for such a small change, but I may work on other MONAI features in the future.

It would be great if the tests in the relevant part could be tested. Of course, I will do more tests before merging.

tim-the-baker:
> > Not sure the integration tests are necessary for such a small change, but I may work on other MONAI features in the future.
> 
> It would be great if the tests in the relevant part could be tested. Of course, I will do more tests before merging.

Yeah I added a test for the new `sample_std` parameter and ran the tests related to `RandGaussianNoise` as well the quick tests. Everything passed locally.

ericspod (approved):
Looks good to me, I don't think these changes will affect compatibility anywhere.

PR #7500, opened 2024-02-27
Add arm support

KumoLiu:
Fixes # .

### Description

Add arm support

### Types of changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,12 @@ FROM ${PYTORCH_IMAGE}
 
 LABEL maintainer="monai.contact@gmail.com"
 
+# TODO: remark for issue [revise the dockerfile](https://github.com/zarr-developers/numcodecs/issues/431)
+WORKDIR /opt
+RUN git clone --recursive https://github.com/zarr-developers/numcodecs.git
+WORKDIR /opt/numcodecs
+RUN pip wheel .
+
 WORKDIR /opt/monai
 
 # install full deps
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,9 +18,7 @@ LABEL maintainer="monai.contact@gmail.com"
 
 # TODO: remark for issue [revise the dockerfile](https://github.com/zarr-developers/numcodecs/issues/431)
 WORKDIR /opt
-RUN git clone --recursive https://github.com/zarr-developers/numcodecs.git
-WORKDIR /opt/numcodecs
-RUN pip wheel .
+RUN git clone --recursive https://github.com/zarr-developers/numcodecs.git && pip wheel numcodecs
 
 WORKDIR /opt/monai
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -26,7 +26,7 @@ mypy>=1.5.0
 ninja
 torchvision
 psutil
-cucim-cu12; platform_system == "Linux"
+cucim-cu12; platform_system == "Linux"; '3.9' <= python_version <= '3.10'
 openslide-python
 imagecodecs; platform_system == "Linux" or platform_system == "Darwin"
 tifffile; platform_system == "Linux" or platform_system == "Darwin"
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -46,7 +46,7 @@ pynrrd
 pre-commit
 pydicom
 h5py
-nni; platform_system == "Linux"
+nni; platform_system == "Linux" and "arm" not in platform_machine and "aarch" not in platform_machine
 optuna
 git+https://github.com/Project-MONAI/MetricsReloaded@monai-support#egg=MetricsReloaded
 onnx>=1.13.0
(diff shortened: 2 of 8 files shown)


requirements-dev.txt
```diff
@@ -26,7 +26,7 @@ mypy>=1.5.0
 ninja
 torchvision
 psutil
-cucim>=23.2.0; platform_system == "Linux"
+cucim-cu12; platform_system == "Linux" and python_version >= "3.9" and python_version <= "3.10"
```
ericspod, line 29: This is related to cucim compatibility issues or specifically related to ARM?

tests/test_convert_to_onnx.py
```diff
@@ -29,6 +30,12 @@
 TESTS = list(itertools.product(TORCH_DEVICE_OPTIONS, [True, False], [True, False]))
 TESTS_ORT = list(itertools.product(TORCH_DEVICE_OPTIONS, [True]))
 
+ON_AARCH64 = platform.machine() == "aarch64"
+if ON_AARCH64:
+    rtol, atol = 1e-1, 1e-2
```
ericspod, line 35: Any idea why ARM is less accurate with these tests? Floating point errors? I think it's fine to have this since the tests aren't about exacting accuracy with results.

requirements-dev.txt
KumoLiu, line 1: Note: to enable ARM, it would be additionally necessary to remove nni requirement within this file.

requirements-dev.txt
KumoLiu: There are two key reasons for the update:
- Firstly, as highlighted by @drbeh, cucim will no longer be supporting PyPI in the future. Hence, we need to transition to a package with a different name.
- Secondly, this new package provides support for ARM architecture, unlike the original one.

tests/test_convert_to_onnx.py
KumoLiu: The discrepancy between the results generated on the ARM and x86 architectures primarily stems from the differing precision handling inherent to each architecture. Additionally, I have discovered that the matmul operation produces varying results when performed on these architectures.

Here is a demonstrative example:
```
a = torch.tensor([[[ 1.0000,  0.0000,  0.0000, -0.5000],
         [ 0.0000,  1.0000,  0.0000, -0.5000],
         [ 0.0000,  0.0000,  0.6667, -0.6667],
         [ 0.0000,  0.0000,  0.0000,  1.0000]]], device='cuda:0')
b = torch.tensor([[[ 1.,  0.,  0.,  0.],
         [-0., -1., -0.,  1.],
         [ 0.,  0.,  1.,  0.],
         [ 0.,  0.,  0.,  1.]]], device='cuda:0')
c = torch.tensor([[[1.0000, 0.0000, 0.0000, 0.5000],
         [0.0000, 1.0000, 0.0000, 0.5000],
         [0.0000, 0.0000, 1.5000, 1.0000],
         [0.0000, 0.0000, 0.0000, 1.0000]]], device='cuda:0')

print(a @ b @ c)
In x86:
tensor([[[ 1.0000,  0.0000,  0.0000,  0.0000],
         [ 0.0000, -1.0000,  0.0000,  0.0000],
         [ 0.0000,  0.0000,  1.0001,  0.0000],
         [ 0.0000,  0.0000,  0.0000,  1.0000]]], device='cuda:0')
In arm:
tensor([[[ 1.0000,  0.0000,  0.0000,  0.0000],
         [ 0.0000, -1.0000,  0.0000,  0.0000],
         [ 0.0000,  0.0000,  0.9998,  0.0000],
         [ 0.0000,  0.0000,  0.0000,  1.0000]]], device='cuda:0')
```

Dockerfile
```diff
@@ -16,6 +16,12 @@ FROM ${PYTORCH_IMAGE}
 
 LABEL maintainer="monai.contact@gmail.com"
 
+# TODO: remark for issue [revise the dockerfile](https://github.com/zarr-developers/numcodecs/issues/431)
+WORKDIR /opt
+RUN git clone --recursive https://github.com/zarr-developers/numcodecs.git
+WORKDIR /opt/numcodecs
+RUN pip wheel .
```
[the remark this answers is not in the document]
KumoLiu: Tested and works well.

tests/test_convert_to_onnx.py
ericspod: The difference is a little surprising, it could be CUDA behaviour on ARM, differences in float32 accuracy, something in Pytorch, or a combination. I don't know if we can do anything about these so let's leave the tests, but if anyone asks why their ARM results are different we'll have something to discuss. Thanks!

requirements-dev.txt
ericspod: From [this](https://github.com/mu-editor/mu/issues/826) I think `nni; platform_system == "Linux" and "arm" not in platform_machine and "aarch" not in platform_machine` would work to exclude it.

KumoLiu: Thanks, works well!

Nic-Ma (approved):

PR #7509, opened 2024-02-29
Geometric transform -- Resize

KumoLiu:
Part of https://github.com/Project-MONAI/MONAI/issues/7486

### Description

Geometric transform -- Resize

### Types of changes

### DIFF
--- a/monai/transforms/spatial/functional.py
+++ b/monai/transforms/spatial/functional.py
@@ -265,8 +265,8 @@ def flip(img, sp_axes, lazy, transform_info):
     return out.copy_meta_from(meta_info) if isinstance(out, MetaTensor) else out
 
 
-def resize(
-    img, out_size, mode, align_corners, dtype, input_ndim, anti_aliasing, anti_aliasing_sigma, lazy, transform_info
+def resize_image(
+    img, out_size, dtype, input_ndim, lazy, transform_info, **kwargs
 ):
     """
     Functional implementation of resize.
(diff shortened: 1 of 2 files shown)


monai/transforms/spatial/functional.py
```diff
@@ -339,6 +363,80 @@ def resize(
...
+        points_affine = torch.matmul(affine_t[:spatial_dims, :spatial_dims], points_affine)
+        points_affine = points_affine.transpose(0, 1)
+
+    # convert tensor back to numpy if needed
+    points_affine, *_ = convert_to_dst_type(src=points_affine, dst=points)
+
+    return points_affine
+
+
+def resize_point(points, out_size, dtype, input_ndim, lazy, transform_info, **kwargs):
```
atbenmurray, line 403: How can we avoid having `*_point(...)` functions for each operation? At this level, shouldn't we be able to always just pass the correct transform? If we think about object space (raster space) and world space transforms then we shouldn't need to be performing these calculations based on spatial size here.

atbenmurray: As far as I understand it, resize should have no effect on point data, as it is entirely a "raster" space operation that preserves the extents of an image or volume, no?

atbenmurray: As discussed in the meeting, the alternative of taking the operation that defines the operation out of the function and putting it into its own implementation function works just as well as what I am suggesting.
```python
def resize_point(data, ...):
    meta_info = get_resize(...)
    out = apply_to_geom(data, metainfo)
    return out
```

PR #7530, opened 2024-03-11
Add Barlow Twins loss for representation learning

Lucas-rbnt:
### Description

Addition of the BarlowTwinsLoss class. This cost function is introduced in the http://proceedings.mlr.press/v139/zbontar21a/zbontar21a.pdf paper with the aim of disentangling the representations learned on two views of the same sample, making it a powerful tool for multimodal and unsupervised learning.
This cost function is similar to the InfoNCE Loss function already implemented in MONAI (https://docs.monai.io/en/latest/_modules/monai/losses/contrastive.html#ContrastiveLoss). However, it differs in several respects: there is no l2-normalisation, but rather a z-normalisation. In addition, rather than working between pairs of embeddings, Barlow Twins seeks to decorrelate the components of the representations.

```math
\mathcal{L}_{BT} := \sum_i (1 - \mathcal{C}_{ii})^2 + \lambda \sum_i \sum_{i\neq j} \mathcal{C}_{ij}^2
```
with $\lambda$ a positive hyperparameters and $\mathcal{C}$ the cross-correlation matrix

### Types of changes


monai/losses/barlow_twins.py
```diff
@@ -0,0 +1,83 @@
...
+
+from warnings import warn
+
+import torch
+from torch.nn.modules.loss import _Loss
+
+
+class BarlowTwinsLoss(_Loss):
+    """
+    Compute the Barlow Twins loss defined in:
```
ericspod, line 22: Thanks for the contribution! Could you please provide some explanation here in the docstring on what the loss is for, what it computes, what applications it would be for, example use case if there's something particular to demonstrate, etc.? It helps at a glance to know what this loss would be for without reading the paper.

tests/test_barlow_twins_loss.py
```diff
@@ -0,0 +1,109 @@
...
+        loss = BarlowTwinsLoss(lambd=5e-3)
+        i = torch.ones((2, 10))
+        j = torch.ones((2, 10))
+        if torch.cuda.is_available():
+            i = i.cuda()
+            j = j.cuda()
+        output = loss(i, j)
+        np.testing.assert_allclose(output.detach().cpu().numpy(), 10.0, atol=1e-4, rtol=1e-4)
+
+    def check_warning_rasied(self):
```
ericspod, line 103:
```suggestion
    def check_warning_raised(self):
```

tests/test_barlow_twins_loss.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @parameterized.expand(TEST_CASES)
+    def test_result(self, input_param, input_data, expected_val):
+        barlowtwinsloss = BarlowTwinsLoss(**input_param)
+        result = barlowtwinsloss(**input_data)
+        np.testing.assert_allclose(result.detach().cpu().numpy(), expected_val, atol=1e-4, rtol=1e-4)
+
+    def test_ill_shape(self):
+        loss = BarlowTwinsLoss(lambd=5e-3)
+        with self.assertRaisesRegex(ValueError, ""):
```
ericspod, line 85:
```suggestion
        with self.assertRaises(ValueError):
```
I think this works?

monai/losses/barlow_twins.py
Lucas-rbnt: Yes, of course!
I've modified the doctring to provide a little more explanation of the usefulness of this loss function.

tests/test_barlow_twins_loss.py
```diff
@@ -0,0 +1,109 @@
...
+        np.testing.assert_allclose(result.detach().cpu().numpy(), expected_val, atol=1e-4, rtol=1e-4)
+
+    def test_ill_shape(self):
+        loss = BarlowTwinsLoss(lambd=5e-3)
+        with self.assertRaises(ValueError):
+            loss(torch.ones((1, 2, 3)), torch.ones((1, 1, 2, 3)))
+
+    def test_ill_batch_size(self):
+        loss = BarlowTwinsLoss(lambd=5e-3, batch_size=1)
+        with self.assertRaisesRegex(ValueError, ""):
```
ericspod, line 90:
```suggestion
        with self.assertRaises(ValueError):
```
The same here.

monai/losses/barlow_twins.py
ericspod: Thanks for this. I think one of the key things here is that it's a loss on the cross-correlation matrix rather than on the images directly. I haven't used this before myself so I would benefit from an example of how this would be used in MONAI. If you're planning a tutorial notebook to show how to use this loss that would go there (I hope you're planning on that!), otherwise something added here to explain that technical detail would help.

I had made minor suggestions in the tests but otherwise it looks good to me.

Lucas-rbnt: This is similar to the contrastive cost function for pre-training encoders.
Yes, perhaps a short tutorial would be more helpful. If I'm not being silly, the MONAI tutorials are on another repo though, should I make a PR on this other repo with a tutorial?

Lucas-rbnt: Hello @ericspod. I have created a notebook tutorial for this loss self supervised as requested. If this PR is merged I will push a PR in the tutorial repo with the corresponding notebook.

Binder: https://mybinder.org/v2/gh/Lucas-rbnt/monai_tutorial/main
Directly on my repo: https://github.com/Lucas-rbnt/monai_tutorial/blob/main/ssl_barlow_twins.ipynb

ericspod: Looks good, thanks. When you do the PR to the tutorials repo we can look at the notebook in more depth, but I would suggest now that it should also have more description on what it's doing step-by-step. You should emphasise what's the main focus or concept you're demonstrating so it doesn't get lost in all the code for training and such. 

 I'm fine with this right now but @Nic-Ma @KumoLiu thoughts?

monai/losses/barlow_twins.py
```diff
@@ -0,0 +1,89 @@
...
+
+    Zbontar, Jure, et al. "Barlow Twins: Self-Supervised Learning via Redundancy Reduction" International
+    conference on machine learning. PMLR, 2020. (http://proceedings.mlr.press/v139/zbontar21a/zbontar21a.pdf)
+
+    Adapted from:
+        https://github.com/facebookresearch/barlowtwins
+
+    """
+
+    def __init__(self, lambd: float = 5e-3, batch_size: int = -1) -> None:
```
KumoLiu, line 38: Looks `batch_size` can be removed now?

monai/losses/barlow_twins.py
```diff
@@ -0,0 +1,89 @@
...
+            ValueError: When an input of dimension length > 2 is passed
+            ValueError: When input and target are of different shapes
+            ValueError: When batch size is less than or equal to 1
+
+        """
+        super().__init__()
+        self.lambd = lambd
+
+        if batch_size != -1:
+            warn("batch_size is no longer required to be set. It will be estimated dynamically in the forward call")
```
KumoLiu, line 53:
```suggestion
```

KumoLiu (approved):
Thanks for the PR! The overall work looks good to me. 
However, I have one comment regarding the `batch_size`. Please take a look when you get a chance

monai/losses/barlow_twins.py
KumoLiu: Agreed, we can discuss more about the notebook after you create a PR.

monai/losses/barlow_twins.py
Lucas-rbnt: Yes definitely, I took inspiration on the Contrastive Loss class where there was this batch size and warning process, but it doesn't need to be here, it's been corrected!

PR #7534, opened 2024-03-12
Update to use `log_sigmoid` in `FocalLoss`

### LINKED ISSUE #7533: AssertionError in test_focal_loss
```
FAIL: test_consistency_with_cross_entropy_2d_no_reduction (tests.test_focal_loss.TestFocalLoss)
For gamma=0 the focal loss reduces to the cross entropy loss
  File "/tmp/MONAI/tests/test_focal_loss.py", line 135, in test_consistency_with_cross_entropy_2d_no_reduction
    assert np.allclose(max_error, 0)
AssertionError
```

KumoLiu:
Fixes #7533

### Description

### Types of changes

### DIFF
--- a/monai/losses/focal_loss.py
+++ b/monai/losses/focal_loss.py
@@ -235,7 +235,7 @@ def sigmoid_focal_loss(
     # computing binary cross entropy with logits
     # equivalent to F.binary_cross_entropy_with_logits(input, target, reduction='none')
     # see also https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/Loss.cpp#L363
-    loss: torch.Tensor = input - input * target  - F.logsigmoid(input)
+    loss: torch.Tensor = input - input * target - F.logsigmoid(input)
 
     # sigmoid(-i) if t==1; sigmoid(i) if t==0 <=>
     # 1-sigmoid(i) if t==1; sigmoid(i) if t==0 <=>
--- a/monai/losses/focal_loss.py
+++ b/monai/losses/focal_loss.py
@@ -234,9 +234,8 @@ def sigmoid_focal_loss(
     """
     # computing binary cross entropy with logits
     # equivalent to F.binary_cross_entropy_with_logits(input, target, reduction='none')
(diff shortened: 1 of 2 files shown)


monai/losses/focal_loss.py
```diff
@@ -234,9 +234,8 @@ def sigmoid_focal_loss(
     """
     # computing binary cross entropy with logits
     # equivalent to F.binary_cross_entropy_with_logits(input, target, reduction='none')
-    # see also https://github.com/pytorch/pytorch/blob/v1.9.0/aten/src/ATen/native/Loss.cpp#L231
-    max_val = (-input).clamp(min=0)
-    loss: torch.Tensor = input - input * target + max_val + ((-max_val).exp() + (-input - max_val).exp()).log()
+    # see also https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/Loss.cpp#L363
+    loss: torch.Tensor = input - input * target - F.logsigmoid(input)
```
ericspod, line 238: Is this change just to speed up our code? It mirrors the PR in Pytorch you mentioned in the issue.

tests/test_focal_loss.py
```diff
@@ -132,7 +132,7 @@ def test_consistency_with_cross_entropy_2d_no_reduction(self):
             error = np.abs(a - b)
             max_error = np.maximum(error, max_error)
 
-        assert np.allclose(max_error, 0)
+        assert np.allclose(max_error, 0, atol=1e-6)
```
ericspod, line 135: This implies that the Pytorch `BCEWithLogitsLoss` is less accurate now? It might not matter during training but it's not a good thing anyway if this is what's happened.

ericspod (approved):
The changes look fine to me but I am concerned if the Pytorch is less accurate now somehow.

monai/losses/focal_loss.py
KumoLiu: To align the latest PyTorch, and when the latest PyTorch is used, the max_error will becomes zero which means our `FocalLoss` will get the same result with `BCEWithLogitsLoss` when gamma is zero.

tests/test_focal_loss.py
ericspod: Makes sense here now thanks!

Issue #7512, opened 2024-03-04
Implementation of Intensity Clipping Transform


KumoLiu:
Based on the implementation of the 'ScaleIntensityRangePercentiles' transform, we can add a new transform that only does intensity clipping.
https://github.com/Project-MONAI/MONAI/blob/e9e273858b6904659b0cd581d357662dc5c782f6/monai/transforms/intensity/array.py#L1205

jmnolte:
I'd be open to create a PR. However, I'd propose adding a soft clipping option as described in this [medium article](https://medium.com/life-at-hopper/clip-it-clip-it-good-1f1bf711b291). The advantage is that it preserves the natural ordering of the intensity values thus loosing less contextual information as in the hard clipped version.

KumoLiu:
Hi @jmnolte, your proposal looks good to me, and we'd be delighted to welcome your contribution. It would indeed be better to offer both soft and hard clipping options for added versatility.
For your convenience, here is the [contribution guide](https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#the-contribution-process).
Thanks.

PR #7535, opened 2024-03-12
Implementation of intensity clipping transform: bot hard and soft approaches

Lucas-rbnt:
Fixes Issue #7512.

### Description

Addition of a transformation allowing values above or below a certain percentile to be clipped.
Clipping can be hard or soft.
With soft clipping, the function remains derivable and the order of the values is respected, with smoother corners.

The soft clipping function is based on this medium article https://medium.com/life-at-hopper/clip-it-clip-it-good-1f1bf711b291

It's important to note that I've chosen to switch from Nones values to percentiles to take account of the fact that soft clipping can be one-sided or two-sided.
In fact, providing percentiles of 100 or 0 doesn't change anything in the case of hard clipping, but it does in the case of soft clipping because the function is smoothed. Hence the interest in introducing the possibility of putting None to avoid smoothing the function on one side or the other.

To implement this we had to define a `softplus` function in `monai.transforms.utils_pytorch_numpy_unification.py`. One of the problems is that `np.logaddexp` do not exactly yields same outputs as `torch.logaddexp`. I've left it as is and lowered the tolerance of the tests slightly, but it's possible to force the conversion to numpy and then switch back to torch to ensure better unification between the frameworks.

I've also added the `soft_clip` function in `monai.transforms.utils.py` with the associated unit tests to ensure that the transformation works properly.

### Types of changes


KumoLiu (approved):

tests/test_clip_intensity_percentiles.py
```diff
@@ -0,0 +1,115 @@
...
+import numpy as np
+from parameterized import parameterized
+
+from monai.transforms import ClipIntensityPercentiles
+from monai.transforms.utils import soft_clip
+from monai.transforms.utils_pytorch_numpy_unification import clip
+from tests.utils import TEST_NDARRAYS, NumpyImageTestCase2D, assert_allclose
+
+
+class TestClipIntensityPercentiles(NumpyImageTestCase2D):
```
dongyang0122, line 24: Please add at least one test example with 3D arrays. Since a lot of medical images have dimensions higher than two. Thanks.

Lucas-rbnt: Hello, of course, I added 3D tests in the new commit

monai/transforms/intensity/array.py
```diff
@@ -1007,6 +1008,118 @@ def __call__(self, img: NdarrayOrTensor) -> NdarrayOrTensor:
...
+
+    backend = [TransformBackends.TORCH, TransformBackends.NUMPY]
+
+    def __init__(
+        self,
+        lower: float | None,
+        upper: float | None,
+        sharpness_factor: float | None = None,
+        channel_wise: bool = False,
+        dtype: DtypeLike = np.float32,
```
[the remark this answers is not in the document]
Lucas-rbnt: Hi, okay, what would be the best way to proceed so as not to interfere too much with the MONAI workflow, by returning the image and a dictionary containing the values? Or by adding the values in the image's meta information?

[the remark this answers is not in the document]
Lucas-rbnt: Because I looked at similar classes that computes percentiles as well such as `ScaleIntensityRangePercentiles` to be consistent but do not provide the values

tests/test_clip_intensity_percentiles.py
dongyang0122: Thanks.

monai/transforms/intensity/array.py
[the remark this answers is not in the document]
dongyang0122: Adding it to meta data would be great if that was easy to implement. I agree that it is unnecessary to directly return additional values from the transform.

monai/transforms/intensity/array.py
```diff
@@ -1007,6 +1008,139 @@ def __call__(self, img: NdarrayOrTensor) -> NdarrayOrTensor:
...
+        img = convert_to_tensor(img, track_meta=get_track_meta())
+        img_t = convert_to_tensor(img, track_meta=False)
+        if self.channel_wise:
+            img_t = torch.stack([self._clip(img=d) for d in img_t])  # type: ignore
+        else:
+            img_t = self._clip(img=img_t)
+
+        img = convert_to_dst_type(img_t, dst=img)[0]
+        if self.return_percentiles:
+            img.meta["percentiles"] = self.percentiles
```
dongyang0122, line 1139: How about changing "percentiles" to "clipping_values" to make the naming clearer in meta tensor?

Lucas-rbnt: Done! You're right, it's clearer that way.

What do you think of the structure and type of these clipping values? As a list of 2d-tuple containing either a NdarrayOrTensor or a None? That's the part I'm a bit dubious about. As far as I'm concerned it's better that way but it's also much less manipulable and perhaps not very clear for users.

dongyang0122 (approved):

monai/transforms/intensity/array.py
dongyang0122: The data structure looks good to me. You can add a few words in the docstrings to explain `None`. And we can unify the data type to "float". Since these values are individual numbers, floats should support numpy arrays or torch tensors for potential use. What do you think?

Lucas-rbnt: Okay, as it is currently done, this returns a list of tuples containing either a float or a None.
so the corresponding typing should be something like: ```List[Tuple[Optional[float], ...]]```

monai/transforms/intensity/array.py
```diff
@@ -1007,6 +1008,139 @@ def __call__(self, img: NdarrayOrTensor) -> NdarrayOrTensor:
...
+                https://medium.com/life-at-hopper/clip-it-clip-it-good-1f1bf711b291.
+            upper: upper intensity percentile.  The same as for lower, but this time with the highest values. If we
+                are looking to perform soft clipping, if None then there will be no effect on this side whereas if set
+                to 100, the values will be passed via the corresponding clipping equation.
+            sharpness_factor: if not None, the intensity values will be soft clipped according to
+                f(x) = x + (1/sharpness_factor)*softplus(- c(x - minv)) - (1/sharpness_factor)*softplus(c(x - maxv)).
+                defaults to None.
+            channel_wise: if True, compute intensity percentile and normalize every channel separately.
+                default to False.
+            return_percentiles: whether to return the calculated percentiles in tensor meta information,
```
dongyang0122, line 1084: Duplicated definition of `return_percentiles`.

Lucas-rbnt: Fixed! I've also changed the name of `return_percentiles` to `return_clipping_values` to enhance clarification

monai/transforms/intensity/array.py
Lucas-rbnt: I've added typing and instead of returning  0-dim tensor or array I chose to keep only the float value via the `.item()`. Does that seem better to you?

dongyang0122: Does it work for numpy arrays?

Lucas-rbnt: Yes, `item()` can also be applied to 0-d numpy arrays but here with the `conversion_to_tensor` if it is a numpy at first it will be converted to a `torch.Tensor`

dongyang0122: I see.

PR #7549, opened 2024-03-15
7540 change bundle workflow args

### LINKED ISSUE #7540: Add args in `BundleWorkflow`
According to my MONAI model zoo experiences, I think `meta_file` and `logging_file` are two necessary args. 
As for `meta_file`, it helps a lot to define the requirements of a bundle, and track each time's change.
As for `logging_file`, it helps to define the logging format, and if we have multiple bundles, it's easier to control the logging data.

These two args are now working well in `ConfigWorkflow`, but are not required when we define a class that extends to  `BundleWorkflow` directly.

yiheng-wang-nv:
Fixes #7540  .

### Description

This PR:
1. add logging file and meta file into BundleWorkflow
2. add the sequence form of meta files check for ConfigWorkflow

### Types of changes


monai/bundle/workflows.py
```diff
@@ -268,6 +293,13 @@ def __init__(
                 f"Cannot find the metadata config file: {meta_file}. "
                 "Please see: https://docs.monai.io/en/stable/mb_specification.html"
             )
+        elif isinstance(meta_file, list):
```
ericspod, line 296: Should we use `self.meta_file` instead? To do you would move `meta_file=str(self.config_root_path / "metadata.json") if meta_file is None else meta_file` to be just before the `super()` call and then pass `meta_file=meta_file` as a new argument. I think the check for files existing here wouldn't be needed in this case.

monai/bundle/workflows.py
```diff
@@ -74,6 +83,22 @@ def __init__(self, workflow_type: str | None = None, workflow: str | None = None
         else:
             raise ValueError(f"Unsupported workflow type: '{workflow_type}'.")
 
+        if logging_file is not None:
+            if not os.path.exists(logging_file):
```
ericspod, line 87: I would suggest `isfile` in case a directory was passed instead.

monai/bundle/workflows.py
```diff
@@ -74,6 +83,22 @@ def __init__(self, workflow_type: str | None = None, workflow: str | None = None
...
+                raise FileNotFoundError(f"Cannot find the logging config file: {logging_file}.")
+            logger.info(f"Setting logging properties based on config: {logging_file}.")
+            fileConfig(logging_file, disable_existing_loggers=False)
+
+        if meta_file is not None:
+            if isinstance(meta_file, str) and not os.path.exists(meta_file):
```
ericspod, line 93: Use `isfile` here too?

monai/bundle/workflows.py
```diff
@@ -74,6 +83,22 @@ def __init__(self, workflow_type: str | None = None, workflow: str | None = None
...
+                raise FileNotFoundError(f"Cannot find the metadata config file: {meta_file}.")
+            if isinstance(meta_file, list):
+                for f in meta_file:
+                    if not os.path.exists(f):
```
ericspod, line 97: Use `isfile` here too?

monai/bundle/workflows.py
yiheng-wang-nv: Hi @ericspod . Thanks for the suggestions. As for `isfile`, I've already updated the code.
However, it seems we cannot do `meta_file=str(self.config_root_path / "metadata.json") if meta_file is None else meta_file` before `super()`.
In order to do that, we also have to produce `self.config_root_path` first, but it would lead to an error in `__getattr__`.
What do you think?

ericspod: The whole `if config_file is not None:...` block could possibly be moved before the `super()` as well. It would set a local variable `config_root_path` (instead of `self.config_root_path`) which the meta file path can be constructed from. Then after `super()` have `self.config_root_path = config_root_path`.  This should work but may be getting messy now...

ericspod: Looks good to me now!

ericspod (approved):
Looks good to me.

KumoLiu:
For the CI error, may add something like this:
https://github.com/Project-MONAI/MONAI/blob/bca11a54fe4631abd2dd1a259a0778a10161be75/monai/bundle/workflows.py#L171

PR #7552, opened 2024-03-15
monai generative: refactor autoencoderkl

marksgraham:
Part of the refactoring in #7227 

### Description
Refactors autoencoderkl. Changes are:

- Introduce `CastToTempType` class for upsampling
- `Downsample` block removed and replaced by a `Sequential`
- The attention block now uses MONAI's `SABlock`, allowing a lot of code to be removed
- Added a `load_old_state_dict` that allows for models trained on MONAI Generative to be loaded in to this model, especially important given some of the MONAI Generative's [model zoo](https://github.com/Project-MONAI/GenerativeModels/tree/main/model-zoo) uses this model. I have tested this works locally.

I discussed with @ericspod inheriting from `AutoEncoder` but after experimentation have decided against it as it introduced changes that made it very hard to ensure we could load model's trained in MONAI Generative.

### Types of changes


marksgraham:
Notes as I go:
Leaving the `pad` argument out of the downsample convolution doesn't give the expected behaviour. It claims `Return the padding value needed to ensure a convolution using the given kernel size produces an output of the same
    shape as the input for a stride of 1, otherwise ensure a shape of the input divided by the stride rounded down.` but I get the size of the input rounded up, e.g.:

```
block(torch.zeros((2,128,65,65)).to('cuda')).shape
torch.Size([2, 128, 33, 33])
```

Nic-Ma:
Hi @marksgraham , @ericspod ,

I imagine this PR is refactoring the Generative AI branch to integrate into dev branch for MONAI 1.4 as we discussed?

Thanks.

marksgraham:
@Nic-Ma  Correct - this will be one of a series of PRs as part of that refactor

monai/networks/nets/autoencoderkl.py
```diff
@@ -320,6 +321,38 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            rearrange_input = Rearrange("b c h w d -> b (h w d) c")
+            rearrange_output = Rearrange("b (h w d) c -> b c h w d", h=h, w=w, d=d)
+
+        x = self.norm(x)
+        x = rearrange_input(x)  # B x (s_dim_1 * ... * s_dim_n) x C
+
+        x = self.attn(x)
+        x = rearrange_output(x)
+        x = x + residual
+        return x
```
ericspod, line 353:
```suggestion
class _AttentionBlockNew(SkipConnection):
    def __init__(self, spatial_dims, num_channels, num_head_channels):
        from monai.networks.blocks.selfattention import SABlock
        from einops.layers.torch import Rearrange

        _,_,h,w,*d=x.shape
        if not d:
            rearrange_input = Rearrange("b c h w -> b (h w) c")
            rearrange_output = Rearrange("b (h w) c -> b c h w", h=h, w=w)
        else:
            rearrange_input = Rearrange("b c h w d -> b (h w d) c")
            rearrange_output = Rearrange("b (h w d) c -> b c h w d", h=h, w=w, d=d[0])
                
        norm = nn.GroupNorm(num_groups=32, num_channels=num_head_channels, eps=1e-6, affine=True)
        attn = SABlock(hidden_size=num_channels, num_heads=1, qkv_bias=True)
        super().__init__(Sequential(norm,rearrange_input,attn,rearrange_output),mode="add")
```

monai/networks/nets/autoencoderkl.py
ericspod, line 1: One idea was to wrap the AMP-fixing code into its own class since it crops up in various places:

```python
class CastTempType(nn.Module):
    def __init__(self,initial_type, temporary_type, submodule):
        super().__init__()
        self.initial_type=initial_type
        self.temporary_type=temporary_type
        self.submodule=submodule

    def forward(self, x):
        dtype = x.dtype
        if dtype == self.initial_type:
            x = x.to(self.temporary_type)
        x = submodule(x)
        if dtype == self.initial_type:
            x = x.to(self.initial_type)
        return x
```

monai/networks/nets/autoencoderkl.py
marksgraham: on further thought, this depends on x, which isn't available at init. this means we will need to put this in the forward, so i don't think we save much code inheriting the skip connection here

ericspod: Ah yes, we'd have to define our own block to use `Rearrange` for a given input shape. If we do that in multiple places this would work but if not then yeah we leave this as it is.

marksgraham:
I have now made use of MONAI's own `Upsample` block rather than implemented a new one, with a few changes to that block needed. The `Downsample` block here isn't very general, I haven't seen it used in other generative networks, so does not seem worth spinning out into its own file.

monai/networks/nets/autoencoderkl.py
```diff
@@ -454,6 +328,7 @@ def __init__(
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         for block in self.blocks:
+            # print(f'{block}: {x[0, 0, 0, 0]}')
```
KumoLiu, line 331:
```suggestion
```

monai/networks/nets/autoencoderkl.py
```diff
@@ -805,3 +694,63 @@ def encode_stage_2_inputs(self, x: torch.Tensor) -> torch.Tensor:
     def decode_stage_2_outputs(self, z: torch.Tensor) -> torch.Tensor:
         image = self.decode(z)
         return image
+
+    def load_old_state_dict(self, old_state_dict: dict, verbose=False) -> None:
+        """
+        Load a state dict from an AutoencoderKL trained with MONAI Generative.
```
KumoLiu, line 700:
```suggestion
        Load a state dict from an AutoencoderKL trained with [MONAI Generative](https://github.com/Project-MONAI/GenerativeModels).
```

monai/networks/nets/autoencoderkl.py
```diff
@@ -569,19 +441,45 @@ def __init__(
...
+                            in_channels=block_in_ch,
+                            out_channels=block_in_ch,
+                            interp_mode="nearest",
+                            scale_factor=2.0,
+                            post_conv=post_conv,
+                            align_corners=None,
+                        )
+                    )
+                    # rename postconv for compatibility with monai-generative
+                    blocks[-1].__dict__["_modules"]["conv"] = blocks[-1].__dict__["_modules"].pop("postconv")
```
KumoLiu, line 482: I think it's better to do the compatibility in the `load_old_state_dict`.  We can load the weight of the `conv` from old model to the `postconv` in the new model instead of modifying the module name. What do you think?
cc @ericspod

monai/networks/nets/spade_autoencoderkl.py
```diff
@@ -241,17 +238,38 @@ def __init__(
...
+                        in_channels=block_in_ch,
+                        out_channels=block_in_ch,
+                        interp_mode="nearest",
+                        scale_factor=2.0,
+                        post_conv=post_conv,
+                        align_corners=None,
+                    )
+                )
+                # rename postconv for compatibility with monai-generative
+                blocks[-1].__dict__["_modules"]["conv"] = blocks[-1].__dict__["_modules"].pop("postconv")
```
KumoLiu, line 272: Same as above.

monai/networks/nets/autoencoderkl.py
```diff
@@ -622,7 +520,6 @@ class AutoencoderKL(nn.Module):
         norm_eps: epsilon for the normalization.
         with_encoder_nonlocal_attn: if True use non-local attention block in the encoder.
         with_decoder_nonlocal_attn: if True use non-local attention block in the decoder.
-        use_flash_attention: if True, use flash attention for a memory efficient attention mechanism.
```
KumoLiu, line 625: May I know remove this will not affect anything here, right?

tests/test_autoencoderkl.py
```diff
@@ -18,11 +18,14 @@
...
 from monai.networks.nets import AutoencoderKL
+from monai.utils import optional_import
 from tests.utils import SkipIfBeforePyTorchVersion
 
+tqdm, has_tqdm = optional_import("tqdm", name="tqdm")
+einops, has_einops = optional_import("einops")
 device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
 
-CASES = [
+CASES_NO_ATTENTION = [
```
KumoLiu, line 28: Can we add a 3d case for this no attention test?

tests/test_spade_autoencoderkl.py
```diff
@@ -12,14 +12,18 @@
...
 from parameterized import parameterized
 
 from monai.networks import eval_mode
 from monai.networks.nets import SPADEAutoencoderKL
+from monai.utils import optional_import
 
-CASES = [
+einops, has_einops = optional_import("einops")
+
+CASES_NO_ATTENTION = [
```
KumoLiu, line 26: Also here.

KumoLiu (commented):
Thanks for the PR. Overall looks good to me.
Several comments inline. Not much seems to need changing, so perhaps we can do the next one. After @ericspod returns, if there are no comments, I will try to merge this one. Thanks.

monai/networks/nets/autoencoderkl.py
marksgraham: `flash_attention` was just a faster version of attention but exactly the same results. the implementation here  never worked that well, we never used it, and it will make sense to use pytorch's inbuilt version in future so i think it is fine to remove. because it was never used i don't think it will affect anything if we remove it

KumoLiu: Got it.

KumoLiu:
Will set DCO pass for this pr: https://github.com/Project-MONAI/MONAI/pull/7552/checks?check_run_id=24109528540

Do you have any concern? @Nic-Ma @ericspod @marksgraham

KumoLiu:
And @marksgraham, could you please take a look at the ci error, looks like many tests failed. Thanks!

marksgraham:
@KumoLiu fixed the CI, shall i merge in dev to get your fix for the code format error?

KumoLiu:
Hi @ericspod, do you have any other comment on this pr otherwise I think we can merge this one?

BTW, next time let we merge dev in a seperate PR which could avoid too much files changed.

Thanks!

ericspod:
@KumoLiu I had discussed things with @marksgraham and we've concluded that whatever refactors we can do are great but there's a lot we can't without breaking compatibility. I'm confident this code is good in general since we've been using it within the other repo for a while now. 

We definitely need a cleaner approach to syncing with dev in these feature branches since 810 changes files is a lot! I hope this doesn't pollute the git history when we go to merge this feature branch.

Thanks both!

ericspod (approved):

PR #7555, opened 2024-03-15
Update setup.py

tejhande:
Variable Naming: Renamed RUN_BUILD to BUILD_MONAI for clarity. Extension Module Handling: Moved the logic for defining extension modules (ext_modules) inside the conditional block, ensuring that they are only defined when BUILD_MONAI is enabled. ImportError Handling: Added handling for ImportError when importing Torch, providing a warning message if Torch is not available. Package Data Inclusion: Included all files in monai/csrc directory in the package_data section to ensure that all relevant files are included in the package distribution. These changes enhance the clarity, correctness, and robustness of the setup script, ensuring that it functions correctly and efficiently for building the MONAI framework.

Fixes # .

### Description

### Types of changes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,13 +10,9 @@
 # limitations under the License.
 
 import os
-import re
-import sys
 import warnings
 
-import pkg_resources
 from setuptools import find_packages, setup
-from setuptools.extension import Extension
 
 import versioneer
 
@@ -29,7 +25,7 @@ ext_modules = []
 if BUILD_MONAI:
     try:
         import torch
-        from torch.utils.cpp_extension import BuildExtension, CUDAExtension
+        from torch.utils.cpp_extension import CUDAExtension
 
         print(f"Setup with torch {torch.__version__}")
 
@@ -54,4 +50,3 @@ setup(
     package_data={"monai": ["py.typed", "monai/csrc/*"]},
     ext_modules=ext_modules,
 )
-
--- a/setup.py
+++ b/setup.py
@@ -9,9 +9,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from __future__ import annotations
-
-import glob
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -9,143 +9,44 @@
...
-try:
-    import torch
-
-    print(f"setup.py with torch {torch.__version__}")
-    from torch.utils.cpp_extension import BuildExtension, CppExtension
-
-    BUILD_CPP = True
-    from torch.utils.cpp_extension import CUDA_HOME, CUDAExtension
-
-    BUILD_CUDA = FORCE_CUDA or (torch.cuda.is_available() and (CUDA_HOME is not None))
```
ericspod, line 41: Removing this variable significantly changes the logic of what's going on here. If CUDA isn't present we don't want to try building against it, this is why some of the CICD tests fail. Below we've used this variable to choose whether to add the CUDAExtension object or not.

setup.py
```diff
@@ -9,143 +9,44 @@
...
-except (ImportError, TypeError, AssertionError, AttributeError) as e:
-    warnings.warn(f"extension build skipped: {e}")
-finally:
-    if not RUN_BUILD:
-        BUILD_CPP = BUILD_CUDA = False
-        print("Please set environment variable `BUILD_MONAI=1` to enable Cpp/CUDA extension build.")
-    print(f"BUILD_MONAI_CPP={BUILD_CPP}, BUILD_MONAI_CUDA={BUILD_CUDA}, TORCH_VERSION={TORCH_VERSION}.")
-
-
-def torch_parallel_backend():
```
ericspod, line 56: This is needed for choosing OMP compiler flags and macros farther down, I don't see this being done in your version and I think we do need them. Further down in the original `get_extensions()` uses this and `omp_flags()` to set compiler flags and the matrco for the backend type.

setup.py
```diff
@@ -9,143 +9,44 @@
...
-            define_macros=define_macros,
-            extra_compile_args=extra_compile_args,
-            extra_link_args=extra_link_args,
+        import torch
+        from torch.utils.cpp_extension import CUDAExtension
+
+        print(f"Setup with torch {torch.__version__}")
+
+        ext_modules.append(
+            CUDAExtension(
```
ericspod, line 33: As mentioned above, if CUDA isn't present this will fail. We do want to be able to build without CUDA to support environments without acceleration but which can still reasonably do inference, eg. macOS.

setup.py
```diff
@@ -9,143 +9,44 @@
...
+                sources=["monai/csrc/*.cpp", "monai/csrc/*.cu"],
+                include_dirs=["monai/csrc"],
+                define_macros=[("WITH_CUDA", None)],
+            )
         )
-    ]
-    return ext_modules
-
-
-def get_cmds():
```
ericspod, line 125: With the choice to build without CUDA, we need to update the commands being returned by `get_cmdclass()` so I think this is necessary.

ericspod (changes requested):
Hi @tejhande thanks for the contribution! I think however this PR takes out a number of things that we do need in the setup file. We do want to be able to build without CUDA but your version I don't believe would allow this. You can see that a number of the CICD checks fail when we try to do this. I'm all for cleaning up this file and making it cleaner but we don't want to lose this functionality. If you want to update your PR I can review it again, there are a few details on the compilation process I may need to get further feedback on from others as well.

PR #7564, opened 2024-03-21
Fixing gradient in sincos positional encoding in monai/networks/blocks/patchembedding.py

Lucas-rbnt:
### Description

When you choose to put the argument `pos_embed_type='sincos'` in the `PatchEmbeddingBlock` class, it still return a learnable positional encoding

To reproduce:
```python
from monai.networks.blocks import PatchEmbeddingBlock
patcher = PatchEmbeddingBlock(
            in_channels=1,
            img_size=(32, 32, 32),
            patch_size=(8, 8, 8),
            hidden_size=96,
            num_heads=8,
            pos_embed_type="sincos",
            dropout_rate=0.5,
        )
print(patcher.position_embeddings.requires_grad) 
>>> True
```

In the literature, we sometimes use either positional encoding in sincos which are fixed and non-trainable as in the original Attention Is All You Need [paper](https://arxiv.org/abs/1706.03762) or a learnable positional embedding as in the ViT [paper](https://arxiv.org/abs/2010.11929).
If you choose to use a sincos, then it seems that is must be fixed which is not the case here.
I'm not completely sure of the desired result in MONAI since there's already a learnable possibility, so if we choose sincos we'd like gradient-free parameters. However the documentation of `build_sincos_position_embedding`in the `pos_embed_utils.py`files stipulate: "The sin-cos position embedding as a learnable parameter" which seems a bit confusing. Especially as the encoding construction function seems to aim to set the require gradient to False (see below)

```python
pos_embed = nn.Parameter(pos_emb)
pos_embed.requires_grad = False

return pos_embed
```
But these changes are not maintained by torch's `copy_` function, which does not copy gradient parameters (see the cpp code https://github.com/pytorch/pytorch/blob/148a8de6397be6e4b4ca1508b03b82d117bfb03c/torch/csrc/lazy/ts_backend/tensor_aten_ops.cpp#L51).
This `copy_`is used in the `PatchEmbeddingBlock` class to instantiate the positional embedding.

I propose a small fix to overcome this problem as well as test cases to ensure that positional embedding behaves correctly.  

### Types of changes


monai/networks/blocks/patchembedding.py
```diff
@@ -123,6 +123,7 @@ def __init__(
             with torch.no_grad():
                 pos_embeddings = build_sincos_position_embedding(grid_size, hidden_size, spatial_dims)
```
marksgraham, line 124: i know this isn't part of the PR but wondering why we don't just do
`self.position_embeddings = build_sincos_position_embedding(grid_size, hidden_size, spatial_dims).float()
` here. @ericspod  any ideas?

marksgraham (approved):

marksgraham:
LGTM, i agree the sincos embeddings should be fixed, it doesn't make sense to have them learnable.

monai/networks/blocks/patchembedding.py
KumoLiu: Agreed, looks like we can directly using the output from `build_sincos_position_embedding`.
https://github.com/Project-MONAI/MONAI/blob/c3a7383494b35f82bc3193797a3b07629e94c0c0/monai/networks/blocks/pos_embed_utils.py#L104

Lucas-rbnt: Yes, that would work, I can commit a more concise and efficient version of the positional encoding creation if ever, the idea of this PR in the first instance was simply to have the desired behavior (a fixed encoding) with a small fix.

ericspod: Setting `requires_grad = False` is the key thing here, being created within a no-grad context won't set it to False. Within the context there won't be a computation of gradients but we wouldn't have this anyway.

Lucas-rbnt: Yes, exactly, but here we're also trying to set the positional encoding `requires_grad=False` in `monai/networks/blocks/pos_embed_utils.py`in the code l103/l104

```python
pos_embed = nn.Parameter(pos_emb)
pos_embed.requires_grad = False
```

But the requires grad attributes is not maintained with the torch `copy_` function

KumoLiu:
Merged now, feel free to create another PR for https://github.com/Project-MONAI/MONAI/pull/7564#discussion_r1535660514.

PR #7568, opened 2024-03-22
MedicalNetPerceptualSimilarity: Add multi-channel

### LINKED ISSUE #7567: Perceptual Loss with mednet 3D multiple channels
**Describe the bug**
Perceptual loss with 3D volumes and the medicalnet network types currently works with one channel only, as MedNet3D takes single channel 3D volumes as input. However the code in MedicalNetPerceptualSimilarity suggests that it is intended to work with multiple channels.

**To Reproduce**
Use PerceptualLoss with spatial_dims=3, is_fake_3d=False,  network_type='medicalnet_resnet50_23datasets' and  3D volumes with multiple channels, e.g. with fMRI, phase-cycle bSSFP, DTI, ... as data.

**Expected behavior**
MedicalNetPerceptualSimilarity passes the input channelwise and averages (the latter is already implemented).

**Environment**
<details>
  <summary>MONAI environment</summary>
  
Printing MONAI config...
MONAI version: 1.3.0
Numpy version: 1.26.4
Pytorch version: 2.2.1+cu121
MONAI rev id: 865972f7a791bf7b42efbcd87c8402bd865b329e
MONAI __file__: /home/<username>/workspace/UNet-bSSFP/unet-venv/lib/python3.11/site-packages/monai/__init__.py

Optional dependencies:
Nibabel version: 5.2.0
scikit-image version: NOT INSTALLED or UNKNOWN VERSION.
scipy version: 1.12.0
Pillow version: 10.2.0
Tensorboard version: 2.16.2
gdown version: 4.7.3
TorchVision version: 0.17.1+cu121
tqdm version: 4.66.2
psutil version: 5.9.8
pandas version: 2.0.3
einops version: NOT INSTALLED or UNKNOWN VERSION.
mlflow version: NOT INSTALLED or UNKNOWN VERSION.
pynrrd version: NOT INSTALLED or UNKNOWN VERSION.
clearml version: NOT INSTALLED or UNKNOWN VERSION.

Printing system config...
System: Linux
Linux version: Arch Linux
Platform: Linux-6.7.4-zen1-1-zen-x86_64-with-glibc2.39
Processor: 
Machine: x86_64
Python version: 3.11.7
Process name: python
Open files: []
Num physical CPUs: 24
Num logical CPUs: 32
Num usable CPUs: 32
CPU usage (%): [93.7, 62.3, 79.0, 40.4, 61.9, 78.7, 56.0, 98.4, 96.5, 19.7, 23.0, 97.6, 96.4, 99.2, 99.6, 30.3, 99.6, 100.0, 100.0, 100.0, 100.0, 100.0, 100.0, 100.0, 100.0, 100.0, 100.0, 100.0, 100.0, 100.0, 100.0, 100.0]
CPU freq. (MHz): 3461
Load avg. in last 1, 5, 15 mins (%): [87.6, 86.7, 82.9]
Disk usage (%): 44.8
Total physical memory (GB): 62.6
Available memory (GB): 25.2
Used memory (GB): 32.0

Printing GPU config...
Num GPUs: 1
Has CUDA: True
CUDA version: 12.1
cuDNN enabled: True
NVIDIA_TF32_OVERRIDE: None
cuDNN version: 8902
Current device: 0
GPU 0 Name: NVIDIA GeForce RTX 4090 Laptop GPU
GPU 0 Is integrated: False
GPU 0 Multi processor count: 76
GPU 0 Total memory (GB): 15.7
GPU 0 CUDA capability (maj.min): 8.9
</details>

**Additional context**

SomeUserName1:
Fixes #7567 .

### Description
 MedicalNetPerceptualSimilarity: Add multi-channel support for 3Dvolumes. 
The current version of the code in the dev branch already largely supports that besides the following:
medicalnet_* require inputs to have a single channel. 
This PR passes the multi-channel volume channel-wise to the networks and concatenates the resulting feature vectors. 
The existing code takes care of averaging over channels and spatially.

### Types of changes

### DIFF
--- a/monai/losses/perceptual.py
+++ b/monai/losses/perceptual.py
@@ -74,6 +74,7 @@ class PerceptualLoss(nn.Module):
         pretrained: bool = True,
         pretrained_path: str | None = None,
         pretrained_state_dict_key: str | None = None,
+        channelwise: bool = False,
     ):
         super().__init__()
 
@@ -102,15 +103,18 @@ class PerceptualLoss(nn.Module):
         self.spatial_dims = spatial_dims
         self.perceptual_function: nn.Module
         if spatial_dims == 3 and is_fake_3d is False:
-            self.perceptual_function = MedicalNetPerceptualSimilarity(net=network_type, verbose=False)
+            self.perceptual_function = MedicalNetPerceptualSimilarity(net=network_type, verbose=False,
+                                                                      channelwise=channelwise)
         elif "radimagenet_" in network_type:
-            self.perceptual_function = RadImageNetPerceptualSimilarity(net=network_type, verbose=False)
+            self.perceptual_function = RadImageNetPerceptualSimilarity(net=network_type, verbose=False,
+                                                                       channelwise=channelwise)
         elif network_type == "resnet50":
             self.perceptual_function = TorchvisionModelPerceptualSimilarity(
                 net=network_type,
                 pretrained=pretrained,
                 pretrained_path=pretrained_path,
                 pretrained_state_dict_key=pretrained_state_dict_key,
+                channelwise=channelwise,
             )
         else:
             self.perceptual_function = LPIPS(pretrained=pretrained, net=network_type, verbose=False)
@@ -185,13 +189,20 @@ class MedicalNetPerceptualSimilarity(nn.Module):
         net: {``"medicalnet_resnet10_23datasets"``, ``"medicalnet_resnet50_23datasets"``}
             Specifies the network architecture to use. Defaults to ``"medicalnet_resnet10_23datasets"``.
         verbose: if false, mute messages from torch Hub load function.
+        channelwise: if True, the loss is returned per channel. Otherwise the loss is averaged over the channels.
+                Defaults to ``False``.
     """
 
-    def __init__(self, net: str = "medicalnet_resnet10_23datasets", verbose: bool = False) -> None:
+    def __init__(self,
+                 net: str = "medicalnet_resnet10_23datasets",
+                 verbose: bool = False,
+                 channelwise: bool = False) -> None:
         super().__init__()
         torch.hub._validate_not_a_forked_repo = lambda a, b, c: True
         self.model = torch.hub.load("warvito/MedicalNet-models", model=net, verbose=verbose)
         self.eval()
+
+        self.channelwise = channelwise
 
         for param in self.parameters():
             param.requires_grad = False
@@ -206,6 +217,7 @@ class MedicalNetPerceptualSimilarity(nn.Module):
         Args:
             input: 3D input tensor with shape BCDHW.
             target: 3D target tensor with shape BCDHW.
+
         """
         input = medicalnet_intensity_normalisation(input)
         target = medicalnet_intensity_normalisation(target)
@@ -227,7 +239,10 @@ class MedicalNetPerceptualSimilarity(nn.Module):
         feats_target = normalize_tensor(outs_target)
 
         results: torch.Tensor = (feats_input - feats_target) ** 2
-        results = spatial_average_3d(results.sum(dim=1, keepdim=True), keepdim=True)
+
+        if self.channelwise:
+            results = results.sum(dim=1, keepdim=True)
+        results = spatial_average_3d(results, keepdim=True)
 
         return results
 
@@ -260,10 +275,12 @@ class RadImageNetPerceptualSimilarity(nn.Module):
         verbose: if false, mute messages from torch Hub load function.
     """
 
-    def __init__(self, net: str = "radimagenet_resnet50", verbose: bool = False) -> None:
+    def __init__(self, net: str = "radimagenet_resnet50", verbose: bool = False, channelwise: bool = False) -> None:
         super().__init__()
         self.model = torch.hub.load("Warvito/radimagenet-models", model=net, verbose=verbose)
         self.eval()
+
+        self.channelwise = channelwise
 
         for param in self.parameters():
             param.requires_grad = False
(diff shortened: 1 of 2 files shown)


KumoLiu:
The implementation appears to be done correctly and it successfully introduces multi-channel input support for PerceptualSimilarity. It effectively allows each channel to be processed independently, which is a substantial addition.
However, I'd like to suggest the consideration of a specific use-case scenario. The new implementation seems to be most useful under situations where independent channel processing is beneficial. Conversely, in the final computation stage, results.sum(dim=1, keepdim=True) consolidates the multi-channel perception into a single scalar evaluation. This approach works well under single modality situations.
For cases involving multi-modal inputs, it might be worth considering the separate evaluation of each modality. As a suggestion, perhaps we could add an optional flag that switches between combined evaluation and individual modality evaluation. This could potentially enhance the generalization capabilities of the implementation and provide users the flexibility to choose based on their specific use case dynamics.

SomeUserName1:
Will adjust accordingly on monday, such that the return types are scalar or 1D array/tensor respectively.

SomeUserName1:
For the case of LPIPS I'm not entirely sure what to do. 
There are two relevant flags, I think: `spatial` and `lpips`

Here are the relevant lines from the forward function of LPIPS:
`        if(self.lpips):
            if(self.spatial):
                res = [upsample(self.lins[kk](diffs[kk]), out_HW=in0.shape[2:]) for kk in range(self.L)]
            else:
                res = [spatial_average(self.lins[kk](diffs[kk]), keepdim=True) for kk in range(self.L)]
        else:
            if(self.spatial):
                res = [upsample(diffs[kk].sum(dim=1,keepdim=True), out_HW=in0.shape[2:]) for kk in range(self.L)]
            else:
                res = [spatial_average(diffs[kk].sum(dim=1,keepdim=True), keepdim=True) for kk in range(self.L)]`

`diffs` is the squared difference of the features.

`spatial` seems to control if spatial averaging is done in dims 2 and 3 so this is not the one we are looking for.
`lpips` controls if summing over dim 1 (i.e. the channels) is done. However the docstring states:
`lpips : bool
            [True] use linear layers on top of base/trunk network
            [False] means no linear layers; each layer is averaged together`

`lpips` false seems to do the averaging without the top layers. True does no averaging but with the input layers. 
From our perspective i think averaging should be independent on other layers, that would mean (a) another PR on lpips would be necessary to support channel-wise averaging independent of the top layers  or (b) an extra elif in the forward function for LPIPS, which does the forward pass channelwise and then either sums or returns the vector. 
@KumoLiu If i understand everything correctly, option (b) would not require chanegs upstream in lpips while (a) would keep the code in MONAI a but cleaner.
Wdyt?

SomeUserName1:
@KumoLiu only the blossom-ci check hasn't passed. Can we merge?

monai/losses/perceptual.py
```diff
@@ -170,9 +176,9 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
             loss = loss_sagittal + loss_axial + loss_coronal
         else:
             # 2D and real 3D cases
-            loss = self.perceptual_function(input, target)
+            loss = self.perceptual_function(input, target).squeeze()
 
-        return torch.mean(loss)
+        return torch.mean(loss, dim=0)
```
KumoLiu, line 181: May I know the reason for these change? The squeeze here is for? And the loss will alway keep the channel dim?

monai/losses/perceptual.py
```diff
@@ -74,6 +74,7 @@ def __init__(
         pretrained: bool = True,
         pretrained_path: str | None = None,
         pretrained_state_dict_key: str | None = None,
+        channelwise: bool = False,
```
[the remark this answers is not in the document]
KumoLiu: Please also update the docstring, thanks.

KumoLiu (commented):
Thanks for the quick update, few comments inline.

tests/test_perceptual_loss.py
```diff
@@ -40,11 +40,26 @@
         (2, 1, 64, 64, 64),
         (2, 1, 64, 64, 64),
     ],
+    [
+        {"spatial_dims": 3, "network_type": "medicalnet_resnet10_23datasets", "is_fake_3d": False},
+        (2, 6, 64, 64, 64),
+        (2, 6, 64, 64, 64),
+    ],
+    [
+        {"spatial_dims": 3, "network_type": "medicalnet_resnet10_23datasets", "is_fake_3d": False, "channelwise": True},
```
KumoLiu, line 49: Could you please review all the instances where "channelwise" is used? It appears there may be a misnaming issue, even with the wrong name, it doesn't affect the passing of the CI tests.

monai/losses/perceptual.py
KumoLiu: Hi @SomeUserName1, could you please help explain more here?
Seems the result will always be [1*C*1]？

tags
KumoLiu, line 1: Added by mistake?

KumoLiu:
Hi @SomeUserName1, I noticed that you add me review the PR again.

In case you didn't noticed the comment here. Looks a little confuse to me here. Could you please help explain it more?
https://github.com/Project-MONAI/MONAI/pull/7568#discussion_r1553071925
Thanks.

monai/losses/perceptual.py
SomeUserName1: Ah sorry, I just figured, that pending means I have to submit a review in files changed...

SomeUserName1: Done

monai/losses/perceptual.py
```diff
@@ -172,7 +181,13 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
             # 2D and real 3D cases
             loss = self.perceptual_function(input, target)
 
-        return torch.mean(loss)
+        if self.channel_wise:
+            loss = loss.squeeze()
+            loss = torch.mean(loss, dim=0)
```
KumoLiu, line 186:
```suggestion
            loss = torch.mean(loss.squeeze(), dim=0)
```

KumoLiu (approved):
Thanks for the quick update, overall looks good to me.
Only a small comment, will merge this ASAP.

PR #7569, opened 2024-03-24
Remove nested error propagation on `ConfigComponent` instantiate

### LINKED ISSUE #7451: [MONAI Bundle] Very long errors during config instantiation - Clean up raised errors
When there is an error in the monai bundle config during instantiation, and the config is large, the `RuntimeError` ends up being extremely long. 
This is counterproductive for debugging and pollutes the terminal with uninformative error messages.

The root cause of this looks to be the `RuntimeError` messages generated here:
- https://github.com/Project-MONAI/MONAI/blob/ff430286c37e78d7592372a5a97377f0cbb0219c/monai/bundle/config_item.py#L295
- https://github.com/Project-MONAI/MONAI/blob/ff430286c37e78d7592372a5a97377f0cbb0219c/monai/utils/module.py#L274

**To Reproduce**
Run any relatively large configuration with a wrongly configured argument, and this will occur. 
Attached is an example config and the error log below. 

The command run for the bundle to run is 
`python -m monai.bundle run system --config_file config.yaml`

[error_log.txt](https://github.com/Project-MONAI/MONAI/files/14202650/error_log.txt)
[config.json](https://github.com/Project-MONAI/MONAI/files/14202657/config.json) [YAML renamed to JSON as github doesn't allow uploading YAML]

A clear and concise error message would be more informative. 

I have the following suggestions: 
1. Remove the `RuntimeError` raised in `ConfigComponent` as the instantiate call will already raise it. Remove this basically: https://github.com/Project-MONAI/MONAI/blob/ff430286c37e78d7592372a5a97377f0cbb0219c/monai/bundle/config_item.py#L295
2. It would be ideal if the RuntimeError raised in https://github.com/Project-MONAI/MONAI/blob/ff430286c37e78d7592372a5a97377f0cbb0219c/monai/utils/module.py#L274 shows only the `__path` and not the full `kwargs` by default. Maybe this could be moved into debug mode instead?

surajpaib:
Fixes #7451

### Description
Reduces the length of error messages and error messages being propagated twice. This helps debug better when long `ConfigComponent`s are being instantiated. Refer to issue #7451 for more details

### Types of changes

### DIFF
--- a/monai/bundle/config_item.py
+++ b/monai/bundle/config_item.py
@@ -289,10 +289,7 @@ class ConfigComponent(ConfigItem, Instantiable):
         mode = self.get_config().get("_mode_", CompInitMode.DEFAULT)
         args = self.resolve_args()
         args.update(kwargs)
-        try:
-            return instantiate(modname, mode, **args)
-        except Exception as e:
-            raise RuntimeError(f"Failed to instantiate {self}") from e
+        return instantiate(modname, mode, **args)
 
 
 class ConfigExpression(ConfigItem):
--- a/monai/utils/module.py
+++ b/monai/utils/module.py
@@ -272,7 +272,7 @@ def instantiate(__path: str, __mode: str, **kwargs: Any) -> Any:
             return pdb.runcall(component, **kwargs)
     except Exception as e:
         raise RuntimeError(
-            f"Failed to instantiate component '{__path}' with kwargs: {kwargs}"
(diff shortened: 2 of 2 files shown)


surajpaib:
@KumoLiu

ericspod (approved):
I think this change is okay to make, it'll be different now for people of course but I hope no one was depending on the output contents we're not showing now.

PR #7574, opened 2024-03-25
Feature added: Get WSI at mpp

### LINKED ISSUE #4980: `WSIReader` support for read by magnification
In MONAI Pathology WG, we have discussed about the need for being able to read whole slide images at specific magnification (mpp or magnification) in addition to reading at a specific level. The feature request has been made by @cooperlab in cucim repo https://github.com/rapidsai/cucim/issues/389.

I have discussed with cuCIM team about this topic and the conclusion was that it may be more useful to implement this functionality in MONA rather than in cucim for few reasons: 
* this feature will be automatically supported for all the backends (cuCIM, OpenSlide, TiffFile, etc)
* the changes in magnification can be tracked in MONAI and be used to invert the process

To achieve this, the following functionalities need to be added to MOANI `WSIReader`: 
- to find the next higher level to the requested mpp or magnification
- load the requested image that level with the appropriate pre-resized size
- resample/resize the image to the target mpp/magnification

NikolasSchmitz:
Fixes: #4980 

### Description

In this pull request, the feature to retrieve a whole slide image at a given mpp (microns per pixel) resolution was implemented for every WSIReader class in the function `get_wsi_at_mpp`. 

While the implementations in the `OpenslideWSIReader` and `CuCIMWSIReader` classes were tested thoroughly, I could not find a suitable TIFF file for testing with the `TiffFileWSIReader` class.

For resizing, I have used `PIL.Image.resize` for Openslide and TiffFile, and `cucim.sklearn.transform.resize` for CuCIM. Originally, I used `cv2.resize`, but since the package isn't listed in requirements-dev.txt, I explored alternative solutions."

### Types of changes

Please let me know what you think and how I can improve this feature.

Best
Niko

(Last PR was closed, because I changed the branch name to include the ticket id)

### DIFF
--- a/monai/data/wsi_reader.py
+++ b/monai/data/wsi_reader.py
@@ -433,6 +433,70 @@ class BaseWSIReader(ImageReader):
                 metadata[key] = [m[key] for m in metadata_list]
         return _stack_images(patch_list, metadata), metadata
 
+    def _compute_mpp_target_res(self, closest_lvl, closest_lvl_dim, mpp_list, mpp: tuple):
+        """
+        Computes the target dimensions for resizing a whole slide image
+        to match a user-specified resolution in microns per pixel (MPP).
+
+        Args:
+            closest_lvl: Whole slide image level closest to user-provided MPP resolution.
+            closest_lvl_dim: Dimensions (height, width) of the image at the closest level.
+            mpp_list: List of MPP values for all levels of the whole slide image.
+            mpp: The MPP resolution at which the whole slide image representation should be extracted.
+
+        Returns:
+            Tuple of (target_res_x, target_res_y) representing the target pixel dimensions.
+
+        """
+        mpp_closest_lvl = mpp_list[closest_lvl]
+        mpp_closest_lvl_x, mpp_closest_lvl_y = mpp_closest_lvl
+
+        ds_factor_x = mpp_closest_lvl_x / mpp[0]
+        ds_factor_y = mpp_closest_lvl_y / mpp[1]
+
+        target_res_x = int(np.round(closest_lvl_dim[1] * ds_factor_x))
+        target_res_y = int(np.round(closest_lvl_dim[0] * ds_factor_y))
+
+        return target_res_x, target_res_y
+
+    def _compute_mpp_tolerances(self, closest_lvl, mpp_list, mpp, atol, rtol) -> tuple[bool, bool]:
+        """
+        Determines if user-provided MPP values are within a specified tolerance of the closest
+        level's MPP and checks if the closest level has higher resolution than desired MPP.
+
+        Args:
+            closest_lvl: Whole slide image level closest to user-provided MPP resolution.
+            mpp_list: List of MPP values for all levels of the whole slide image.
+            mpp: The MPP resolution at which the whole slide image representation should be extracted.
+            atol: Absolute tolerance for MPP comparison.
+            rtol: Relative tolerance for MPP comparison.
+
+        Returns:
+            Tuple of (is_within_tolerance, closest_level_is_bigger) where first element indicates
+            if MPP is within tolerance and second indicates if closest level has higher resolution.
+
+        """
+        user_mpp_x, user_mpp_y = mpp
+        mpp_closest_lvl_x, mpp_closest_lvl_y = mpp_list[closest_lvl]
+
+        # Define tolerance intervals for x and y of closest level
+        lower_bound_x = mpp_closest_lvl_x * (1 - rtol) - atol
+        upper_bound_x = mpp_closest_lvl_x * (1 + rtol) + atol
+        lower_bound_y = mpp_closest_lvl_y * (1 - rtol) - atol
+        upper_bound_y = mpp_closest_lvl_y * (1 + rtol) + atol
+
+        # Check if user-provided mpp_x and mpp_y fall within the tolerance intervals for closest level
+        is_within_tolerance_x = (user_mpp_x >= lower_bound_x) and (user_mpp_x <= upper_bound_x)
+        is_within_tolerance_y = (user_mpp_y >= lower_bound_y) and (user_mpp_y <= upper_bound_y)
+        is_within_tolerance = is_within_tolerance_x and is_within_tolerance_y
+
+        # If mpp_closest_level < mpp -> closest_level has higher res than img at mpp => downscale from closest_level to mpp
+        closest_level_is_bigger_x = mpp_closest_lvl_x < user_mpp_x
+        closest_level_is_bigger_y = mpp_closest_lvl_y < user_mpp_y
+        closest_level_is_bigger = closest_level_is_bigger_x and closest_level_is_bigger_y
+
+        return is_within_tolerance, closest_level_is_bigger
+
     def verify_suffix(self, filename: Sequence[PathLike] | PathLike) -> bool:
         """
         Verify whether the specified file or files format is supported by WSI reader.
@@ -605,6 +669,31 @@ class WSIReader(BaseWSIReader):
         """
         return self.reader.get_mpp(wsi, level)
 
+    def get_wsi_at_mpp(
+        self, wsi, mpp: float | tuple[float, float], atol: float = 0.00, rtol: float = 0.05
+    ) -> np.ndarray:
+        """
+        Returns the representation of the whole slide image at a given micro-per-pixel (mpp) resolution.
+        The optional tolerance parameters are considered at the level whose mpp value is closest to the one provided by the user.
+        If the user-provided mpp is larger than the mpp of the closest level,
+        the image is downscaled to a resolution that matches the user-provided mpp.
+        Otherwise, if the closest level's resolution is not sufficient to meet the user's requested resolution,
+        the next lower level (which has a higher resolution) is chosen.
+        The image from this level is then down-scaled to achieve a resolution at the user-provided mpp value.
+
+        Args:
+            wsi: whole slide image object from WSIReader
+            mpp: the resolution in micron per pixel at which the representation of the whole slide image should be extracted.
+            atol: the acceptable absolute tolerance for resolution in micro per pixel.
+            rtol: the acceptable relative tolerance for resolution in micro per pixel.
+
+        Returns:
+            Numpy array containing the whole slide image at the requested MPP resolution.
+
+        """
+        mpp = ensure_tuple_rep(mpp, 2)
+        return self.reader.get_wsi_at_mpp(wsi, mpp, atol, rtol)
+
     def get_power(self, wsi, level: int) -> float:
         """
         Returns the micro-per-pixel resolution of the whole slide image at a given level.
@@ -746,6 +835,75 @@ class CuCIMWSIReader(BaseWSIReader):
 
         raise ValueError("`mpp` cannot be obtained for this file. Please use `level` instead.")
 
+    def get_wsi_at_mpp(self, wsi, mpp: float | tuple[float, float], atol: float = 0.00, rtol: float = 0.05) -> Any:
+        """
+        Returns the representation of the whole slide image at a given micro-per-pixel (mpp) resolution.
+        The optional tolerance parameters are considered at the level whose mpp value is closest to the one provided by the user.
+        If the user-provided mpp is larger than the mpp of the closest level,
+        the image is downscaled to a resolution that matches the user-provided mpp.
+        Otherwise, if the closest level's resolution is not sufficient to meet the user's requested resolution,
+        the next lower level (which has a higher resolution) is chosen.
+        The image from this level is then down-scaled to achieve a resolution at the user-provided mpp value.
+
+        Args:
+            wsi: whole slide image object from WSIReader
+            mpp: the resolution in micron per pixel at which the representation of the whole slide image should be extracted.
+            atol: the acceptable absolute tolerance for resolution in micro per pixel.
+            rtol: the acceptable relative tolerance for resolution in micro per pixel.
+
+        Returns:
+            Cupy array containing the whole slide image at the requested MPP resolution.
+
+        """
+        cp, _ = optional_import("cupy")
+        mpp = ensure_tuple_rep(mpp, 2)
+
+        mpp_list = [self.get_mpp(wsi, lvl) for lvl in range(wsi.resolutions["level_count"])]
+        closest_lvl = self._find_closest_level("mpp", mpp, mpp_list, 0, 5)
+
+        within_tolerance, closest_level_is_bigger = self._compute_mpp_tolerances(closest_lvl, mpp_list, mpp, atol, rtol)
+
+        if within_tolerance:
+            # If the image at the desired mpp resolution is within tolerances, return the image at closest_level.
+            closest_lvl_wsi = wsi.read_region(
+                (0, 0),
+                level=closest_lvl,
+                size=wsi.resolutions["level_dimensions"][closest_lvl],
+                num_workers=self.num_workers,
+            )
+
+        elif closest_level_is_bigger:
+            # Otherwise, select the level closest to the desired mpp with a higher resolution and downsample it.
+            closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+
+        else:
+            # If both checks fail, increase resolution (i.e., decrement level) and then downsample it.
+            if closest_lvl == 0:
+                closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+            else:
+                closest_lvl = closest_lvl - 1
+                closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+
+        wsi_arr = cp.asnumpy(closest_lvl_wsi)
+
+        # Ensure channel dimension exists for grayscale
+        if wsi_arr.ndim < 3:
+            wsi_arr = wsi_arr[..., None]
+
+        # Apply channel dimension
+        wsi_arr = np.moveaxis(wsi_arr, -1, self.channel_dim)
+
+        # Apply mode (RGB / RGBA)
+        if self.mode == "RGB":
+            if wsi_arr.shape[self.channel_dim] not in [3, 4]:
+                raise ValueError(
+                    f"The image is expected to have three or four color channels in '{self.mode}' mode but has "
+                    f"{wsi_arr.shape[self.channel_dim]}. "
+                )
+            wsi_arr = np.take(wsi_arr, [0, 1, 2], self.channel_dim)
+
+        return wsi_arr
+
     def get_power(self, wsi, level: int) -> float:
         """
         Returns the objective power of the whole slide image at a given level.
@@ -829,6 +987,36 @@ class CuCIMWSIReader(BaseWSIReader):
             patch = np.take(patch, [0, 1, 2], self.channel_dim)
 
         return patch
+
+    def _resize_to_mpp_res(self, wsi, closest_lvl, mpp_list, user_mpp: tuple):
+        """
+        Resizes the whole slide image to the specified resolution in microns per pixel (mpp).
+
+        Args:
+            wsi: whole slide image object from WSIReader
+            user_mpp: the resolution in microns per pixel at which the whole slide image representation should be extracted.
+            closest_lvl: the wsi level that is closest to the user-provided mpp resolution.
+            mpp_list: list of mpp values for all levels of a whole slide image.
+
+        Returns:
+            Resized cupy image array at the target MPP resolution.
+
+        """
+        cucim_resize, _ = optional_import("cucim.skimage.transform", name="resize")
+        cp, _ = optional_import("cupy")
+
+        closest_lvl_dim = wsi.resolutions["level_dimensions"][closest_lvl]
+
+        target_res_x, target_res_y = self._compute_mpp_target_res(closest_lvl, closest_lvl_dim, mpp_list, user_mpp)
+
+        wsi_arr = cp.array(
+            wsi.read_region((0, 0), level=closest_lvl, size=closest_lvl_dim, num_workers=self.num_workers)
+        )
+        closest_lvl_wsi = cucim_resize(
+            wsi_arr, (target_res_x, target_res_y), order=1, preserve_range=True, anti_aliasing=False
+        ).astype(cp.uint8)
+
+        return closest_lvl_wsi
 
 
 @require_pkg(pkg_name="openslide")
@@ -942,6 +1130,66 @@ class OpenSlideWSIReader(BaseWSIReader):
 
         raise ValueError("`mpp` cannot be obtained for this file. Please use `level` instead.")
 
+    def get_wsi_at_mpp(
+        self, wsi, mpp: float | tuple[float, float], atol: float = 0.00, rtol: float = 0.05
+    ) -> np.ndarray:
+        """
+        Returns the representation of the whole slide image at a given micro-per-pixel (mpp) resolution.
+        The optional tolerance parameters are considered at the level whose mpp value is closest to the one provided by the user.
+        If the user-provided mpp is larger than the mpp of the closest level,
+        the image is downscaled to a resolution that matches the user-provided mpp.
+        Otherwise, if the closest level's resolution is not sufficient to meet the user's requested resolution,
+        the next lower level (which has a higher resolution) is chosen.
+        The image from this level is then down-scaled to achieve a resolution at the user-provided mpp value.
+
+        Args:
+            wsi: whole slide image object from WSIReader
+            mpp: the resolution in micron per pixel at which the representation of the whole slide image should be extracted.
+            atol: the acceptable absolute tolerance for resolution in micro per pixel.
+            rtol: the acceptable relative tolerance for resolution in micro per pixel.
+
+        Returns:
+            Numpy array containing the whole slide image at the requested MPP resolution.
+
+        """
+        mpp = ensure_tuple_rep(mpp, 2)
+
+        mpp_list = [self.get_mpp(wsi, lvl) for lvl in range(wsi.level_count)]
+        closest_lvl = self._find_closest_level("mpp", mpp, mpp_list, 0, 5)
+
+        within_tolerance, closest_level_is_bigger = self._compute_mpp_tolerances(closest_lvl, mpp_list, mpp, atol, rtol)
+
+        if within_tolerance:
+            # If the image at the desired mpp resolution is within tolerances, return the image at closest_level.
+            closest_lvl_wsi = wsi.read_region((0, 0), level=closest_lvl, size=wsi.level_dimensions[closest_lvl])
+
+        elif closest_level_is_bigger:
+            # Otherwise, select the level closest to the desired mpp with a higher resolution and downsample it.
+            closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+
+        else:
+            # If both checks fail, increase resolution (i.e., decrement level) and then downsample it.
+            if closest_lvl == 0:
+                closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+            else:
+                closest_lvl = closest_lvl - 1
+                closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+
+        # Convert to specified mode (e.g., RGBA to RGB)
+        closest_lvl_wsi = closest_lvl_wsi.convert(self.mode)
+
+        wsi_arr = np.array(closest_lvl_wsi)
+
+        # Ensure channel dimension exists for grayscale (defensive: PIL.convert
+        # to "L" produces a 2-D array without a trailing channel axis).
+        if wsi_arr.ndim < 3:
+            wsi_arr = wsi_arr[..., None]
+
+        # Apply channel dimension
+        wsi_arr = np.moveaxis(wsi_arr, -1, self.channel_dim)
+
+        return wsi_arr
+
     def get_power(self, wsi, level: int) -> float:
         """
         Returns the objective power of the whole slide image at a given level.
@@ -1011,6 +1259,31 @@ class OpenSlideWSIReader(BaseWSIReader):
         patch = np.moveaxis(patch, -1, self.channel_dim)
 
         return patch
+
+    def _resize_to_mpp_res(self, wsi, closest_lvl, mpp_list, user_mpp: tuple):
+        """
+        Resizes the whole slide image to the specified resolution in microns per pixel (mpp).
+
+        Args:
+            wsi: whole slide image object from WSIReader
+            user_mpp: the resolution in microns per pixel at which the whole slide image representation should be extracted.
+            closest_lvl: the wsi level that is closest to the user-provided mpp resolution.
+            mpp_list: list of mpp values for all levels of a whole slide image.
+
+        Returns:
+            PIL Image object resized to the target MPP resolution.
+
+        """
+        pil_image, _ = optional_import("PIL", name="Image")
+
+        closest_lvl_dim = wsi.level_dimensions[closest_lvl]
+
+        target_res_x, target_res_y = self._compute_mpp_target_res(closest_lvl, closest_lvl_dim, mpp_list, user_mpp)
+
+        closest_lvl_wsi = wsi.read_region((0, 0), level=closest_lvl, size=closest_lvl_dim)
+        closest_lvl_wsi = closest_lvl_wsi.resize((target_res_y, target_res_x), pil_image.BILINEAR)  # row, col order
+
+        return closest_lvl_wsi
 
 
 @require_pkg(pkg_name="tifffile")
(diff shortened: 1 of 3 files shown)


NikolasSchmitz:
Thank you @drbeh ! I now marked it as ready for review.

monai/data/wsi_reader.py
```diff
@@ -744,6 +763,101 @@ def get_mpp(self, wsi, level: int) -> tuple[float, float]:
...
+        upper_bound_y = mpp_closest_lvl_y * (1 + rtol) + atol
+
+        # Check if user-provided mpp_x and mpp_y fall within the tolerance intervals for closest level
+        within_tolerance_x = (user_mpp_x >= lower_bound_x) & (user_mpp_x <= upper_bound_x)
+        within_tolerance_y = (user_mpp_y >= lower_bound_y) & (user_mpp_y <= upper_bound_y)
+        within_tolerance = within_tolerance_x & within_tolerance_y
+
+        if within_tolerance:
+            # Take closest_level and continue with returning img at level
+            print(f"User-provided MPP lies within tolerance of level {closest_lvl}, returning wsi at this level.")
```
ericspod, line 811: Print statements generally shouldn't be in production code, if you want to keep track of things you can log this instead. This goes for all the other print statements below.

monai/data/wsi_reader.py
```diff
@@ -940,6 +1054,91 @@ def get_mpp(self, wsi, level: int) -> tuple[float, float]:
 
         raise ValueError("`mpp` cannot be obtained for this file. Please use `level` instead.")
 
+    def get_wsi_at_mpp(self, wsi, mpp: tuple, atol: float = 0.00, rtol: float = 0.05) -> np.ndarray:
```
ericspod, line 1057: This code appears to duplicate a lot of that in the other `get_wsi_at_mpp` method. Can this be factored out into a separate function to eliminate the duplication?

ericspod:
@drbeh would you be able to review this? I think you're best qualified here. I made some minor comments about print and code duplication but otherwise it seems fine to me without having tested it. Thanks!

monai/data/wsi_reader.py
NikolasSchmitz: Hi Eric, thanks for feedback!

monai/data/wsi_reader.py
```diff
@@ -744,6 +763,101 @@ def get_mpp(self, wsi, level: int) -> tuple[float, float]:
...
+        # -> Should not throw ValueError, instead just return the closest value; how to select tolerances?
+
+        mpp_closest_lvl = mpp_list[closest_lvl]
+        closest_lvl_dim = wsi.resolutions["level_dimensions"][closest_lvl]
+
+        print(f"Closest Level: {closest_lvl} with MPP: {mpp_closest_lvl}")
+        mpp_closest_lvl_x, mpp_closest_lvl_y = mpp_closest_lvl
+
+        # Define tolerance intervals for x and y of closest level
+        lower_bound_x = mpp_closest_lvl_x * (1 - rtol) - atol
```
JHancox, line 799: I wonder whether the annotation ought to clarify the interaction between the absolute and relative tolerances. For example, someone might understandably assume that if either of the two values are exceeded then the value is deemed out of bounds, as opposed to the proposed behaviour which evaluates their combined tolerance.

monai/data/wsi_reader.py
```diff
@@ -744,6 +763,101 @@ def get_mpp(self, wsi, level: int) -> tuple[float, float]:
...
+                closest_lvl_dim = wsi.resolutions["level_dimensions"][closest_lvl]
+                closest_lvl_wsi = wsi.read_region(
+                    (0, 0), level=closest_lvl, size=closest_lvl_dim, num_workers=self.num_workers
+                )
+                wsi_arr = cp.array(closest_lvl_wsi)
+
+                target_res_x = int(np.round(closest_lvl_dim[1] * ds_factor_x))
+                target_res_y = int(np.round(closest_lvl_dim[0] * ds_factor_y))
+
+                closest_lvl_wsi = cucim_resize(wsi_arr, (target_res_x, target_res_y), order=0)
```
JHancox, line 855: I think order=0 equates to nearest neighbour interpolation. The other methods seem to be using bilinear interpolation (order=1). If so, this would seem to be preferable. An alternative would be to have an optional parameter - possibly overkill.

JHancox (commented):
Thanks for the effort @NikolasSchmitz. A welcome addition! I agree with @ericspod about the refactoring to remove duplication - that would be great.

NikolasSchmitz:
Thank you for the feedback, @JHancox . Unfortunately, I couldn't join the meeting because I was on my way home from ICLR. 
I will review the code and provide my updates as soon as possible.

monai/data/wsi_reader.py
NikolasSchmitz: We could clarify the interaction of both tolerances in the function's docstrings. Additionally, to prevent the two values from exceeding a certain limit, we could implement a validation check that raises a ValueError if either value is out of bounds. I will implement it, and we can review and decide later if we want to adopt this approach.

monai/data/wsi_reader.py
NikolasSchmitz: Ah yes, you're right. That should be order=1 here. Thanks!
If wanted, I could keep bilinear interpolation as default, and add an optional parameter to enable the user to overwrite the default. Potentially, the user might also want to decide on which resize method to use (currently `cucim.skimage.transform.resize` and `PIL.Image.resize`), which could be another optional parameter.

monai/data/wsi_reader.py
NikolasSchmitz: Hi Eric, I factored out the actual resizing of the wsi to a given mpp resolution into a function called _resize_to_mpp_res. The tolerance checks still happen in the main function get_wsi_at_mpp.

NikolasSchmitz:
I factored out the actual resizing of the image into _resize_to_mpp_res() and kept the tolerance checks in get_wsi_at_mpp(). I had some problems with specific tiff files and the TiffFile backend. Some files had zero as X/Y Resolution, which led to a division by zero error in the function get_mpp (added a check for that in the new commit). 
One tiff file from a Philips scanner that I tested had unobtainable mpp values using get_mpp. In this case, the tags do exist (openslide can find them), but they must be read and parsed from the property philips_metadata as XML. I have not implemented a fix for this yet, but please let me know if this is needed. In this particular case, the best option would be to use Openslide as the backend, although there might be users who prefer Tifffile.

monai/data/wsi_reader.py
ericspod: I'm looking at this method vs the one implemented in `CuCIMWSIReader` and it seems identical (or virtually so) except for the last line before return. What I'd suggest is that this method be moved to `BaseWSIReader` and then the implementation here and in `CuCIMWSIReader` calls it and converts its output to a numpy or cupy array as appropriate.

monai/data/wsi_reader.py
```diff
@@ -827,6 +911,36 @@ def _get_patch(
             patch = np.take(patch, [0, 1, 2], self.channel_dim)
 
         return patch
+    
+    def _resize_to_mpp_res(self, wsi, closest_lvl, mpp_list, user_mpp: tuple):
```
ericspod, line 915: This is definitely better to reduce duplication, however this and the other method of the same name below are virtually identical except for some type conversions. I think this method could be moved to `BaseWSIReader` and then here we'd have an override method calling the base one which does the conversion of outputs:

```python
def _resize_to_mpp_res(self, wsi, closest_lvl, mpp_list, user_mpp: tuple):
  cucim_resize, _ = optional_import("cucim.skimage.transform", name="resize")
  cp, _ = optional_import("cupy")

  closest_lvl_wsi, target_res_x, target_res_y = super()._resize_to_mpp_res(wsi, closest_lvl, mpp_list, user_mpp)
  wsi_arr = cp.array(closest_lvl_wsi)
  closest_lvl_wsi = cucim_resize(wsi_arr, (target_res_x, target_res_y), order=1)

  return closest_lvl_wsi
```

The `_resize_to_mpp_res` version in `BaseWSIReader` wouldn't have any Numpy/Cupy code in it but just implement the base algorithm and return the result plus the target resolution values. The version in `OpenSlideWSIReader` would use PIL to do the resizing instead of Cupy. 

Does this make sense to do?

ericspod:
There are lines added/removed to two other files that don't need to be changed in this PR, tests/test_regularization.py is causing a conflict for example. I would remove those and we can use black here through Github to resolve formatting issues if we need to.

NikolasSchmitz:
> There are lines added/removed to two other files that don't need to be changed in this PR, tests/test_regularization.py is causing a conflict for example. I would remove those and we can use black here through Github to resolve formatting issues if we need to.

Thanks for the info. Interesting though, haven't noticed that these files were changed. Perhaps through one of the coding style checks.

monai/data/wsi_reader.py
NikolasSchmitz: That sounds like a good solution to me. However, I would keep the computation of `closest_lvl_dim` and `closest_lvl_wsi` in the class specific function, and use `super()._resize_to_mpp_res` to compute `target_res_x` and `target_res_y`.

ericspod: Sure, feel free to update things.

NikolasSchmitz: Hi Eric, I have added a function to compute the target resolution to the BaseWSIReader class. Based on your idea, _resize_to_mpp_res then calls super()._compute_target_res, which returns target_x and target_y. I would suggest to keep the image extraction methods (read_region etc.) in _resize_to_mpp_res since they depend on the backend. Feel free to check out the result. Though, I still have to modify the docstring of _compute_target_res.

monai/data/wsi_reader.py
NikolasSchmitz: Based on the idea to factor out the function to compute the target resolution, we could also factor out the tolerance checks into a BaseWSIReader function. That would again reduce redundancy, since they are indeed identical.

monai/data/wsi_reader.py
ericspod: This looks much better with much less redundancy, thanks!

monai/data/wsi_reader.py
ericspod: If you could please do this that would greatly improve things, thanks.

ericspod:
There's a few changes to do but it looks good in general. We do need tests for these new methods as well so please look at the existing WSI tests to see what needs to be added to cover new functionality. The flake8 and DCO fixes can be done at the end. Thanks!

monai/data/wsi_reader.py
NikolasSchmitz: Hi Eric, this is now implemented. There is now a function _compute_mpp_tolerances to check tolerances in BaseWSIReader which is called in every reader's get_wsi_at_mpp function. I think the latter is now minimized as much as possible. Still need to finish the test functions though.

monai/data/wsi_reader.py
```diff
@@ -744,6 +820,48 @@ def get_mpp(self, wsi, level: int) -> tuple[float, float]:
...
+            atol: the acceptable absolute tolerance for resolution in micro per pixel.
+            rtol: the acceptable relative tolerance for resolution in micro per pixel.
+
+        """
+        cp, _ = optional_import("cupy")
+
+        mpp_list = [self.get_mpp(wsi, lvl) for lvl in range(wsi.resolutions["level_count"])]
+        closest_lvl = self._find_closest_level("mpp", mpp, mpp_list, 0, 5)
+
+        within_tolerance, closest_level_is_bigger = super()._compute_mpp_tolerances(closest_lvl, mpp_list, mpp, atol, rtol)
```
ericspod, line 845:
```suggestion
        within_tolerance, closest_level_is_bigger = self._compute_mpp_tolerances(closest_lvl, mpp_list, mpp, atol, rtol)
```
If you're not overriding the method in this class you can refer to it through `self` as it will be inherited.

ericspod: Same in other places below.

monai/data/wsi_reader.py
ericspod: Looks good now thanks. If you're still working on tests I don't need to review it all quite yet.

NikolasSchmitz:
> There's a few changes to do but it looks good in general. We do need tests for these new methods as well so please look at the existing WSI tests to see what needs to be added to cover new functionality. The flake8 and DCO fixes can be done at the end. Thanks!

Hi Eric, 
I’ve written a few tests to validate the core functionality of get_wsi_at_mpp. However, I’ve encountered an issue with the metadata in the TIFF image used for testing. It appears the MPP value is incorrect or missing. The metadata includes:

'tiff.ResolutionUnit': 'centimeter', 'tiff.XResolution': '10', 'tiff.YResolution': '10'

Here’s a screenshot of the TIFF file opened in QuPath:
image

For comparison, here’s a screenshot of the SVS file opened in QuPath:
image

It seems the SVS file was converted to TIFF, but during the conversion process, the resolution metadata got altered or lost. This discrepancy is likely affecting the accuracy of the tests.

Let me know how you’d like to proceed—should we regenerate the test image with proper metadata or use a different TIFF file for testing? I’ll continue refining the tests in the meantime.

Thanks!

ericspod:
> I’ve written a few tests to validate the core functionality of get_wsi_at_mpp. However, I’ve encountered an issue with the metadata in the TIFF image used for testing. It appears the MPP value is incorrect or missing. The metadata includes:

I'm also seeing some very different information in the two header from these files I've included below. I don't know how the tiff was generated but maybe @bhashemian would and know what's the discrepancy is.

<details>

In [11]: dict(openslide.open_slide("CMU-1.tiff").properties)
Out[11]:
{'openslide.level-count': '9',
 'openslide.level[0].downsample': '1',
 'openslide.level[0].height': '32914',
 'openslide.level[0].tile-height': '256',
 'openslide.level[0].tile-width': '256',
 'openslide.level[0].width': '46000',
 'openslide.level[1].downsample': '2',
 'openslide.level[1].height': '16457',
 'openslide.level[1].tile-height': '256',
 'openslide.level[1].tile-width': '256',
 'openslide.level[1].width': '23000',
 'openslide.level[2].downsample': '4.0001215362177929',
 'openslide.level[2].height': '8228',
 'openslide.level[2].tile-height': '256',
 'openslide.level[2].tile-width': '256',
 'openslide.level[2].width': '11500',
 'openslide.level[3].downsample': '8.0002430724355857',
 'openslide.level[3].height': '4114',
 'openslide.level[3].tile-height': '256',
 'openslide.level[3].tile-width': '256',
 'openslide.level[3].width': '5750',
 'openslide.level[4].downsample': '16.000486144871171',
 'openslide.level[4].height': '2057',
 'openslide.level[4].tile-height': '256',
 'openslide.level[4].tile-width': '256',
 'openslide.level[4].width': '2875',
 'openslide.level[5].downsample': '32.014322017605849',
 'openslide.level[5].height': '1028',
 'openslide.level[5].tile-height': '256',
 'openslide.level[5].tile-width': '256',
 'openslide.level[5].width': '1437',
 'openslide.level[6].downsample': '64.050935911470475',
 'openslide.level[6].height': '514',
 'openslide.level[6].tile-height': '256',
 'openslide.level[6].tile-width': '256',
 'openslide.level[6].width': '718',
 'openslide.level[7].downsample': '128.10187182294095',
 'openslide.level[7].height': '257',
 'openslide.level[7].tile-height': '256',
 'openslide.level[7].tile-width': '256',
 'openslide.level[7].width': '359',
 'openslide.level[8].downsample': '257.06193261173183',
 'openslide.level[8].height': '128',
 'openslide.level[8].tile-height': '256',
 'openslide.level[8].tile-width': '256',
 'openslide.level[8].width': '179',
 'openslide.mpp-x': '1000',
 'openslide.mpp-y': '1000',
 'openslide.quickhash-1': '428aa6abf42c774234a463cb90e2cbf88423afc0217e46ec2e308f31e29f1a9f',
 'openslide.vendor': 'generic-tiff',
 'tiff.ResolutionUnit': 'centimeter',
 'tiff.XResolution': '10',
 'tiff.YResolution': '10'}

In [12]: dict(openslide.open_slide("Aperio-CMU-1.svs").properties)
Out[12]:
{'aperio.AppMag': '20',
 'aperio.Date': '12/29/09',
 'aperio.Filename': 'CMU-1',
 'aperio.Filtered': '5',
 'aperio.Focus Offset': '0.000000',
 'aperio.ICC Profile': 'ScanScope v1',
 'aperio.ImageID': '1004486',
 'aperio.Left': '25.691574',
 'aperio.LineAreaXOffset': '0.019265',
 'aperio.LineAreaYOffset': '-0.000313',
 'aperio.LineCameraSkew': '-0.000424',
 'aperio.MPP': '0.4990',
 'aperio.OriginalWidth': '46920',
 'aperio.Originalheight': '33014',
 'aperio.Parmset': 'USM Filter',
 'aperio.ScanScope ID': 'CPAPERIOCS',
 'aperio.StripeWidth': '2040',
 'aperio.Time': '09:59:15',
 'aperio.Top': '23.449873',
 'aperio.User': 'b414003d-95c6-48b0-9369-8010ed517ba7',
 'openslide.associated.label.height': '463',
 'openslide.associated.label.width': '387',
 'openslide.associated.macro.height': '431',
 'openslide.associated.macro.width': '1280',
 'openslide.associated.thumbnail.height': '732',
 'openslide.associated.thumbnail.width': '1024',
 'openslide.comment': 'Aperio Image Library v10.0.51\r\n46920x33014 [0,100 46000x32914] (256x256) JPEG/RGB Q=30|AppMag = 20|StripeWidth = 2040|ScanScope ID = CPAPERIOCS|Filename = CMU-1|Date = 12/29/09|Time = 09:59:15|User = b414003d-95c6-48b0-9369-8010ed517ba7|Parmset =
 USM Filter|MPP = 0.4990|Left = 25.691574|Top = 23.449873|LineCameraSkew = -0.000424|LineAreaXOffset = 0.019265|LineAreaYOffset = -0.000313|Focus Offset = 0.000000|ImageID = 1004486|OriginalWidth = 46920|Originalheight = 33014|Filtered = 5|ICC Profile = ScanScope v1',
 'openslide.icc-size': '141992',
 'openslide.level-count': '3',
 'openslide.level[0].downsample': '1',
 'openslide.level[0].height': '32914',
 'openslide.level[0].tile-height': '256',
 'openslide.level[0].tile-width': '256',
 'openslide.level[0].width': '46000',
 'openslide.level[1].downsample': '4.0001215362177929',
 'openslide.level[1].height': '8228',
 'openslide.level[1].tile-height': '256',
 'openslide.level[1].tile-width': '256',
 'openslide.level[1].width': '11500',
 'openslide.level[2].downsample': '16.000486144871171',
 'openslide.level[2].height': '2057',
 'openslide.level[2].tile-height': '256',
 'openslide.level[2].tile-width': '256',
 'openslide.level[2].width': '2875',
 'openslide.mpp-x': '0.499',
 'openslide.mpp-y': '0.499',
 'openslide.objective-power': '20',
 'openslide.quickhash-1': '30f1a38031fc0e21d81f9d01435ac4af848f6fe2bbf8f7768184336ee5d7e796',
 'openslide.vendor': 'aperio',
 'tiff.ImageDescription': 'Aperio Image Library v10.0.51\r\n46920x33014 [0,100 46000x32914] (256x256) JPEG/RGB Q=30|AppMag = 20|StripeWidth = 2040|ScanScope ID = CPAPERIOCS|Filename = CMU-1|Date = 12/29/09|Time = 09:59:15|User = b414003d-95c6-48b0-9369-8010ed517ba7|Parms
et = USM Filter|MPP = 0.4990|Left = 25.691574|Top = 23.449873|LineCameraSkew = -0.000424|LineAreaXOffset = 0.019265|LineAreaYOffset = -0.000313|Focus Offset = 0.000000|ImageID = 1004486|OriginalWidth = 46920|Originalheight = 33014|Filtered = 5|ICC Profile = ScanScope v1'
,
 'tiff.ResolutionUnit': 'inch'}

bhashemian:
Hi @NikolasSchmitz and @ericspod,

Your observation is spot-on! The TIFF file lacks the correct MPP. We’ve been using these files from OpenSlide test cases without making any modifications.

Previously, we simply assumed the files were correct and adjusted our test cases to read them with those metadata. For instance, we would read TIFF files at a specific MPP of 1000 and SVS files at a MPP of 0.4999.

You can continue with the same approach. However, if you can convert the SVS file to TIFF with the correct metadata and update the test cases accordingly, that would be fantastic.

Thanks for your contribution!

NikolasSchmitz:
Hi @bhashemian, thanks for the feedback! 

I have converted the SVS file to TIFF with the correct metadata. According to the Contributing document, the testing data is stored in the Project-MONAI/MONAI-extra-test-data repository. However, I'm unsure about the process for uploading files there—I've noticed that others have opened issues for similar cases. Should I follow the same approach?

In the meantime, I’ll proceed with updating the other test cases accordingly.

Thanks again for your guidance!

NikolasSchmitz:
Hi all, the feature and a few associated tests have been implemented. I ran into an issue with the resizing functions, as they expected different channel orders for the target resolution. That’s been resolved.
The code now passes all ruff checks. However, a few commits are still missing sign-offs. What’s the recommended way to handle that at this point?

I’ve also uploaded the updated TIFF file here: https://gigamove.rwth-aachen.de/en/download/722a2879de053fd9a0ddd7f6e4874624

ericspod:
Hi @NikolasSchmitz thanks for getting back to this. There's a few tests which still aren't passing, a few say just that a timeout occurred so it might be internal but the code formatting test did find something. As for DCO, clicking on that brings up advice on how to resolve, specifically doing an empty remedial commit. If that doesn't work we will set the DCO to pass just before merging. Please have a look at the tests and see if we can get that through.

ericspod:
We also are putting more test data on Huggingface now so you can propose to upload your image to here https://huggingface.co/datasets/MONAI/testing_data if you like. We're planning to adjust our tests to pull from this dataset.

ericspod:
Your new tiff file is added to the hf dataset now, so you need to change `setUpModule` in `test_wsireader.py` to be:

```python
@skipUnless(has_cucim or has_osl or has_tiff, "Requires cucim, openslide, or tifffile!")
def setUpModule():
    download_url_or_skip_test(
        testing_data_config("images", WSI_GENERIC_TIFF_KEY, "url"),
        WSI_GENERIC_TIFF_PATH,
        hash_type=testing_data_config("images", WSI_GENERIC_TIFF_KEY, "hash_type"),
        hash_val=testing_data_config("images", WSI_GENERIC_TIFF_KEY, "hash_val"),
    )
    download_url_or_skip_test(
        testing_data_config("images", WSI_GENERIC_TIFF_CORRECT_MPP_KEY, "url"),
        WSI_GENERIC_TIFF_CORRECT_MPP_PATH,
        hash_type=testing_data_config("images", WSI_GENERIC_TIFF_CORRECT_MPP_KEY, "hash_type"),
        hash_val=testing_data_config("images", WSI_GENERIC_TIFF_CORRECT_MPP_KEY, "hash_val"),
    )
    download_url_or_skip_test(
        testing_data_config("images", WSI_APERIO_SVS_KEY, "url"),
        WSI_APERIO_SVS_PATH,
        hash_type=testing_data_config("images", WSI_APERIO_SVS_KEY, "hash_type"),
        hash_val=testing_data_config("images", WSI_APERIO_SVS_KEY, "hash_val"),
    )
```

Add the following to `tests/testing_data/data_config.json` just after the first item in the `images` section:

```json

        "wsi_generic_tiff_corrected": {
            "url": "https://huggingface.co/datasets/MONAI/testing_data/resolve/main/CMU-1_correct_mpp.tiff",
            "hash_type": "sha256",
            "hash_val": "65306e3f8f7f5282d19d942dadc525cd06a80d5fd8268053939751365226c65f"
        },
```

As you can see, our process for downloading data needs to be revamped to improved this process! Thanks.

bhashemian:
@NikolasSchmitz thanks for your contribution. Please let me know if you need help fixing the DCO issues.

bhashemian:
@coderabbitai full review

bhashemian:
@NikolasSchmitz please let me know if I can help to finalize and merge this PR. Thanks

NikolasSchmitz:
Hey Bruce, I will finalize this PR in the next days and will let you know whether I need help. Thanks!

monai/data/wsi_reader.py
```diff
@@ -431,6 +431,63 @@ def get_data(
...
+        is_within_tolerance_x = (user_mpp_x >= lower_bound_x) & (user_mpp_x <= upper_bound_x)
+        is_within_tolerance_y = (user_mpp_y >= lower_bound_y) & (user_mpp_y <= upper_bound_y)
+        is_within_tolerance = is_within_tolerance_x & is_within_tolerance_y
+
+        # If mpp_closest_level < mpp -> closest_level has higher res than img at mpp => downscale from closest_level to mpp
+        closest_level_is_bigger_x = mpp_closest_lvl_x < user_mpp_x
+        closest_level_is_bigger_y = mpp_closest_lvl_y < user_mpp_y
+        closest_level_is_bigger = closest_level_is_bigger_x & closest_level_is_bigger_y
+
+        return is_within_tolerance, closest_level_is_bigger
```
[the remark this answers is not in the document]
NikolasSchmitz: Added this suggestion.

monai/data/wsi_reader.py
```diff
@@ -744,6 +820,48 @@ def get_mpp(self, wsi, level: int) -> tuple[float, float]:
...
+
+        elif closest_level_is_bigger:
+            # Otherwise, select the level closest to the desired mpp with a higher resolution and downsample it.
+            closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+
+        else:
+            # If both checks fail, increase resolution (i.e., decrement level) and then downsample it.
+            closest_lvl = closest_lvl - 1
+            closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+
```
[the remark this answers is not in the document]
NikolasSchmitz: Added this suggestion.

NikolasSchmitz:
Hi @bhashemian, I’ve just pushed my current progress which includes two additions suggested by coderabbitai.
I could use a hand to sort out the DCO issues, or some hint on how I can solve this easily.
Also, I’ve noticed issues with the output whenever the user-specified MPP is below the image’s native MPP resolution. I’ll try to sort those out.

Thanks!

siemdejong:
Hey, cool PR! Super useful! Would love to see this incorporated into MONAI soon.

Could it be that the method doesn't respect the default RGB mode set for the reader?

```python
>>> from monai.data.wsi_reader import WSIReader
>>> reader = WSIReader("openslide", mode="RGB")
>>> image = reader.read("CMU-3.tiff")
>>> reader.get_wsi_at_mpp(image, (10, 10)).shape
(2307, 3272, 4)
```

Also, I hit my feet against something pointed out in this [coderabbit comment](https://github.com/Project-MONAI/MONAI/pull/7574#discussion_r2527103494). This indeed didn't work while users might expect it to work as the exception is not obvious what to change.
```python
>>> from monai.data.wsi_reader import WSIReader
>>> reader = WSIReader("openslide", mode="RGB")
>>> image = reader.read("CMU-3.tiff")
>>> reader.get_wsi_at_mpp(image, 10).shape
  File "/home/siemdejong/segmenteer/.venv/lib/python3.10/site-packages/monai/data/wsi_reader.py", line 690, in get_wsi_at_mpp
    return self.reader.get_wsi_at_mpp(wsi, mpp, atol, rtol)
  File "/home/siemdejong/segmenteer/.venv/lib/python3.10/site-packages/monai/data/wsi_reader.py", line 1130, in get_wsi_at_mpp
    closest_lvl = self._find_closest_level("mpp", mpp, mpp_list, 0, 5)
  File "/home/siemdejong/segmenteer/.venv/lib/python3.10/site-packages/monai/data/wsi_reader.py", line 161, in _find_closest_level
    closest_value = min(value_at_levels, key=lambda a_value: sum([abs(x - y) for x, y in zip(a_value, value)]))
  File "/home/siemdejong/segmenteer/.venv/lib/python3.10/site-packages/monai/data/wsi_reader.py", line 161, in <lambda>
    closest_value = min(value_at_levels, key=lambda a_value: sum([abs(x - y) for x, y in zip(a_value, value)]))
TypeError: 'int' object is not iterable

```

ericspod:
Hi @NikolasSchmitz we still want to get this merged into MONAI, there's a few things to resolve flagged by Coderabbit and users as well as the DCO. Would it be possible to get back to this? Thanks!

monai/data/wsi_reader.py
```diff
@@ -744,6 +830,54 @@ def get_mpp(self, wsi, level: int) -> tuple[float, float]:
...
+        else:
+            # If both checks fail, increase resolution (i.e., decrement level) and then downsample it.
+            if closest_lvl == 0:
+                closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+            else:
+                closest_lvl = closest_lvl - 1
+                closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+
+        wsi_arr = cp.asnumpy(closest_lvl_wsi)
+        return wsi_arr
```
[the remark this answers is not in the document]
NikolasSchmitz: Added suggested changes.

monai/data/wsi_reader.py
```diff
@@ -433,6 +433,70 @@ def get_data(
...
+        Args:
+            closest_lvl: Whole slide image level closest to user-provided MPP resolution.
+            mpp_list: List of MPP values for all levels of the whole slide image.
+            mpp: The MPP resolution at which the whole slide image representation should be extracted.
+            atol: Absolute tolerance for MPP comparison.
+            rtol: Relative tolerance for MPP comparison.
+
+        Returns:
+            Tuple of (is_within_tolerance, closest_level_is_bigger) where first element indicates
+            if MPP is within tolerance and second indicates if closest level has higher resolution.
```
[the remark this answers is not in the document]
NikolasSchmitz: Fixed! I updated the return type hint from bool to tuple[bool, bool] to accurately reflect the returned values. The docstring was already correct.

monai/data/wsi_reader.py
```diff
@@ -746,6 +832,54 @@ def get_mpp(self, wsi, level: int) -> tuple[float, float]:
...
+        else:
+            # If both checks fail, increase resolution (i.e., decrement level) and then downsample it.
+            if closest_lvl == 0:
+                closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+            else:
+                closest_lvl = closest_lvl - 1
+                closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+
+        wsi_arr = cp.asnumpy(closest_lvl_wsi)
+        return wsi_arr
```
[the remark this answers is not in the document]
NikolasSchmitz: I've updated get_wsi_at_mpp across all three backends (CuCIM, OpenSlide, TiffFile) to correctly apply the configured mode and channel_dim, ensuring the output is consistent with _get_patch() and get_data().

NikolasSchmitz:
> Hey, cool PR! Super useful! Would love to see this incorporated into MONAI soon.
> 
> Could it be that the method doesn't respect the default RGB mode set for the reader?
> 
> ```python
> >>> from monai.data.wsi_reader import WSIReader
> >>> reader = WSIReader("openslide", mode="RGB")
> >>> image = reader.read("CMU-3.tiff")
> >>> reader.get_wsi_at_mpp(image, (10, 10)).shape
> (2307, 3272, 4)
> ```
> 
> Also, I hit my feet against something pointed out in this [coderabbit comment](https://github.com/Project-MONAI/MONAI/pull/7574#discussion_r2527103494). This indeed didn't work while users might expect it to work as the exception is not obvious what to change.
> 
> ```python
> >>> from monai.data.wsi_reader import WSIReader
> >>> reader = WSIReader("openslide", mode="RGB")
> >>> image = reader.read("CMU-3.tiff")
> >>> reader.get_wsi_at_mpp(image, 10).shape
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "/home/siemdejong/segmenteer/.venv/lib/python3.10/site-packages/monai/data/wsi_reader.py", line 690, in get_wsi_at_mpp
>     return self.reader.get_wsi_at_mpp(wsi, mpp, atol, rtol)
>   File "/home/siemdejong/segmenteer/.venv/lib/python3.10/site-packages/monai/data/wsi_reader.py", line 1130, in get_wsi_at_mpp
>     closest_lvl = self._find_closest_level("mpp", mpp, mpp_list, 0, 5)
>   File "/home/siemdejong/segmenteer/.venv/lib/python3.10/site-packages/monai/data/wsi_reader.py", line 161, in _find_closest_level
>     closest_value = min(value_at_levels, key=lambda a_value: sum([abs(x - y) for x, y in zip(a_value, value)]))
>   File "/home/siemdejong/segmenteer/.venv/lib/python3.10/site-packages/monai/data/wsi_reader.py", line 161, in <lambda>
>     closest_value = min(value_at_levels, key=lambda a_value: sum([abs(x - y) for x, y in zip(a_value, value)]))
> TypeError: 'int' object is not iterable
> ```

Thanks for the feedback and the testing! Both issues are fixed now: The RGB mode is properly enforced before returning the array, and I've added a guard (ensure_tuple_rep(mpp, 2)) to safely handle scalar inputs for mpp without throwing a TypeError.

NikolasSchmitz:
> Hi @NikolasSchmitz we still want to get this merged into MONAI, there's a few things to resolve flagged by Coderabbit and users as well as the DCO. Would it be possible to get back to this? Thanks!

Hi @ericspod! Yes, I'm still on board. I just force-pushed an update addressing all feedback, including the DCO issue. Could you please trigger a new CodeRabbit review from your end? Just to make sure I didn't forget anything. Let me know if anything else is needed!

ericspod:
@coderabbitai review this PR and summarise changes as bullet points.

monai/data/wsi_reader.py
```diff
@@ -746,6 +835,75 @@ def get_mpp(self, wsi, level: int) -> tuple[float, float]:
 
         raise ValueError("`mpp` cannot be obtained for this file. Please use `level` instead.")
 
+    def get_wsi_at_mpp(self, wsi, mpp: float | tuple[float, float], atol: float = 0.00, rtol: float = 0.05) -> Any:
```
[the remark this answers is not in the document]
ericspod: Propose this fix as a suggestion comment which can be signed and merged directly

tests/utils/enums/test_wsireader.py
```diff
@@ -407,17 +470,47 @@ class WSIReaderTests:
...
-            reader = WSIReader(self.backend, level=level)
-            with reader.read(file_path) as img_obj:
-                img, meta = reader.get_data(img_obj)
-            self.assertTupleEqual(img.shape, expected_shape)
-            self.assertEqual(meta["backend"], self.backend)
-            self.assertEqual(meta[WSIPatchKeys.PATH].lower(), str(os.path.abspath(file_path)).lower())
-            self.assertEqual(meta[WSIPatchKeys.LEVEL], level)
-            assert_allclose(meta[WSIPatchKeys.SIZE], expected_shape[1:], type_test=False)
-            assert_allclose(meta[WSIPatchKeys.LOCATION], (0, 0), type_test=False)
+        # @parameterized.expand([TEST_CASE_WHOLE_0])
```
ericspod: Let's remove this commented test if there's no reason to keep it as a note.

ericspod (dismissed):
Hi @NikolasSchmitz I think we're good to go here. The coderabbit comments should be looked at, they look minor now. Codeformat is failing so just run `./runtests.sh --autofix` as indicated, however the other fails may relate to the tests you've added being very slow. You may want to add `skip_if_quick` to the new tests to skip them for quick tests. If this doesn't work we'll dig deeper into the issue.

monai/data/wsi_reader.py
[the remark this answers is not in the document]
NikolasSchmitz: Applied the proposed fix.

monai/data/wsi_reader.py
```diff
@@ -1105,12 +1378,84 @@ def get_mpp(self, wsi, level: int) -> tuple[float, float]:
...
+        # Apply mode (RGB / RGBA)
+        if self.mode == "RGB":
+            if wsi_arr.shape[self.channel_dim] not in [3, 4]:
+                raise ValueError(
+                    f"The image is expected to have three or four color channels in '{self.mode}' mode but has "
+                    f"{wsi_arr.shape[self.channel_dim]}. "
+                )
+            wsi_arr = np.take(wsi_arr, [0, 1, 2], self.channel_dim)
+
+        return wsi_arr
```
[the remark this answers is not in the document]
NikolasSchmitz: The raw array is now passed through PIL (.convert(self.mode)) to guarantee a consistent uint8 numpy output, matching the other branches.

tests/utils/enums/test_wsireader.py
NikolasSchmitz: Hi Eric, I commented it out temporarily to save some time when running tests and forgot to change it back. I've uncommented it now so it matches the original test_wsireader.py file.

monai/data/wsi_reader.py
```diff
@@ -746,6 +835,77 @@ def get_mpp(self, wsi, level: int) -> tuple[float, float]:
...
+
+        else:
+            # If both checks fail, increase resolution (i.e., decrement level) and then downsample it.
+            if closest_lvl == 0:
+                closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+            else:
+                closest_lvl = closest_lvl - 1
+                closest_lvl_wsi = self._resize_to_mpp_res(wsi, closest_lvl, mpp_list, mpp)
+
+        wsi_arr = cp.asnumpy(closest_lvl_wsi)
```
ericspod:
```suggestion
        wsi_arr: np.ndarray = cp.asnumpy(closest_lvl_wsi)
```

monai/data/wsi_reader.py
```diff
@@ -746,6 +835,77 @@ def get_mpp(self, wsi, level: int) -> tuple[float, float]:
...
+
+        Returns:
+            Numpy array containing the whole slide image at the requested MPP resolution.
+
+        """
+        cp, _ = optional_import("cupy")
+        mpp = ensure_tuple_rep(mpp, 2)
+
+        mpp_list = [self.get_mpp(wsi, lvl) for lvl in range(wsi.resolutions["level_count"])]
+        closest_lvl = self._find_closest_level("mpp", mpp, mpp_list, 0, 5)
```
bhashemian, line 864: @NikolasSchmitz could you please elaborate on the logic of using atol=0 and rtol=5 here? Thanks

NikolasSchmitz: Hi @bhashemian! The high rtol=5 is used as a deliberate catch-all. We need to ensure that `_find_closest_level` always returns some level (the mathematically closest one) to act as a starting point for the (down)scaling process. If we used strict tolerances here and the user's mpp was far off, the function would return nothing and crash the pipeline. I used 5 as a sufficiently large number to bypass any strict filtering at this stage.

NikolasSchmitz:
> Hi @NikolasSchmitz I think we're good to go here. The coderabbit comments should be looked at, they look minor now. Codeformat is failing so just run `./runtests.sh --autofix` as indicated, however the other fails may relate to the tests you've added being very slow. You may want to add `skip_if_quick` to the new tests to skip them for quick tests. If this doesn't work we'll dig deeper into the issue.

Hi @ericspod, I've addressed the coderabbit comments and ran `./runtests.sh --autofix` which passed all checks! I've also added the decorator skip_if_quick and the necessary import for that.

monai/data/wsi_reader.py
ericspod: I'm going to mark this as resolved now so we can merge, but we can come back to the tolerance settings later if we want.

ericspod (approved):
Looks good now, thanks!

Issue #7575, opened 2024-03-25
Ignore the 'CUDA_VISIBLE_DEVICE' setting after importing Monai in the latest Pytorch


KumoLiu:
```
FAIL: test_visible_devices (tests.test_set_visible_devices.TestVisibleDevices)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/MONAI/tests/test_set_visible_devices.py", line 37, in test_visible_devices
    self.assertEqual(num_gpus_before, num_gpus_after)
AssertionError: 0 != 2
```

The root cause could be issues listed here:
[https://github.com/pytorch/pytorch/issues/122085](https://nam11.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fpytorch%2Fpytorch%2Fissues%2F122085&data=05%7C02%7Cyunl%40nvidia.com%7Cbe8ae127a4f14e6fb93508dc4a1d8eb9%7C43083d15727340c1b7db39efd9ccc17a%7C0%7C0%7C638466738026714791%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=FIU8Awmy3gV%2FvamjyRtNJ2fCFL7jH9NFHETfMdhks%2FE%3D&reserved=0)

PR #7576, opened 2024-03-25
Utilizing subprocess for nnUNet training.

KumoLiu:
Workaround for #7575 

### Description
- Due to the impact of  #7575, the operation to set the device within nnUNetV2Runner will become ineffective. This PR is intended to resolve this issue.
- Add a version check for #7575, will revisit after the update from pytorch team.

### Types of changes

### DIFF
--- a/monai/apps/nnunet/nnunetv2_runner.py
+++ b/monai/apps/nnunet/nnunetv2_runner.py
@@ -496,17 +496,16 @@ class nnUNetV2Runner:  # noqa: N801
             fold: fold of the 5-fold cross-validation. Should be an int between 0 and 4.
             gpu_id: an integer to select the device to use, or a tuple/list of GPU device indices used for multi-GPU
                 training (e.g., (0,1)). Default: 0.
-        from nnunetv2.run.run_training import run_training
             kwargs: this optional parameter allows you to specify additional arguments in
                 ``nnunetv2.run.run_training.run_training``. Currently supported args are
-                    - plans_identifier: custom plans identifier. Default: "nnUNetPlans".
+                    - p: custom plans identifier. Default: "nnUNetPlans".
                     - pretrained_weights: path to nnU-Net checkpoint file to be used as pretrained model. Will only be
                         used when actually training. Beta. Use with caution. Default: False.
-                    - use_compressed_data: True to use compressed data for training. Reading compressed data is much
+                    - use_compressed: True to use compressed data for training. Reading compressed data is much
                         more CPU and (potentially) RAM intensive and should only be used if you know what you are
                         doing. Default: False.
-                    - continue_training: continue training from latest checkpoint. Default: False.
-                    - only_run_validation: True to run the validation only. Requires training to have finished.
+                    - c: continue training from latest checkpoint. Default: False.
+                    - val: True to run the validation only. Requires training to have finished.
                         Default: False.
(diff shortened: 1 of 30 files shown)


monai/apps/nnunet/nnunetv2_runner.py
```diff
@@ -522,38 +523,33 @@ def train_single_model(self, config: Any, fold: int, gpu_id: tuple | list | int
             kwargs.pop("export_validation_probabilities")
             logger.warning("please specify the `export_validation_probabilities` in the __init__ of `nnUNetV2Runner`.")
 
+        # from nnunetv2.run.run_training import run_training
```
Nic-Ma, line 526: Please remove commented code

Nic-Ma (approved):

monai/apps/nnunet/nnunetv2_runner.py
```diff
@@ -522,38 +523,33 @@ def train_single_model(self, config: Any, fold: int, gpu_id: tuple | list | int
...
+        cmd = self.train_single_model_command(config, fold, gpu_id, kwargs)
+        run_cmd(cmd, shell=True)
```
mingxin-zheng, line 528: There are a couple of environment variable settings in:
https://github.com/Project-MONAI/MONAI/blob/e26cab4cd1e340b839bbcfd1f2e4906599a3242f/monai/apps/nnunet/nnunetv2_runner.py#L190

Do we need to copy them to the environment variables in this subprocess run under `run_cmd`?

KumoLiu: It looks like it should be put in init, so that preprocessing, training, and inference can use.
What do you think?

mingxin-zheng: Hi @KumoLiu , to clarify my previous question, I mean maybe we need to:
```
ps_environ = os.environ.copy()
run_cmd(cmd.split(), env=ps_environ, ...)
```
I agree with your comment that they should stay where they were (init).

KumoLiu: Got it.
If env is None, it inherits the current process's environment. So it looks like we don't have to pass it?

mingxin-zheng: Thanks, that resolves my question!

mingxin-zheng (approved):

monai/apps/nnunet/nnunetv2_runner.py
```diff
@@ -495,65 +496,61 @@ def train_single_model(self, config: Any, fold: int, gpu_id: tuple | list | int
             fold: fold of the 5-fold cross-validation. Should be an int between 0 and 4.
             gpu_id: an integer to select the device to use, or a tuple/list of GPU device indices used for multi-GPU
                 training (e.g., (0,1)). Default: 0.
-        from nnunetv2.run.run_training import run_training
             kwargs: this optional parameter allows you to specify additional arguments in
                 ``nnunetv2.run.run_training.run_training``. Currently supported args are
```
dongyang0122, line 500: Please change ``nnunetv2.run.run_training.run_training`` to  ``nnunetv2.run.run_training.run_training_entry``. Thanks.

dongyang0122 (approved):

PR #7587, opened 2024-03-27
harmonization and clarification of dice losses variants docs and associated tests

Lucas-rbnt:
### Description

This PR aims to clarify and harmonise the code for the DiceLoss variants in the `monai/losses/dice.py` file. With the `to_onehot_y` `softmax` and `sigmoid` arguments, I didn't necessarily understand the ValueError that occurred when I passed a target of size NH[WD]. I had a bit of trouble reading the documentation and understanding it. I thought that they had to be the same shape as they are displayed, unlike the number of dimensions in the input, so I added that.
Besides, in the documentation is written:
```python
"""
raises:
      ValueError: When number of channels for target is neither 1 nor the same as input.

"""
```
Trying to reproduce this, we give an input with a number of channels $N$ and target a number of channels of $M$, with $M \neq N$ and $M > 1$.
```python
loss = DiceCELoss()
input = torch.rand(1, 4, 3, 3)
target = torch.randn(1, 2, 3, 3)
loss(input, target)
>: AssertionError: ground truth has different shape (torch.Size([1, 2, 3, 3])) from input (torch.Size([1, 4, 3, 3]))
```
This error in the Dice is an `AssertionError` and not a `ValueError` as expected and the explanation can be confusing and doesn't give a clear idea of the error here. The classes concerned and harmonised are `DiceFocalLoss`, `DiceCELoss` and `GeneralizedDiceFocalLoss` with the addition of tests that behave correctly and handle this harmonisation. 

Also, feel free to modify or make suggestions regarding the changes made in the docstring to make them more understandable (in my opinion, but other readers and users will probably have a different view).

### Types of changes

### DIFF
--- a/monai/losses/dice.py
+++ b/monai/losses/dice.py
@@ -776,14 +776,23 @@ class DiceCELoss(_Loss):
             input: the shape should be BNH[WD].
             target: the shape should be BNH[WD] or B1H[WD].
 
-        Raises:
-            ValueError: When number of dimensions for input and target are different.
-            ValueError: When number of channels for target is neither 1 nor the same as input.
+        ValueError: When number of dimensions for input and target are different.
+        ValueError: When the target channel isn't either one-hot encoded or categorical with the same shape of the input.
+
+        Returns:
+            torch.Tensor: value of the loss.
 
         """
-        if len(input.shape) != len(target.shape):
+        if input.dim() != target.dim():
             raise ValueError(
                 "the number of dimensions for input and target should be the same, "
+                f"got shape {input.shape} (nb dims: {len(input.shape)}) and {target.shape} (nb dims: {len(target.shape)}). "
+                "if target is not one-hot encoded, please provide a tensor with shape B1H[WD]."
+            )
+
+        if target.shape[1] != 1 and target.shape[1] != input.shape[1]:
+            raise ValueError(
+                "number of channels for target is neither 1 (with no one-hot encoding) nor the same as input, "
                 f"got shape {input.shape} and {target.shape}."
             )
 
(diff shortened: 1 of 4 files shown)


monai/losses/dice.py
```diff
@@ -778,12 +778,22 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
 
         Raises:
             ValueError: When number of dimensions for input and target are different.
-            ValueError: When number of channels for target is neither 1 nor the same as input.
+            ValueError: When the target channel isn't either one-hot encoded or categorical with the same shape of the input.
```
KumoLiu, line 781: The error message seems to indicate that the data must either be in one-hot format or consistent with the original image, which appears to slightly differ from the initial implication.

monai/losses/dice.py
```diff
@@ -1038,3 +1066,11 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
...
 generalized_dice_focal = GeneralizedDiceFocalLoss
 generalized_wasserstein_dice = GeneralizedWassersteinDiceLoss
+
+if __name__ == "__main__":
+    loss = DiceFocalLoss(to_onehot_y=False, softmax=True)
+    input = torch.randn(1, 3, 128, 128, 128)
+    target = torch.randn(1, 4, 128, 128, 128)
+    print(input.dim())
+    print(target.dim())
+    loss(input, target)
```
KumoLiu, line 1076:
```suggestion
```

KumoLiu (commented):
Thank you for this pull request, overall, the updates are looking good to me. However, I do have two minor comments inline for your consideration.

monai/losses/dice.py
Lucas-rbnt: Yes, sorry, on re-reading the original message is clearer, I've changed that, hoping it's better now!

KumoLiu (approved):

Issue #6704, opened 2023-07-11
A utility function to generate sampling points


wyli:
**Is your feature request related to a problem? Please describe.**
Currently there are a few functions in monai.transforms to help sampling image patches:

1. `map_classes_to_indices` to identify all feasible sampling indices
2. `generate_label_classes_crop_centers` to subsample and generate spatial coordinates
3. `RandCropByLabelClasses` combines 1, 2 and output image patches

**Describe the solution you'd like**

Would be great to have a helper function to run step 1 and 2 as a single command. The function should output sampling coordinates instead of the final patch

**Describe alternatives you've considered**

Write the logic with multiple commands: https://gist.github.com/wyli/50c65e686beaf418a32c9cc131f0bdc7

Same idea applies to the other sampling transforms such as RandCropByPosNegLabel

CCU612410030:
Hi @wyli , How can I obtain the file "post_label0002.h5" in "example_sample.py"?

CCU612410030:
I wrote this helper function with the expectation that the output would be the coordinates of the centers. I hope it meets your requirements.

`def combineMCtIandGLCCC(
    label: NdarrayOrTensor,
    spatial_size: Sequence[int] | int,
    num_samples: int,
    label_spatial_shape: Sequence[int],
    indices: Sequence[NdarrayOrTensor],
    num_classes: int | None = None,
    image: NdarrayOrTensor | None = None,
    image_threshold: float = 0.0,
    max_samples_per_class: int | None = None,
    ratios: list[float | int] | None = None,
    rand_state: np.random.RandomState | None = None,
    allow_smaller: bool = False,
    warn: bool = True,
) -> list[NdarrayOrTensor]:
    """
    Combine "map_classes_to_indices" and "generate_pos_neg_label_crop_centers" functions, return crop center coordinates.

     Args:
        spatial_size: spatial size of the ROIs to be sampled.
        num_samples: total sample centers to be generated.
        pos_ratio: ratio of total locations generated that have center being foreground.
        label_spatial_shape: spatial shape of the original label data to unravel selected centers.
        fg_indices: pre-computed foreground indices in 1 dimension.
        bg_indices: pre-computed background indices in 1 dimension.
        rand_state: numpy randomState object to align with other modules.
        allow_smaller: if `False`, an exception will be raised if the image is smaller than
            the requested ROI in any dimension. If `True`, any smaller dimensions will be set to
            match the cropped size (i.e., no cropping in that dimension).
[...]

PR #7588, opened 2024-03-27
Update utils.py

Yu0610:
Add combineMCtIandGLCCC()
From the request "A utility function to generate sampling points #6704"
https://github.com/Project-MONAI/MONAI/issues/6704

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -778,10 +778,19 @@ def run(
             https://docs.python.org/3/library/logging.config.html#logging.config.fileConfig.
             Default to None.
         tracking: if not None, enable the experiment tracking at runtime with optionally configurable and extensible.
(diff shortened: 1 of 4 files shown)


monai/transforms/utils.py
```diff
@@ -642,6 +643,63 @@ def generate_label_classes_crop_centers(
     return ensure_tuple(centers)
 
 
+def combineMCtIandGLCCC(
```
KumoLiu, line 646: Would be better to use a more meaningful name, such as `map_and_generate_sampling_centers`.

ericspod: I would second here that we need a much more descriptive name here. It would also help to state the dimensions input tensors must have and what the dimensions mean in the docstring. Thanks!

Yu0610: I'll change the name, Thanks!

PR #7604, opened 2024-04-01
7601 fix mlflow artifacts

### LINKED ISSUE #7601: Default MLFlow tracking cannot set if tracking the config artifacts
When running a typical MONAI model zoo bundle training with `tracking="mlflow"`, it will always try to record all the config files as artifacts due to [this line](https://github.com/Project-MONAI/MONAI/blob/c885100e951e5b7b1fedf5e9bc965201592e8541/monai/bundle/workflows.py#L494). If the `execute_config` is set, it will track config files in the specific path. When the `execute_config` parameter is not set it will track the default path. 

The `execute_config` parameter should be a `bool| string` type, so that users can set it as `False`, if they do not want to record any configs.  They can set the parameter to `True` to record the default path or to a real path if config files are saved in other places.

### LINKED ISSUE #7610: Inaccurate error message in regression.py
**Describe the bug**

On line 306 of monai.metrics.regression, the printed error message states that y_pred should have 4 dimensions when in fact the surrounding if-statement is looking for the case where y_pred should have 5 dimensions. This confused me because I got an error message saying:

 "y_pred should have 4 dimensions (batch, channel, height, width, depth) when using 3 spatial dimensions, got 4."

**Expected behavior**

All that needs to be changed is the number 4 to a 5

binliunls:
Fixes #7601 .

### Description

Support not save artifacts.

### Types of changes

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -778,10 +778,19 @@ def run(
             https://docs.python.org/3/library/logging.config.html#logging.config.fileConfig.
             Default to None.
         tracking: if not None, enable the experiment tracking at runtime with optionally configurable and extensible.
-            if "mlflow", will add `MLFlowHandler` to the parsed bundle with default tracking settings,
-            if other string, treat it as file path to load the tracking settings.
-            if `dict`, treat it as tracking settings.
-            will patch the target config content with `tracking handlers` and the top-level items of `configs`.
+            If "mlflow", will add `MLFlowHandler` to the parsed bundle with default tracking settings where a set of
+            common parameters shown below will be added and can be passed through the `override` parameter of this method.
+
+            - ``"output_dir"``: the path to save mlflow tracking outputs locally, default to "<bundle root>/eval".
+            - ``"tracking_uri"``: uri to save mlflow tracking outputs, default to "/output_dir/mlruns".
+            - ``"experiment_name"``: experiment name for this run, default to "monai_experiment".
+            - ``"run_name"``: the name of current run.
+            - ``"save_execute_config"``: whether to save the executed config files. It can be `False`, `/path/to/artifacts`
+              or `True`, default to `True`.
+
+            If other string, treat it as file path to load the tracking settings.
+            If `dict`, treat it as tracking settings.
+            Will patch the target config content with `tracking handlers` and the top-level items of `configs`.
             for detailed usage examples, please check the tutorial:
             https://github.com/Project-MONAI/tutorials/blob/main/experiment_management/bundle_integrate_mlflow.ipynb.
         args_file: a JSON or YAML file to provide default values for `run_id`, `meta_file`,
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -786,7 +786,7 @@ def run(
             - ``"experiment_name"``: experiment name for this run, default to "monai_experiment".
             - ``"run_name"``: the name of current run.
             - ``"save_execute_config"``: whether to save the executed config files. It can be `False`, `/path/to/artifacts`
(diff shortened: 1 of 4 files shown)


monai/bundle/utils.py
```diff
@@ -113,7 +113,7 @@
         "experiment_name": "monai_experiment",
         "run_name": None,
         # may fill it at runtime
-        "execute_config": None,
+        "save_execute_config": True,
```
Nic-Ma, line 116: As the `DEFAULT_MLFLOW_SETTINGS` is more and more complicated, I would suggest to write a clear doc-string for the description of every parameter here and add to the API-DOC: https://docs.monai.io/en/latest/bundle.html.

monai/bundle/workflows.py
```diff
@@ -506,13 +506,19 @@ def patch_bundle_tracking(parser: ConfigParser, settings: dict) -> None:
...
-        if filepath is None:
-            if "output_dir" not in parser:
-                # if no "output_dir" in the bundle config, default to "<bundle root>/eval"
-                parser["output_dir"] = f"{EXPR_KEY}{ID_REF_KEY}bundle_root + '/eval'"
-            # experiment management tools can refer to this config item to track the config info
-            parser["execute_config"] = parser["output_dir"] + f" + '/{default_name}'"
-            filepath = os.path.join(parser.get_parsed_content("output_dir"), default_name)
-        Path(filepath).parent.mkdir(parents=True, exist_ok=True)
-        parser.export_config_file(parser.get(), filepath)
+        # Users can set the `save_execute_config` to `False`, `/path/to/artifacts` or `True`.
```
Nic-Ma, line 509: Move these description comments to the `DEFAULT_MLFLOW_SETTINGS` docstring. otherwise, users can't see it in the API-DOC https://docs.monai.io/en/latest/bundle.html.

Thanks.

Nic-Ma (approved):
Overall looks good to me, put minor comments inline.
Please remember to update the associated tutorial notebook.

Thanks.

monai/bundle/utils.py
binliunls: Ok will update the doc.

monai/bundle/workflows.py
binliunls: Will do this.

binliunls: Done

monai/bundle/utils.py
binliunls: Done

monai/bundle/scripts.py
```diff
@@ -778,10 +778,19 @@ def run(
...
-            will patch the target config content with `tracking handlers` and the top-level items of `configs`.
+            If "mlflow", will add `MLFlowHandler` to the parsed bundle with default tracking settings where a set of
+            common parameters shown below will be added and can be passed through the `override` parameter of this method.
+
+            - ``"output_dir"``: the path to save mlflow tracking outputs locally, default to "<bundle root>/eval".
+            - ``"tracking_uri"``: uri to save mlflow tracking outputs, default to "/output_dir/mlruns".
+            - ``"experiment_name"``: experiment name for this run, default to "monai_experiment".
+            - ``"run_name"``: the name of current run.
+            - ``"save_execute_config"``: whether to save the executed config files. It can be `False`, `/path/to/artifacts`
+              or `True`, default to `True`.
```
Nic-Ma, line 789: Please explain what the default path if `True`.

Thanks.

Nic-Ma (approved):

Nic-Ma:
Hi @KumoLiu ,

Could you please help include this bug fix PR in the next MONAI 1.3.1RC release?

Thanks in advance.

PR #7605, opened 2024-04-02
simplification of the sincos positional encoding in patchembedding.py

Lucas-rbnt:
Fixes #7564 .

### Description

As discussed, a small simplification for the creation of sincos positional encoding where we don't need to use the `torch.no_grad()` context or copy the tensor with `copy_` from torch which doesn't preserve the `requires_grad` attribute here.

The changes are simple and are linked to the corresponding comment #7564, the output is already in float32 so it doesn't seem particularly necessary to apply the conversion previously done.

### Types of changes


marksgraham (approved):
LGTM

PR #7640, opened 2024-04-13
Enable dynamo onnx export of BasicUNet

asfiyab-nvidia:
### Enable dynamo onnx export of BasicUNet models

TorchDynamo based ONNX export does not support `torch.jit.isinstance()`. To enable the export via dynamo, I've updated the calls to `torch.jit.isinstance` to `isinstance`. I've also verified that the TorchScript based ONNX export still works with this change. These changes enable the export of BasicUNet, BasicUNetPlusPlus, FlexibleUNet, TorchvisionFC models.

### Types of changes

- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`. - Errors caused due to dataloader
- [ ] Quick tests passed locally by running `./runtests.sh --quick --unittests  --disttests`. - Errors caused due to dataloader

### DIFF
--- a/monai/networks/nets/basic_unet.py
+++ b/monai/networks/nets/basic_unet.py
@@ -159,7 +159,7 @@ class UpCat(nn.Module):
         """
         x_0 = self.upsample(x)
 
-        if x_e is not None and torch.jit.isinstance(x_e, torch.Tensor):
+        if x_e is not None and isinstance(x_e, torch.Tensor):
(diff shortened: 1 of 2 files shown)


ericspod:
Judging from the errors you do need to specify `Dict` with the contained types even though you're explicitly checking for these and Python's `isinstance` won't use them.

PR #7662, opened 2024-04-18
Fixed four test issues within test code.

freddiewanah:
### Description

This PR fixed the three issues within the test code that could cause problems during CICD.

Three issues:

tests/test_pad_collation.py line 120: supposed to be assertEqual, but not assertTure? or it will always be true.
tests/test_to_numpy.py line 74: Same as above.
tests/test_auto3dseg.py line 370: typo? two same asserts next to each other.
tests/test_compose.py line 719, 725, 727: Should be assertEqual instead of assertTrue

### Types of changes

### DIFF
--- a/tests/test_auto3dseg.py
+++ b/tests/test_auto3dseg.py
@@ -367,7 +367,6 @@ class TestDataAnalyzer(unittest.TestCase):
         for batch_data in self.dataset:
             d = transform(batch_data[0])
             assert DataStatsKeys.BY_CASE_IMAGE_PATH in d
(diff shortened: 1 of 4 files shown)


atbenmurray (approved):
These changes all look sensible and correct

atbenmurray:
@KumoLiu the failures don't look related to the changes at a glance. Do we have some test configuration issue at the moment?

KumoLiu:
Yes, the update from mlflow has caused some issues. The fix will be included in this PR: https://github.com/Project-MONAI/MONAI/pull/7659.
Will merge this one after that.
Thanks.

PR #7670, opened 2024-04-18
Adapt to use assert raises

freddiewanah:
### Description

Some test cases use try/except command, instead of catch exceptions with assertRaise. This will cause extra execution time according to the experimental tests.

### Types of changes


KumoLiu (approved):
Thanks for the PR, looks good to me.

PR #7671, opened 2024-04-18
Refactored test assertions that have suboptimal tests with numbers

freddiewanah:
### Description

As discussed in PR #7609, I tried to break the suboptimal test refactors to smaller pieces.

Suboptimal Assert: Instead of using statements such as assertIsNone, assertIsInstance, always simply use assertTrue or assertFalse. This will decrease the code overall readability and increase the execution time as extra logic needed.

### Types of changes

### DIFF
--- a/tests/test_bundle_get_data.py
+++ b/tests/test_bundle_get_data.py
@@ -53,7 +53,7 @@ class TestGetBundleData(unittest.TestCase):
             output = get_all_bundles_list(**params)
             self.assertTrue(isinstance(output, list))
             self.assertTrue(isinstance(output[0], tuple))
-            self.assertEqual(output[0], 2)
+            self.assertEqual(len(output[0]), 2)
 
     @parameterized.expand([TEST_CASE_1, TEST_CASE_5])
     @skip_if_quick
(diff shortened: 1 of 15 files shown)


tests/test_bundle_get_data.py
```diff
@@ -53,7 +53,7 @@ def test_get_all_bundles_list(self, params):
             output = get_all_bundles_list(**params)
             self.assertTrue(isinstance(output, list))
             self.assertTrue(isinstance(output[0], tuple))
-            self.assertTrue(len(output[0]) == 2)
+            self.assertEqual(output[0], 2)
```
KumoLiu, line 56:
```suggestion
            self.assertEqual(len(output[0]), 2)
```

Looks like len should be added here?

KumoLiu (approved):
Thanks for the PR , overall looks good to me.
Just a minor comment inline.

tests/test_bundle_get_data.py
freddiewanah: You're absolutely right, and sorry for making this mistake.

Issue #7616, opened 2024-04-10
Test error in test_soft_clipping_one_sided_high


KumoLiu:
```
[2024-04-10T01:27:20.324Z] ======================================================================
[2024-04-10T01:27:20.324Z] FAIL: test_soft_clipping_one_sided_high_2 (tests.test_clip_intensity_percentiles.TestClipIntensityPercentiles2D)
[2024-04-10T01:27:20.324Z] ----------------------------------------------------------------------
[2024-04-10T01:27:20.324Z] Traceback (most recent call last):
[2024-04-10T01:27:20.324Z]   File "/usr/local/lib/python3.10/dist-packages/parameterized/parameterized.py", line 620, in standalone_func
[2024-04-10T01:27:20.324Z]     return func(*(a + p.args), **p.kwargs, **kw)
[2024-04-10T01:27:20.324Z]   File "/opt/monai/tests/test_clip_intensity_percentiles.py", line 71, in test_soft_clipping_one_sided_high
[2024-04-10T01:27:20.324Z]     assert_allclose(result, p(expected), type_test="tensor", rtol=5e-5, atol=0)
[2024-04-10T01:27:20.324Z]   File "/opt/monai/tests/utils.py", line 135, in assert_allclose
[2024-04-10T01:27:20.324Z]     np.testing.assert_allclose(actual, desired, *args, **kwargs)
[2024-04-10T01:27:20.324Z]   File "/usr/local/lib/python3.10/dist-packages/numpy/testing/_private/utils.py", line 1592, in assert_allclose
[2024-04-10T01:27:20.324Z]     assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
[2024-04-10T01:27:20.324Z]   File "/usr/lib/python3.10/contextlib.py", line 79, in inner
[2024-04-10T01:27:20.324Z]     return func(*args, **kwds)
[2024-04-10T01:27:20.324Z]   File "/usr/local/lib/python3.10/dist-packages/numpy/testing/_private/utils.py", line 862, in assert_array_compare
[2024-04-10T01:27:20.324Z]     raise AssertionError(msg)
[2024-04-10T01:27:20.324Z] AssertionError: 
[2024-04-10T01:27:20.324Z] Not equal to tolerance rtol=5e-05, atol=0
[2024-04-10T01:27:20.324Z] 
[2024-04-10T01:27:20.324Z] Mismatched elements: 8192 / 8192 (100%)
[...]

PR #7675, opened 2024-04-19
Always convert input to C-order in distance_transform_edt

### LINKED ISSUE #7660: distance_transform_edt not always working with cuCIM backend
**Describe the bug**
The distance_transform_edt returns all zero in special cases when cuCIM is installed and used as a backend. 
This happens when convert_to_cupy() decides that the data array needs to be copied [here:](https://github.com/Project-MONAI/MONAI/blob/16d4e2fcdbc6a7be5bf9df02fc12f34eee5f26b9/monai/transforms/utils.py#L2199)

**To Reproduce**
```
!pip install torch
!pip install cucim
!pip install monai
!pip install numpy

from monai.transforms import distance_transform_edt
import torch
import numpy as np

# create a c-contiguous array
mask = np.zeros((1, 4, 4))

# make it f-contiguous to force convert_to_cupy to copy the data array later on
mask2 = np.asfortranarray(mask)

# convert to torch tensor to make distance_transform_edt use cuCIM backend
mask = torch.tensor(mask, device='cuda')
mask2 = torch.tensor(mask2, device='cuda')

# set half of the mask to foreground
mask[0, 0:2, :,] = 1
mask2[0, 0:2, :,] = 1

# confirm masks are the same
print(mask)
print(mask2)

# calculate distance transform
edt = distance_transform_edt(mask)
edt2 = distance_transform_edt(mask2)

# check if the results are the same
print(edt)
print(edt2)
```
```
tensor([[[1., 1., 1., 1.],
         [1., 1., 1., 1.],
         [0., 0., 0., 0.],
         [0., 0., 0., 0.]]], device='cuda:0', dtype=torch.float64)
tensor([[[1., 1., 1., 1.],
         [1., 1., 1., 1.],
         [0., 0., 0., 0.],
         [0., 0., 0., 0.]]], device='cuda:0', dtype=torch.float64)
tensor([[[2., 2., 2., 2.],
         [1., 1., 1., 1.],
         [0., 0., 0., 0.],
         [0., 0., 0., 0.]]], device='cuda:0')
tensor([[[0., 0., 0., 0.],
         [0., 0., 0., 0.],
         [0., 0., 0., 0.],
         [0., 0., 0., 0.]]], device='cuda:0')
```

**Expected behavior**
Results should be the same, but the second output is all zero.

KumoLiu:
Fixes #7660

`indices_` may not always be a shallow copy for F-order input as it constantly converts to C-order
https://github.com/Project-MONAI/MONAI/blob/ffd4454b576abd4eaae30b364f41c213e30dca4c/monai/transforms/utils.py#L2209
https://github.com/Project-MONAI/MONAI/blob/ffd4454b576abd4eaae30b364f41c213e30dca4c/monai/utils/type_conversion.py#L262

### Types of changes

### DIFF
--- a/monai/transforms/utils.py
+++ b/monai/transforms/utils.py
@@ -2190,7 +2190,7 @@ def distance_transform_edt(
         if return_distances:
             dtype = torch.float64 if float64_distances else torch.float32
             if distances is None:
-                distances = torch.zeros_like(img, dtype=dtype)  # type: ignore
+                distances = torch.zeros_like(img.contiguous(), dtype=dtype)  # type: ignore
             else:
                 if not isinstance(distances, torch.Tensor) and distances.device != img.device:
                     raise TypeError("distances must be a torch.Tensor on the same device as img")
--- a/monai/transforms/utils.py
+++ b/monai/transforms/utils.py
@@ -2190,7 +2190,7 @@ def distance_transform_edt(
         if return_distances:
             dtype = torch.float64 if float64_distances else torch.float32
             if distances is None:
-                distances = torch.zeros_like(img.contiguous(), dtype=dtype)  # type: ignore
+                distances = torch.zeros_like(img, memory_format=torch.contiguous_format, dtype=dtype)  # type: ignore
             else:
                 if not isinstance(distances, torch.Tensor) and distances.device != img.device:
                     raise TypeError("distances must be a torch.Tensor on the same device as img")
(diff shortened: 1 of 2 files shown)


monai/transforms/utils.py
```diff
@@ -2190,7 +2190,7 @@ def distance_transform_edt(
         if return_distances:
             dtype = torch.float64 if float64_distances else torch.float32
             if distances is None:
-                distances = torch.zeros_like(img, dtype=dtype)  # type: ignore
+                distances = torch.zeros_like(img.contiguous(), dtype=dtype)  # type: ignore
```
ericspod, line 2193: This is going to make a copy of `img` if it isn't contiguous, instead I think we can use `torch.zeros_like(img, memory_format=torch.contiguous_format)`.

ericspod: Looks good to me, I'm not sure if the CI/CD issues are related however.

KumoLiu: Hi Eric, not related to this PR but for this one: https://github.com/Project-MONAI/MONAI/issues/7616.

ericspod (approved):

PR #7698, opened 2024-04-23
Fix download failing on FIPS machines

MattTheCuber:
### Description
This PR fixes downloads failing on FIPS enabled machines due to insecure MD5 hashing. The two solutions are to disable MD5 hashing (SHA1 is allowed and faster), or use the `usedforsecurity=False` flag. This PR uses the second method. However, the `usedforsecurity` flag only works for Python 3.9 and later (which was accounted for). Let me know if you have a better implementation to solve this issue.

The error thrown on FIPS enabled machine is:
```ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS```

### Types of changes

### DIFF
--- a/monai/apps/utils.py
+++ b/monai/apps/utils.py
@@ -135,7 +135,12 @@ def check_hash(filepath: PathLike, val: str | None = None, hash_type: str = "md5
         logger.info(f"Expected {hash_type} is None, skip {hash_type} check for file {filepath}.")
         return True
     actual_hash_func = look_up_option(hash_type.lower(), SUPPORTED_HASH_TYPES)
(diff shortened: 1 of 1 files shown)


ericspod:
I think this change is fine. @Nic-Ma @KumoLiu just to check with you that there isn't any security implications to consider here.

KumoLiu:
> I think this change is fine. @Nic-Ma @KumoLiu just to check with you that there isn't any security implications to consider here.

I checked the usage for this `check_hash`, looks only used for downloading, will merge this one first.

KumoLiu (approved):

PR #7702, opened 2024-04-23
fixed unit tests with random function

### LINKED ISSUE #7697: AssertionError in test_regularization
Random;y occur this error:
```
FAIL: test_cutmixd (tests.test_regularization.TestCutMix)
  File "/home/runner/work/MONAI/MONAI/tests/test_regularization.py", line 88, in test_cutmixd
    self.assertTrue(not torch.allclose(output["a"], output["b"]))

```

juampatronics:
Fixes #7697 .

### Description

### Types of changes

### DIFF
--- a/tests/test_regularization.py
+++ b/tests/test_regularization.py
@@ -20,7 +20,6 @@ from monai.utils import set_determinism
 
 
 class TestMixup(unittest.TestCase):
-
     def setUp(self) -> None:
         set_determinism(seed=0)
 
@@ -60,7 +59,6 @@ class TestMixup(unittest.TestCase):
 
 
 class TestCutMix(unittest.TestCase):
-
     def setUp(self) -> None:
         set_determinism(seed=0)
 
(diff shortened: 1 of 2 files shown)


tests/test_regularization.py
```diff
@@ -75,23 +73,32 @@ def test_cutmix(self):
             output = cutmix(sample)
             self.assertEqual(output.shape, sample.shape)
             self.assertTrue(any(not torch.allclose(sample, cutmix(sample)) for _ in range(10)))
+            # croppings are different on each application... most of the times!
+            checks = [torch.allclose(sample, cutmix(sample)) for _ in range(1000)]
```
KumoLiu, line 77: While this check might help mitigate the issue, it's not addressing the core problem. After `set_determinism(seed=0)`, we should guarantee consistent transform results every time.

cc @ericspod

ericspod: If we're setting a seed before the test then we should get the same results every time. Any behaviour dependent on a random variable should be dependent by this seed so that we can enforce determinism in cases like this. These transforms should inherit form `RandomizableTransform` but also set the random state of the internal delegate transform appropriately like [here](https://github.com/Project-MONAI/MONAI/blob/c3e4457f98e80785a946892bdd6741b9fead9113/monai/transforms/smooth_field/dictionary.py#L94). This ensures the seed value is propagated to the delegate. 

Sorry this wasn't discussed in the initial PR for these transforms, it's something we missed.

juampatronics: Thank you @ericspod for the detailed explanation of how to correctly implement it

Also, I still think that the unit test was missing the case that no transform happens when the mixing only considers one of the samples. Now this is exactly tested and it is checked that it happens as often as expected, so I would leave it as it is now.

Best

KumoLiu:
Already fixed in https://github.com/Project-MONAI/MONAI/pull/7813.

PR #7725, opened 2024-04-29
feat(img2tensorboard): plot multiple 2D images as cards in the same tag

### LINKED ISSUE #7724: Support for multiple 2D images in a single tag for TensorBoardImageHandler/plot_2d_or_3d_image
When using the [`TensorBoardImageHandler`](https://github.com/Project-MONAI/MONAI/blob/dev/monai/handlers/tensorboard_handlers.py#L289-L454) and passing a series of 2D images to the transforms, the [`plot_2d_or_3d_image`](https://github.com/Project-MONAI/MONAI/blob/dev/monai/visualize/img2tensorboard.py#L146-L211) plots each 2D image under a separate tag. When I do the same thing with a series of 3D images, they are plotted as multiple cards in the same tag.

It would be really nice if both series of 2D images and series of 3D images behaved the same way. Plotting the series under a single tag is great as it reduces the number of tags in my TensorBoard.

I know that with a series of 3D images, the [`add_animated_gif`](https://github.com/Project-MONAI/MONAI/blob/dev/monai/visualize/img2tensorboard.py#L116-L143) function is called. Which plots using [`writer._get_file_writer().add_summary()`](https://github.com/Project-MONAI/MONAI/blob/dev/monai/visualize/img2tensorboard.py#L143) for each image. However, the 2D images are plotted using [`writer.add_image()`](https://github.com/Project-MONAI/MONAI/blob/dev/monai/visualize/img2tensorboard.py#L195) for each image. I tried figuring out why this happens and couldn't figure it out.

At the moment I just attach another dimension to the front of my 2D images so that they are plotted as 3D GIFs.

**Additional context**
Here's an example of default series of 2D images behavior:
image

Here's an example of default series of 3D images behavior:
image

MattTheCuber:
Fixes #7724.

### Description
This will change the [`plot_2d_or_3d_image`](https://github.com/Project-MONAI/MONAI/blob/dev/monai/visualize/img2tensorboard.py#L146-L211) function to plot multiple 2D images under the same tag as multiple cards rather than each image as a separate tag in TensorBoard. See #7724 for more information and examples at the bottom.

This might not be the optimal way to solve this issue, let me know if you have other ideas.

### Types of changes

### DIFF
--- a/monai/visualize/img2tensorboard.py
+++ b/monai/visualize/img2tensorboard.py
@@ -192,7 +192,7 @@ def plot_2d_or_3d_image(
         dataformats = "HW"
         for j, d2 in enumerate(d[:max_channels]):
             d2 = rescale_array(d2, 0, 1)
-            writer.add_image(f"{tag}_{dataformats}_{j}", d2, step, dataformats=dataformats)
+            writer.add_image(f"{tag}_{dataformats}/{j}", d2, step, dataformats=dataformats)
         return
 
     if d.ndim >= 4:


atbenmurray:
Hi @MattTheCuber, taking a look

monai/visualize/img2tensorboard.py
```diff
@@ -192,7 +192,7 @@ def plot_2d_or_3d_image(
         dataformats = "HW"
         for j, d2 in enumerate(d[:max_channels]):
             d2 = rescale_array(d2, 0, 1)
-            writer.add_image(f"{tag}_{dataformats}_{j}", d2, step, dataformats=dataformats)
+            writer.add_image(f"{tag}_{dataformats}/{j}", d2, step, dataformats=dataformats)
```
atbenmurray, line 195: In order to preserve existing behaviour for users, this change should be controlled by a flag passed in as an argument to `plot_2d_or_3d_image`. with the default being the existing behaviour.

atbenmurray (changes requested):
Let's add a flag to control this behaviour. I'm also having a look at the tests to see if there are tests for this function that need augmenting

monai/visualize/img2tensorboard.py
MattTheCuber: The current behavior is inconsistent between different image types. Currently, multiple 3D images are plotted in the same tag while multiple 2D images are plotted in different tags. I would think consistency would be preferred over preserving the current behavior? Also, the flag should probably control all image sizes and not just 2D?

KumoLiu: Hi @MattTheCuber, we can simply add a flag for this multi-channel 2d image plt. Thanks.

MattTheCuber: I'm kind of confused what behavior you desire. Should I make a flag for controlling all plotting (2d & 3d)? Should I make flags for 2D and 3D with 2D default on and 3D default off? Should I simply add a flag for the change I made?

MattTheCuber: I agree, refactoring this would be nice. As I said previously, the current behavior is quite disjoint.

PR #7730, opened 2024-04-30
gh-7563: add config wrapper

### LINKED ISSUE #7563: Add _decorator_ keyword to Monai Bundle Configuration
When using Monai Bundle Configuration, I frequently have the need to wrap some component or function call. Examples include torch.compile or loading some weights in the instantiated model/modify a layer. 
If I now want to use torch.compile for my model, a key such previously as model#channels now needs to be changed to model#model#channels. This makes referring to the config more error-prone, less intuitive and repetitive (IMHO).  If I ever decide to not use torch.compile anymore, I now need to either use an identity lambda under model or I need to modify all configs again.

I could imagine a keyword such as \_decorator\_ or \_wrapper\_ in addition to the existing special words. Under this, a single or (maybe a list of) decorators could be defined that are added to the item in question when it is instantiated.

Keep using nested configs.

**Additional context**
Now I fully understand that there are two sides to this coin. And adding this feature has potential drawbacks such as making the config more complicated, it might lead to errors where there are unexpected decorators in place when overwriting parts of a config, and more possible support requests using this functionality.
This feature could be implemented without breaking changes AFAIK, and should be very clearly marked as experimental to gather support and provide the option to change or remove it in the future if it does not provide the desired outcomes.

If there is support for such a feature and we agree on a API, I'd be happy to work on a PR introducing this feature.

johnzielke:
Fixes #7563 .

### Description
This PR introduces a _wrapper_ special keyword that can be used to wrap/decorate a existing model key. It is intended to be used with functions that follow the decorator API in python (i.e. expect the first positional argument of the function to be the instance/function to be decorated).
This can be used to keep the hierarchy inside a config when using tools like `torch.compile` or wrapping other datasets/models.

This is still a draft, but I'd like to gather opinions on this feature and its implementation.
Open questions (non-exhaustive):
- name of the special keyword (e.g. \_wrapper\_, \_wrap\_, \_decorator\_)
- How to enable/handle this experimental feature

This feature should be handled as experimental, meaning subject to change and removal in the future. This way we can gain some experience using it and change it without worrying about backwards compatibility.

#### Breaking changes
The only breaking change I see is that configs that used the special keyword in their configs before, would have this value be interpreted differently now. In the current state, with the feature flag disabled (the default), the config will warn about usage of this special key since it's meaning could change in the future.

### Types of changes


ericspod:
Hi @johnzielke thanks for the contribution! This looks like an interesting addition with some clearly useful use cases. When you are done mark the PR as ready and we can consider it though I'd want to the feedback of a few other developers on this feature.

johnzielke:
Hey @ericspod, thank you. I fixed a few minor issues but I think it's ready for review/discussion now.

monai/utils/feature_flag.py
```diff
@@ -0,0 +1,61 @@
...
+
+from __future__ import annotations
+
+import os
+from contextlib import contextmanager
+
+FEATURE_FLAG_PREFIX = "MONAI_FEATURE_ENABLED_"
+
+
+class FeatureFlag:
```
ericspod, line 21: This class could use docstring to describe what it's doing and what it's for. There's a number of other places that we use environment variables we could potentially adapt later to use this.

tests/test_config_parser.py
```diff
@@ -124,6 +126,30 @@ def __call__(self, a, b):
...
+TEST_CASE_WRAPPER_ENABLED = [
+    {
+        "dataset": {
+            "_target_": "Dataset",
+            "data": [1, 2],
+            CONFIG_COMPONENT_KEY_WRAPPER: {"_target_": "CacheDataset", "_mode_": "callable"},
+        }
+    },
+    ["dataset", f"dataset#{CONFIG_COMPONENT_KEY_WRAPPER}"],
+    [CacheDataset, type(CacheDataset)],
```
ericspod, line 138:
```suggestion
    [CacheDataset, CacheDataset],
```
The expression `type(CacheDataset)` returns `type` which allows the test to pass but isn't specific enough to be testing what we want, right?

tests/test_config_parser.py
```diff
@@ -124,6 +126,30 @@ def __call__(self, a, b):
...
+TEST_CASE_WRAPPER_DISABLED = [
+    {
+        "dataset": {
+            "_target_": "collections.OrderedDict",
+            "data": [1, 2],
+            CONFIG_COMPONENT_KEY_WRAPPER: {"_target_": "CacheDataset", "_mode_": "callable"},
+        }
+    },
+    ["dataset", f"dataset#{CONFIG_COMPONENT_KEY_WRAPPER}"],
+    [OrderedDict, type(CacheDataset)],
```
ericspod, line 150:
```suggestion
    [OrderedDict, CacheDataset],
```

Same as above.

monai/bundle/config_item.py
```diff
@@ -19,13 +19,24 @@
...
+from typing import Any, Callable
 
 from monai.bundle.utils import EXPR_KEY
 from monai.utils import CompInitMode, ensure_tuple, first, instantiate, optional_import, run_debug, run_eval
 
 __all__ = ["ComponentLocator", "ConfigItem", "ConfigExpression", "ConfigComponent", "Instantiable"]
 
+from monai.utils.feature_flag import FeatureFlag
+
+CONFIG_COMPONENT_KEY_MODE = "_mode_"
```
ericspod, line 31: I would suggest using [StrEnum](https://github.com/Project-MONAI/MONAI/blob/daf2e714bb8a1cfa7c17d9aa7fb066dd37f5429e/monai/utils/enums.py#L69) for these variables so that they don't need `CONFIG_COMPONENT_` suffix and can be cast to `set` to produce `non_arg_keys` set below.

monai/bundle/config_item.py
```diff
@@ -210,7 +227,14 @@ class ConfigComponent(ConfigItem, Instantiable):
 
     """
 
-    non_arg_keys = {"_target_", "_disabled_", "_requires_", "_desc_", "_mode_"}
+    non_arg_keys = {
```
ericspod, line 230: So if you had a `StrEnum` object called `ConfigComponents`, you could have `non_arg_keys = set(ConfigComponents)` instead I believe.

tests/test_config_parser.py
johnzielke: The problem is that this is not a instance of this class, since it has mode: callable and therefore will just be the class. And reusing the existing test setups, this is the the workaround that I found. I agree it's not very specific, and if you think it makes sense I can introduce some custom test wrapper class instead of trying to use some pre-existing classes from another library.

ericspod: I think the test might have to instantiate things further, checking that something is an instance of `type` is going to be true for any class.

johnzielke: I now added a helper class that makes a more specific assert possible

monai/bundle/config_item.py
```diff
@@ -262,34 +282,78 @@ def resolve_args(self):
         Utility function used in `instantiate()` to resolve the arguments from current config content.
 
         """
-        return {k: v for k, v in self.get_config().items() if k not in self.non_arg_keys}
+        return {
+            k: v
+            for k, v in self.get_config().items()
+            if (k not in self.non_arg_keys)
+            or (k == ConfigComponentReservedKeys.WRAPPER and not _wrapper_feature_flag.enabled)
```
KumoLiu, line 289: I don't think we should return the key and value under `k == ConfigComponentReservedKeys.WRAPPER and not _wrapper_feature_flag.enabled` this case, it will always wrong since the k will always not be the arguments for the instance?
Correct me if I misunderstand.

monai/bundle/config_item.py
```diff
@@ -262,34 +282,78 @@ def resolve_args(self):
...
+        instantiate_kwargs = self.resolve_args()
+        instantiate_kwargs.update(kwargs)
+        wrapper = self._get_wrapper()
+        if wrapper is not None:
+            return wrapper(instantiate(modname, mode, **instantiate_kwargs))
+
+        try:
+            return instantiate(modname, mode, **instantiate_kwargs)
+        except Exception as e:
+            if _wrapper_feature_flag.enabled and self.get_id().endswith(ConfigComponentReservedKeys.WRAPPER):
```
KumoLiu, line 348: `self.get_id()` will only return a str, will this check worked?

KumoLiu (commented):
Thanks for the PR! It appears to be a useful and great enhancement.
Leave few comments inline and will also let more people review.

monai/bundle/config_item.py
```diff
@@ -262,34 +282,78 @@ def resolve_args(self):
...
-        mode = self.get_config().get("_mode_", CompInitMode.DEFAULT)
-        args = self.resolve_args()
-        args.update(kwargs)
-        return instantiate(modname, mode, **args)
+        mode = self.get_config().get(ConfigComponentReservedKeys.MODE, CompInitMode.DEFAULT)
+        instantiate_kwargs = self.resolve_args()
+        instantiate_kwargs.update(kwargs)
+        wrapper = self._get_wrapper()
+        if wrapper is not None:
+            return wrapper(instantiate(modname, mode, **instantiate_kwargs))
```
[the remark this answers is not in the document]
johnzielke: Yes, the wrapper key should be "instantiated" like any other key, and therefore will accept kwargs. So by using mode: "callable" for torch.compile, you can bind any kwargs to it, e.g.
_target_: nets.Unet
spatial_dim: 3
_wrapper_:
  _target_: torch.compile
  _mode_: callable
  dynamic: true
  fullgraph: true

docs/source/config_syntax.md
```diff
@@ -175,6 +175,63 @@ _Description:_ `_requires_`, `_disabled_`, `_desc_`, and `_mode_` are optional k
...
+{
+  "model": {
+      "_target_": "monai.networks.nets.BasicUNet",
+      "spatial_dims": 3,
+      "in_channels": 1,
+      "out_channels": 2,
+      "features": [16, 16, 32, 32, 64, 64],
+      "_wrapper_": {
+          "_target_": "torch::jit::compile",
+          "_mode_": "callable"
```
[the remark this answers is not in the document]
johnzielke: There might/probably are use cases where you would instantiate a factory here (class with a __call__ method) or have a function that returns another function (in fact torch.compile probably does this, as most decorators supporting arguments do)

monai/bundle/config_item.py
johnzielke: This is meant for backwards compatibility. What if someone is already using the key _wrapper_ in one of their configs? SInce it was not reserved before, it might have been used in normal config. It's unlikely, but I wanted to make sure to cover this case. But I'm happy to always remove it. But in that case I would probably add a regex to check for "_\w+_" and make all keys of this pattern reserved for the future.

monai/bundle/config_item.py
johnzielke:
```
>>> "test::_wrapper_".endswith(ConfigComponentReservedKeys.WRAPPER)
True
```
Yes this should work, this evaluates to true for me

docs/source/config_syntax.md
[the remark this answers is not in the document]
KumoLiu: Hi @johnzielke, after discuss with @Nic-Ma offline, we find that such wrapper can be easily achieved by using code like this:
```

config = {
  "model_base": {
      "_target_": "monai.networks.nets.BasicUNet",
      "spatial_dims": 3,
      "in_channels": 1,
      "out_channels": 2,
      "features": [16, 16, 32, 32, 64, 64]
  },
  "model": "$torch.compile(@model_base)"

}
parser.parse()
net = parser.get_parsed_content("model")
print(net)
```

Does this meet your needs?
Like how we use DDP here:
https://github.com/Project-MONAI/model-zoo/blob/dev/models/spleen_ct_segmentation/configs/multi_gpu_train.json#L3

cc @Nic-Ma @ericspod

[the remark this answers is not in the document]
johnzielke: Yes that patterns works in many cases, but it requires you to move definitions to other keys, which might be undesirable for multiple reasons:
- A lot of references to this in code and overriding configs need to be adjusted if this has not been planned from the beginning
- If the thing you are adding is used as a parameter to a parent instantiation using _target_, you cannot just add the "model_base" key to the same level if that parent class does not handle other kwargs gracefully.
- If your wrapping function has parameters (for example the dynamic shapes in torch.compile), you need to specify those in a single string making it harder to modify these with configs. Of course you could also introduce another config called model_decorator, and then have model be `@model_decorator(@model_base)`, but that adds a lot of visual noise in my opinion

I'll try to compile some other use cases later, but one I could see would be to wrap existing Datasets in CacheDatasets or similar. In that case you would have often have a dictionary of train, val, test and the datasets. Of course the same pattern can be applied again, but it makes the configs harder to read in my opinion.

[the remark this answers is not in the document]
ibro45: Hi all, is there still interest in adding this to MONAI?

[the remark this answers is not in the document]
ericspod: I would like to come back to this as it looks like a good feature, we all seem to have gotten distracted on this one! As mentioned there's a few use cases where the proposed mechanism works out but we should consider the added complexity. As I mentioned the current pattern (`"model": "$torch.compile(@model_base)"`) described above works for me and can be used to apply multiple wrapper functions, the proposed one as it is permits only one I think. The current conflicts look minor to resolve so if we want to revive this we'll look at outstanding issues and comments.

PR #7738, opened 2024-05-03
Add direct links to github source code to docs

johnzielke:
### Description

This adds direct links to the respective source code of classes, methods, etc. to the docs.
In my opinion these nicer and more useful than viewing the source code in copy in the docs.
image

I currently added it in addition to the links to the source files copied into the docs. If desired this could also be done instead of.
I also used a github logo instead of another `[source]` link.
I am not sure whether there is any easier way to change that into the logo than the way I have done, but this one seems to work.

### Types of changes

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -2,3 +2,19 @@
 body{font-family:'Roboto',sans-serif;}.wy-menu-vertical p.caption{color:#7cccc7;}
 *{font-variant-ligatures: none;}.autoclasstoc td {padding:0.2rem;line-height:normal;}
 dl.field-list>dt{word-break: normal}
+
+
+/* Instead of showing the [source] link for the github link, show a github icon */
+/* Give the empty span some content to make the background image show up */
+a.external > span.viewcode-link:after {
+    content: "\00a0\00a0\00a0";
+}
+a.external > span.viewcode-link {
+    background: url(https://github.com/favicon.ico);
+    background-size: contain;
+    background-repeat: no-repeat;
+}
+/* Hide the [source] text */
+a.external > span.viewcode-link > span.pre {
+    display:none;
+}
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -2,19 +2,3 @@
 body{font-family:'Roboto',sans-serif;}.wy-menu-vertical p.caption{color:#7cccc7;}
 *{font-variant-ligatures: none;}.autoclasstoc td {padding:0.2rem;line-height:normal;}
 dl.field-list>dt{word-break: normal}
-
-
-/* Instead of showing the [source] link for the github link, show a github icon */
-/* Give the empty span some content to make the background image show up */
-a.external > span.viewcode-link:after {
-    content: "\00a0\00a0\00a0";
-}
-a.external > span.viewcode-link {
-    background: url(https://github.com/favicon.ico);
-    background-size: contain;
-    background-repeat: no-repeat;
-}
-/* Hide the [source] text */
-a.external > span.viewcode-link > span.pre {
-    display:none;
-}
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -177,11 +177,14 @@ repository = os.environ.get("GITHUB_REPOSITORY", DEFAULT_REPOSITORY)
 base_code_url = f"https://github.com/{repository}/blob/{git_ref}"
 MODULE_ROOT_FOLDER = "monai"
 
+
 # Adjusted from https://github.com/python-websockets/websockets/blob/main/docs/conf.py
 def linkcode_resolve(domain, info):
     if domain != "py":
-        raise ValueError(f"expected domain to be 'py', got {domain}."
-                         "Please adjust linkcode_resolve to either handle this domain or ignore it.")
+        raise ValueError(
+            f"expected domain to be 'py', got {domain}."
+            "Please adjust linkcode_resolve to either handle this domain or ignore it."
+        )
 
     mod = importlib.import_module(info["module"])
     if "." in info["fullname"]:
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -13,6 +13,8 @@
 import os
 import subprocess
 import sys
+import importlib
+import inspect
 
 sys.path.insert(0, os.path.abspath(".."))
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..")))
@@ -96,6 +98,7 @@ extensions = [
     "sphinx.ext.napoleon",
     "sphinx.ext.autodoc",
     "sphinx.ext.viewcode",
+    "sphinx.ext.linkcode",
     "sphinx.ext.autosectionlabel",
     "sphinx.ext.autosummary",
     "sphinx_autodoc_typehints",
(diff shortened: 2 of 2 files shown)


docs/_static/custom.css
```diff
@@ -2,3 +2,19 @@
 body{font-family:'Roboto',sans-serif;}.wy-menu-vertical p.caption{color:#7cccc7;}
 *{font-variant-ligatures: none;}.autoclasstoc td {padding:0.2rem;line-height:normal;}
 dl.field-list>dt{word-break: normal}
+
+
+/* Instead of showing the [source] link for the github link, show a github icon */
```
KumoLiu, line 7: Looks like the [source] link is more common used in other repo, such as PyTorch? What do you think?

cc @ericspod @Nic-Ma

johnzielke: I can find some examples of sphinx.ext.linkcode later

johnzielke:
In response to @KumoLiu and as a general question:
The 
If we link to the github source, do we want to even keep the existing "viewcode" extension, i.e. the copied version of the source code in the docs? I kept this mainly for backwards compatibility, IMHO it makes sense to switch exclusively to the links in github.
Or would you rather just have the **[source]** link go directly to github and remove "viewcode"?

I did a quick search on what other repos use this style of source code linking (using [this](https://github.com/search?q=sphinx.ext.linkcode+path%3A%2F%28%5E%7C%5C%2F%29conf%5C.py%24%2F&type=code) search), and examples include:
- scikit-learn
- jax
- numpy

I mainly styled the link to github as a github icon to differentiate it from the other **[source]** link (from viewcode). If we use the github link exclusively, it's personal preference IMO. I think that the icon uses up a little less space and makes it clear you are linking to an "external" site, but on the other hand **[source]** is widely used already. And changing to the icon is a bit hacky (which is also why I think almost no one has bothered doing it so far) since the linkcode extension does not easily support it.

docs/_static/custom.css
ericspod: I prefer linking to the file in the repo, it's cleaner and lets one browse to other places quickly given how Github is adding more dynamic features to code browsing. I've often found readthedocs and other pages which never link to the project's repo anywhere which I find really annoying when I just want to get to the code so this is helpful for that.

johnzielke: As a general comment: The line of code or rather the whole change in custom.css is just so that we get one link to github using an icon and one linking to the current code copy in the docs using the usual [source] link. If we wouldn't have this change in custom.css, there would be [source][source] in this location.

ericspod:
I think this is an improvement in general, with two things to discuss:

1. Should we keep the existing generated source code view? If not we remove "sphinx.ext.viewcode" from the Sphinx config file so that only the Github link shows up. This is slightly less helpful if people want to generate offline docs but this is a minor point, I just think it's confusing having both links.
2. The choice of line ranges to link to isn't always correct. For example, going to `docs/build/html/networks.html#unet` in the generated docs and clicking the Github link wil take one to the correct line range for UNet, but clicking on `docs/build/html/networks.html#resnet` produces an incorrect line range. It seems to be just below the docstring but other things I've seen have started mid-string rather than just below. Let's double check the algorithm here for computing that range. 

Thanks @johnzielke for this. @zephyrie please a look to see if this makes sense for our documentation, thanks.

johnzielke:
Thanks @ericspod. I agree with both points. Do you know of any people creating offline docs? If we really needed those, we can of course create a Env variable or similar to change what we generate.

Regarding the second point, if there is agreement to merge this I'll have a look at some of the mentioned repositories again and see if there are some cases that this function doesn't cover (decorators might be one of those). Another thing I have to check up on is how the mapping to source code positions might be improved by newer python versions.
And thanks for the examples, I'll check those.
While of course we should fix as many errors as possible, IMO a small error in the mapping is no deal breaker, if it's just one or two lines.

PS: I was considering creating a small pip package (sphinx extension) that does this and would prevent having to copy this into all repos needing this, since it'squite a common use case nowadays. But I'm not sure how the sentiment for an extra dev dependency for something like this would be in light of recent events.

ericspod:
> Thanks @ericspod. I agree with both points. Do you know of any people creating offline docs? If we really needed those, we can of course create a Env variable or similar to change what we generate.

I don't think people are but some feedback from others would be helpful before making this change.

> Regarding the second point, if there is agreement to merge this I'll have a look at some of the mentioned repositories again and see if there are some cases that this function doesn't cover (decorators might be one of those). Another thing I have to check up on is how the mapping to source code positions might be improved by newer python versions. And thanks for the examples, I'll check those. While of course we should fix as many errors as possible, IMO a small error in the mapping is no deal breaker, if it's just one or two lines.

The example I gave for an incorrect line range wasn't using decorators in the definition, so I suspect there's just something mildly wrong with the algorithm that can be fixed. Perhaps someone else has it in their versions of this plugin so yeah checking around would help. If not we can still merge this since I think it's an improvement over the reformatted code.

> PS: I was considering creating a small pip package (sphinx extension) that does this and would prevent having to copy this into all repos needing this, since it'squite a common use case nowadays. But I'm not sure how the sentiment for an extra dev dependency for something like this would be in light of recent events with xz.

This would be a dependency of the sphinx environment and not for MONAI so much less impactful. If it's something other people would want it could be a good project to start. New dependencies and libraries are unavoidable so one famous example of an attempted exploit shouldn't put us off. We should be cautious and have some way of verifying contributions, but that's a community-wide issue.

johnzielke:
Ok great, than I'll wait for feedback from the rest.

> The example I gave for an incorrect line range wasn't using decorators in the definition, so I suspect there's just something mildly wrong with the algorithm that can be fixed. Perhaps someone else has it in their versions of this plugin so yeah checking around would help. If not we can still merge this since I think it's an improvement over the reformatted code.

Thanks for noticing, in the end the error wasn't in the algorithm but in the default branch that I hard-coded.

> This would be a dependency of the sphinx environment and not for MONAI so much less impactful. If it's something other people would want it could be a good project to start. New dependencies and libraries are unavoidable so one famous example of an attempted exploit shouldn't put us off. We should be cautious and have some way of verifying contributions, but that's a community-wide issue.

I will try to get a package like this available. Once it's ready, I'll open up another PR here to switch to that, but in the meantime, the simple implementation here should suffice.

zephyrie:
I haven't seen any demand for offline documentation, so removing the viewcode extension seems reasonable, and any minor errors in mapping line ranges can be addressed.

And I think linking directly to GitHub source code is a great idea and it offers a better user experience overall!

ericspod (approved):
I think this looks good now and the links go to the right places for me. Thanks all!

johnzielke:
@KumoLiu On https://docs.monai.io/en/latest/metrics.html I am now not seeing any more links to source anymore on "latest". Is this normal or is there some error in the build? I was unable to find the CI job deploying the pages.

KumoLiu:
> @KumoLiu On https://docs.monai.io/en/latest/metrics.html I am now not seeing any more links to source anymore on "latest". Is this normal or is there some error in the build? I was unable to find the CI job deploying the pages.

here is the job link: https://github.com/Project-MONAI/MONAI/actions/runs/9069259715/job/24918422082

johnzielke:
I think the issue is with read the docs. See #7779

PR #7745, opened 2024-05-07
Enhance logging logic in `ConfigWorkflow`

KumoLiu:
When using nvflare with monai bundle, the logging logic may be overrided by the logging logic in the bundle. 
Add an option to disable logging logic in the bundle.

### Types of changes

### DIFF
--- a/monai/bundle/workflows.py
+++ b/monai/bundle/workflows.py
@@ -318,7 +318,7 @@ class ConfigWorkflow(BundleWorkflow):
                     raise FileNotFoundError(f"Cannot find the logging config file: {logging_file}.")
             else:
                 logger.info(f"Setting logging properties based on config: {logging_file}.")
-                fileConfig(logging_file, disable_existing_loggers=False)
+                fileConfig(logging_file, disable_existing_loggers=False)  # type: ignore[arg-type]
 
         self.parser = ConfigParser()
         self.parser.read_config(f=config_file)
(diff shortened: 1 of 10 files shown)


ericspod (approved):
Looks good with some minor comments.

holgerroth (approved):

Nic-Ma (approved):

monai/bundle/workflows.py
```diff
@@ -307,15 +308,17 @@ def __init__(
         super().__init__(workflow_type=workflow_type, meta_file=meta_file, properties_path=properties_path)
         self.config_root_path = config_root_path
         logging_file = str(self.config_root_path / "logging.conf") if logging_file is None else logging_file
-        if logging_file is not None:
+        if not logging_file:
```
YuanTingHsieh, line 311: We might just want explicit say "if logging file is False"?

Otherwise if logging_file="" this if statement also evaluated to true

monai/bundle/workflows.py
```diff
@@ -307,15 +308,17 @@ def __init__(
...
+            logger.warn(f"Logging file is set to {logging_file}, skipping logging.")
+        elif logging_file is not None:
```
YuanTingHsieh, line 313: I think After L310 the logging file cannot be None anymore.

So we could just put else here?

monai/bundle/workflows.py
```diff
@@ -307,15 +308,17 @@ def __init__(
...
             if not os.path.isfile(logging_file):
                 if logging_file == str(self.config_root_path / "logging.conf"):
                     logger.warn(f"Default logging file in {logging_file} does not exist, skipping logging.")
                 else:
                     raise FileNotFoundError(f"Cannot find the logging config file: {logging_file}.")
             else:
                 logger.info(f"Setting logging properties based on config: {logging_file}.")
-                fileConfig(logging_file, disable_existing_loggers=False)
+                fileConfig(str(logging_file), disable_existing_loggers=False)
```
YuanTingHsieh, line 321: Do we need this change? But this is fine either way.

KumoLiu: Yes, will fix the mypy error in the ci.

YuanTingHsieh (approved):

PR #7749, opened 2024-05-08
Add activation parameter to ResNet

Pkaps25:
Fixes #7653 .

### Description
Includes an `act` parameter to `ResNet` and its submodules to allow for passing the `inplace` param. 

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [x] New tests added to cover the changes.
- [x] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [x] Quick tests passed locally by running `./runtests.sh --quick --unittests  --disttests`.
- [x] In-line docstrings updated.
- [x] Documentation updated, tested `make html` command in the `docs/` folder.

PR #7752, opened 2024-05-09
Add norm param to ResNet

### LINKED ISSUE #7294: Implement the ability to use layernorm with resnet models
Batch norm can be problematic with small batch sizes and certain applications due to computing running batch statistics. Layernorm does not have this problem but it is not available to use in with monai's resnet models:

```
from monai.networks.nets.resnet import ...
```

Implement the ability to use layer norm with resnet models

I have tried using other models which support layer norm such as densenet but it seems to be broken also. See https://github.com/Project-MONAI/MONAI/issues/7293

Pkaps25:
Fixes #7294  .

### Description
Adds a `norm` param to ResNet

### Types of changes

### DIFF
--- a/monai/networks/nets/daf3d.py
+++ b/monai/networks/nets/daf3d.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 from collections import OrderedDict
 from collections.abc import Callable, Sequence
+from functools import partial
 
 import torch
 import torch.nn as nn
@@ -25,6 +26,7 @@ from monai.networks.blocks.backbone_fpn_utils import BackboneWithFPN
 from monai.networks.blocks.convolutions import Convolution
 from monai.networks.blocks.feature_pyramid_network import ExtraFPNBlock, FeaturePyramidNetwork
 from monai.networks.layers.factories import Conv, Norm
+from monai.networks.layers.utils import get_norm_layer
 from monai.networks.nets.resnet import ResNet, ResNetBottleneck
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


ericspod:
Hi @Pkaps25 thanks for this, is it ready to review? We should add at least one test for the new argument in Daf3D however.

Pkaps25:
> Hi @Pkaps25 thanks for this, is it ready to review? We should add at least one test for the new argument in Daf3D however.

Daf3D doesn't currently have a `norm` parameter and it looks like adding that in might be tricky since certain parts of the network expect group norm and others use ResNet's batch norm. I'd like to avoid adding that param in this PR, so short of that, would the existing tests cover my change?

KumoLiu:
Hi @Pkaps25, after we discussed from the dev meeting.
One thing I overlooked in this [PR](https://github.com/Project-MONAI/MONAI/pull/7749/files) is whether the change of the name of act will affect the weights loading of the original model, can you help check that, if so, maybe we need to revert the name change for the variable and add an alias.

Thanks!

Pkaps25:
Ah, nice catch. I will take a look.

Pkaps25:
> Hi @Pkaps25, after we discussed from the dev meeting. One thing I overlooked in this [PR](https://github.com/Project-MONAI/MONAI/pull/7749/files) is whether the change of the name of act will affect the weights loading of the original model, can you help check that, if so, maybe we need to revert the name change for the variable and add an alias.
> 
> Thanks!

Hi @KumoLiu, I trained ResNet with latest MONAI and loaded the model state using my branch. The model loaded successfully, had the same layers, and a forward pass with a random input produced the same result on both branches. Let me know if that's a sufficient check and if we need to add tests for this compatibility in the future

KumoLiu:
> > Hi @Pkaps25, after we discussed from the dev meeting. One thing I overlooked in this [PR](https://github.com/Project-MONAI/MONAI/pull/7749/files) is whether the change of the name of act will affect the weights loading of the original model, can you help check that, if so, maybe we need to revert the name change for the variable and add an alias.
> > Thanks!
> 
> Hi @KumoLiu, I trained ResNet with latest MONAI and loaded the model state using my branch. The model loaded successfully, had the same layers, and a forward pass with a random input produced the same result on both branches. Let me know if that's a sufficient check and if we need to add tests for this compatibility in the future

Thanks to the quick verification, and make sense it works since the activation function usually has no parameters to learn in the neural network. I guess we don't need to add a test for this. But if the norm layer name changed, it should be worth to add the test. What do you think? cc @ericspod

Pkaps25:
> > > Hi @Pkaps25, after we discussed from the dev meeting. One thing I overlooked in this [PR](https://github.com/Project-MONAI/MONAI/pull/7749/files) is whether the change of the name of act will affect the weights loading of the original model, can you help check that, if so, maybe we need to revert the name change for the variable and add an alias.
> > > Thanks!
> > 
> > 
> > Hi @KumoLiu, I trained ResNet with latest MONAI and loaded the model state using my branch. The model loaded successfully, had the same layers, and a forward pass with a random input produced the same result on both branches. Let me know if that's a sufficient check and if we need to add tests for this compatibility in the future
> 
> Thanks to the quick verification, and make sense it works since the activation function usually has no parameters to learn in the neural network. I guess we don't need to add a test for this. But if the norm layer name changed, it should be worth to add the test. What do you think? cc @ericspod

Agreed that if the name changes we should add that test, but I don't change the name in this PR.

monai/networks/nets/daf3d.py
```diff
@@ -170,27 +172,31 @@ class Daf3dResNetBottleneck(ResNetBottleneck):
...
+                norm_layer(channels=planes * self.expansion),
             )
 
         super().__init__(in_planes, planes, spatial_dims, stride, downsample)
 
         # change norm from batch to group norm
-        self.bn1 = norm_type(num_groups=32, num_channels=planes)
-        self.bn2 = norm_type(num_groups=32, num_channels=planes)
-        self.bn3 = norm_type(num_groups=32, num_channels=planes * self.expansion)
+        self.bn1 = norm_layer(channels=planes)
```
KumoLiu, line 197: Looks like the norm arg added in these two blocks will be useless since it can not be used in `Daf3d`. Do we still need add them here?

KumoLiu (commented):
The norm type added in ResNet look good to me. But have a concern regarding the norm layer added in daf3d added inline. Thanks!

ericspod:
> Thanks to the quick verification, and make sense it works since the activation function usually has no parameters to learn in the neural network. I guess we don't need to add a test for this. But if the norm layer name changed, it should be worth to add the test. What do you think? cc @ericspod

This looks good then, I thought that a layer with no parameters wouldn't show up in the state dict but it's best to double check.

ericspod (approved):
i think things are fine if all comments are addressed.

PR #7754, opened 2024-05-10
Neater use off nn.Sequential in controlnet

marksgraham:
Part of #7227  .

### Description
Tidies up some of controlnet

### Types of changes

### DIFF
--- a/monai/networks/nets/controlnet.py
+++ b/monai/networks/nets/controlnet.py
@@ -141,7 +141,6 @@ class ControlNet(nn.Module):
         num_class_embeds: if specified (as an int), then this model will be class-conditional with `num_class_embeds`
             classes.
         upcast_attention: if True, upcast attention operations to full precision.
-        use_flash_attention: if True, use flash attention for a memory efficient attention mechanism.
         conditioning_embedding_in_channels: number of input channels for the conditioning embedding.
         conditioning_embedding_num_channels: number of channels for the blocks in the conditioning embedding.
     """
@@ -162,7 +161,6 @@ class ControlNet(nn.Module):
         cross_attention_dim: int | None = None,
         num_class_embeds: int | None = None,
         upcast_attention: bool = False,
-        use_flash_attention: bool = False,
         conditioning_embedding_in_channels: int = 1,
         conditioning_embedding_num_channels: Sequence[int] = (16, 32, 96, 256),
     ) -> None:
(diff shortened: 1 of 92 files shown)


monai/networks/nets/controlnet.py
```diff
@@ -34,84 +34,71 @@
...
-            out_channels=channels[0],
-            strides=1,
-            kernel_size=3,
-            padding=1,
-            conv_only=True,
-        )
-
-        self.blocks = nn.ModuleList([])
-
+        convs = [
```
KumoLiu, line 50: Will this name changing here influence the weight loading?

monai/networks/nets/controlnet.py
```diff
@@ -34,84 +34,71 @@
...
+            zero_module(
+                Convolution(
+                    spatial_dims=spatial_dims,
+                    in_channels=channels[-1],
+                    out_channels=out_channels,
+                    strides=1,
+                    kernel_size=3,
+                    padding=1,
+                    adn_ordering="A",
+                    act="SWISH",
```
KumoLiu, line 97: Previous the `conv_out` didn't follow the activation layer, is this adding expected here?
Also will the name changing influence anything here?

monai/networks/nets/controlnet.py
KumoLiu, line 1: I just noticed the license information in the header of the file, do we need to add L17-L30, if the license is Apache2.0?
https://github.com/Project-MONAI/MONAI/blob/56373f6df602cae0499bbe62f665c3c7fb5c2be9/monai/networks/nets/controlnet.py#L13-L29

cc @ericspod @Nic-Ma @zephyrie

monai/networks/nets/controlnet.py
marksgraham: hm i just checked and it does. i could implement `load_old_state_dict` here, but maybe it isn't worth placing an extra burden on users when the gains are quite small: just a slight reduction in the amount of code, rather than any functional differences. what do you think?

KumoLiu: Yes, no need to such a small update, but do we have a `load_old_state_dict` for `ControlNet`?

KumoLiu: I would suggest against this change but also add activation layer in the `Convolution` which can also make code cleaner. What do you think?
cc @ericspod

marksgraham: sorry i have done some tests and realised we do need to add a `load_old_state dict`. i will add it + a test for compatibility. I'll also make the suggested change with the activation layer

monai/networks/nets/controlnet.py
KumoLiu: Let's merge this PR first and discuss the license info issue later.

KumoLiu:
Hi @marksgraham,
For the CI error here: https://github.com/Project-MONAI/MONAI/actions/runs/9110192660/job/25044692031?pr=7754#step:12:21194
Could you please help cherry pick the update from this PR: https://github.com/Project-MONAI/MONAI/pull/7780?

Thanks!

PR #7769, opened 2024-05-14
Add a mapping function in image_reader.py and image_writer.py

### LINKED ISSUE #7557: Update the SaveImage transform to support saving input-output mapping.
When I run a segmentation bundle like spleen segmentation, I would like to have a new json/yaml/csv file to explicitly show the input-output mapping between images and labels. For example if the input and output folders are like:
```
--input
   --input_0.nii.gz
   --input_1.nii.gz
   -- ...
   --input_n.nii.gz

--output
   --input_0/input_0_seg.nii.gz
   --input_1/input_1_seg.nii.gz
   --...
   --input_n/input_n_seg.nii.gz
```

A json file with content shown below should be generated if the parameter is specified:
```
[
    {image: ["/path/to/image/input/input_0.nii.gz"], label: ["/path/to/seg/output/input_0/input_0_seg.nii.gz"]},
    {image: ["/path/to/image/input/input_1.nii.gz"], label: ["/path/to/seg/output/input_1/input_1_seg.nii.gz"]},
    ...,
    {image: ["/path/to/image/input/input_n.nii.gz"], label: ["/path/to/seg/output/input_n/input_n_seg.nii.gz"]},
]
```
The `image` and `label` parameters are set to lists in case there are multiple inputs/outputs.

staydelight:
Add a function to create a JSON file that maps input and output paths.

Fixes #7557  .

### Description

### Types of changes


monai/data/image_reader.py
```diff
@@ -148,6 +149,25 @@ def _stack_images(image_list: list, meta_dict: dict):
     return np.stack(image_list, axis=0)
 
 
+def update_json(input_file=None, output_file=None):
+        record_path = "img-label.json"
```
KumoLiu, line 153: Hi @staydelight, thank you for the PR. I have a few concerns:
1. Do we really need to make any changes to read images? It seems that we can already support reading paired data using LoadImaged.
2. I suggest adding a flag to `SaveImage` to allow users to choose whether or not to write a JSON file.
3. Can you clarify the purpose of the record_path? We can directly obtain the label path from the input of SaveImage, and for the image path, we can retrieve it from the metadata of the data (since we have introduced MetaTensor).

Let me know if you need further clarification on any of these points.

ericspod: Hi @staydelight thanks as well but I also have concerns similar to @KumoLiu. In the general case users are not going to want these JSON files being generated so I don't think we should add this to the readers and writers at all. This is very use-case specific so writing some custom transform or doing things in a different way would be the solution. 

One other thing to mention is that it's not thread-safe since multiple parallel transforms may be reading/writing the file at the same time. In this case you also cannot rely on the ordering of LoadImage/SaveImage operations to ensure you match the right input with output.

As mentioned you can access original paths with the metadata present in the MetaTensor objects:

```python
trans = monai.transforms.LoadImaged(keys="image")
d = trans({"image": "/path/to/file.nii.gz"})
print(d["image"].meta["filename_or_obj"])
```

It should be possible to access this value in the postprocessing transform sequence were SaveImage is used since the network output should be a MetaTensor with these values included. You should be able to define a transform after SaveImage which logs these values to a file. This would be the much more modular approach versus adding specific code to the loader/saver classes, so I'd strongly suggest investigating how best to go about that.

staydelight: Hi @ericspod @KumoLiu,Thank you for your advice. According to what you said, since the network output should be a MetaTensor, is it possible to add a `save_log` function to `SaveImage` like this:

```python
self.log_data.append({
    "input": meta_data.get("filename_or_obj", "(unknown)"),
    "output": filename
})


def save_log(self):
    try:
        with open(self.log_json_path, 'r') as f:
            existing_log_data = json.load(f)
    except FileNotFoundError:
        existing_log_data = []

    with open(self.log_json_path, 'w') as f:
        existing_log_data.extend(self.log_data)
        json.dump(existing_log_data, f, indent=4)
```

KumoLiu: It seems like a workable solution, but it involves repeatedly reading and writing JSON files. I haven't thought of a better way yet. We could add a utility function, but it might not be very effective. @ericspod, do you have any better suggestions for saving a mapping?

BTW, the save path can also be record in the meta data:
https://github.com/Project-MONAI/MONAI/blob/4029c422f1da84cdf5e4ce546dbff33245cbe025/monai/transforms/io/array.py#L507

ericspod: This is a similar sort of change we discussed with the input-output mapping file generation in another PR. I think this would be better implemented in a transform which will appear after SaveImage(d) in your pipeline which will handle saving to file(s). The code here can be put into that and keep SaveImage(d) focused on saving the image data only. You would also have to take into account multiprocessing when doing this and write to separate files for each subprocess, this implementation as-is contains a race condition if multiple processes attempt to save to the same file.

staydelight: Hi @KumoLiu @ericspod, I am writing a test file for `MappingJson`. I tried to create a numpy array to simulate an image and mimic the actual usage where `LoadImage` and `SaveImage` would be used to obtain the contents of meta_data. However, during this process, I needed to import external dependencies. In min_tests.py, any files that require importing external dependencies are skipped. Therefore, I would like to ask if I can add my test file to the skipped list, or should I write a test file that does not require importing external dependencies?

ericspod: I would add your file to the skipped list, otherwise you can use the same [pattern with the nrrd tests](https://github.com/Project-MONAI/MONAI/blob/7a8680e84457cb374859639ab6a078313da85926/tests/test_nrrd_reader.py#L49) to make the test optional.

monai/transforms/io/array.py
```diff
@@ -514,3 +515,46 @@ def __call__(
...
+
+    def __call__(self, img: MetaTensor):
+        """
+        Args:
+            img (MetaTensor): The input image with metadata.
+        """
+        if "saved_to" not in img.meta:
+            raise KeyError("Missing 'saved_to' key in metadata. Check SaveImage savepath_in_metadict.")
+
+        input_path = img.meta["filename_or_obj"]
```
binliunls, line 557: For the string `filename_or_obj`, I would recommand to use the enum [ImageMetaKey.FILENAME_OR_OBJ](https://github.com/Project-MONAI/MONAI/blob/2e53df78e580131046dc8db7f7638063db1f5045/monai/utils/misc.py#L577). For the `saved_to` string, @KumoLiu  I would suggest to create a new enum to record metadata keys that are optional. What do you think about this?

Thanks  
Bin

monai/transforms/io/array.py
```diff
@@ -514,3 +515,46 @@ def __call__(
...
+
+    def write_json(self, input_path: str, output_path: str):
+        """
+        Args:
+            input_path (str): The path of the input image file.
+            output_path (str): The path of the output image file.
+        """
+        log_data = {"input": input_path, "output": output_path}
+        try:
+            with self.mapping_json_path.open("r") as f:
```
binliunls, line 539: I think this part is unsafe for the multiprocess environment. Could you please add a multiprocess test case reading/writing the same file here? If the test turns out failure, I think we should consider to use the [filelock](https://py-filelock.readthedocs.io/en/latest/index.html) to make this one safe. @ericspod  Please correct me if I am wrong about this one.

Thanks  
Bin

ericspod: If nothing is done to either ensure the file names are unique for every thread/process then yes writing to a file by multiple processes is going to be unsafe. Very fast writes can usually be done atomically so you don't often see the issue, but they are still unsafe. 

The solution is to ensure files are locked when written to, this requires a third party library like filelock (which seems well supported so we could consider adding it to requirements), writing to a database like sqlite, or writing to unique files that can be combined later. 

My thought on how output transforms would be used was that there would be a single thread/process doing the IO and so there would be no race issues, or if multiples they'd be guaranteed to write to unique files. This issue here of wanting to accumulate a JSON file from multiple processes is unique I think in MONAI.

staydelight: Hi, @binliunls @ericspod . I have added a multiprocess test case. The original `MappingJson` indeed results in failures. Therefore, I added filelock. What do you think about the new one?

monai/transforms/io/array.py
staydelight: If @KumoLiu thinks this is feasible, I will modify `MappingJson` and use ImageMetaKey.

monai/transforms/io/array.py
```diff
@@ -514,3 +516,49 @@ def __call__(
             "    https://docs.monai.io/en/latest/installation.html#installing-the-recommended-dependencies.\n"
             f"   The current registered writers for {self.output_ext}: {self.writers}.\n{msg}"
         )
+
+
+class MappingJson(Transform):
```
ericspod, line 521: I think `WriteFileMapping` or some other name describes what the class is doing a little better, specifically it states that it writes something.

monai/transforms/io/array.py
```diff
@@ -514,3 +516,49 @@ def __call__(
...
+            with self.mapping_json_path.open("w") as f:
+                json.dump(existing_log_data, f, indent=4)
+
+    def __call__(self, img: MetaTensor):
+        """
+        Args:
+            img (MetaTensor): The input image with metadata.
+        """
+        if "saved_to" not in img.meta:
+            raise KeyError("Missing 'saved_to' key in metadata. Check SaveImage savepath_in_metadict.")
```
ericspod, line 560:
```suggestion
            raise KeyError("Missing 'saved_to' key in metadata. Check SaveImage argument savepath_in_metadict is True.")
```

tests/test_mapping_json.py
```diff
@@ -0,0 +1,112 @@
...
+            cause_exception = cm.exception.__cause__
+            self.assertIsInstance(cause_exception, KeyError)
+            self.assertIn(
+                "Missing 'saved_to' key in metadata. Check SaveImage savepath_in_metadict.", str(cause_exception)
+            )
+
+    def test_multiprocess_mapping_json(self):
+        num_processes, num_images = 3, 50
+
+        with multiprocessing.Pool(processes=num_processes) as pool:
```
ericspod, line 96: It might good to test with `DataLoader` as well with multiple workers to ensure that the transform behaves under that environment. You would have to check with a data loader that would produce more than one batch so that multiple workers are used.

tests/test_mapping_json.py
```diff
@@ -0,0 +1,112 @@
...
+
+        with open(self.mapping_json_path) as f:
+            mapping_data = json.load(f)
+
+        self.assertEqual(len(mapping_data), num_images, f"Expected {num_images} entries, but got {len(mapping_data)}")
+        unique_entries = {tuple(sorted(entry.items())) for entry in mapping_data}
+        self.assertEqual(len(mapping_data), len(unique_entries), "Duplicate entries exist")
+        for entry in mapping_data:
+            self.assertIn("input", entry, "Entry missing 'input' key")
+            self.assertIn("output", entry, "Entry missing 'output' key")
```
ericspod, line 108: Another possible way of checking this is to generate a json file with only one process, then do things again with multiple processes and compare the contents.

ericspod:
I had a few minor added comments but otherwise it looks good. If you could please address what I and @KumoLiu have mentioned we can get this through.

monai/transforms/io/array.py
KumoLiu: Agreed, and for `saved_to`, we can add it in https://github.com/Project-MONAI/MONAI/blob/e85580af2267404ff0f022e5372a44e8effe6316/monai/utils/enums.py#L536

staydelight:
> I had a few minor added comments but otherwise it looks good. If you could please address what I and @KumoLiu have mentioned we can get this through.

@ericspod @KumoLiu All comments have been addressed.If there are any additioanal changes,please let me know.

ericspod:
> @ericspod @KumoLiu All comments have been addressed.If there are any additioanal changes,please let me know.

Do we not also need a dictionary version of this transform? If not we can approve this as it is now.

KumoLiu:
> Do we not also need a dictionary version of this transform? If not we can approve this as it is now.

Adding a dictionary version would be better since dictionary version pipeline may be more commonly used.

ericspod:
> > Do we not also need a dictionary version of this transform? If not we can approve this as it is now.
> 
> Adding a dictionary version would be better since dictionary version pipeline may be more commonly used.

The dictionary version can be something very simple I think, like this:

```python
class WriteFileMappingd(MapTransform):
    def __init__(self, keys: KeysCollection, mapping_file_path: Path | str = "mapping.json", allow_missing_keys: bool = False) -> None:
        self.mapping = WriteFileMapping(mapping_file_path)

    def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> dict[Hashable, NdarrayOrTensor]:
        d = dict(data)
            d[key] = self.mapping(d[key])
        return d
```

This would needed added tests as well.

KumoLiu:
Hi @staydelight, just a quick note—if you want this new feature included in version 1.4, we only have one or two weeks left to add new features.

staydelight:
> > > Do we not also need a dictionary version of this transform? If not we can approve this as it is now.
> > 
> > 
> > Adding a dictionary version would be better since dictionary version pipeline may be more commonly used.
> 
> The dictionary version can be something very simple I think, like this:
> 
> ```python
> class WriteFileMappingd(MapTransform):
>     def __init__(self, keys: KeysCollection, mapping_file_path: Path | str = "mapping.json", allow_missing_keys: bool = False) -> None:
>         super().__init__(keys, allow_missing_keys)
>         self.mapping = WriteFileMapping(mapping_file_path)
> 
>     def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> dict[Hashable, NdarrayOrTensor]:
>         d = dict(data)
>         for key in self.key_iterator(d):
>             d[key] = self.mapping(d[key])
>         return d
> ```
> 
> This would needed added tests as well.

I would like to ask if data: Mapping[Hashable, NdarrayOrTensor]) -> dict[Hashable, NdarrayOrTensor] must include Ndarray. Since both the input and output paths require metadata, the input parameter for WriteFileMapping is MetaTensor. In this case, is it possible to use data: Mapping[Hashable, torch.Tensor]) -> dict[Hashable, torch.Tensor] or data: Mapping[Hashable, MetaTensor]) -> dict[Hashable, MetaTensor]?

Also, I've added the test for the dictionary. Please let me know if it's okay. Thank you!

tests/test_mapping_filed.py
```diff
@@ -0,0 +1,118 @@
...
+            ]
+        )
+
+        dataset = Dataset(data=data, transform=test_transforms)
+        dataloader = DataLoader(dataset, batch_size=1)
+        device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        model = UNet(spatial_dims=3, in_channels=1, out_channels=2, channels=(16, 32), strides=(2,)).to(device)
+        model.eval()
+
+        try:
```
KumoLiu, line 89: Instead of using try-except in test case, I would suggest let's clearly distinguish between cases that may cause errors and not.

KumoLiu (approved):
Thanks for the quick update, overall looks good to me.
@ericspod @binliunls, do you have any further comments on this one?

monai/transforms/io/array.py
```diff
@@ -514,3 +523,49 @@ def __call__(
...
+        if MetaKeys.SAVED_TO not in meta_data:
+            raise KeyError(
+                "Missing 'saved_to' key in metadata. Check SaveImage argument 'savepath_in_metadict' is True."
+            )
+
+        input_path = meta_data[Key.FILENAME_OR_OBJ]
+        output_path = meta_data[MetaKeys.SAVED_TO]
+        log_data = {"input": input_path, "output": output_path}
+
+        lock = FileLock(str(self.mapping_file_path) + ".lock")
```
binliunls, line 557: Should we check the `FileLock` existence before use it? cc @KumoLiu 
Looks good to me except this comment.

Thanks,  
Bin

binliunls (approved):

KumoLiu (approved):

PR #7775, opened 2024-05-16
Addition of SPADE Network  + tests and modification of SPADE normalisation 

virginiafdez:
- spade_network, and SPADENet (VAE-GAN)
- test_spade_vaegan (to test previously mentioned model) Modification of:
- spade_diffusion_model_unet.py to change namings.
- SPADE normalisation layer, to use get_norm_layer function instead of defining such layers directly.

Fixes # .

### Description

### Types of changes

- [] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.

### DIFF
--- a/monai/networks/nets/spade_network.py
+++ b/monai/networks/nets/spade_network.py
@@ -21,8 +21,8 @@ import torch.nn.functional as F
 from monai.networks.blocks import Convolution
 from monai.networks.blocks.spade_norm import SPADE
 from monai.networks.layers import Act
-from monai.utils.enums import StrEnum
 from monai.networks.layers.utils import get_act_layer
+from monai.utils.enums import StrEnum
 
 
 class UpsamplingModes(StrEnum):
@@ -280,7 +280,7 @@ class SPADEDecoder(nn.Module):
                     spade_intermediate_channels=spade_intermediate_channels,
                     norm=norm,
                     kernel_size=kernel_size,
-                    act = act
+                    act=act,
                 )
             )
 
--- a/monai/networks/nets/spade_network.py
+++ b/monai/networks/nets/spade_network.py
@@ -321,7 +321,7 @@ class SPADEDecoder(nn.Module):
         else:
             if (
                 z is None and self.z_dim is not None
-            ):  # is_vae is Truee, but we can use the VAE-GAN as GAN in this function.
+            ):  # Even though this network is a VAE (self.is_vae), you should be able to sample from noise as well.
                 z = torch.randn(seg.size(0), self.z_dim, dtype=torch.float32, device=seg.get_device())
             x = self.fc(z)
             x = x.view(*[-1, self.num_channels[0]] + self.latent_spatial_shape)
--- a/monai/networks/nets/spade_network.py
+++ b/monai/networks/nets/spade_network.py
@@ -237,8 +237,8 @@ class SPADEDecoder(nn.Module):
         out_channels: int,
         label_nc: int,
         input_shape: Sequence[int],
-        channels: Sequence[int],
-        z_dim: int | None = None,
+        channels: list[int],
+        z_dim: int  | None = None,
         is_gan: bool = False,
         spade_intermediate_channels: int = 128,
         norm: str | tuple = "INSTANCE",
@@ -265,6 +265,7 @@ class SPADEDecoder(nn.Module):
         if self.is_gan:
             self.fc = nn.Linear(label_nc, np.prod(self.latent_spatial_shape) * channels[0])
         else:
+            assert z_dim is not None
             self.fc = nn.Linear(z_dim, np.prod(self.latent_spatial_shape) * channels[0])
 
         blocks = []
@@ -295,7 +296,7 @@ class SPADEDecoder(nn.Module):
             act=last_act,
         )
 
-    def forward(self, seg, z: torch.Tensor = None):
+    def forward(self, seg, z: torch.Tensor | None = None):
         if self.is_gan:
             x = F.interpolate(seg, size=tuple(self.latent_spatial_shape))
             x = self.fc(x)
@@ -343,7 +344,7 @@ class SPADENet(nn.Module):
         out_channels: int,
         label_nc: int,
         input_shape: Sequence[int],
-        channels: Sequence[int],
+        channels: list[int],
         z_dim: int | None = None,
         is_vae: bool = True,
         spade_intermediate_channels: int = 128,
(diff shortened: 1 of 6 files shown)


tests/test_spade_vaegan.py
```diff
@@ -0,0 +1,135 @@
...
+            out, z_mu, z_logvar = net(in_label, in_image)
+            self.assertEqual(
+                False,
+                True in torch.isnan(out)
+                or True in torch.isinf(out)
+                or True in torch.isinf(z_mu)
+                or True in torch.isnan(z_mu)
+                or True in torch.isinf(z_logvar)
+                or True in torch.isnan(z_logvar),
+            )
```
ericspod, line 123:
```suggestion
            self.assertTrue(torch.all(torch.isfinite(out)))
            self.assertTrue(torch.all(torch.isfinite(z_mu)))
            self.assertTrue(torch.all(torch.isfinite(z_logvar)))
```

ericspod: Same thing for the similar statement in `test_forward_2d`.

ericspod:
The quick tests are failing for some weird issue with Matplotlib that may go away if they're rerun anyway, it isn't related to this PR. The DCO can be fixed as well by following the remediation instructions.

marksgraham:
This is the final piece of #7227 

The need for the addition of the spade network here is my fault - I missed it out in the initial port, and @virginiafdez spotted it.

The matplotlib errors have been popping up on other PRs too, recently. 

Will do a review in a moment. I can help with the mypy errors too

marksgraham (approved):
looks good to me

monai/networks/nets/spade_network.py
```diff
@@ -0,0 +1,412 @@
...
+            spatial_dims=spatial_dims,
+            in_channels=channels[-1],
+            out_channels=out_channels,
+            padding=(kernel_size - 1) // 2,
+            kernel_size=kernel_size,
+            norm=None,
+            act=last_act,
+        )
+
+    def forward(self, seg, z: torch.Tensor | None = None):
```
KumoLiu, line 299: Can we add docstring here mention what z represents here? Looks like it's output from the encoder, and will random create one if not specified with `self.is_gan=False`. Correct me if I'm wrong. Thanks.

monai/networks/nets/spade_network.py
```diff
@@ -0,0 +1,412 @@
...
+        if self.is_gan:
+            x = F.interpolate(seg, size=tuple(self.latent_spatial_shape))
+            x = self.fc(x)
+        else:
+            if z is None:
+                z = torch.randn(seg.size(0), self.opt.z_dim, dtype=torch.float32, device=seg.get_device())
```
KumoLiu, line 305: I didn't find where this `self.opt` define.

KumoLiu (commented):
Minor comments inline.

dzenanz:
Perhaps rename this PR to `SPADENet` or `SPADENet (VAE-GAN)`? The current name, `Addition of:`, is quite non-descriptive.

monai/networks/nets/spade_network.py
virginiafdez: This is a bug! Thanks for spotting. Should be self.z_dim, which has to be set-up in the init method.

tests/test_spade_vaegan.py
```diff
@@ -0,0 +1,123 @@
...
+
+    # One hot encode label
+    out_label_ = torch.zeros([semantic_regions] + list(out_label.shape))
+    for ch in range(semantic_regions):
+        out_label_[ch, ...] = out_label == ch
+
+    return out_label_.unsqueeze(0), out_image.unsqueeze(0).unsqueeze(0)
+
+
+class TestDiffusionModelUNet2D(unittest.TestCase):
```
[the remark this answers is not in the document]
virginiafdez: This shouldn't be named like this indeed. Name changed to TestSpadeNet.

tests/test_spade_vaegan.py
```diff
@@ -0,0 +1,123 @@
...
+
+import numpy as np
+import torch
+from parameterized import parameterized
+
+from monai.networks import eval_mode
+from monai.networks.nets import SPADENet
+
+CASE_2D = [[[2, 1, 1, 3, [64, 64], [16, 32, 64, 128], 16, True]]]
+CASE_2D_BIS = [[[2, 1, 1, 3, [64, 64], [16, 32, 64, 128], 16, True]]]
```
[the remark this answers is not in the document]
virginiafdez: None, CASE_2D could be used instead for the test in question. Removed it.

tests/test_spade_vaegan.py
```diff
@@ -0,0 +1,123 @@
...
+CASE_3D = [[[3, 1, 1, 3, [64, 64, 64], [16, 32, 64, 128], 16, True]]]
```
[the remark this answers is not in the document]
virginiafdez: Multiple configs can be place within each case, but in this scenario, the cases had only one config. Now there are two per case.

monai/networks/nets/spade_network.py
```diff
@@ -297,12 +306,23 @@ def __init__(
...
+        if not self.is_vae:
             x = F.interpolate(seg, size=tuple(self.latent_spatial_shape))
-            x = self.fc(x)
+            x = self.conv_init(x)
         else:
-            if z is None:
-                z = torch.randn(seg.size(0), self.opt.z_dim, dtype=torch.float32, device=seg.get_device())
+            if (
+                z is None and self.z_dim is not None
+            ):  # is_vae is Truee, but we can use the VAE-GAN as GAN in this function.
```
KumoLiu, line 324:
```suggestion
            ):  # is_vae is True, but we can use the VAE-GAN as GAN in this function.
```

I didn't get the point of this line of comment.

tests/test_spade_vaegan.py
```diff
@@ -69,7 +74,7 @@ def create_semantic_data(shape: list, semantic_regions: int):
     return out_label_.unsqueeze(0), out_image.unsqueeze(0).unsqueeze(0)
 
 
-class TestDiffusionModelUNet2D(unittest.TestCase):
+class TestSpadeNet(unittest.TestCase):
```
KumoLiu, line 77: Do we need to add a `load_old_state_dict` function for this network?
cc @ericspod @marksgraham

ericspod: This being a test network we can regenerate weights for it without worrying about user compatibility.

ericspod: We decided on not doing a old state dict loader for this network considering its use.

monai/networks/nets/spade_network.py
virginiafdez: When flag is_vae is True, the network is a VAE-GAN, but it should be possible to sample from random noise as well, even when the network is a VAE-GAN (you either encode an input and then decode the latent resulting of that, or sample from noise). I've clarified this.

PR #7779, opened 2024-05-16
Fix doc source links for read the docs

johnzielke:
Fixes the current docs build.
Currently no [source] links are shown anymore.
It seems like read the docs uses a different working directory, therefore breaking the source links. It also will not have the correct git tag referenced. This should fix both of these issues. 

If there is a way to test if this fix works without merging to dev, let me know.

### Description

### Types of changes

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -168,7 +168,11 @@ def setup(app):
 
 # -- Linkcode configuration --------------------------------------------------
 DEFAULT_REF = "dev"
-if os.environ.get("GITHUB_REF_TYPE", "branch") == "tag":
+if git_ref := os.environ.get("READTHEDOCS_GIT_IDENTIFIER", None):
+    # When building on ReadTheDocs, link to the specific commit
+    # https://docs.readthedocs.io/en/stable/reference/environment-variables.html#envvar-READTHEDOCS_GIT_IDENTIFIER
+    git_ref = git_ref
+elif os.environ.get("GITHUB_REF_TYPE", "branch") == "tag":
     # When building a tag, link to the tag itself
     git_ref = os.environ.get("GITHUB_REF", DEFAULT_REF)
 else:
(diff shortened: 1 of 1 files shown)


KumoLiu:
> If there is a way to test if this fix works without merging to dev, let me know.

A easy way may be cat one of the html in the job workflow to see whether the source link generated correctly.
https://github.com/Project-MONAI/MONAI/blob/dev/.github/workflows/pythonapp.yml#L238

johnzielke:
The problem is that the github workflow and the read the docs workflow are not identical.

johnzielke:
Maybe enabling this might help recognize these issues in the future: https://docs.readthedocs.io/en/stable/pull-requests.html Although this still requires you to manually check the created build, if you make changes you can confirm that they also succeed in the read the docs build

docs/source/conf.py
```diff
@@ -168,7 +168,11 @@ def setup(app):
 
 # -- Linkcode configuration --------------------------------------------------
 DEFAULT_REF = "dev"
-if os.environ.get("GITHUB_REF_TYPE", "branch") == "tag":
+if git_ref := os.environ.get("READTHEDOCS_GIT_IDENTIFIER", None):
```
ericspod, line 171: We don't use the walrus operator anywhere else in MONAI so I'd suggest not using it here to be consistent with our code standard. Thanks!

johnzielke: Ok, no problem. Then I can get rid of the weird noop assignment in the next line as well

KumoLiu:
Hi @johnzielke, I'm considering reverting this [PR](https://github.com/Project-MONAI/MONAI/pull/7738) for the 1.3.1 release and instead merging it into 1.4. I want to avoid any potential disruptions in the upcoming release. Thanks for understanding.

KumoLiu (approved):

johnzielke:
Hi @KumoLiu, it looks like it's working now. If you still end up doing that, I totally understand. I did setup my own read the docs to test, and it seems like the reference to stable versions/tags works as well and will then point to the commit of that version.

PR #7784, opened 2024-05-20
Refactor Dataset to use Compose for transforms

### LINKED ISSUE #7646: Provide ability to override `map_items=True` set by default for `Dataset`
https://github.com/Project-MONAI/MONAI/blob/bff4b1559a147f570fc881717eadc7b3e8a98b96/monai/data/dataset.py#L98

In the `Dataset` class, for any `transform` applied, `map_items=True` is the default behavior in the `apply_transform` helper function.

 I'm trying to use some other datasets in MONAI which have an (input, target) tuple as the return from `_getitem_`. When I superclass this with the MONAI dataset, the tuple is unpacked, and the input and target are operated on as different items in the dataset. 

Ability to provide `kwargs` to `Dataset` that propagates to the `apply_transform` function, allowing disabling this mapping behavior. 

Not using the Dataset class. 

**Additional context**
A lot of datasets have the (input, target) behavior, and it would be useful to be able to wrap these in MONAI dataset classes.

surajpaib:
Fixes #7646 

### Description

### Types of changes

### DIFF
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -73,7 +73,9 @@ class Dataset(_TorchDataset):
         """
         Args:
             data: input data to load and transform to generate dataset for model.
-            transform: a callable data transform on input data.
+            transform: a callable, sequence of callables or None. The transform will be passed
+            into a `Compose` wrapper by default. Sequence of callables are applied in order and
+            if `None` is passed, the data is returned as is.
         """
         self.data = data
         try:
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -76,13 +76,10 @@ class Dataset(_TorchDataset):
             transform: a callable data transform on input data.
         """
         self.data = data
-        if transform is None:
-            self.transform = None
-        else:
-            try:
-                self.transform = Compose(transform) if not isinstance(transform, Compose) else transform
-            except Exception as e:
-                raise ValueError("`transform` must be a callable or a list of callables that is Composable") from e
+        try:
+            self.transform = Compose(transform) if not isinstance(transform, Compose) else transform
+        except Exception as e:
+            raise ValueError("`transform` must be a callable or a list of callables that is Composable") from e
 
     def __len__(self) -> int:
         return len(self.data)
@@ -92,7 +89,7 @@ class Dataset(_TorchDataset):
         Fetch single data item from `self.data`.
         """
         data_i = self.data[index]
-        return self.transform(data_i) if self.transform is not None else data_i
+        return self.transform(data_i)
 
     def __getitem__(self, index: int | slice | Sequence[int]):
         """
@@ -262,8 +259,6 @@ class PersistentDataset(Dataset):
 
         """
         super().__init__(data=data, transform=transform)
-        if self.transform is None:
-            raise ValueError("transform must not be None when provided to a PersistentDataset")
         self.cache_dir = Path(cache_dir) if cache_dir is not None else None
         self.hash_func = hash_func
         self.pickle_module = pickle_module
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -77,15 +77,17 @@ class Dataset(_TorchDataset):
          },                           },                           }]
     """
 
-    def __init__(self, data: Sequence, transform: Callable | None = None) -> None:
+    def __init__(self, data: Sequence, transform: Callable | None = None, **kwargs) -> None:
         """
         Args:
             data: input data to load and transform to generate dataset for model.
             transform: a callable data transform on input data.
+            kwargs: other arguments for the `apply_transform` function called to apply provided transforms. For ex. `map_items=False`
 
         """
         self.data = data
         self.transform: Any = transform
+        self.apply_transform_kwargs = kwargs
 
     def __len__(self) -> int:
         return len(self.data)
@@ -95,7 +97,11 @@ class Dataset(_TorchDataset):
         Fetch single data item from `self.data`.
         """
         data_i = self.data[index]
-        return apply_transform(self.transform, data_i) if self.transform is not None else data_i
+        return (
+            apply_transform(self.transform, data_i, **self.apply_transform_kwargs)
+            if self.transform is not None
+            else data_i
+        )
 
     def __getitem__(self, index: int | slice | Sequence[int]):
         """
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -36,15 +36,7 @@ from torch.utils.data import Subset
 
 from monai.data.meta_tensor import MetaTensor
 from monai.data.utils import SUPPORTED_PICKLE_MOD, convert_tables_to_dicts, pickle_hashing
-from monai.transforms import (
-    Compose,
-    Randomizable,
-    RandomizableTrait,
-    Transform,
-    apply_transform,
-    convert_to_contiguous,
-    reset_ops_id,
-)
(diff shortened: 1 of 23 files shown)


ericspod:
Hi @surajpaib thanks for the contribution. I think that the other classes inheriting from `Dataset` should be modified as well to pass along these arguments. They should have kwargs added to their constructors which is then passed to the superconstructor calls. There are other dataset classes which call `apply_transform` which maybe should be modified as well. The tests I think need to be a bit more thorough, currently they check the type of produced data which isn't indicative of the differing behaviour, you should be inspecting the return value somehow to demonstrate the added arguments have made a difference.

monai/data/dataset.py
```diff
@@ -77,15 +69,17 @@ class Dataset(_TorchDataset):
...
         """
         Args:
             data: input data to load and transform to generate dataset for model.
             transform: a callable data transform on input data.
-
         """
         self.data = data
-        self.transform: Any = transform
+        try:
+            self.transform = Compose(transform) if not isinstance(transform, Compose) else transform
```
KumoLiu, line 80: Looks we miss support for transform to be None here.

KumoLiu:
ci error log:
```
ERROR: test_shape_2 (tests.test_arraydataset.TestArrayDataset)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/parameterized/parameterized.py", line 620, in standalone_func
  File "/Users/runner/work/MONAI/MONAI/tests/test_arraydataset.py", line 100, in test_shape
    data1 = dataset[0]
  File "/Users/runner/work/MONAI/MONAI/monai/data/dataset.py", line 1367, in __getitem__
    return self.dataset[index]
  File "/Users/runner/work/MONAI/MONAI/monai/data/dataset.py", line 106, in __getitem__
    return self._transform(index)
  File "/Users/runner/work/MONAI/MONAI/monai/data/dataset.py", line 1262, in _transform
    data.extend(to_list(dataset[index]))
  File "/Users/runner/work/MONAI/MONAI/monai/data/dataset.py", line 106, in __getitem__
    return self._transform(index)
  File "/Users/runner/work/MONAI/MONAI/monai/data/dataset.py", line 92, in _transform
    return self.transform(data_i)
TypeError: __call__() missing 1 required positional argument: 'lazy'

```

monai/data/dataset.py
surajpaib: I've intentionally left this out as the `Compose` should handle the case where `transform` happens to be `None`. 

See:
https://github.com/Project-MONAI/MONAI/blob/e5afa4341b18149ca8d44f680f669b33ab423722/monai/transforms/compose.py#L238

https://github.com/Project-MONAI/MONAI/blob/e5afa4341b18149ca8d44f680f669b33ab423722/monai/transforms/compose.py#L105

surajpaib:
Hi @ericspod @KumoLiu 

The updated PR is ready for review.

Noting some of the changes from the PR below,
1. All dataset `transform`s use `Compose` now. I've replaced the `apply_transform` way of applying the transforms with `Compose` for the `Dataset`, `ZipDataset` and `NPZDictItemDataset` classes.
2. Removed the `Compose` checks from `PersistentDataset` and `CacheDataset` as these are now done in the `Dataset` base class.
3. Added tests for the `map_items=False` and `tuple` input scenario for `Dataset` in `test_dataset.py`. Modified tests for `test_profiling.py` and `test_arraydataset.py` for the `Compose` changes made. The functionality tested does not change. 

Overall, I think these changes make the `transform` application across `Datasets` more uniform and simplify underlying code as well.

monai/data/dataset.py
KumoLiu: Still suggest add a check in the dataset, it's not easy to understand even it's None, it will still return a compose object. What do you think?

monai/data/dataset.py
```diff
@@ -323,9 +315,6 @@ def _pre_transform(self, item_transformed):
             random transform object
 
         """
-        if not isinstance(self.transform, Compose):
```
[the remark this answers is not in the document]
surajpaib: Since both these datasets inherit from `Dataset` and the check is now moved to `Dataset`, this point in the code would always receive a `Compose` object, making this check redundant.

monai/data/dataset.py
surajpaib: Is there any reason we need to have an explicit `None` check here?

Since `Compose` will handle the case when `transform = None` internally by assigning `self.transforms = []` which will give the exact same output as before (returning the data as is), I do not see the need to add an explicit `None` check. 

Moreover, the `Compose` logic is now consistent across dataset implementations making the behaviour much more understandable in my opinion.

KumoLiu: Yes, I know the point, just feel it's not straightforward when users check ds.transform, it will still return a compose object.
```
import torch
from monai.data import Dataset

data = [torch.rand(10, 10) for _ in range(100)]
ds = Dataset(data=data, transform=None)
print(ds.transform) # --> <monai.transforms.compose.Compose object at 0x7ffa6088b250>
print(ds.transform.transforms) # --> ()
```
But it's ok if other developers also agree on this.
cc @ericspod @atbenmurray @Nic-Ma

monai/data/dataset.py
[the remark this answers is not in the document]
KumoLiu: When considering it along with the previous question, `PersistentDataset` must ensure that the transform is always not None. However, it will always convert to a compose in the base class, even with an empty transform, results in behavior that is inconsistent with the previous implementation.

monai/data/dataset.py
ericspod: I think this behaviour without a check is fine so long as the docstring line documenting `transform` states this, currently it makes no mention of what `None` would imply or what a sequence of callables would imply.

monai/data/dataset.py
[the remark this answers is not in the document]
ericspod: Correct me if I'm wrong but I wonder if these checks were redundant in the original class as well? The lines in the constructor that this PR removes were converting `transform` to a `Compose` object if it wasn't already, so that is the behaviour that now is inherited. In these methods we're checking that `self.transform` is a `Compose` object, but wouldn't that always be the case in the existing code and in this PR? Am I missing some situation where this check would prevent some condition?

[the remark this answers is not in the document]
KumoLiu: No, you are right. The previous implementation is also redundant. Mark as resolved.

monai/data/dataset.py
surajpaib: @ericspod @KumoLiu Docstring for this is updated in the latest commit.

monai/data/dataset.py
```diff
@@ -77,15 +69,19 @@ class Dataset(_TorchDataset):
...
-    def __init__(self, data: Sequence, transform: Callable | None = None) -> None:
+    def __init__(self, data: Sequence, transform: Sequence[Callable] | Callable | None = None) -> None:
         """
         Args:
             data: input data to load and transform to generate dataset for model.
-            transform: a callable data transform on input data.
-
+            transform: a callable, sequence of callables or None. The transform will be passed
+            into a `Compose` wrapper by default. Sequence of callables are applied in order and
+            if `None` is passed, the data is returned as is.
```
atbenmurray, line 78:
```suggestion
            transform: a callable, sequence of callables or None. If transform is not
            a `Compose` instance, it will be wrapped in a `Compose` instance. Sequences
            of callables are applied in order and if `None` is passed, the data is returned as is.
```

ericspod (approved):
I'm good with this now if everyone else is.

KumoLiu (approved):

PR #7811, opened 2024-05-27
Modify _log_input_summary function

### LINKED ISSUE #7513: Enhance bundle logging logic
There are issues with using logging.conf to document the arguments recorded by _log_input_summary. Additionally, when more than one GPU is utilized, the output becomes messy and unmanageable.

https://github.com/Project-MONAI/model-zoo/blob/dev/models/brats_mri_generative_diffusion/configs/logging.conf
https://github.com/Project-MONAI/MONAI/blob/02c7f53a2ea5835623df6ad12943d58202a42e04/monai/bundle/scripts.py#L118

hyacinth97223:
Fixes #7513

Fixes # .

### Description

### Types of changes

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -10,6 +10,7 @@
 # limitations under the License.
 
 from __future__ import annotations
+from io import StringIO
 
 import ast
 import json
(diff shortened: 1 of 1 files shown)


monai/bundle/scripts.py
```diff
@@ -116,10 +117,14 @@ def _pop_args(src: dict, *args: Any, **kwargs: Any) -> tuple:
...
+    log_buffer = StringIO()
+
+    log_buffer.write(f"--- input summary of monai.bundle.scripts.{tag} ---\n")
     for name, val in args.items():
-        logger.info(f"> {name}: {pprint_edges(val, PPRINT_CONFIG_N)}")
-    logger.info("---\n\n")
+        log_buffer.write(f"> {name}: {pprint_edges(val, PPRINT_CONFIG_N)}\n")
+    log_buffer.write("---\n\n")
+
+    logger.info(log_buffer.getvalue())
```
KumoLiu, line 127: This change may still not solve the issue mentioned in https://github.com/Project-MONAI/MONAI/issues/7513?

ericspod: What it seems to me will happen here is the message will be written all at once with one call to `logger.info()` which has a greater chance of appearing all at once in the output in the presence of multiple GPUs. This is only by chance and isn't guaranteed. What we need in the case of multiple GPUs is to identify which process is rank 0 and only it writes to the logger by default but with an option to allow everyone to write if desired. With the `torch.distributed` package this would be:

```python
import torch.distributed as dist
...
rank = dist.get_rank() if dist.is_initialized() else 0
```

hyacinth97223: @ericspod  Thanks for your advise. I will try it.

PR #7813, opened 2024-05-29
Ensure deterministic in MixUp, CutMix, CutOut

### LINKED ISSUE #7697: AssertionError in test_regularization

KumoLiu:
Fixes #7697

### Description

### Types of changes

### DIFF
--- a/monai/transforms/regularization/array.py
+++ b/monai/transforms/regularization/array.py
@@ -60,7 +60,7 @@ class Mixer(RandomizableTransform):
         self._params = (
             torch.from_numpy(self.R.beta(self.alpha, self.alpha, self.batch_size)).type(torch.float32),
             self.R.permutation(self.batch_size),
-            [torch.from_numpy(self.R.randint(0, d, size=(1,))) for d in data.shape[2:]] if data is not None else None,
+            [torch.from_numpy(self.R.randint(0, d, size=(1,))) for d in data.shape[2:]] if data is not None else [],
         )
 
 
--- a/monai/transforms/regularization/array.py
+++ b/monai/transforms/regularization/array.py
@@ -16,7 +16,9 @@ from math import ceil, sqrt
 
 import torch
 
+from monai.data.meta_obj import get_track_meta
 from ..transform import RandomizableTransform
+from monai.utils.type_conversion import convert_to_tensor, convert_to_dst_type
 
 __all__ = ["MixUp", "CutMix", "CutOut", "Mixer"]
 
@@ -53,9 +55,11 @@ class Mixer(RandomizableTransform):
         as needed. You need to call this method everytime you apply the transform to a new
         batch.
         """
+        super().randomize(None)
         self._params = (
             torch.from_numpy(self.R.beta(self.alpha, self.alpha, self.batch_size)).type(torch.float32),
             self.R.permutation(self.batch_size),
+            [torch.from_numpy(self.R.randint(0, d, size=(1,))) for d in data.shape[2:]] if data is not None else None,
         )
 
 
@@ -69,7 +73,7 @@ class MixUp(Mixer):
     """
 
     def apply(self, data: torch.Tensor):
-        weight, perm = self._params
+        weight, perm, _ = self._params
         nsamples, *dims = data.shape
         if len(weight) != nsamples:
             raise ValueError(f"Expected batch of size: {len(weight)}, but got {nsamples}")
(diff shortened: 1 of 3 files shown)


monai/transforms/regularization/array.py
```diff
@@ -80,11 +85,20 @@ def apply(self, data: torch.Tensor):
...
         return mixweight * data + (1 - mixweight) * data[perm, ...]
 
-    def __call__(self, data: torch.Tensor, labels: torch.Tensor | None = None):
-        self.randomize()
+    def __call__(self, data: torch.Tensor, labels: torch.Tensor | None = None, randomize=True):
+        data = convert_to_tensor(data, track_meta=get_track_meta())
+        data_t = convert_to_tensor(data, track_meta=False)
+        if labels is not None:
+            labels_t = convert_to_tensor(labels, track_meta=get_track_meta())
+            labels_t = convert_to_tensor(labels, track_meta=False)
```
ericspod, line 93:
```suggestion
        data_t = convert_to_tensor(data, track_meta=get_track_meta())
        if labels is not None:
            labels_t = convert_to_tensor(labels, track_meta=get_track_meta())
```
Unless I missed something these lines were redundant?

monai/transforms/regularization/array.py
```diff
@@ -113,33 +127,40 @@ class CutMix(Mixer):
...
-    def __call__(self, data: torch.Tensor, labels: torch.Tensor | None = None):
-        self.randomize()
-        augmented = self.apply(data)
-        return (augmented, self.apply_on_labels(labels)) if labels is not None else augmented
+    def __call__(self, data: torch.Tensor, labels: torch.Tensor | None = None, randomize=True):
+        data = convert_to_tensor(data, track_meta=get_track_meta())
+        data_t = convert_to_tensor(data, track_meta=False)
+        if labels is not None:
+            labels_t = convert_to_tensor(labels, track_meta=get_track_meta())
+            labels_t = convert_to_tensor(labels, track_meta=False)
```
ericspod, line 157:
```suggestion
        data_t = convert_to_tensor(data, track_meta=get_track_meta())
        if labels is not None:
            labels_t = convert_to_tensor(labels, track_meta=get_track_meta())
```

monai/transforms/regularization/array.py
KumoLiu: It was probably written this way in the beginning to avoid having a lot of meta data constantly recorded in the computation. It's fine to move here.

monai/transforms/regularization/array.py
```diff
@@ -155,20 +176,22 @@ class CutOut(Mixer):
...
             mask[s][idx] = 0
 
         return mask * data
 
-    def __call__(self, data: torch.Tensor):
-        self.randomize()
-        return self.apply(data)
+    def __call__(self, data: torch.Tensor, randomize=True):
+        data = convert_to_tensor(data, track_meta=get_track_meta())
+        data_t = convert_to_tensor(data, track_meta=False)
```
ericspod, line 194:
```suggestion
        data_t = convert_to_tensor(data, track_meta=get_track_meta())
```

monai/transforms/regularization/dictionary.py
```diff
@@ -31,18 +38,24 @@ class MixUpd(MapTransform):
...
+
     def __call__(self, data):
-        self.mixup.randomize()
-        result = dict(data)
-        for k in self.keys:
-            result[k] = self.mixup.apply(data[k])
-        return result
+        d = dict(data)
+        # all the keys share the same random state
+        self.mixup.randomize(None)
```
ericspod, line 52: Should we add a `randomize` keyword argument for this and the other `__call__` methods so we can only optionally call `randomize`?

tests/test_regularization.py
```diff
@@ -85,30 +92,31 @@ def test_cutmixd(self):
             label = torch.randint(0, 1, shape)
             sample = {"a": t, "b": t, "lbl1": label, "lbl2": label}
             cutmix = CutMixd(["a", "b"], 6, label_keys=("lbl1", "lbl2"))
+            cutmix.set_random_state(seed=123)
             output = cutmix(sample)
-            # croppings are different on each application
-            self.assertTrue(not torch.allclose(output["a"], output["b"]))
```
KumoLiu, line 90: Behaviour changed here since for dictionary version, it will apply all the same parameters on different keys. It will align with other transforms.

monai/transforms/regularization/dictionary.py
KumoLiu: Do you mean always make these transforms generate the same results? If we want to do that, the better way may be introduce RandMixUp, RandCutMix, RandCutOut and make existing transforms not inherit from the `RandomizableTransform`. But it seems that these transforms were originally designed for augmentation, so maybe deterministic is not commonly used?

ericspod: OK, let's not do that then and see if anyone requests this change later.

ericspod (approved):
Looks good now, the test changes look good too and capture the actual semantics we want. Thanks!

tests/test_regularization.py
ericspod: Yes we can deterministically inspect the results now.

monai/transforms/regularization/array.py
```diff
@@ -113,33 +123,40 @@ class CutMix(Mixer):
...
+        if labels is not None:
+            labels_t = convert_to_tensor(labels, track_meta=get_track_meta())
+            labels_t = convert_to_tensor(labels, track_meta=False)
+        if randomize:
+            self.randomize(data)
+        augmented = convert_to_dst_type(self.apply(data_t), dst=data)[0]
+        if labels is not None:
+            augmented_label = convert_to_dst_type(self.apply(labels_t), dst=labels)[0]
+        return (augmented, augmented_label) if labels is not None else augmented
```
maxime915: Hello, I'm a researcher currently working with CutMix and other similar method. My group and I noticed an inconsistent behavior in CutMix on the segmentation labels on our different machines. After some digging, we think it may come from this PR: before this change, `self.apply_on_labels(labels)` is used for the label, but the modified change use `self.apply(labels_t)` which uses the same cropping mechanism as for the image.

This PR does not mention the reason for this change, could you enlighten me please ?

Thank you in advance for your time.

PR #7814, opened 2024-05-30
Make MetaTensor optional printed in DataStats and DataStatsd #5905

### LINKED ISSUE #5905: verbose DataStats for metatensor
the DataStats transform can print basic info about input tensors:
https://github.com/Project-MONAI/MONAI/blob/59593bb21652e084f4e5d83eb6b58a85c5f99d44/monai/transforms/utility/array.py#L681-L684

but it can't call the `print_verbose`, which makes it impossible to inspect metadata. would be great to make the metadata optionally printed.

https://github.com/Project-MONAI/MONAI/blob/59593bb21652e084f4e5d83eb6b58a85c5f99d44/monai/data/meta_tensor.py#L559-L561

slicepaste:
Fixes #5905 

### Description

We simply add one argument for DataStats and DataStatsd to make MetaTensor optional printed.

### Types of changes


monai/transforms/utility/array.py
```diff
@@ -654,6 +654,7 @@ def __init__(
         data_shape: bool = True,
         value_range: bool = True,
         data_value: bool = False,
+        meta_tensor: bool = False,
```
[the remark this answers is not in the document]
ericspod: This sounds good both otherwise this looks good to me overall.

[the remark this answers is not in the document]
slicepaste: Thanks for the suggestion!  I think `meta_info` would be better in this case.

KumoLiu:
@slicepaste, after change the argument, please also try run `./runtests.sh --autofix` to fix the format issue.
https://github.com/Project-MONAI/MONAI/actions/runs/9312809059/job/25634185420?pr=7814#step:7:375

Thanks!

slicepaste:
Thanks for the reminder. It still some issue, I'll figure out the problem as fast as possible.

slicepaste:
@KumoLiu 
I have fixed the argument and formatting problem, I would appreciate it if you could review this issue.
If there are any additional changes needed, please let me know.
Thank you very much!

monai/transforms/utility/array.py
```diff
@@ -725,6 +729,8 @@ def __call__(
                 lines.append(f"Value range: (not a PyTorch or Numpy array, type: {type(img)})")
         if self.data_value if data_value is None else data_value:
             lines.append(f"Value: {img}")
+        if self.meta_info if meta_info is None else meta_info:
+            lines.append(f"Meta_Info: {MetaTensor(img).meta.__repr__()}")
```
ericspod, line 733:
```suggestion
        if self.meta_info if meta_info is None else meta_info:
            lines.append(f"Meta_Info: {repr(MetaTensor(img).meta)}")
```
Unless you need to you should use the function rather than the special function. Perhaps we should check if `img` is a `MetaTensor` instance here and do nothing if it's not? Always converting to `MetaTensor` like this maybe isn't ideal. If you agree we would have:

```suggestion
        if isinstance(img, MetaTensor) and (self.meta_info if meta_info is None else meta_info):
            lines.append(f"Meta_Info: {repr(img.meta)}")
```

slicepaste: Sorry for the procrastination. Thanks for the suggestion, maybe I should change it to the following code:

```
if self.meta_info if meta_info is None else meta_info:
  metadata = getattr(img, "meta", None)
  lines.append(f"Meta_Info: {repr(metadata)}")
```

Does this look better?

ericspod: I would say that we shouldn't print anything if `img` is not a MetaTensor rather than imply it is and whose `meta` value is `None`. I would also write "Meta info" instead of "Meta_Info" in the string as well.

slicepaste: You're right, but I think users have already set the argument to `meta_info=True` in `Datastat` or `Datastatd`. Perhaps we should give some feedback after checking. Would it be too verbose if we responded more explicitly, like `Meta info: (input is not a MetaTensor)`?
```
if self.meta_info if meta_info is None else meta_info:
  metadata = getattr(img, "meta", "(input is not a MetaTensor)")
  lines.append(f"Meta info: {repr(metadata)}")
```

KumoLiu: Agree, you can add something like this. What do you think? @ericspod

KumoLiu:
Hi @slicepaste, do you plan to finish this PR? If we want to include the change in version 1.4, we only have about one or two weeks left to add new things. Thanks!

slicepaste:
Hi @KumoLiu , 
Sure, I'll finish this PR in the next few days. Thanks for the reminder!

KumoLiu (approved):
LGTM now, thanks for the quick update.

Issue #7682, opened 2024-04-19
Missing `alpha` parameter of Focal loss in DiceFocalLoss


dkalsan:
**Describe the bug**
The available class [monai.losses.DiceFocalLoss](https://docs.monai.io/en/stable/_modules/monai/losses/dice.html#DiceFocalLoss) is missing the `alpha` parameter for Focal loss. Is this a result of some design choice or just got overlooked in the implementation?

**Expected behavior**
`alpha` parameter should be passed to the monai.losses.FocalLoss or have the reason for its absence documented.

KumoLiu:
Hi @dkalsan, what does `alpha` mean here, can you be more explicit? Thanks.

dkalsan:
I mean the argument passed down to the FocalLoss class during construction, which results in an alpha-balanced Focal loss as per the paper. I believe this argument should also be available in the DiceFocalLoss since it's effectively just a wrapper around the two losses.

From the [docstring of FocalLoss](https://docs.monai.io/en/stable/losses.html#monai.losses.FocalLoss):
> alpha – value of the alpha in the definition of the alpha-balanced Focal loss. The value should be in [0, 1]. Defaults to None.

PR #7841, opened 2024-06-12
Add alpha parameter to DiceFocalLoss

kephale:
Fixes https://github.com/Project-MONAI/MONAI/issues/7682.

### Description

This PR introduces the `alpha` parameter from `FocalLoss` into `DiceFocalLoss`. 

### Types of changes


monai/losses/dice.py
```diff
@@ -835,6 +835,7 @@ def __init__(
         gamma: float = 2.0,
         focal_weight: Sequence[float] | float | int | torch.Tensor | None = None,
         weight: Sequence[float] | float | int | torch.Tensor | None = None,
+        alpha: float | None = None,
```
ericspod, line 838: To avoid breaking any existing code, new parameters should appear at the end of the list. I know it makes sense for the lambda parameters to be last but this avoids problems if anyone called the constructor with all positional arguments.

tests/test_dice_focal_loss.py
```diff
@@ -91,6 +91,28 @@ def test_script(self):
         test_input = torch.ones(2, 1, 8, 8)
         test_script_save(loss, test_input, test_input)
 
+    def test_result_with_alpha(self):
```
ericspod, line 94: Can we use `@parameterized.expand` for this test instead of the loops? It reduces the size of the test and breaks them up into separate tests. If this tests only the `alpha` parameter it would be helpful to have fewer such values if they don't really make a large difference, we don't need to test `FocalLoss` itself but just verify `DiceFocalLoss` uses it correctly.

ericspod:
Hi @kephale thanks for the contribution! I've made a few inline comments but in general it looks good. Please address these and I can come back to it for approval.

kephale:
Okey doke, the requested changes have been addressed. I'm not sure what is going on with the min-dep tests because that part of the code hasn't been modified.

ericspod:
The test fails are related to Numpy 2.0 being released a few days ago and we aren't compatible yet. I have a fix I need to push then these will work.

ericspod:
I have pushed a temporary fix now so I'm rerunning your tests after merging `dev` into this branch.

ericspod:
There is a DCO fail that you do need to fix, @kephale. The instructions are there if you click on the check.

kephale:
The DCO should be addressed.

KumoLiu (approved):

KumoLiu:
Hi @ericspod, do you have any other comments on this one?

ericspod (approved):
I'm good here!

PR #7857, opened 2024-06-17
Updating to match Numpy 2.0 requirements

### LINKED ISSUE #7856: Numpy 2.0 Behaviour Changes
**Describe the bug**
Numpy 2.0 was just released with some behavioural changes which are showing up as bugs. One that is seen in this trace:

```
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/runner/work/MONAI/MONAI/tests/min_tests.py", line 239, in <module>
    result = test_runner.run(run_testsuit())
  File "/home/runner/work/MONAI/MONAI/tests/min_tests.py", line 226, in run_testsuit
    test_suite = unittest.TestLoader().loadTestsFromNames(cases)
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/unittest/loader.py", line 220, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/unittest/loader.py", line 220, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
  File "/home/runner/work/MONAI/MONAI/tests/test_inverse_collation.py", line 55, in <module>
    TESTS_3D = [
  File "/home/runner/work/MONAI/MONAI/tests/test_inverse_collation.py", line 62, in <listcomp>
    Compose([RandRotate90d(keys=KEYS, spatial_axes=(1, 2))]),
  File "/home/runner/work/MONAI/MONAI/monai/transforms/compose.py", line 251, in __init__
    self.set_random_state(seed=get_seed())
  File "/home/runner/work/MONAI/MONAI/monai/transforms/compose.py", line 263, in set_random_state
    _transform.set_random_state(seed=self.R.randint(MAX_SEED, dtype="uint32"))
  File "/home/runner/work/MONAI/MONAI/monai/transforms/transform.py", line 207, in set_random_state
    _seed = _seed % MAX_SEED
OverflowError: Python integer 4294967296 out of bounds for uint32
``` 

The underlying issue is that the `MAX_SEED` value is 2**32 which is an overflow value for uint32. It is stored as a Python `int` but when used as the right hand side of modulus in this expression where `_seed` is a uint32 of any value, the exception shown above is raised. In Numpy before 2.0, the expression would silently produce a int64 result.

**To Reproduce**
1. Import Numpy 2.0 as np
2. Attempt evaluate `np.uint32(1)%(np.iinfo(np.uint32).max+1)`
3. This will raise an exception rather than silently produce a np.int64 result as happens in Numpy<2.0

**Expected behavior**
No exception should be raise for this operations if the MAX_SEED value is set correctly. The result from the expression should be a uint32.

**Environment**

```
pip install numpy\>=2.0
python -c "import numpy as np;print(np.uint32(1)%(np.iinfo(np.uint32).max+1))"
```

**Additional context**
This was encountered when doing a PR on the `gen-ai-dev` branch. @virginiafdez

ericspod:
Fixes #7856. 

### Description

This introduces changes to meet Numpy 2.0 requirements. MONAI itself is compatible with Numpy 2.0 however some dependencies are not such as older versions of Pytorch. This PR adjusts the MAX_SEED value to be compatible with Numpy 2.0 behaviour changes, uses the `ptp` function, and some other minor tweaks. The versions for dependencies are also fixed to exclude Numpy 2.0.

### Types of changes


Nic-Ma:
Hi @ericspod ,

Thanks for your quick update for Numpy 2.0.
May I know whether you found any test failures due to 3rd party dependency can't support Numpy 2.0 so far?

Thanks in advance.

ericspod:
These appear to be compatible with Numpy2:
* Pytorch<2.3
* Pytorch 2.4.0+cpu on Windows?
* Pyarrow
* ClearML when using Pyarrow

ericspod:
@KumoLiu We'll leave the requirements excluding Numpy2 but otherwise updating the minimum version.

environment-dev.yml
```diff
@@ -5,10 +5,10 @@ channels:
   - nvidia
   - conda-forge
 dependencies:
-  - numpy>=1.20
+  - numpy>=1.25,<2.0
```
KumoLiu, line 8: Hi @ericspod, is there a specific reason to use numpy>=1.25? The base image for MONAI 1.4 will be PyTorch 24.08, which will continue to use numpy == 1.24.4.

ericspod: There wasn't except to use a more recent version than 1.20, so it can be set to 1.24.

requirements.txt
```diff
@@ -1,2 +1,2 @@
 torch>=1.9
-numpy>=1.20,<=1.26.0
+numpy>=1.25,<2.0
```
KumoLiu, line 2: Here should also be updated.

setup.cfg
```diff
@@ -42,7 +42,7 @@ setup_requires =
     ninja
 install_requires =
     torch>=1.9
-    numpy>=1.20
+    numpy>=1.25,<2.0
```
KumoLiu, line 45: Same here.

KumoLiu (approved):
Thanks for the PR, looks good to me.

michael-camilleri:
Will Numpy 2 be supported soon? Especially as the dependency is due to older versions of pytorch/other libraries and not explicitly support within MONAI itself? is there a reason why MONAI does not drop older versions of pytorch?

ericspod:
Hi @michael-camilleri The dependency issue for us is that we do want to continue supporting older versions of Pytorch because  many users may need older versions for compatibility or because their environment requires it. The problem that everyone has discovered is that their versions strings for Numpy would set a minimum requirement but not a max, so Numpy 2.0 with its incompatibility isn't excluded. Older versions of libraries are unlikely to be updated with support for Numpy 2.0, and we have no way of forcing Numpy 1.x when these are present. For the moment we've chosen to just restrict the version although MONAI itself should be compatible now with the changes made. I think it would work for you to try installing MONAI in an environment (venv or conda) and then installing Numpy 2.0 subsequently.

ericspod:
I agree, we can improve the documentation and mention this issue with Numpy 2. Beyond that, we would need a way of directly controlling version resolution in pip to account for known incompatibilities, but this isn't currently possible and would rely on guessing which libraries have a 2.0 problem.

michael-camilleri:
Yes, I see your point. I think a documentation of the issue within the
installation instructions is enough to be honest, given that it is (in most
cases) an easy fix.

Michael

PR #7859, opened 2024-06-17
Fixing Numpy requirements to exclude 2.0

ericspod:
### Description

This fixes the version of Numpy to be less than 2.0 so that we can get a working setup guaranteed. Numpy 2.0 isn't supported yet by MONAI and will need a number of changes to be compatible. 

This is only sometimes a problem because a number of dependencies we have require Numpy<2.0. When these are absent however, such as with minimal installs, Numpy 2.0 can be installed and causes problems. This is a temporary fix until we have worked out an update that's compatible with versions before and after 2.0.

### Types of changes


ericspod:
@wyli @pxLi @YanxuanLiu @KumoLiu @Nic-Ma I need one of you to initiate the Blossom tests to let this through. Other PRs can't pass because certain tests are failing if they install Numpy 2.0. We should update things to be compatible but this is the temporary fix we need for now.

Nic-Ma (approved):

MattTheCuber:
Is there plans to make MONAI support Numpy v2 in the near future?

ericspod:
Hi @MattTheCuber we are discussing developing a dependency update policy at #8303 so we can balance keeping older versions for those that need them against updating requirements appropriately. For Numpy 2 if you don't need to use any dependencies which do not work with it, you can install MONAI then upgrade the version of Numpy that gets installed. That should work as we've made MONAI compatible but added this exclusion to ensure the install process would produce a valid configuration.

PR #7876, opened 2024-06-25
fix implementation mistakes and add conjugate gradients solver

### LINKED ISSUE #6767: faster solver for computing confidence maps
follow up of #6709
             
 > The issue though is the process for installing Octave and oct2py is more complex than other things we rely on, requiring a separate install and a correctly configured PATH variable. This is also an issue for our CICD system since we do need to test this transform. The library you're using, oct2py, also seems like a rather small project whose long term support is not assured. I'd still recommend we look into some other solver (something wrapping eigen?) to avoid this dependency.

Hello, for now I've removed the octave dependency, now the code only uses scipy solver.
As far as my knowledge goes, there is no decomposition or iterative method that would make
the algorithm faster while preserving the results. I am also sharing some metrics for the comparison
of octave and scipy. I believe SciPy is not all bad, considering they also would optimize their sparse solver
over time.

image

For an image with 118677 pixels (around 350 x 350), using i7-13700K with 10 runs each:

```
Scipy mean time: 5.763958525657654 +- 0.2072540601999276
Octave mean time: 0.7772286891937256 +- 0.018855939984020915
```

_Originally posted by @MrGranddy in https://github.com/Project-MONAI/MONAI/issues/6709#issuecomment-1641391810_

MrGranddy:
Fixes #6767 
### Description

Fixed two minor implementation differences between the official MATLAB code and the MONAI implementation, to confirm also added a new test case where two images and their confidence maps calculated by the official code is added to tests and the MONAI implementation results are checked against there results created by the official code.

Also fixing the issue: added the conjugate gradients solver option. Now the users can utilize it to run the algorithm faster with a trade-off of accuracy of the end result, a range of speed-ups can be achieved with little to no quality loss by tweaking the parameters, the optimal parameters between quality and speed in my experience is set as the default parameters, namely 'cg_tol' and 'cg_maxiter'.

For the CG solver installing PyAMG (https://github.com/pyamg/pyamg) is a requirement, this is because we use it to generate a preconditioner, without it CG does not provide any speed-ups, even slows down the algorithm. This part can be changed if the requirement is not ideal, yet this was the best solution as far as my knowledge goes.

### Types of changes


monai/data/ultrasound_confidence_map.py
```diff
@@ -228,17 +246,18 @@ def confidence_laplacian(self, padded_index: NDArray, padded_image: NDArray, bet
...
+        # on the edge penalty. We directly implement what the official code does.
 
         # Normalize differences
         s = self.normalize(s)
 
         # Gaussian weighting function
         s = -(
-            (np.exp(-beta * s, dtype="float64")) + 1.0e-6
-        )  # --> This epsilon changes results drastically default: 1.e-6
+            (np.exp(-beta * s, dtype="float64")) + 10e-6
```
KumoLiu, line 258: I guess it's a default value provided by the paper. Could you please point it out here? Thanks.

requirements-dev.txt
```diff
@@ -26,7 +26,7 @@ mypy>=1.5.0
 ninja
 torchvision
 psutil
-cucim-cu12; platform_system == "Linux" and python_version >= "3.9" and python_version <= "3.10"
+#cucim-cu12; platform_system == "Linux" and python_version >= "3.9" and python_version <= "3.10"
```
KumoLiu, line 29: I suppose it was a mistaken comment.
```suggestion
cucim-cu12; platform_system == "Linux" and python_version >= "3.9" and python_version <= "3.10"
```

tests/test_ultrasound_confidence_map_transform.py
```diff
@@ -683,6 +662,49 @@ def test_func(self):
...
+
+            transform = UltrasoundConfidenceMapTransform(
+                sink_mode="all", use_cg=True, cg_tol=1.0e-6, cg_maxiter=300, **params
+            )
+            output = transform(input_img)
+
+            import matplotlib.pyplot as plt
+
+            plt.imshow(output)
+            plt.show()
```
KumoLiu, line 704:
```suggestion
```

monai/transforms/intensity/array.py
```diff
@@ -2796,14 +2796,30 @@ class UltrasoundConfidenceMapTransform(Transform):
         mode (str, optional): 'RF' or 'B' mode data. Defaults to 'B'.
         sink_mode (str, optional): Sink mode. Defaults to 'all'. If 'mask' is selected, a mask must be when
             calling the transform. Can be one of 'all', 'mid', 'min', 'mask'.
+        use_cg (bool, optional): Use Conjugate Gradient method for solving the linear system. Defaults to False.
+        cg_tol (float, optional): Tolerance for the Conjugate Gradient method. Defaults to 1e-6.
+        cg_maxiter (int, optional): Maximum number of iterations for the Conjugate Gradient method. Defaults to 200.
```
KumoLiu, line 2801: It would be better mentioned that these args are only used when `use_cg=True`.

tests/test_ultrasound_confidence_map_transform.py
```diff
@@ -683,6 +662,49 @@ def test_func(self):
...
+
+            transform = UltrasoundConfidenceMapTransform(sink_mode="all", **params)
+            output = transform(input_img)
+
+            assert_allclose(output, result_img, rtol=1e-4, atol=1e-4)
+
+    def test_against_official_code_using_cg(self):
+        # This test is to compare the output of the transform with the official code
+        # The official code is available at:
+        # https://campar.in.tum.de/Main/AthanasiosKaramalisCode
```
KumoLiu, line 686: Could you please also mention this in the class docstring? Thanks!

KumoLiu (commented):
Thanks for the PR! Overall looks good to me, several minor comments inline.

@ericspod and @Nic-Ma, do you have any concern to include this third package pyamg?

KumoLiu:
Could you please also take a look at the CI error: https://github.com/Project-MONAI/MONAI/actions/runs/9655433015/job/26631317948?pr=7876#step:13:19809?

monai/data/ultrasound_confidence_map.py
MrGranddy: Hi, sadly there is no mention of this value in the paper, since normally it should serve only as an epsilon value, close to zero, yet in the implementation it does not behave like that at all, I mostly relate this to the problem being not very well conditioned. So in the end, to my best knowledge, this value is just a magic number which changes the results drastically.

That being said, this is exactly how it is implemented in the official MATLAB code: https://campar.in.tum.de/Main/AthanasiosKaramalisCode

I first implemented it as 1.e-6 because 99% of the time you expect it to be 0.000001 yet in this case I also strongly believe that the authors wanted 0.000001 and mistakenly made it 0.00001, so it is a value with lots to speculate about, also made me think a lot about itself, I guess we will never know unless we ask the author, and only if he remembers.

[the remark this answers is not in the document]
MrGranddy: Yes they are exactly equivalent, for the explanation of the value please kindly refer to the answer above.

requirements-dev.txt
MrGranddy: Exactly, my CUDA version was too old to install it, I forgot to revert it to the original, thank you.

tests/test_ultrasound_confidence_map_transform.py
MrGranddy: Added

monai/transforms/intensity/array.py
MrGranddy: Added

ericspod:
The current twine issue appears related to a new version of a dependent package: https://github.com/pypa/twine/issues/977#issuecomment-2186916888 We could add a value in one of our config files to include the missing key or fix the version of `importlib_metadata` to be less than 8.0.

KumoLiu:
> The current twine issue appears related to a new version of a dependent package: [pypa/twine#977 (comment)](https://github.com/pypa/twine/issues/977#issuecomment-2186916888) We could add a value in one of our config files to include the missing key or fix the version of `importlib_metadata` to be less than 8.0.

twine==5.1.0 has been yanked: https://pypi.org/project/twine/#history.
I just rerun the job, it works well now.

KumoLiu (approved):

KumoLiu:
Hi @ericspod, do you have any more comments on this one?

ericspod (approved):
Looks good now.

PR #7887, opened 2024-07-01
Merging Dev Into gen-ai-dev and Undeclared Variable Fixes

ericspod:
### Description

This merges `dev` into `gen-ai-dev` and modifies/fixes some places in code to remove undeclared variable errors raised by pylint. This should all be non-breaking and not changing anything functional in this branch. When completed, the branch should be ready for merging into `dev`. 

This also resolves various conflicts which need to be checked for correctness.

### Types of changes


ericspod:
@KumoLiu There's something really wrong with the DCO check. It's flagging a lot of people, however every commit in the history for this PR appears to be from a PR or is one I've signed. Any idea?

ericspod:
@virginiafdez This PR should address the issue we saw of many files being marked as changed because it merges in `dev`. Please check files over to make sure this is all correct, then when we go to merge `gen-ai-dev` into `dev` we have to check that the changed files are only those with the generative additions and nothing's missing.

KumoLiu:
> @KumoLiu There's something really wrong with the DCO check. It's flagging a lot of people, however every commit in the history for this PR appears to be from a PR or is one I've signed. Any idea?

Hi @ericspod, yes, we have encountered this situation many times before, all due to the problem of email mismatch, for this PR we can manually set the DCO pass first, and then see if there is still a DCO problem when merge into dev.

ericspod:
> all due to the problem of email mismatch,

I'm fine with setting it to pass, though this seems funny since one of the "users" flagged is the monai-bot which wouldn't ever have used a different email. Thanks!

KumoLiu (approved):

PR #7907, opened 2024-07-09
Support download bundles from ngc private registry

KumoLiu:
### Description

Support download from ngc private registry, this download option requires ngc api key.

### Types of changes

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -12,10 +12,12 @@
 from __future__ import annotations
 
 import ast
+import io
 import json
 import os
 import re
 import warnings
+import zipfile
 from collections.abc import Mapping, Sequence
 from pathlib import Path
 from pydoc import locate
@@ -171,6 +173,13 @@ def _get_ngc_bundle_url(model_name: str, version: str) -> str:
     return f"https://api.ngc.nvidia.com/v2/models/nvidia/monaitoolkit/{model_name.lower()}/versions/{version}/zip"
 
 
+def _get_nvstaging_bundle_url(model_name: str, version: str, org: str, team: str) -> str:
+    team_str = ""
+    if team != "no-team":
+        team_str = f"team/{team}"
+    return f"https://api.ngc.nvidia.com/v2/org/{org}/{team_str}/models/{model_name.lower()}/versions/{version}/zip"
+
+
 def _get_monaihosting_bundle_url(model_name: str, version: str) -> str:
     monaihosting_root_path = "https://api.ngc.nvidia.com/v2/models/nvidia/monaihosting"
     return f"{monaihosting_root_path}/{model_name.lower()}/versions/{version}/files/{model_name}_v{version}.zip"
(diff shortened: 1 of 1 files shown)


mingxin-zheng:
Please also fix the docstring here:
```
            In default, the value is achieved from the environment variable BUNDLE_DOWNLOAD_SRC, and
            it should be "ngc", "monaihosting", "github", or "huggingface_hub".
```

Thanks!

monai/bundle/scripts.py
```diff
@@ -171,6 +172,10 @@ def _get_ngc_bundle_url(model_name: str, version: str) -> str:
     return f"https://api.ngc.nvidia.com/v2/models/nvidia/monaitoolkit/{model_name.lower()}/versions/{version}/zip"
 
 
+def _get_nvstaging_bundle_url(model_name: str, version: str, repo: str) -> str:
```
Nic-Ma, line 175: `_get_ngc_private_bundle_url()`

monai/bundle/scripts.py
```diff
@@ -219,6 +224,47 @@ def _download_from_ngc(
     extractall(filepath=filepath, output_dir=extract_path, has_base=True)
 
 
+def _download_from_nvstaging(
```
Nic-Ma, line 227: `ngc_private`

Nic-Ma (approved):
put some comments inline to remove `nvstaging` naming. Others look good to me.

Thanks.

monai/bundle/scripts.py
KumoLiu: Thanks, updated it in the latest commit.

monai/bundle/scripts.py
KumoLiu: Thanks, updated it in the latest commit.

monai/bundle/scripts.py
```diff
@@ -219,6 +224,47 @@ def _download_from_ngc(
...
+
+
+def _get_ngc_token(api_key, retry=0):
+    """Try to connect to NGC."""
+    url = "https://authn.nvidia.com/token?service=ngc"
+    headers = {"Accept": "application/json", "Authorization": "ApiKey " + api_key}
+    if has_requests:
+        response = requests_get(url, headers=headers)
+        if not response.ok:
+            if retry < 3:
```
yiheng-wang-nv, line 259: magic number here? should use a variable to replace 3?

monai/bundle/scripts.py
```diff
@@ -232,7 +278,23 @@ def _get_latest_bundle_version_monaihosting(name):
...
+def _get_latest_bundle_version_private_registry(name, repo, headers=None):
+    url = f"https://api.ngc.nvidia.com/v2/{repo}/models"
+    full_url = f"{url}/{name.lower()}"
+    requests_get, has_requests = optional_import("requests", name="get")
+    if has_requests:
+        headers = {} if headers is None else headers
+        resp = requests_get(full_url, headers=headers)
+        resp.raise_for_status()
+    else:
+        raise ValueError("NGC API requires requests package.  Please install it.")
```
yiheng-wang-nv, line 290: Here has one extra space

KumoLiu: Updated, thanks!

monai/bundle/scripts.py
```diff
@@ -171,6 +172,10 @@ def _get_ngc_bundle_url(model_name: str, version: str) -> str:
     return f"https://api.ngc.nvidia.com/v2/models/nvidia/monaitoolkit/{model_name.lower()}/versions/{version}/zip"
 
 
+def _get_ngc_private_bundle_url(model_name: str, version: str, repo: str) -> str:
+    return f"https://api.ngc.nvidia.com/v2/{repo}/models/{model_name.lower()}/versions/{version}/zip"
```
yiheng-wang-nv, line 176: I'm not sure but should repo be after models?
Like 
```
f"https://api.ngc.nvidia.com/v2/models/{repo}/{model_name.lower()}/versions/{version}/zip"
```

monai/bundle/scripts.py
KumoLiu: I just don't want to introduce a variable here. If retry three time still failed will raise an error. Do you think we need to introduce an argument here?

yiheng-wang-nv: ok, if so I suggest to add a comment above this line

monai/bundle/scripts.py
KumoLiu: No, I guess the link is different with the catlog one.

yiheng-wang-nv (approved):

monai/bundle/scripts.py
KumoLiu: Added, thanks!

ericspod (approved):

PR #7914, opened 2024-07-11
Cleanup warnings when collecting tests

hjmjohnson:
Minimize warning information so that important warnings and errors are more easily accessible

warnings identified with
```bash
python3  -W ignore::DeprecationWarning -m pytest -k dummy
```

Fixing warnings like:
cannot collect test class 'TestTimeAugmentation' because it has a __init__ constructor

Use `__test__ = False` to suppress attempted collection in those non test-classes

### Description

### Types of changes


monai/data/test_time_augmentation.py
```diff
@@ -106,6 +106,8 @@ class TestTimeAugmentation:
             mode, mean, std, vvc = tt_aug(test_data)
     """
 
+    __test__ = False  # indicate to pytest that this class is not intended for collection
```
ericspod, line 109: This class isn't a unit test class, it's used at test time right after training though I'm sure it's used in unit tests. I don't think pytest should ignore this file as it's correctness is part of the tests.

hjmjohnson: This is needed to minimize warnings when using pytest.  It indicates to pytest that this class is not a unit test during the test collection process.  During the test running, these classes are used.  

An alternate solution is to change the name of the class to remove "Test" from the class name so that pytest does not attempt to register it during the test collection phase.

ericspod (approved):

PR #7916, opened 2024-07-14
Try to fix experiment already exist issue in `MLFlowHandler`

KumoLiu:
Try to fixes https://github.com/NVIDIA/NVFlare/issues/2698.

### Types of changes

### DIFF
--- a/monai/handlers/mlflow_handler.py
+++ b/monai/handlers/mlflow_handler.py
@@ -29,6 +29,7 @@ mlflow, _ = optional_import("mlflow", descriptor="Please install mlflow before u
 mlflow.entities, _ = optional_import(
     "mlflow.entities", descriptor="Please install mlflow.entities before using MLFlowHandler."
 )
+MlflowException, _ = optional_import("mlflow.exceptions", name="MlflowException", descriptor="Please install mlflow before using MLFlowHandler.")
 pandas, _ = optional_import("pandas", descriptor="Please install pandas for recording the dataset.")
 tqdm, _ = optional_import("tqdm", "4.47.0", min_version, "tqdm")
 
@@ -236,14 +237,25 @@ class MLFlowHandler:
     def _set_experiment(self):
         experiment = self.experiment
         if not experiment:
-            experiment = self.client.get_experiment_by_name(self.experiment_name)
-            if not experiment:
-                experiment_id = self.client.create_experiment(self.experiment_name)
-                experiment = self.client.get_experiment(experiment_id)
+            for _ in range(3):
+                try:
+                    experiment = self.client.get_experiment_by_name(self.experiment_name)
+                    if not experiment:
+                        experiment_id = self.client.create_experiment(self.experiment_name)
+                        experiment = self.client.get_experiment(experiment_id)
+                    break
+                except MlflowException as e:
+                    if "RESOURCE_ALREADY_EXISTS" in str(e):
+                        time.sleep(1)
+                        continue
+                    else:
+                        raise e
 
         if experiment.lifecycle_stage != mlflow.entities.LifecycleStage.ACTIVE:
             raise ValueError(f"Cannot set a deleted experiment '{self.experiment_name}' as the active experiment")
         self.experiment = experiment
+
+
 
     @staticmethod
     def _get_pandas_dataset_info(pandas_dataset):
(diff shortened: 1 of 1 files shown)


monai/handlers/mlflow_handler.py
```diff
@@ -236,15 +237,26 @@ def start(self, engine: Engine) -> None:
...
+            for _ in range(3):
+                try:
+                    experiment = self.client.get_experiment_by_name(self.experiment_name)
+                    if not experiment:
+                        experiment_id = self.client.create_experiment(self.experiment_name)
+                        experiment = self.client.get_experiment(experiment_id)
+                    break
+                except MlflowException as e:
+                    if "RESOURCE_ALREADY_EXISTS" in str(e):
+                        time.sleep(1)
```
mingxin-zheng, line 249: Do we need a warning here?
```
logger.warning("Experiment already exists; delaying before retrying.")
```

monai/handlers/mlflow_handler.py
```diff
@@ -236,15 +237,26 @@ def start(self, engine: Engine) -> None:
     def _set_experiment(self):
         experiment = self.experiment
         if not experiment:
-            experiment = self.client.get_experiment_by_name(self.experiment_name)
-            if not experiment:
-                experiment_id = self.client.create_experiment(self.experiment_name)
-                experiment = self.client.get_experiment(experiment_id)
+            for _ in range(3):
+                try:
+                    experiment = self.client.get_experiment_by_name(self.experiment_name)
```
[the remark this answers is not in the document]
binliunls: This is caused by multiprocessing issue. Resolved after discussion.

binliunls:
I think one more thing to discuss is whether MONAI should handle the multiprocessing scene of the mlflow handler or users should implement it in their code. If we'd like to do this, we'd better to use a more formal way like a lock for this.

Thanks  
Bin

monai/handlers/mlflow_handler.py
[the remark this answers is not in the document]
KumoLiu: No special reason, I just didn't introduce a new arg like `retry_time` here.

monai/handlers/mlflow_handler.py
KumoLiu: Thanks, added in the latest commit.

KumoLiu:
> I think one more thing to discuss is whether MONAI should handle the multiprocessing scene of the mlflow handler or users should implement it in their code. If we'd like to do this, we'd better to use a more formal way like a lock for this.

A good point. I didn't find a multiprocessing scene, the issue here is that in the nvflare real world tests, it will raise an error during creating experiments in different sites. Perhaps we can add try-catch block here and if there is a real-world case, we can refactor the logic. What do you think? Thanks.

binliunls:
> > I think one more thing to discuss is whether MONAI should handle the multiprocessing scene of the mlflow handler or users should implement it in their code. If we'd like to do this, we'd better to use a more formal way like a lock for this.
> 
> A good point. I didn't find a multiprocessing scene, the issue here is that in the nvflare real world tests, it will raise an error during creating experiments in different sites. Perhaps we can add try-catch block here and if there is a real-world case, we can refactor the logic. What do you think? Thanks.

Sure, we can leave this one as it is. But I think from the flare side, they need to handle this multiprocessing issue.

monai/handlers/mlflow_handler.py
```diff
@@ -236,10 +242,20 @@ def start(self, engine: Engine) -> None:
...
+                    if not experiment:
+                        experiment_id = self.client.create_experiment(self.experiment_name)
+                        experiment = self.client.get_experiment(experiment_id)
+                    break
+                except MlflowException as e:
+                    if "RESOURCE_ALREADY_EXISTS" in str(e):
+                        logger.warning("Experiment already exists; delaying before retrying.")
+                        time.sleep(1)
+                        continue
```
ericspod, line 256: I don't think this line is needed. Furthermore, if the exception is raised three times in a row the loop will silently do nothing. In place of this line we should check if this is the last time through the loop and raise the exception if so instead of sleeping. The check after this loop isn't sufficient since it raises a different exception which won't have the  information that the MlflowException would have.

KumoLiu: Thanks @ericspod! A good point, I removed the line and raise error if _retry_time meet 3. Could you please help review it again? Thanks.

Nic-Ma (approved):

monai/handlers/mlflow_handler.py
ericspod: I think it's good now, thanks!

ericspod (approved):

PR #7934, opened 2024-07-20
Refactor Export for Model Conversion and Saving

### LINKED ISSUE #6375: Make the _export function support saver input

Han123su:
Fixes #6375 .

### Description
Changes to be made based on the [previous discussion #7835](https://github.com/Project-MONAI/MONAI/pull/7835).

Modify the `_export` function to call the `saver` parameter for saving different models. Rewrite the `onnx_export` function using the updated `_export` to achieve consistency in model format conversion and saving.

* Rewrite `onnx_export` to call `_export` with `convert_to_onnx` and appropriate `kwargs`.
* Add a `saver: Callable` parameter to `_export`, replacing `save_net_with_metadata`.
* Pass `save_net_with_metadata` function wrapped with `partial` to set parameters like `include_config_vals` and `append_timestamp`.

### Types of changes


monai/bundle/scripts.py
```diff
@@ -1173,6 +1175,8 @@ def _export(
     Args:
         converter: a callable object that takes a torch.nn.module and kwargs as input and
             converts the module to another type.
+        saver: a callable object that takes the converted model and a filepath as input and
```
ericspod, line 1178: Here you could state the callable accepts the network to save, the path to save to, and a dictionary of extra JSON files (name -> contents) to include with this.

monai/bundle/scripts.py
```diff
@@ -1212,14 +1216,8 @@ def _export(
...
 
-    save_net_with_metadata(
-        jit_obj=net,
-        filename_prefix_or_stream=filepath,
-        include_config_vals=False,
-        append_timestamp=False,
-        meta_values=parser.get().pop("_meta_", None),
-        more_extra_files=extra_files,
-    )
+    saver(net, filepath, more_extra_files=extra_files)
```
ericspod, line 1219: When calling `saver`, you could include the argument `meta_values=parser.get().pop("_meta_", None)` so that this line isn't needed in the partial constructs you've added, which could then fit onto one line. You would have to state that `saver` expects the `meta_values` argument in the doctring above. You can leave this unchanged if you want, only a suggestion.

ericspod (commented):
It looks much better now. I have a comment on docstrings and something optional about moving arguments around, but otherwise if all the tests pass this is good to go I feel.

ericspod:
@binliunls This PR is response to an issue you've raised previously, would you have any comment here? Thanks!

monai/bundle/scripts.py
```diff
@@ -1212,14 +1216,8 @@ def _export(
     # add .json extension to all extra files which are always encoded as JSON
     extra_files = {k + ".json": v for k, v in extra_files.items()}
```
binliunls, line 1217: I would recommand to move the whole `extra_files` logic from the `_export` function to the `ckpt_export` function , since this parameter is unique for the latter one.  You can fix it as a partial function parameter as `include_config_vals`.  Then the saver call in the next line will only need two parameters which looks more consistent to me.

Thanks  
Bin

monai/bundle/scripts.py
```diff
@@ -1318,17 +1316,23 @@ def onnx_export(
...
-        Checkpoint.load_objects(to_load={key_in_ckpt_: net}, checkpoint=ckpt_file_)
-    else:
-        ckpt = torch.load(ckpt_file_)
-        copy_model_state(dst=net, src=ckpt if key_in_ckpt_ == "" else ckpt[key_in_ckpt_])
 
     converter_kwargs_.update({"inputs": inputs_, "use_trace": use_trace_})
-    onnx_model = convert_to_onnx(model=net, **converter_kwargs_)
-    onnx.save(onnx_model, filepath_)
+
+    def save_onnx(onnx_obj: Any, filename_prefix_or_stream: str | IO[Any], **kwargs: Any) -> None:
```
binliunls, line 1322: In this version, the `_export` function only supports exporting as files. I am not sure why the `| IO[Any]` is added here. If this is for exporting to IO, please also add this to the `_export` function and implement the IO exporting in it. Otherwise, I think we should remove this type hints.

Thanks  
Bin

monai/bundle/scripts.py
Han123su: Alright, your suggestion is very good. After confirming all the changes, I will clearly explain the `saver` parameters. Thank you!

monai/bundle/scripts.py
Han123su: You are right. The `save_onnx` function here in `onnx_export` focuses only on saving the ONNX model. It handles file paths instead of `IO` objects, so it can be removed to avoid confusion.

Han123su:
Hi, @ericspod  and @binliunls, thank you very much for your suggestions. However, it seems that there might be a slight conflict in the ideas behind the two proposed changes, [suggestion 1](https://github.com/Project-MONAI/MONAI/pull/7934/files/5798d72f92929c09561e607a5972e5616b876666#r1688144849) and [suggestion 2](https://github.com/Project-MONAI/MONAI/pull/7934/files/5798d72f92929c09561e607a5972e5616b876666#r1689234159). I would like to explain my thoughts on these two modifications. Thank you!

---
* [suggestion 1](https://github.com/Project-MONAI/MONAI/pull/7934/files/5798d72f92929c09561e607a5972e5616b876666#r1688144849)

`ckpt_export`/`trt_export` :
```python
save_ts = partial(
        save_net_with_metadata,
        include_config_vals=False,
        append_timestamp=False,
)
```
`_export` :
```python
def _export(
    ...
) -> None:
    ...
    ...
    meta_values = parser.get().pop("_meta_", None)  
    saver(net, filepath, more_extra_files=extra_files, meta_values=meta_values)    

    logger.info(f"exported to file: {filepath}.")
```
When calling `saver`, directly passing `meta_values=parser.get().pop("_meta_", None)` as a parameter is feasible and makes partial more concise. Although this parameter may not be required by ONNX and might be slightly inconsistent, it does not affect the saving process.


* [suggestion 2](https://github.com/Project-MONAI/MONAI/pull/7934/files/5798d72f92929c09561e607a5972e5616b876666#r1689234159)

`ckpt_export`/`trt_export` :
```python
def ckpt_export/trt_export(
    ...
    
) -> None:
    ...
    ...
    ################## Repeat part ####################
    extra_files: dict = {}
    for i in ensure_tuple(config_file_):
        filename = os.path.basename(i)
        filename, _ = os.path.splitext(filename)
        if filename in extra_files:
            raise ValueError(f"Filename part '{filename}' is given multiple times in config file list.")
        extra_files[filename] = json.dumps(ConfigParser.load_config_file(i)).encode()
    extra_files = {k + ".json": v for k, v in extra_files.items()}
    ################################################

    save_ts = partial(
        save_net_with_metadata,
        include_config_vals=False,
        append_timestamp=False,
        meta_values=parser.get().pop("_meta_", None),
        more_extra_files=extra_files,   
    )
    ...
```
`_export` :
```python
def _export(
    ...
) -> None:
    ...
    ...
    saver(net, filepath,)   
    ...
```
Moving the `extra_files` logic from `_export` to `ckpt_export`/`trt_export` is a viable approach since ONNX storage indeed does not require this parameter. This method would require two parameters when calling `saver`, making it more consistent. However, doing so might introduce previously discussed issues and increase code redundancy (as mentioned in the "Repeat part" section above).

---
My thought is that we might perhaps adopt  [suggestion 1](https://github.com/Project-MONAI/MONAI/pull/7934/files/5798d72f92929c09561e607a5972e5616b876666#r1688144849), calling the `saver` function and passing `meta_values=parser.get().pop("_meta_", None)` to make the partial simpler, without excessive repetition. Although this parameter is not required for ONNX storage, I have used the `**kwargs: Any` parameter in the definition of `save_onnx` so that it can accept other parameters. This won't actually affect the original ONNX storage.
```python
def save_onnx(onnx_obj: Any, filename_prefix_or_stream: str , **kwargs: Any) -> None:
    onnx.save(onnx_obj, filename_prefix_or_stream)
```

Thank you both for reading this explanation. I hope to receive feedback again to make this even better. Thank you very much!

monai/bundle/scripts.py
Han123su: Thank you for your suggestion! This idea will be discussed along with another proposed change in the comments below.

monai/bundle/scripts.py
Han123su: Thank you for your suggestion! This idea will be discussed along with another proposed change in the comments below.

KumoLiu:
Hi @Han123su, thanks for the detailed explanation. I would prefer the first proposal since parser has been passed to `_export` which make the meta_values more straightforward to get.
@ericspod @binliunls Could you please help review it again and share your comments. Thanks!

ericspod:
> @ericspod @binliunls Could you please help review it again and share your comments. Thanks!

I am for suggestion 1, I think it looks good now with that added and with the type issue fixed mentioned by @binliunls. Thanks!

Han123su:
> Hi @Han123su, thanks for the detailed explanation. I would prefer the first proposal since parser has been passed to `_export` which make the meta_values more straightforward to get. 

Ok. Thank you so much for your suggestions!

Han123su:
> I am for suggestion 1, I think it looks good now with that added and with the type issue fixed mentioned by @binliunls. Thanks!

Got it, thank you for your valuable feedback!

Han123su:
> Hi, @ericspod and @binliunls, thank you very much for your suggestions. However, it seems that there might be a slight conflict in the ideas behind the two proposed changes, [suggestion 1](https://github.com/Project-MONAI/MONAI/pull/7934/files/5798d72f92929c09561e607a5972e5616b876666#r1688144849) and [suggestion 2](https://github.com/Project-MONAI/MONAI/pull/7934/files/5798d72f92929c09561e607a5972e5616b876666#r1689234159). I would like to explain my thoughts on these two modifications. Thank you!
> 
> * [suggestion 1](https://github.com/Project-MONAI/MONAI/pull/7934/files/5798d72f92929c09561e607a5972e5616b876666#r1688144849)
> 
> `ckpt_export`/`trt_export` :
> 
> ```python
> save_ts = partial(
>         save_net_with_metadata,
>         include_config_vals=False,
>         append_timestamp=False,
> )
> ```
> 
> `_export` :
> 
> ```python
> def _export(
>     ...
> ) -> None:
>     ...
>     ...
>     meta_values = parser.get().pop("_meta_", None)  
>     saver(net, filepath, more_extra_files=extra_files, meta_values=meta_values)    
> 
>     logger.info(f"exported to file: {filepath}.")
> ```
> 
> When calling `saver`, directly passing `meta_values=parser.get().pop("_meta_", None)` as a parameter is feasible and makes partial more concise. Although this parameter may not be required by ONNX and might be slightly inconsistent, it does not affect the saving process.
> 
> * [suggestion 2](https://github.com/Project-MONAI/MONAI/pull/7934/files/5798d72f92929c09561e607a5972e5616b876666#r1689234159)
> 
> `ckpt_export`/`trt_export` :
> 
> ```python
> def ckpt_export/trt_export(
>     ...
>     
> ) -> None:
>     ...
>     ...
>     ################## Repeat part ####################
>     extra_files: dict = {}
>     for i in ensure_tuple(config_file_):
>         filename = os.path.basename(i)
>         filename, _ = os.path.splitext(filename)
>         if filename in extra_files:
>             raise ValueError(f"Filename part '{filename}' is given multiple times in config file list.")
>         extra_files[filename] = json.dumps(ConfigParser.load_config_file(i)).encode()
>     extra_files = {k + ".json": v for k, v in extra_files.items()}
>     ################################################
> 
>     save_ts = partial(
>         save_net_with_metadata,
>         include_config_vals=False,
>         append_timestamp=False,
>         meta_values=parser.get().pop("_meta_", None),
>         more_extra_files=extra_files,   
>     )
>     ...
> ```
> 
> `_export` :
> 
> ```python
> def _export(
>     ...
> ) -> None:
>     ...
>     ...
>     saver(net, filepath,)   
>     ...
> ```
> 
> Moving the `extra_files` logic from `_export` to `ckpt_export`/`trt_export` is a viable approach since ONNX storage indeed does not require this parameter. This method would require two parameters when calling `saver`, making it more consistent. However, doing so might introduce previously discussed issues and increase code redundancy (as mentioned in the "Repeat part" section above).
> 
> My thought is that we might perhaps adopt [suggestion 1](https://github.com/Project-MONAI/MONAI/pull/7934/files/5798d72f92929c09561e607a5972e5616b876666#r1688144849), calling the `saver` function and passing `meta_values=parser.get().pop("_meta_", None)` to make the partial simpler, without excessive repetition. Although this parameter is not required for ONNX storage, I have used the `**kwargs: Any` parameter in the definition of `save_onnx` so that it can accept other parameters. This won't actually affect the original ONNX storage.
> 
> ```python
> def save_onnx(onnx_obj: Any, filename_prefix_or_stream: str , **kwargs: Any) -> None:
>     onnx.save(onnx_obj, filename_prefix_or_stream)
> ```
> 
> Thank you both for reading this explanation. I hope to receive feedback again to make this even better. Thank you very much!

@binliunls, sorry to bother you, could you please provide some suggestions in this discussion? Thank you!

Han123su:
@binliunls, sorry to bother you, could you please provide some suggestions on the above discussion? Thank you!

binliunls:
> monai/bundle/scripts.py

I am OK with the suggestion 1.

Han123su:
Hi @ericspod @KumoLiu @binliunls, I have made modifications based on the discussions above. Please help review them again, and let me know if there are any further changes needed. Thanks for taking the time to check it out.

ericspod (approved):
I'm good now if @mingxin-zheng @KumoLiu are. Thanks!

KumoLiu (approved):
Thanks for the update, looks good to me now.
Hi @binliunls, do you have any further comments here, otherwise I will try to merge this one.

binliunls (approved):

binliunls:
> Thanks for the update, looks good to me now. Hi @binliunls, do you have any further comments here, otherwise I will try to merge this one.

LGTM.

PR #7942, opened 2024-07-23
Image space and World Space convert transforms

KumoLiu:
Fixes # .

### Description
Add `ImageToWorldSpace` and `WorldToImageSpace`

### Types of changes


monai/transforms/utility/array.py
```diff
@@ -1715,3 +1715,84 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+
+        homogeneous = concatenate((data_, torch.ones((data_.shape[0], 1))), axis=1)
+        transformed_homogeneous = torch.matmul(_affine, homogeneous.T)
+        transformed_coordinates = transformed_homogeneous[:-1].T
+        out, *_ = convert_to_dst_type(transformed_coordinates, data, dtype=self.dtype)
+
+        extra_info = {
+            "mode": self.mode,
+            "dtype": str(self.dtype)[6:],  # dtype as string; remove "torch": torch.float32 -> float32
+            "affine": affine if invert else affine_t,
```
aylward, line 1768: This logic seems backwards to me, but I haven't parsed the full intent of this yet.

KumoLiu: Yeah, it may lead confusing here. I'm not convert the affine back. The affine here saved in `extra_info` is only used for inverse.
Given an example, if the transformed affine is `A`, then the affine logged here is the inverse matrix of `A` which will be used for the inverse.
I will think more and see whether I can make it clearer here.

Thanks!

vikashg:
Hey @KumoLiu, I am going over the code. Can you please share the data `spleen_60.nii.gz` along with the `.mrk.json` file in a google drive? 
Did you use Monai Label to annotate this?

KumoLiu:
> Hey @KumoLiu, I am going over the code. Can you please share the data `spleen_60.nii.gz` along with the `.mrk.json` file in a google drive? Did you use Monai Label to annotate this?

Sure, I have uploaded the Google Drive link. I annotated it by using 3d slicer.

atbenmurray:
My apologies, I had to write slides for a presentation today and I've had no time to review this yet.

vikashg:
Thanks @KumoLiu I am going over it.

vikashg:
Hello Yun, 
Thanks for sharing the images. The code looks all right to me and it is quite clear to me what you are trying to do. I will suggest to test 2 things if you can (I would have but I do not have slicer installed) 
1. I looked at the image and it has the spacing as `(0.792, 0.792, 5.0)`. In the test case you have marked all the points in the same plane. For the sake of completeness can you please mark few points in a different plane and check that they too are transformed correctly. I do not see a reason why it would not. But it is just a good idea to make sure. 

2. I see that you have marked the points on the axial plane (lets call it xy plane), where the pixel spacing is same in both x and y directions. Can you please mark few points in the coronal or sagittal planes and do the transformation. It will ensure that we are transforming the points correctly when the spacing is not same in both directions. 

Again, I must say, I do not think that there will be a problem, but just to make sure it works in the aforementioned cases. 
Thanks once again

vikashg (approved):
Thanks for testing the coronal and sagittal pieces @KumoLiu.

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1715,101 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+        mode=CoordinateTransformMode.IMAGE_TO_WORLD,
+        affine_lps_to_ras: bool = False,
+    ) -> None:
+        self.dtype = dtype
+        self.mode = mode
+        self.affine_lps_to_ras = affine_lps_to_ras
+
+
+    @staticmethod
+    def apply_affine_to_points(data, affine, dtype):
```
atbenmurray, line 1746: I think apply_affine_to_points should be a standalone utility implementation (see PR #7955) as we will need this functionality deep in the pipeline execution.
I also think that we would be better off loading points into a form suitable for multiplication with transformation matrices (i.e. (x,y,1) or (x,y,z,1) rather than constantly adding and removing the final element (see monai/transforms/lazy/functional.py L299 in PR #7955)

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1715,101 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+        out, *_ = convert_to_dst_type(transformed_coordinates, data, dtype=dtype)
+
+        return out
+
+    def transform_coordinates(self, data, i2w_affine):
+        """
+        Transform coordinates using an affine transformation matrix.
+
+        Args:
+            data: The input coordinates, assume to be in shape (C, N, 3 or 4).
```
mingxin-zheng, line 1762: Hi @KumoLiu , thank you for the PR.

The `transform_coordinates` method assumes that the input `data` has shape `(C, N, 3 or 4)`, but does not validate this. Consider adding input validation to ensure correct usage.

monai/transforms/utility/array.py
KumoLiu: > apply_affine_to_points should be a standalone utility implementation

`apply_affine_to_points ` can be reused as it's a staticmethod. I also reused it in the Flip transform.

> I also think that we would be better off loading points into a form suitable for multiplication with transformation matrices (i.e. (x,y,1) or (x,y,z,1) rather than constantly adding and removing the final element

I didn't get the point for this one. The axis is only added and removed in this transform, not affect any place else.

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1715,101 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
         if self._do_transform:
             img = self.filter(img)
         return img
+
+
+class CoordinateTransform(InvertibleTransform, Transform):
```
mingxin-zheng, line 1720: Please add the type hints for the member function methods. Thanks!

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1715,101 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+        Args:
+            data: The input coordinates, assume to be in shape (C, N, 3 or 4).
+            i2w_affine: A 3x3 or 4x4 affine transformation matrix.
+            invert: Whether to invert the affine matrix.
+
+        Returns:
+            Transformed coordinates.
+        """
+        data = convert_to_tensor(data, track_meta=get_track_meta())
+        data_: torch.Tensor = convert_to_tensor(data, track_meta=False, dtype=self.dtype)
```
mingxin-zheng, line 1770: The following code is duplicated in `apply_affine_to_points` and `transform_coordinates`
```
data = convert_to_tensor(data, track_meta=get_track_meta())
data_: torch.Tensor = convert_to_tensor(data, track_meta=False, dtype=dtype)
```
Please consider if one of them is not necessary.

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1715,101 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+            i2w_affine = orientation_ras_lps(i2w_affine)
+
+        w2i_affine = linalg_inv(i2w_affine)
+        _affine = w2i_affine if self.mode == CoordinateTransformMode.WORLD_TO_IMAGE else i2w_affine
+
+        out = self.apply_affine_to_points(data_, _affine, self.dtype)
+
+        extra_info = {
+            "mode": self.mode,
+            "dtype": str(self.dtype)[6:],  # dtype as string; remove "torch": torch.float32 -> float32
```
mingxin-zheng, line 1783: How about having a small util function for this?

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1715,101 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+        data_: torch.Tensor = convert_to_tensor(data, track_meta=False, dtype=dtype)
+
+        homogeneous = concatenate((data_[0], torch.ones((data_[0].shape[0], 1))), axis=1)
+        transformed_homogeneous = torch.matmul(affine, homogeneous.T)
+        transformed_coordinates = transformed_homogeneous[:-1].T.unsqueeze(0)
+        out, *_ = convert_to_dst_type(transformed_coordinates, data, dtype=dtype)
+
+        return out
+
+    def transform_coordinates(self, data, i2w_affine):
```
mingxin-zheng, line 1757: `i2w` seems difficult to understand. I would suggest to use the full term `image_to_world`

monai/transforms/utility/array.py
atbenmurray: Ahh, but the points themselves will be transformed on multiple occasions in a typical pipeline. This means that each time, they are transformed, the final element needs to be added and then removed. If we just add it up front and only remove it on export, then there is less tensor churn

monai/transforms/utility/array.py
atbenmurray, line 1720: I'm struggling with why this is a class. Is it specifically for converting points from "image" to "world" and vice versa or is it a base class for geometry? We don't have a base raster class and I'm not sure a base class can be defined that would cover all of the specific transforms cleanly. We have a functional layer for most of our operations these days; I'd argue that all baseline geometry should be defined at the functional level and wrapped by classes, rather than defined as classes.

monai/transforms/utility/array.py
KumoLiu: Yes, Ben. I know your points. But after deeper investigation, I feel we didn't enhance other spatial transforms to be compatible with the geometric data, we can just use this transform once to convert the point back to the transformed image space, right?

atbenmurray: I don't think so.
Example pipeline:
LoadImaged, LoadGeometryd, Rotated, Zoomd, Spacingd
Rotated will need to transform the points, Zoomd will need to transform the points, Zoomd will need to transform the points, and Spacingd will need to transform the raster data but apply a "no-op" to the geometry data so the applied entries are all the same length. We probably don't want to modify the geometry tensors twice by adding a tensor of [1s] then removing it after each operation. Please look at PR #7955. I've enabled geometry for many of the spatial transforms already.

atbenmurray:
I think that we will need to have three tensor kinds given that we have expressed the need to support both image and world space coordinates for geometry:

```
class KindKeys(StrEnum):
    """
    This class provides an effective way to reference data types such as pixel-based data
    and point-like data that consists of point coordinates.

    - PIXEL: Represents data that corresponds to pixel-based data.
    - IMAGE_SPACE_POINT: Image-space point refers to geometry data whose point coordinates
      are specified relative to an image's pixel coordinate system
    - WORLD_SPACE_POINT: World-space point refers to geometry data whose point coordinates
      are specified in world space in some physical units such as mm
    """

    PIXEL = "pixel"
    IMAGE_SPACE_POINT = "image-space point"
    WORLD_SPACE_POINT = "world-space point"
```

monai/transforms/utility/array.py
KumoLiu: Thanks, have updated.

monai/transforms/utility/array.py
KumoLiu: Have added a util function for this one.

monai/transforms/utility/array.py
KumoLiu: Thanks, moved duplicated code here.

monai/transforms/utility/array.py
KumoLiu: Added, thanks.

monai/transforms/utility/array.py
KumoLiu: Add shape check here, thanks.

monai/transforms/utility/dictionary.py
```diff
@@ -1740,6 +1742,57 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> dict[Hashable, N
         return d
 
 
+class CoordinateTransformd(MapTransform, InvertibleTransform):
```
ericspod, line 1745: Hi @KumoLiu,
As we discussed today in regards to https://github.com/Project-MONAI/tutorials/pull/1795 I'd propose changing the name of this class and it's operations to make it clearer what's going on. What the class does is take the affine from whatever `refer_key` references and applying that to those in `keys` which for our purposes at the moment are points in space. I would recommend a name like `ApplyTransformToPointsd` or just `ApplyTransformd` if we want to add other things besides points in the future (normals, for example). 

I would also not have a `mode` keyword but instead a boolean stating whether the affine is applied or the inverse of the affine. If `inverse` is True instead of its default False, the inverse of the affine is applied to points (and so of course the `inverse()` operation applies the original affine). I think it helps to be explicit about what the transform is doing rather tha using "image-to-world" and other terms, this was what confused me a bit earlier. If people have reason to apply transforms in some bizarre order this makes it clear what is being done so they can choose what is a forward or inverse transformation. 

I'm not sure if we can do something about the `affine_lps_to_ras` argument here which doesn't seem clean here. I know what when loading Nifti files we fix the header and can then use `Orientationd` to fix the orientation, can we not do something with ITK-loaded images and not need this value? You've looked into this longer than I have so I may be off here, if it's harder to fix we can come back to this later.

It might also be helpful if in this class and the array version (whatever we end calling it) we can explicitly specify the affine to apply, either as a provided `__init__` or `__call__` argument, or as a dictionary value referred to by key. 

Thanks!

KumoLiu: Hi @ericspod, Thank you for all the suggestions. I try to address all of them.

- I renamed the transform to `ApplyTransformToPointsd`, as it currently only applies to points. The name `ApplyTransformd` felt too broad—what do you think?

- I removed the `mode` argument and opted to use `invert_affine` instead. I avoided using `inverse` to prevent conflicts with the `inverse` function.

- I also added `affine` in both the `__call__` and `__init__` methods.

- The `affine_lps_to_ras` parameter is used to ensure that points and images are aligned within the same coordinate system. This parameter is set to True only if the data is loaded by ITK with affine_lps_to_ras=True. For example, if users save annotations (points) using 3DSlicer, they are saved in LPS format. However, when loading images with ITKReader and using affine_lps_to_ras=True, the images are in RAS format. Therefore, we need this flag to ensure both points and images are in the same coordinate system. 
To be honest, it might be better to add this argument in something like LoadPoint, similar to what we do when loading images. However, since we currently don’t have a transform specifically for loading points, could we leave this argument here for now? Once we develop a loading function for geometric data, we can move this argument there.
https://github.com/Project-MONAI/MONAI/blob/a5fbe716378948630783deef8ee435e7e3bdc918/monai/data/image_reader.py#L176

Could you please help review the PR again? I also included the unittests. Thanks again!

KumoLiu:
Hi @vikashg @atbenmurray @aylward @Nic-Ma @mingxin-zheng, I tried to address all the comments we discussing before and also included the test. Could you please help review the PR again and let me know if you have further suggestion? Thanks.

monai/transforms/utility/dictionary.py
ericspod: Let's look into adding `affine_lps_to_ras` to `LoadPoints` or whatever loading class we define, you're right that leaving it here for now makes sense.

vikashg: 100 percent agree with this statement "To be honest, it might be better to add this argument in something like `LoadPoint`, similar to what we do when loading images." 
I would say lets leave it here for now. But we should consider expanding the `LoadPoint` definition to like `LoadGeometry` (or something like that), where we can pass an additional parameter like `primitive_type` with options like `box`,  `circle`. I don't think we need to cover anything more. `box`, `circle` and `points` would be sufficient.

vikashg:
Thanks @KumoLiu, I am going over it again.

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1715,126 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+    def apply_affine_to_points(data: torch.Tensor, affine: torch.Tensor, dtype: DtypeLike | torch.dtype):
+        data_: torch.Tensor = convert_to_tensor(data, track_meta=False, dtype=torch.float64)
+        affine = to_affine_nd(data_.shape[-1], affine)
+
+        homogeneous = concatenate((data_, torch.ones((data_.shape[0], data_.shape[1], 1))), axis=2)
+        transformed_homogeneous = torch.matmul(affine, homogeneous.transpose(1, 2))
+        transformed_coordinates = transformed_homogeneous[:, :-1, :].transpose(1, 2)
+        out, *_ = convert_to_dst_type(transformed_coordinates, data, dtype=dtype)
+
+        return out
```
ericspod, line 1761: This could be in a utility function instead, there may be need to apply such transforms in other contexts. Also you can avoid doing transpositions of the data by doing matmul with the transposed affine as the second arg, ie. `torch.matmul(homogeneous, affine.T)` or something like this. Also (but check my math) if the bottom row of the matrix is `[0, 0, 0, 1]` you don't need to add the fourth component, just matmul with the upper 3x3 of `affine` then add the translation component of the matrix.

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1715,126 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+
+        affine_inv = linalg_inv(affine)
+
+        if self.invert_affine:
+            _affine = affine_inv
+            # consider the affine transformation already applied to the data in the world space
+            if applied_affine is not None:
+                _affine = _affine @ linalg_inv(applied_affine)
+        else:
+            _affine = affine
```
ericspod, line 1794:
```suggestion
        _affine = affine
        if self.invert_affine:
            _affine = linalg_inv(affine)
            # consider the affine transformation already applied to the data in the world space
            if applied_affine is not None:
                _affine = _affine @ linalg_inv(applied_affine)            
```

I remember that some checkers have issue with variables declared in `if-else` blocks and not beforehand, these are erroneously flagged as being non-existent.

KumoLiu: Thanks, updated it.

monai/transforms/utility/array.py
KumoLiu: Hi @ericspod, agreed and I moved it under transforms/utils.

monai/transforms/utils.py
```diff
@@ -2509,5 +2511,25 @@ def distance_transform_edt(
...
+        data: input data to apply affine transformation, should be a tensor of shape (C, N, 2 or 3).
+        affine: affine matrix to be applied, should be a tensor of shape (3, 3) or (4, 4).
+        dtype: output data dtype.
+    """
+    data_: torch.Tensor = convert_to_tensor(data, track_meta=False, dtype=torch.float64)
+    affine = to_affine_nd(data_.shape[-1], affine)
+
+    homogeneous = concatenate((data_, torch.ones((data_.shape[0], data_.shape[1], 1))), axis=2)
+    transformed_homogeneous = torch.matmul(homogeneous, affine.T)
+    transformed_coordinates = transformed_homogeneous[:, :, :-1]
```
ericspod, line 2528:
```suggestion
    transformed_coordinates = transformed_homogeneous[:, :, :-1] / transformed_homogeneous[:, :, -1:]
```
Ideally the `w` component of the output vectors should be all ones but this isn't always the case and vectors should be scaled by this.

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1716,111 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
         if self._do_transform:
             img = self.filter(img)
         return img
+
+
+class ApplyTransformToPoints(InvertibleTransform, Transform):
+    """
+    Transform points between image coordinates and world coordinates.
```
ericspod, line 1723: I think here and in the dictionary version we should state that the point tensor should have shape (C, N, 2|3) and state what C and N dimensions represent.

ericspod: We should also state that affine matrices are 3x3 if 2D points and 4x4 if 3D.

monai/transforms/utils.py
```diff
@@ -2509,5 +2511,25 @@ def distance_transform_edt(
     return convert_data_type(r_vals[0] if len(r_vals) == 1 else r_vals, output_type=type(img), device=device)[0]
 
 
+def apply_affine_to_points(data: torch.Tensor, affine: torch.Tensor, dtype: DtypeLike | torch.dtype):
+    """
+    apply affine transformation to a set of points.
+
+    Args:
+        data: input data to apply affine transformation, should be a tensor of shape (C, N, 2 or 3).
```
ericspod, line 2519: Here also we should mention what C and N represent.

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1716,111 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+            2) The `ITKReader` has `affine_lps_to_ras=True`,
+            3) The data is in world coordinates.
+            This ensures the correct application of the affine transformation between LPS (left-posterior-superior)
+            and RAS (right-anterior-superior) coordinate systems. This argument ensures the points and the affine
+            matrix are in the same coordinate system.
+    """
+
+    def __init__(
+        self,
+        dtype: DtypeLike | torch.dtype = torch.float64,
```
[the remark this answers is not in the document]
KumoLiu: Agreed, updated, thanks!

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1716,111 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+            Transformed coordinates.
+        """
+        data = convert_to_tensor(data, track_meta=get_track_meta())
+        applied_affine = data.affine if isinstance(data, MetaTensor) else None
+
+        if affine is None and self.invert_affine:
+            raise ValueError("affine must be provided when invert_affine is True.")
+
+        affine = applied_affine if affine is None else affine
+        affine = convert_data_type(affine, dtype=torch.float64)[0]  # always convert to float64 for affine
```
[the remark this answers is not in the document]
KumoLiu: Hi @ericspod, I have updated this part and mentioned it in the docstring.

monai/transforms/utility/dictionary.py
```diff
@@ -1740,6 +1742,68 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> dict[Hashable, N
...
+        )
+
+    def __call__(self, data: Mapping[Hashable, torch.Tensor]):
+        d = dict(data)
+        refer_data = d[self.refer_key] if self.refer_key is not None else None
+        if isinstance(refer_data, MetaTensor):
+            affine = refer_data.affine
+        else:
+            warnings.warn("No reference affine find in the refer key, will use the affine derived from the data.")
+            affine = self.affine
```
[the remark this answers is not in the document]
KumoLiu: Hi @ericspod, good point, I updated it to be `affine = getattr(refer_data, "affine", refer_data)`, and also update the docstring for the `refer_key` to be:
```
        refer_key: The key of the reference item used for transformation.
            It can directly refer to an affine or an image from which the affine can be derived.
```
Do you think this good now?

monai/transforms/utility/dictionary.py
```diff
@@ -1740,6 +1742,68 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> dict[Hashable, N
...
+        refer_key: str | None = None,
+        dtype: DtypeLike | torch.dtype = torch.float64,
+        affine: torch.Tensor | None = None,
+        invert_affine: bool = True,
+        affine_lps_to_ras: bool = False,
+        allow_missing_keys: bool = False,
+    ):
+        MapTransform.__init__(self, keys, allow_missing_keys)
+        self.refer_key = refer_key
+        self.affine = affine
```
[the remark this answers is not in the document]
KumoLiu: Agreed, should directly pass it to the array version, updated it! Thanks.

monai/transforms/utility/array.py
KumoLiu: Thanks! Update the docstring, and for affine I only mention it should be `3x3 or 4x4` since `to_affine_nd` will modify the affine based on the input shape.
https://github.com/Project-MONAI/MONAI/blob/1a8afd18d5fb132efa26d6f73c3bf1fdbccd985d/monai/data/utils.py#L945

monai/transforms/utils.py
KumoLiu: Also update this one!

monai/transforms/utils.py
KumoLiu: Hi @ericspod, I didn't get the point for this update here, this change will scale transformed_coordinates which make output not as expected.

monai/transforms/utility/dictionary.py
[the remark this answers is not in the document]
ericspod: I think this is good, thanks!

monai/transforms/utility/array.py
ericspod: `to_affine_nd` will ensure that the affine will be correct but I think it's important to mention that 3x3 and 4x4 matrices can be used and when, so that it's clear when one or the other is needed or what happens to a 4x4 when 2D points are provided. It should be clear that users can provide a 3x3 for 2D, and don't have to use a 4x4 nor add an unneeded Z dimension that's all 0.

KumoLiu: Got it, update the docstring!

monai/transforms/utils.py
ericspod: I might be wrong here but this is what I thought the math was so someone else please check. For a given affine matrix, if the bottom row isn't $(0\ 0\ 0\ 1)$ then the point you'll get won't necessarily have 1 as the value of its w component. The idea is that the resulting vector should be scaled by that $W$ value to get the correct coordinate. 

If we have some 4x4 affine matrix $A$ and some vector $P$, this vector has to be 1x4 with a $W$ component added which is typically 1:

$$
P = \begin{pmatrix} 
X \\ 
Y \\ 
Z \\ 
W
\end{pmatrix}
$$

Once we apply the matrix multiplication we get $P'$:

$$
P' = AP = \begin{pmatrix} 
X' \\ 
Y' \\ 
Z' \\ 
W'
\end{pmatrix}
$$

This is a 4D vector, so to get the 3D vector we divide by $W'$:

$$
P'_{3D} =\begin{pmatrix} 
X'/W' \\ 
Y'/W' \\ 
Z'/W' 
\end{pmatrix}
$$

If the bottom row of $A$ is $(0\ 0\ 0\ 1)$ then $W'$ will be all 1, in which case it's safe to just remove it, but in the general case the above should be done. 

I'm also surprised how much I can do with Latex here.

KumoLiu: > I'm also surprised how much I can do with Latex here.

Amazing!

Got it now. I generally assume that w is 1 by default, and I'm uncertain about when it will be different. Moreover, I explicitly adding W in line 2526, so the user doesn’t need to handle it.

ericspod: For what we're doing adding $W$ is fine, though you could check if you're given 4D points and not add it in that case (or if you have 3D points and a 3x3 matrix), but since we're dealing with 2D/3D cartesian points we know adding a dimension of ones is correct.

monai/transforms/utils.py
ericspod: I don't see the updated description for C and N, if it's here just marked this as resolved.

KumoLiu: Sorry for missing this one, updated in https://github.com/Project-MONAI/MONAI/pull/7942/commits/ffab3900ca31b08a0db29036fa5d50e872d1e2bf

KumoLiu:
Hi @vikashg @atbenmurray @aylward @mingxin-zheng @Nic-Ma, do you have any further comments? Otherwise I will try to merge this one. Thanks.

vikashg:
Hey Yun, 
Give me couple of hours I am going over it right now. ok

Thanks

vikashg:
Hey @KumoLiu, 
I am trying to run this new changes. I get the following error
```
  File "/workspace/app/MONAI/monai/transforms/utility/temp_test2.py", line 4, in <module>
    from monai.transforms import LoadImage
  File "/workspace/app/MONAI/monai/__init__.py", line 61, in <module>
    load_submodules(sys.modules[__name__], True, exclude_pattern=excludes)
  File "/workspace/app/MONAI/monai/utils/module.py", line 211, in load_submodules
    mod = import_module(name)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/workspace/app/MONAI/monai/transforms/utility/temp_test.py", line 86, in <module>
    world_points_transformed_axial = image_to_world_transform(data)["coordinates_image_axial"]
  File "/workspace/app/MONAI/monai/transforms/utility/dictionary.py", line 1795, in __call__
  File "/workspace/app/MONAI/monai/transforms/transform.py", line 475, in key_iterator
    raise KeyError(
KeyError: 'Key `point` of transform `ApplyTransformToPointsd` was missing in the data and allow_missing_keys==False.'
```
I think its something small. if you can have a look while I try to setup a test on my end.

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1716,124 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+            Transformed coordinates.
+        """
+        data = convert_to_tensor(data, track_meta=get_track_meta())
+        applied_affine = getattr(data, "affine", None)
+
+        if affine is None and self.invert_affine:
+            raise ValueError("affine must be provided when invert_affine is True.")
+
+        affine = applied_affine if affine is None else affine
+        affine = convert_data_type(affine, dtype=torch.float64)[0]  # always convert to float64 for affine
```
mingxin-zheng, line 1782: Hi @KumoLiu , may I know what the difference between `affine` and `applied_affine` is? When I am reading this part, I got lost about what I am looking at. May some in-line comments will be helpful for readability.

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1716,124 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+        if self.affine_lps_to_ras:
+            affine = orientation_ras_lps(affine)
+
+        _affine: torch.Tensor = affine
+        if self.invert_affine:
+            _affine = linalg_inv(affine)
+            # consider the affine transformation already applied to the data in the world space
+            if applied_affine is not None:
+                _affine = _affine @ linalg_inv(applied_affine)
+        out = apply_affine_to_points(data, _affine, self.dtype)
```
mingxin-zheng, line 1793:
```suggestion
        out = apply_affine_to_points(data, _affine, dtype=self.dtype)
```

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1716,124 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+
+        extra_info = {
+            "invert_affine": self.invert_affine,
+            "dtype": get_dtype_string(self.dtype),
+            "image_affine": original_affine,
+            "affine_lps_to_ras": self.affine_lps_to_ras,
+        }
+        xform: torch.Tensor = original_affine if self.invert_affine else linalg_inv(original_affine)
```
mingxin-zheng, line 1801: Similarly to my comment above, we used `affine` above, and the introduction of `xform` confused me a bit about the difference between it and `affine`

monai/transforms/utility/dictionary.py
```diff
@@ -1740,6 +1741,69 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> dict[Hashable, N
...
+    The input coordinates are assumed to be in the shape (C, N, 2 or 3),
+    where C represents the number of channels and N denotes the number of points.
+
+    Args:
+        keys: keys of the corresponding items to be transformed.
+            See also: monai.transforms.MapTransform
+        refer_key: The key of the reference item used for transformation.
+            It can directly refer to an affine or an image from which the affine can be derived.
+        dtype: The desired data type for the output.
+        affine: A 3x3 or 4x4 affine transformation matrix applied to points. This matrix typically originates
```
mingxin-zheng, line 1756: In `ApplyTransformToPoints`, the `affine` parameter is optional, but it's not clear what happens when it's `None`. It would be good to add a docstring explaining the behavior in this case.

monai/transforms/utility/dictionary.py
```diff
@@ -1740,6 +1741,69 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> dict[Hashable, N
...
+    ):
+        MapTransform.__init__(self, keys, allow_missing_keys)
+        self.refer_key = refer_key
+        self.converter = ApplyTransformToPoints(
+            dtype=dtype, affine=affine, invert_affine=invert_affine, affine_lps_to_ras=affine_lps_to_ras
+        )
+
+    def __call__(self, data: Mapping[Hashable, torch.Tensor]):
+        d = dict(data)
+        refer_data = d[self.refer_key] if self.refer_key is not None else None
```
mingxin-zheng, line 1793: It would be good to add a check to ensure that the `refer_key` is a valid key in the input dictionary.

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1716,124 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
         if self._do_transform:
             img = self.filter(img)
         return img
+
+
+class ApplyTransformToPoints(InvertibleTransform, Transform):
```
mingxin-zheng, line 1721: The `ApplyTransformToPoints` class could benefit from additional description of the transformation being applied and the expected input and output formats.

monai/transforms/utility/dictionary.py
vikashg: If the parameter is `None`, does it not use the affine matrix from the image ? I believe that should be expected behavior, right ?

vikashg:
Hey @KumoLiu, 
I have created this [gist](https://gist.github.com/vikashg/8cb1d79c8c56f6425acd73a4f76988ba) to test the workings of the code. I wanted to test how an inplane rotation is working. Please have a look at the code to ensure that I am using all the options in `ApplyTransformToPointsd` correctly.
This was the original image
test
After rotation I get the following image
test_rot

Also, please add `ApplyTransformToPointsd` and `ApplyTransformToPoints` in the `/monai/transforms/__init__.py` file.
I think we are almost there, just a few things as @mingxin-zheng also mentioned about the documentation needs to be corrected/modified. 
Thanks so much for working on this.

monai/transforms/utility/dictionary.py
```diff
@@ -1740,6 +1741,69 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> dict[Hashable, N
...
+        affine: A 3x3 or 4x4 affine transformation matrix applied to points. This matrix typically originates
+            from the image. For 2D points, a 3x3 matrix can be provided, avoiding the need to add an unnecessary
+            Z dimension. While a 4x4 matrix is required for 3D transformations, it's important to note that when
+            applying a 4x4 matrix to 2D points, the additional dimensions are handled accordingly.
+            The matrix is always converted to float64 for computation, which can be computationally
+            expensive when applied to a large number of points.
+        invert_affine: Whether to invert the affine transformation matrix applied to the points. Defaults to ``True``.
+            Typically, the affine matrix is derived from the image, while the points are in world coordinates.
+            If you want to align the points with the image, set this to ``True``. Otherwise, set it to ``False``.
+        affine_lps_to_ras: Defaults to ``False``. Set this to ``True`` if all of the following are true:
```
vikashg, line 1765: Can you please make a note here that we need to check this point a bit more. It makes sense that `affine_lps_to_ras` is `True` when **all** the statements are true. But MONAI also used readers other than ITK. So what happens when we use nibabel as the image reader. 
Or maybe add few more guidelines as in what happens when only a few conditions are met. 

@ericspod Just to confirm that the world coordinate system is RAS but nifti has an LPS coordinate system right?

monai/transforms/utility/array.py
KumoLiu: Hi @mingxin-zheng, yes I mentioned here: https://github.com/Project-MONAI/MONAI/blob/ffab3900ca31b08a0db29036fa5d50e872d1e2bf/monai/transforms/utility/array.py#L1790
The applied_affine can also be recorded in the point itself, if users want to convert points back to the world space, it will have affine info in it's meta.

monai/transforms/utility/dictionary.py
KumoLiu: It will only be True when all the statements are true so all other case will be False here.

monai/transforms/utility/array.py
KumoLiu: Add more descriptions in the docstring.

monai/transforms/utility/dictionary.py
KumoLiu: Thanks, add the check.

monai/transforms/utility/dictionary.py
KumoLiu: Yes, you are right @vikashg. Add more description.

KumoLiu:
Hi @mingxin-zheng @vikashg, add the use case description in the docstring. please help take a look whether it makes clearer to you, thanks!
https://github.com/Project-MONAI/MONAI/blob/48707eded0f00a7481ca314ac78597b84368c89e/monai/transforms/utility/array.py#L1748

mingxin-zheng:
> Hi @mingxin-zheng @vikashg, add the use case description in the docstring. please help take a look whether it makes clearer to you, thanks!
> 
> https://github.com/Project-MONAI/MONAI/blob/48707eded0f00a7481ca314ac78597b84368c89e/monai/transforms/utility/array.py#L1748

Thanks and it looks good to me!

mingxin-zheng (approved):

monai/transforms/utility/array.py
mingxin-zheng: Closing this as we had already discussed this in meetings. Please feel free to re-open if there are still questions.

monai/transforms/utility/array.py
mingxin-zheng: Closing this as we already discussed this in meetings. Please feel free to re-open it if there are questions @atbenmurray

monai/transforms/utility/array.py
mingxin-zheng: Closing this as we had already discussed it in meetings. I think @KumoLiu has already addressed this. Please feel free to re-open it if there are questions @aylward

monai/transforms/utility/dictionary.py
ericspod: We're assuming world coordinates is RAS which I believe matches the scanner space DICOM expects, Nifti are generally in RAS as well and is the way [Nibabel produces outputs](https://nipy.org/nibabel/coordinate_systems.html#nibabel-always-uses-an-ras-output-space).

vikashg: Thanks @ericspod

KumoLiu:
Hi @vikashg, I take a look at your gist. The issue is that you need transform the point in the world space but you are transform the data already in the image space.
you can get world space point using:
```
coordinates_world_axial = np.array([i['position'] for i in data_axial['markups'][0]['controlPoints']])
```

vikashg:
Thanks @KumoLiu. That solves it.

monai/transforms/utility/array.py
```diff
@@ -1715,3 +1717,135 @@ def __call__(self, img: NdarrayOrTensor, meta_dict: Mapping | None = None) -> Nd
...
+            while the points are in world coordinates. A value of ``True`` represents transforming these
+            world space coordinates to the image's coordinate space, and ``False`` the inverse of this operation.
+        affine_lps_to_ras: Defaults to ``False``. Set this to ``True`` if all of the following are true:
+            1) The image is read by `ITKReader`,
+            2) The `ITKReader` has `affine_lps_to_ras=True`,
+            3) The data is in world coordinates.
+            This ensures the correct application of the affine transformation between LPS (left-posterior-superior)
+            and RAS (right-anterior-superior) coordinate systems. This argument ensures the points and the affine
+            matrix are in the same coordinate system.
+
```
vikashg, line 1748: Hey @KumoLiu 
So I tested this aspect of the code. Here is what I did. Instead of using `ITKReader` I used `NibabelReader` for loading the images. As I used a different reader I changed the `apply_lps_to_ras` to `False` as indicated in the docstring. 
I get the following result as transformed points 
```
metatensor([[[1001.9410,  -36.8047,   78.0000],
         [1028.0130,  -32.9858,   78.0000],
         [1059.5454,  -50.9436,   78.0000],
         [1087.0344,  -61.4492,   78.0000],
         [1063.8782, -132.3439,   78.0000],
         [ 996.8077, -168.7347,   78.0000],
         [1011.6713, -141.4345,   78.0000],
         [1032.8000, -110.6216,   78.0000],
         [1020.5166,  -78.3613,   78.0000],
         [ 999.7059,  -41.3652,   78.0000]]], dtype=torch.float64)
```
You can try the same by changing my gist. However, if I change the `apply_lps_to_ras` as `True` with `NibabelReader`.

KumoLiu: > Hey @KumoLiu So I tested this aspect of the code. Here is what I did. Instead of using `ITKReader` I used `NibabelReader` for loading the images. As I used a different reader I changed the `apply_lps_to_ras` to `False` as indicated in the docstring.

Hi @vikashg, thanks for pointing this out. Let me clarify, this argument should set to true when the input point data is annotated in the RAS coordinate. I just updated the docstring, please help take a look again to see whether it's clearer now.

vikashg (approved):
Thanks @KumoLiu Really appreciate all the hard work you have put on this. Looks great to me.

PR #7969, opened 2024-07-31
Forced `Fourier` class to output contiguous tensors.

bwittmann:
Forced `Fourier` class to output contiguous tensors, which potentially fixes a performance bottleneck.

### Description

Some transforms, such as `RandKSpaceSpikeNoise`, rely on the `Fourier` class.
In its current state, the `Fourier` class returns non-contiguous tensors, which potentially limits performance.
For example, when followed by `RandHistogramShift`, the following warning occurs:

```
<path_to_monai>/monai/transforms/intensity/array.py:1852: UserWarning: torch.searchsorted(): input value tensor is non-contiguous, this will lower the performance due to extra data copy when converting non-contiguous tensor to contiguous, please use contiguous input value tensor if possible. This message will only appear once per program. (Triggered internally at /opt/conda/conda-bld/pytorch_1716905975447/work/aten/src/ATen/native/BucketizationUtils.h:32.)
  indices = ns.searchsorted(xp.reshape(-1), x.reshape(-1)) - 1
```

A straightforward fix is to force the `Fourier` class to output contiguous tensors (see commit).
To reproduce, please run:

```
from monai.transforms import RandKSpaceSpikeNoise
from monai.transforms.utils import Fourier
import numpy as np

### TEST WITH TRANSFORMS ###   
t = RandKSpaceSpikeNoise(prob=1)

# for torch tensors
a_torch = torch.rand(1, 128, 128, 128)
print(a_torch.is_contiguous())

a_torch_mod = t(a_torch)
print(a_torch_mod.is_contiguous())

# for np arrays
a_np = np.random.rand(1, 128, 128, 128)
print(a_np.flags['C_CONTIGUOUS'])

a_np_mod = t(a_np)  # automatically transformed to torch.tensor
print(a_np_mod.is_contiguous())

### TEST DIRECTLY WITH FOURIER ###
f = Fourier()

# inv_shift_fourier
# for torch tensors
real_torch = torch.randn(1, 128, 128, 128)
im_torch = torch.randn(1, 128, 128, 128)
k_torch = torch.complex(real_torch, im_torch)
print(k_torch.is_contiguous())

out_torch = f.inv_shift_fourier(k_torch, spatial_dims=3)
print(out_torch.is_contiguous())

# for np arrays
real_np = np.random.randn(1, 100, 100, 100)
im_np = np.random.randn(1, 100, 100, 100)
k_np = real_np + 1j * im_np
print(k_np.flags['C_CONTIGUOUS'])

out_np = f.inv_shift_fourier(k_np, spatial_dims=3)
print(out_np.flags['C_CONTIGUOUS'])

# shift_fourier
# for torch tensors
a_torch = torch.rand(1, 128, 128, 128)
print(a_torch.is_contiguous())

out_torch = f.shift_fourier(a_torch, spatial_dims=3)
print(out_torch.is_contiguous())

# for np arrays
a_np = np.random.rand(1, 128, 128, 128)
print(a_np.flags['C_CONTIGUOUS'])

out_np = f.shift_fourier(a_np, spatial_dims=3)
print(out_np.flags['C_CONTIGUOUS'])
```

### Types of changes


monai/transforms/utils.py
```diff
@@ -1649,12 +1649,12 @@ def shift_fourier(x: NdarrayOrTensor, spatial_dims: int) -> NdarrayOrTensor:
         k: NdarrayOrTensor
         if isinstance(x, torch.Tensor):
             if hasattr(torch.fft, "fftshift"):  # `fftshift` is new in torch 1.8.0
-                k = torch.fft.fftshift(torch.fft.fftn(x, dim=dims), dim=dims)
+                k = torch.fft.fftshift(torch.fft.fftn(x, dim=dims), dim=dims).contiguous()
```
KumoLiu, line 1652: Calling contiguous() results in a new tensor, meaning memory copying. For large tensors, this can increase memory consumption and computational overhead.
If contiguous() is called frequently, especially on large datasets, it can impact performance. Therefore, use this method only when necessary.

I would prefer if we can add an arg here `as_contiguous` such as what we did here: https://github.com/Project-MONAI/MONAI/blob/f1ef3e88f4bc594779c0c4188883f78bf6d1efff/monai/data/grid_dataset.py#L214

And also, we can use `ascontiguousarray` to only change one line during return.
https://github.com/Project-MONAI/MONAI/blob/f1ef3e88f4bc594779c0c4188883f78bf6d1efff/monai/transforms/utils_pytorch_numpy_unification.py#L398

What do you think?
cc @Nic-Ma @ericspod

ericspod: I think this makes sense. It's not often clear when an array/tensor needs to be contiguous, so it's likely an experimental thing to figure out where needed. An argument would let people do that.

KumoLiu:
Hi @bwittmann, do you plan to finish this PR? If we want to include the change in version 1.4, we only have about one or two weeks left to add new things.

bwittmann:
Dear @KumoLiu,

following your recommendation, I added `as_contiguous` as an argument to propose an experimental option.

Best, 
Bastian

KumoLiu (approved):
Thanks for the update, LGTM.

KumoLiu:
Hi @bwittmann, please help fix the DCO and format issue. Thanks.

KumoLiu:
You can refer to the instruction here to fix the DCO issue.
https://github.com/Project-MONAI/MONAI/pull/7969/checks?check_run_id=29602735375

PR #7970, opened 2024-07-31
Fix generalized dice computation

### LINKED ISSUE #7966: `GeneralizedDiceScore`  reductions do not work as expected
**Describe the bug**
`compute_generalized_dice` does not return values as highlighted in the docs. 

```python
import torch
from monai.metrics import compute_generalized_dice, GeneralizedDiceScore

a = torch.ones((20, 10, 64, 128, 128))
b = torch.ones((20, 10, 64, 128, 128))


compute_generalized_dice(a, b).shape
```
This returns a shape of `torch.Size([20])` whereas per the documentation it should return `torch.Size([20, 10])`
https://docs.monai.io/en/stable/metrics.html#generalized-dice-score

This leads to problems that propagate to `GeneralizedDiceScore` as below,
```python
generalized_dice_score = GeneralizedDiceScore()
generalized_dice_score(a, b)
```
Aggregating this over different reductions shows unexpected results and errors.
Case 1: `generalized_dice_score.aggregate(reduction="sum")` and `generalized_dice_score.aggregate(reduction="mean")` return this error `IndexError: Dimension out of range (expected to be in range of [-1, 0], but got 1)`

Case 2: `generalized_dice_score.aggregate(reduction="sum_batch")` returns a single-element tensor but it should return a tensor with the size as number of classes containing dice scores summed across all batches for each class. Similar for mean_batch.

Case 3: `generalized_dice_score.aggregate(reduction="none")` should give dice scores across batch and classes unreduced (similar to `compute_generalized_dice`) but it gives values reduced over all classes. 

**Environment**

```
Printing MONAI config...
MONAI version: 1.3.2
Numpy version: 1.24.4
Pytorch version: 2.4.0+cu121
MONAI rev id: 59a7211070538586369afd4a01eca0a7fe2e742e
MONAI __file__: /home/<username>/miniconda3/lib/python3.10/site-packages/monai/__init__.py
```

**Suggested Solution**

It looks like removing the `.sum()` in the following code should give expected behaviour
https://github.com/Project-MONAI/MONAI/blob/59a7211070538586369afd4a01eca0a7fe2e742e/monai/metrics/generalized_dice.py#L162-L163

https://github.com/Project-MONAI/MONAI/blob/59a7211070538586369afd4a01eca0a7fe2e742e/monai/metrics/generalized_dice.py#L170

Tried it out and the shapes and values look as expected.

surajpaib:
Fixes #7966 

### Description

### Types of changes


monai/metrics/generalized_dice.py
```diff
@@ -20,109 +20,108 @@
...
         return f
 
 
 def compute_generalized_dice(
-    y_pred: torch.Tensor, y: torch.Tensor, include_background: bool = True, weight_type: Weight | str = Weight.SQUARE
+    y_pred: torch.Tensor,
+    y: torch.Tensor,
+    include_background: bool = True,
+    weight_type: Weight | str = Weight.SQUARE,
+    sum_over_labels: bool = False,
```
KumoLiu, line 104: I would suggest we set this arg as `True` by default to be consistent with previous behaviour.

surajpaib: @KumoLiu 

After looking a bit more at the `meandice` implementation as well, its probably best to make both consistent. 


```python
import torch
from monai.metrics import compute_generalized_dice, compute_dice

a = torch.randn((20, 10, 64, 128, 128))
b = torch.ones((20, 10, 64, 128, 128))



compute_generalized_dice(a, b).shape, compute_dice(a, b).shape
```

The current output looks like so:
```
(torch.Size([20]), torch.Size([20, 10]))
```

KumoLiu: @surajpaib, Thanks for the details, agreed, please just go ahead.

monai/metrics/generalized_dice.py
```diff
@@ -20,109 +20,108 @@
...
         return compute_generalized_dice(
-            y_pred=y_pred, y=y, include_background=self.include_background, weight_type=self.weight_type
+            y_pred=y_pred,
+            y=y,
+            include_background=self.include_background,
+            weight_type=self.weight_type,
+            sum_over_labels=self.sum_over_labels,
         )
 
-    def aggregate(self, reduction: MetricReduction | str | None = None) -> torch.Tensor:
```
KumoLiu, line 81: Is there any specific reason remove the reduction here? It seems that the keeping this argument maintains the same logic as other metrics.

monai/metrics/generalized_dice.py
```diff
@@ -20,109 +20,108 @@
...
             remaining are the spatial dimensions.
-        y (torch.Tensor): binarized ground-truth. It should be binarized, in one-hot format and have the same shape as `y_pred`.
-        include_background (bool, optional): whether to include score computation on the first channel of the
+        y: Binarized ground-truth. It should be binarized, in one-hot format and have the same shape as `y_pred`.
+        include_background: Whether to include score computation on the first channel of the
             predicted output. Defaults to True.
-        weight_type (Union[Weight, str], optional): {``"square"``, ``"simple"``, ``"uniform"``}. Type of function to
+        weight_type: {``"square"``, ``"simple"``, ``"uniform"``}. Type of function to
             transform ground truth volume into a weight factor. Defaults to ``"square"``.
+        sum_over_labels: Whether to sum the numerator and denominator across all labels before the final computation.
```
KumoLiu, line 118: Do you thinks sum_over_classes would be better? Just a suggestion, open for this one.
```suggestion
        sum_over_labels: Whether to sum the numerator and denominator across all classes before the final computation.
```

KumoLiu (commented):
Thanks for the contribution, overall looks good to me.
Once you finished the PR, please let me know, I will add more people help review it, thanks.

monai/metrics/generalized_dice.py
```diff
@@ -20,109 +20,108 @@
...
-        ValueError: when the `weight_type` is not one of {``"none"``, ``"mean"``, ``"sum"``}.
+        ValueError: When the `reduction` is not one of MetricReduction enum.
     """
 
     def __init__(
-        self,
-        include_background: bool = True,
-        reduction: MetricReduction | str = MetricReduction.MEAN_BATCH,
-        weight_type: Weight | str = Weight.SQUARE,
+        self, include_background: bool = True, reduction: str = "mean", weight_type: Weight | str = Weight.SQUARE
```
ericspod, line 48: The `reduction` default value should not be a string literal (this is like a magic number essentially) and should use the value from the `MetricReduction` enum. The original default was also mean_batch and not mean, this shouldn't be changed or if it must there should be a deprecated warning using `deprecated_arg_default`.

surajpaib: I'll make the changes for the enum. 

For the default reduction, the current value specified is `mean_batch` but the calling aggregate actually returns the `mean` reduction. 

This is because the `compute_generalized_dice` function reduces over classes already so applying a mean batch after it results in an overall mean.

monai/metrics/generalized_dice.py
surajpaib: Agreed! Will make the change

monai/metrics/generalized_dice.py
surajpaib: I've removed the reduction here because the value of `sum_over_labels` needs to be set during the `_compute_tensor` and the tensor will already be reduced over labels at that point. 

So I'm not sure how we can handle a different reduction passed to `aggregate` right now - unless we store an intermediary tensor for the `intersection` and `denominator` in the buffer and make the computation in aggregate. 

Do you have any suggestions here?

surajpaib: I agree that the inconsistency might be an issue. We could offer lesser options for reduction as a solution - providing only `mean_batch` and `mean` reductions. 

This would mean you can't compute GeneralizedDice per class but that is probably fine as per the original formulation in the paper which reduces over classes by default.

surajpaib: Was originally inspired by this implementation: https://lightning.ai/docs/torchmetrics/stable/segmentation/generalized_dice.html

which does provide values per class but the `per_class` flag is defined during init

monai/metrics/generalized_dice.py
KumoLiu: Like @ericspod suggested, it would also be better to add a `deprecated_arg_default`, and mentioned that the default result of `mean_batch` before is not correct, if user want to achieve before results, they should use `mean`. Does this make sense?

KumoLiu:
Hi @surajpaib, do you plan to address the comments on the PR? I’m hoping we can get this merged into version 1.4, but we only have about one or two weeks left to add new features.

surajpaib:
Hi @KumoLiu, sorry for the gap, was away on vacation. I think I can work on this in the next few days and get it in.

monai/metrics/generalized_dice.py
surajpaib: @KumoLiu Do you think we can follow the `deprecated_arg_default` convention here as well for the reduction on aggregate?

KumoLiu: Sure, we can do it here. Just ensure we can get the exactly same result with before and let users aware if any behaviour will be changed.

surajpaib:
@KumoLiu I've incorporated the comments we discussed. 

`deprecated_arg_default` is added to the initialization and for the reduction, I've added `deprecated_arg`. 

The tests seem to be failing because of the version string - I assume we can set this appropriately during the release. 

Let me know if you see any other changes

monai/metrics/generalized_dice.py
```diff
@@ -14,115 +14,131 @@
...
+        ValueError: When the `reduction` is not one of MetricReduction enum.
     """
 
+    @deprecated_arg_default(
+        "reduction",
+        old_default=MetricReduction.MEAN_BATCH,
+        new_default=MetricReduction.MEAN,
+        since="1.3.3",
+        replaced="1.3.3",
+        msg_suffix="Old versions computed `mean` when `mean_batch` was provided due to bug in reduction",
```
KumoLiu, line 53: Instead of immediately updating the default behavior, we need to provide users with some time to adapt to this change. We should also issue a warning to inform users that the mean_batch behavior is no longer the same as before, since it previously calculated the mean. If users want to retain the previous behavior, they will need to explicitly set the parameter to "mean".
```suggestion
        since="1.4.0",
        replaced="1.5.0",
        msg_suffix="Old versions computed `mean` when `mean_batch` was provided due to bug in reduction， If you want to retain the old behavior (calculating the mean), please explicitly set the parameter to 'mean'.",
```

cc @ericspod

monai/metrics/generalized_dice.py
```diff
@@ -14,115 +14,131 @@
...
+        new_default=MetricReduction.MEAN,
+        since="1.4.0",
+        replaced="1.5.0",
+        msg_suffix="Old versions computed `mean` when `mean_batch` was provided due to bug in reduction， If you want to retain the old behavior (calculating the mean), please explicitly set the parameter to 'mean'.",
+    )
     def __init__(
         self,
         include_background: bool = True,
-        reduction: MetricReduction | str = MetricReduction.MEAN_BATCH,
+        reduction: MetricReduction | str = MetricReduction.MEAN,
```
KumoLiu, line 58:
```suggestion
        reduction: MetricReduction | str = MetricReduction.MEAN_BATCH,
```

monai/metrics/generalized_dice.py
KumoLiu: Hi @surajpaib, I may misunderstand before, seems we don't need to deprecated the reduction in aggregation. sum_over_classes will only sum over class, then if users also want calculate batch, what should they specify?

surajpaib: The `sum_over_classes` is only so we can enable computing gen.dice considering all classes in the numerator and denominator - the default setting for gen dice as in the original paper. 

So if the user wants to reduce `mean_batch` then the `sum_over_classes` will be set to False. But if they want to compute `mean_channel` or overall `mean`, this should be set to True as it will consider all classes in the computation. 

P.S: I realize that I need to add `mean_channel` for setting the `sum_over_classes` flag as well. Will do so asap

monai/metrics/generalized_dice.py
```diff
@@ -50,79 +60,85 @@ def __init__(
...
-            MetricReduction.NONE,
-            MetricReduction.MEAN_BATCH,
-            MetricReduction.SUM_BATCH,
-        ]
-        self.reduction = reduction
-        if self.reduction not in reduction_options:
-            raise ValueError(f"reduction must be one of {reduction_options}")
+        self.reduction = look_up_option(reduction, MetricReduction)
         self.weight_type = look_up_option(weight_type, Weight)
+        self.sum_over_classes = self.reduction == MetricReduction.SUM or self.reduction == MetricReduction.MEAN
```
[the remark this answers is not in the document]
surajpaib: @KumoLiu Just added it! Missed it on the previous, added a `mean_channel` computation in the test as well

tests/test_compute_generalized_dice.py
```diff
@@ -22,7 +22,7 @@
 _device = "cuda:0" if torch.cuda.is_available() else "cpu"
 
 # keep background
-TEST_CASE_1 = [  # y (1, 1, 2, 2), y_pred (1, 1, 2, 2), expected out (1)
```
KumoLiu, line 25: Please also update the shape for Line 31, [0.8] --> [[0.8]]

KumoLiu (approved):
Thanks for the update, LGTM.
Could you please help fix the ci issue and dco issue then I can help trigger more tests and merge this one?
https://github.com/Project-MONAI/MONAI/actions/runs/10733660015/job/29767399315?pr=7970#step:7:428

KumoLiu:
Hi @surajpaib, you can fix the dco issue by following the method here: https://github.com/Project-MONAI/MONAI/pull/7970/checks?check_run_id=29787940944
Also please update here: https://github.com/Project-MONAI/MONAI/pull/7970/files#r1746799238
Then I will try to merge this one, thanks!

PR #7974, opened 2024-08-01
7973-add-ngc-prefix

### LINKED ISSUE #7973: Enhance `_download_from_ngc` to support no-prefix or different prefix data
So far, it's alway use the default prefix:
```
def _download_from_ngc(
    download_path: Path, filename: str, version: str, remove_prefix: str | None, progress: bool
) -> None:
    # ensure prefix is contained
    filename = _add_ngc_prefix(filename)
...
```
We can make use of `_add_ngc_prefix`'s `prefix` arg

yiheng-wang-nv:
Fixes #7973 .

### Description

### Types of changes

### DIFF
--- a/monai/bundle/scripts.py
+++ b/monai/bundle/scripts.py
@@ -217,10 +217,15 @@ def _remove_ngc_prefix(name: str, prefix: str = "monai_") -> str:
 
 
 def _download_from_ngc(
-    download_path: Path, filename: str, version: str, remove_prefix: str | None, progress: bool
+    download_path: Path,
+    filename: str,
+    version: str,
+    prefix: str = "monai_",
+    remove_prefix: str | None = "monai_",
+    progress: bool = True,
 ) -> None:
     # ensure prefix is contained
-    filename = _add_ngc_prefix(filename)
+    filename = _add_ngc_prefix(filename, prefix=prefix)
     url = _get_ngc_bundle_url(model_name=filename, version=version)
     filepath = download_path / f"{filename}_v{version}.zip"
     if remove_prefix:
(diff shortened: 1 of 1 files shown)


monai/bundle/scripts.py
```diff
@@ -231,10 +236,16 @@ def _download_from_ngc(
 
 
 def _download_from_ngc_private(
-    download_path: Path, filename: str, version: str, remove_prefix: str | None, repo: str, headers: dict | None = None
+    download_path: Path,
+    filename: str,
+    version: str,
+    repo: str,
+    prefix: str = "monai_",
+    remove_prefix: str | None = "monai_",
```
KumoLiu, line 244: Look like the change of the default value here will make the behavior not consistent with previous.
If no special reason, I would suggest revert it back.
cc @mingxin-zheng

yiheng-wang-nv: no, the default is always monai_. Please check: https://github.com/Project-MONAI/MONAI/blob/23ade147fd9a3c32d754b28c2ae4c365b924b04a/monai/bundle/scripts.py#L440

KumoLiu: My bad, closed as resolved. Thanks.

monai/bundle/scripts.py
```diff
@@ -217,10 +217,15 @@ def _remove_ngc_prefix(name: str, prefix: str = "monai_") -> str:
 
 
 def _download_from_ngc(
-    download_path: Path, filename: str, version: str, remove_prefix: str | None, progress: bool
+    download_path: Path,
+    filename: str,
+    version: str,
+    prefix: str = "monai_",
```
KumoLiu, line 223: Do we also need add this arg in the download and pass it to this one?

KumoLiu: Discussed this offline with @yiheng-wang-nv 
Will leave this one for this PR. If we have a case need this arg in the download and load, will enhance them later.

Nic-Ma (approved):

PR #7977, opened 2024-08-01
Flash attention

### LINKED ISSUE #7944: Consideration of Flash Attention in Generative Components
Regarding the comments here: https://github.com/Project-MONAI/MONAI/pull/7715#issuecomment-2088149866

We have removed the flash attention from the generative components merged into the core. However, based on experiments conducted by @dongyang0122, there appears to be a significant difference between using and not using flash attention. We should consider adding this option back. @dongyang0122 will share more detailed comparison results from the experiments.

virginiafdez:
Fixes #7944.

### Description

In response to Issue https://github.com/Project-MONAI/MONAI/issues/7944, I added the new functionality scaled_dot_product_attention from PyTorch to re-enable flash attention, present in the original MONAI Generative Models repository. This is allowed for torch >= 2.0 and when argument save_attn = False. Errors are raised otherwise. I ran quick tests and added some checks on test_selfattention and test_crossattention scripts to make sure the outputs are the same as not using flash attention. 

### Types of changes


monai/networks/blocks/selfattention.py
```diff
@@ -130,23 +145,28 @@ def forward(self, x):
...
+        if self.use_flash_attention:
+            x = F.scaled_dot_product_attention(q, k, v)
+        else:
+            att_mat = torch.einsum("blxd,blyd->blxy", q, k) * self.scale
 
-        # apply relative positional embedding if defined
-        att_mat = self.rel_positional_embedding(x, att_mat, q) if self.rel_positional_embedding is not None else att_mat
+            # apply relative positional embedding if defined
+            att_mat = (
+                self.rel_positional_embedding(x, att_mat, q) if self.rel_positional_embedding is not None else att_mat
```
ericspod, line 155:
```suggestion
            if self.rel_positional_embedding is not None:
                att_mat = self.rel_positional_embedding(x, att_mat, q)
```

monai/networks/blocks/crossattention.py
```diff
@@ -62,6 +63,7 @@ def __init__(
             input_size (tuple(spatial_dim), optional): Input resolution for calculating the relative
                 positional parameter size.
             attention_dtype: cast attention operations to this dtype.
+            use_flash_attention: if True, use flash attention for a memory efficient attention mechanism.
```
ericspod, line 66: I think we should explicitly state here and elsewhere that we're using Pytorch's inbuilt flash attention implementation.

ericspod:
A few minor comments but looks good otherwise. I would like others to test this change and see if memory performance improves. Thanks!

tests/test_selfattention.py
```diff
@@ -49,11 +50,17 @@ class TestResBlock(unittest.TestCase):
 
     @parameterized.expand(TEST_CASE_SABLOCK)
     @skipUnless(has_einops, "Requires einops")
+    @SkipIfBeforePyTorchVersion((0, 2))
```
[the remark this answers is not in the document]
virginiafdez: I also changed the cross_attention one, it was wrong.

monai/networks/blocks/crossattention.py
KumoLiu: Agreed, would be great if we could also mention the link here: https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html

monai/networks/blocks/crossattention.py
```diff
@@ -145,23 +159,29 @@ def forward(self, x: torch.Tensor, context: Optional[torch.Tensor] = None):
...
-        # apply relative positional embedding if defined
-        att_mat = self.rel_positional_embedding(x, att_mat, q) if self.rel_positional_embedding is not None else att_mat
+        if self.use_flash_attention:
+            x = torch.nn.functional.scaled_dot_product_attention(q, k, v).contiguous()
+        else:
+
+            att_mat = torch.einsum("blxd,blyd->blxy", q, k) * self.scale
+            # apply relative positional embedding if defined
+            att_mat = (
+                self.rel_positional_embedding(x, att_mat, q) if self.rel_positional_embedding is not None else att_mat
```
KumoLiu, line 170: Looks we can only add `rel_pos_embedding` when not using flash attention. I'm not sure whether it's same with the version in the generative repo, I didn't find the reference.
If that's the case, we may also need add a warning if `self.rel_positional_embedding is not None` and `use_flash_attention=True`.

monai/networks/blocks/selfattention.py
```diff
@@ -130,23 +145,28 @@ def forward(self, x):
             q = q.to(self.attention_dtype)
             k = k.to(self.attention_dtype)
 
-        att_mat = torch.einsum("blxd,blyd->blxy", q, k) * self.scale
+        if self.use_flash_attention:
+            x = F.scaled_dot_product_attention(q, k, v)
```
KumoLiu, line 149: We may also inlcude these args here to be consistent with before.
https://github.com/Project-MONAI/GenerativeModels/blob/7428fce193771e9564f29b91d29e523dd1b6b4cd/generative/networks/blocks/selfattention.py#L121-L123

monai/networks/blocks/transformerblock.py
```diff
@@ -36,15 +36,15 @@ def __init__(
...
     ) -> None:
         """
-        Args:
-            hidden_size (int): dimension of hidden layer.
-            mlp_dim (int): dimension of feedforward layer.
-            num_heads (int): number of attention heads.
-            dropout_rate (float, optional): fraction of the input units to drop. Defaults to 0.0.
-            qkv_bias (bool, optional): apply bias term for the qkv linear layer. Defaults to False.
-            save_attn (bool, optional): to make accessible the attention matrix. Defaults to False.
+        Args: hidden_size (int): dimension of hidden layer. mlp_dim (int): dimension of feedforward layer. num_heads
```
KumoLiu, line 42: Looks the format here has been changed by mistake.

monai/networks/blocks/crossattention.py
virginiafdez: Is it better to have a warning on an error, as we've done with save_attn?

KumoLiu: error would be better.

monai/networks/blocks/selfattention.py
virginiafdez: Scale and dropout are fine. I am a bit confused when it comes to causal and attn_bias. attn_bias is a mask, which is set to xops.LowerTriangularMask()  in the old implementation. Nonetheless, here you cannot set it up if causal is True. It's automatically derived. It seems that if causal is True, we can just pass "is_causal = True" and the function will derive the mask. However, when causal is not True, we can pass a mask, but in the init method, self.causal_mask is only set up when causal is True. I think to be safe, I'd just pass the causal value and not any mask.

virginiafdez: If you use causal = True, and you run with and without flash_attention, you do not get a similar output

virginiafdez: Also, the test test_causal cannot be reproduced for flash_attention because save_attn has to be False. There's no way the attention matrix can be derived and therefore, we cannot verify that it's masking things correctly.

virginiafdez: I also had to add self.dropout_rate as an attribute, as it's necessary for the call to the function.

KumoLiu: For the casual one, we can use `is_casual` in the torch: https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html

```
import torch
import xformers.ops as xops
from copy import deepcopy

torch.manual_seed(0)

query = torch.rand(8, 2560, 1, 32).to("cuda")
key = torch.rand(8, 2560, 1, 32).to("cuda")
value = torch.rand(8, 2560, 1, 32).to("cuda")

x = xops.memory_efficient_attention(query, key, value, attn_bias=xops.LowerTriangularMask())
print(x.squeeze()[0,0,...])
y = F.scaled_dot_product_attention(query.transpose(1, 2), key.transpose(1, 2), value.transpose(1, 2), is_causal=True).transpose(1, 2)
print(y.squeeze()[0, 0, ...])

torch.sum(torch.abs(x.squeeze()-y.squeeze()))
```

monai/networks/blocks/transformerblock.py
virginiafdez: Yes, it was the "fill paragraph" button that changed these. Same for the self attention block.

monai/networks/blocks/crossattention.py
mingxin-zheng: Hi @KumoLiu and @virginiafdez , when I read this thread:
https://github.com/pytorch/pytorch/issues/96099#issuecomment-1480430583
It seems to me that the relative positional embedding can be integrated with `scaled_dot_product_attention` 's `attn_mask` argument. However, it can be slow as it's not taking the "fast path".

Do you think we can keep this option open for users who wants to use flash_attention and rel_pos_embedding?

tests/test_crossattention.py
```diff
@@ -31,25 +32,28 @@
...
-                    TEST_CASE_CABLOCK.append(test_case)
+                    for flash_attn in [True, False]:
+                        test_case = [
+                            {
+                                "hidden_size": hidden_size,
+                                "num_heads": num_heads,
+                                "dropout_rate": dropout_rate,
+                                "rel_pos_embedding": rel_pos_embedding if not flash_attn else None,
+                                "input_size": input_size,
+                                "use_flash_attention": flash_attn,
```
mingxin-zheng, line 43: Hi @virginiafdez , do you think it would make sense to compare the output consistency with versus without `use_flash_attention`?

Nic-Ma:
@yiheng-wang-nv Could you please help also review this PR?

Thanks in advance.

monai/networks/blocks/crossattention.py
```diff
@@ -52,16 +53,19 @@ def __init__(
             dropout_rate (float, optional): fraction of the input units to drop. Defaults to 0.0.
             hidden_input_size (int, optional): dimension of the input tensor. Defaults to hidden_size.
             context_input_size (int, optional): dimension of the context tensor. Defaults to hidden_size.
-            dim_head (int, optional): dimension of each head. Defaults to hidden_size // num_heads.
+            dim_head (int, optional): dimension of each head. Defaults to  hidden_size // num_heads.
```
yiheng-wang-nv, line 56: seems add one more (unnecessary) space

monai/networks/blocks/selfattention.py
```diff
@@ -153,8 +153,13 @@ def forward(self, x):
 
         if self.use_flash_attention:
             x = F.scaled_dot_product_attention(
-                q, k, v, scale=self.scale, dropout_p=self.dropout_rate, is_causal=self.causal
-            )
+                query=q.transpose(1, 2),
```
KumoLiu, line 156: You can use rearrange to to avoid add transpose here for the q k and v, thanks.

monai/networks/blocks/crossattention.py
```diff
@@ -145,23 +165,34 @@ def forward(self, x: torch.Tensor, context: Optional[torch.Tensor] = None):
         q = q.view(b, t, self.num_heads, c // self.num_heads).transpose(1, 2)  # (b, nh, t,  hs)
         k = k.view(b, kv_t, self.num_heads, c // self.num_heads).transpose(1, 2)  # (b, nh, kv_t, hs)
         v = v.view(b, kv_t, self.num_heads, c // self.num_heads).transpose(1, 2)  # (b, nh, kv_t, hs)
-        att_mat = torch.einsum("blxd,blyd->blxy", q, k) * self.scale
 
-        # apply relative positional embedding if defined
-        att_mat = self.rel_positional_embedding(x, att_mat, q) if self.rel_positional_embedding is not None else att_mat
+        if self.use_flash_attention:
+            x = torch.nn.functional.scaled_dot_product_attention(
+                query=q.transpose(1, 2),
```
KumoLiu, line 171: Perhaps you can directly remove the transpose in Line165-Line167. I didn't test cross attention. Could you please test whether the  shape change here is right.

KumoLiu: The cases for cross attention and self attention are not fully equal.

monai/networks/blocks/selfattention.py
```diff
@@ -130,23 +151,34 @@ def forward(self, x):
...
-        # apply relative positional embedding if defined
-        att_mat = self.rel_positional_embedding(x, att_mat, q) if self.rel_positional_embedding is not None else att_mat
+            # apply relative positional embedding if defined
+            if self.rel_positional_embedding is not None:
+                att_mat = self.rel_positional_embedding(x, att_mat, q)
 
-        if self.causal:
-            att_mat = att_mat.masked_fill(self.causal_mask[:, :, : x.shape[1], : x.shape[1]] == 0, float("-inf"))
+            if self.causal:
+                att_mat = att_mat.masked_fill(self.causal_mask[:, :, : x.shape[1], : x.shape[1]] == 0, float("-inf"))
```
ericspod, line 171: Should be `att_mat = att_mat.masked_fill(causal_mask[:, :, : query.shape[-2], : query.shape[-2]] == 0, float("-inf"))` ???

KumoLiu: I guess so, should be consistent with the generative repo. Should be sequence length.
https://github.com/Project-MONAI/GenerativeModels/blob/7428fce193771e9564f29b91d29e523dd1b6b4cd/generative/networks/blocks/selfattention.py#L136

KumoLiu:
Perhaps we need also include the test @ericspod shared which compare the result between without flash attention and attention that Pytorch says is the same as `F.scaled_dot_product_attention`. The dimensions appear to be highly prone to error, and additional comprehensive testing is necessary to prevent them. Thanks!

ericspod:
Related issues: #7991 #7992

tests/test_crossattention.py
virginiafdez: @ericspod should we add more use cases here?

monai/networks/blocks/crossattention.py
ericspod: We're working on getting the cross and self attention modules to replicate the results from GenerativeModels so slightly suboptimal code like this will be here. I'd suggest we finish up this PR, integrate the other PRs relating to these classes, verify it performs, and then work on cleaning up these sorts of things.

monai/networks/blocks/transformerblock.py
```diff
@@ -36,15 +36,17 @@ def __init__(
         causal: bool = False,
         sequence_length: int | None = None,
         with_cross_attention: bool = False,
+        use_flash_attention: bool = False,
     ) -> None:
         """
         Args:
-            hidden_size (int): dimension of hidden layer.
-            mlp_dim (int): dimension of feedforward layer.
+            hidden_size (int): dimension of hidden layer. mlp_dim (int): dimension of feedforward layer.
```
KumoLiu, line 43:
```suggestion
            hidden_size (int): dimension of hidden layer.
            mlp_dim (int): dimension of feedforward layer.
```

monai/networks/blocks/transformerblock.py
```diff
@@ -36,15 +36,17 @@ def __init__(
...
             num_heads (int): number of attention heads.
-            dropout_rate (float, optional): fraction of the input units to drop. Defaults to 0.0.
-            qkv_bias (bool, optional): apply bias term for the qkv linear layer. Defaults to False.
+            dropout_rate (float, optional): fraction of the input units to drop. Defaults to 0.0. qkv_bias
+             (bool, optional): apply bias term for the qkv linear layer. Defaults to False.
```
KumoLiu, line 46:
```suggestion
            dropout_rate (float, optional): fraction of the input units to drop. Defaults to 0.0.
            qkv_bias (bool, optional): apply bias term for the qkv linear layer. Defaults to False.
```

monai/networks/blocks/crossattention.py
ericspod: @mingxin-zheng can you create an issue for that feature please? We've got a number of moving parts already with this PR and later ones building on it, but we can come back to this afterward.

tests/test_crossattention.py
ericspod: I'm working on the consistency but we've seen that results are different with and without flash attention, and with CPU vs GPU.

monai/networks/blocks/crossattention.py
ericspod:
```suggestion
            dim_head (int, optional): dimension of each head. Defaults to hidden_size // num_heads.
```

ericspod:
@KumoLiu @mingxin-zheng @yiheng-wang-nv we have the flash attention added here. There are other changes we want to do in further PRs such as #7996 but we should merge this one first. I know there's issues of consistency between this layer and the one in GenerativeModels, I'm working on narrowing down where that's coming from. For now I would say we resolve whatever conversations are outstanding, run blossom, and hopefully merge soon. We can come back to these issues if there are any. Also @virginiafdez will be away from tomorrow and less able to work on things.

KumoLiu:
run more test and will address the remaining issues in the next PR.

monai/networks/blocks/crossattention.py
KumoLiu: Created, closed this one.

monai/networks/blocks/selfattention.py
KumoLiu: Will address in https://github.com/Project-MONAI/MONAI/pull/7996

monai/networks/blocks/crossattention.py
KumoLiu: Will address in https://github.com/Project-MONAI/MONAI/pull/7996

tests/test_crossattention.py
KumoLiu: Will try add more tests in https://github.com/Project-MONAI/MONAI/pull/7996

KumoLiu (approved):

ericspod (approved):

monai/networks/blocks/crossattention.py
```diff
@@ -142,26 +162,39 @@ def forward(self, x: torch.Tensor, context: Optional[torch.Tensor] = None):
             q = q.to(self.attention_dtype)
             k = k.to(self.attention_dtype)
 
-        q = q.view(b, t, self.num_heads, c // self.num_heads).transpose(1, 2)  # (b, nh, t,  hs)
+        q = q.view(b, t, self.num_heads, c // self.num_heads).transpose(1, 2)  # (b, nh, t,  hs) #
```
mingxin-zheng, line 165: The `#` in the end seems unnecessary (This is non-blocker comment for a merge)

PR #7981, opened 2024-08-02
Adding a network CellSamWrapper

myron:
Adding a network CellSamWrapper, a thin wrapper around SAM, which can be used for 2D segmentation tasks.

### Types of changes

### DIFF
--- a/monai/networks/nets/cell_sam_wrapper.py
+++ b/monai/networks/nets/cell_sam_wrapper.py
@@ -33,7 +33,7 @@ class CellSamWrapper(torch.nn.Module):
             (currently SAM expects 1024x1024)
         checkpoint: checkpoint file to load the SAM weights from.
             (this can be downloaded from SAM repo https://dl.fbaipublicfiles.com/segment_anything/sam_vit_b_01ec64.pth)
-        return_features: whether to return features from SAM encoder 
+        return_features: whether to return features from SAM encoder
             (without using decoder/upsampling to the original input size)
 
     """
--- a/monai/networks/nets/cell_sam_wrapper.py
+++ b/monai/networks/nets/cell_sam_wrapper.py
@@ -1,0 +1,87 @@
+# Copyright (c) MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import torch
+from torch import nn
+from torch.nn import functional as F
+
+try:
+    from segment_anything.build_sam import build_sam_vit_b
+except ImportError:
+    raise AssertionError("SAM is not installed, please run: pip install git+https://github.com/facebookresearch/segment-anything.git")
+
+
+_all__ = ["CellSamWrapper"]
+
+class CellSamWrapper(torch.nn.Module):
+    """
+    CellSamWrapper is thin wrapper around SAM model https://github.com/facebookresearch/segment-anything
+    with an image only decoder, that can be used for segmentation tasks.
+
+
+    Args:
+        auto_resize_inputs: whether to resize inputs before passing to the network.
+        network_resize_roi: expected input size for the network.
+        checkpoint: checkpoint file to load the SAM weights from.
+        return_features: whether to return features 
+
+    """
+        
+    def __init__(
+        self,
+        auto_resize_inputs=True,
+        network_resize_roi=[1024, 1024],
+        checkpoint="sam_vit_b_01ec64.pth",
+        return_features=False,
+        *args,
+        **kwargs,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+
+        print(
+            f"CellSamWrapper auto_resize_inputs {auto_resize_inputs} network_resize_roi {network_resize_roi} checkpoint {checkpoint}"
+        )
+        self.network_resize_roi = network_resize_roi
+        self.auto_resize_inputs = auto_resize_inputs
+        self.return_features = return_features
+
+        model = build_sam_vit_b(checkpoint=checkpoint)
+
+        model.prompt_encoder = None
+        model.mask_decoder = None
+
+        model.mask_decoder = nn.Sequential(
+            nn.BatchNorm2d(num_features=256),
+            nn.ReLU(inplace=True),
+            nn.ConvTranspose2d(256, 128, kernel_size=3, stride=2, padding=1, output_padding=1, bias=False),
+            nn.BatchNorm2d(num_features=128),
+            nn.ReLU(inplace=True),
+            nn.ConvTranspose2d(128, 3, kernel_size=3, stride=2, padding=1, output_padding=1, bias=True),
+        )
+
+        self.model = model
+
+    def forward(self, x):
+        sh = x.shape[2:]
+
+        if self.auto_resize_inputs:
+            x = F.interpolate(x, size=self.network_resize_roi, mode="bilinear")
+
+        x = self.model.image_encoder(x)  
+
+        if not self.return_features:
+            x = self.model.mask_decoder(x)
+            if self.auto_resize_inputs:
+                x = F.interpolate(x, size=sh, mode="bilinear")
+
+        return x
--- a/monai/networks/nets/cell_sam_wrapper.py
+++ b/monai/networks/nets/cell_sam_wrapper.py
@@ -28,9 +28,13 @@ class CellSamWrapper(torch.nn.Module):
 
     Args:
         auto_resize_inputs: whether to resize inputs before passing to the network.
+            (usually they need be resized, unless they are already at the expected size)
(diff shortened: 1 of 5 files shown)


monai/networks/nets/cell_sam_wrapper.py
```diff
@@ -0,0 +1,87 @@
...
+# limitations under the License.
+
+from __future__ import annotations
+
+import torch
+from torch import nn
+from torch.nn import functional as F
+
+try:
+    from segment_anything.build_sam import build_sam_vit_b
```
ericspod, line 19: Please use `optional_import` to have optional packages included in code. If this package isn't present nothing should be said or raised, but the class should raise an exception upon instantiation.

monai/networks/nets/cell_sam_wrapper.py
```diff
@@ -0,0 +1,87 @@
...
+        auto_resize_inputs=True,
+        network_resize_roi=[1024, 1024],
+        checkpoint="sam_vit_b_01ec64.pth",
+        return_features=False,
+        *args,
+        **kwargs,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+
+        print(
```
ericspod, line 51: Please don't have print statements in code, you can write to the log if you wish though this is typically not necessary.

monai/networks/nets/cell_sam_wrapper.py
myron: ok I changed it, and added a unit test. plz check. thank you!

monai/networks/nets/cell_sam_wrapper.py
myron: removed. thank you!

monai/networks/nets/cell_sam_wrapper.py
```diff
@@ -0,0 +1,84 @@
...
+
+        model = build_sam_vit_b(checkpoint=checkpoint)
+
+        model.prompt_encoder = None
+        model.mask_decoder = None
+
+        model.mask_decoder = nn.Sequential(
+            nn.BatchNorm2d(num_features=256),
+            nn.ReLU(inplace=True),
+            nn.ConvTranspose2d(256, 128, kernel_size=3, stride=2, padding=1, output_padding=1, bias=False),
```
KumoLiu, line 63: Can we add these hard-coded feature num as arguments? Or is it that the user does not need to modify these parameters for some reason?

monai/networks/nets/cell_sam_wrapper.py
```diff
@@ -0,0 +1,84 @@
...
+            nn.ConvTranspose2d(128, 3, kernel_size=3, stride=2, padding=1, output_padding=1, bias=True),
+        )
+
+        self.model = model
+
+    def forward(self, x):
+        sh = x.shape[2:]
+
+        if self.auto_resize_inputs:
+            x = F.interpolate(x, size=self.network_resize_roi, mode="bilinear")
```
KumoLiu, line 75: Also for the interpolate mode here.

monai/networks/nets/cell_sam_wrapper.py
```diff
@@ -0,0 +1,84 @@
...
+
+        x = self.model.image_encoder(x)
+
+        if not self.return_features:
```
KumoLiu, line 79: Could you please help explain more for this `return_features`? According to the code, you will decode the data and interpolate it back to its original size if the arg sets to `False`. Otherwise it will return the encoder's result directly. Perhaps we can write this more thoroughly in docstring to help users understand. Thanks!

tests/test_cell_sam_wrapper.py
```diff
@@ -0,0 +1,62 @@
...
+    def test_shape(self, input_param, input_shape, expected_shape):
+        net = CellSamWrapper(**input_param).to(device)
+        with eval_mode(net):
+            result = net(torch.randn(input_shape).to(device))
+            self.assertEqual(result.shape, expected_shape, msg=str(input_param))
+
+    def test_ill_arg0(self):
+        with self.assertRaises(RuntimeError):
+            net = CellSamWrapper(auto_resize_inputs=False, checkpoint=None).to(device)
+            net(torch.randn([1, 3, 256, 256]))
```
KumoLiu, line 54: The error show here for `test_ill_arg0` and `test_ill_arg0` are both due to the network and the data are not on the same device. I guess that's not what you're trying to test here. 
Does the input have to be resize to 1024?

monai/networks/nets/cell_sam_wrapper.py
```diff
@@ -0,0 +1,84 @@
...
+        checkpoint: checkpoint file to load the SAM weights from.
+        return_features: whether to return features
+
+    """
+
+    def __init__(
+        self,
+        auto_resize_inputs=True,
+        network_resize_roi=[1024, 1024],
+        checkpoint="sam_vit_b_01ec64.pth",
```
KumoLiu, line 41: Where can users find these checkpoint and could you please mention all supported checkpoint here? The CI didn't throw the error since no sam installed in the ci env.

KumoLiu (commented):
Thanks for the PR, overall looks good me, several minor comments inline.

monai/networks/nets/cell_sam_wrapper.py
myron: mmm, SAM output is 256 features, if the user changes it , it will not work. let's keep hardcoded

KumoLiu:
Hi @myron, please help include the segment_anything in the [requirements-dev.txt](https://github.com/Project-MONAI/MONAI/blob/dev/requirements-dev.txt), then it will be installed in the ci env and test all the things. Thanks!

myron:
> Hi @myron, please help include the segment_anything in the [requirements-dev.txt](https://github.com/Project-MONAI/MONAI/blob/dev/requirements-dev.txt), then it will be installed in the ci env and test all the things. Thanks!

Hi, I can do it, but I don't think we should do it, it's a very limited use case, and I don't think we should ask every developer to install it.

monai/networks/nets/cell_sam_wrapper.py
myron: We don't need it as argument

monai/networks/nets/cell_sam_wrapper.py
myron: The checkpoint is optional, and can be downloaded from SAM repository

myron: We don't use any other model configurations at the moment. We could support it, but we don't have use cases for it

tests/test_cell_sam_wrapper.py
myron: Sam expects 1024 input, otherwise it will fail

monai/networks/nets/cell_sam_wrapper.py
KumoLiu: Yes, then could you please mention where users can download it in the docstring? Thanks.

tests/test_cell_sam_wrapper.py
KumoLiu: Got it, then could you please help mention it in the docstring? Thanks.

KumoLiu:
> Hi, I can do it, but I don't think we should do it, it's a very limited use case, and I don't think we should ask every developer to install it.

Hi @myron, yes we need to include that, then all the test cases can be tested correctly. Users don't need to install "requirement-dev", they can choose what they want to install like `pip install monai["nibabel"]`.

monai/networks/nets/cell_sam_wrapper.py
myron: add info

tests/test_cell_sam_wrapper.py
myron: added

myron:
> Hi @myron, please help include the segment_anything in the [requirements-dev.txt](https://github.com/Project-MONAI/MONAI/blob/dev/requirements-dev.txt), then it will be installed in the ci env and test all the things. Thanks!

added

monai/networks/nets/cell_sam_wrapper.py
myron: done

myron:
all done, please check

KumoLiu (approved):
@myron Thanks for the quick update. Just push a commit help fix format issue.

requirements-dev.txt
```diff
@@ -59,3 +59,4 @@ nvidia-ml-py
 huggingface_hub
 pyamg>=5.0.0
 git+https://github.com/Project-MONAI/GenerativeModels.git@7428fce193771e9564f29b91d29e523dd1b6b4cd
```
Nic-Ma, line 61: @KumoLiu please remove this GenerativeModels dependency after refactoring MAISI.

Thanks.

requirements-dev.txt
```diff
@@ -59,3 +59,4 @@ nvidia-ml-py
...
+git+https://github.com/facebookresearch/segment-anything.git@6fdee8f2727f4506cfbbe553e23b895e27956588
```
Nic-Ma, line 62: Hi @KumoLiu ,

Should we add all the required places for this new optional dependency?
Referring to: https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#adding-new-optional-dependencies.

Thanks.

requirements-dev.txt
KumoLiu: Sure, will remove it in maisi-related prs.

requirements-dev.txt
KumoLiu: I try to add it in the setup.cfg.  Other place should be ok.

PR #7984, opened 2024-08-02
Fix: Small logic mistake in the `AsDiscrete.__call__` method

dcfidalgo:
Hi MONAI Team! 
Thank you very much for this super nice framework, really appreciate it!
Just found a small logic mistake in one of the transform classes. To reproduce:
```python
import torch
from monai.transforms.post.array import AsDiscrete

transform = AsDiscrete(argmax=True)
prediction = torch.rand(2, 3, 3)

transform(prediction, argmax=False)
# will still apply argmax
```

### Description

Proposed fix: `argmax` is explicitly checked for `None` in the `__cal__` method.

### Types of changes


ericspod (approved):
Thanks for the contribution! I think this looks good.

PR #7996, opened 2024-08-06
Add include_fc and `use_combined_linear` argument in the `SABlock`

### LINKED ISSUE #7991: Unused proj_attn in the attention block
`proj_attn` is not used in the original generative repo, we may add an argument in `SABlock` to make the behaviour consistent with before.

https://github.com/Project-MONAI/GenerativeModels/blob/7428fce193771e9564f29b91d29e523dd1b6b4cd/generative/networks/nets/diffusion_model_unet.py#L383

https://github.com/Project-MONAI/MONAI/blob/56ee32e36c5c0c7a5cb10afa4ec5589c81171e6b/monai/networks/blocks/selfattention.py#L87

### LINKED ISSUE #7992: Inconsistent usage of the linear layer in the SABlock
There are three linear in the original generative repo which now is combined within one linear layer. We need add an argument to make it consistent with before.

https://github.com/Project-MONAI/GenerativeModels/blob/main/generative/networks/nets/diffusion_model_unet.py#L379-L381

https://github.com/Project-MONAI/MONAI/blob/56ee32e36c5c0c7a5cb10afa4ec5589c81171e6b/monai/networks/blocks/selfattention.py#L89

KumoLiu:
Fixes #7991
Fixes #7992

### Description
Add `include_fc` and `use_combined_linear` argument in the `SABlock`.

### Types of changes

### DIFF
--- a/monai/networks/blocks/selfattention.py
+++ b/monai/networks/blocks/selfattention.py
@@ -182,7 +182,7 @@ class SABlock(nn.Module):
                 att_mat = self.rel_positional_embedding(x, att_mat, q)
 
             if self.causal:
-                att_mat = att_mat.masked_fill(self.causal_mask[:, :, :x.shape[-2], :x.shape[-2]] == 0, float("-inf"))
+                att_mat = att_mat.masked_fill(self.causal_mask[:, :, : x.shape[-2], : x.shape[-2]] == 0, float("-inf"))
 
             att_mat = att_mat.softmax(dim=-1)
 
@@ -193,10 +193,7 @@ class SABlock(nn.Module):
 
             att_mat = self.drop_weights(att_mat)
             x = torch.einsum("bhxy,bhyd->bhxd", att_mat, v)
-        
-            y = torch.nn.functional.scaled_dot_product_attention(
-                query=q, key=k, value=v, scale=self.scale, dropout_p=self.dropout_rate, is_causal=self.causal
-            )
+
 
         x = self.out_rearrange(x)
         if self.include_fc:
--- a/monai/networks/blocks/selfattention.py
+++ b/monai/networks/blocks/selfattention.py
@@ -15,9 +15,10 @@ from typing import Tuple
 
 import torch
 import torch.nn as nn
+import torch.nn.functional as F
 
 from monai.networks.layers.utils import get_rel_pos_embedding_layer
-from monai.utils import optional_import
+from monai.utils import optional_import, pytorch_after
 
 Rearrange, _ = optional_import("einops.layers.torch", name="Rearrange")
 
@@ -44,6 +45,7 @@ class SABlock(nn.Module):
         attention_dtype: torch.dtype | None = None,
         include_fc: bool = True,
         use_combined_linear: bool = True,
+        use_flash_attention: bool = False,
     ) -> None:
         """
         Args:
@@ -63,6 +65,9 @@ class SABlock(nn.Module):
             attention_dtype: cast attention operations to this dtype.
             include_fc: whether to include the final linear layer. Default to True.
             use_combined_linear: whether to use a single linear layer for qkv projection, default to True.
+            use_flash_attention: if True, use Pytorch's inbuilt
+                flash attention for a memory efficient attention mechanism (see
+                https://pytorch.org/docs/2.2/generated/torch.nn.functional.scaled_dot_product_attention.html).
 
         """
 
@@ -86,6 +91,20 @@ class SABlock(nn.Module):
         if causal and sequence_length is None:
             raise ValueError("sequence_length is necessary for causal attention.")
 
+        if use_flash_attention and not pytorch_after(minor=13, major=1, patch=0):
+            raise ValueError(
+                "use_flash_attention is only supported for PyTorch versions >= 2.0."
+                "Upgrade your PyTorch or set the flag to False."
+            )
+        if use_flash_attention and save_attn:
+            raise ValueError(
+                "save_attn has been set to True, but use_flash_attention is also set"
+                "to True. save_attn can only be used if use_flash_attention is False."
+            )
+
+        if use_flash_attention and rel_pos_embedding is not None:
+            raise ValueError("rel_pos_embedding must be None if you are using flash_attention.")
+
         self.num_heads = num_heads
         self.hidden_input_size = hidden_input_size if hidden_input_size else hidden_size
         self.out_proj = nn.Linear(self.inner_dim, self.hidden_input_size)
@@ -103,6 +122,7 @@ class SABlock(nn.Module):
         self.out_rearrange = Rearrange("b l h d -> b h (l d)")
         self.drop_output = nn.Dropout(dropout_rate)
         self.drop_weights = nn.Dropout(dropout_rate)
+        self.dropout_rate = dropout_rate
         self.scale = self.dim_head**-0.5
         self.save_attn = save_attn
         self.att_mat = torch.Tensor()
@@ -111,6 +131,7 @@ class SABlock(nn.Module):
         self.sequence_length = sequence_length
         self.include_fc = include_fc
         self.use_combined_linear = use_combined_linear
+        self.use_flash_attention = use_flash_attention
 
         if causal and sequence_length is not None:
             # causal mask to ensure that attention is only applied to the left in the input sequence
@@ -149,23 +170,34 @@ class SABlock(nn.Module):
             q = q.to(self.attention_dtype)
             k = k.to(self.attention_dtype)
 
-        att_mat = torch.einsum("blxd,blyd->blxy", q, k) * self.scale
+        if self.use_flash_attention:
+            x = F.scaled_dot_product_attention(
+                query=q.transpose(1, 2),
+                key=k.transpose(1, 2),
+                value=v.transpose(1, 2),
+                scale=self.scale,
+                dropout_p=self.dropout_rate,
+                is_causal=self.causal,
+            ).transpose(1, 2)
+        else:
+            att_mat = torch.einsum("blxd,blyd->blxy", q, k) * self.scale
 
-        # apply relative positional embedding if defined
-        att_mat = self.rel_positional_embedding(x, att_mat, q) if self.rel_positional_embedding is not None else att_mat
+            # apply relative positional embedding if defined
+            if self.rel_positional_embedding is not None:
+                att_mat = self.rel_positional_embedding(x, att_mat, q)
 
-        if self.causal:
-            att_mat = att_mat.masked_fill(self.causal_mask[:, :, : x.shape[1], : x.shape[1]] == 0, float("-inf"))
+            if self.causal:
+                att_mat = att_mat.masked_fill(self.causal_mask[:, :, : x.shape[-2], : x.shape[-2]] == 0, float("-inf"))
 
-        att_mat = att_mat.softmax(dim=-1)
+            att_mat = att_mat.softmax(dim=-1)
 
-        if self.save_attn:
-            # no gradients and new tensor;
-            # https://pytorch.org/docs/stable/generated/torch.Tensor.detach.html
-            self.att_mat = att_mat.detach()
+            if self.save_attn:
+                # no gradients and new tensor;
+                # https://pytorch.org/docs/stable/generated/torch.Tensor.detach.html
+                self.att_mat = att_mat.detach()
 
-        att_mat = self.drop_weights(att_mat)
-        x = torch.einsum("bhxy,bhyd->bhxd", att_mat, v)
+            att_mat = self.drop_weights(att_mat)
+            x = torch.einsum("bhxy,bhyd->bhxd", att_mat, v)
         x = self.out_rearrange(x)
         if self.include_fc:
             x = self.out_proj(x)
--- a/monai/networks/blocks/selfattention.py
+++ b/monai/networks/blocks/selfattention.py
@@ -65,9 +65,8 @@ class SABlock(nn.Module):
             attention_dtype: cast attention operations to this dtype.
             include_fc: whether to include the final linear layer. Default to True.
             use_combined_linear: whether to use a single linear layer for qkv projection, default to True.
-            use_flash_attention: if True, use Pytorch's inbuilt
-                flash attention for a memory efficient attention mechanism (see
-                https://pytorch.org/docs/2.2/generated/torch.nn.functional.scaled_dot_product_attention.html).
+            use_flash_attention: if True, use Pytorch's inbuilt flash attention for a memory efficient attention mechanism
+                (see https://pytorch.org/docs/2.2/generated/torch.nn.functional.scaled_dot_product_attention.html).
 
         """
 
--- a/monai/networks/blocks/selfattention.py
+++ b/monai/networks/blocks/selfattention.py
@@ -194,7 +194,6 @@ class SABlock(nn.Module):
             att_mat = self.drop_weights(att_mat)
             x = torch.einsum("bhxy,bhyd->bhxd", att_mat, v)
 
-
         x = self.out_rearrange(x)
         if self.include_fc:
             x = self.out_proj(x)
--- a/monai/networks/blocks/selfattention.py
+++ b/monai/networks/blocks/selfattention.py
@@ -172,12 +172,7 @@ class SABlock(nn.Module):
 
         if self.use_flash_attention:
             x = F.scaled_dot_product_attention(
-                query=q,
-                key=k,
-                value=v,
-                scale=self.scale,
-                dropout_p=self.dropout_rate,
-                is_causal=self.causal,
+                query=q, key=k, value=v, scale=self.scale, dropout_p=self.dropout_rate, is_causal=self.causal
             )
         else:
             att_mat = torch.einsum("blxd,blyd->blxy", q, k) * self.scale
--- a/monai/networks/blocks/selfattention.py
+++ b/monai/networks/blocks/selfattention.py
@@ -43,6 +43,7 @@ class SABlock(nn.Module):
         input_size: Tuple | None = None,
         attention_dtype: torch.dtype | None = None,
         include_fc: bool = True,
+        use_combined_linear: bool = True,
     ) -> None:
         """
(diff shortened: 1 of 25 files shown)


KumoLiu:
`RelPosEmbedding.DECOMPOSED` can not convert to torchscript, leave it in this PR. Error shows as below.
```
'__torch__.torch.nn.modules.container.ParameterList (of Python compilation unit at: 0x55c00fc6db20)' object has no attribute or method '__len__'. Did you forget to initialize an attribute in __init__()?:
  File "/workspace/Code/MONAI/monai/networks/blocks/attention_utils.py", line 98
```

KumoLiu:
Remaining parts need to be addressed.
https://github.com/Project-MONAI/MONAI/pull/7977#discussion_r1704176137
https://github.com/Project-MONAI/MONAI/pull/7977#discussion_r1705752334

monai/networks/blocks/selfattention.py
```diff
@@ -86,9 +90,17 @@ def __init__(
...
+            self.qkv = nn.Linear(self.hidden_input_size, self.inner_dim * 3, bias=qkv_bias)
+            self.to_q = self.to_k = self.to_v = nn.Identity()  # add to enable torchscript
+            self.input_rearrange = Rearrange("b h (qkv l d) -> qkv b l h d", qkv=3, l=num_heads)
+        else:
+            self.to_q = nn.Linear(self.hidden_input_size, self.inner_dim, bias=qkv_bias)
+            self.to_k = nn.Linear(self.hidden_input_size, self.inner_dim, bias=qkv_bias)
+            self.to_v = nn.Linear(self.hidden_input_size, self.inner_dim, bias=qkv_bias)
+            self.qkv = nn.Identity()  # add to enable torchscript
+            self.input_rearrange = Rearrange("b h (l d) -> b l h d", l=num_heads)
+        self.out_rearrange = Rearrange("b l h d -> b h (l d)")
```
ericspod, line 103: If we don't have a compelling reason for this to be an option, we can just use the 3 linear layers without this being a choice. We haven't released 1.4 yet so nothing here should be dependent yet.

KumoLiu: The combined way was introduced since SABlock added. I've mentioned [here](https://github.com/Project-MONAI/MONAI/issues/7992#issuecomment-2270802435) that there will be a big gap between the two, and maybe we should make this an option, otherwise for the user, the behavior will change without any notice. What do you think?
https://github.com/Project-MONAI/MONAI/commit/d41af7e498677ed813aeb39bc541bc36f72e86cb#diff-bb1ecca81545a8d2ce544bd6115ddb48e733ad85ea4fdb1edf245ce73c0db6b2

Also cc @Nic-Ma for vis.

ericspod: Yes, it's true the behaviour is different but I don't know if there's any users yet who would be impacted by making it 3 linears.

mingxin-zheng: In an earlier conversation with @KumoLiu , it seems that the trained MAISI model is already using the combined way.

ericspod: @mingxin-zheng you had me worried for a moment! I would propose we just use the 3 linears and not bother with an option. The combined way seems to have been an optimisation that isn't necessary.

KumoLiu: I'm not sure about this one, add it as an option or directly use three linears.
And for your reference.
https://github.com/microsoft/Swin-Transformer/blob/main/models/swin_transformer.py#L117

Open for this one, want to hear more opinion from other person.
cc @Nic-Ma @mingxin-zheng @heyufan1995 @yiheng-wang-nv @dongyang0122 @guopengf

guopengf: It seems like treating it as an option is more flexible.

tests/test_selfattention.py
```diff
@@ -175,6 +179,39 @@ def count_sablock_params(*args, **kwargs):
...
+                    "use_combined_linear": use_combined_linear,
+                }
+        net = SABlock(**input_param)
+        input_shape = (2, 512, 360)
+        test_data = torch.randn(input_shape)
+        test_script_save(net, test_data)
+
+    @skipUnless(has_einops, "Requires einops")
+    @SkipIfBeforePyTorchVersion((2, 0))
+    def test_flash_attention(self):
```
KumoLiu, line 202: Have add test based on https://github.com/Project-MONAI/MONAI/pull/7977#discussion_r1701827839 for selfattention and cross attention.
Please help review, thanks! @mingxin-zheng @ericspod

monai/networks/blocks/selfattention.py
mingxin-zheng: Since `use_combined_layer=True` is for backward compatibility, so I did some search in tag `1.3.1` for the usage of `SABlock`. 

It turns out that `TransformerBlock` is using it, and `TransformerBlock` is used utilized by `ViT`(`UNETR` depends on it) and `ViTAutoEnc`

https://github.com/Project-MONAI/MONAI/blob/96bfda00c6bd290297f5e3514ea227c6be4d08b4/monai/networks/blocks/transformerblock.py#L56

Removing the combined layer, in my observation, will introduce weight loading issue if users are using the blocks and models above, although it is not a fundamental change (user can combine the 3 inner_dim channels manually).

monai/networks/blocks/spatialattention.py
```diff
@@ -46,6 +46,8 @@ def __init__(
         norm_eps: float = 1e-6,
         attention_dtype: Optional[torch.dtype] = None,
         use_flash_attention: bool = False,
+        include_fc: bool = True,
+        use_combined_linear: bool = True,
```
mingxin-zheng, line 50: Since `SpatialAttentionBlock` is introduced from the `gen-ai` feature branch, I think we don't need to be backward compatible and can set `use_combined_linear` to `False`

tests/test_unetr.py
```diff
@@ -123,7 +123,7 @@ def test_ill_arg(self):
             )
 
     @parameterized.expand(TEST_CASE_UNETR)
-    @SkipIfBeforePyTorchVersion((1, 9))
+    @SkipIfBeforePyTorchVersion((2, 0))
```
mingxin-zheng, line 126: The `SkipIfBeforePyTorchVersion` decorator has been modified to require PyTorch version 2.0 or later. However, it's not clear why this change was made

KumoLiu: Torchscript convert will fail since the new introduced flash attention.

mingxin-zheng: Shall we fallback to not using flash attention if version < 2.0?

mingxin-zheng: For example:
```
if use_flash_attention and not pytorch_after(minor=13, major=1, patch=0):
    raise ValueError(
        "use_flash_attention is only supported for PyTorch versions >= 2.0."
        "Upgrade your PyTorch or set the flag to False."
    )
```

to:

```
if use_flash_attention and not pytorch_after(minor=13, major=1, patch=0):
    log.warning(...)
    use_flash_attention = False
```

KumoLiu: Yes, it's true. But it may not influence torchscript convert which will go through all the sub-branch.
https://github.com/Project-MONAI/MONAI/blob/49a1e343129909f7e01c67882a71c12beed3ef30/monai/networks/blocks/selfattention.py#L90

mingxin-zheng (approved):

tests/test_crossattention.py
```diff
@@ -166,6 +166,26 @@ def test_access_attn_matrix(self):
...
+            }
+        device = "cuda:0" if torch.cuda.is_available() else "cpu"
+        block_w_flash_attention = CrossAttentionBlock(**input_param, use_flash_attention=True).to(device)
+        block_wo_flash_attention = CrossAttentionBlock(**input_param, use_flash_attention=False).to(device)
+        block_wo_flash_attention.load_state_dict(block_w_flash_attention.state_dict())
+        test_data = torch.randn(1, 16, 128).to(device)
+
+        out_1 = block_w_flash_attention(test_data)
+        out_2 = block_wo_flash_attention(test_data)
+        assert_allclose(out_1, out_2, atol=1e-4)
```
ericspod, line 187:
```suggestion
    @parameterized.expand([[True], [False]])
    @skipUnless(has_einops, "Requires einops")
    @SkipIfBeforePyTorchVersion((2, 0))
    def test_flash_attention(self, causal):
        input_param = {
            "hidden_size": 128,
            "num_heads": 1,
            "causal": causal,
            "sequence_length": 16 if causal else None,
        }
        device = "cuda:0" if torch.cuda.is_available() else "cpu"
        block_w_flash_attention = CrossAttentionBlock(**input_param, use_flash_attention=True).to(device)
        block_wo_flash_attention = CrossAttentionBlock(**input_param, use_flash_attention=False).to(device)
        block_wo_flash_attention.load_state_dict(block_w_flash_attention.state_dict())
        test_data = torch.randn(1, 16, 128).to(device)

        out_1 = block_w_flash_attention(test_data)
        out_2 = block_wo_flash_attention(test_data)
        assert_allclose(out_1, out_2, atol=1e-4)
```
I would expand this into multiple tests so if either case fails we can more easily identify which one it was. We may also want to test with and without CUDA, I've noticed some differences on CPU between using flash attention and not.

tests/test_selfattention.py
```diff
@@ -175,6 +178,40 @@ def count_sablock_params(*args, **kwargs):
...
+                    "dropout_rate": 0.0,
+                    "rel_pos_embedding": None,
+                    "input_size": (16, 32),
+                    "include_fc": include_fc,
+                    "use_combined_linear": use_combined_linear,
+                }
+        net = SABlock(**input_param)
+        input_shape = (2, 512, 360)
+        test_data = torch.randn(input_shape)
+        test_script_save(net, test_data)
```
ericspod, line 198:
```suggestion
    @parameterized.expand([[True,False], [True,True], [False, True],[False, False]])
    @skipUnless(has_einops, "Requires einops")
    @SkipIfBeforePyTorchVersion((2, 0))
    def test_script(self, include_fc, use_combined_linear):
        input_param = {
            "hidden_size": 360,
            "num_heads": 4,
            "dropout_rate": 0.0,
            "rel_pos_embedding": None,
            "input_size": (16, 32),
            "include_fc": include_fc,
            "use_combined_linear": use_combined_linear,
        }
        net = SABlock(**input_param)
        input_shape = (2, 512, 360)
        test_data = torch.randn(input_shape)
        test_script_save(net, test_data)
```
Similar idea here. Also the loops didn't include the lines after `input_param` definition so only tested the last combination of values.

ericspod:
For reference, I had attempted to test the similarity between the `SABlock` in core versus that in GenerativeModels and didn't have any luck narrowing down where the differences were. I have this at the end of `test_selfattension.py` which defines some failing tests that should pass if I haven't missed anything:

```python
from tests.utils import SkipIfBeforePyTorchVersion, assert_allclose
from monai.utils import set_determinism

generative, has_generative = optional_import("generative")
xops, has_xformers = optional_import("xformers.ops")


class TestComparison(unittest.TestCase):
    @parameterized.expand([["cuda:0", True], ["cuda:0", False], ["cpu", False]])
    @skipUnless(has_einops, "Requires einops")
    @skipUnless(has_generative, "Requires generative")
    @skipUnless(has_xformers, "Requires xformers")
    @SkipIfBeforePyTorchVersion((2, 0))
    def test_generative_vs_core(self, _device, use_flash_attention):
        device = torch.device(_device)
        input_shape = (2, 512, 360)
        input_param = {
            "hidden_size": 360,
            "num_heads": 4,
            "dropout_rate": 0,
            "use_flash_attention": use_flash_attention,
        }

        set_determinism(0)
        net_gen = generative.networks.blocks.SABlock(**input_param).to(device)
        set_determinism(0)
        net_monai = SABlock(**input_param).to(device)
        set_determinism(0)
        input = torch.randn(input_shape).to(device)

        net_monai.load_state_dict(net_gen.state_dict())

        with eval_mode(net_gen, net_monai):
            set_determinism(0)
            r1 = net_gen(input)
            set_determinism(0)
            r2 = net_monai(input)
            assert_allclose(r1.detach().cpu().numpy(), r2.detach().cpu().numpy())
```

monai/networks/blocks/selfattention.py
ericspod: I think we've concluded that we leave the option in, I'm good with that.

monai/networks/blocks/selfattention.py
```diff
@@ -179,7 +197,9 @@ def forward(self, x):
 
             att_mat = self.drop_weights(att_mat)
             x = torch.einsum("bhxy,bhyd->bhxd", att_mat, v)
+
         x = self.out_rearrange(x)
-        x = self.out_proj(x)
+        if self.include_fc:
+            x = self.out_proj(x)
```
ericspod, line 203: Note that `out_proj` is defined even if not used, this is compatible with previously saved weights. It seems we have to keep it for this compatibility reason.

tests/test_selfattention.py
ericspod: Similar idea to the previous comment, I would put the choice of `causal` into `parameterized.expand`, this also only tests with `causal==False` case as it currently is.

ericspod:
Hi @KumoLiu I had a few comments about the tests that need to be addressed to test the actual cases you want. I'm otherwise good with things.

guopengf:
Hi @KumoLiu, shall we also add those parameters (`use_flash_attention`,  `include_fc` and `use_combined_linear`) to `monai.networks.nets.controlnet`?

tests/test_selfattention.py
KumoLiu: Oh, my bad here, thanks for pointing it out.

monai/networks/blocks/selfattention.py
KumoLiu: Yes, and also we need defined it to make torchscript can be converted correctly.

tests/test_selfattention.py
KumoLiu: Agree, good points, addressed in the latest commit.

KumoLiu:
> For reference, I had attempted to test the similarity between the `SABlock` in core versus that in GenerativeModels and didn't have any luck narrowing down where the differences were.

Actually I also tested it locally, I've already ensure they return the same result, the issue from your script may due to the initial weight is not the same, so I copy and reuse the state dict. I didn't include this test since we don't want to add generative as a new dependency.
Here is my test code for your reference:
```
import torch
from copy import deepcopy
from generative.networks.nets.diffusion_model_unet import CrossAttention, AttentionBlock
from monai.utils import set_determinism
from monai.networks.blocks import CrossAttentionBlock, SpatialAttentionBlock

set_determinism(0)
use_flash_attention = True

cross_block_gen = CrossAttention(query_dim=512, num_attention_heads=16, num_head_channels=32, use_flash_attention=use_flash_attention).to('cuda')
cross_block_core = CrossAttentionBlock(hidden_size=512, num_heads=16, dim_head=32, use_flash_attention=use_flash_attention).to('cuda')
state_dict_gen_cross = cross_block_gen.state_dict()
state_dict_core_cross = deepcopy(state_dict_gen_cross)
state_dict_core_cross["out_proj.weight"] = state_dict_gen_cross["to_out.0.weight"]
state_dict_core_cross["out_proj.bias"] = state_dict_gen_cross["to_out.0.bias"]
del state_dict_core_cross["to_out.0.weight"]
del state_dict_core_cross["to_out.0.bias"]

atten_block_gen = AttentionBlock(spatial_dims=2, num_channels=512, num_head_channels=32, use_flash_attention=use_flash_attention).to('cuda')
atten_block_core = SpatialAttentionBlock(spatial_dims=2, num_channels=512, num_head_channels=32, use_combined_linear=False, include_fc=False, use_flash_attention=use_flash_attention).to('cuda')
state_dict_gen_atten = atten_block_gen.state_dict()
state_dict_core_atten = atten_block_core.state_dict()
for key in state_dict_core_atten.keys():
    if 'attn' in key and "proj" not in key:
        state_dict_core_atten[key] = state_dict_gen_atten[key[5:]]
    if key in state_dict_gen_atten:
        state_dict_core_atten[key] = state_dict_gen_atten[key]
state_dict_core_atten["attn.out_proj.weight"] = state_dict_gen_atten["proj_attn.weight"]
state_dict_core_atten["attn.out_proj.bias"] = state_dict_gen_atten["proj_attn.bias"]


input_cross = torch.rand(1, 256, 512).to("cuda")
input_cross2 = deepcopy(input_cross)
cross_block_gen.load_state_dict(state_dict_gen_cross)

out_gen = cross_block_gen(input_cross)
cross_block_core.load_state_dict(state_dict_core_cross)
out_core = cross_block_core(input_cross2)

print('diff in cross', (out_gen-out_core).abs().sum())

input_atten = torch.rand(1, 512, 16, 16).to("cuda")
input_atten2 = deepcopy(input_atten)
atten_block_gen.load_state_dict(state_dict_gen_atten)

out_gen = atten_block_gen(input_atten)
atten_block_core.load_state_dict(state_dict_core_atten)
out_core = atten_block_core(input_atten2)

print('diff in atten', (out_gen-out_core).abs().sum())
```

Result:
```
diff in cross tensor(0., device='cuda:0', grad_fn=<SumBackward0>)
diff in atten tensor(0., device='cuda:0', grad_fn=<SumBackward0>)
```


And for this one :
> We may also want to test with and without CUDA, I've noticed some differences on CPU between using flash attention and not.

Yes, we also include these tests. The gpu pipeline have been removed to blossom, but it will stilled be tested.

monai/networks/blocks/selfattention.py
Nic-Ma: @ericspod Thanks, I also agree to keep the option.

KumoLiu:
> Hi @KumoLiu, shall we also add those parameters (`use_flash_attention`, `include_fc` and `use_combined_linear`) to `monai.networks.nets.controlnet` and `monai.networks.nets.autoencoderkl`? Those two networks also use attention layers.

Hi @guopengf, added in the latest commit.

PR #8025, opened 2024-08-19
Move PyType test to weekly test

### LINKED ISSUE #8022: Consider Not Using Pytype in PR tests
Pytype is very slow and there's not much that seems to be done about this. One of the checks in PRs is to run this for everything which significantly slows down the review process occasionally. Would we be better off moving this to an occasional test instead? Typically if mypy is ok with code pytype is as well so there's not that much value.

Move pytype to be weekly test or something like this. Optionally we could test with pyright but this is throwing up a lot of errors currently owing to our import system (which may benefit from a revamp itself).

KumoLiu:
Fixes #8022

### Description

- Add format test to weekly test
- Set pytype test as not required in each PR
- Add packaging in weekly-preview pipeline

### Types of changes

### DIFF
--- a/.github/workflows/weekly-preview.yml
+++ b/.github/workflows/weekly-preview.yml
@@ -1,8 +1,11 @@
 name: weekly-preview
 
 on:
-  schedule:
-  - cron: "0 2 * * 0"  # 02:00 of every Sunday
+  # schedule:
+  # - cron: "0 2 * * 0"  # 02:00 of every Sunday
+  pull_request:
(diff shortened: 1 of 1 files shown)


ericspod:
Hi @Nic-Ma If you're OK with this change so am I. This will speed up doing PRs since we don't have to wait for Pytype to finish but doesn't remove it from our CICD system. Thanks both!

ericspod (approved):

PR #8041, opened 2024-08-22
Fix AttributeError when using torch.min and max

### LINKED ISSUE #8040: AttributeError: 'torch.return_types.max' object has no attribute 'shape'"
```
import torch
from monai.transforms.utils_pytorch_numpy_unification import max

a = torch.rand(6, 8, 3)
b = max(a, dim=1)
print(type(b), (isinstance(b, tuple)))
print(b.shape)
```
Return:
```
<class 'torch.return_types.max'> True
{
	"name": "AttributeError",
	"message": "'torch.return_types.max' object has no attribute 'shape'",
	"stack": "---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[8], line 7
      5 b = max(a, dim=1)
      6 print(type(b), (isinstance(b, tuple)))
----> 7 print(b.shape)

AttributeError: 'torch.return_types.max' object has no attribute 'shape'"
}
```

KumoLiu:
Fixes #8040.

### Description

Only return values if got a namedtuple when using torch.min and max

### Types of changes

### DIFF
--- a/tests/test_utils_pytorch_numpy_unification.py
+++ b/tests/test_utils_pytorch_numpy_unification.py
@@ -17,7 +17,7 @@ import numpy as np
 import torch
 from parameterized import parameterized
 
-from monai.transforms.utils_pytorch_numpy_unification import min, max, mode, percentile
+from monai.transforms.utils_pytorch_numpy_unification import max, min, mode, percentile
 from monai.utils import set_determinism
 from tests.utils import TEST_NDARRAYS, assert_allclose, skip_if_quick
 
--- a/tests/test_utils_pytorch_numpy_unification.py
+++ b/tests/test_utils_pytorch_numpy_unification.py
@@ -17,7 +17,7 @@ import numpy as np
 import torch
 from parameterized import parameterized
 
-from monai.transforms.utils_pytorch_numpy_unification import mode, percentile, min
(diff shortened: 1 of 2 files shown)


tests/test_utils_pytorch_numpy_unification.py
```diff
@@ -27,6 +27,13 @@
     TEST_MODE.append([p(np.array([3.1, 4.1, 4.1, 5.1])), p(4.1), False])
     TEST_MODE.append([p(np.array([3.1, 4.1, 4.1, 5.1])), p(4), True])
 
+TEST_MIN_MAX = []
+for p in TEST_NDARRAYS:
+    TEST_MIN_MAX.append([p(np.array([1, 2, 3, 4, 4, 5])), {}, min, p(1)])
```
mingxin-zheng, line 32: Hi @KumoLiu , thanks for the fix and it looks good me. My only question is if we should cover torch array in the test cases?

KumoLiu: all test cases are covered by using `TEST_NDARRAYS`.
https://github.com/Project-MONAI/MONAI/blob/cea80a686b907b22d90c8024cea5a17ef9d9f58a/tests/utils.py#L839

mingxin-zheng: I see. Thanks!

Issue #8061, opened 2024-09-02
test_trt_compile test error in pytorch 23.08


KumoLiu:
```
[2024-09-02T09:35:38.944Z] ======================================================================
[2024-09-02T09:35:38.944Z] ERROR: test_unet_value_0_fp32 (tests.test_trt_compile.TestTRTCompile)
[2024-09-02T09:35:38.944Z] ----------------------------------------------------------------------
[2024-09-02T09:35:38.944Z] Traceback (most recent call last):
[2024-09-02T09:35:38.944Z]   File "/usr/local/lib/python3.10/dist-packages/parameterized/parameterized.py", line 620, in standalone_func
[2024-09-02T09:35:38.944Z]     return func(*(a + p.args), **p.kwargs, **kw)
[2024-09-02T09:35:38.944Z]   File "/home/jenkins/agent/workspace/Monai-latest-image/tests/test_trt_compile.py", line 91, in test_unet_value
[2024-09-02T09:35:38.944Z]     trt_output = model(input_example)
[2024-09-02T09:35:38.944Z]   File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1505, in _wrapped_call_impl
[2024-09-02T09:35:38.944Z]     return self._call_impl(*args, **kwargs)
[2024-09-02T09:35:38.944Z]   File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1514, in _call_impl
[2024-09-02T09:35:38.944Z]     return forward_call(*args, **kwargs)
[2024-09-02T09:35:38.944Z]   File "/home/jenkins/agent/workspace/Monai-latest-image/monai/networks/trt_compiler.py", line 493, in trt_forward
[2024-09-02T09:35:38.944Z]     return self._trt_compiler.forward(self, argv, kwargs)
[2024-09-02T09:35:38.944Z]   File "/home/jenkins/agent/workspace/Monai-latest-image/monai/networks/trt_compiler.py", line 350, in forward
[2024-09-02T09:35:38.944Z]     raise e
[2024-09-02T09:35:38.944Z]   File "/home/jenkins/agent/workspace/Monai-latest-image/monai/networks/trt_compiler.py", line 342, in forward
[2024-09-02T09:35:38.944Z]     self._build_and_save(model, build_args)
[...]

PR #8062, opened 2024-09-02
Ignore warning from nptyping as workaround

KumoLiu:
workaround for #8061

### Description

Ignore warning from nptyping as workaround

### Types of changes

### DIFF
--- a/monai/data/image_reader.py
+++ b/monai/data/image_reader.py
@@ -34,7 +34,7 @@ from monai.data.utils import (
 )
 from monai.utils import MetaKeys, SpaceKeys, TraceKeys, ensure_tuple, optional_import, require_pkg
 
-# workaround for https://github.com/Project-MONAI/MONAI/pull/8062
+# workaround for https://github.com/Project-MONAI/MONAI/issues/8061
 warnings.filterwarnings("ignore", category=DeprecationWarning, module="nptyping")
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


KumoLiu:
We didn't update the version of nptyping, which is a dependency for pynrrd, yet the error suddenly started occurring. As a workaround, I suppress the warning.
https://github.com/ramonhagenaars/nptyping/blob/785cd07e65f992f47256398fd01f62067928d29c/nptyping/typing_.py#L51

monai/data/image_reader.py
```diff
@@ -34,6 +34,9 @@
 )
 from monai.utils import MetaKeys, SpaceKeys, TraceKeys, ensure_tuple, optional_import, require_pkg
 
+# workaround for https://github.com/Project-MONAI/MONAI/pull/8062
```
ericspod, line 37:
```suggestion
# workaround for https://github.com/Project-MONAI/MONAI/issues/8061
```
Should reference the issue? Odd that something like this would come up without change, so long as this doesn't cover up anything else we're fine.

KumoLiu: Yes, my bad. Thanks for pointing out!

ericspod:
I'm not seeing from the mentioned issue what the nptyping problem is, I guess this fix is fine so long as nothing else is hidden. I would just check that it's not related to Numpy2 or something else being updated elsewhere.

ericspod (approved):

PR #8064, opened 2024-09-03
Trt compiler fixes

borisfom:
Fixes https://github.com/Project-MONAI/MONAI/issues/8061.

### Description

Post-merge fixes for trt_compile()

### Types of changes


KumoLiu (approved):
Thanks for the quick fix, LGTM.

tests/test_trt_compile.py
```diff
@@ -20,10 +20,10 @@
 from monai.handlers import TrtHandler
 from monai.networks import trt_compile
 from monai.networks.nets import UNet, cell_sam_wrapper, vista3d132
-from monai.utils import optional_import
+from monai.utils import min_version, optional_import
 from tests.utils import skip_if_no_cuda, skip_if_quick, skip_if_windows
 
-trt, trt_imported = optional_import("tensorrt")
+trt, trt_imported = optional_import("tensorrt", "10.1.0", min_version)
```
yiheng-wang-nv, line 26: Hi @borisfom , may I ask why set 10.1 instead of 10.3 as the min version here?

binliunls (approved):

tests/test_trt_compile.py
KumoLiu: I guess 10.3 mentioned here is only for best performance, not the least requirement. Correct me if I'm wrong. @borisfom 
https://github.com/Project-MONAI/MONAI/blob/9e98f667bf6655cd0806f920a86d2581534a9d5e/monai/networks/trt_compiler.py#L508

Issue #8060, opened 2024-09-02
 DeprecationWarning when import monai


KumoLiu:
```
/usr/local/lib/python3.10/dist-packages/ignite/handlers/checkpoint.py:16: DeprecationWarning: `TorchScript` support for functional optimizers is deprecated and will be removed in a future PyTorch release. Consider using the `torch.compile` optimizer instead.
  from torch.distributed.optim import ZeroRedundancyOptimizer
/usr/local/lib/python3.10/dist-packages/modelopt/torch/quantization/tensor_quant.py:92: FutureWarning: `torch.library.impl_abstract` was renamed to `torch.library.register_fake`. Please use that instead; we will remove `torch.library.impl_abstract` in a future version of PyTorch.
  scaled_e4m3_abstract = torch.library.impl_abstract("trt::quantize_fp8")(
WARNING:py.warnings:`np.bool8` is a deprecated alias for `np.bool_`.  (Deprecated NumPy 1.24)

WARNING:py.warnings:`np.object0` is a deprecated alias for ``np.object0` is a deprecated alias for `np.object_`. `object` can be used instead.  (Deprecated NumPy 1.24)`.  (Deprecated NumPy 1.24)

WARNING:py.warnings:`np.int0` is a deprecated alias for `np.intp`.  (Deprecated NumPy 1.24)

WARNING:py.warnings:`np.uint0` is a deprecated alias for `np.uintp`.  (Deprecated NumPy 1.24)

WARNING:py.warnings:`np.void0` is a deprecated alias for `np.void`.  (Deprecated NumPy 1.24)

WARNING:py.warnings:`np.bytes0` is a deprecated alias for `np.bytes_`.  (Deprecated NumPy 1.24)

WARNING:py.warnings:`np.str0` is a deprecated alias for `np.str_`.  (Deprecated NumPy 1.24)
```

KumoLiu:
Some new warning:
```
WARNING:py.warnings:pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html

WARNING:py.warnings:Pydantic V1 style `@validator` validators are deprecated. You should migrate to Pydantic V2 style `@field_validator` validators, see the migration guide for more details. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.7/migration/

[...]

PR #8067, opened 2024-09-05
Suppress deprecated warning when import monai

KumoLiu:
- add packaging in setup.cfg
- fix test_gdsdataset.py issue
- add test_matshow3d to the skip list for min test
- suppress deprecated warning when import monai (workaround for #8060)

### Types of changes

### DIFF
--- a/monai/__init__.py
+++ b/monai/__init__.py
@@ -27,6 +27,8 @@ def custom_warning_handler(message, category, filename, lineno, file=None, line=
         "nptyping/typing_",
         "pydantic/_internal/_config",
         "tempfile",
+        "mlflow/gateway",
+        "mlflow/utils",
     ]
     if any(ignore in filename for ignore in ignore_files):
         return
--- a/monai/__init__.py
+++ b/monai/__init__.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 import os
 import sys
+import logging
 import warnings
 from ._version import get_versions
 
@@ -24,19 +25,42 @@ def custom_warning_handler(message, category, filename, lineno, file=None, line=
     ignore_files = [
         "ignite/handlers/checkpoint",
         "modelopt/torch/quantization/tensor_quant",
-        "nptyping/typing_",
-        "pydantic/_internal/_config",
-        "tempfile",
-        "mlflow/gateway",
-        "mlflow/utils",
     ]
     if any(ignore in filename for ignore in ignore_files):
         return
     old_showwarning(message, category, filename, lineno, file, line)
 
+class DeprecatedTypesWarningFilter(logging.Filter):
+    def filter(self, record):
+        deprecated_aliases = [
+            "np.bool8",
+            "np.object0",
+            "np.int0",
+            "np.uint0",
+            "np.void0",
+            "np.str0",
+            "np.bytes0",
+            "@validator",
+            "@root_validator",
+            "class-based `config`",
+            "pkg_resources",
+            "Implicitly cleaning up"
+            ]
+        for alias in deprecated_aliases:
+            if alias in record.getMessage():
+                return False
+        return True
 
 # workaround for https://github.com/Project-MONAI/MONAI/issues/8060
+# Set the custom warning handler to filter warning
 warnings.showwarning = custom_warning_handler
+# Get the logger for warnings
+logger = logging.getLogger("py.warnings")
+# Create and add the filter to the logger
+filter = DeprecatedTypesWarningFilter()
+logger.addFilter(filter)
+
+
 PY_REQUIRED_MAJOR = 3
 PY_REQUIRED_MINOR = 9
 
--- a/monai/__init__.py
+++ b/monai/__init__.py
@@ -30,7 +30,7 @@ def custom_warning_handler(message, category, filename, lineno, file=None, line=
 
 class DeprecatedTypesWarningFilter(logging.Filter):
     def filter(self, record):
-        deprecated_aliases = [
+        message_bodies_to_ignore = [
             "np.bool8",
             "np.object0",
             "np.int0",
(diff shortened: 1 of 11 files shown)


KumoLiu:
I aim to make the suppression of warnings as specific as possible to avoid impacting other parts of the code. For warnings that I can trace to specific imports, like `nptyping`, I apply filtering at the point of import. For warnings where the source is harder to locate, I suppress them based on the warning message in the __init__ file. For those I can't suppress certain warnings directly in __init__, I handle them by adding the suppression logic to optional imports.
cc @ericspod @Nic-Ma @mingxin-zheng, please help take a look to see if you have any concern or better suggestions, thanks.

monai/__init__.py
```diff
@@ -13,9 +13,19 @@
...
 import sys
-
+import warnings
 from ._version import get_versions
 
+
+def custom_warning_handler(message, category, filename, lineno, file=None, line=None):
+    message1 = "`TorchScript` support for functional optimizers is deprecated and will be removed in a future PyTorch release. Consider using the `torch.compile` optimizer instead."
+    message2 = "`torch.library.impl_abstract` was renamed to `torch.library.register_fake`. Please use that instead; we will remove `torch.library.impl_abstract` in a future version of PyTorch."
+    if message1 or message2 in message:
```
ericspod, line 23:
```suggestion
    if message1 in message or message2 in message:
```

ericspod:
It's a little worrying doing a blanket "ignore" on warnings from a library, but I know they're annoying to deal with in any other way. Even though some of the suppression is for optional packages, could we still use your filter function to ignore them? Eg. whether we import mlflow or not we can filter the warnings that it may create.

KumoLiu:
> could we still use your filter function to ignore them? Eg. whether we import mlflow or not we can filter the warnings that it may create.

Looks like I can't use this custom filter handler due to this RecursionError.
```
  File "/Users/runner/work/MONAI/MONAI/monai/__init__.py", line 25, in custom_warning_handler
BasicUNet features: (8, 16, 32, 64, 128, 8).
    warnings.showwarning(message, category, filename, lineno, file, line)
  [Previous line repeated 975 more times]
RecursionError: maximum recursion depth exceeded
```

ericspod:
> Looks like I can't use this custom filter handler due to this RecursionError.

This is possibly because you're calling `showwarning` within it which is itself, you can save the older function as a separate variable and call it, or call `formatwarning` https://docs.python.org/3/library/warnings.html#warnings.showwarning

ericspod:
I don't think it would be better to have this handler added in multiple places. Alternatively can we put the `filterwarnings` calls into `__init__.py` in one place?

KumoLiu:
> I don't think it would be better to have this handler added in multiple places. Alternatively can we put the `filterwarnings` calls into `__init__.py` in one place?

Actually I didn't understand why some of the warning message can be captured or shown by `warnings.showwarning` from the `__init__.py` I add, but some of them may not shown. So I'm not sure where we can put this handler, to my best of the knowledge, project-level __init__.py file should be be best place, but it can not capture all warning.

ericspod:
> Actually I didn't understand why some of the warning message can be captured or shown by `warnings.showwarning` from the `__init__.py` I add, but some of them may not shown. So I'm not sure where we can put this handler, to my best of the knowledge, project-level **init**.py file should be be best place, but it can not capture all warning.

It's possible other libraries are using other functions in `warnings` so bypasses the override you're using, such as `warn_explicit`. I'm trying to read through the source of the warnings module and there's a few things in there like that which could be used to raise warnings, but filtering should work with that.

KumoLiu:
Hi @ericspod, I discovered that the showwarning function has been modified, which is why some warnings are no longer displayed. I attempted to add a reset warning to the optional import, and set True in tensorrt and itk. Could you please help take another look at it. This may resolve your concern you mentioned before.

mingxin-zheng:
Hi @KumoLiu , I think the reason that you can't disable some warnings is partially because they're sent by logging.warning module.

This seems to be working to cancel out some of the warnings:

```python
import logging

# Create a custom filter
class DeprecatedTypesWarningFilter(logging.Filter):
    def filter(self, record):
        deprecated_aliases = ["np.bool8", "np.object0", "np.int0", "np.uint0", "np.void0", "np.str0", "np.bytes0"]
        for alias in deprecated_aliases:
            if alias in record.getMessage():
                return False
        return True

# Get the logger for warnings
logger = logging.getLogger("py.warnings")

# Create and add the filter to the logger
filter = DeprecatedTypesWarningFilter()
logger.addFilter(filter)

import monai

print(monai.__version__)
```

mingxin-zheng:
Another example why the filter doesn't work is because the library forces it:

https://github.com/pytorch/pytorch/blob/758d515d982a7fe00afe78060fecec3e55b72fe8/torch/distributed/optim/__init__.py#L30

KumoLiu:
Thanks @mingxin-zheng,I added it in the latest commit. A good catch then I don't need to locate where shown the warning. Thanks!

monai/__init__.py
```diff
@@ -13,9 +13,53 @@
...
-
+import logging
+import warnings
 from ._version import get_versions
 
+
+old_showwarning = warnings.showwarning
+
+
+def custom_warning_handler(message, category, filename, lineno, file=None, line=None):
```
mingxin-zheng, line 24: Another option to override maybe:

```
old_showwarning = warnings.showwarning

def custom_warning_handler(message, category, filename, lineno, file=None, line=None):
    ignore_files = [
        "ignite/handlers/checkpoint",
        "modelopt/torch/quantization/tensor_quant",
    ]
    if any(ignore in filename for ignore in ignore_files):
        with open(os.devnull, "w") as f:
            return old_showwarning(message, category, filename, lineno, file=f, line=line)
    return old_showwarning(message, category, filename, lineno, file, line)

warnings.showwarning = custom_warning_handler
```

monai/__init__.py
```diff
@@ -13,9 +13,53 @@
...
+    ignore_files = ["ignite/handlers/checkpoint", "modelopt/torch/quantization/tensor_quant"]
```
mingxin-zheng, line 25: I guess we can expect this is not needed after the packages are upgraded to fix these warnings. So please add some notes or TODO item here. Thanks!

mingxin-zheng (approved):
Overall it looks good to me. Thanks!

monai/__init__.py
```diff
@@ -13,9 +13,53 @@
...
+
+
+# workaround for https://github.com/Project-MONAI/MONAI/issues/8060
+# Set the custom warning handler to filter warning
+warnings.showwarning = custom_warning_handler
+# Get the logger for warnings
+logger = logging.getLogger("py.warnings")
+# Create and add the filter to the logger
+filter = DeprecatedTypesWarningFilter()
+logger.addFilter(filter)
```
mingxin-zheng, line 60: Maybe we can condense it to one line: 
```
logging.getLogger("py.warnings").addFilter(DeprecatedTypesWarningFilter())
```

ericspod:
> Another example why the filter doesn't work is because the library forces it:
> 
> https://github.com/pytorch/pytorch/blob/758d515d982a7fe00afe78060fecec3e55b72fe8/torch/distributed/optim/__init__.py#L30

This strikes me as just being really bad behaviour. The point of filtering things is so that one can suppress warnings after they've seen them so they don't continue to annoy, people will abuse that to sweep warnings under the rug but this is the same thing in mirror.

ericspod (approved):
I'm a little worried that we're having to do all this to shut up other warnings. Pytorch and others abuse the API of warnings I feel to force their messages to show, this is bad design but then we're doing the same thing here to counteract this. I hope this is all temporary until other dependencies update.

PR #8093, opened 2024-09-18
Add R2 score to metrics

### LINKED ISSUE #8085: Add Average Precision and R2 to the metrics
### Discussed in https://github.com/Project-MONAI/MONAI/discussions/8081

<sup>Originally posted by **thibaultdvx** September 13, 2024</sup>
Hi, 
I would like to suggest to add the [**Average Precision**](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.average_precision_score.html) in the classification metrics. I think that the implementation would be very similar to ROC AUC, so it wouldn't be a big deal to integrate it.
I would be happy to make a PR to do so.

I also feel that an important regression metric is missing:[ **R2 score**](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.r2_score.html). Is there any reason why?
Once again, I would be happy to help to implement it!

WDYT?</div>

thibaultdvx:
Fixes #8085.

### Description

I created:
- `R2Metric` and `compute_r2_score` in `monai.metrics`,
- a handler called `R2Score` in `monai.handlers`,
- three unittest modules: `test_compute_r2_score.py`, `test_handler_r2_score.py` and `test_handler_r2_score_dist.py`.

I also modified the docs to mention R2 score.

Note:  
- R2 score expects 1D or 2D data, i.e. I didn't allow the computation of R2 at pixel level in an image, like it can be done with the other regression metrics. 
- One needs all the data to compute R2 score (because of the mean), so tensors are simply accumulated and the metric computation is made once on all the data.

### Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,12 +18,13 @@
 
 MONAI is a [PyTorch](https://pytorch.org/)-based, [open-source](https://github.com/Project-MONAI/MONAI/blob/dev/LICENSE) framework for deep learning in healthcare imaging, part of the [PyTorch Ecosystem](https://pytorch.org/ecosystem/).
 Its ambitions are as follows:
+
 - Developing a community of academic, industrial and clinical researchers collaborating on a common foundation;
 - Creating state-of-the-art, end-to-end training workflows for healthcare imaging;
 - Providing researchers with the optimized and standardized way to create and evaluate deep learning models.
 
-
 ## Features
+
 > _Please see [the technical highlights](https://docs.monai.io/en/latest/highlights.html) and [What's New](https://docs.monai.io/en/latest/whatsnew.html) of the milestone releases._
 
 - flexible pre-processing for multi-dimensional medical imaging data;
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -53,6 +53,12 @@ ROC AUC metrics handler
     :members:
 
 
+Average Precision metric handler
+--------------------------------
+.. autoclass:: AveragePrecision
+    :members:
+
+
 Confusion matrix metrics handler
 --------------------------------
 .. autoclass:: ConfusionMatrix
--- a/docs/source/metrics.rst
+++ b/docs/source/metrics.rst
@@ -80,6 +80,13 @@ Metrics
 .. autoclass:: ROCAUCMetric
     :members:
 
+`Average Precision`
+-------------------
+.. autofunction:: compute_average_precision
+
+.. autoclass:: AveragePrecisionMetric
+    :members:
+
 `Confusion matrix`
 ------------------
 .. autofunction:: get_confusion_matrix
--- a/monai/handlers/__init__.py
+++ b/monai/handlers/__init__.py
@@ -11,6 +11,7 @@
 
 from __future__ import annotations
 
+from .average_precision import AveragePrecision
 from .checkpoint_loader import CheckpointLoader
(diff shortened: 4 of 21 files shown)


thibaultdvx:
hi @KumoLiu, could you please have a look at this PR?

tests/test_handler_r2_score.py
```diff
@@ -0,0 +1,41 @@
...
+
+import unittest
+
+import numpy as np
+import torch
+
+from monai.handlers import R2Score
+
+
+class TestHandlerR2Score(unittest.TestCase):
```
KumoLiu, line 22: Base on the ci :https://github.com/Project-MONAI/MONAI/actions/runs/10926409391/job/30330218578?pr=8093#step:8:13844

You need skip the tests if no ignite installed.

monai/metrics/r2_score.py
```diff
@@ -0,0 +1,184 @@
...
+    if not isinstance(p, int) or p < 0:
+        raise ValueError(f"`p` must be an integer larger or equal to 0, got {p}.")
+
+    return multi_output, p
+
+
+def _calculate(y_pred: np.ndarray, y: np.ndarray, p: int) -> float:
+    num_obs = len(y)
+    rss = np.sum((y_pred - y) ** 2)
+    tss = np.sum(y**2) - np.sum(y) ** 2 / num_obs
```
KumoLiu, line 114: Should be `tss = np.sum((y - np.mean(y)) ** 2)` here?

KumoLiu (commented):
Thanks for the PR, leave few comments inline.

monai/handlers/r2_score.py
```diff
@@ -0,0 +1,56 @@
...
+from collections.abc import Callable
+
+from monai.handlers.ignite_metric import IgniteMetricHandler
+from monai.metrics import R2Metric
+from monai.utils import MultiOutput
+
+
+class R2Score(IgniteMetricHandler):
+    """
+    Computes :math:`R^{2}` score accumulating predictions and the ground-truth during an epoch and applying `compute_r2_score`.
```
ericspod, line 23: As elsewhere it would help to explain where this is used and what it would be for. The math is explained in the function which is fine so no need here, just a high level idea.

monai/metrics/r2_score.py
thibaultdvx: Both formula works, but I agree that yours is clearer! I made the change

tests/test_handler_r2_score_dist.py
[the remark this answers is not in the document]
thibaultdvx: Done!

monai/handlers/r2_score.py
thibaultdvx: Hi @ericspod, tell me if the new version looks good to you!

tests/test_handler_r2_score.py
thibaultdvx: Done, I changed `min_tests.py`.

thibaultdvx:
@KumoLiu @ericspod ready for review

ericspod (approved):
I think this is good to go.

monai/handlers/r2_score.py
ericspod: I mean more to explain what the R2 means, what it would be used with (eg. compare classifications vs. segmentations), what it's value is vs other metrics. I would put this in the docstring for `R2Metric` however. Other than that this is fine here.

PR #8097, opened 2024-09-19
Fix RandomWeightedCrop for Integer Weightmap Handling

### LINKED ISSUE #7949: RandomWeightedCrop doesn't randomly generate crops for a weightmap  of integers
**Describe the bug**
monai.transforms.RandomWeightedCrop uses the function monai.transforms.utils.weighted_patch_samples. At line 576  in utils the  weighted_patch_samples function converts the datatype of the randomly generate number in the range of (0,1) to match the datatype of the input weightmap.
"""
        r, *_ = convert_to_dst_type(r_state.random(n_samples), v)
"""
If the weightmap consisted of integers then the random value will always be the integer of 0, which disables random sampling and instead samples from the first valid region continuously.  

**To Reproduce**
1. Create a numpy weightmap with integer datatype
2. Pass weightmap into RandomWeightedCrop
3. Observe the "crop_center" value in the meta data of the cropped image after multiple iterations it will remain constant

**Expected behavior**
The random factor "r" should never be an converted to an integer.

Han123su:
Fixes #7949 .

### Description
Regardless of the type of `weight map`, random numbers should be kept as floating-point numbers for calculating the sampling location. However, `searchsorted` requires matching data structures. I have modified `convert_to_dst_type` to control converting only the data structure while maintaining the original data type. Additionally, I have included an example with integer weight maps in the test file.

### Types of changes


monai/transforms/utils.py
```diff
@@ -582,7 +582,7 @@ def weighted_patch_samples(
     if not v[-1] or not isfinite(v[-1]) or v[-1] < 0:  # uniform sampling
         idx = r_state.randint(0, len(v), size=n_samples)
     else:
-        r, *_ = convert_to_dst_type(r_state.random(n_samples), v)
+        r, *_ = convert_to_dst_type(r_state.random(n_samples), v, structure_only=True)
```
ericspod, line 585:
```suggestion
        r_samples = r_state.random(n_samples)
        r, *_ = convert_to_dst_type(r_samples, v, dtype=r_samples.dtype)
```

Would this not be equivalent to what you're wanting to accomplish? I think `convert_to_dst_type` is getting rather complicated now with various options. In the docstring you provide with `structure_only` it's not clear what the implications are of using it, you could say more exlicitly that the dtype will be that of `src` if it has one or `dtype` if not (then the user has to realise the `dtype` will be taken from `dest` if this is `None`). If my suggestion works for you we should avoid adding this complexity.

Han123su: You’re right, this does have the same effect. I shouldn’t have made the function more complex. The current logic already meets the requirements, and I’ve made the necessary adjustments.

ericspod (approved):
I think this is fine as it is now, thanks.

KumoLiu (approved):
Thanks for the fix, LGTM.

PR #8112, opened 2024-09-25
Add multi-channel conversion transforms for brats23

dani-capellan:
Hi team!

I added some new transforms to convert from BraTS2023 labels to multi-channel - both when performing region-based and label-based training. I thought it could be useful because right now the only one existing in monai.transforms (`ConvertToMultiChannelBasedOnBratsClasses`) is based on BraTS2018 labels, which have changed over the past editions. If you find this useful I could also try to implement them for BraTS2024, although here we have more heterogenity across the different tasks.

New classes:
- `ConvertToMultiChannelBasedOnBrats23Classes`: returns labels -> regions -> multi-channel tensors. Simulates the previous mentioned class (ConvertToMultiChannelBasedOnBratsClasses)
- `ConvertToMultiChannelBasedOnBrats23ClassesNoReg`: returns labels -> multi-channel tensors

I've quickly tested the changes by running it in some experiments I'm working on, but I haven't run any integration tests yet. Please let me know if this is helpful. I'm happy to assist further with the implementation or help finalize the pull request.

### Types of changes


KumoLiu (commented):
Thanks for the PR!
Instead of introducing several new transforms here, I would suggest we can extend the existing one such as add an argument which can refer to what dataset you want to convert.
Also we may need to include the test for these update.
Does this makes sense to you? Thanks again!

PR #8118, opened 2024-09-27
Re-adding load_submodules 

ericspod:
### Description

This restores load_modules for now to resolve importation issues. Doing a two-pass loading process seems to allow some references which normally Python wouldn't permit. The fact that these issues weren't caught in testing is rather strange, and only showed up when symbols in `monai.apps` were references in bundles. I'm not sure if this is all related to circular imports, if parts of MONAI aren't being tested properly, or the way the symbol resolution works within `monai.bundle` should be improved. 

### Types of changes

### DIFF
--- a/monai/__init__.py
+++ b/monai/__init__.py
@@ -80,24 +80,28 @@ if sys.version_info.major != PY_REQUIRED_MAJOR or sys.version_info.minor < PY_RE
     )
 
 
-from . import (  # noqa: E402
-    apps,
-    auto3dseg,
-    bundle,
-    config,
-    data,
-    engines,
-    fl,
-    handlers,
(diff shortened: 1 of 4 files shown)


KumoLiu:
Let's also include this test back then let me trigger more tests.
https://github.com/Project-MONAI/MONAI/blob/1.4.0rc9/tests/min_tests.py#L235

ericspod:
> Let's also include this test back then let me trigger more tests.
> https://github.com/Project-MONAI/MONAI/blob/1.4.0rc9/tests/min_tests.py#L235

Ok that's back in now.

KumoLiu (approved):

PR #8123, opened 2024-09-30
Optimize VISTA3D

### LINKED ISSUE #8122: Optimize the VISTA3D latency
The VISTA3D is a very brilliant model for everything segmentation in a medical image. However it suffers a slowdown [issue](https://github.com/Project-MONAI/model-zoo/tree/dev/models/vista3d/docs#inference-gpu-benchmarks) when running on a GPU platform, which takes minutes to run a single inference. Therefore, a lower latency version of the VISTA3D model should be a user-friendly model and gain more usage.

1. Update the VISTA3D bundle parameters to achieve a better latency.
2. Update the VISTA3D network to solve some latency bottlenecks.
3. Try to accelerate the bundle with acceleration tools like TensorRT.

**Additional context**
Some analyzation has already done in [this PR](https://github.com/Project-MONAI/model-zoo/pull/671).
The corresponding MONAI core code change is included [here](https://github.com/binliunls/MONAI/pull/1)

binliunls:
Fixes #8122 .

### Description
As shown in [this PR](https://github.com/Project-MONAI/model-zoo/pull/671), the memory malloc and mask embedding for-loop are the bottlenecks that caused the vista3d slow inference. Therefore, this PR fixed them by adding the logic for malloc and replacing the for-loop with a tensor multiplication.

### Types of changes


borisfom:
@binliunls : Why wasn't  this PR integrated ?

binliunls:
> @binliunls : Why wasn't this PR integrated ?

Need a reviewer to confirm these changes are reasonale. Will contact with them.

monai/networks/nets/segresnet_ds.py
```diff
@@ -508,8 +508,10 @@ def forward(  # type: ignore
 
         outputs: list[torch.Tensor] = []
         outputs_auto: list[torch.Tensor] = []
-        x_ = x.clone()
+        x_ = x
         if with_point:
+            if with_label:
+                x_ = x.clone()
```
yiheng-wang-nv, line 514: Hi @binliunls , could you explain the changes here? Why remove deepcopy when with_label is False?

binliunls: The input `x` is cloned here because there are two branches for inference. Therefore, in order to make the inference result correct, we need two copies of input to go through difference branch. However, when we only have one branch, there is no need to maintain this copy. So I move this copy logic here to avoid invalid copy.

yiheng-wang-nv (approved):

yiheng-wang-nv:
Hi @heyufan1995 , this PR looks good to me, could you help to double confirm it?

KumoLiu:
Hi @binliunls, do we want include this change in 1.4?

yiheng-wang-nv:
I contacted with he vista3d author @heyufan1995 offline, and he confirmed that the changes look good

PR #8124, opened 2024-09-30
Update Dockerfile

kartikshastrakar:
correct relate update

Fixes # .

### Description

### Types of changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,60 +1,43 @@
-# Copyright (c) MONAI Consortium
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#     http://www.apache.org/licenses/LICENSE-2.0
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


ericspod:
Hi @kartikshastrakar thanks for the contribution but could you explain the rationale behind these changes? I don't think any of them should be accepted as they are, so if there's something you needed to accomplish here we can review and see what to do.

kartikshastrakar:
Hi ,

Actuall i was trying to change the dockerfile in it and had updated it

ericspod:
> also only one check is skipped else all are fine i have attached a snap because while using the monai ai i face some difficulties so i have changed

The other checks will be fine because the code base wasn't changed, only the Dockerfile. Was there something specific with the issues you were having? As I said I don't think these changes should be accepted as they are, we need to see what wasn't working for you and then if we can modify the Dockerfile without breaking it for others. For example, we shouldn't remove the license header, nor remove the aarch64 install in the first RUN section. If you wanted a modified Dockerfile for your own purposes only, I would suggest to keep it in your own fork and not to propose the PR here.

kartikshastrakar:
Ok, thanks for the clarification.

kartikshastrakar:
thank you for clarification

PR #8138, opened 2024-10-10
Modify Dice, Jaccard and Tversky losses

### LINKED ISSUE #8094: Jaccard, Dice and Tversky losses are incompatible with soft labels
**Describe the bug**
The Jaccard, Dice and Tversky losses are incompatible with soft labels [1, 2]. For example, with a ground truth value of 0.5 for a single pixel, the Dice loss is minimized when the predicted value is 1, which is clearly erroneous.

**To Reproduce**
```
import torch
from monai.losses.dice import DiceLoss
from monai.losses.tversky import TverskyLoss

torch.manual_seed(0)

B, C, H, W = 7, 5, 3, 2
input = torch.rand(B, C, H, W).softmax(1)
jaccard_loss = DiceLoss(jaccard=True, reduction='mean')
dice_loss = DiceLoss(reduction='mean')
tversky_loss = TverskyLoss(reduction='mean')

jaccard_loss_value = jaccard_loss(input, input)
dice_loss_value = dice_loss(input, input)
tversky_loss_value = tversky_loss(input, input)

print(jaccard_loss_value, dice_loss_value, tversky_loss_value)

# tensor(0.8817) tensor(0.7888) tensor(0.7888)
```

**Expected behavior**
When the input is equal to the target, the loss should be minimized and equals 0.  

**Environment**
```
Printing MONAI config...
MONAI version: 0+unknown
Numpy version: 1.26.4
Pytorch version: 2.2.2
MONAI rev id: 25589c377ac63d6be28ab9bb65dd8cb52f2bebdf
MONAI __file__: /Users/<username>/Desktop/loss/MONAI-dev/monai/__init__.py

Optional dependencies:
Pytorch Ignite version: 0.4.11
ITK version: 5.4.0
Nibabel version: 5.2.1
scikit-image version: 0.24.0
scipy version: 1.13.1
Pillow version: 10.4.0
Tensorboard version: 2.17.1
gdown version: 5.2.0
TorchVision version: 0.17.2
tqdm version: 4.66.5
lmdb version: 1.5.1
psutil version: 6.0.0
pandas version: 2.2.2
einops version: 0.8.0
transformers version: 4.40.2
mlflow version: 2.16.2
pynrrd version: 1.0.0
clearml version: 1.16.5rc0

Printing system config...
System: Darwin
Mac version: 10.16
Platform: macOS-10.16-x86_64-i386-64bit
Processor: i386
Machine: x86_64
Python version: 3.9.19
Process name: python3.9
Open files: []
Num physical CPUs: 8
Num logical CPUs: 8
Num usable CPUs: UNKNOWN for given OS
CPU usage (%): [55.2, 54.9, 31.8, 54.8, 12.8, 13.5, 7.7, 7.0]
CPU freq. (MHz): 2400
Load avg. in last 1, 5, 15 mins (%): [59.1, 62.0, 55.6]
Disk usage (%): 95.6
Total physical memory (GB): 16.0
Available memory (GB): 1.3
Used memory (GB): 1.9

Printing GPU config...
Num GPUs: 0
Has CUDA: False
cuDNN enabled: False
NVIDIA_TF32_OVERRIDE: None
```

**References**
[1] Dice Semimetric Losses: Optimizing the Dice Score with Soft Labels. Zifu Wang, Teodora Popordanoska, Jeroen Bertels, Robin Lemmens, Matthew B. Blaschko. *MICCAI 2023*.

[2] Jaccard Metric Losses: Optimizing the Jaccard Index with Soft Labels. Zifu Wang, Xuefei Ning, Matthew B. Blaschko. *NeurIPS 2023*.

zifuwanggg:
Fixes #8094.

### Description

The Dice, Jaccard and Tversky losses in `monai.losses.dice` and `monai.losses.tversky` are modified based on [JDTLoss](https://github.com/zifuwanggg/JDTLosses/blob/master/losses/jdt_loss.py) and [segmentation_models.pytorch](https://github.com/qubvel-org/segmentation_models.pytorch/blob/main/segmentation_models_pytorch/losses/_functional.py).

In the original versions, when `squared_pred=False`, the loss functions are incompatible with soft labels. For example, with a ground truth value of 0.5 for a single pixel, the Dice loss is minimized when the predicted value is 1, which is clearly erroneous. To address this, the intersection term is rewritten as $\frac{\|x\|_p^p + \|y\|_p^p - \|x-y\|_p^p}{2}$. When $p$ is 2 (`squared_pred=True`), this reformulation becomes the classical inner product: $\langle x,y \rangle$. When $p$ is 1 (`squared_pred=False`), the reformulation has been proven to retain equivalence with the original versions when the ground truth is binary (i.e. one-hot hard labels). Moreover, since the new versions are minimized if and only if the prediction is identical to the ground truth, even when the ground truth include fractional numbers, they resolves the issue with soft labels [1, 2].

In summary, there are three scenarios:
* [Scenario 1] $x$ is nonnegative and $y$ is binary: The new versions are the same as the original versions.
* [Scenario 2] Both $x$ and $y$ are nonnegative: The new versions differ from the original versions. The new versions are minimized if and only if $x=y$, while the original versions may not, making them incorrect.
* [Scenario 3] Either $x$ or $y$ is negative: The new versions differ from the original versions. The new versions are minimized if and only if $x=y$, while the original versions may not, making them incorrect.

Due to these differences, particularly in Scenarios 2 and 3, some tests fail with the new versions:
* The target is non-binary: `test_multi_scale`
* The input is negative: `test_dice_loss`, `test_tversky_loss`, `test_generalized_dice_loss`, `test_masked_loss`, `test_seg_loss_integration`

The failures in `test_multi_scale` are expected since the original versions are incorrectly defined for non-binary targets. Furthermore, because Dice, Jaccard, and Tversky losses are fundamentally defined over probabilities—which should be nonnegative—the new versions should not be tested against negative input or target values.

### Example
```
import torch
import torch.linalg as LA

torch.manual_seed(0)

b, c, h, w = 4, 3, 32, 32
dims = (0, 2, 3)

pred = torch.rand(b, c, h, w).softmax(dim=1)
soft_label = torch.rand(b, c, h, w).softmax(dim=1)
hard_label = torch.randint(low=0, high=c, size=(b, h, w))
one_hot_label = F.one_hot(hard_label, c).permute(0, 3, 1, 2).float()

def dice_old(x, y, ord, dims):
    cardinality = LA.vector_norm(x, ord=ord, dim=dims) ** ord + LA.vector_norm(y, ord=ord, dim=dims) ** ord
    intersection = torch.sum(x * y, dim=dims)
    return 2 * intersection / cardinality

def dice_new(x, y, ord, dims):
    cardinality = LA.vector_norm(x, ord=ord, dim=dims) ** ord + LA.vector_norm(y, ord=ord, dim=dims) ** ord
    difference = LA.vector_norm(x - y, ord=ord, dim=dims) ** ord
    intersection = (cardinality - difference) / 2
    return 2 * intersection / cardinality

print(dice_old(pred, one_hot_label, 1, dims), dice_new(pred, one_hot_label, 1, dims))
print(dice_old(pred, soft_label, 1, dims), dice_new(pred, soft_label, 1, dims))
print(dice_old(pred, pred, 1, dims), dice_new(pred, pred, 1, dims))

print(dice_old(pred, one_hot_label, 2, dims), dice_new(pred, one_hot_label, 2, dims))
print(dice_old(pred, soft_label, 2, dims), dice_new(pred, soft_label, 2, dims))
print(dice_old(pred, pred, 2, dims), dice_new(pred, pred, 2, dims))

# tensor([0.3345, 0.3310, 0.3317]) tensor([0.3345, 0.3310, 0.3317])
# tensor([0.3321, 0.3333, 0.3350]) tensor([0.8680, 0.8690, 0.8700])
# tensor([0.3487, 0.3502, 0.3544]) tensor([1., 1., 1.])

# tensor([0.4921, 0.4904, 0.4935]) tensor([0.4921, 0.4904, 0.4935])
# tensor([0.9489, 0.9499, 0.9503]) tensor([0.9489, 0.9499, 0.9503])
# tensor([1., 1., 1.]) tensor([1., 1., 1.])
```

### References
[1] Dice Semimetric Losses: Optimizing the Dice Score with Soft Labels. Zifu Wang, Teodora Popordanoska, Jeroen Bertels, Robin Lemmens, Matthew B. Blaschko. *MICCAI 2023*.

[2] Jaccard Metric Losses: Optimizing the Jaccard Index with Soft Labels. Zifu Wang, Xuefei Ning, Matthew B. Blaschko. *NeurIPS 2023*.

### Types of changes


ericspod:
Hi @zifuwanggg thanks for the contribution. I have an issue with this change in that the behaviour of the losses is very different now as seen in the CICD errors. I would instead suggest adding new loss functions in a "soft_losses.py" file or something like that instead of changing existing losses. Other uses may rely on existing behaviour, and in situations where non-binarises values are accidentally used due to incorrect postprocessing there is less feedback about the problem.

zifuwanggg:
Hi @ericspod, thank you for reviewing my code. While adding new loss functions as separate .py files could be a workaround, my concern is that this approach would lead to a lot of duplicated code, as the core differences are only in 2-3 lines.

Would it make sense to add an attribute to the existing loss classes and create a new helper function, so that the default behavior remains unchanged? Something like the following.

```
class DiceLoss(_Loss):
    def __init__(
        ...
        binary_label: bool = True,
    ):
        ...
        self.binary_label = binary_label

    def forward(...):
        ...
        f = compute_score(self.binary_label)
        ...


class GeneralizedDiceLoss(_Loss):
    def __init__(
        ...
        binary_label: bool = True,
    ):
        ...
        self.binary_label = binary_label
    
    def forward(...):
        ...
        f = compute_score(self.binary_label)
        ...


def compute_score(binary_label):
    if binary_label == True:
        ...
    else:
        ...
```

ericspod:
> Hi @ericspod, thank you for reviewing my code. While adding new loss functions as separate .py files could be a workaround, my concern is that this approach would lead to a lot of duplicated code, as the core differences are only in 2-3 lines.
> 

Hi @zifuwanggg I appreciate wanting to reduce duplicate code, we have too much of that in these loss functions as it stands so yes adding more isn't great. I think we can try to parameterise the loss functions in some way, either a function as you suggest or some other way, so long as the default behaviour is preserved. If you want to have a go at refactoring to do that we can return to it, I think in the future we do need to refactor all these loss functions to reduce duplication anyway.

zifuwanggg:
Hi @ericspod, I've created `losses/utils.py` and put a helper function that is shared by both `dice.py` and `tversky.py`.

Unit tests pass, but `mypy` tests fail. This seems related to #8161.

zifuwanggg:
Hi @ericspod, all CICD tests pass. @KumoLiu, thanks for the commit.

monai/losses/dice.py
```diff
@@ -89,6 +99,7 @@ def __init__(
                 of the sequence should be the same as the number of classes. If not ``include_background``,
                 the number of classes should not include the background category class 0).
                 The value/values should be no less than 0. Defaults to None.
+            soft_label: whether the target contains non-binary values or not
```
ericspod, line 102:
```suggestion
            soft_label: whether the target contains non-binary values (soft labels) or not. If True a soft label formulation of the loss will be used.
```
This clarifies a little bit I feel, the same should be done with the other modified losses.

ericspod (commented):
I have a minor comment about the check in the helper function being expensive to compute, but otherwise we do also need tests for soft labels to ensure that formulation of the losses works. I do want to get others to review this as well to be doubly sure the changes are compatible. Thanks again.

zifuwanggg:
Sorry for the late response.

I remove the costly check and modify the description of `soft_label` as you suggested. I also add some test cases. When `input` is the same as `target`, the loss value becomes zero when `soft_label=True`.

monai/losses/utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+def compute_tp_fp_fn(
+    input: torch.Tensor,
+    target: torch.Tensor,
+    reduce_axis: list[int],
+    ord: int,
+    soft_label: bool,
+    decoupled: bool = True,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    """
```
KumoLiu, line 26: Could you please also add docstring for these arguments?

KumoLiu (commented):
Thanks for the contribution! Overall looks good to me.
Jus leave one comment and please sign off to fix the DCO error.
https://github.com/Project-MONAI/MONAI/pull/8138/checks?check_run_id=33736878625

monai/losses/utils.py
zifuwanggg: Thanks for the comment. I've added some docstring and the DCO error is fixed.

KumoLiu: Thanks for the quick update.

KumoLiu (approved):

PR #8141, opened 2024-10-11
Modify MeanEnsemble to accumulate results in-place.

### LINKED ISSUE #6366: MeanEnsemble to accumulate results in-place
**Describe the bug**

https://github.com/Project-MONAI/MONAI/blob/825b8db9b20e0a85b2df327e3f4f23379fef7b2e/monai/transforms/post/array.py#L628-L636

the current implementation keeps all the predictions which take up unnecessary space compared with maintaining an output and updating it in-place.

staydelight:
Fixes #6366 .

### Description
According to the description in issue 6366, modify MeanEnsemble.

### Types of changes


monai/transforms/post/array.py
```diff
@@ -679,17 +679,21 @@ def __init__(self, weights: Sequence[float] | NdarrayOrTensor | None = None) ->
...
+        for i, pred in enumerate(img):
+            pred = torch.as_tensor(pred)
+            if out_pt is None:
+                out_pt = torch.zeros_like(pred)
+
+            if self.weights is not None:
+                weight = self.weights[i].to(pred.device)
 
-            img_ = img_ * weights / weights.mean(dim=0, keepdim=True)
+            out_pt += pred * weight
```
KumoLiu, line 693: If `self.weights` is None, this line will show error. You may need set a default value or move the computation in the for-loop.

monai/transforms/post/array.py
```diff
@@ -679,17 +679,21 @@ def __init__(self, weights: Sequence[float] | NdarrayOrTensor | None = None) ->
...
-        if self.weights is not None:
-            self.weights = self.weights.to(img_.device)
-            shape = tuple(self.weights.shape)
-            for _ in range(img_.ndimension() - self.weights.ndimension()):
-                shape += (1,)
-            weights = self.weights.reshape(*shape)
+        out_pt = None
+        total_weight = 0.0
+
+        for i, pred in enumerate(img):
```
KumoLiu, line 685: You may need check the the type for img what if it's a single array. Something like the change here: 
https://github.com/Project-MONAI/MONAI/blob/bae8f59a6d42f0264b083832e543bb671b07a918/monai/transforms/post/array.py#L639-L644

monai/transforms/post/array.py
```diff
@@ -679,17 +679,21 @@ def __init__(self, weights: Sequence[float] | NdarrayOrTensor | None = None) ->
...
+            pred = torch.as_tensor(pred)
+            if out_pt is None:
+                out_pt = torch.zeros_like(pred)
+
+            if self.weights is not None:
+                weight = self.weights[i].to(pred.device)
```
KumoLiu, line 691: Did you handle whether `self.weights` has the same shape with the img?

KumoLiu (commented):
Thanks for the PR, leave several comments.
And also please ensure the unittest pass, the error log shown here: https://github.com/Project-MONAI/MONAI/actions/runs/11289371448/job/31399089589?pr=8141#step:8:14508

PR #8142, opened 2024-10-11
Enable redirection of all loggers by configuring a FileHandler within the bundle

KumoLiu:
Fix https://github.com/Project-MONAI/model-zoo/issues/658, part of #7513

### Description
Enable redirection of all loggers by configuring a FileHandler within the bundle

### Types of changes


monai/apps/utils.py
```diff
@@ -54,20 +54,25 @@ def get_logger(
...
-    logger.propagate = False
     logger.setLevel(logging.INFO)
-    if adds_stdout_handler:  # don't add multiple stdout or add to the root
-        handler = logging.StreamHandler(sys.stdout)
-        formatter = logging.Formatter(fmt=fmt, datefmt=datefmt)
-        handler.setFormatter(formatter)
-        logger.addHandler(handler)
+
+    # Check if stdout handler should be added (avoid adding multiple times)
+    if module_name and module_name not in logging.root.manager.loggerDict:
```
Nic-Ma, line 67: Hi @ericspod ,

I am not very familiar with these python libs, could you please help double confirm it?

Thanks in advance.

Nic-Ma (approved):

monai/apps/utils.py
ericspod: This is different from the current code in that a `StreamHandler` is never created in this function. Previously, if the function is called and there wasn't a handler for the given module name, this would be created, but here we only try to find that handler and then set it's formatter if found. No code is now automatically creating the `StreamHandler` instance writing to stdout. I don't see how this solves the problem of there being things logged to stdout and not files (or vice versa). I have noticed this discrepency in the past and felt at some point we should have a logging module to keep better track of these sorts of things, and have a single logger rather than one per module which seems needless.

monai/apps/utils.py
```diff
@@ -54,20 +54,25 @@ def get_logger(
...
+        if fmt is not None or datefmt is not None:
+            stdout_handler = next((h for h in logging.root.handlers if isinstance(h, logging.StreamHandler)), None)
```
ericspod, line 69:
```suggestion
            stdout_handler = first(h for h in logging.root.handlers if isinstance(h, logging.StreamHandler))
```
`first` would also need to be imported: `from monai.utils import first`.

monai/apps/utils.py
ericspod: On another note, there's a number of things in this file that should go into their own places in `monai.utils`, such as moving the download functions to a `monai/utils/download.py` file. I was working some importation to clean up how modules are loaded but this broke some bundles, but I want to return to that to clean things up and reduce this file in size.

KumoLiu: Hi @ericspod, Let me briefly explain the changes I've made. When we use `logging.getLogger(module_name)` to create a logger, it gets registered under ``logging.root. Previously, we would check if the logger wasn't already in the root and, if not, add a `StreamHandler` while also formatting the output.

However, I noticed that we set `propagate = False`, which means the existing `StreamHandler` in the root won't be used. Now, I’ve adjusted the code by not setting this argument, which allows us to utilize the root's existing `StreamHandler` and only update the format if needed.

This change helps resolve the issue because in the bundle, we configure root as the logger. If we set `propagate = False`, logging won’t happen automatically, and we would need to manually configure loggers for each module (e.g., monai.apps).
https://github.com/Project-MONAI/model-zoo/blob/dev/models/brats_mri_segmentation/configs/logging.conf#L2

Hope I have clarified the change here.

ericspod: Hi @KumoLiu I get what you're wanting to do now thanks, however my other point on the logic being quite different stands in that there isn't a handler being created at the same times now as was the case in the previous version of this code. The conditions have been moved around and this effects the final logic, so I'd double check through tracing or some other way that the behaviour is the same.

monai/bundle/scripts.py
```diff
@@ -1969,7 +1968,7 @@ def create_workflow(
         workflow_ = workflow_class(**_args)
 
     workflow_.initialize()
-
+    _log_input_summary(tag="run", args=_args)
```
KumoLiu, line 1971: move it here then the logging summary can be logged by filehandler used in the bundle.

monai/bundle/workflows.py
```diff
@@ -316,8 +316,8 @@ def __init__(
                 else:
                     raise FileNotFoundError(f"Cannot find the logging config file: {logging_file}.")
             else:
-                logger.info(f"Setting logging properties based on config: {logging_file}.")
                 fileConfig(str(logging_file), disable_existing_loggers=False)
+                logger.info(f"Setting logging properties based on config: {logging_file}.")
```
KumoLiu, line 320: The same as above.

monai/apps/utils.py
KumoLiu: Hi @ericspod, I reverted the changes in get_logger after discovering that the issue with StatsHandler not logging through the FileHandler was due to the logger name not reflecting the correct hierarchy. I've now updated the logger name to monai.xxxx.stats_handlers, and it's being logged correctly.

please help review it again, thanks.

ericspod: Ok I see the changes that were done and I'm good with them, but did you not also want to remove the `propagate = False` part from the `get_logger` function?

KumoLiu: Yes, we don't need to remove that since for now all the logging logic can be handled by "monai" this root logger. So what we need to add in the logging.conf is some thing like:
Add a new logger named monai and also apply the filehandler to it, and set `propagate=0` in it.
```
[loggers]
keys=root, monai

[logger_monai]
level=INFO
handlers=consoleHandler, fileHandler
qualname=monai
propagate=0
```

KumoLiu: If I remove that from the `get_logger`, the logging output will be **duplicated** by default, since MONAI also adds a `StreamHandler`.

ericspod: Sounds good, thanks!

ericspod (approved):

PR #8146, opened 2024-10-12
8134 Add unit test for responsive inference

### LINKED ISSUE #8134: Add responsive inference support in bundle
Hi @ericspod , @KumoLiu ,

Recently, I got more and more feature requests to run bundle for real-time inference in MONAI Label, MONAI Deploy, and NVIDIA NIMs, etc.
There are 2 main blockers to support it:
1. Our current inference examples are for `batch inference`, for example:
https://github.com/Project-MONAI/model-zoo/blob/dev/models/spleen_ct_segmentation/configs/inference.json
We have lazy-instantiation for all the components in the config and pre-define all the datalist in the config.
But for real-time inference, we should instantiate all the python components defined in the config, and keep the model idle in GPU, then waiting for input data request.
Our current design can't change the config content once instantiated, because we do `deep copy` during parsing:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/config_parser.py#L347
I made a very hacky method to replace input data, it works but obviously not general for all bundles:
```py
ConfigWorkflow.parser.ref_resolver.items["dataset"].config["data"][0] = input_data
```
2. [Optional] 3rd party applications usually have their own `input` and `output` pipelines, they need to remove or replace the `LoadImage` and `SaveImage` transforms in the bundle config. We only have `MERGE_KEY`, missing the `delete` key:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/utils.py#L252 

Could you please help investigate this problem and make an ideal solution together?
It can be an important feature for MONAI 1.5.

Thanks in advance.

Nic-Ma:
Fixes #8134 .

### Description

This PR added unit test to cover the realtime inference with bundles.
And updated `BundleWorkflow` to support cyclically calling the `run` function with all components instantiated.

### Types of changes


ericspod:
Hi @Nic-Ma, we haven't had time to discuss the ideas around realtime/streaming inference yet due to MICCAI, but I feel we should look into doing this with specialised versions of datasets and transforms to represent streams of incoming and outgoing items. This would implement streaming without modifying existing workflows if we can get it performant.

Nic-Ma:
Hi @ericspod ,

Yes, I totally agree with you, I was thinking both the streaming way and repeatedly calling way.
Actually, for these 2 calling methods:
(1) pending `IterableDataset` in a separate thread for the real-time inference input streaming data.
(2) Repeatedly call `BundleWorkflow.run()` with replaced input data, just what I made the simple test case in this PR.

This PR is just a draft implementation for discussion, could you please help also write draft code for your idea and let's discuss later?
When you design the logic, please consider below typical use cases:
(1) `Triton backend`: https://github.com/triton-inference-server/python_backend?tab=readme-ov-file#usage
(2) `MONAI deploy`: https://github.com/Project-MONAI/monai-deploy-app-sdk/blob/main/monai/deploy/operators/monai_bundle_inference_operator.py#L277
(3) `Holoscan inference`: https://github.com/nvidia-holoscan/holohub/tree/main/tutorials/dicom_to_usd_with_monai_and_holoscan#monai-deploy--holoscan-pipeline

Thanks in advance.

monai/bundle/reference_resolver.py
```diff
@@ -192,6 +192,16 @@ def get_resolved_content(self, id: str, **kwargs: Any) -> ConfigExpression | str
         """
         return self._resolve_one_item(id=id, **kwargs)
 
+    def remove_resolved_content(self, id: str) -> Any | None:
```
ericspod, line 195: This may be a good addition regardless of the intent here.

monai/bundle/workflows.py
```diff
@@ -393,8 +393,15 @@ def check_properties(self) -> list[str] | None:
...
-    def _run_expr(self, id: str, **kwargs: dict) -> Any:
-        return self.parser.get_parsed_content(id, **kwargs) if id in self.parser else None
+    def _run_expr(self, id: str, **kwargs: dict) -> list[Any]:
+        ret = []
+        if id in self.parser:
+            # suppose all the expressions are in a list, run and reset the expressions
+            for i in range(len(self.parser[id])):
+                sub_id = f"{id}{ID_SEP_KEY}{i}"
+                ret.append(self.parser.get_parsed_content(sub_id, **kwargs))
+                self.parser.ref_resolver.remove_resolved_content(sub_id)
```
ericspod, line 403: This prevents keeping large objects created by the expression evaluation which aren't used anymore, so might be an optimisation in some cases.

tests/test_bundle_workflow.py
```diff
@@ -115,6 +119,35 @@ def test_inference_config(self, config_file):
         self._test_inferer(inferer)
         self.assertEqual(inferer.workflow_type, None)
 
+    @parameterized.expand([TEST_CASE_4])
+    def test_realtime_inference_config(self, config_file):
+        input_loader = LoadImaged(keys="image")
```
binliunls, line 124: Hi @Nic-Ma ,
   When using `LoadImage` , all the meta data is correct for the consequencial preprocessing. However, in a streaming environment, sometimes images are loading from other libraries, how could we ensure that all meta data is ready for using?

Thanks,  
Bin

monai/bundle/workflows.py
```diff
@@ -394,8 +394,15 @@ def check_properties(self) -> list[str] | None:
...
```
KumoLiu, line 404: Based on error here: https://github.com/Project-MONAI/MONAI/actions/runs/11946274965/job/33300343831?pr=8146#step:12:23028
If `self.parser[id]` is not a list, it will throw error when remove_resolved_content.
```suggestion
            if isinstance(self.parser[id], list):
                for i in range(len(self.parser[id])):
                    sub_id = f"{id}{ID_SEP_KEY}{i}"
                    ret.append(self.parser.get_parsed_content(sub_id, **kwargs))
                    self.parser.ref_resolver.remove_resolved_content(sub_id)
            else:
                ret.append(self.parser.get_parsed_content(id, **kwargs))
                self.parser.ref_resolver.remove_resolved_content(sub_id)
```

KumoLiu:
Hi @ericspod, if you don't have any further comments, I will try to merge this one.

monai/bundle/workflows.py
```diff
@@ -394,8 +394,19 @@ def check_properties(self) -> list[str] | None:
             ret.extend(wrong_props)
         return ret
 
-    def _run_expr(self, id: str, **kwargs: dict) -> Any:
-        return self.parser.get_parsed_content(id, **kwargs) if id in self.parser else None
+    def _run_expr(self, id: str, **kwargs: dict) -> list[Any]:
+        ret = []
```
ericspod, line 398:
```suggestion
    def _run_expr(self, id: str, **kwargs: dict) -> list[Any]:
        """
        Evaluate the expression or expression list given by `id`. The resolved values from the evaluations are not stored, allowing this to be evaluated repeatedly (eg. in streaming applications) without restarting the hosting process.
        """
        ret = []
```

ericspod (approved):

PR #8150, opened 2024-10-15
Apply pyupgrade fixes for Python 3.9+ syntax

jamesobutler:
### Description

Included is a commit here applying pyupgrade changes for Python 3.8+ syntax. This should have been included in https://github.com/Project-MONAI/MONAI/commit/104a360f953b5d720b3eb8f5b1a0546540e5e391 when Python 3.7 support was dropped.

Also included is a commit here apply pyupgrade changes for Python 3.9+ syntax. This should have been included in https://github.com/Project-MONAI/MONAI/commit/14b086b553693f5d344ff054f37d12ce6839da06 when Python 3.8 support was dropped.

I've also run the pre-commit autoupdate command to use the latest versions of the various pre-commit hook repos. It appears that pre-commit.ci bot has not been doing this on the quarterly schedule as expected? It appears the last time it submitted the PR to update the pre-commit hook repos was back in https://github.com/Project-MONAI/MONAI/pull/6286 from April 2023.

### Types of changes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ ci:
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.6.0
+    rev: v5.0.0
     hooks:
       - id: end-of-file-fixer
       - id: trailing-whitespace
@@ -26,15 +26,15 @@ repos:
         args: ['--autofix', '--no-sort-keys', '--indent=4']
       - id: end-of-file-fixer
       - id: mixed-line-ending
-  - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.3.5
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.7.0
     hooks:
     -   id: ruff
         args:
         - --fix
 
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.3.1
+    rev: v3.19.0
     hooks:
       - id: pyupgrade
         args: [--py37-plus]
@@ -52,7 +52,7 @@ repos:
         files: (?x)(^monai/networks/)
 
   - repo: https://github.com/asottile/yesqa
-    rev: v1.4.0
+    rev: v1.5.0
     hooks:
       - id: yesqa
         name: Unused noqa
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
     rev: v3.19.0
     hooks:
       - id: pyupgrade
-        args: [--py37-plus]
+        args: [--py38-plus]
         name: Upgrade code excluding monai networks
         exclude: |
           (?x)(
@@ -49,7 +49,7 @@ repos:
               ^monai/data/grid_dataset.py
           )
       - id: pyupgrade
-        args: [--py37-plus, --keep-runtime-typing]
+        args: [--py38-plus, --keep-runtime-typing]
         name: Upgrade monai networks
         files: (?x)(
                    ^monai/networks/|
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,25 +37,16 @@ repos:
     rev: v3.19.0
     hooks:
       - id: pyupgrade
-        args: [--py38-plus]
-        name: Upgrade code excluding monai networks
+        args: [--py38-plus, --keep-runtime-typing]
+        name: Upgrade code with exceptions
         exclude: |
           (?x)(
               ^versioneer.py|
               ^monai/_version.py|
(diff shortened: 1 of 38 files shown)


KumoLiu:
Hi @jamesobutler, thanks for the PR.
Could you please help take a look at the ci error: https://github.com/Project-MONAI/MONAI/actions/runs/11355141675/job/31583909844?pr=8150#step:8:13833

jamesobutler:
I've attempted to avoid the typing related changes, but yeah I'm not sure what is going here with the code not being able to handle the post-upgrade.

.pre-commit-config.yaml
```diff
@@ -26,33 +26,28 @@ repos:
...
       - id: pyupgrade
-        args: [--py37-plus]
-        name: Upgrade code excluding monai networks
+        args: [--py39-plus, --keep-runtime-typing]
+        name: Upgrade code with exceptions
         exclude: |
           (?x)(
               ^versioneer.py|
-              ^monai/_version.py|
-              ^monai/networks/|  # no PEP 604 for torchscript tensorrt
```
KumoLiu, line 46: Will revert this two lines help?

jamesobutler: No. There was previously two runs of upgrade specified in the config yaml where the first exclude those files and then the second included only those files with the “keep-runtime-typing” flag used. After observing the same issues I attempted removing the exception and running pyupgrade once with “keep-runtime-typing” applied for all files.

monai/networks/nets/dynunet.py
```diff
@@ -32,7 +33,7 @@ class DynUNetSkipLayer(nn.Module):
     forward passes of the network.
     """
 
-    heads: Optional[List[torch.Tensor]]
+    heads: Optional[list[torch.Tensor]]
```
KumoLiu, line 36: A feasible way may be revert the change in this file or try to set `Union[None, List[Tensor]]` due to TorchScript has limited support for type annotations, especially when dealing with complex types, such as nested generic types.

jamesobutler:
I have rebased this branch against latest `dev` so that mass bulk changes would apply to latest code.

I don't generally like large file exclusions, but I've excluded some files from the pyupgrade pre-commit action and as a result monai tests have begun to pass again. This PR now represents an improvement and removal of the pyupgrade exceptions can come in the future along with more debugging. Maybe the issue will resolve itself upon dropping support for Python 3.9 and the typing support that was available in that version.

.pre-commit-config.yaml
```diff
@@ -26,33 +26,30 @@ repos:
...
-        name: Upgrade code excluding monai networks
+        args: [--py39-plus, --keep-runtime-typing]
+        name: Upgrade code with exceptions
         exclude: |
           (?x)(
               ^versioneer.py|
               ^monai/_version.py|
               ^monai/networks/|  # no PEP 604 for torchscript tensorrt
-              ^monai/losses/  # no PEP 604 for torchscript tensorrt
+              ^monai/apps/detection/utils/anchor_utils.py|
```
KumoLiu, line 47: Could you please also add a comment here for next time we know why we are here and remind us to update it in the future? Thanks!

KumoLiu (approved):

jamesobutler:
@KumoLiu Should the GPU-quick-py3 still be a required status check for the protected `dev` branch or should it be removed? It appears it was set to not run as of https://github.com/Project-MONAI/MONAI/commit/18a671a7f12389dd8263a4c2232196dcca814bb9 which was well over a year ago. It is listed as a required status check, but it has been constantly been getting skipped (https://github.com/Project-MONAI/MONAI/actions/workflows/pythonapp-gpu.yml). It seems like the pythonapp-gpu.yml could be removed from the source repository so that it no longer runs or at least modified to not run on pull request opened.

https://github.com/Project-MONAI/MONAI/actions/runs/11500638643/workflow?pr=8150

{C7002F56-F6BF-4A02-86BF-4996E8628D63}

KumoLiu:
> @KumoLiu Should the GPU-quick-py3 still be a required status check for the protected `dev` branch or should it be removed? 

Hi @jamesobutler, yes it's required and the workflow has been move to the blossom, so it is run in this pipeline: https://github.com/Project-MONAI/MONAI/actions/runs/11502209707.

jamesobutler:
@KumoLiu I see that blossom-ci is a required status check, but it doesn't seem like GPU-quick-py3 should be set as a required status check for a PR. Enforcing all status checks to pass prior to integrating into a protected branch would require it to be constantly overruled by an admin enforcement since the rule isn't being respected.

{1F186E76-66BE-4DC9-8AB7-B4915D27BB28}

KumoLiu:
@jamesobutler, Thanks for pointing out, I just updated that part, seems it's out of the date.

PR #8151, opened 2024-10-15
Add PythonicWorkflow

KumoLiu:
Fixes # .

### Description

Add PythonicWorkflow

### Types of changes


monai/bundle/workflows.py
```diff
@@ -225,6 +225,115 @@ def check_properties(self) -> list[str] | None:
...
 
 
+class PythonicWorkflow(BundleWorkflow):
+    """
+    Base class for the pythonic workflow specification in bundle, it can be a training, evaluation or inference workflow.
+    It defines the basic interfaces for the bundle workflow behavior: `initialize`, `run`, `finalize`, etc.
+    And also provides the interface to get / set public properties to interact with a bundle workflow.
+
+    Args:
+        workflow_type: specifies the workflow type: "train" or "training" for a training workflow,
```
yiheng-wang-nv, line 235: I think for pythonic workflow, we can also enable read config file contents.
It's common to prepare a config file to save hyper parameters, otherwise it's not convenient to adjust them.

yiheng-wang-nv: we can discuss offline about it

Nic-Ma:
Hi @KumoLiu ,

Could you please add some unit tests and maybe we can start reviewing and merging this PR?

Thanks.

monai/bundle/workflows.py
KumoLiu: added in the latest change, please help review it again, thanks.

monai/bundle/workflows.py
```diff
@@ -226,6 +242,138 @@ def check_properties(self) -> list[str] | None:
...
+
+        Args:
+            name: the name of target property.
+            property: other information for the target property, defined in `TrainProperties` or `InferProperties`.
+        """
+        value = None
+        id = self.properties.get(name, None).get(BundlePropertyConfig.ID, None)
+        if name in self._set_props:
+            value = self._set_props[name]
+            self._props[name] = value
```
ericspod, line 346: If `name` is in `_set_props` the value will always be queried from `_set_props` and never from `_props`. Either that's intended and so this line doesn't accomplish anything, or `name` should be removed from `_set_props` here.

ericspod: Unless I've missed something?

monai/bundle/workflows.py
```diff
@@ -226,6 +242,138 @@ def check_properties(self) -> list[str] | None:
...
+        elif name in self._props:
+            value = self._props[name]
+        elif name in self.parser.config[self.parser.meta_key]:  # type: ignore[index]
+            value = self.parser.__getitem__(id)
```
ericspod, line 350:
```suggestion
            value = self.parser[id]
```
This is equivalent here?

monai/bundle/workflows.py
```diff
@@ -226,6 +242,138 @@ def check_properties(self) -> list[str] | None:
...
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def finalize(self, *args: Any, **kwargs: Any) -> Any:
+        """
+        Finalize step after the running of bundle workflow.
+
+        """
+        raise NotImplementedError()
```
ericspod, line 329: These are inherited from `BundleWorkflow` and so don't need to be implemented again if they're identical.

monai/bundle/workflows.py
```diff
@@ -226,6 +242,138 @@ def check_properties(self) -> list[str] | None:
...
+        If the property is already generated, return from the bucket directly.
+        If user explicitly set the property, return it directly.
+        Otherwise, generate the expected property as a class private property with prefix "_".
+
+        Args:
+            name: the name of target property.
+            property: other information for the target property, defined in `TrainProperties` or `InferProperties`.
+        """
+        value = None
+        id = self.properties.get(name, None).get(BundlePropertyConfig.ID, None)
```
ericspod, line 343: This line is only needed in the body of the `elif` clause on line 349.

monai/bundle/workflows.py
```diff
@@ -226,6 +242,138 @@ def check_properties(self) -> list[str] | None:
         return [n for n, p in self.properties.items() if p.get(BundleProperty.REQUIRED, False) and not hasattr(self, n)]
 
 
+class PythonicWorkflow(BundleWorkflow):
+    """
+    Base class for the pythonic workflow specification in bundle, it can be a training, evaluation or inference workflow.
+    It defines the basic interfaces for the bundle workflow behavior: `initialize`, `run`, `finalize`, etc.
+    And also provides the interface to get / set public properties to interact with a bundle workflow.
```
ericspod, line 249:
```suggestion
    This also provides the interface to get / set public properties to interact with a bundle workflow through defined `get_<property>` accessor methods or directly defining members of the object.
```
There should be something said about this behaviour which is only seen in the body of `_get_property`.

monai/bundle/workflows.py
```diff
@@ -44,12 +44,14 @@ class BundleWorkflow(ABC):
...
-            default to `train` for train workflow.
+            default to `None` for only using meta properties.
         workflow: specifies the workflow type: "train" or "training" for a training workflow,
             or "infer", "inference", "eval", "evaluation" for a inference workflow,
             other unsupported string will raise a ValueError.
             default to `None` for common workflow.
-        properties_path: the path to the JSON file of properties.
+        properties_path: the path to the JSON file of properties. If workflow type is specified, the properties
+            will be loaded from the file based on the workflow type and meta, otherwise, the properties will be loaded from
+            "train". If the file does not exist, the default properties will be used.
```
ericspod, line 54: We should explain here what these properties are, that is they specify what the components of the bundle are such as the definitions that would be found in the config files for a config workflow. We should mention what the default properties are, that is it's those in `monai/bundle/properties.py`.

monai/bundle/workflows.py
```diff
@@ -226,6 +242,138 @@ def check_properties(self) -> list[str] | None:
...
+        meta_file = str(Path(os.getcwd()) / "metadata.json") if meta_file is None else meta_file
+        super().__init__(
+            workflow_type=workflow_type,
+            workflow=workflow,
+            properties_path=properties_path,
+            meta_file=meta_file,
+            logging_file=logging_file,
+        )
+        self._props: dict = {}
+        self._set_props: dict = {}
```
ericspod, line 299: It seemed a little unclear to me what the difference is between `self.properties` and `self._props` just from the names. The `_props` dictionary stores the property values, that is an entry in `properties` states that a property with a certain name exists and what it does, then the value for it is in `_props`. Perhaps a different name for this such as `_prop_vals` or something would help (and change `_set_props` as well)?

monai/bundle/workflows.py
KumoLiu: Hi @ericspod, the `_set_props` won't be re-initialized once the workflow is initialized, meaning that `_props` will be completely removed. As a result, this line doesn’t serve any purpose unless users want to review all currently used properties.
Do you think we should remove this line?
also cc @Nic-Ma

monai/bundle/workflows.py
KumoLiu: Yes, thanks!

monai/bundle/workflows.py
```diff
@@ -226,6 +242,138 @@ def check_properties(self) -> list[str] | None:
...
+    supported_train_type: tuple = ("train", "training")
+    supported_infer_type: tuple = ("infer", "inference", "eval", "evaluation")
+
+    @deprecated_arg(
+        "workflow",
+        since="1.2",
+        removed="1.5",
+        new_name="workflow_type",
+        msg_suffix="please use `workflow_type` instead.",
+    )
```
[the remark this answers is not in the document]
KumoLiu: Thanks for pointing this out, updated.

monai/bundle/workflows.py
KumoLiu: Moved, thanks.

monai/bundle/workflows.py
KumoLiu: Update the docstring, please help review it again, thanks.

monai/bundle/workflows.py
KumoLiu: Agreed, I have updated the name based on your comments, please help review it again, thanks.

monai/bundle/workflows.py
ericspod: As the code stands `_set_props` has priority, so if the `name` is in both it'll never be looked up in `_props` so this line isn't needed. Users can look at both dictionaries to see what the state is I suppose.

monai/bundle/workflows.py
ericspod: I think that's better thanks.

monai/bundle/workflows.py
KumoLiu: Removed it in the latest commit, let's see whether @Nic-Ma has more concern here. Thanks.

tests/test_bundle_workflow.py
```diff
@@ -164,6 +167,29 @@ def test_non_config_wrong_log_cases(self, meta_file, logging_file, expected_erro
         with self.assertRaisesRegex(FileNotFoundError, expected_error):
             NonConfigWorkflow(self.filename, self.data_dir, meta_file, logging_file)
 
+    def test_pythonic_workflow(self):
+        device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        config_file = {"roi_size": (64, 64, 32)}
+        meta_file = os.path.join(os.path.dirname(__file__), "testing_data", "metadata.json")
+        workflow = PythonicWorkflowImpl(workflow_type="infer", config_file=config_file, meta_file=meta_file)
```
binliunls, line 174: Can we use the `create_workflow` API to create a pythonic workflow? If so, could you please provide an example here?

Thanks,  
Bin

tests/test_bundle_workflow.py
```diff
@@ -164,6 +167,29 @@ def test_non_config_wrong_log_cases(self, meta_file, logging_file, expected_erro
...
+        workflow.initialize()
+        # Load input data
+        input_loader = LoadImaged(keys="image")
+        workflow.dataflow.update(input_loader({"image": self.filename}))
```
binliunls, line 178: Does this `workflow.dataflow.update` suitable for all kinds of workflows? I think we should enable all type workflows with this streaming IO.

Thanks,  
Bin

monai/bundle/workflows.py
```diff
@@ -226,6 +246,114 @@ def check_properties(self) -> list[str] | None:
...
+
+        Args:
+            name: the name of target property.
+            property: other information for the target property, defined in `TrainProperties` or `InferProperties`.
+            value: value to set for the property.
+
+        """
+        self._set_props_vals[name] = value
+
+
```
binliunls, line 356: Cannot instantiate this class directly since it does not have the `run` and `finalize` methods as a abstract class.

tests/test_bundle_workflow.py
KumoLiu: For config workflow, it should be fixed in https://github.com/Project-MONAI/MONAI/pull/8146/files.

monai/bundle/workflows.py
```diff
@@ -44,12 +44,18 @@ class BundleWorkflow(ABC):
...
             other unsupported string will raise a ValueError.
             default to `None` for common workflow.
-        properties_path: the path to the JSON file of properties.
+        properties_path: the path to the JSON file of properties. If `workflow_type` is specified, properties will be
+            loaded from the file based on the provided `workflow_type` and meta. If no `workflow_type` is specified,
+            properties will default to loading from "train". If the specified file is unavailable, default properties
+            will be sourced from "monai/bundle/properties.py" based on the workflow_type:
+            For a training workflow, properties load from `TrainProperties` and `MetaProperties`.
+            For a inference workflow, properties load from `InferProperties` and `MetaProperties`.
+            For workflow_type = None : only `MetaProperties` will be loaded.
```
Nic-Ma, line 58: I think if `properties_path` is not None and file is unavailable, we should raise error directly.
And if no `workflow_type` is specified, properties should not default to loading from "train", I would suggest the same behavior as `workflow_type` (default to `None` for only using meta properties)

monai/bundle/workflows.py
```diff
@@ -226,6 +246,114 @@ def check_properties(self) -> list[str] | None:
...
+        With specified property name and information, set value for the expected property.
+        Stores user-reset initialized objects that should not be re-initialized.
+
+        Args:
+            name: the name of target property.
+            property: other information for the target property, defined in `TrainProperties` or `InferProperties`.
+            value: value to set for the property.
+
+        """
+        self._set_props_vals[name] = value
```
Nic-Ma, line 354: After `set property`, I think we should force to `initialize` again, right?
Refer to: https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/workflows.py#L437

Thanks.

monai/bundle/workflows.py
```diff
@@ -226,6 +246,114 @@ def check_properties(self) -> list[str] | None:
         return [n for n, p in self.properties.items() if p.get(BundleProperty.REQUIRED, False) and not hasattr(self, n)]
 
 
+class PythonicWorkflow(BundleWorkflow):
+    """
+    Base class for the pythonic workflow specification in bundle, it can be a training, evaluation or inference workflow.
+    It defines the basic interfaces for the bundle workflow behavior: `initialize`, `run`, `finalize`, etc.
+    This also provides the interface to get / set public properties to interact with a bundle workflow through
+    defined `get_<property>` accessor methods or directly defining members of the object.
```
Nic-Ma, line 254: Please also describe how to `set` properties.
And better to talk about the lifecycle of properties with some examples: set, initialize, get, re-set, initialize, add property, etc.
Add property: https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/workflows.py#L200
Refer to this example for all the usage cases:
https://github.com/Project-MONAI/tutorials/blob/main/bundle/python_bundle_workflow/scripts/train.py#L84

Thanks.

tests/test_bundle_workflow.py
KumoLiu: Thanks, I added related test, please help review again.

monai/bundle/workflows.py
KumoLiu: Hi @Nic-Ma, I have updated to load meta properties if workflow_type is None, please help review it again, thanks.

monai/bundle/workflows.py
KumoLiu: Update this part, please help review it again, thanks.

monai/bundle/workflows.py
KumoLiu: Add more description in the docstring and in the test case, please take a look at it, thanks.

monai/bundle/workflows.py
Nic-Ma: As my previous comment, I think if properties_path is not None and file is unavailable, we should raise error directly. What do you think?

Thanks.

KumoLiu: Hi @Nic-Ma, yes. The logic is here:
https://github.com/Project-MONAI/MONAI/blob/ec202e0c75d6008326c055ac8df013405167ad29/monai/bundle/workflows.py#L116

KumoLiu:
Hi @ericspod @Nic-Ma @binliunls @yiheng-wang-nv, do you have any further comments on this PR?

ericspod (approved):
I think we're good now, let's get people using this and see about feedback. Thanks!

Nic-Ma (approved):

PR #7598, opened 2024-03-31
implementation of masked autoencoder as a monai network

Lucas-rbnt:
### Description

Implementation of the Masked Autoencoder as described in the paper: [Masked Autoencoders Are Scalable Vision Learners](https://arxiv.org/pdf/2111.06377.pdf) from Kaiming et al.

Masked autoencoders are a type of deep learning model that learn to reconstruct input data from partially masked or obscured versions of that data. They are structured to first "mask" or remove parts of the input image, and then attempt to reconstruct the missing pieces based solely on the available (unmasked) information. 

This allows us to be much faster with fewer tokens being passed through the encoder. In addition, most representation learning methods are based on augmentations or different views of the same image and the quality of the representations depends heavily on the augmentations in question, which can be restrictive in the context of medical imaging. Masked autoencoders go some way towards overcoming this problem.
Its effectiveness has already been demonstrated in the literature for medical tasks in the paper [Self Pre-training with Masked Autoencoders for Medical Image Classification and Segmentation](https://arxiv.org/abs/2203.05573).

The PR concerns the implementation of this method and the associated tests (note currently the tests take 48seconds to pass that might be a bit long tell me).

So far 2D training on CIFAR data 
cifar_mae
went well

and 3d training on BraTS2021 data yields this
brain_mae

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [x] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [x] Quick tests passed locally by running `./runtests.sh --quick --unittests  --disttests`.
- [x] In-line docstrings updated.
- [x] Documentation updated, tested `make html` command in the `docs/` folder.

PR #8152, opened 2024-10-15
Implementation of a Masked Autoencoder for representation learning

Lucas-rbnt:
This follows a previous PR (#7598).

In the previous PR, the official implementation was under a non-compatible license. This is a clean-sheet implementation I developed. The code is fairly straightforward, involving a transformer, encoder, and decoder. The primary changes are in how masks are selected and how patches are organized as they pass through the model.

In the official masked autoencoder implementation, noise is first generated and then sorted twice using `torch.argsort`. This rearranges the tokens and identifies which ones are retained, ultimately selecting only a subset of the shuffled indices.

In our implementation, we use `torch.multinomial` to generate mask indices, followed by simple boolean indexing to manage the sub-selection of patches for encoding and the reordering with mask tokens in the decoder.

**Let me know if you need a detailed, line-by-line explanation of the new code, including how it works and how it differs from the previous version.**
### Description

Implementation of the Masked Autoencoder as described in the paper: [Masked Autoencoders Are Scalable Vision Learners](https://arxiv.org/pdf/2111.06377.pdf) from Kaiming et al.

Its effectiveness has already been demonstrated in the literature for medical tasks in the paper [Self Pre-training with Masked Autoencoders for Medical Image Classification and Segmentation](https://arxiv.org/abs/2203.05573).
The PR contains the architecture and associated unit tests.

**Note:** The output includes the prediction, which is a tensor of size: ($BS$, $N_{tokens}$, $D$), and the associated mask ($BS$, $N_{tokens}$). The mask is used to apply loss only to masked patches, but I'm not sure it's the “best” output format, what do you think?

### Types of changes


ericspod:
Hi @Lucas-rbnt thanks for the effort on this followup PR. @atbenmurray could you please re-review the content here?

atbenmurray:
@Lucas-rbnt @atbenmurray I shall do so

monai/networks/nets/masked_autoencoder_vit.py
```diff
@@ -0,0 +1,221 @@
...
+            >>> net = MaskedAutoEncoderViT(in_channels=3, img_size=128, patch_size=16, pos_embed_type='learnable')
+            # for 3-channel with image size of (224,224) and a masking ratio of 0.25
+            >>> net = MaskedAutoEncoderViT(in_channels=3, img_size=(224,224), patch_size=(16,16), masking_ratio=0.25,
+            spatial_dims=2)
+        """
+
+        super().__init__()
+
+        if not (0 <= dropout_rate <= 1):
+            raise ValueError("dropout_rate should be between 0 and 1.")
```
KumoLiu, line 97: Could you please also mention the value in the error message which will show more informations for the users? The same as below. Thanks!
Something like:
```suggestion
            raise ValueError(f"dropout_rate should be between 0 and 1, got {dropout_rate}.")
```

KumoLiu (commented):
Thanks for the PR.

> In the official masked autoencoder implementation, noise is first generated and then sorted twice using torch.argsort. This rearranges the tokens and identifies which ones are retained, ultimately selecting only a subset of the shuffled indices.
In our implementation, we use torch.multinomial to generate mask indices, followed by simple boolean indexing to manage the sub-selection of patches for encoding and the reordering with mask tokens in the decoder.

As you mentioned here, I wonder have you verified that whether there is a big difference between the two different implementations? Does it have any impact on the final performance? Thanks.

tests/test_masked_autoencoder_vit.py
```diff
@@ -0,0 +1,260 @@
...
+                img_size=(96, 96, 96),
+                patch_size=(16, 16, 16),
+                hidden_size=768,
+                mlp_dim=3072,
+                num_layers=12,
+                num_heads=12,
+                proj_type="conv",
+                dropout_rate=0.3,
+                masking_ratio=-0.1,
+            )
```
ericspod, line 200: This could be implemented by putting the arguments of each construction into a list and using parameterized.expand.

monai/networks/nets/masked_autoencoder_vit.py
Lucas-rbnt: @KumoLiu 
Concerning the request for comparison: I've carried out an experiment on CIFAR-10 in 2D with 1000 pre-training epochs.

Here are the validation losses obtained and an example reconstruction: (left) this implementation, (right) official implementation.
Capture d’écran 2024-10-23 à 10 58 37
Capture d’écran 2024-10-23 à 10 59 06


For a comparison in terms of fine-tuning accuracy and computational performance, see the table below.

|                                | This implem. | Official implem. |
|:------------------------------:|:------------:|:----------------:|
|   N parameters (in millions)   |     15.60    |       15.63      |
|      GPU consumption (MiB)     |     8733     |       9185       |
|      Training Time (hours)     |       9      |         8        |
| 5 epoch Fine-tuning  accuracy |     0.86     |       0.86       |

The same training from scratch gives 0.6 accuracy.

Note that what changes a little is the calculation time, which is longer with this implementation. Indeed, in the official implementation, token selection via shuffling + removing the last part is extremely fast, whereas here boolean indexing is a little slower.
Knowing that this selection is really the part apart from the official code, I haven't kept it for licensing reasons, but a similar process coded differently could be envisaged.
Although, I don't know whether this would be acceptable in terms of license compatibility.

tests/test_masked_autoencoder_vit.py
Lucas-rbnt: Yes, thanks for the comment, I've corrected it now in the new push!

monai/networks/nets/masked_autoencoder_vit.py
```diff
@@ -0,0 +1,221 @@
...
+            dropout_rate=dropout_rate,
+            spatial_dims=self.spatial_dims,
+        )
+        self.blocks = nn.ModuleList(
+            [
+                TransformerBlock(hidden_size, mlp_dim, num_heads, dropout_rate, qkv_bias, save_attn)
+                for _ in range(num_layers)
+            ]
+        )
+        self.norm = nn.LayerNorm(hidden_size)
```
ericspod, line 137:
```suggestion
        blocks = [
            TransformerBlock(hidden_size, mlp_dim, num_heads, dropout_rate, qkv_bias, save_attn)
            for _ in range(num_layers)
        ]
        self.blocks = nn.Sequential(*blocks, nn.LayerNorm(hidden_size))
```
This is a suggested simplication you can ignore if you want. 

Then in the `forward` method this:

```python
        for blk in self.blocks:
            x = blk(x)
        x = self.norm(x)
```
becomes:
```python
        x=self.blocks(x)
````

The same can be done for `self.decoder_blocks`, in that this becomes a `Sequential` with `self.decoder_norm` and `self.decoder_pred` being removed and those layers added to the end.

ericspod:
I think this is fine now though the comments should be looked at the conflict resolved, then we can trigger the blossom tests. Thanks!

monai/networks/nets/masked_autoencoder_vit.py
Lucas-rbnt: Yes no problem, it simplifies things a bit, I added it in the last commit.

ericspod (approved):
I'm good with this and look forward to an example notebook in Tutorials demonstrating its use!

Issue #8149, opened 2024-10-14
MyPy issue


KumoLiu:
```
mypy 1.12.0 (compiled: yes)
monai/data/utils.py:562:17: error: Item "ItemsView[Any, Any]" of "Union[Iterator[tuple[Any, Any]], ItemsView[Any, Any], enumerate[Any]]" has no attribute "__next__"  [union-attr]
monai/transforms/utils.py:1130:25: error: Incompatible types in assignment (expression has type "Union[int, float, Any]", target has type "int")  [assignment]
monai/transforms/utils.py:1131:23: error: Incompatible types in assignment (expression has type "Union[int, float, Any]", target has type "int")  [assignment]
monai/transforms/croppad/functional.py:194:17: error: Incompatible types in assignment (expression has type "Union[Any, Size]", variable has type "list[Any]")  [assignment]
monai/transforms/utility/dictionary.py:738:23: error: Item "Hashable" of "Union[Any, Collection[Hashable], Hashable]" has no attribute "__iter__" (not iterable)  [union-attr]
monai/transforms/io/array.py:503:30: error: "str" not callable  [operator]
monai/transforms/io/array.py:507:35: error: Incompatible types in assignment (expression has type "Union[Any, ImageWriter]", variable has type "None")  [assignment]
monai/metrics/utils.py:206:20: error: Incompatible return value type (got "Union[tuple[Any, Any], tuple[Any, Any, Any, Any]]", expected "tuple[NdarrayTensor, NdarrayTensor]")  [return-value]
monai/bundle/reference_resolver.py:245:21: error: Item "dict_items[Any, Any]" of "Union[Iterator[tuple[Any, Any]], dict_items[Any, Any], enumerate[Any]]" has no attribute "__next__"  [union-attr]
monai/bundle/reference_resolver.py:247:13: error: Incompatible types in "yield" (actual type "tuple[Union[Any, int], str, Any]", expected type "tuple[str, str, Any]")  [misc]
monai/apps/deepgrow/dataset.py:159:45: error: Argument 1 to "sum" has incompatible type "Optional[Any]"; expected "Union[_SupportsArray[dtype[Any]], _NestedSequence[_SupportsArray[dtype[Any]]]]"  [arg-type]
monai/apps/deepgrow/dataset.py:178:35: error: Item "None" of "Optional[Any]" has no attribute "flatten"  [union-attr]
[...]

PR #8153, opened 2024-10-16
TRT support for MAISI

borisfom:
### Description
Added trt_compile() support for Lists and Tuples in arguments for forward() - needed for MAISI. 
Did not add support for grouping return results yet - MAISI worked with explicit workaround unrolling the return results.

### Notes
To successfully export MAISI, either latest Torch nightly is needed, or this patch needs to be applied to 24.09-based container: 

```
--- /usr/local/lib/python3.10/dist-packages/torch/onnx/symbolic_opset14.bak     2024-10-09 01:38:04.920316673 +0000                                                   
+++ /usr/local/lib/python3.10/dist-packages/torch/onnx/symbolic_opset14.py      2024-10-09 01:38:25.228053951 +0000                                                   
@@ -148,7 +148,6 @@                                                                                                                                                   
         is_causal and symbolic_helper._is_none(attn_mask)                                                                                                            
     ), "is_causal and attn_mask cannot be set at the same time"                                                                                                      
                                                                                                                                                                      
-    scale = symbolic_helper._maybe_get_const(scale, "f")                                                                                                             
     if symbolic_helper._is_none(scale):                                                                                                                              
         scale = _attention_scale(g, query)                                                                                                                           
```

### Types of changes
<!--- Put an `x` in all the boxes that apply, and remove the not applicable items -->


borisfom:
Also, I did not do any results verification. If any results depend on Meta tensors operation, that part may be lost. Please check!

monai/networks/utils.py
```diff
@@ -693,7 +695,7 @@ def convert_to_onnx(
             f = io.BytesIO()
```
KumoLiu, line 695: Could you please also modify this part based on the latest api from `torch.onnx.export`? Thanks!
https://github.com/Project-MONAI/MONAI/issues/8149#issuecomment-2422165354

KumoLiu:
Hi @binliunls, please also review the trt related parts in this PR, thanks.

monai/apps/generation/maisi/networks/controlnet_maisi.py
```diff
@@ -119,7 +119,7 @@ def forward(
         down_block_res_samples = [h * conditioning_scale for h in down_block_res_samples]
         mid_block_res_sample *= conditioning_scale
 
-        return down_block_res_samples, mid_block_res_sample
+        return [*down_block_res_samples, mid_block_res_sample]
```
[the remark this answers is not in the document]
borisfom: Oh right, I did not revert that line after adding output_lists - will do!

Dockerfile
```diff
@@ -11,7 +11,7 @@
 
 # To build with a different base image
 # please run `docker build` using the `--build-arg PYTORCH_IMAGE=...` flag.
-ARG PYTORCH_IMAGE=nvcr.io/nvidia/pytorch:24.08-py3
+ARG PYTORCH_IMAGE=nvcr.io/nvidia/pytorch:24.09-py3
```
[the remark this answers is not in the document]
borisfom: Well it does not make real difference (patch  I mentioned in the description is needed for 24.09 anyway), so  I may revert this one for now, too. 24.10 (and 2.5.0) won't require exporter patch.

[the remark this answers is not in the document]
KumoLiu: Does this mean we'll need to update to version 24.10 once it's released, since 24.09 still doesn't meet the requirements, and MAISI still lacks TRT support?
I try to update the base image and trigger more test in this PR #8164, shown an error below:
https://github.com/Project-MONAI/MONAI/pull/8164#issuecomment-2428493586

monai/apps/generation/maisi/networks/controlnet_maisi.py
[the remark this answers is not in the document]
KumoLiu: I noticed your update, seems this file is outdated. Marking it as resolved, thanks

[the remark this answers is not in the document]
borisfom: Right. That line was reverted after I added output_lists option.

Dockerfile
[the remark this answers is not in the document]
borisfom: Yes I believe it's better to skip 24.09 as it still requires a patch.

monai/networks/trt_compiler.py
```diff
@@ -255,6 +345,7 @@ def __init__(
                     'torch_trt' may not work for some nets. Also AMP must be turned off for it to work.
             input_names: Optional list of input names. If None, will be read from the function signature.
             output_names: Optional list of output names. Note: If not None, patched forward() will return a dictionary.
+            output_lists: Optional list of output lists.
```
binliunls, line 348: Could you please add more details about this parameter and the relation between this one and the output_names? Now it's hard to understand the meaning of it.

Thanks,  
Bin

monai/networks/trt_compiler.py
```diff
@@ -233,13 +321,15 @@ def __init__(
         method="onnx",
         input_names=None,
         output_names=None,
+        output_lists=None,
```
binliunls, line 324: Could you please also add a simple test case to the unit test to show case how to use this parameter?

Thanks,  
Bin

monai/networks/trt_compiler.py
```diff
@@ -329,18 +428,22 @@ def forward(self, model, argv, kwargs):
...
             try:
                 self._load_engine()
                 if self.engine is None:
-                    build_args = kwargs.copy()
-                    if len(argv) > 0:
-                        build_args.update(self._inputs_to_dict(argv))
-                    self._build_and_save(model, build_args)
-                    # This will reassign input_names from the engine
+                    build_args = args.copy()
+                    with torch.no_grad():
```
binliunls, line 444: May I ask the reason for adding the `torch.no_grad()` here? Was it caused some issues in the previous version?

Thanks,  
Bin

borisfom: Yes there were some issues with export. As TRT is inference-only, it makes sense to do the whole export with torch.no_grad() - this is the recommended way.

monai/networks/trt_compiler.py
```diff
@@ -180,7 +184,8 @@ def try_set_inputs():
                     raise
                 self.cur_profile = next_profile
                 ctx.set_optimization_profile_async(self.cur_profile, stream)
-
+            except Exception:
+                raise
```
binliunls, line 188: Could you please add more info to explain this exception?
Thanks,  
Bin

monai/networks/trt_compiler.py
```diff
@@ -217,6 +222,89 @@ def infer(self, stream, use_cuda_graph=False):
...
+        # print(f"Removing non-tensor input: {key})")
+        input_example.pop(key)
+    return non_tensors
+
+
+def unroll_input(input_names, input_example):
+    # Simulate list/tuple unrolling during ONNX export
+    unrolled_input = {}
+    for name in input_names:
+        val = input_example[name]
```
binliunls, line 249: I think `input_example.get(name, None)` is a better choice here, in case there are any illegal keys.

Thanks,  
Bin

Dockerfile
```diff
@@ -41,6 +41,10 @@ RUN cp /tmp/requirements.txt /tmp/req.bak \
 COPY LICENSE CHANGELOG.md CODE_OF_CONDUCT.md CONTRIBUTING.md README.md versioneer.py setup.py setup.cfg runtests.sh MANIFEST.in ./
 COPY tests ./tests
 COPY monai ./monai
+
+# TODO: remove this line and torch.patch for 24.11
+RUN patch -R -d /usr/local/lib/python3.10/dist-packages/torch/onnx/ < ./monai/torch.patch
```
KumoLiu, line 46: Seems the patch not included in 24.10, right?

borisfom: The proper fix is not included with 24.10, yes, so we have to patch.

monai/networks/trt_compiler.py
borisfom: Yes we can look more into making this robust for the odd cases.

monai/networks/trt_compiler.py
borisfom: This would be exception trying to set input shapes for which the engine was not built ; previously I had a logic there that would try rotating trt optimization profile index on such an exception - we do not use multiple profiles with MONAI so I should probably simplify the code.

borisfom: Actually, let's just leave the profile selection code here, it does not hurt.

monai/networks/trt_compiler.py
borisfom: Done.

monai/networks/trt_compiler.py
borisfom: Done.

tests/test_trt_compile.py
```diff
@@ -46,7 +60,7 @@ def tearDown(self):
         if current_device != self.gpu_device:
             torch.cuda.set_device(self.gpu_device)
 
-    @SkipIfAtLeastPyTorchVersion((2, 5, 0))
+    @unittest.skipUnless(torch_trt_imported, "torch_tensorrt is required")
```
KumoLiu, line 63: Does this change mean even before 2.5, the test will pass as well? Since we have multiple pytorch version test in the ci.

KumoLiu: The test will still failed in 24.10 container: https://prod.blsm.nvidia.com/dlmed-clara-jenkins/blue/organizations/jenkins/Monai-latest-image/detail/Monai-latest-image/1306/pipeline/227/

borisfom: In this PR, I used arg_inputs instead of inputs argument in call to torch_tensorrt.convert_method_to_trt_engine, to work around  Torch-TRT bug that forced us to introduce that skip.

KumoLiu: Hi @borisfom, I don’t think this approach will work with Torch versions prior to 2.4. I can run more tests once you’ve resolved the conflict. Thanks!

borisfom: @KumoLiu - sure please do try with older versions - we may then have to add another guard on minimal Torch version if if fails with < 2.4.

KumoLiu (approved):

PR #8169, opened 2024-10-23
Fix the logging of a nested dictionary metric in MLflow

KumoLiu:
Fix https://github.com/Project-MONAI/model-zoo/issues/697

### Description
Flatten the metric dict when the metric is a nested dictionary.

### Types of changes

### DIFF
--- a/monai/handlers/mlflow_handler.py
+++ b/monai/handlers/mlflow_handler.py
@@ -22,7 +22,7 @@ import torch
 from torch.utils.data import Dataset
 
 from monai.apps.utils import get_logger
-from monai.utils import CommonKeys, IgniteInfo, ensure_tuple, min_version, optional_import
+from monai.utils import CommonKeys, IgniteInfo, ensure_tuple, flatten_dict, min_version, optional_import
 
 Events, _ = optional_import("ignite.engine", IgniteInfo.OPT_IMPORT_VERSION, min_version, "Events")
 mlflow, _ = optional_import("mlflow", descriptor="Please install mlflow before using MLFlowHandler.")
@@ -303,7 +303,9 @@ class MLFlowHandler:
 
         run_id = self.cur_run.info.run_id
         timestamp = int(time.time() * 1000)
-        metrics_arr = [mlflow.entities.Metric(key, value, timestamp, step or 0) for key, value in metrics.items()]
+        metrics_arr = [
+            mlflow.entities.Metric(key, value, timestamp, step or 0) for key, value in flatten_dict(metrics).items()
+        ]
         self.client.log_batch(run_id=run_id, metrics=metrics_arr, params=[], tags=[])
 
     def _parse_artifacts(self):
--- a/monai/networks/utils.py
+++ b/monai/networks/utils.py
@@ -692,7 +692,7 @@ def convert_to_onnx(
             onnx_inputs = tuple(inputs)
         if filename is None:
             temp_file = tempfile.NamedTemporaryFile(delete=False)
-            f = f"{temp_file.name}/model.onnx"
+            f = temp_file.name
         else:
             f = filename
 
--- a/monai/networks/utils.py
+++ b/monai/networks/utils.py
@@ -15,7 +15,6 @@ Utilities and types for defining networks, these depend on PyTorch.
 from __future__ import annotations
 
 import io
-import os
 import re
 import tempfile
 import warnings
--- a/monai/networks/utils.py
+++ b/monai/networks/utils.py
@@ -15,7 +15,9 @@ Utilities and types for defining networks, these depend on PyTorch.
 from __future__ import annotations
 
 import io
+import os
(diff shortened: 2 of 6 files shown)


monai/utils/misc.py
```diff
@@ -916,3 +916,16 @@ def unsqueeze_right(arr: NT, ndim: int) -> NT:
...
+def flatten_dict(metrics: dict[str, Any]) -> dict[str, Any]:
+    """
+    Flatten the nested dictionary to a flat dictionary.
+    """
+    result = {}
+    for key, value in metrics.items():
+        if isinstance(value, dict):
+            result.update(flatten_dict(value))
+        else:
+            result[f"{key}"] = value
```
binliunls, line 930: May I ask why not use `result[key]` directly since the input type hint has already indicate all keys are in `str` type.

Thanks,  
Bin

monai/handlers/mlflow_handler.py
```diff
@@ -303,7 +303,9 @@ def _log_metrics(self, metrics: dict[str, Any], step: int | None = None) -> None
 
         run_id = self.cur_run.info.run_id
         timestamp = int(time.time() * 1000)
-        metrics_arr = [mlflow.entities.Metric(key, value, timestamp, step or 0) for key, value in metrics.items()]
+        metrics_arr = [
```
binliunls, line 306: Could you please also update the MLflow test case to cover this issue?

Thanks,  
Bin

monai/utils/misc.py
```diff
@@ -916,3 +916,16 @@ def unsqueeze_right(arr: NT, ndim: int) -> NT:
 def unsqueeze_left(arr: NT, ndim: int) -> NT:
     """Prepend 1-sized dimensions to `arr` to create a result with `ndim` dimensions."""
     return arr[(None,) * (ndim - arr.ndim)]
+
+
+def flatten_dict(metrics: dict[str, Any]) -> dict[str, Any]:
```
binliunls, line 921: I think there is a potential risk when the input dict has the same key in different dict level. For example:
```
input_dict = {"a":1, "b":2, "c": {"a":3, "d":4}}
print(flatten_dict(input_dict))
```

The output will be:
```
{'a': 3, 'b': 2, 'd': 4}
```
I would suggest to reserve the key as a prefix when its value is a dict.

Thanks,  
Bin

KumoLiu: The purpose is to keep only one if there are duplicate keys.

monai/handlers/mlflow_handler.py
KumoLiu: Added, please help take a look at it again, thanks.

monai/utils/misc.py
KumoLiu: updated, thanks.

monai/networks/utils.py
```diff
@@ -688,27 +690,26 @@ def convert_to_onnx(
             onnx_inputs = (inputs,)
         else:
             onnx_inputs = tuple(inputs)
-
         if filename is None:
-            f = io.BytesIO()
+            temp_file = tempfile.NamedTemporaryFile(delete=False)
```
KumoLiu, line 694: Hi @borisfom, based on your comments here: https://github.com/Project-MONAI/MONAI/issues/8149#issuecomment-2433177783
I try to create a temp file when filename is None, could you please take a look at whether the change here make senses? Thanks.

monai/utils/misc.py
binliunls: I am not sure if users want this. Some of them may want to keep all metrics. But we can try your way first. If there is no bug report, maybe we can keep it.

Thanks  
Bin

tests/test_handler_mlflow.py
```diff
@@ -138,10 +143,12 @@ def _update_metric(engine):
                 state_attributes=["test"],
                 experiment_param=experiment_param,
                 artifacts=[artifact_path],
-                close_on_complete=True,
+                close_on_complete=False,
```
binliunls, line 146: Not sure if this will cause other mlflow cases error. Let's run the test and see the result.

monai/networks/utils.py
borisfom: Well the way it is now, it would crash if filename in not None on temp_file.close() below 
I suggest not using delete=False then close/remove is not needed. You may do temp_file=None above to keep it in the scope : 
```
temp_file = None
if filename is None:
            temp_file = tempfile.NamedTemporaryFile(delete=False)
            ...
```

KumoLiu: Updated, please help take a look at it again, thanks!

borisfom (approved):

tests/test_handler_mlflow.py
KumoLiu: All test passed.

binliunls (approved):

PR #8176, opened 2024-10-24
Add onnx output option for trt_compiler

yiheng-wang-nv:
Fixes # .

### Description

So far, it's onnx model is in a tmp folder and will be removed after execution. For users that need onnx model, it will be more convenient to add an option to save it to a specified path.

### Types of changes


yiheng-wang-nv:
close as it's not needed

PR #8177, opened 2024-10-24
Fix `torch.load()` frequently warning in PersistentDataset and GDSDataset

bnbqq8:
### Description
Frequently getting warning massage in a newer Pytorch version (2.4.1 in my case):
"_You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature._"

This pull request fixes an issue with the `torch.load()` function in the `PersistentDataset` and `GDSDataset` classes by adding `weights_only=False`. The fix ensures that the `torch.load()` function maintains load consistency in future versions of PyTorch.

### Types of changes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ ci:
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.6.0
+    rev: v5.0.0
     hooks:
       - id: end-of-file-fixer
       - id: trailing-whitespace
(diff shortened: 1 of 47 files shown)


ericspod:
Thanks for the contribution, this is a fix that we do need to remove that warning. I'm not sure if the fails recorded here are related to this PR rather than something else wrong with our CICD system but please run them locally to see if you have the same result. 

Could you also try with `weights_only=True` instead? We should be only loading tensors and not code with the datasets so this should work and be more secure. Thanks!

bnbqq8:
> Thanks for the contribution, this is a fix that we do need to remove that warning. I'm not sure if the fails recorded here are related to this PR rather than something else wrong with our CICD system but please run them locally to see if you have the same result.
> 
> Could you also try with `weights_only=True` instead? We should be only loading tensors and not code with the datasets so this should work and be more secure. Thanks!

Thanks for the comment! I believe the failed checks occurred because the `weights_only` parameter was not introduced in `torch.load()` until v1.13.0. [This is why the checks for vresion bellow v1.13.0 failed](https://github.com/Project-MONAI/MONAI/actions/runs/11496794922/job/31999110020?pr=8177#step:8:9440).

So, some compatibility work still needs to be done, and I’m working on it. :)

As for setting `weights_only=True`, I don’t think it's strictly necessary at this point, since the default setting is `weights_only=False`. We should maintain consistency for now, right? I mean some users might be caching objects other than tensors, which would require `weights_only=False`.

KumoLiu:
Hi @bnbqq8, could you please help address the format issue and DCO?
https://github.com/Project-MONAI/MONAI/actions/runs/11511448642/job/32044911037?pr=8177
https://github.com/Project-MONAI/MONAI/pull/8177/checks?check_run_id=32044901330

KumoLiu (approved):

bnbqq8:
@KumoLiu  Of course! Thank you for the reminder!

PR #8179, opened 2024-10-24
Fix Ruff Numpy2 deprecation rules

jamesobutler:
### Description

This addresses expired numpy deprecations that happened with the numpy 2 release. This code is compatible with numpy 1 and 2 to support future enablement of numpy2 for monai.

See https://docs.astral.sh/ruff/rules/numpy2-deprecation/

### Types of changes


ericspod (approved):
Thanks for the contribution, this all looks fine to me.

KumoLiu (approved):

PR #8188, opened 2024-11-02
enable gpu load nifti

yiheng-wang-nv:
Related to https://github.com/Project-MONAI/MONAI/issues/8241 .

### Description

### Types of changes


yiheng-wang-nv:
related topic in: https://github.com/nipy/nibabel/discussions/1385

Will modify my draft PR accordingly.

monai/data/image_reader.py
```diff
@@ -880,12 +896,15 @@ def __init__(
         channel_dim: str | int | None = None,
         as_closest_canonical: bool = False,
         squeeze_non_spatial_dims: bool = False,
+        gpu_load: bool = False,
```
yiheng-wang-nv, line 899: to_gpu

monai/data/image_reader.py
```diff
@@ -864,12 +874,19 @@ class NibabelReader(ImageReader):
...
             this is used to set original_channel_dim in the metadata, EnsureChannelFirstD reads this field.
             if None, `original_channel_dim` will be either `no_channel` or `-1`.
             most Nifti files are usually "channel last", no need to specify this argument for them.
+        as_closest_canonical: if True, load the image as closest to canonical axis format.
+        squeeze_non_spatial_dims: if True, non-spatial singletons will be squeezed, e.g. (256,256,1,3) -> (256,256,3)
+        to_gpu: If True, load the image into GPU memory using CuPy and Kvikio. This can accelerate data loading.
+            Default is False. CuPy and Kvikio are required for this option.
+            Note: For compressed NIfTI files, some operations may still be performed on CPU memory,
+            and the acceleration may not be significant. In some cases, it may be slower than loading on CPU.
+            #TODO: the first kvikio call is slow since it will initialize internal buffers, cuFile, GDS, etc.
```
yiheng-wang-nv, line 887: Hi @Nic-Ma @KumoLiu @ericspod , here I added a TODO, and later after finishing an acceleration tutorial, I will submit a separate PR to update the docstrings. What do you think?

KumoLiu:
Thank you for your contribution!

I have a question about this enhancement: Will it load NIfTI/DICOM files to the GPU faster compared to caching all the data on the GPU? If not, does this option use less GPU memory since the metadata is stored on the CPU? Thanks.

yiheng-wang-nv:
> Thank you for your contribution!
> 
> I have a question about this enhancement: Will it load NIfTI/DICOM files to the GPU faster compared to caching all the data on the GPU? If not, does this option use less GPU memory since the metadata is stored on the CPU? Thanks.

Hi @KumoLiu , these are two different things. The enhanced reader can load nii images from disk to GPU directly (the old way is to load from disk to cpu, and then convert from cpu to gpu), and the loaded data can be cached.

yiheng-wang-nv:
Hi @KumoLiu , do you know what is the root cause of the doc error: https://github.com/Project-MONAI/MONAI/actions/runs/12311294929/job/34361260922?pr=8188#step:7:247
cupy and kvikio are only imported optionally, I'm not sure why we have the error. Do I need to modify some other places?

KumoLiu:
> Hi @KumoLiu , do you know what is the root cause of the doc error: https://github.com/Project-MONAI/MONAI/actions/runs/12311294929/job/34361260922?pr=8188#step:7:247 cupy and kvikio are only imported optionally, I'm not sure why we have the error. Do I need to modify some other places?

Yes, you may need include `cupy` here: https://github.com/Project-MONAI/MONAI/blob/dev/docs/requirements.txt

monai/data/image_reader.py
```diff
@@ -1015,14 +1043,34 @@ def _get_spatial_shape(self, img):
         spatial_rank = max(min(ndim, 3), 1)
         return np.asarray(size[:spatial_rank])
 
-    def _get_array_data(self, img):
+    def _get_array_data(self, img, filename):
         """
         Get the raw array data of the image, converted to Numpy array.
 
         Args:
             img: a Nibabel image object loaded from an image file.
```
KumoLiu, line 1051: Please also add docstring for filename here.

monai/data/image_reader.py
```diff
@@ -1015,14 +1043,34 @@ def _get_spatial_shape(self, img):
...
 
         """
+        if self.to_gpu:
+            file_size = os.path.getsize(filename)
+            image = cp.empty(file_size, dtype=cp.uint8)
```
KumoLiu, line 1056: Will it be an issue when data dtype is not uint8?

monai/data/image_reader.py
ericspod: Is there any way to do that warm up call in some other way, such as in the constructor?

monai/data/image_reader.py
```diff
@@ -880,12 +897,22 @@ def __init__(
...
         self.channel_dim = float("nan") if channel_dim == "no_channel" else channel_dim
         self.as_closest_canonical = as_closest_canonical
         self.squeeze_non_spatial_dims = squeeze_non_spatial_dims
+        if to_gpu is True:
+            if not has_cp:
+                warnings.warn("CuPy is not installed, fall back to use cpu load.")
+                to_gpu = False
+            if not has_kvikio:
+                warnings.warn("Kvikio is not installed, fall back to use cpu load.")
+                to_gpu = False
```
ericspod, line 913:
```suggestion
        if to_gpu and (not has_cp or not has_kvikio):
            warnings.warn("NibabelReader: CuPy and/or Kvikio not installed for GPU loading, falling back to CPU loading.")
            to_gpu = False
```
I'd suggest something like this so we have fewer warnings. It's less clear which library is missing, but the user is going to have to sort out their environment anyway so they'll find out which one it is.

monai/data/image_reader.py
```diff
@@ -924,7 +952,7 @@ def read(self, data: Sequence[PathLike] | PathLike, **kwargs):
             img_.append(img)  # type: ignore
         return img_ if len(filenames) > 1 else img_[0]
 
-    def get_data(self, img) -> tuple[np.ndarray, dict]:
+    def get_data(self, img) -> tuple[np.ndarray | cp.ndarray, dict]:
```
ericspod, line 955: Does the typing work if CuPy isn't installed? I don't know if our CICD tests are going to catch when it's not.

monai/data/image_reader.py
yiheng-wang-nv: Hi @ericspod , thanks for the suggestions.
Yes, we can do in the constructor, and trigger if `to_gpu=True`. However, there is no standard warm up way according to kvikio team's information. My current method is:
```
def warm_up():
    a = cp.arange(100)
    with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
        tmp_file_name = tmp_file.name
        f = kvikio.CuFile(tmp_file_name, "w")
        # Write whole array to file
        f.write(a)
        f.close()

        b = cp.empty_like(a)
        f = kvikio.CuFile(tmp_file_name, "r")
        f.read(b)
```
I think it can be used in practical, or in a tutorial, but I'm not sure if it's formal to add this kind of code in our reader. What do you think?

monai/data/image_reader.py
yiheng-wang-nv: Hi @KumoLiu , thanks for the question.
CUPY uses uint8 for handling bytes data. Since the loaded data is bytes, cp.uint8 is used here.

monai/data/image_reader.py
yiheng-wang-nv: Thanks @ericspod . TYPE_CHECKING requires the library is installed, but install cupy and kvikio in our CI/CD machine seems not easy (cannot simply add in requirements.txt)
Therefore, I remove this type hint and remove the import from TYPE_CHECKING. Do you think it's fine?

monai/data/image_reader.py
ericspod: This is the probably only way if there's nothing to explicitly init things, so it's up to you if avoiding the slow initial call is worth adding this. I would add a comment about why this is here and set `delete_on_close` to `False` instead of `delete` in the context manager.

yiheng-wang-nv: Thanks @ericspod . I added the warm up. As for the delete/delete_on_close arg, after rerun tests, I found that we can delete in default and it will not impact the warm up performance. Therefore, now I use `tempfile.NamedTemporaryFile()`

monai/data/image_reader.py
yiheng-wang-nv: Thank, I added a TODO in the code

monai/data/image_reader.py
```diff
@@ -880,14 +897,43 @@ def __init__(
...
+        self.to_gpu = to_gpu
         self.kwargs = kwargs
 
+    def warmup_kvikio(self):
+        """
+        Warm up the Kvikio library to initialize the internal buffers, cuFile, GDS, etc.
+        This can accelerate the data loading process when `to_gpu` is set to True.
+        """
+        if has_cp and has_kvikio:
+            print("warm up")
```
ericspod, line 925:
```suggestion
```

monai/data/image_reader.py
ericspod, line 887:
```suggestion
```
We can remove this TODO now I guess if we have the warmup code.

monai/data/image_reader.py
```diff
@@ -1015,14 +1065,35 @@ def _get_spatial_shape(self, img):
...
+            if filename.endswith(".nii.gz"):
+                # for compressed data, have to tansfer to CPU to decompress
+                # and then transfer back to GPU. It is not efficient compared to .nii file
+                # and may be slower than CPU loading in some cases.
+                warnings.warn("Loading compressed NIfTI file into GPU may not be efficient.")
+                compressed_data = cp.asnumpy(image)
+                with gzip.GzipFile(fileobj=io.BytesIO(compressed_data)) as gz_file:
+                    decompressed_data = gz_file.read()
+
+                file_size = len(decompressed_data)
```
ericspod, line 1091:
```suggestion
```
Not needed?

yiheng-wang-nv: Thanks @ericspod for the detailed check! I forgot to remove some of the experimental code

monai/data/meta_tensor.py
```diff
@@ -565,7 +572,7 @@ def ensure_torch_and_prune_meta(
         if simple_keys:
             # ensure affine is of type `torch.Tensor`
             if MetaKeys.AFFINE in meta:
-                meta[MetaKeys.AFFINE] = convert_to_tensor(meta[MetaKeys.AFFINE])  # bc-breaking
+                meta[MetaKeys.AFFINE] = convert_to_tensor(meta[MetaKeys.AFFINE], device=device)  # bc-breaking
```
KumoLiu, line 575: May don't need move affine to device here.

yiheng-wang-nv: Hi @KumoLiu , thanks for the comments, I removed the device arg, and re-tested the changes with decathlon Task03_Liver dataset.

KumoLiu (approved):

PR #8216, opened 2024-11-16
enhance download_and_extract

### LINKED ISSUE #5463: parameter `filepath` of `download_and_extract`
**Describe the bug**
```py
>>> import monai
>>> url = "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/MedNIST.tar.gz"

>>> monai.apps.utils.download_and_extract(url, filepath=".")
2022-11-03 15:15:04,858 - INFO - Expected md5 is None, skip md5 check for file ..
2022-11-03 15:15:04,858 - INFO - File exists: ., skipped downloading.
2022-11-03 15:15:04,858 - INFO - Non-empty folder exists in ., skipped extracting.

>>> monai.apps.utils.download_and_extract(url, filepath="./test")
test: 59.0MB [00:09, 6.22MB/s]                                                                                                                                                                                     
2022-11-03 15:16:23,355 - INFO - Downloaded: test
2022-11-03 15:16:23,356 - INFO - Expected md5 is None, skip md5 check for file test.
  File "/Users/Documents/MONAI/monai/apps/utils.py", line 313, in download_and_extract
    extractall(filepath=filename, output_dir=output_dir, file_type=file_type, has_base=has_base)
  File "/Users/Documents/MONAI/monai/apps/utils.py", line 256, in extractall
    if cache_dir.exists() and next(cache_dir.iterdir(), None) is not None:
  File "/usr/local/anaconda3/envs/py38/lib/python3.8/pathlib.py", line 1122, in iterdir
    for name in self._accessor.listdir(self):
NotADirectoryError: [Errno 20] Not a directory: 'test'

```

it's not easy to set a valid `filepath` for `download_and_extract` and the error messages are not very intuitive.

would be great if the argument can be used to cache the downloaded file easily

Jerome-Hsieh:
Fixes #5463 

### Description

According to issue, the error messages are not very intuitive. 
I think maybe we can check if the file name matches the downloaded file’s base name before starting the download. 
If it doesn’t match, it will notify user.

### Types of changes


ericspod:
Hi @Jerome-Hsieh If you're having issues with putting this PR together we can discuss how to resolve them, but please do not close and then open a new PR. Thanks for the contribution!

monai/apps/utils.py
```diff
@@ -327,7 +327,17 @@ def download_and_extract(
...
+    urlFilenameExtension = ''.join(Path(".", _basename(url)).resolve().suffixes)
+    if filepath:
+        FilepathExtenstion = ''.join(Path(".", _basename(filepath)).resolve().suffixes)
+        if urlFilenameExtension != FilepathExtenstion:
+            raise NotImplementedError(
+                f'The file types do not match: url={urlFilenameExtension}, but filepath={FilepathExtenstion}'
+            )
     with tempfile.TemporaryDirectory() as tmp_dir:
-        filename = filepath or Path(tmp_dir, _basename(url)).resolve()
+        if filepath:
```
KumoLiu, line 338: Didn't get the idea for this change here. Seems equivalent to me.

monai/apps/utils.py
```diff
@@ -327,7 +327,17 @@ def download_and_extract(
             be False.
         progress: whether to display progress bar.
     """
+    urlFilenameExtension = ''.join(Path(".", _basename(url)).resolve().suffixes)
+    if filepath:
+        FilepathExtenstion = ''.join(Path(".", _basename(filepath)).resolve().suffixes)
+        if urlFilenameExtension != FilepathExtenstion:
+            raise NotImplementedError(
```
KumoLiu, line 334: In my opinion, the issue [here](https://github.com/Project-MONAI/MONAI/issues/5463) is that if filepath can directly get the name of the downloaded file, the `download_and_extract` function would work as expected, rather than only raising an error at this point.

KumoLiu: I know your point here.
Do you think when user specify `filepath='./test'`, it would be better to give a warning and automatically catch the filename?

cc @ericspod @Nic-Ma

ericspod: I would warn instead of forcing extensions, it takes control away from people to force things like this that aren't strictly necessary.

ericspod, line 334: For reference, `NotImplementedError` indicates some method or something isn't implemented, it's not meant for this kind of error, so instead you should use `ValueError` or some other things more closely aligned with this sort of condition.

monai/apps/utils.py
```diff
@@ -327,7 +327,17 @@ def download_and_extract(
             be False.
         progress: whether to display progress bar.
     """
+    urlFilenameExtension = ''.join(Path(".", _basename(url)).resolve().suffixes)
```
ericspod, line 330: Please also stick to Pythonic variable name conventions, that is all lower case with words separated by _ typically. This variable could be instead `url_filename_ext` to align with this convention.

monai/apps/utils.py
Jerome-Hsieh: But I have a question.
Is it necessary for the user to specify the file extension when setting the `filepath` for the downloaded file?

ericspod: If the user gives a path without extension then you'll get a file with no extension, the user would have to account for that and provide an extension. You could add an extension if none is given (and warn that this happened), but if the user has given a different extension than what's expected, this should be a warning and not an error. If you add an extension you should also be careful not to overwrite an existing file. I realise this is getting rather complicated now, sorry!

Jerome-Hsieh:
@ericspod @KumoLiu 
Here's newest change. like we discussed before. I check the `filepath` every time at the very beginning. 
If no extension is provided, the system will automatically add the appropriate file extension with a warning.
A warning will also appear if provided extension doesn't match the downloaded file's extension .

KumoLiu:
> @ericspod @KumoLiu Here's newest change. like we discussed before. I check the `filepath` every time at the very beginning. If no extension is provided, the system will automatically add the appropriate file extension with a warning. A warning will also appear if provided extension doesn't match the downloaded file's extension .

Hi @Jerome-Hsieh, thank you for the quick update. I tested your latest changes, but neither "." nor "./test" as the file path seems to work. Is this the expected behavior? In my opinion, both should work. The logic could be adjusted so that if the file path is a directory and not empty, it automatically captures the name and downloads the file into the specified directory. What do you think?
```
url = "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/MedNIST.tar.gz"
monai.apps.utils.download_and_extract(url, filepath="./test")
```

Jerome-Hsieh:
Hi @KumoLiu my test results：
```python
>>> url = "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/MedNIST.tar.gz"
>>> monai.apps.utils.download_and_extract(url, filepath=".")
2024-12-06 18:06:56,333 - INFO - Expected md5 is None, skip md5 check for file ..
2024-12-06 18:06:56,334 - INFO - File exists: ., skipped downloading.
2024-12-06 18:06:56,334 - INFO - Non-empty folder exists in ., skipped extracting.
```
and
``` python
>>> monai.apps.utils.download_and_extract(url, filepath="./test")
2024-12-06 18:08:41,246 - WARNING - filepath=./test, which missing file extension. Auto-appending extension to: ./test.tar.gz
test.tar.gz: 59.0MB [00:10, 6.07MB/s]
2024-12-06 18:08:51,443 - INFO - Downloaded: test.tar.gz
2024-12-06 18:08:51,443 - INFO - Expected md5 is None, skip md5 check for file test.tar.gz.
2024-12-06 18:08:51,443 - INFO - Writing into directory: ..
```
If the results like above, that's expected behavior.

Your opinion sounds like
`url=example.com/file.tar.gz, filepath=./test`
the program will  work ：`./test/file.tar.gz`.
If I don't misunderstanding your thoughts, this method although can automatically captures the name, not only capture the strange filename make user confused if the URL don't have appropriate naming, but also can't allow user set the filename they want.

mingxin-zheng:
Regarding file downloading from URL, sometimes we can infer the name from the content deposition. Here is one example snippet: https://github.com/Project-MONAI/VLM/blob/7be688aa457a1806f908eb758f2f3ee816fea017/m3/demo/experts/utils.py#L135 @KumoLiu

Jerome-Hsieh:
Thank @mingxin-zheng provide information to find file name.
But I want to say that at first my initial approach to this issue that `filepath` should be a compressed file rather than a directory.
I wanted to focus on making filepath valid. Method provided by @KumoLiu it's great, but which somewhat deviates from my original intent.
Sorry for not providing a clear explanation.

Jerome-Hsieh:
Hi @KumoLiu the newest commit I change the logic that if the file path is a directory and not empty, it will automatically capture the name from content deposition, if it doses't  have any file name, will use url basename  and download the file into the specified directory then gives a warning.
Besides, I also add if set the wrong file extension, it will raise the error and stop downloading.
Please let me know if you have any suggestions. Thx!

KumoLiu:
> Hi @KumoLiu the newest commit I change the logic that if the file path is a directory and not empty, it will automatically capture the name from content deposition, if it doses't have any file name, will use url basename and download the file into the specified directory then gives a warning. Besides, I also add if set the wrong file extension, it will raise the error and stop downloading. Please let me know if you have any suggestions. Thx!

Hi @Jerome-Hsieh, thank you for the update. After our discussion in the development meeting, I realized that there is an `output_dir` argument in `download_and_extract`. I apologize for any confusion my previous comments may have caused. It would be great to proceed by considering only the extension as before. Sorry again for the misunderstanding.

monai/apps/utils.py
```diff
@@ -27,6 +28,8 @@
 from urllib.parse import urlparse
 from urllib.request import urlopen, urlretrieve
 
+import requests
```
KumoLiu, line 31: Regarding the ci error here: https://github.com/Project-MONAI/MONAI/actions/runs/12338243392/job/34432967743?pr=8216
Please consider using `optional_import` here and 
https://github.com/Project-MONAI/MONAI/blob/21920a34bc00a114e430e1943e1fd1f572880919/monai/bundle/scripts.py#L61

And skip in the related tests.
https://github.com/Project-MONAI/MONAI/blob/21920a34bc00a114e430e1943e1fd1f572880919/tests/test_bundle_get_data.py#L46

Please also include related test! Thanks again.

Jerome-Hsieh:
> > Hi @KumoLiu the newest commit I change the logic that if the file path is a directory and not empty, it will automatically capture the name from content deposition, if it doses't have any file name, will use url basename and download the file into the specified directory then gives a warning. Besides, I also add if set the wrong file extension, it will raise the error and stop downloading. Please let me know if you have any suggestions. Thx!
> 
> Hi @Jerome-Hsieh, thank you for the update. After our discussion in the development meeting, I realized that there is an `output_dir` argument in `download_and_extract`. I apologize for any confusion my previous comments may have caused. It would be great to proceed by considering only the extension as before. Sorry again for the misunderstanding.

Hi @KumoLiu  I see, and what do you think about previous version [a9a0171](https://github.com/Project-MONAI/MONAI/pull/8216/commits/a9a01715480caa6d45a2275e328b0f11ed5e17dd)  I commit ?
Could you please identify any areas that require improvement or could be further enhance ?

KumoLiu:
> Hi @KumoLiu I see, and what do you think about previous version [a9a0171](https://github.com/Project-MONAI/MONAI/pull/8216/commits/a9a01715480caa6d45a2275e328b0f11ed5e17dd) I commit ? Could you please identify any areas that require improvement or could be further enhance ? Thx!

Hi @Jerome-Hsieh, [a9a0171](https://github.com/Project-MONAI/MONAI/pull/8216/commits/a9a01715480caa6d45a2275e328b0f11ed5e17dd) looks good. The only update needed is how to extract the file extension from the URL.

For example, given a URL like "https://drive.google.com/u/1/uc?id=1KntZge40tAHgyXmHYVqZZ5d2p_4Qr2l5&export=download", you can use `get_filename_from_url` to determine the exact extension based on the filename. What do you think? Please also include the tests, thanks!

monai/apps/utils.py
```diff
@@ -298,6 +301,35 @@ def extractall(
...
+            response = requests.head(data_url, allow_redirects=True)
+            content_disposition = response.headers.get("Content-Disposition")
+            if content_disposition:
+                filename = re.findall("filename=(.+)", content_disposition)
+                return filename[0].strip('"').strip("'")
+            else:
+                filename = _basename(data_url)
+                return filename
+    except Exception as e:
+        raise Exception(f"Error processing URL: {e}") from e
```
KumoLiu, line 330: Thank you for the quick update! It looks much better now. Here’s a simplified version that achieves the same functionality for your reference:
```suggestion
def get_filename_from_url(data_url: str) -> str:
    """
    Get the filename from the URL link.
    """
    try:
        response = requests.head(data_url, allow_redirects=True)
        content_disposition = response.headers.get("Content-Disposition")
        if content_disposition:
            filename = re.findall("filename=\"?([^\";]+)\"?", content_disposition)
            if filename:
                return str(filename[0])
        if "drive.google.com" in data_url:
            response = requests.get(data_url)
            if "text/html" in response.headers.get("Content-Type", ""):
                soup = BeautifulSoup(response.text, "html.parser")
                filename_div = soup.find("span", {"class": "uc-name-size"})
                if filename_div:
                    return str(filename_div.find("a").text)
        return _basename(data_url)
    except Exception as e:
        raise Exception(f"Error processing URL: {e}") from e
```

Jerome-Hsieh:
@KumoLiu   Thanks a lot for your assistance！It looks more cleaner than before!

KumoLiu (approved):

KumoLiu:
@ericspod do you have any further comments here? Otherwise I will try to merge this one, thanks.

PR #8263, opened 2024-12-13
Modify Workflow to Allow IterableDataset Inputs

ericspod:
### Description

This modifies the behaviour of `Workflow` to permit `IterableDataset` to be used correctly. A check against the `epoch_length` value is removed, to allow that value to be `None`, and a test is added to verify this. The length of a data loader is not defined when using iterable datasets, so try/raise is added to allow that to be queried safely. This is related to my work on the streaming support, in my [prototype gist](https://gist.github.com/ericspod/1904713716b45631260784ac3fcd6fb3) I had to provide a bogus epoch length value in the then change it to `None` later once the evaluator object was created. This PR will remove the need for this hack.

### Types of changes

### DIFF
--- a/monai/engines/workflow.py
+++ b/monai/engines/workflow.py
@@ -12,9 +12,8 @@
 from __future__ import annotations
 
 import warnings
-from collections.abc import Callable, Iterable, Sequence
+from collections.abc import Callable, Iterable, Sequence, Sized
 from typing import TYPE_CHECKING, Any
-from collections.abc import Sized
 
 import torch
 import torch.distributed as dist
--- a/monai/engines/workflow.py
+++ b/monai/engines/workflow.py
@@ -128,6 +128,7 @@ class Workflow(Engine):
 
             # set the epoch value for DistributedSampler objects when an epoch starts
             if isinstance(sampler, DistributedSampler):
+
                 @self.on(Events.EPOCH_STARTED)
                 def set_sampler_epoch(engine: Engine) -> None:
                     sampler.set_epoch(engine.state.epoch)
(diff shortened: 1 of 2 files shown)


monai/engines/workflow.py
```diff
@@ -121,24 +121,24 @@ def __init__(
...
 
+            # if the epoch_length isn't given, attempt to get it from the length of the data loader
             if epoch_length is None:
-                epoch_length = len(data_loader)
-        else:
-            if epoch_length is None:
-                raise ValueError("If data_loader is not PyTorch DataLoader, must specify the epoch_length.")
+                try:
+                    epoch_length = len(data_loader)
+                except TypeError:  # raised when data_loader is given an iterable dataset which has no length
```
KumoLiu, line 140: Do you think we should add a warning here or validate the dataset instance in the `data_loader`?
```
isinstance(data_loader.dataset, Iterable)
```

KumoLiu (approved):
Thanks for the pr, looks good to me. One minor comments inline.

monai/engines/workflow.py
ericspod: The `data_loader` argument itself isn't always a `DataLoader` instance so may not have a `dataset` member, it can be a `Iterable` itself which might have a length. It think there's a use case where you'd pass in just a list of items you already have a so not use a `DataLoader` at all and so this is the more robust way of attempting to get a length.

KumoLiu: So it looks like `if epoch_length is None:` this if should be moved to the outside level, right?

ericspod:
I just noticed pre-commit.ci made changes to the source file that were incompatible with isort's expectations. It added a new import line instead of appending to an existing one.

monai/engines/workflow.py
ericspod: I've tried a different method here actually that avoid checking the exception throw.

ericspod: And then I reverted to something similar again.

Issue #6832, opened 2023-08-07
Spacing transform does not update `pixdim`


KumoLiu:
### Discussed in https://github.com/Project-MONAI/MONAI/discussions/6831

<div type='discussions-op-text'>

<sup>Originally posted by **agaldran** August  7, 2023</sup>
Hello,

Sorry about the naive question, I am trying to wrap my head around transforms and meta_dicts. Right now I have the following operations:
```
import monai.transforms as t

data_dict = {'image': 'data/ct/images/.whatever.nii.gz'}
loader = t.LoadImaged(keys=('image'), image_only=False)
loaded_data_dict = loader(data_dict)
print(loaded_data_dict['image'].shape, loaded_data_dict['image_meta_dict']['pixdim'][1:4])
```
which returns a shape and voxel sizes of `[281, 313, 182]` and `(0.50,0.50,0.75)` respectively. Now, if I run:
```
spacing = t.Spacingd(keys=['image'], pixdim=(0.50, 0.50, 1.5), mode=('bilinear')) 
spaced_data_dict = spacing(loaded_data_dict)
print(spaced_data_dict['image'].shape, spaced_data_dict['image_meta_dict']['pixdim'][1:4])
```
I am expecting to see approx half the voxels in the Z direction, as the original spacing was 0.75 and I am asking for twice that spacing. For the other directions, nothing should be changed^* . This is what I get:
```
[281, 313, 92], (0.50,0.50,0.75)
```
So, although the number of voxels are as expected, the spacing remains the same. I cannot find any other key in the meta data dictionary that tells me what is the *current* spacing for this volume. Am I doing something wrong here?
</div>

wyli:
yes, the image's original `'pixdim'` is not modified, the current pixdim is a property: `spaced_data_dict["image"].pixdim`

KumoLiu:
> yes, the image's original `'pixdim'` is not modified, the current pixdim is a property: `spaced_data_dict["image"].pixdim`

Does it mean we don't need to modify it or we may enhance it in the future?

wyli:
We don't have a good reason to modify it at the moment, a single source of truth for the original and current pixdim seems to be consistent and easy to maintain..

PR #8269, opened 2024-12-23
Fixes #6840

### LINKED ISSUE #6840: Modify `pixdim` to `original_pixdim` in meta dict
Follow up of #6832, after `Spacing`, the original `pixdim` is not modified in meta dict, which may cause some confusion.

Would be better to update it to `original_pixdim`.
Similar to `affine`.
https://github.com/Project-MONAI/MONAI/blob/8e99af5f96df0746d6cbcaed88feaea0e51abd56/monai/data/image_reader.py#L279-L281

slicepaste:
Fixes #6840

### Description
According to the issue, this PR addresses on the meta dictionary `data['pixdim']` of NIfTI images does not update after applying the `spacing` or `spacingd`. To align with `affine`, we update `data['pixdim']` and keep the original metainfo in `data['original_pixdim']`. Additionally, this PR also update the metainfo `key_{meta_key_postfix}['pixdim']` in NIfTI images, consistent with `spaced_data_dict['image_meta_dict']['pixdim']` in issue #6832.

### Types of changes

### DIFF
--- a/monai/data/meta_tensor.py
+++ b/monai/data/meta_tensor.py
@@ -477,6 +477,10 @@ class MetaTensor(MetaObj, torch.Tensor):
             return [affine_to_spacing(a) for a in self.affine]
         return affine_to_spacing(self.affine)
 
+    def set_pixdim(self) -> None:
+        """Update pixdim based on current affine."""
+        self.meta[MetaKeys.PIXDIM][1 : 1 + len(self.pixdim)] = affine_to_spacing(self.affine)
+
     def peek_pending_shape(self):
         """
         Get the currently expected spatial shape as if all the pending operations are executed.
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -535,6 +535,9 @@ class Spacing(InvertibleTransform, LazyTransform):
             dtype=dtype,
             lazy=lazy_,
         )
+        if isinstance(data_array, MetaTensor) and "pixdim" in data_array.meta:
+            data_array = data_array.clone()
+            data_array.set_pixdim()
         if self.recompute_affine and isinstance(data_array, MetaTensor):
             if lazy_:
                 raise NotImplementedError("recompute_affine is not supported with lazy evaluation.")
--- a/monai/transforms/spatial/dictionary.py
+++ b/monai/transforms/spatial/dictionary.py
@@ -24,11 +24,13 @@ from typing import Any, cast
 import numpy as np
 import torch
 
+import monai.transforms as transforms
 from monai.config import DtypeLike, KeysCollection, SequenceStr
 from monai.config.type_definitions import NdarrayOrTensor
 from monai.data.box_utils import BoxMode, StandardMode
 from monai.data.meta_obj import get_track_meta
 from monai.data.meta_tensor import MetaTensor
+from monai.data.utils import is_supported_format
 from monai.networks.layers.simplelayers import GaussianFilter
 from monai.transforms.croppad.array import CenterSpatialCrop
 from monai.transforms.inverse import InvertibleTransform
@@ -520,6 +522,9 @@ class Spacingd(MapTransform, InvertibleTransform, LazyTransform):
                 output_spatial_shape=output_shape_k if should_match else None,
                 lazy=lazy_,
             )
+            if isinstance(d[key], MetaTensor) and "filename_or_obj" in d[key].meta:
+                if is_supported_format(d[key].meta["filename_or_obj"], ["nii", "nii.gz"]):
+                    d = transforms.sync_meta_info(key, d)
             if output_shape_k is None:
                 output_shape_k = d[key].peek_pending_shape() if isinstance(d[key], MetaTensor) else d[key].shape[1:]
         return d
--- a/monai/data/image_reader.py
+++ b/monai/data/image_reader.py
@@ -989,6 +989,8 @@ class NibabelReader(ImageReader):
 
         for i, filename in zip(ensure_tuple(img), self.filenames):
             header = self._get_meta_dict(i)
+            if MetaKeys.PIXDIM in header:
+                header[MetaKeys.ORIGINAL_PIXDIM] = np.array(header[MetaKeys.PIXDIM], copy=True)
             header[MetaKeys.AFFINE] = self._get_affine(i)
             header[MetaKeys.ORIGINAL_AFFINE] = self._get_affine(i)
             header["as_closest_canonical"] = self.as_closest_canonical
--- a/monai/utils/enums.py
+++ b/monai/utils/enums.py
@@ -528,6 +528,8 @@ class MetaKeys(StrEnum):
     Typical keys for MetaObj.meta
     """
 
+    PIXDIM = "pixdim"  # MetaTensor.pixdim
+    ORIGINAL_PIXDIM = "original_pixdim"  # the pixdim after image loading before any data processing
     AFFINE = "affine"  # MetaTensor.affine
     ORIGINAL_AFFINE = "original_affine"  # the affine after image loading before any data processing
     SPATIAL_SHAPE = "spatial_shape"  # optional key for the length in each spatial dimension


monai/data/meta_tensor.py
```diff
@@ -477,6 +477,10 @@ def pixdim(self):
             return [affine_to_spacing(a) for a in self.affine]
         return affine_to_spacing(self.affine)
 
+    def set_pixdim(self) -> None:
```
ericspod, line 480: Does it make sense for this to be a method? If it's only going to be called in one place it's simple code could just be put there. If there's anticipation that this would be called by other things then that's fine.

monai/transforms/spatial/array.py
```diff
@@ -535,6 +535,9 @@ def __call__(
             dtype=dtype,
             lazy=lazy_,
         )
+        if isinstance(data_array, MetaTensor) and "pixdim" in data_array.meta:
+            data_array = cast(MetaTensor, data_array.clone())
```
ericspod, line 539: Why is it necessary to clone the data array here? This is going to have a cost and I think isn't compatible with lazy resampling. Perhaps this is code that should be `SpactialResample` instead? @atbenmurray If you could please check if this is going to interact with laziness, thanks.

monai/transforms/spatial/array.py
```diff
@@ -535,6 +535,9 @@ def __call__(
...
+            data_array.set_pixdim()
```
KumoLiu, line 540: It seems that the pixel dimensions (pixdim) are only updated in the Spacing transformation. This is why I previously suggested that we should only retain the `original_pixdim`, as the latest pixdim can be derived from the metadata in the MetaTensor and the affine transformation.

Additionally, the use of `meta_dict` for logging metadata is becoming outdated. Perhaps we no longer need to maintain it? What are your thoughts on this?

```
import torch
from monai.transforms import Spacing

data = torch.randn(2, 1, 32, 32, 32)
trans = Spacing(pixdim=(1.5, 1.5, 1.0))
out = trans(data)
print(out.pixdim) -- > (1.5, 1.5, 1.0)
```
cc @ericspod @Nic-Ma

KumoLiu (commented):
Thanks for the contribution! Leave several comments inline.

monai/transforms/spatial/array.py
slicepaste: @ericspod Thank you for your suggestion.

We initially considered using `data_array.clone()` based on the following issue:
```
# LoadImage
data = {'image', 'data1.nii'}
imgloader = LoadImaged(keys=('image'), image_only=False, ensure_channel_first=True)
input_data_dict = imgloader(data)

# Spacing
respacing = transforms.Spacingd(keys=['image', 'label'], pixdim=(1, 1, 10), mode=('bilinear'))
spaced_data_dict = respacing(input_data_dict)
```
Originally, if we didn't use `data_array.clone()` and directly modified the data, the `MetaTensor` in both `input_data_dict` and `spaced_data_dict` would be affected simultaneously. 
This means `input_data_dict` would lose its original `input_data_dict["pixdim"]` information.

However, as suggested above, using `.clone()` in this way is not ideal as it introduces additional costs.
If we perform this modification within `TraceableTransform.track_transform_meta()` method, which is executed by `SpatialResample`, it might no longer be a concern.

monai/data/meta_tensor.py
slicepaste: Thanks for your quick reply.
Originally, we considered other files such as DICOM might use `Spacing`, which could involve the usage and access `meta_tensor.py` property. Therefore, we decided to define a method.
However, after we reevaluating the entire codebase this week, it might be better to modify `data["pixdim"]` directly within `TraceableTransform` instead.

monai/transforms/spatial/dictionary.py
```diff
@@ -520,6 +522,11 @@ def __call__(self, data: Mapping[Hashable, torch.Tensor], lazy: bool | None = No
                 output_spatial_shape=output_shape_k if should_match else None,
                 lazy=lazy_,
             )
+            if isinstance(d[key], MetaTensor) and f"{key}_meta_dict" in d:
+                if "filename_or_obj" in d[key].meta and is_supported_format(
+                    d[key].meta["filename_or_obj"], ["nii", "nii.gz"]
+                ):
+                    d = transforms.sync_meta_info(key, d)
```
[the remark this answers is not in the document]
einsyang723: Thank you for your question. In `MapTransform`, it only synchronizes MetaTensor, but since `{key}_meta_dict` is not a MetaTensor format, it won't be updated automatically. That's why we added this line of code to ensure the information in `"{key}_meta_dict` gets synchronized as well.

monai/transforms/spatial/array.py
einsyang723: @KumoLiu  Thank you for your detailed feedback. 

Our original thinking was that all related information should be update, including:

data['image']['pixdim']
data['image'].pixdim
data['image_meta_dict']['pixdim']

However, after reading your response, we're a bit uncertain: are you suggesting that we don't need to update pixdim in `{key}_meta_dict`?
Or will there be a new way to handle `{key}_meta_dict` in future releases? 
If possible, could you please elaborate on this? Thank you.

cc @slicepaste  @IamTingTing

atbenmurray:
Hi folks. I need to do a deeper evaluation of how this intersects with lazy resampling.

atbenmurray:
In the middle of extending test scenarios. Please bear with me.

slicepaste:
@einsyang723 @IamTingTing The current PR is not our final version, and it has remained unchanged for a while.
We should continue working on this issue and update the final version ASAP so we can move forward with this PR.  
Let me know if you need any help. Thanks!

atbenmurray:
Picking this up again. Will have a review for you all today.

monai/transforms/spatial/dictionary.py
```diff
@@ -520,6 +522,11 @@ def __call__(self, data: Mapping[Hashable, torch.Tensor], lazy: bool | None = No
                 output_spatial_shape=output_shape_k if should_match else None,
                 lazy=lazy_,
             )
+            if isinstance(d[key], MetaTensor) and f"{key}_meta_dict" in d:
```
atbenmurray, line 525: I think this may need to be more generalized. There are options in LoadImaged to change the default naming convention between a key and the meta_dict for that key. `meta_keys` and `meta_key_postfix` can change this behaviour. I don't know how well changing those from their defaults is supported, however.

atbenmurray (changes requested):

monai/transforms/spatial/dictionary.py
einsyang723: @atbenmurray Thank you for your suggestion. We plan to implement checking all dictionary keys that start with `{key}_` to support custom settings of `meta_keys` and `meta_key_postfix`. This ensures that no matter how users configure the naming conventions in LoadImaged, we can correctly synchronize metadata from the MetaTensor to the corresponding meta dictionary.

During implementation, we discovered that `sync_meta_info()` doesn't properly synchronize meta dictionaries with custom postfixes, and it creates a new key using the default postfix. 

For example, if the original key is 'image' and the postfix is changed to 'dict', there would be two keys: 'image' and 'image_dict'. However, after executing `sync_meta_info()`, a third key 'image_meta_dict' is created, using the default postfix. 

In our implementation, we've worked around this by directly updating the existing meta dictionaries, which avoids the creation of additional keys. Would you consider this behavior something that should be addressed in a separate issue, or is our current approach sufficient? We'd appreciate your thoughts on this.

cc @slicepaste @IamTingTing

atbenmurray: Whilst it isn't directly connected to this item, it does expose a design issue that we ought to find a solution to. Given that there is currently no way for anything downstream of LoadImaged to know what key was used for the metadata dictionary, can we instead add an entry to the metadata dictionary that indicates what key it relates to?

einsyang723: Thank you for your reply. We plan to enhance the `LoadImaged.call` method by adding a `key_meta_dict_map` dictionary to map data keys to their corresponding metadata keys, considering both user-defined `meta_keys` and the default naming convention (`{key}_{meta_key_postfix}`).

Since this differs from the focus of the current PR, we thought that opening a new issue to specifically address it may be better. This will keep the PR focused on the `pixdim` update.

Would this approach be feasible? We’d appreciate your feedback.

cc @slicepaste @IamTingTing

atbenmurray:
@einsyang723 @slicepaste @ericspod @KumoLiu @IamTingTing
Hi folks. Sorry it has been a while but I have a candidate PR #8411 that I think is important to address as a prerequisite to this PR. It is still in draft phase but take a look when you have a moment

atbenmurray (approved):
The change made in monai/transforms/spatial/dictionary.py means the issue that I raised is much less relevant now

slicepaste:
> The change made in monai/transforms/spatial/dictionary.py means the issue that I raised is much less relevant now

@atbenmurray Thank you for the approval. Just to confirm our understanding — this PR can be merged first, and the remaining concerns regarding `meta_keys`, `meta_key_postfix`, and the additional metadata dictionary changes will be addressed in PR #8411  , right?

atbenmurray:
> > The change made in monai/transforms/spatial/dictionary.py means the issue that I raised is much less relevant now
> 
> @atbenmurray Thank you for the approval. Just to confirm our understanding — this PR can be merged first, and the remaining concerns regarding `meta_keys`, `meta_key_postfix`, and the additional metadata dictionary changes will be addressed in PR #8411 , right?

Exactly. Yes. This PR is decoupled from PR #8411 now.

ericspod (approved):
I think everything has been addressed here, @KumoLiu please trigger blossom and we'll integrate this soon. Thanks!

yakovdan:
Hi, is there any progress on merging this? Perhaps blossom-ci is stuck?

PR #8286, opened 2025-01-08
8267 fix normalize intensity

### LINKED ISSUE #8267: monai.transforms.intensity.array.NormalizeIntensity bug when __call__  with channel_wise=True
**Describe the bug**
In the `__call__` method of `monai.transforms.intensity.array.NormalizeIntensity` class, the conversion into torch.float32 performed in self._normalize line 911, is not taken into account when `channel_wise = True`. The following code does not define a new variable img with the right type but it keeps the type of the original img variable. Therefore, if the original image was filled with int, the normalization will be performed on int and the normalization will fail.
```
for i, d in enumerate(img):
    img[i] = self._normalize(  # type: ignore
        d,
        sub=self.subtrahend[i] if self.subtrahend is not None else None,
        div=self.divisor[i] if self.divisor is not None else None,
    )
```

**To Reproduce**
Run commands
```
import torch
from monai.transforms.intensity.array import NormalizeIntensity
tensor = torch.arange(24).reshape(2,3,4)
normalized_correctly = NormalizeIntensity()(tensor)
normalized_corrupted = NormalizeIntensity(channel_wise=True)(tensor)
print(normalized_correctly)
print(normalized_corrupted)
````

**Expected behavior**
```
>>> normalized_correctly
metatensor([[[-1.6613, -1.5169, -1.3724, -1.2279],
         [-1.0835, -0.9390, -0.7945, -0.6501],
         [-0.5056, -0.3612, -0.2167, -0.0722]],

        [[ 0.0722,  0.2167,  0.3612,  0.5056],
         [ 0.6501,  0.7945,  0.9390,  1.0835],
         [ 1.2279,  1.3724,  1.5169,  1.6613]]])
>>> normalized_corrupted
metatensor([[[-1., -1., -1.,  0.],
         [ 0.,  0.,  0.,  0.],
         [ 0.,  1.,  1.,  1.]],

        [[-1., -1., -1.,  0.],
         [ 0.,  0.,  0.,  0.],
         [ 0.,  1.,  1.,  1.]]])
```

### PR DESCRIPTION
**Pull request by advcu987**
Fixes #8267 .

### Description

Fix channel-wise intensity normalization for integer type inputs. 

### Types of changes

### CODE REVIEW

File: monai/transforms/intensity/array.py [hunk 1]
```diff
@@ -907,6 +907,8 @@ def __call__(self, img: NdarrayOrTensor) -> NdarrayOrTensor:
             if self.divisor is not None and len(self.divisor) != len(img):
                 raise ValueError(f"img has {len(img)} channels, but divisor has {len(self.divisor)} components.")
 
+            img, *_ = convert_data_type(img, dtype=torch.float32)
```
KumoLiu, line 910: Thanks for the fix. Seems we can remove the convert out side of the `self._normalize` to avoid convert multiple times here.

ericspod: Should we also check that the dtype of `img` isn't float16/float32/float64 and only do the conversion in that case?

KumoLiu: Yes, it would be a little more rigorous. Although in the non channel-wise version, we do the conversion directly. But yes if we could do the check, would be better.

ericspod:
```suggestion
            if not img.dtype.is_floating_point:
                img, *_ = convert_data_type(img, dtype=torch.float32)
```
Something like this?

advcu987:
@KumoLiu , @ericspod  can this be merged?

KumoLiu:
Hi @advcu987, could you please try to add some comments based on @ericspod's comments? Then I could help merge the pr, thanks.

PR #8301, opened 2025-01-15
Add Transform to Subtract Images

anamatoso:
Hello all,

I've created a transform to subtract two images. 
I've created the tests and passed all integration tests.
Hope you find it useful.

Let me know if there is anything else still to do.

### Description

Adds a transform for subtracting images.

### Types of changes


anamatoso:
Hi @ericspod.

I guess it is indeed specific to my case as I used it to highlight longitudinal changes between images. I don't know whether other arithmetic functions would be useful, but I agree with you that a generic function that aggregates all these callables is perhaps the way to go. However, in the meantime, I won't have much time to implement that. So, if you agree, I would prefer to commit just this for now, as it does not break existing functionality. Then, later down the line, I could write the aggregator function that would replace this one and ConcatItemsd.  What do you think?

ericspod:
> Hi @ericspod.
> 
> I guess it is indeed specific to my case as I used it to highlight longitudinal changes between images. I don't know whether other arithmetic functions would be useful, but I agree with you that a generic function that aggregates all these callables is perhaps the way to go. However, in the meantime, I won't have much time to implement that. So, if you agree, I would prefer to commit just this for now, as it does not break existing functionality. Then, later down the line, I could write the aggregator function that would replace this one and ConcatItemsd. What do you think?

Hi @anamatoso, I agree that there's a specific use case for subtraction but much less so for other operations, I'm still concerned about not adding code we wouldn't expect to keep in the code base since we have to follow our deprecation process. This is minor code so not a large problem, if @KumoLiu or @Nic-Ma have any input I'd like to hear but otherwise I think  this is good. 

For discussion, I would have suggested something like the following to add a lambda transform to apply to multiple members together:

```python
class LambdaItemsd(MapTransform):
    """
    Apply a callable to specified items from data dictionary together. This will collect the items together into a list
    and pass this as the single argument to the supplied callable. If an output name is given, this will key to the call
    result in the produced dictionary, otherwise the result value is discarded.
    """

    def __init__(
        self,
        keys: KeysCollection,
        output_name: str | None,
        func: Callable | None = None,
        track_meta: bool = True,
    ) -> None:
        """
        Args:
            keys: keys of the corresponding items to be passed to `func` as a single list.
                See also: :py:class:`monai.transforms.compose.MapTransform`
            output_name: the output_name corresponding to the key to store the result data.
            func: function to apply to collected values.
            track_meta: if true, the result of the function application will be converted to a MetaTensor.
        """
        self.output_name = output_name
        self._lambd = Lambda(func=func, track_meta=track_meta)

    def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> dict[Hashable, NdarrayOrTensor]:
        """ """
        d = dict(data)
        inputs = [d[key] for key in self.key_iterator(d)]

        result = self._lambd(inputs)
        if self.output_name is not None:
            d[self.output_name] = result

        return d
    
test={"a":torch.rand(1,64,64),"b":torch.rand(1,64,64)}
t=LambdaItemsd(["a","b"],"out",lambda items:items[0]+items[1])
result=t(test)
print(result.keys())
```

Issue #8303, opened 2025-01-17
Dependency Version/Deprecation Policy


ericspod:
MONAI has an existing policy for deprecating components it defines [here](https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md#backwards-compatibility). For dependencies there isn't as clear a policy to follow in regards to which versions of Python, Numpy, Pytorch, etc. to support. It's important to support older environments as much as is practical so that users with frozen environments have the fewest issues with new versions of MONAI. There are a few considerations that vary by library which need to be composed into a clear policy:

* [Python versions of have end-of-life dates](https://devguide.python.org/versions/) and so we should support only those versions that haven't reached that date.
* Pytorch's releases have for a long time never been out of sync, that is there have been no bug fixes for old versions for a very long time, and so old versions immediately become out-of-support. This is an issue with known security vulnerabilities in older versions which won't be fixed, so #8296 has been proposed to remove old versions from the CICD system. A subsequent PR will remove more versions further as discussed [here ](https://github.com/Project-MONAI/MONAI/pull/8296#issuecomment-2587338931).
* Numpy has released version 2.x which introduced some incompatibilities with other dependencies, though MONAI itself should be compatible. This has created the situation where we have to specify that MONAI requires 1.x to prevent pip from installing incompatible combinations of libraries. At some point this will be fixed and the version restriction can be removed.
* Other dependencies may deprecate versions, or vulnerabilities discovered for them.

All this means we need to determine:
* How old are the versions of dependencies we want to support.
* How to decide when to remove support for certain dependency versions and by whom.
* How to respond to discovered vulnerabilities.
* How and when to re-analize the dependencies that we have to determine if version requirements are still suitable.
* Who may add or remove dependencies, even optional ones.
* Who may mark other repositories as archived or de-archive them (I'm thinking specifically of archiving [GenerativeModels](https://github.com/Project-MONAI/GenerativeModels/)).

jamesobutler:
> This has created the situation where we have to specify that MONAI requires 1.x to prevent pip from installing incompatible combinations of libraries.

[...]

PR #8309, opened 2025-01-20
Enable Pytorch 2.6

ericspod:
Partially addresses #8303.

### Description

This changes the maximum Numpy version to be below 3.0 for testing with 2.x compatibility. This appears to be resolved with newer versions of dependencies. This will also include fixes for Pytorch 2.6 mostly relating to `torch.load` and `autocast` usage.

### Types of changes


jamesobutler (commented):
@ericspod I would suggest updating the automated workflow so that when it tests older torch versions (pre-2.3.0 which don’t have numpy 2 support), to `pip install numpy<2` just before the torch install linked below. This should keep this numpy 1 version as torch is installed and as well as when MONAI gets installed afterwards. With this PR’s change, having numpy<2 already installed with an older torch so fine and will meet the install_requires for MONAI when it goes to get installed and won’t force the upgrade of them.

https://github.com/Project-MONAI/MONAI/blob/df1ba5d1e6aa9a0a1744b7ae3ff37ca114cec7bb/.github/workflows/pythonapp-min.yml#L157

ericspod:
> @ericspod I would suggest updating the automated workflow so that when it tests older torch versions (pre-2.3.0 which don’t have numpy 2 support), to `python -m pip install numpy<2`

Thanks for the suggestion, that would fix this PR however we wouldn't resolve the version issue during installation. Users would still need to pre-install Numpy like this. This PR was more an experiment to see what version updates we can do and where the pain points were. We're still discussing the version policy which will change what we do here as well, and hopefully resolve this Numpy issue.

ericspod:
We probably need to merge #8368 first given the test results.

jamesobutler:
Based on the current title of this PR, the intention is to limit numpy to only numpy 2 instead of supporting numpy 1 and 2?

ericspod:
> Based on the current title of this PR, the intention is to limit numpy to only numpy 2 instead of supporting numpy 1 and 2?

Actually just an outdated title, I've kept your previous PR's changes so it's just to do with Pytorch 2.6 now. Fixed now, thanks.

jamesobutler:
This PR should then remove the upper pin found in the requirements.txt to allow PyTorch 2.6. The diff in this PR currently doesn't include those changes:
https://github.com/Project-MONAI/MONAI/blob/2e391c82d9e5fc565917ac46fdeb3a900c96acaf/requirements.txt#L1-L2

ericspod:
> This PR should then remove the upper pin found in the requirements.txt to allow PyTorch 2.6. The diff in this PR currently doesn't include those changes:
> 
> https://github.com/Project-MONAI/MONAI/blob/2e391c82d9e5fc565917ac46fdeb3a900c96acaf/requirements.txt#L1-L2

Good spot thanks!

monai/networks/nets/swin_unetr.py
```diff
@@ -272,53 +272,50 @@ def __init__(
         self.out = UnetOutBlock(spatial_dims=spatial_dims, in_channels=feature_size, out_channels=out_channels)
 
     def load_from(self, weights):
+        layers1_0: BasicLayer = self.swinViT.layers1[0]  # type: ignore[assignment]
```
KumoLiu, line 275: May I ask the reason for this change here?

KumoLiu (approved):
Only one minor comment.

monai/networks/nets/swin_unetr.py
ericspod: This was to try and force better typing on the layers coming from the swinViT object. These get typed as tensor or module which isn't correct as it should be BasicLayer. This isn't the perfect fix for the typing issue but it's a little better, we'll revist typing as discussed.

KumoLiu: Got it, thanks for the explaination!

KumoLiu:
Will fix https://github.com/Project-MONAI/MONAI/issues/8355

PR #8313, opened 2025-01-24
Fix deprecated usage in zarr

### LINKED ISSUE #8298: AttributeError: module 'zarr.storage' has no attribute 'DirectoryStore'
```
  File "/home/runner/work/MONAI/MONAI/tests/runner.py", line 135, in <module>
    tests = unittest.TestLoader().loadTestsFromNames(cases)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.11/x64/lib/python3.11/unittest/loader.py", line 232, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.11/x64/lib/python3.11/unittest/loader.py", line 232, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.11/x64/lib/python3.11/unittest/loader.py", line 162, in loadTestsFromName
    module = __import__(module_name)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/MONAI/MONAI/tests/test_zarr_avg_merger.py", line 157, in <module>
    dict(merged_shape=TENSOR_4x4.shape, store=zarr.storage.DirectoryStore("test.zarr")),
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'zarr.storage' has no attribute 'DirectoryStore'
```

KumoLiu:
Fixes #8298

### Types of changes

### DIFF
--- a/monai/inferers/merger.py
+++ b/monai/inferers/merger.py
@@ -247,6 +247,7 @@ class ZarrAvgMerger(Merger):
         self.value_dtype = value_dtype
         self.count_dtype = count_dtype
         self.store = store
+        self.tmpdir: TemporaryDirectory | None
         if version_geq(get_package_version("zarr"), "3.0.0"):
             if value_store is None:
                 self.tmpdir = TemporaryDirectory()
--- a/monai/inferers/merger.py
+++ b/monai/inferers/merger.py
@@ -15,12 +15,13 @@ import threading
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from contextlib import nullcontext
+from tempfile import TemporaryDirectory
 from typing import TYPE_CHECKING, Any
 
 import numpy as np
 import torch
 
-from monai.utils import ensure_tuple_size, optional_import, require_pkg
+from monai.utils import ensure_tuple_size, get_package_version, optional_import, require_pkg, version_geq
 
 if TYPE_CHECKING:
     import zarr
@@ -233,7 +234,7 @@ class ZarrAvgMerger(Merger):
         store: zarr.storage.Store | str = "merged.zarr",
         value_store: zarr.storage.Store | str | None = None,
         count_store: zarr.storage.Store | str | None = None,
-        compressor: str = "default",
+        compressor: str | None = None,
         value_compressor: str | None = None,
         count_compressor: str | None = None,
         chunks: Sequence[int] | bool = True,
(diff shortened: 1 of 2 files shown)


monai/inferers/merger.py
```diff
@@ -246,8 +247,20 @@ def __init__(
         self.value_dtype = value_dtype
         self.count_dtype = count_dtype
         self.store = store
-        self.value_store = zarr.storage.TempStore() if value_store is None else value_store
-        self.count_store = zarr.storage.TempStore() if count_store is None else count_store
+        if version_geq(get_package_version("zarr"), "3.0.0"):
+            if value_store is None:
+                with TemporaryDirectory() as tmpdir:
```
ericspod, line 252: `TemporaryDirectory` should delete the directory once the context block if left, is this a problem here? It might not be immediate but we can't rely on it existing outside the block I don't think. If you store an instance of `TemporaryDirectory` as an attribute it will get cleaned up when the object is collected.

tests/test_zarr_avg_merger.py
```diff
@@ -289,13 +289,17 @@ class ZarrAvgMergerTests(unittest.TestCase):
     def test_zarr_avg_merger_patches(self, arguments, patch_locations, expected):
         if "compressor" in arguments:
             if arguments["compressor"] != "default":
-                arguments["compressor"] = zarr.codec_registry[arguments["compressor"].lower()]()
+                arguments["compressor"] = numcodecs.registry.codec_registry[arguments["compressor"].lower()]()
         if "value_compressor" in arguments:
             if arguments["value_compressor"] != "default":
-                arguments["value_compressor"] = zarr.codec_registry[arguments["value_compressor"].lower()]()
+                arguments["value_compressor"] = numcodecs.registry.codec_registry[
```
ericspod, line 295: I'd suggest putting `codec_reg = numcodecs.registry.codec_registry` at the top of this method then using that variable in the three places the registry is needed just to shorten the lines a bit.

KumoLiu: good point, updated it in the latest commit. Thanks!

monai/inferers/merger.py
KumoLiu: Thanks! should be an issue, updated it in the latest commit.

monai/inferers/merger.py
```diff
@@ -246,8 +247,20 @@ def __init__(
...
+                tmpdir = TemporaryDirectory()
+                self.value_store = zarr.storage.LocalStore(tmpdir.name)
+            else:
+                self.value_store = value_store
+            if count_store is None:
+                tmpdir = TemporaryDirectory()
+                self.count_store = zarr.storage.LocalStore(tmpdir.name)
+            else:
+                self.count_store = count_store
+        else:
```
ericspod, line 261:
```suggestion
        if version_geq(get_package_version("zarr"), "3.0.0"):
            if value_store is None:
                self.tmpdir = TemporaryDirectory()
                self.value_store = zarr.storage.LocalStore(self.tmpdir.name)
            else:
                self.value_store = value_store
            if count_store is None:
                self.tmpdir = TemporaryDirectory()
                self.count_store = zarr.storage.LocalStore(self.tmpdir.name)
            else:
                self.count_store = count_store
        else:
            self.tmpdir = None
```

What I meant was that the lifecycle of the tmpdir object should follow that of the containing object since this will be the same as that for the storage objects.

PR #8323, opened 2025-02-02
modified videodataset for window multiprocessing

### LINKED ISSUE #8322: VideoDataset Deadlock and error
**Describe the bug**

OpenCV VideoCapture object deadlocks the data loader or throws different errors when multiple workers try to access the same video file. The current logic implemented in the dataset is workable for Linux env but not Windows. The self.multiprocessing attribute doesn't prevent num_worker > 0 from being handled either. 

**To Reproduce**
dataset = VideoFileDataset(video_source = "video.mp4", multiprocessing = True)
dataloader = torch.utils.data.Dataloader(dataset, batch_size = 2, num_worker = 2)
for batch in dataloader:
print(hello)

**Expected behaviour**
Dataloader will deadlock when trying to create the same cv2.VideoCapture object on same video (behaviour could be different in different video files)

**Screenshots**

**Environment**

Windows
Pyhton > 3.9

JK-rez:
Fixes #8322

### Description

Slight workaround to VideoCapure behaviours when num worker > 0 in dataloader.
Explicitly assign video capture object to release if after reading frame. Should prevent deadlock and other cv2 errors from occurring.
Not tested exhaustively as current pipelines in CV for video pre process videos into frame beforehand. Current dataclass not scalable 

### Types of changes

- [] Non-breaking change (fix or new feature that would not break existing functionality).
- [x ] Breaking change (fix or new feature that would cause existing functionality to change).

### DIFF
--- a/monai/data/video_dataset.py
+++ b/monai/data/video_dataset.py
@@ -140,7 +140,13 @@ class VideoDataset:
     def get_frame(self) -> Any:
         """Get next frame. For a file, this will be the next frame, whereas for a camera
         source, it will be the next available frame."""
-        ret, frame = self._get_cap().read()
+        #Assign explicit videocapture object to cap to realease file to avoid error in multiprocessing
+        if self.multiprocessing:
+            cap = self._get_cap()
+            ret, frame = cap.read()
+            cap.release()
+        else:   
+            ret, frame = self._get_cap().read()
         if not ret:
             raise RuntimeError("Failed to read frame.")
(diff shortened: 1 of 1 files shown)


ericspod:
Thanks for the update. Could we add a test to capture this issue somehow? The problem seems to be multiprocessing in Windows from what we discussed, can we add a unit test using multiprocessing to ensure this works in Windows and elsewhere? Thanks!

PR #8325, opened 2025-02-04
selfattention block: Remove the fc linear layer if it is not used

johnzielke:
### Description
when include_fc = False, the nn.Linear layer is unused. This leads to errors and warning when training with the pytorch Distributed Data Parallel infrastructure, since the parameters for the nn.Linear layer will not have gradients attached.
### Types of changes


ericspod:
Thanks for the contribution! In itself I think it's fine however we have to check that this won't break old weights. We have this method [load_old_state_dict](https://github.com/Project-MONAI/MONAI/blob/8dcb9dc0e594059d87ce9882c20fe5a59340a6b2/monai/networks/nets/diffusion_model_unet.py#L1807) for doing this with DiffusionModelUNet that might not work if `out_proj` doesn't have weight or bias components. There are other `load_old_state_dict` methods doing this for other networks that should be looked at. 

We still want to maintain backwards compatibility with old stored weights at least for now, but we should discuss about when to deprecate these methods. CC @virginiafdez

johnzielke:
If that's the only concern, I could update that method to ignore that key in the appropriate cases

ericspod:
> If that's the only concern, I could update that method to ignore that key in the appropriate cases

Please do have a look at how old state is loaded and see if there's any issues, otherwise yes we should be good here. I've updated your branch after we've done a lot of test refactoring, we should perhaps also include a test that checks the network does or does not have the fc layer when appropriate.

johnzielke:
I pushed the discussed changes. I wanted to test the load_old_state_dict as well, but it seems there is no test yet that loads the code without cross-attention. I did not dive all the way into where the old state dicts are stored etc. Is there an easy to use old state dict I could use in the `test_compatibility_with_monai_generative()` test?

KumoLiu:
The `load_old_state_dict` function is likely not a concern here, as the default value for `include_fc` is set to True. This matches the requirement when utilizing the old state dictionary, which also necessitates that `include_fc` remains True.

KumoLiu (approved):

PR #8343, opened 2025-02-17
Inferer modification - save_intermediates clashes with latent shape adjustment in latent diffusion inferers

### LINKED ISSUE #8334: LatentDiffusionInferer errors out when save_intermediates = False and padding and cropping is needed
**Describe the bug**
The latent space of the autoencoder can be padded (to avoid it not being multiple of 2**(num_downsamplings) on the DM). When save_intermediates is False, the code still tries to crop them, even though they aren't created.

**To Reproduce**
- Create diffusion model
- Create inferer with ldm_latent_shape and autoencoder_latent_shape different
- Run the sample function with save_intermediates = False.

**Expected behavior**
No error. 

**Environment**
Python 3.9
MONAI 1.4.0

virginiafdez:
Fixes #8334 

### Description

There was an if save_intermediates missing in the code that was trying to run crop of the latent spaces on the sample function of the Latent Diffusion Inferers (normal one and ControlNet one) even when intermediates aren't created. 

### Types of changes


ericspod:
Thanks for the fix, could we add a unit test to capture this behaviour? We should have one test at least that would fail without this fix but works with it I think.

KumoLiu (approved):
Thanks for the fix, LGTM.

monai/inferers/inferer.py
```diff
@@ -1211,8 +1211,10 @@ def sample(  # type: ignore[override]
         decode = autoencoder_model.decode_stage_2_outputs
         if isinstance(autoencoder_model, SPADEAutoencoderKL):
             decode = partial(autoencoder_model.decode_stage_2_outputs, seg=seg)
+        print("Decoding latents...")
+        print(latent.shape)
```
KumoLiu, line 1215:
```suggestion
```

monai/inferers/inferer.py
```diff
@@ -1211,8 +1211,10 @@ def sample(  # type: ignore[override]
...
         image = decode(latent / self.scale_factor)
-
+        print(image.shape)
```
KumoLiu, line 1217:
```suggestion

```

ericspod (approved):

PR #8351, opened 2025-02-18
8201 Fix DataFrame subsets indexing in CSVDataset()

### LINKED ISSUE #8201: CSVDataset behaves unexpectedly if src is a dataframe unexpected index
**Describe the bug**
CSVDataset accepts pandas DataFrames as input for src. But it makes assumptions about the index.

This is because `convert_tables_to_dicts` uses `.loc` instead of `.iloc`. It generates ordinal indexes to subset on but treats them as names indices.

https://github.com/Project-MONAI/MONAI/blob/0bb20a88ec7869f6453aa58890df50ad6b2b6271/monai/data/utils.py#L1494

**To Reproduce**
```
import numpy
import pandas
import monai

df = pandas.DataFrame(numpy.random.random((50, 3)))
df_subset = df.iloc[numpy.arange(0, 50, 5)]
print(df_subset.shape)  # (10, 3)

ds = monai.data.CSVDataset(df_subset)
print(len(ds))  # 3
```

**Expected behavior**
`print(len(ds))` should return 10.
It returns 3 because it looks up indices slice(10), which match indices 0, 5 and 10 from the subset.

**Environment**
Shouldn't be relevant?

**Additional context**
Simple fix: https://github.com/Project-MONAI/MONAI/blob/0bb20a88ec7869f6453aa58890df50ad6b2b6271/monai/data/utils.py#L1494

The first .loc should be .iloc, and the second should be .iloc[rows][col_names]

bartosz-grabowski:
Fixes #8201 .

### Description

`convert_tables_to_dicts` was using `.loc` to index DataFrames which was changed to `.iloc`. It was causing unexpected behavior in `CSVDataset` as demonstrated in #8201 because `.loc` expects labels, but was instead provided positions of the rows.
Unittest was added which fails before the change and passes afterwards.

### Types of changes

### DIFF
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -1500,7 +1500,7 @@ def convert_tables_to_dicts(
     if col_groups is not None:
         groups: dict[str, list] = {}
         for name, cols in col_groups.items():
-            groups[name] = df.loc[rows, cols].values
+            groups[name] = df.iloc[rows][cols].values
         # invert items of groups to every row of data
         data = [dict(d, **{k: v[i] for k, v in groups.items()}) for i, d in enumerate(data)]
 
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -1484,7 +1484,7 @@ def convert_tables_to_dicts(
                 rows.append(i)
 
     # convert to a list of dictionaries corresponding to every row
-    data_ = df.loc[rows] if col_names is None else df.loc[rows, col_names]
+    data_ = df.iloc[rows] if col_names is None else df.iloc[rows][col_names]
     if isinstance(col_types, dict):
         # fill default values for NaN
         defaults = {k: v["default"] for k, v in col_types.items() if v is not None and v.get("default") is not None}
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -1473,7 +1473,7 @@ def convert_tables_to_dicts(
     # parse row indices
     rows: list[int | str] = []
     if row_indices is None:
-        rows = slice(df.shape[0])  # type: ignore
+        rows = df.index.tolist()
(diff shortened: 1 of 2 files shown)


bartosz-grabowski:
@KumoLiu @ericspod could you take a look at this?

ericspod (approved):
I think this is good to go, we're waiting on a large PR to be merged, then we'll update this one and move forward.

monai/data/utils.py
```diff
@@ -1484,7 +1484,7 @@ def convert_tables_to_dicts(
                 rows.append(i)
 
     # convert to a list of dictionaries corresponding to every row
-    data_ = df.loc[rows] if col_names is None else df.loc[rows, col_names]
+    data_ = df.iloc[rows] if col_names is None else df.iloc[rows][col_names]
```
KumoLiu, line 1487: Hi @bartosz-grabowski, there might be a problem with updating in this manner. The `.loc` method relies on a label index, whereas `.iloc` uses a positional index. So it may not works when rows are generated with specific `row_indices`.

```
import numpy
import pandas
import monai

df = pandas.DataFrame(numpy.random.random((50, 3)))
df_subset = df.iloc[numpy.arange(0, 50, 2)]

row_indices = [2, 10, 40]
df_subset.iloc[row_indices]
# Throw out-of-bounds index error.
```

What do you think? also cc @ericspod

bartosz-grabowski: I think you are right. I updated the code so that it should now fix the original issue while also working with `row_indices != None`.

KumoLiu: looks good to me now, thanks for the quick update.

KumoLiu (approved):

PR #8360, opened 2025-02-21
Modify ControlNet inferer so that it takes in context when the diffus…

### LINKED ISSUE #8344: ControlNetLatentDiffusionInferer errors out when conditioning is not False
**Describe the bug**
ControlNets are essentially copies of the diffusion model encoders. However, on the ControlNetLatentDiffusionInferer, the context (conditioning) isn't passed to the ControlNet even though the DiffusionModel might take it in. If ControlNet has been defined with conditioning, this causes an error. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Create DiffusionModel and ControlNet with conditioning = True.
2. Create ControlNetDiffusionInferer 
3. Run the call() function on the inferer passing dummy random tensors. 

**Expected behaviour**
ControlNet should take in the conditioning tensor.

virginiafdez:
Fixes #8344 

### Description
The ControlNet inferers (latent and not latent) work in such a way that, when conditioning is used, the ControlNet does not take in the conditioning. It should, in theory, exhibit the same behaviour as the diffusion model. 
I've changed this behaviour, which has included modifying ControlNetDiffusionInferer and ControlNetLatentDiffusionInferer; the methods call, sample and get_likelihood. 
I've also modified the tests to take this into account. 

### Types of changes

- [x] New tests added to cover the changes (modified, rather than new)

### DIFF
--- a/monai/inferers/inferer.py
+++ b/monai/inferers/inferer.py
@@ -1334,12 +1334,14 @@ class ControlNetDiffusionInferer(DiffusionInferer):
             raise NotImplementedError(f"{mode} condition is not supported")
 
         noisy_image = self.scheduler.add_noise(original_samples=inputs, noise=noise, timesteps=timesteps)
-        down_block_res_samples, mid_block_res_sample = controlnet(
-            x=noisy_image, timesteps=timesteps, controlnet_cond=cn_cond
-        )
+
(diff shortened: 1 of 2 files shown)


ericspod (approved):
LGTM!

PR #8388, opened 2025-03-12
Update Dice Metric Docs

### LINKED ISSUE #8385: Update Dice Metric/Loss Documentation
The docstring descriptions for `DiceLoss`, `DiceMetric`, `DiceHelper`, and `compute_dice` should include more information about the expectations of inputs, implications of various arguments, and better examples. There's been some confusion about why results from the loss don't match expectations given the value of metrics, this is related to the extra features in the loss such as the smoothing values as well as misunderstanding about activation or thresholding/one-hot formatting. 

A few things should be changed or added:
* Docstring for `DiceMetric` should make it clear the ground truth and prediction can be either single-channel label maps or multi-channel one-hot tensors.
* The values of the inputs do matter since no thresholding is done in the metrics themselves.
* The `softmax` parameter of `DiceHelper` is confusing in that it appears to indicate `softmax` will be applied if it's true, rather than the expectation that this was already done to the prediction. This should be renamed to `argmax` and the old name deprecated. It should also be noted that this parameter becomes the opposite value to `sigmoid` if not given a value.
* Using `compute_dice` probably should be discouraged in favour of the `DiceMetric` class, and it should be mentioned that they do the same computation. It's possible this function isn't used by anyone and can be deprecated entirely.
* An example in the docstring of how to use the metric and the loss to get expected results should be included.

ericspod:
Fixes #8385.

### Description

This updates the docstrings for the Dice metric classes and tweak some constructor parameter names to improve comprehension.

### Types of changes

### DIFF
--- a/monai/metrics/meandice.py
+++ b/monai/metrics/meandice.py
@@ -40,6 +40,29 @@ class DiceMetric(CumulativeIterationMetric):
 
     An example of the typical execution steps of this metric class follows :py:class:`monai.metrics.metric.Cumulative`.
 
+    Example:
+
+    .. code-block:: python
+
+        import torch
+        from monai.metrics import DiceMetric
+
+        batch_size, n_classes, h, w = 7, 5, 128, 128
+
+        y_pred = torch.rand(batch_size, n_classes, h, w)  # network predictions
+        y_pred = torch.argmax(y_pred, 1, True)  # convert to label map
+
+        # ground truth as label map
+        y = torch.randint(0, n_classes, size=(batch_size, 1, h, w))  
+
+        dm = DiceMetric(
+            reduction="mean_batch", return_with_label=True, num_classes=n_classes
+        )
+
+        raw_scores = dm(y_pred, y)
+        print(dm.aggregate())
+
+
     Args:
         include_background: whether to include Dice computation on the first channel/category of the prediction and
             ground truth. Defaults to ``True``, use ``False`` to exclude the background class.  
--- a/monai/metrics/meandice.py
+++ b/monai/metrics/meandice.py
@@ -23,35 +23,39 @@ __all__ = ["DiceMetric", "compute_dice", "DiceHelper"]
 
 class DiceMetric(CumulativeIterationMetric):
     """
-    Compute average Dice score for a set of pairs of prediction-groundtruth segmentations.
+    Computes Dice score for a set of pairs of prediction-groundtruth labels. It supports single-channel label maps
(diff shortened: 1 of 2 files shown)


monai/metrics/meandice.py
```diff
@@ -188,49 +228,65 @@ class DiceHelper:
...
     """
 
+    @deprecated_arg("softmax", "1.5", "1.7", "Use `apply_argmax` instead.", new_name="apply_argmax")
+    @deprecated_arg("sigmoid", "1.5", "1.7", "Use `threshold` instead.", new_name="threshold")
     def __init__(
         self,
         include_background: bool | None = None,
-        sigmoid: bool = False,
-        softmax: bool | None = None,
+        threshold: bool = False,
```
KumoLiu, line 256: Do you think we can also support float for this threshold if user want to use another threshold instead of 0.5?

monai/metrics/meandice.py
```diff
@@ -77,22 +118,21 @@ def __init__(
...
     def _compute_tensor(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor:  # type: ignore[override]
         """
+        Compute the dice value using ``DiceHelper``.
+
         Args:
-            y_pred: input data to compute, typical segmentation model output.
-                It must be one-hot format and first dim is batch, example shape: [16, 3, 32, 32]. The values
-                should be binarized.
-            y: ground truth to compute mean Dice metric. `y` can be single-channel class indices or
-                in the one-hot format.
```
KumoLiu, line 92: These restrictions are not really necessary, such as binarized, or one-hot, etc, right?

monai/metrics/meandice.py
ericspod: We could, the purpose here was to threshold values that were activated with sigmoid already and so that would be the sensible one to use. It would be a minor change here but isn't use the metric itself actually, unless we want to change it as well.

KumoLiu: Got it, so let's leave it as it for now unless there are any requirement.

monai/metrics/meandice.py
ericspod: This is coded into the way the `DiceHelper` class computes the metrics on a per-item and per-channel basis so it is expected. [Here](https://github.com/Project-MONAI/MONAI/blob/90de55b1340ffc8eccf960d111229535bcb2f81d/monai/metrics/meandice.py#L274) and the next line down there's assumptions about the shape of the incoming tensors.

KumoLiu (approved):
LGTM.
@virginiafdez, do you have any comments on this one?

Issue #8220, opened 2024-11-19
SliceInferer may need to handle conditional-based models


SanBast:
**Is your feature request related to a problem? Please describe.**
Lately I'm working with LDMs to generate samples conditioned by a specific conditional vector. My model learned to generate samples in an axial-2D fashion. The conditioning embeddings are based on 2D axial slices as well. Following the tutorial [here](https://github.com/Project-MONAI/tutorials/blob/1dfa20ca373a582f404856b3ff509958655b183f/modules/2d_inference_3d_volume.ipynb), I noticed that [**SliceInferer**](https://docs.monai.io/en/stable/inferers.html) might come in handy. However, as the documentation suggests, you can just teak a little bit the network's method you will use for inference, and not the external conditions. What's happening in the SliceInferer is, as the name suggests, passing the input slice-by-slice to the considered network, yet the condition is not processed in the same way.

**Describe the solution you'd like**
It would be nice to adjust the code to allow models' conditioning. An idea can be to add a "forced" parameter in the SliceInferer (e.g., "condition") that may call another SliceWindowInferer dedicated to the condition.

**Describe alternatives you've considered**

**Additional context**
Given the number of axial slices Z and a bottleneck output that matches of my network that matches the dimensionality D of the conditional vector, the main error behind using SliceInferer is that the input model expects a [1, D] but what it got is a [Z,D] of course.

Thanks for helping me out!

PR #8400, opened 2025-03-25
Add support for optional conditioning in PatchInferer, SliceInferer, and SlidingWindowInferer

FinnBehrendt:
Fixes [#8220](https://github.com/Project-MONAI/MONAI/issues/8220)

### Description
This PR adds support for optional conditioning in MONAI’s inferers, allowing models to receive auxiliary inputs for conditioning that are processed (patched, sliced) the same way as the inputs. This is particularly relevant for generative models like conditional GANs or DMs.

Example Usage:
```python
# Given a conditioned model, inputs of shape (1, C, H, W, D) and condition of shape (1, C, H, W, D)
output = SliceInferer(...)(inputs, model, condition=cond_tensor)
```

### Types of changes
- Extended `PatchInferer`, `SliceInferer`, and `SlidingWindowInferer` to optionally accept a `condition` tensor (passed as a kwarg).
- The `condition` can now be:
  - `None` (default)
  - A tensor of the same shape as `inputs`
- The inferers now slice/patch the conditions alongside the corresponding inputs and feed them to the network.
- Updated unit tests for each inferer:
  - Verified with and without conditioning

Additional extensions such as support for dense vector conditioning (e.g., (1, C, Z), with Z being the conditional dimension) could be explored in a follow-up PR if there’s interest.

## Summary by CodeRabbit

* **New Features**
  * Added support for an optional "condition" tensor in patch-based, sliding window, and slice inference, allowing conditional inference workflows.
  * The "condition" tensor is validated for shape and type consistency with inputs and is processed in sync during inference.

* **Tests**
  * Introduced extensive new tests for conditional inference across patch, sliding window, and slice inferers to ensure correct behavior and output validation when using the "condition" argument.


monai/inferers/inferer.py
```diff
@@ -344,20 +347,39 @@ def __call__(
...
-            # initialize the mergers
-            if not mergers:
-                mergers, ratios = self._initialize_mergers(inputs, outputs, patches, batch_size)
-            # aggregate outputs
-            self._aggregate(outputs, locations, batch_size, mergers, ratios)
+        if condition is not None:
+            for (patches, locations, batch_size), (condition_patches, _, _) in zip(
+                self._batch_sampler(patches_locations), self._batch_sampler(condition_locations)
+            ):
+                # add patched condition to kwargs
```
virginiafdez, line 365: Will this require condition to have the same shape as the input? in that case, if condition is None, perhaps there should be size check.

virginiafdez (changes requested):
The PR looks good to me overall, with appropriate tests that pass. Not sure if it'll be better to derive the condition from an argument defaulted to None instead of kwargs, to be more transparent. If leaving to kwargs, I believe there should still be some mention on the dosctrings that it is possible to pass a "condition". Not sure about this either, but  If the shape of both input and output have to match, then checking for this at the beginning of the inferer sampler would be good (+test).

virginiafdez (approved):

PR #8403, opened 2025-03-26
8394 Update bundle download API

### LINKED ISSUE #8394: Update bundle download API to align with model zoo changes
The bundle hosting change is that any new bundles/versions will be hosted in Hugging Face.
take `maisi_ct_generative` for example, it's v0.4.6 is in:
https://api.ngc.nvidia.com/v2/models/nvidia/monaihosting/maisi_ct_generative/versions/0.4.6/files/maisi_ct_generative_v0.4.6.zip
it's latest v1.0.0 is in:
https://huggingface.co/MONAI/maisi_ct_generative/tree/1.0.0

Therefore, MONAI bundle download API should also be updated. For each bundle (each version), the API should be able to download not matter which host it is on. To enable it, we should define the priority of download: NGC or Hugging Face, which  host should be accessed first?

Currently, most of the bundles (most of the versions) are in NGC, and only a few of them in Hugging Face, but in the future, new versions will all be in Hugging Face.

My suggestion is that we can prepare a new version for each bundle, to ensure the latest versions are all in Hugging Face. In the API side, let the program try to access to Hugging Face first.

Hi @Nic-Ma @ericspod @KumoLiu , could you help to give some suggestions?

Related model zoo PR: related model zoo changes: https://github.com/Project-MONAI/model-zoo/pull/735

yiheng-wang-nv:
Fixes #8394  .

### Description

Add support to download monaihosting bundles from Huggingface.

### Types of changes


yiheng-wang-nv:
This PR can be merged after https://github.com/Project-MONAI/model-zoo/issues/739 is solved

monai/bundle/scripts.py
```diff
@@ -600,11 +600,15 @@ def download(
             _download_from_github(repo=repo_, download_path=bundle_dir_, filename=name_ver, progress=progress_)
         elif source_ == "monaihosting":
             try:
+                extract_path = os.path.join(bundle_dir_, name_)
```
mingxin-zheng, line 603: Would it make more sense if we use the huggingface cache path, instead of the MONAI cache path?

yiheng-wang-nv: Hi @mingxin-zheng , I followed: https://github.com/Project-MONAI/MONAI/pull/8403/files#diff-a715c9207f1e727d7a6ed7c6b25dfb57045d4429fc0b4eed3f5396e26ab34f80L630
maybe we can update to hf cache path in next version, what do you think?

mingxin-zheng: I guess this can stay open to discussion. Maybe we should hear more from the users. I'll closing the comment here

mingxin-zheng (approved):

monai/bundle/scripts.py
KumoLiu, line 1: Thanks for the PR.
If I understand correctly, when the source is set to "monaihosting," we will initially check the Hugging Face Hub, which should yield the same result as when the source is "huggingface_hub." Is that correct?

If so, please include this clarification in the docstring. Thanks!
Other part looks good to me.

yiheng-wang-nv: Hi @KumoLiu , the required args are different. monaihosting bundles need `name`, and `huggingface_hub` bundles need `repo_`

yiheng-wang-nv: pushed a new commit based on discussions offline

KumoLiu (approved):

monai/bundle/scripts.py
KumoLiu: LGTM, thanks for the update.

PR #8411, opened 2025-04-04
Initial commit to make codebase robust to use of non-standard meta_dict names

atbenmurray:
This PR adds a field to the meta dict indicating which tensor it belongs to, and a method get_meta_dict_key to find the meta_dict when it is present in a dictionary.

### Description

### Types of changes

### DIFF
--- a/monai/transforms/intensity/dictionary.py
+++ b/monai/transforms/intensity/dictionary.py
@@ -24,7 +24,7 @@ import numpy as np
 
 from monai.config import DtypeLike, KeysCollection
 from monai.config.type_definitions import NdarrayOrTensor
-from monai.data.meta_obj import get_track_meta
+from monai.data.meta_obj import get_meta_dict_name, get_track_meta
 from monai.transforms.intensity.array import (
     AdjustContrast,
     ClipIntensityPercentiles,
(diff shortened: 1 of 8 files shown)


monai/transforms/io/dictionary.py
```diff
@@ -169,6 +169,7 @@ def __call__(self, data, reader: ImageReader | None = None):
                         f"loader must return a tuple or list (because image_only=False was used), got {type(data)}."
                     )
                 d[key] = data[0]
+                data[1]['tensor_name'] = key
```
einsyang723, line 172: @atbenmurray , 
Hi, since the method you proposed involves looping through the data, resulting in O(n) time complexity. We would like to ask if it’s possible to add a dictionary to make the lookup more intuitive and faster, achieving O(1) time complexity.

```
key_to_meta_key_map[key] = meta_key
...
d["key_meta_dict_map"] = key_to_meta_key_map
```

Then our `d` would like:
```
--- image{}
|
--- image_meta_dict{}
|
--- lebel{}
|
--- lebel_meta_dict{}
|
--- key_meta_dict_map{"image": "image_meta_dict", "lebel": "lebel_meta_dict"}
```

monai/transforms/intensity/dictionary.py
```diff
@@ -358,6 +358,8 @@ def __call__(self, data) -> dict[Hashable, NdarrayOrTensor]:
             d, self.factor_key, self.meta_keys, self.meta_key_postfix
         ):
             meta_key = meta_key or f"{key}_{meta_key_postfix}"
+            if meta_key not in d:
+                meta_key = get_meta_dict_name(key, d)
```
IamTingTing, line 362: Is this part of the code to prevent meta_keys from being modified or deleted?
However, if a meta_key were deleted, the get_meta_dict_name() function would not be able to find the corresponding meta_key.
As for the possibility of meta_keys being modified, we have not found any function that could alter it.
Therefore, the condition here might be overlapping with the original code.
Could you kindly share the reason for adding it?

PR #8460, opened 2025-05-23
Classifier free guidance

### LINKED ISSUE #8448: Classifier-free guidance added to the Inferers
Diffusion inferers currently do not support classifier-free guidance; the classifier-free guidance tutorial (https://github.com/Project-MONAI/tutorials/tree/main/generation/classifier_free_guidance) has to implement the sample method outside of the inferer.

Perhaps, there could be a classifier-free guidance attribute to the inferer that branches the sample logic so that, if not None, the sampling method samples an unconditioned / conditioned output for every t in the for loop as in the tutorial. This would not alter the behaviour if the attribute is defaulted to None.

virginiafdez:
Fixes #8448 

### Description

Classifier-free guidance scale can be used in the sampling methods of diffusion models to strengthen the conditioning. It's been used in one of the generative MONAI tutorials, but the actual Inferers (DiffusionInferer, and ControlNetDiffusionInferer) do not support this feature. This means that, whenever users want to use CFG, they have to either copy the Inferer object or write their own sampling method. 
This PR incorporates classifier-free guidance into the inferer objects by modifying their sampling method and adding an argument cfg to the sampling method to control this.
This should not change the default behaviour (cfg=None), although some rewriting has been necessary.

### Types of changes

- [x] New tests added to cover the changes (to test_controlnet_inferers.py, test_diffusion_inferer.py and test_latent_diffusion_inferer.py). 

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -105,9 +105,9 @@ MONAI tests are located under `tests/`.
 A bash script (`runtests.sh`) is provided to run all tests locally.
 Please run ``./runtests.sh -h`` to see all options.
 
-To run a particular test, for example `tests/test_dice_loss.py`:
+To run a particular test, for example `tests/losses/test_dice_loss.py`:
 ```
-python -m tests.test_dice_loss
+python -m tests.losses.test_dice_loss
 ```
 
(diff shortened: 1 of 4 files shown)


ericspod (approved):
Looks good to me!

PR #8470, opened 2025-06-03
fix: support decollate for numpy scalars

arthurdjn:
Fixes https://github.com/Project-MONAI/MONAI/issues/8471

### Description

This PR supports numpy scalars (e.g. in the form of `np.array(1)` ) in the `decollate_batch` function (fix issue https://github.com/Project-MONAI/MONAI/issues/8471).

### Types of changes

### DIFF
--- a/.github/workflows/blossom-ci.yml
+++ b/.github/workflows/blossom-ci.yml
@@ -51,7 +51,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v5
+        uses: actions/checkout@v4
         with:
           repository: ${{ fromJson(needs.Authorization.outputs.args).repo }}
           ref: ${{ fromJson(needs.Authorization.outputs.args).ref }}
--- a/.github/workflows/chatops.yml
+++ b/.github/workflows/chatops.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: dispatch
-        uses: peter-evans/slash-command-dispatch@v4.0.0
+        uses: peter-evans/slash-command-dispatch@v5.0.2
         with:
           token: ${{ secrets.PR_MAINTAIN }}
           reaction-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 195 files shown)


monai/data/utils.py
```diff
@@ -625,6 +625,8 @@ def decollate_batch(batch, detach: bool = True, pad=True, fill_value=None):
         type(batch).__module__ == "numpy" and not isinstance(batch, Iterable)
     ):
         return batch
+    if isinstance(batch, np.ndarray) and batch.ndim == 0:
```
KumoLiu, line 628: Thanks for the pr! Do you think it might be beneficial to convert the array into a tensor? This way, the data could be handled more consistently.

arthurdjn: We could, I think it does not matter for my use cases. As long as the function handles numpy scalars in the form of an array it is good for me!

I will add this change and convert it as a tensor there (L629) if you prefer :)

ericspod:
Could we consider a more complete solution? The issue it seems is that 0-d arrays are iterable but can't be iterated over. We already check for non-iterable things in `decollate_batch` [here](https://github.com/Project-MONAI/MONAI/blob/c3a317d2bcb486199f40bda0d722a41e3869712a/monai/data/utils.py#L624). Can we modeify this to correctly pick up when the `batch` is a 0-d array and just return it in that case? Or return its contents?

ericspod:
> Is this something that you find more complete?

What I had in mind was more of the following change:

```python
...
    if batch is None or isinstance(batch, (float, int, str, bytes)):
        return batch
    if getattr(batch, "ndim", -1) == 0:  # assumes only Numpy objects and Pytorch tensors have ndim
        return batch.item() if detach else batch
    if isinstance(batch, torch.Tensor):
        if detach:
            batch = batch.detach()
# REMOVE
#      if batch.ndim == 0:
#          return batch.item() if detach else batch
...
```

arthurdjn:
Thanks! I will update the PR to include these changes.

ericspod:
Sorry for the long delay again. If this is all that's needed in the end we can merge this shortly. I think lines [605-6](https://github.com/arthurdjn/MONAI/blob/0e43b28422e710134d232b808b08556b1a38290a/monai/data/utils.py#L605) are not redundant since the added if condition will catch this case?

ericspod:
Also please have a look at the DCO message for doing a remedial commit.

PR #8475, opened 2025-06-09
7951 anchors are not centered on grid cells

### LINKED ISSUE #7951: Anchors are not centered on grid cells
**Describe the bug**
Anchors should be placed ideally in the middle of a grid cell. 
However, this is not the case here:
https://github.com/Project-MONAI/MONAI/blob/12d00ce1369e37cb06f483735ef83674a208b031/monai/apps/detection/utils/anchor_utils.py#L253-L258

**Expected behavior**
The anchors should be placed ideally in the middle of a grid cell.

**Possible fix**
This could probably be fixed by simply changing these lines to:
```python
shifts_centers = [
    torch.arange(0, size[axis], dtype=torch.int32, device=device) * stride[axis] + int(stride[axis] / 2)
    for axis in range(self.spatial_dims)
]
```

hachoj:
This pull request resolves issue #7951 by correctly centering anchor boxes within their grid cells.

### Description

**Changes Made:**
- Modified the anchor generation logic in `monai/apps/detection/utils/anchor_utils.py` to add a `stride // 2` offset.
- Refactored the corresponding unit test in `tests/apps/detection/utils/test_anchor_box.py` to validate this new, correct behavior against the torchvision baseline by accounting for the offset.

Fixes #7951

### Types of changes

- [ ]


ericspod:
Hi @hachoj thanks for the contribution, it looks fine to me as far as I can tell. Could you please check that this works with the [luna16 tutorial example](https://github.com/Project-MONAI/tutorials/tree/main/detection)? If so I think we're good. @Can-Zhao I think you were the original author of this file, any thoughts?

Cado87:
Hi @ericspod 

I ran a smoke test of the LUNA16 detection tutorial pipeline with this patch applied (MONAI commit [8f6e2c7](https://github.com/Project-MONAI/MONAI/pull/8475/commits/8f6e2c71245d626cffc5cf69c4997feefc998be7), PyTorch
2.10.0, Tesla T4 GPU).

What was tested:
- Verified the anchor fix is active: with a 2×2×2 grid and stride=4, the first anchor center is now at (2.0, 2.0,
2.0) instead of (0.0, 0.0, 0.0)
- Ran [luna16_training.py](https://github.com/Project-MONAI/tutorials/blob/main/detection/luna16_training.py) (fold 0, 3 epochs, synthetic volumes matching the fold-0 datasplit paths) — training
completed without errors, loss was finite and decreasing
- Ran [luna16_testing.py](https://github.com/Project-MONAI/tutorials/blob/main/detection/luna16_testing.py) on the saved checkpoint — inference processed 67 images, produced 94 predicted boxes with
scores in a valid range

The full pipeline (data loading → ATSS matching → training → sliding window inference → box post-processing) runs
without issues with the centered anchors.

Note: since real LUNA16 data was not used, detection quality metrics are not meaningful here. The test only
validates that the fix does not break the tutorial pipeline. With real data, a full training run would be needed to
confirm the expected improvement in detection accuracy from properly centered anchors.

ericspod (approved):

ericspod:
Hi @Cado87 thanks for the help here. I think the change is fine with a sanity check like what you've done so we're good to merge. We can return to this later of course, the Coderabbit comment didn't need necessary to me but it might be worth looking at later if anyone has issues with this change.

PR #8493, opened 2025-06-20
Added focal_loss_with_probs and full activation selection in Dice+FL.

### LINKED ISSUE #8242: DiceFocalLoss does not have the option to run softmax focal loss because softmax is not being passed to FocalLoss class
The 'use_softmax' arg is missing in FocalLoss in the DiceFocalLoss class. See the screenshot.

Screenshot 2024-11-28 at 10 42 24

j-morano:
Fixes #8242 

### Description

Added option to compute Focal Loss (FL) directly from class probabilities instead of logits.
Added parameters to the different loss classes combining Dice and FL to allow choosing the activation function (if any) for FL.
Added corresponding tests passing with `pytest tests/losses/ -v`.

### Types of changes

### DIFF
--- a/monai/losses/dice.py
+++ b/monai/losses/dice.py
@@ -823,8 +823,10 @@ class DiceFocalLoss(_Loss):
         self,
         include_background: bool = True,
         to_onehot_y: bool = False,
-        sigmoid: bool = False,
-        softmax: bool = False,
+        sigmoid_dice: bool = False,
+        softmax_dice: bool = False,
+        sigmoid_focal: bool = True,
+        softmax_focal: bool = False,
         other_act: Callable | None = None,
         squared_pred: bool = False,
         jaccard: bool = False,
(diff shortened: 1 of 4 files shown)


monai/losses/dice.py
```diff
@@ -823,8 +823,10 @@ def __init__(
         self,
         include_background: bool = True,
         to_onehot_y: bool = False,
-        sigmoid: bool = False,
-        softmax: bool = False,
+        sigmoid_dice: bool = False,
```
KumoLiu, line 826: Hi @j-morano, thanks for the contributions!
We would suggest using `deprecated_arg` to add such changes.
https://github.com/Project-MONAI/MONAI/blob/b92b2ce8aa25e53c8e1824da0a063b9e5239f1e4/monai/utils/deprecate_utils.py#L123

You can find an example here:
https://github.com/Project-MONAI/MONAI/blob/b92b2ce8aa25e53c8e1824da0a063b9e5239f1e4/monai/metrics/meandice.py#L251

PR #8517, opened 2025-07-22
fix bug in maisi vae

Can-Zhao:
Fixes # .

### Description

### Types of changes

### DIFF
--- a/monai/apps/generation/maisi/networks/autoencoderkl_maisi.py
+++ b/monai/apps/generation/maisi/networks/autoencoderkl_maisi.py
@@ -232,7 +232,7 @@ class MaisiConvolution(nn.Module):
         if self.print_info:
             logger.info(f"Number of splits: {self.num_splits}")
 
-        if self.dim_split <= 1 and self.num_splits <= 1:
+        if self.num_splits <= 1:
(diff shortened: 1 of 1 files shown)


monai/apps/generation/maisi/networks/autoencoderkl_maisi.py
```diff
@@ -232,7 +232,7 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         if self.print_info:
             logger.info(f"Number of splits: {self.num_splits}")
 
-        if self.dim_split <= 1 and self.num_splits <= 1:
+        if self.num_splits <= 1:
```
KumoLiu, line 235: Hi @Can-Zhao, could you please explain the reason for this change? It would be helpful to mention it here, so when we review the commit in the future, it won't confuse others. Thank you!

Can-Zhao: Because if self.num_splits <= 1, it means there is no tensor split, so we directly call convolution. There is no need to add condition on self.dim_split. Thank you!

KumoLiu (approved):

PR #8522, opened 2025-07-29
Fix LoadImage to raise OptionalImportError when specified reader is not available

### LINKED ISSUE #7437: Raise the exception when LoadImage has a reader specified but it is not installed

its-serah:
## Description
This PR fixes issue #7437 by making LoadImage raise an OptionalImportError when a specified reader is not installed, instead of silently falling back to another reader.

## Changes
- Modified `LoadImage.__init__` to catch ValueError from `look_up_option` when reader name is not recognized
- Raise `OptionalImportError` instead of just warning when specified reader is not installed
- Added test case to verify the new behavior

## Why this is needed
Previously, when a user specified `LoadImage(reader='ITKReader')` without ITK installed, it would just warn and use PILReader instead. This could lead to confusion and unexpected behavior. Now it properly raises an OptionalImportError to make it clear that the requested reader is not available.

Fixes #7437


ericspod:
Hi @its-serah thanks for the contribution. As you see from the failed tests the expectation of the reader is to not raise an exception if an optional package isn't found, and this is correct in the cases when a fallback reader does exist for some formats. If we raise an exception whenever a reader can't be loaded then this fallback behaviour can't happen. I would suggest that we add a flag as a member to the class to enable the exception behaviour, but whose default state retains the existing behaviour. We would also need tests to check that turning this on correctly raises exceptions. What do you think? There's interest in the associated issue so I'm keen to find a solution that works for everyone. Thanks!

its-serah:
Hi @ericspod  thanks for the excellent feedback! I've implemented exactly what you suggested. I added a raise_on_missing_reader flag to both LoadImage and LoadImaged classes that defaults to False to maintain existing behavior and backward compatibility. When set to True, it raises OptionalImportError for missing readers. When False (default), it preserves the current fallback behavior with warnings. I also added comprehensive tests to verify the flag correctly raises exceptions when enabled. This gives users control while maintaining the important fallback functionality for existing codebases. The failed tests should now pass since the default behavior is unchanged. What do you think of this approach?

monai/transforms/io/array.py
```diff
@@ -206,13 +210,25 @@ def __init__(
                 if not has_built_in:
                     the_reader = locate(f"{_r}")  # search dotted path
                 if the_reader is None:
-                    the_reader = look_up_option(_r.lower(), SUPPORTED_READERS)
+                    try:
+                        the_reader = look_up_option(_r.lower(), SUPPORTED_READERS)
+                    except ValueError:
+                        # If the reader name is not recognized at all, raise OptionalImportError
+                        raise OptionalImportError(
```
ericspod: I think we need `if self.raise_on_missing_reader` here as well.

ericspod:
Hi @its-serah it looks better now, but I think the two raises need to be guarded by the same mechanism. Other than that it looks good though you'll have to fix your DCO issue and the formatting issue (`./runtests.sh --autofix` will do it). Thanks!

its-serah:
Hi @ericspod  thanks for the feedback! I've addressed both issues you mentioned:

Fixed the guarding mechanism: Both `raise OptionalImportError` statements now use the same `raise_on_missing_reader` flag mechanism for consistent behavior.

Fixed DCO and formatting: Added proper DCO sign-off and applied code formatting fixes using isort, black, and ruff.

The changes ensure that both exception cases (unrecognized reader name and missing dependencies) are handled consistently through the same flag. When `raise_on_missing_reader=False`, both cases will show warnings and allow fallback behavior. When `True`, both will raise `OptionalImportError`.

monai/transforms/io/array.py
```diff
@@ -206,13 +210,32 @@ def __init__(
...
+                        # If the reader name is not recognized at all, raise OptionalImportError
+                        if self.raise_on_missing_reader:
+                            raise OptionalImportError(
+                                f"Cannot find reader '{_r}'. It may not be installed or recognized."
+                            )
+                        else:
+                            warnings.warn(
+                                f"Cannot find reader '{_r}'. It may not be installed or recognized. "
+                                f"Will use fallback readers if available."
+                            )
```
ericspod, line 238:
```suggestion
                        msg = f"Cannot find reader '{_r}'. It may not be installed or recognized."
                        if self.raise_on_missing_reader:
                            raise OptionalImportError(msg)
                        else:
                            warnings.warn(f"{msg} Will use fallback readers if available.")
```

monai/transforms/io/array.py
```diff
@@ -206,13 +210,32 @@ def __init__(
...
+                except OptionalImportError as e:
+                    if self.raise_on_missing_reader:
+                        raise OptionalImportError(
+                            f"required package for reader {_r} is not installed, or the version doesn't match requirement."
+                        ) from e
+                    else:
+                        warnings.warn(
+                            f"required package for reader {_r} is not installed, or the version doesn't match requirement. "
+                            f"Will use fallback readers if available."
+                        )
```
ericspod, line 253:
```suggestion
                    msg = f"Required package for reader {_r} is not installed, or the version doesn't match requirement."
                    if self.raise_on_missing_reader:
                        raise OptionalImportError(msg) from e
                    else:
                        warnings.warn(f"{msg} Will use fallback readers if available.")
```

ericspod:
Hi @its-serah thanks for continuing to work on this, I think it's much better with some minor suggestions. The tests are failing however so please look again and what your tests are testing to see what the issue is. The DCO still doesn't seem happy but we can sort that last and just force it to pass if needed.

its-serah:
Hey @ericspod 

Thank you for the code review feedback! I've addressed both of your suggestions in the latest commits:

Refactored duplicate error messaging:
•  Lines 217-225: Extracted the common message f"Cannot find reader '{_r}'. It may not be installed or recognized." into a variable and reused it for both the exception and warning paths
•  Lines 230-238: Similarly extracted f"Required package for reader {_r} is not installed, or the version doesn't match requirement." to eliminate duplication

Fixed the failing test:
•  Updated test_raise_on_missing_reader_flag to correctly verify the intended behavior
•  The test now properly expects warnings (not exceptions) when raise_on_missing_reader=False
•  Added proper warning verification using warnings.catch_warnings()

The changes maintain backward compatibility while implementing the cleaner code structure you suggested. I've also verified that the functionality works as expected with custom tests.

The CI failures should now be resolved since the test logic correctly matches the implementation behavior. Please let me know if you need any clarifications or have additional feedback!

ericspod:
Hi @its-serah thanks for the update, I think this addresses the suggested code I had. The tests are failing however since your tests are attempting to load a non-existent file. There's also a few things from Coderabbit but we're just about there.

dzenanz:
@its-serah are you fine with me finishing this PR?

its-serah:
> @dzenanz No that's fine I'll handle it, thanks for the reminder!

its-serah:
pre-commit.ci run

its-serah:
PR Updated - Ready for Review! Added raise_on_missing_reader flag to LoadImage/LoadImaged. All CodeRabbit feedback addressed, tests passing, backward compatible. Defaults to False for graceful fallback behavior.

PR #8531, opened 2025-08-08
added ReduceTrait and FlattenSequence 

### LINKED ISSUE #8528: Stacking two or more MultiSampleTrait-like transforms breaks execute_compose()
**Describe the bug**
If a list of transforms includes more than one `MultiSampleTrait`-like operations (e.g. each returning a list of generated items), putting this as an argument to `execute_compose()` will lead to a list of lists being returned by [this line of apply_transform](https://github.com/Project-MONAI/MONAI/blob/b92b2ce8aa25e53c8e1824da0a063b9e5239f1e4/monai/transforms/transform.py#L146). The next transform in the list would then break, because it would try to interpret one entry of the outer list as a single item, e.g. a dict.

**To Reproduce**
Steps to reproduce the behavior (a sketch):
```python
import torch
from monai.transforms.traits import MultiSampleTrait
from monai.transforms.compose import execute_compose
from torchvision.transforms import CenterCrop

class Foo(MultiSampleTrait):
    def __init__(self):
        super().__init__()
    def __call__(self, data):
        return [data]

center_crop = CenterCrop([128,128])  # just some "normal" transform

img = torch.zeros([1, 512, 512])

res = execute_compose(img, [center_crop])  # -> tensor 1x128x128 (as expected)
res = execute_compose(img, [Foo(), center_crop])  # -> list, containing one element, tensor 1x128x128 (as expected)
res = execute_compose(img, [Foo(), Foo(), center_crop])  # -> TypeError: Unexpected type <class 'list'>
```

**Expected behavior**
Since input list or tuple unwrapping is [implemented in `apply_transform`](https://github.com/Project-MONAI/MONAI/blob/b92b2ce8aa25e53c8e1824da0a063b9e5239f1e4/monai/transforms/transform.py#L146), and `MultiSampleTrait` indicates the possibility of such transforms (I know it does not do anything by itself, nor is it used to inform any actions inside MONAI as of now), I expect it to handle return values that are lists or tuples correctly.

**Environment**
(providing a partial output since the rest seems irrelevant) 

```
Printing MONAI config...
MONAI version: 1.4.0
Numpy version: 1.26.4
Pytorch version: 2.7.1+cu126
MONAI rev id: 46a5272196a6c2590ca2589029eed8e4d56ff008
MONAI __file__: /opt/venv/lib/python3.12/site-packages/monai/__init__.py
```
(I checked that the main-branch MONAI has the same issue in its code).

**Suggested fix**

Replace [these lines](https://github.com/Project-MONAI/MONAI/blob/b92b2ce8aa25e53c8e1824da0a063b9e5239f1e4/monai/transforms/transform.py#L146-L149) with

```python
            res = []
            for item in data:
                res_item = _apply_transform(transform, item, unpack_items, lazy, overrides, log_stats)
                if isinstance(res_item, list | tuple):
                    res.extend(res_item)
                else:
                    res.append(res_item)
            return res
```

lukas-folle-snkeos:
Fixes #8528.

### Description

This PR adds the `FlatttenSequence` transform (a flavor of the also newly added `ReduceTrait`) which can flatten a nested data structure by one level. This way, #8528 can be tackled without the need to change the `apply_transform` of `Compose` significantly.

### Types of changes


lukas-folle-snkeos:
All remaining failing CI runs are not coming from this change.

monai/transforms/transform.py
```diff
@@ -143,10 +143,19 @@ def apply_transform(
...
-                apply_transform(transform, item, map_items_ - 1, unpack_items, log_stats, lazy, overrides)
-                for item in data
-            ]
+            res: list[Any] = []
+            for item in data:
+                res_item = apply_transform(transform, item, map_items_ - 1, unpack_items, log_stats, lazy, overrides)
+                # Only extend if we're at the leaf level (map_items_ == 1) and the transform
+                # actually returned a list (not preserving nested structure)
+                if isinstance(res_item, list) and map_items_ == 1:
+                    if not isinstance(item, (list, tuple)):
```
ericspod: Within the containing block `res_item` will always be a list (line 151) and so this if statement's condition is always false. I'm not sure the logic is what you're expecting it to be here?

ericspod:
Hi @lukas-folle-snkeos thanks for the contribution! Please check the logic I flagged and we do need tests to check that this behaviour is correct. You can take the mentioned issue as an example of what to check for but I think we need to test this logic thoroughly and more clearly document what's going on in comments.

monai/transforms/transform.py
lukas-folle-snkeos: @ericspod I am not sure I understand what you mean. In case `res_item` is `list`, `item` can still be something else.

ericspod: Sorry for the delay, I might have not read correctly what was going on here. I think this block 151-157 can be simplified to:

```python
if isinstance(res_item, list) and map_items_ == 1 and not isinstance(item, (list, tuple)):
    res.extend(res_item)
else:
    res.append(res_item)
```

The result list is extended when `transform` turns a non-list/tuple `item` into a list specifically, and appended otherwise. I think what I'm not sure about is under what situation `item` would be a list/tuple, and even if it was why we'd append in that case. Am I missing something with that?

We should have an added test as well that would catch the case described in the original issue, ideally using existing transforms rather than dummy ones for testing. Thanks!

atbenmurray:
Taking a look.

lukas-folle-snkeos:
Thanks for the review @ericspod! I added a test with a real transform and refactored the code as suggested.

atbenmurray:
Hi folks. I've taken an initial look at this. I'm always a little afraid of these kind of changes that mandate a particular behaviour. Are there no situations where someone might multi-sample something and then multi-sample the multi-sampled thing and expect the lists to be nested? Should this be fixed in execute_compose instead? Please give me a couple of hours today to take a look at that.

lukas-folle-snkeos:
@atbenmurray @ericspod could you please check the status of this PR again?

atbenmurray:
@lukas-folle-snkeos My apologies; taking a look now

atbenmurray:
Okay, so apologies for the long wait. I finally managed to get my head into it properly:

I'd like to suggest a slight change. I think that, if a user wants to run multisamples on top of multisamples there will be scenarios in which they don't want to next the output and scenarios in which they do. I'd rather we didn't remove the ability to do so:

I propose that we do the following instead:
1. Add ReduceTrait to the list of traits
2. Tweak the code in apply_transform, but slightly differently
3. Add a FlattenSequence transform that can be inserted when the user wants multiply nested transforms to not generate nested lists

```
        # modified apply_transform snippet that is ReduceTrait aware
        map_items_ = int(map_items) if isinstance(map_items, bool) else map_items
        if isinstance(data, (list, tuple)) and map_items_ > 0 and not isinstance(transform, ReduceTrait):
            return [
                apply_transform(transform, item, map_items_ - 1, unpack_items, log_stats, lazy, overrides)
                for item in data
            ]
        return _apply_transform(transform, data, unpack_items, lazy, overrides, log_stats)
```

```
class ReduceTrait:
    """
    An interface to indicate that the transform has the capability to reduce multiple samples
    into a single sample.
    This interface can be extended from by people adapting transforms to the MONAI framework as well
    as by implementors of MONAI transforms.
    """

    pass
```

```
class FlattenSequence(mt.Transform, mtt.ReduceTrait):
    def __init__(self):
        super().__init__()

    def __call__(self, data):
        if isinstance(data, (list, tuple)):
            if len(data) == 0:
                return data
            if isinstance(data[0], (list, tuple)):
                return [item for sublist in data for item in sublist]
        return data
```

The footprint has a slightly larger change but then we don't lose the flexibility

It would be used as follows:

```    transforms_2 = [
        mt.CenterSpatialCrop((128, 128)),
        mt.RandSpatialCropSamples((96, 96), num_samples=3),
        mt.RandSpatialCropSamples((64, 64), num_samples=2),
        FlattenSequence(),
 ```
 
 We can make interesting uses of ReduceTrait, allowing it to be a trait that can do things like aggregations, catting, etc, as well as flattening lists. WDYT?

lukas-folle-snkeos:
@atbenmurray I like that idea; looks a whole lot cleaner to me compared to the initially proposed change! I will bring the changes into this PR.

lukas-folle-snkeos:
@atbenmurray do you want to have a look at the proposed changes?

atbenmurray (approved):
Looks good to me

lukas-folle-snkeos:
@atbenmurray not sure how to make the blossom-ci run.. Seems to be required to merge.

atbenmurray:
> @atbenmurray not sure how to make the blossom-ci run.. Seems to be required to merge.

I don't have the powers, sadly. @ericspod @KumoLiu @Nic-Ma?

lukas-folle-snkeos:
bump @ericspod @KumoLiu @Nic-Ma @atbenmurray

ericspod:
@KumoLiu I think we're good to go here, please trigger blossom when you can, thanks.

Issue #8366, opened 2025-02-25
Channel last format


pure-rgb:
Does MONAI support the following format 3D data, (batch-size, depth, h, w, channel)? For utility like sliding_window_inference, and various built-in metrics, i.e. DICE?

mattlin1124:
Hi @pure-rgb , thanks for opening this issue! Quick question: in your (batch, depth, h, w, channel) tensor, what cues or metadata do you use to identify that the last axis is indeed the channel dimension? For example, do you rely on a fixed small size (1/3/4), loader metadata (e.g. original_channel_dim), or something else? Clarifying this would help me reproduce and debug your setup. Thanks!

PR #8532, opened 2025-08-09
Fix: add ensure_channel_first in utils and integrate into DiceHelper (refs #8366)

mattlin1124:
What does this PR do?
Adds ensure_channel_first utility in monai/metrics/utils.py and integrates it into DiceHelper.__call__ in monai/metrics/meandice.py to normalize input tensors to channel-first layout before Dice computation. This prevents metric calculation errors when the input is provided in channel-last format.

Issue reference
Refs #8366

Motivation and context
Currently, DiceHelper assumes the channel dimension is at index 1. If a user passes channel-last tensors (e.g., [N, H, W, C]), it can lead to incorrect indexing and metric values. This PR ensures that y_pred and (when applicable) y are converted to channel-first format before further processing, preserving original metric behavior while adding channel-last support.

How I tested

Verified that channel-last inputs now produce identical results to channel-first inputs.

Ran pytest -q tests/metrics -k dice locally without failures.

Checked that no changes are introduced to the outputs when inputs are already channel-first.


ericspod:
Hi @mattlin1124 thanks for the contribution but please do look at the Coderabbit complaints to see if there's things to address as well as the CICD fails. The .gitignore file also shouldn't be changed, and we'd need tests for this added behaviour.

I do feel that being a PyTorch library means the running assumption with all data is that it is formatted with channel first. If you have data that's channel last it should be on your to manually account for this, so forcing you to be explicitly aware of what your data is and how you're handling it. Using heuristics as you do here for checking if the channels are first or last is going to lead to false positive cases or hide what is actually being done. 

This problem is also present everywhere that we ingest data, that's why we have the [EnsureChannelFirst/EnsureChannelFirstd](https://github.com/Project-MONAI/MONAI/blob/b5bc69d0306c16cba61de58a403190689bb192b3/monai/transforms/utility/array.py#L174) transforms which don't attempt to guess at the input format. If we do want to integrate this heuristic it should be optionally used in places like this as well. 

We're planning a release soon which won't include major semantic changes but we will discuss this change for the following release.

ericspod:
I see your other PR @mattlin1124 [here related to sliding window and Dice](https://github.com/Project-MONAI/MONAI/pull/8550). This raises the same concerns of course about false positive which I think what Coderabbit is on about.

mattlin1124:
Thanks @ericspod  for the clear feedback.  
I’ll try to add strict shape validation without any implicit guessing.  
If the input isn’t channel-first, the function will raise a clear error telling users to apply `EnsureChannelFirst/EnsureChannelFirstd` upstream.  
I’ll also revert the unintended `.gitignore` change as suggested.  

Does this approach sound reasonable?

ericspod:
> Thanks @ericspod for the clear feedback. I’ll try to add strict shape validation without any implicit guessing. If the input isn’t channel-first, the function will raise a clear error telling users to apply `EnsureChannelFirst/EnsureChannelFirstd` upstream. I’ll also revert the unintended `.gitignore` change as suggested.
> 
> Does this approach sound reasonable?

Hi @mattlin1124 this all sounds good. We can review this shortly but we're working on a minor release to go out soon so this will be delayed a bit. Thanks!

PR #8535, opened 2025-08-10
Tests Cleanup 

### LINKED ISSUE #8534: Further Test Cleanup
A few areas of the tests can be easily cleaned up currently to reduce complexity and about of code:

* Remove obsolete checks for enabling/disabling tests for certain versions of PyTorch, eg. a test being disabled for PyTorch versions less than 2.2 can have that check removed.
* Some legacy bug fixes can be removed since these have been fixed or aren't relevant now.
* Some checks disable tests if `torch.fft` isn't present, this always is now for currently supported versions.
* Moving some files into more appropriate subdirectories.

ericspod:
Fixes #8534.

### Description

This cleans up the tests further as discussed in the issue. This also additionally updates the zarr tests to save the .zarr directories to a temp directory rather than the current one, this is just cleaner when running local tests.

### Types of changes


tests/inferers/test_zarr_avg_merger.py
```diff
@@ -11,7 +11,9 @@
 
 from __future__ import annotations
 
+import os
 import unittest
+from tempfile import TemporaryDirectory
```
garciadias, line 16: I like the use of the `tempfile` module here. There is at least one test file that relies on creating temp files at `tests/testing_data/temp`, the tests at `tests/apps/pathology/test_lesion_froc.py`.
I understand many of the files in `tests/testing_data` are testing assets that can be reused, but it would be nice to implement this `tempfile` module behaviour to all temporary files.

garciadias (approved):

PR #8542, opened 2025-08-19
Change SupervisedTrainer decollation default.

### LINKED ISSUE #8541: Default for SupervisedTrainer slows down training almost 2x
https://github.com/Project-MONAI/MONAI/blob/d388d1c6fec8cb3a0eebee5b5a0b9776ca59ca83/monai/engines/trainer.py#L121-L123

The default behaviour for the SupervisedTrainer slows down each training step quite a bit. The cause of this slow-down is not apparent until profiling each step. With larger batch sizes, the delay from the default decollation step increases as more tensors have to move from GPU/CPU and back.

Image
In this flame-chart, there are two steps. A forward, backward and the decollate.

It may be better to set the default for decollation in the supervised trainer to `False` or to add a warning that the default behaviour can significantly impact training time (in our case, 2x speed up from 4 days to 2 days by disabling decollation).

id-b3:
Fixes #8541 .

### Description

By default, the SupervisedTrainer has `decollation` enabled by default. The decollation step moves the tensors to CPU for processing which significantly impacts the speed of each training step as described in issue #8541 

This pull request proposes to change the default of decollation to `False` and provide more information in the documentation as to the impact of the decollation option with regard to the training speed.

### Types of changes

### DIFF
--- a/monai/engines/trainer.py
+++ b/monai/engines/trainer.py
@@ -120,7 +120,7 @@ class SupervisedTrainer(Trainer):
             #ignite.engine.engine.Engine.register_events.
         decollate: whether to decollate the batch-first data to a list of data after model computation,
             recommend `decollate=True` when `postprocessing` uses components from `monai.transforms`.
-            default to `True`.
+            default to `False` as training slows due to tensor movement to CPU for decollation when enabled.
         optim_set_to_none: when calling `optimizer.zero_grad()`, instead of setting to zero, set the grads to None.
             more details: https://pytorch.org/docs/stable/generated/torch.optim.Optimizer.zero_grad.html.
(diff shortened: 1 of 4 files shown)


ericspod:
Hi @id-b3 thanks for the contribution. This can be a significant change for users who have written their own handlers or metrics so we need to do some evaluation beyond just the unit tests. Can you run the tests in the Tutorials repo with your change? 

Also please use the [deprecated_arg_default](https://github.com/Project-MONAI/MONAI/blob/b5bc69d0306c16cba61de58a403190689bb192b3/monai/utils/deprecate_utils.py#L230) decorator to mark the changing default vlaue.

PR #8543, opened 2025-08-20
fix: HistogramNormalized doc

gauthamk02:
### Description

The docstring for `min` parameter for `HistogramNormalized` wrongly shows the default value to be 255 when the correct default value is 0. 

### Types of changes


ericspod (approved):
Thanks for spotting this! Looks good to me.

PR #8545, opened 2025-08-22
fix(transformerblock): conditionally initialize cross attention compo…

yunhaoli24:
Transformerblock

This PR improves the initialization logic of TransformerBlock by adding conditional layer creation for cross-attention.
Previously, the cross-attention components (norm_cross_attn and cross_attn) were always initialized, even when with_cross_attention=False. This introduced unnecessary memory overhead and additional parameters that were never used during forward passes.

With this change, the cross-attention layers are initialized only when with_cross_attention=True.
This ensures cleaner module definitions, reduced memory usage, and avoids confusion about unused layers in models that do not require cross-attention.

Fixes # .

### Description

Before
```python
...
class TransformerBlock(nn.Module):


    def __init__(
        self,
...
    ) -> None:
       ...
        self.norm2 = nn.LayerNorm(hidden_size)

            hidden_size=hidden_size,
            num_heads=num_heads,
            dropout_rate=dropout_rate,
            qkv_bias=qkv_bias,
            causal=False,
        )
```

After
```python
...
class TransformerBlock(nn.Module):


    def __init__(
        self,
...
    ) -> None:
       ...
        self.norm2 = nn.LayerNorm(hidden_size)

        if with_cross_attention:
                hidden_size=hidden_size,
                num_heads=num_heads,
                dropout_rate=dropout_rate,
                qkv_bias=qkv_bias,
                causal=False,
            )
```
### Types of changes

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,11 +3,15 @@
 __pycache__/
 docs/
 
+.vscode
+.git
+.mypy_cache
+.ruff_cache
+.pytype
 .coverage
 .coverage.*
 .coverage/
 coverage.xml
 .readthedocs.yml
-*.toml
 
 !README.md
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -380,7 +380,8 @@ All code review comments should be specific, constructive, and actionable.
 
 ### Release a new version
 
-The `dev` branch's `HEAD` always corresponds to MONAI docker image's latest tag: `projectmonai/monai:latest`.
+The `dev` branch's `HEAD` always corresponds to MONAI Docker image's latest tag: `projectmonai/monai:latest`. (No
+release is currently done for the slim MONAI image, this is built locally by users.)
 The `main` branch's `HEAD` always corresponds to the latest MONAI milestone release.
 
 When major features are ready for a milestone, to prepare for a new release:
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -1,0 +1,93 @@
+# Copyright (c) MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This is a slimmed down version of the MONAI Docker image using a smaller base image and multi-stage building. Not all
+# NVIDIA tools will be present but all libraries and compiled code are included. This image isn't provided through
+# Dockerhub so users must build locally: `docker build -t monai_slim -f Dockerfile.slim .`
+# Containers may require more shared memory, eg.: `docker run -ti --rm --gpus all --shm-size=10gb monai_slim /bin/bash`
+
+ARG IMAGE=debian:12-slim
+
+FROM ${IMAGE} AS build
+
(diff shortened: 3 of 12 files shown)


ericspod:
Hi @yunhaoli24 thanks for the contribution, but as the Coderabbit comment notes this will break existing stored weight files since the members will not always exist. Your change is more efficient but I think requires more consideration to avoid this issue, and also should be documented in the docstring that this is happening. We are working on a quick release to be done shortly but then we can return to this PR.

yunhaoli24:
Thanks for the feedback! I agree with your point about checkpoint compatibility; that’s indeed an important concern.
At the same time, I’d like to note that in a DDP multi-GPU setting, initializing but not using the cross_attn layers can lead to torch errors about unused parameters during training. So this change could also help avoid those issues in addition to saving memory.
I think a good compromise might be to add backward-compatibility handling when loading checkpoints, or at least documenting this behavior clearly in the docstring.

ericspod:
> Thanks for the feedback! I agree with your point about checkpoint compatibility; that’s indeed an important concern. At the same time, I’d like to note that in a DDP multi-GPU setting, initializing but not using the cross_attn layers can lead to torch errors about unused parameters during training. So this change could also help avoid those issues in addition to saving memory. I think a good compromise might be to add backward-compatibility handling when loading checkpoints, or at least documenting this behavior clearly in the docstring.

Yes I have DDP errors related to this so you're right, this is one advantage to consider. Please do update the docstring to record what's going on here. One other thing to consider is Torchscript compatibility which I can't say is being tested thoroughly enough for this class. I know it has in the past complained about absent members even if the Python code works, you may need to always define the members but use `nn.Identity` as their values when `with_cross_attention` is False (in which case the condition check isn't needed in `forward` either).

yunhaoli24:
Thanks for the suggestion! Regarding defining cross_attn as nn.Identity, I have a few concerns:

* Interface mismatch: The current forward unconditionally calls self.cross_attn(self.norm_cross_attn(x), context=context), whereas nn.Identity.forward only accepts a single input and doesn’t support a context positional or keyword argument. This would raise an “unexpected keyword argument 'context'” error in both Eager and TorchScript.
* Semantic change: Even if we change the call to pass only x to avoid the error, nn.Identity returns the normalized x as-is. That adds a non-zero residual when with_cross_attention=False, which breaks the expectation that disabling cross-attn should not alter the output.
* TorchScript compatibility: Scripting performs static signature checks. Invoking a module with a context argument when the target doesn’t accept it will fail at script/trace time or produce an inconsistent graph.

ericspod:
> Thanks for the suggestion! Regarding defining cross_attn as nn.Identity, I have a few concerns:

Sorry for the delay in replying. I see the issue you're mentioning about `nn.Identity`, one possible solution is to define a new `nn.Module` subclass whose `forward` method accepts two arguments but returns 0 always, it's less computationally efficient however. We should double check that Torchscript works with your change as it is first and that the unit tests are thorough enough in that regard, but a second solution is to define `self.cross_attn` as `nn.Identity` in the constructor (and always define `self.norm_cross_attn`) when `with_cross_attention` is false, and just not use it in the `forward` method. 

@virginiafdez could you please assist here as well? Thanks!

ericspod:
I would like to mention that another PR https://github.com/Project-MONAI/MONAI/pull/8584 is looking at the same problem. I suggested we consolidate efforts here.

yunhaoli24:
I’d be happy to continue working on this PR. Please let me know if there’s anything that needs to be modified or improved. @ericspod

virginiafdez:
Hi @yunhaoli24, many thanks for spotting this and apologies for the delay. Since this issue is duplicated and also addressed in the PR @ericspod linked, which also addresses the checkpoint problem, we will move the fix to that PR for the time being. Further tests and fixes are required, but I will keep these two PRs open for the time being and make sure I update as soon as the fix's done.

Issue #8369, opened 2025-02-26
box_iou does not work as expected when boxes1 dtype is int


xkuubix:
**Describe the bug**
`box_iou()` does not work as expected when `boxes1` `dtype` is `int`, resulting in returned IoU=`0` (after converting back to `int` `dtype`). No warning messages are generated in this case.

**To Reproduce**
1. Install `monai`.
2. Run commands:
```
import torch
from monai.data.box_utils import box_iou
a = torch.tensor([1., 1., 1., 2., 2., 2.]).unsqueeze(0)
b = torch.tensor([1, 1, 1, 2, 2, 2]).unsqueeze(0)

print(a, a.dtype)
print(b, b.dtype)

iou_aa = box_iou(a,a)
print(f"{iou_aa=}")
iou_ab = box_iou(a,b)
print(f"{iou_ab=}")
iou_bb = box_iou(b,b)
print(f"{iou_bb=}")
iou_ba = box_iou(b,a)
print(f"{iou_ba=}")
```
OUTPUT:
```
tensor([[1., 1., 1., 2., 2., 2.]]) torch.float32
tensor([[1, 1, 1, 2, 2, 2]]) torch.int64
iou_aa=tensor([[1.0000]])
iou_ab=tensor([[1.0000]])
iou_bb=tensor([[0]])
iou_ba=tensor([[0]])
```
**Expected behavior**
All IoUs  should be (in this case) values close to one. However, if `boxes1.dtype` is of type `int`
the line in the definition of `box_iou()`:
`iou_t = iou_t.to(dtype=box_dtype)`
makes those IoUs to be 0. Maybe casting the IoU result back to integer should be avoided or at least warning message can be generated (`iou_t` is always in range `[0.; 1.]` just before type conversion).

**Environment**

================================
Printing MONAI config...
================================
MONAI version: 1.4.0
Numpy version: 1.26.4
Pytorch version: 2.5.0
MONAI flags: HAS_EXT = False, USE_COMPILED = False, USE_META_DICT = False
MONAI rev id: 46a5272196a6c2590ca2589029eed8e4d56ff008
MONAI __file__: /home/<username>/anaconda3/envs/crc/lib/python3.9/site-packages/monai/__init__.py

Optional dependencies:
Pytorch Ignite version: 0.4.11
ITK version: 5.4.2
Nibabel version: 5.3.2
scikit-image version: 0.22.0
scipy version: 1.13.1
Pillow version: 10.4.0
[...]

PR #8553, opened 2025-09-02
Fix box_iou returning 0 for floating-point results less than 1. #8369

reworld223:
Fixes # 8369

### Description
Fixes an issue where the result of `box_iou` was 0 when the first argument was an integer, as described in #8369.
IOU values range from 0 to 1. If the first argument is an integer, the function could return an integer less than 1, resulting in a return value of 0.
This pull request changes the return type of `box_iou` to `float32`, or to match the float data type of the first argument.

### Types of changes


monai/data/box_utils.py
```diff
@@ -842,16 +844,18 @@ def box_iou(boxes1: NdarrayOrTensor, boxes2: NdarrayOrTensor) -> NdarrayOrTensor
...
+        box_dtype = COMPUTE_DTYPE
     iou_t = iou_t.to(dtype=box_dtype)
 
     # check if NaN or Inf
     if torch.isnan(iou_t).any() or torch.isinf(iou_t).any():
         raise ValueError("Box IoU is NaN or Inf.")
 
     # convert tensor back to numpy if needed
-    iou, *_ = convert_to_dst_type(src=iou_t, dst=boxes1)
+    iou, *_ = convert_to_dst_type(src=iou_t, dst=boxes1, dtype=box_dtype)
```
ericspod, line 859: Will this return a Numpy array if the inputs are arrays? We should have a test for this new behaviour to verify the result is a float array/tensor even when the inputs are integer boxes.

ericspod:
Hi @reworld223 thanks for the contribution. I had a minor comment but please also look at what Coderabbit is saying if it's relevant. We should have a test though I think for this behaviour, and if @xkuubix could have a look to see if this solves the initial posted issue.

xkuubix:
> Hi @reworld223 thanks for the contribution. I had a minor comment but please also look at what Coderabbit is saying if it's relevant. We should have a test though I think for this behaviour, and if @xkuubix could have a look to see if this solves the initial posted issue.

It does solve the initial issue, @ericspod.
Regarding what @coderabbit mentioned about `giou`, I checked and it’s the same case (minimal example the same as in #8553 but with `box_giou` instead of `box_iou`). I didn’t come across it before since I was only using `box_iou`.

monai/data/box_utils.py
xkuubix: Yes, I checked, `box_iou` works as expected. It correctly handles **NumPy arrays** and **Torch tensors** with int or float dtypes, including mixed types, and always returns a float array/tensor.  

If the inputs are mixed types (one NumPy array, one Torch tensor), `box_iou` raises a **UserWarning** indicating that the result type follows the type of the first input, e.g.:
`boxes1 is <class 'numpy.ndarray'>, while boxes2 is <class 'torch.Tensor'>. The result will be <class 'numpy.ndarray'>.` or vice versa.

reworld223:
@ericspod Do I need to complete the unit test mentioned in Code Rabbit? If so, should I modify tests/data/test_box_utils.py?

ericspod:
> @ericspod Do I need to complete the unit test mentioned in Code Rabbit? If so, should I modify tests/data/test_box_utils.py?

Sorry for the delay. I think we do need at least one unit test that checks for the situation your PR is meant to fix. This should check that integer box inputs still produce floating point results. I would do these without pytest however as we don't use that library, but it's a good suggestion to ensure your fix doesn't get broken in the future. Also, there was mention of improving the docstring comments you may want to look at and accept if you think they help. Thanks!

ericspod (approved):
Thanks for the contribution, I think we're good now to merge.

PR #8561, opened 2025-09-08
Refactoring min-dep tests

ericspod:
### Description

This PR is an effort at refactoring the CICD tests to reduce code volume in the yaml files. Branch rules will have to be temporarily changed to not expect jobs by names which are no longer used, and once accepted changed again to add the jobs with the new names back in. This would be a good opportunity to move branch rules to rule sets which is the newer Github mechanism. This also adds code for removing unused tools in the Linux runners to avoid running out of disk space during testing, this has increasingly become an issue for testing as the image has gotten larger and more libraries get installed. 

This also removes testing support for Python 3.9 and updates which versions of PyTorch are tested with in line with our support policy. This also bumps versions of some actions and dependencies, specifically black for security reasons. Testing Windows with full dependencies requires a much later version of PyTorch due to a gloo bug, and macOS requires the explicit installation of pybind11 for now. A number of other formatting fixes had to be made to account for newer versions of black and other tools, and a number of imports were moved around to match Python 3.10 usage. A number of type annotations were changed to remove Union and Optional.

Windows tests with versions of PyTorch between 2.8 and 2.10 seem to suffer from the gloo bug mentioned in  https://github.com/pytorch/pytorch/issues/150381.

### Types of changes


ericspod:
@coderabbitai summarise the changes in this PR as a serices of bullet points.

monai/apps/detection/transforms/dictionary.py
```diff
@@ -814,17 +809,17 @@ def __init__(
         allow_missing_keys: bool = False,
     ) -> None:
         box_keys_tuple = ensure_tuple(box_keys)
-        if len(box_keys_tuple) != 1:
+        if len(box_keys_tuple) < 1:
```
garciadias: Constructor now accepts `len(box_keys_tuple) >= 1` and `len(box_ref_image_keys_tuple) >= 1`, but still only uses index `0` (`self.box_keys = box_keys_tuple[0]`, `self.box_ref_image_keys = box_ref_image_keys_tuple[0]`). Previously, passing multiple keys was rejected; now extra keys are just ignored, it may be ok, I don't know enough to tell, but just highlinghing in case you didn't mean to allow this.

tests/test_utils.py
```diff
@@ -562,15 +562,17 @@ def _wrapper(*args, **kwargs):
...
+                run_args = (func, proc_rank, args, kwargs, results)
+                p = tmp.Process(target=self.run_process, args=run_args, daemon=self.daemon)
                 p.start()
                 processes.append(p)
+
             for p in processes:
                 p.join()
-                assert results.get(), "Distributed call failed."
+                pr = results.get()
+                assert pr is True, f"Distributed call failed: {pr}"
```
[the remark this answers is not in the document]
garciadias: Yes, this is relevant because `multiprocessing.Queue` delivery is asynchronous; even after process join, non-blocking get can still race and raise `queue.Empty`, causing intermittent CI/test failures unrelated to functional regressions. If you use blocking `get(timeout=...)`  you could avoid hangs.

.github/workflows/build_docs.yml
```diff
@@ -0,0 +1,45 @@
+name: tests
```
garciadias: Are you sure this should be called tests? It seems we are just building docs.

garciadias (approved):
Great improvement, Eric. I am adding some minor comments, please take a look.

ericspod:
@coderabbitai Update this comment https://github.com/Project-MONAI/MONAI/pull/8561#issuecomment-4224326855 with changes made since it was written.

Nic-Ma (approved):

PR #8562, opened 2025-09-08
Classifier free guidance unconditioned value

### LINKED ISSUE #8560: Classifier-free guidance value in diffusion inferers is forced to -1
Classifier-free guidance is useful to make conditioning stronger in diffusion models. For this, a portion of the conditioning passed as input is set to a value that means "no condition". Currently, this is set to -1. However, this value might not be suitable for certain applications (e.g. text representations are usually normally-distributed, so -1 would not make sense here).

Give users the possibility of specifying the "unconditional" value.

virginiafdez:
Fixes #8560.

### Description

This PR adds an argument cfg_fill_value so that users can control the value that replaces the conditioning tensor during classifier-free guidance inference. 
Previously, this was set to -1, which might not be ideal for certain application where the conditioning could have -1 as a normal value. 

### Types of changes

- [x] New tests added to cover the changes **modified existing ones

### DIFF
--- a/monai/inferers/inferer.py
+++ b/monai/inferers/inferer.py
@@ -916,6 +916,7 @@ class DiffusionInferer(Inferer):
         verbose: bool = True,
         seg: torch.Tensor | None = None,
         cfg: float | None = None,
+        cfg_fill_value: float = -1.0,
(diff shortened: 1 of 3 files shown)


ericspod (approved):
Looks good to me!

PR #8568, opened 2025-09-12
Path traversal issue security fix

h3rrr:
The path traversal issue in the previously submitted vulnerability report has been fixed and has been verified and fixed locally.
https://github.com/Project-MONAI/MONAI/security/advisories/GHSA-x6ww-pf9m-m73m

### Description

Added a security function to fix a security issue caused by directly using extractall in zip_file.extractall(output_dir).
This has been verified locally and does not affect previous functionality.
image
image
Figure 1: Malicious zip file loaded, Figure 2: Normal zip file loaded

### Types of changes


ericspod:
This PR includes the security file as well but this should go away once we merge the other PR. This happened because you made your `h3rr-patch-1` branch from the `dev` branch you were using with the previous. No worries, this will work out eventually.

h3rrr:
Hi, @ericspod  the pull request for SECURITY.md mentioned the need for testing.
I've already tested this locally, and the image above shows the results.
You can see that when a zip file contains malicious content, an error is thrown.
Normal zip files, however, are successfully extracted, with the same results as before.

ericspod:
> Hi, @ericspod the pull request for SECURITY.md mentioned the need for testing. I've already tested this locally, and the image above shows the results. You can see that when a zip file contains malicious content, an error is thrown. Normal zip files, however, are successfully extracted, with the same results as before.

What I mean to say is that we need unit tests added to the `tests` directory which will demonstrate this is the case for a valid and a malicious zip file. You can find tests for download_and_extract [here](https://github.com/Project-MONAI/MONAI/blob/dev/tests/apps/test_download_and_extract.py). I suggest we add two more tests in a new class in this file, one constructs a valid zip file in a temporary directory and shows `extractall` does not raise an exception, and a second with a maliciously defined zip file which does raise an exception as expected. This should suffice to capture the condition we're trying to cover with this change. You can use the `tempfile` and `zipfile` standard libraries to do this.

h3rrr:
> > Hi, @ericspod the pull request for SECURITY.md mentioned the need for testing. I've already tested this locally, and the image above shows the results. You can see that when a zip file contains malicious content, an error is thrown. Normal zip files, however, are successfully extracted, with the same results as before.你好，关于 SECURITY.md 中提到的拉取请求，提到了需要测试。我已经在本地上进行了测试，上面的图片显示了结果。你可以看到，当 zip 文件包含恶意内容时，会抛出错误。然而，正常的 zip 文件则成功提取，结果与之前相同。
> 
> What I mean to say is that we need unit tests added to the `tests` directory which will demonstrate this is the case for a valid and a malicious zip file. You can find tests for download_and_extract [here](https://github.com/Project-MONAI/MONAI/blob/dev/tests/apps/test_download_and_extract.py). I suggest we add two more tests in a new class in this file, one constructs a valid zip file in a temporary directory and shows `extractall` does not raise an exception, and a second with a maliciously defined zip file which does raise an exception as expected. This should suffice to capture the condition we're trying to cover with this change. You can use the `tempfile` and `zipfile` standard libraries to do this.我的意思是，我们需要在 `tests` 目录中添加单元测试，以证明对于有效和恶意的 zip 文件，情况都是如此。你可以在这里的 download_and_extract 中找到测试。我建议我们在这个文件中添加两个新的测试，在一个新类中，一个在临时目录中构建一个有效的 zip 文件并显示 `extractall` 不会抛出异常，另一个使用恶意定义的 zip 文件，它确实会抛出异常，正如预期的那样。这应该足以捕捉我们通过这个更改想要覆盖的条件。你可以使用 `tempfile` 和 `zipfile` 标准库来完成这个任务。

I understand what you mean. In the latest pull request, I fixed some potential exploits and added the corresponding testing module.
It seems the first submission was submitted as a previous pull request... Sorry.

tests/apps/test_download_and_extract.py
```diff
@@ -66,5 +68,187 @@ def test_default(self, key, file_type):
...
+        """Test protection against absolute paths in archives."""
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            # Create zip with absolute path
+            zip_path = Path(tmp_dir) / "absolute_path_test.zip"
+            extract_dir = Path(tmp_dir) / "extract"
+            extract_dir.mkdir()
+
+            with zipfile.ZipFile(zip_path, 'w') as zf:
+                # Try to use absolute path
+                zf.writestr("/etc/passwd", "malicious content")
```
ericspod: These tests all look good, but please change this to be something other than "/etc/passwd" just in case someone somehow overwrites that file. Perhaps use "/etc/passwd_bad" here and elsewhere instead.

ericspod:
This looks good now, just run `./runtests.sh --autofix` from the root MONAI directory to fix the failing test and we should be good to go.

tests/apps/test_download_and_extract.py
h3rrr: ok

monai/apps/utils.py
```diff
@@ -287,14 +319,30 @@ def extractall(
...
+            for member in zip_file.infolist():
+                safe_path = safe_extract_member(member, output_dir)
+                if member.is_dir():
+                    continue
+
+                os.makedirs(os.path.dirname(safe_path), exist_ok=True)
+                with zip_file.open(member) as source:
+                    with open(safe_path, "wb") as target:
+                        shutil.copyfileobj(source, target)
         return
```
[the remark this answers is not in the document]
ericspod: I think losing +x permissions is in fact a good thing in this case.

ericspod (approved):

PR #8607, opened 2025-10-30
Generalize AsymmetricUnifiedFocalLoss for multi-class and align interface

### LINKED ISSUE #8603: Add sigmoid/softmax interface for `AsymmetricUnifiedFocalLoss`

No

**Additional context**
No

ytl0623:
Fixes #8603  .

### Description

Addresses the limitations of `AsymmetricUnifiedFocalLoss` and its dependencies (`AsymmetricFocalLoss`, `AsymmetricFocalTverskyLoss`). The previous implementations were hard-coded for binary segmentation and lacked the standard `sigmoid`/`softmax`/`to_onehot_y` interface common to other MONAI losses.

### Types of changes

### DIFF
--- a/monai/losses/unified_focal_loss.py
+++ b/monai/losses/unified_focal_loss.py
@@ -192,29 +192,29 @@ class AsymmetricUnifiedFocalLoss(_Loss):
 
     def __init__(
         self,
-        include_background: bool = True,
         to_onehot_y: bool = False,
-        sigmoid: bool = False,
-        softmax: bool = False,
+        use_sigmoid: bool = False,
+        use_softmax: bool = False,
         lambda_focal: float = 0.5,
         focal_loss_gamma: float = 2.0,
         focal_loss_delta: float = 0.7,
         tversky_loss_gamma: float = 0.75,
         tversky_loss_delta: float = 0.7,
+        include_background: bool = True,
         reduction: LossReduction | str = LossReduction.MEAN,
     ):
         """
         Args:
-            include_background: whether to include loss computation for the background class. Defaults to True.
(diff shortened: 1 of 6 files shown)


ytl0623:
Hi @KumoLiu, @Nic-Ma  and @ericspod,

Sorry to bother.
Could you please review the PR or give some tips?

Thanks in advance!

monai/losses/unified_focal_loss.py
```diff
@@ -23,218 +23,272 @@
...
     Reimplementation of the Asymmetric Focal Tversky Loss described in:
 
     - "Unified Focal Loss: Generalising Dice and Cross Entropy-based Losses to Handle Class Imbalanced Medical Image Segmentation",
-    Michael Yeung, Computerized Medical Imaging and Graphics
+      Michael Yeung, Computerized Medical Imaging and Graphics
     """
 
     def __init__(
         self,
-        to_onehot_y: bool = False,
```
ericspod, line 37: Why is `to_onehot_y` being removed? I think users will still want this functionality so I would leave it in and add `include_background` as a new last argument. Even if we do want to remove it we need to use the [deprecation decorators to mark the argument removed](https://github.com/Project-MONAI/MONAI/blob/69f3dd26ed2a65e89ae89d951bb16f2dcb4d7c5d/monai/utils/deprecate_utils.py#L123) but still keep it for a version or two.

ericspod (commented):
Hi @ytl0623 thanks for the contribution. I had only a minor comment but please look at the greptile comments. A few look spurious they should all be resolved.

PR #8626, opened 2025-11-06
modify unified_focal_loss.py

### LINKED ISSUE #8603: Add sigmoid/softmax interface for `AsymmetricUnifiedFocalLoss`

ytl0623:
Fixes #8603

### Description

Asymmetric Unified Focal Loss with sigmoid/softmax interface and multiclass support.

### Types of changes


ericspod:
Hi @ytl0623, appreciate the contribution with these losses. As we're adding new arguments such as `include_background`, we should have tests for these being either True or False so that we're sure the classes work in both cases. These can be quite simple tests but should go with the existing tests for these classes. Thanks!

PR #8628, opened 2025-11-12
Fix #8599: Add track_meta and weights_only arguments to PersistentDataset for MetaTensor support.

### LINKED ISSUE #8599: PersistentDataset not usable anymore (v1.5.1) ?
Since the modifications to the PersistentDataset made in version 1.5.1, the persistent dataset does not allow to save and load metadata. This mean we cannot invert back the initial transforms or even know the original filename (and probably a lot more).

**To Reproduce**
1. Install MONAI version 1.5.1
2. Load and transform an image through the PersistentDataset
3. Re-load the image from cache
4. Try to invert and save the image

**Expected behavior**
The original image and outputted image should be the same

**Code used**

```
data = ["./image.nrrd"]
transforms = Compose([LoadImage(ensure_channel_first=True), Rotate90()])
ds=PersistentDataset(data, transform=transforms, cache_dir="./cache")
reloaded_data = ds[0]
reloaded_data = ds[0]
inverter = Invert(transforms)
inverted = inverter(reloaded_data)
saver = SaveImage("./out", output_ext=".nrrd")
saver(inverted)
```

Let me know if there is a way to bypass this behavior, thank you.

mccle:
Fixes #8599.

### Description

`PersistentDataset` currently casts all `MetaTensor` objects to `torch.Tensor` objects and forces the use of `torch.load` with `weights_only=True`. This makes it impossible to save or load metadata to cached files, which may be necessary for accurate post-transform operations. 

To address this, this PR introduces the `track_meta` and `weights_only` arguments directly to `PersistentDataset`. They are internally passed to `convert_to_tensor` and `torch.load`, respectively. A `ValueError` is raised when `track_meta=True` and `weights_only=True`, since `MetaTensor` objects cannot be loaded with `weights_only=True` and the cached files would be continually deleted and rewritten. 

These changes restore the ability to cache `MetaTensor` objects by allowing explicit control over data casting and `torch.load` behavior. The default values of `track_meta=False` and `weights_only=True` will preserve the current behavior of `PersistentDataset`.

### Types of changes

### DIFF
--- a/monai/data/dataset.py
+++ b/monai/data/dataset.py
@@ -230,6 +230,8 @@ class PersistentDataset(Dataset):
         pickle_protocol: int = DEFAULT_PROTOCOL,
         hash_transform: Callable[..., bytes] | None = None,
         reset_ops_id: bool = True,
+        track_meta: bool = False,
+        weights_only: bool = True
     ) -> None:
         """
         Args:
@@ -264,7 +266,10 @@ class PersistentDataset(Dataset):
                 When this is enabled, the traced transform instance IDs will be removed from the cached MetaTensors.
                 This is useful for skipping the transform instance checks when inverting applied operations
                 using the cached content and with re-created transform instances.
-
+            track_meta: whether to track the meta information, if `True`, will convert to `MetaTensor`.
+                default to `False`.
+            weights_only: keyword argument passed to `torch.load` when reading cached files.
+                default to `True`.
         """
         super().__init__(data=data, transform=transform)
         self.cache_dir = Path(cache_dir) if cache_dir is not None else None
@@ -280,6 +285,8 @@ class PersistentDataset(Dataset):
         if hash_transform is not None:
             self.set_transform_hash(hash_transform)
         self.reset_ops_id = reset_ops_id
+        self.track_meta = track_meta
+        self.weights_only = weights_only
 
     def set_transform_hash(self, hash_xform_func: Callable[..., bytes]):
         """Get hashable transforms, and then hash them. Hashable transforms
@@ -377,7 +384,7 @@ class PersistentDataset(Dataset):
 
         if hashfile is not None and hashfile.is_file():  # cache hit
             try:
-                return torch.load(hashfile, weights_only=True)
+                return torch.load(hashfile, weights_only=self.weights_only)
             except PermissionError as e:
                 if sys.platform != "win32":
                     raise e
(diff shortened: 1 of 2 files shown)


monai/data/dataset.py
```diff
@@ -264,7 +266,16 @@ def __init__(
                 When this is enabled, the traced transform instance IDs will be removed from the cached MetaTensors.
                 This is useful for skipping the transform instance checks when inverting applied operations
                 using the cached content and with re-created transform instances.
+            track_meta: whether to track the meta information, if `True`, will convert to `MetaTensor`.
+                default to `False`. Cannot be used with `weights_only=True`.
+            weights_only: keyword argument passed to `torch.load` when reading cached files.
+                default to `True`. When set to `True`, `torch.load` restricts loading to tensors and
+                other safe objects. Setting this to `False` is required for loading `MetaTensor`
+                objects saved with `track_meta=True`.
```
ericspod:
```suggestion
                other safe objects. Setting this to `False` is required for loading `MetaTensor`
                objects saved with `track_meta=True`, however this creates the possibility of remote
                code execution through `torch.load` so be aware of the security implications of doing so.
```

I'd like to explicitly state the security concern here that motivated the previous changes to the datasets.

tests/data/test_persistentdataset.py
```diff
@@ -168,6 +176,40 @@ def test_different_transforms(self):
...
+                    )
+
+            else:
+                test_dataset = PersistentDataset(
+                    data=test_data,
+                    transform=transform,
+                    cache_dir=cache_dir,
+                    track_meta=track_meta,
+                    weights_only=weights_only,
+                )
```
ericspod:
```suggestion

            cm = self.assertRaises(ValueError) if expected_error else contextlib.nullcontext()
            with cm:
                test_dataset = PersistentDataset(
                    data=test_data,
                    transform=transform,
                    cache_dir=cache_dir,
                    track_meta=track_meta,
                    weights_only=weights_only,
                )
```
I think this works and is equivalent, you'll have to import `contextlib`.

ericspod (approved):
Hi @mccle thanks for this fix. We had changed things in the past in response to a security concern so it's important to discuss the implication of loading more than just weights. Since `torch.load` still retains this functionality it should be enough to do as they do and make `weights_only` True by default as you have. We are investigating alternative storage representations that would be safe as well.

mccle:
Hello @ericspod, thank you for your quick response! I completely understand your concern about the potential security implications of using `torch.load` with `weights_only=False`, so I agree that your warning in the documentation will be helpful for users. I have also verified that your modification to the new test runs successfully with the addition of `import contextlib`, so I have committed both of those changes as well.

ericspod (approved):
I think this is good to merge. We will revisit this in the future after considering better alternatives to storing metatensors. @KumoLiu please trigger blossom when you can. Thanks!

PR #8633, opened 2025-11-14
Fix index using tuple for image cropping operation

johnzielke:
I'm getting a warning at this line:
"Using a non-tuple sequence for multidimensional indexing is deprecated and will be changed in pytorch 2.9; use x instead of x. In pytorch 2.9 this will be interpreted as     _tensor.py:1654
tensor index, x, which will result either in an error or a different result".

### Types of changes


ericspod (approved):
Thanks for spotting this, @johnzielke I think we can merge this one soon, @KumoLiu please trigger blossom when possible.

PR #8639, opened 2025-11-21
Fix #8350: Clarify LocalNormalizedCrossCorrelationLoss docstring

### LINKED ISSUE #8350: "LocalNormalizedCrossCorrelationLoss" Docstring is Ambiguous
The problem is about the docstring of "LocalNormalizedCrossCorrelationLoss" under losses.
1. It is unclear if the returned value is in the range [-1, 1] as a standard correlation metric.
2. It is not clear if the returned value will be negative or positive, also it is not clear if it is subject to minimization or maximization.

Let me elaborate on an example:

If for any project I want to increase the correlation between two images, I would calculate the correlation first then I would
negate it since the autograd library I would use would do minimization by default. Then I would also add +1 to it since
having negative loss terms when doing minimization is generally frowned upon, I am not sure for a good reason or not.

But this loss function directly returns a negative value, without going into the code it is not clear if this is because
the correlation is actually negative or not, should minimize or maximize the value, these are not at all clear and
absolutely can't be assumed, so this information should clearly be stated in the docstring, possibly with an additional
explanation like "smaller loss means high correlation" or as such.

1. In the docstring, state the range of the values that can be returned by the function if there is such a closed range.
2. State the "real calculation" to "loss function" conversion steps, what is the actual correlation value, and what are you applying to it before returning (applying log, multiplying with -1, etc.). So that people can interpret what a high or low loss value actually means in terms of correlation.
3. Optionally also state what a low or high loss value indicated in terms of correlation between the images for ease of usage.

I think there are endless ways of doing it, the main goal is clearing out any ambiguity, in whichever way possible. What I suggest seems reasonable to me.

**Additional context**
None

engmohamedsalah:
## Description
  This PR improves the docstring for `LocalNormalizedCrossCorrelationLoss` to address the ambiguities identified in
  #8350.

  ## Problem
  The current docstring does not clearly document:
  - The range of returned loss values
  - Whether the loss should be minimized or maximized
  - How to interpret high vs. low loss values

  ## Solution
  Enhanced the class docstring with comprehensive documentation including Returns section (value range, optimization
   direction) and Note section (implementation details, interpretation guidelines).

  ## Changes
  - Added `Returns` section with explicit value range and optimization direction
  - Added `Note` section explaining transformations and interpretation
  - Reorganized Args to class level for better discoverability
  - Followed MONAI formatting conventions

  ## Testing
  - [x] Verified docstring syntax is correct
  - [x] Confirmed technical accuracy by analyzing implementation
  - [x] Validated all three issue requirements are addressed

### DIFF
--- a/monai/losses/image_dissimilarity.py
+++ b/monai/losses/image_dissimilarity.py
@@ -51,6 +51,7 @@ kernel_dict = {
 class LocalNormalizedCrossCorrelationLoss(_Loss):
     """
     Local squared zero-normalized cross-correlation.
+
     The loss is based on a moving kernel/window over the y_true/y_pred,
     within the window the square of zncc is calculated.
     The kernel can be a rectangular / triangular / gaussian window.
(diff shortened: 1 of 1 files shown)


ericspod (approved):
Hi @engmohamedsalah thanks for the contribution. I think this is fine as it is so we can merge this soon.

Issue #7901, opened 2024-07-04
No space left on device when test with conda


KumoLiu:
```
        ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device
        
        [end of output]
    
    note: This error originates from a subprocess, and is likely not a problem with pip.
  error: subprocess-exited-with-error
  
  × pip subprocess to install backend dependencies did not run successfully.
  │ exit code: 1
  ╰─> See above for output.
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  
  Warning: 
  CondaEnvException: Pip failed
```

PR #8646, opened 2025-11-30
Fix type hint for image reader

Kirscher:
### Description
This PR fixes the type hint for image reader to correctly handle Cupy arrays and removes the TODO.

### Types of changes

### DIFF
--- a/monai/data/image_reader.py
+++ b/monai/data/image_reader.py
@@ -22,7 +22,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable, Iterator, Sequence
 from dataclasses import dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Union
 
 import numpy as np
 from torch.utils.data._utils.collate import np_str_obj_array_pattern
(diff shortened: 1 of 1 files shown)


monai/data/image_reader.py
```diff
@@ -57,6 +57,13 @@
 cp, has_cp = optional_import("cupy")
 kvikio, has_kvikio = optional_import("kvikio")
 
+if TYPE_CHECKING:
+    import cupy
+
+    NdarrayOrCupy = Union[np.ndarray, cupy.ndarray]
```
ericspod, line 63: This was done to get around Cupy being an optional dependency, this makes it necessary when type checking but this might not be avoidable right now. We do need to consider a way of defining placeholder types for optional dependencies so we can type check correctly without these workarounds.

ericspod (approved):
Hi @Kirscher thanks for these updates, I think it's fine though I agree we need a better way of doing the type checking when optional libraries are used.

ericspod:
Hi @KumoLiu The packaging test is failing when running out of space, is this possibly related to #7901? I remember this was solved with the fix where we clean out some things in the image being used.

PR #8651, opened 2025-12-01
Fix Invertd

### LINKED ISSUE #8396: `Invertd` Confused by Invertible Transforms in Postprocessing
**Describe the bug**
The `Invertd` class relies on a list of applied transforms stored in dictionaries or `MetaTensor` instances to know which transforms to invert and with what parameters. This gets confused if the postprocessing sequence of transforms contains an invertible transform before this class is used. 

**To Reproduce**
Run the following with the current version of MONAI:

```python
from monai.data import MetaTensor, create_test_image_2d
from monai.transforms import Compose, EnsureChannelFirstd, Invertd, Spacingd
from monai.transforms.utility.dictionary import Lambdad

img, _ = create_test_image_2d(60, 60, 2, 10, num_seg_classes=2)
img = MetaTensor(img, meta={"original_channel_dim": float("nan"), "pixdim": [1.0, 1.0]})
key = "image"
preprocessing = Compose([
        EnsureChannelFirstd(key),
        Spacingd(key, pixdim=[2.0, 2.0]),
])

postprocessing = Compose([
    Lambdad(key, func=lambda x: x),  # adds something to the applied operations of img
    Invertd(key, transform=preprocessing, orig_keys=key)  # gets confused by this item
])

item = {key: img}
pre = preprocessing(item)
post = postprocessing(pre)  # exception raised here
```

This raises the following exception:

```
  File "/home/localek10/workspace/monai/MONAI_mine/monai/transforms/transform.py", line 150, in apply_transform
    return _apply_transform(transform, data, unpack_items, lazy, overrides, log_stats)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/localek10/workspace/monai/MONAI_mine/monai/transforms/transform.py", line 98, in _apply_transform
    return transform(data, lazy=lazy) if isinstance(transform, LazyTrait) else transform(data)
                                                                               ^^^^^^^^^^^^^^^
  File "/home/localek10/workspace/monai/MONAI_mine/monai/transforms/spatial/dictionary.py", line 530, in inverse
    d[key] = self.spacing_transform.inverse(cast(torch.Tensor, d[key]))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/localek10/workspace/monai/MONAI_mine/monai/transforms/spatial/array.py", line 546, in inverse
    return self.sp_resample.inverse(data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/localek10/workspace/monai/MONAI_mine/monai/transforms/spatial/array.py", line 239, in inverse
    transform = self.pop_transform(data)
                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/localek10/workspace/monai/MONAI_mine/monai/transforms/inverse.py", line 348, in pop_transform
    return self.get_most_recent_transform(data, key, check, pop=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/localek10/workspace/monai/MONAI_mine/monai/transforms/inverse.py", line 330, in get_most_recent_transform
    self.check_transforms_match(all_transforms[idx])
  File "/home/localek10/workspace/monai/MONAI_mine/monai/transforms/inverse.py", line 287, in check_transforms_match
    raise RuntimeError(
RuntimeError: Error SpatialResample getting the most recently applied invertible transform Lambda 129323001074736 != 129322998584224.
```

**Expected behavior**
The `Invertd` should only try to invert those transforms in `preprocessing` and ignore the `Lambdad`.

**Environment**

```
Printing MONAI config...
MONAI version: 1.4.1rc1+30.g7c26e5af.dirty
Numpy version: 1.26.4
Pytorch version: 2.6.0+cu124
MONAI rev id: 7c26e5af385eb5f7a813fa405c6f3fc87b7511fa
MONAI __file__: /home/<username>/workspace/monai/MONAI_mine/monai/__init__.py

Optional dependencies:
Pytorch Ignite version: 0.4.11
ITK version: 5.4.2
Nibabel version: 5.3.2
scikit-image version: 0.25.2
scipy version: 1.15.2
Pillow version: 11.1.0
Tensorboard version: 2.19.0
gdown version: 5.2.0
TorchVision version: 0.21.0+cu124
tqdm version: 4.67.1
lmdb version: 1.6.2
psutil version: 7.0.0
pandas version: 2.2.3
einops version: 0.8.1
mlflow version: 2.20.2
pynrrd version: 1.1.3
clearml version: 1.17.2rc0

Printing system config...
System: Linux
Linux version: Ubuntu 24.04.2 LTS
Platform: Linux-6.11.0-17-generic-x86_64-with-glibc2.39
Processor: x86_64
Machine: x86_64
Python version: 3.12.9
Process name: python
Open files: [popenfile(path='/home/localek10/.vscode-server/data/logs/20250320T125602/remoteTelemetry.log', fd=19, position=0, mode='a', flags=33793), popenfile(path='/home/localek10/.vscode-server/data/logs/20250320T125602/ptyhost.log', fd=20, position=2102, mode='a', flags=33793), popenfile(path='/home/localek10/.vscode-server/data/logs/20250320T125602/remoteagent.log', fd=22, position=491, mode='a', flags=33793)]
Num physical CPUs: 8
Num logical CPUs: 16
Num usable CPUs: 16
CPU usage (%): [3.4, 3.0, 3.4, 3.9, 3.4, 3.4, 6.4, 31.6, 3.0, 3.0, 3.0, 3.0, 3.0, 3.0, 3.8, 70.9]
CPU freq. (MHz): 1653
Load avg. in last 1, 5, 15 mins (%): [1.5, 1.2, 0.8]
Disk usage (%): 10.6
Total physical memory (GB): 62.7
Available memory (GB): 51.7
Used memory (GB): 10.0

Printing GPU config...
Num GPUs: 2
Has CUDA: True
CUDA version: 12.4
cuDNN enabled: True
NVIDIA_TF32_OVERRIDE: None
cuDNN version: 90100
Current device: 0
GPU 0 Name: NVIDIA TITAN X (Pascal)
GPU 0 Is integrated: False
GPU 0 Multi processor count: 28
GPU 0 Total memory (GB): 11.9
GPU 0 CUDA capability (maj.min): 6.1
GPU 1 Name: NVIDIA GeForce GTX 980
GPU 1 Is integrated: False
GPU 1 Multi processor count: 16
GPU 1 Total memory (GB): 3.9
GPU 1 CUDA capability (maj.min): 5.2
```

**Additional context**
This is a similar issue to that raised in #8056 but is unrelated to threading.

eclipse0922:
Fixes #8396.

### Description

This PR fixes a regression where `Invertd` and direct `Compose.inverse()` calls could fail when later post-processing steps had appended unrelated entries to the same transform history.

Root cause:
- inverse execution assumed that the most recent history entry had to belong to the transform currently being inverted
- once post-processing appended additional history entries, inverse would pop from the tail, hit the wrong transform, and raise an ID mismatch

Solution:
- centralize inverse matching in the shared `TraceableTransform` history lookup path
- search backward for the most recent matching transform entry instead of assuming the tail entry must match
- pop only the matched preprocessing entry, leaving unrelated post-processing history intact
- remove the group-tracking plumbing that was previously introduced in `Compose`, `TraceKeys`, and `Invertd`

Regression coverage added for:
- the original `Invertd` reproducer from #8396
- multiple unrelated trailing post-processing transforms
- same-class trailing transform history on `MetaTensor`
- trace-key history with a different instance of the same invertible transform class
- direct `Compose.inverse()` with mixed post-processing history

### Types of changes


eclipse0922:
Hi @KumoLiu @Nic-Ma @ericspod,  
Sorry for the ping — just checking if you might have time to review this PR.  
It’s been up for a while and all CI checks are green.  
Thanks in advance!

atbenmurray:
I can take a look at this.

ericspod:
> I can take a look at this.

Please do have a look @atbenmurray, thanks. Sorry @eclipse0922 for the delay but we have a CI issue we need to sort with a patch release anyway.

monai/transforms/post/dictionary.py
```diff
@@ -859,6 +859,29 @@ def __init__(
...
+        for xform in all_transforms:
+            xform_group = xform.get(TraceKeys.GROUP)
+            if xform_group == target_group:
+                filtered.append(xform)
+
+        # If no transforms match (backward compatibility), return all transforms
+        if not filtered:
+            return all_transforms
+
+        return filtered
```
ericspod:
```suggestion
        filtered = [xform for xform in all_transforms if xform.get(TraceKeys.GROUP) == target_group]

        # If no transforms match (backward compatibility), return all transforms
        return filtered if filtered else all_transforms
```

Ignore if you think this is harder to read.

monai/transforms/post/dictionary.py
```diff
@@ -859,6 +859,29 @@ def __init__(
         self.post_func = ensure_tuple_rep(post_func, len(self.keys))
         self._totensor = ToTensor()
 
+    def _filter_transforms_by_group(self, all_transforms: list[dict]) -> list[dict]:
+        """
+        Filter applied_operations to only include transforms from the target Compose instance.
+        Uses automatic group tracking where Compose assigns its ID to child transforms.
+        """
+        from monai.utils import TraceKeys
```
ericspod: Imports should always be at the top of the source file and sorted according to isort. If you need to import this here to avoid circular importation please note that with a comment.

tests/transforms/inverse/test_invertd.py
```diff
@@ -137,6 +137,224 @@ def test_invert(self):
...
+
+        # This should NOT raise an error (was failing before the fix)
+        try:
+            post = postprocessing(pre)
+            # If we get here, the bug is fixed
+            self.assertIsNotNone(post)
+            self.assertIn(key, post)
+            print("SUCCESS! Automatic group tracking fixed the bug.")
+            print(f"  Preprocessing group ID: {id(preprocessing)}")
+            print(f"  Postprocessing group ID: {id(postprocessing)}")
```
ericspod: Tests in general should be silent unless they fail, this keeps the log easier to read so please remove these.

tests/transforms/inverse/test_invertd.py
```diff
@@ -137,6 +137,224 @@ def test_invert(self):
...
 
+    def test_invertd_with_postprocessing_transforms(self):
+        """Test that Invertd ignores postprocessing transforms using automatic group tracking.
+
+        This is a regression test for the issue where Invertd would fail when
+        postprocessing contains invertible transforms before Invertd is called.
+        The fix uses automatic group tracking where Compose assigns its ID to child transforms.
+        """
+        from monai.data import MetaTensor, create_test_image_2d
+        from monai.transforms.utility.dictionary import Lambdad
```
ericspod: These imports should all be at the top of the file and not in methods, same for those below.

tests/transforms/inverse/test_invertd.py
```diff
@@ -137,6 +137,224 @@ def test_invert(self):
...
+
+        # Should have operations from both pipelines with different groups
+        groups = [op.get("group") for op in pre2[key].applied_operations]
+        self.assertIn(str(id(preprocessing1)), groups)
+        self.assertIn(str(id(preprocessing2)), groups)
+
+        # Inverting preprocessing1 should only invert its transforms
+        inverter = Invertd(key, transform=preprocessing1, orig_keys=key)
+        inverted = inverter(pre2)
+        self.assertIsNotNone(inverted)
```
ericspod: Here it might be good to check the size of items in `inverted` to ensure the correct transforms were inverted only.

ericspod (changes requested):
Hi @eclipse0922 sorry for the delay on this one. I have had a look through and I think this solution will work. I'm a little worried about how complex our transform implementation is getting with fixes like these being added on, but this looks like a good solution to the problem. I would like @atbenmurray to have a look as well and I made a few comments.

monai/transforms/post/dictionary.py
eclipse0922: Applied in commit 5734cffe5: switched to a list comprehension for `filtered` and kept the backward-compatible fallback `return filtered if filtered else all_transforms`.

monai/transforms/post/dictionary.py
eclipse0922: Updated in commit 5734cffe5: moved `TraceKeys` to the module-level imports in `monai/transforms/post/dictionary.py` and removed the local import in `_filter_transforms_by_group`.

tests/transforms/inverse/test_invertd.py
eclipse0922: Removed the debug `print(...)` statements from `test_invertd_with_postprocessing_transforms` in commit 5734cffe5 so the test remains silent on success.

tests/transforms/inverse/test_invertd.py
eclipse0922: Updated in commit 5734cffe5: hoisted method-local imports to the file top (`MetaTensor`, `create_test_image_2d`, and `Lambdad`) and removed the in-method imports in these new tests.

tests/transforms/inverse/test_invertd.py
eclipse0922: Added explicit checks in `test_invertd_group_isolation` (commit 5734cffe5): `pre2[key].shape == (1, 40, 40)`, per-group occurrence counts for `preprocessing1`/`preprocessing2` in `applied_operations`, and `inverted[key].shape == (1, 60, 60)`.

monai/transforms/post/dictionary.py
```diff
@@ -859,6 +859,20 @@ def __init__(
...
+        Uses automatic group tracking where Compose assigns its ID to child transforms.
+        """
+        # Get the group ID of the transform (Compose instance)
+        target_group = str(id(self.transform))
+
+        # Filter transforms that match the target group
+        filtered = [xform for xform in all_transforms if xform.get(TraceKeys.GROUP) == target_group]
+
+        # If no transforms match (backward compatibility), return all transforms
+        return filtered if filtered else all_transforms
```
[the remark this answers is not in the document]
eclipse0922: Updated in commit f6777abd: `_filter_transforms_by_group` docstring now uses Google-style sections with explicit `Args` and `Returns`, including the `TraceKeys.GROUP` matching/backward-compat behavior.

monai/transforms/post/dictionary.py
```diff
@@ -894,8 +908,11 @@ def __call__(self, data: Mapping[Hashable, Any]) -> dict[Hashable, Any]:
...
             if orig_key in d and isinstance(d[orig_key], MetaTensor):
-                transform_info = d[orig_key].applied_operations
+                all_transforms = d[orig_key].applied_operations
                 meta_info = d[orig_key].meta
+
+                # Automatically filter by Compose instance group ID
+                transform_info = self._filter_transforms_by_group(all_transforms)
             else:
                 transform_info = d[InvertibleTransform.trace_key(orig_key)]
                 meta_info = d.get(orig_meta_key, {})
```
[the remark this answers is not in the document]
eclipse0922: Applied in commit f6777abd: the non-MetaTensor path now also calls `_filter_transforms_by_group(...)` before using `transform_info`, so group filtering is consistent for both metadata-storage paths.

tests/transforms/inverse/test_invertd.py
```diff
@@ -137,6 +138,209 @@ def test_invert(self):
...
+
+        # This should NOT raise an error (was failing before the fix)
+        try:
+            post = postprocessing(pre)
+            # If we get here, the bug is fixed
+            self.assertIsNotNone(post)
+            self.assertIn(key, post)
+        except RuntimeError as e:
+            if "getting the most recently applied invertible transform" in str(e):
+                self.fail(f"Invertd still has the postprocessing transform bug: {e}")
```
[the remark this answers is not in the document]
eclipse0922: Updated in commit f6777abd: removed the `try/except RuntimeError` wrapper in `test_invertd_with_postprocessing_transforms` and now call `post = postprocessing(pre)` directly so any exception fails the test naturally.

eclipse0922:
@ericspod 
I agree with the concern about transform complexity increasing as these fixes accumulate. For this PR, I kept the scope limited to the regression fix and tests so we can unblock the issue safely.

As follow-up work, I will open a separate refactor thread to reduce complexity in this area (without changing behavior):

centralize inverse transform selection/filtering logic in one shared path for both MetaTensor and trace-key data,
reduce implicit grouping mechanics by moving toward a clearer, explicit tracking flow,
add focused regression coverage around multi-pipeline and mixed post-processing cases.
This should let us keep this bugfix small while improving maintainability in the next step.

atbenmurray:
Taking a look

atbenmurray:
Okay, so I've started reviewing this finally, my apologies for the delay.

atbenmurray:
So, I've had a deeper look:

Going back to the motivating example in #8396, I'd observe the following:

```python
from monai.data import MetaTensor, create_test_image_2d
from monai.transforms import Compose, EnsureChannelFirstd, Invertd, Spacingd
from monai.transforms.utility.dictionary import Lambdad

img, _ = create_test_image_2d(60, 60, 2, 10, num_seg_classes=2)
img = MetaTensor(img, meta={"original_channel_dim": float("nan"), "pixdim": [1.0, 1.0]})
key = "image"
preprocessing = Compose([
        EnsureChannelFirstd(key),
        Spacingd(key, pixdim=[2.0, 2.0]),
])

postprocessing = Compose([
    Lambdad(key, func=lambda x: x),  # adds something to the applied operations of img
    Invertd(key, transform=preprocessing, orig_keys=key)  # gets confused by this item
])

item = {key: img}
pre = preprocessing(item)
post = postprocessing(pre)  # exception raised here
```

There seems to be some design tension between `Invertd` taking a sequence of transforms to be inverted relative to the logging of invertable transforms as applied operations. Ideally, we shouldn't have to pass the transforms at all, if all the required information is logged in the operations that have been performed on the data, but I'll proceed assuming we need them for now.

A cleaner approach might be to tweak the invert process so that it skips entries in the applied operations list that don't match the entries in the transforms passed to `Invertd`. I think that would achieve the same effect as the current PR without having to introduce a new group concept.

If we were to modify `pop_transform` and `get_most_recent_transform` so that each call to `pop_transform` pops the corresponding entry in `applied_operations` rather than requiring it to be the first entry in the list. It has the same effect as filtering on groups except it doesn't require groups


What Do You Think?

eclipse0922:
@atbenmurray I checked this more closely and I agree with your direction here.

The current group-based change fixes the motivating `Invertd` path from `#8396`, but it is adding complexity around tracing/filtering instead of addressing the inverse-selection problem directly. It also leaves `Compose.inverse()` on mixed histories relying on the original stack ordering.

I'm going to rework this PR in the direction you suggested: drop the group plumbing and move the fix into the inverse stack handling so we match/pop the entries corresponding to the transform sequence being inverted, while still failing if an expected entry cannot be found.

That keeps the scope on `#8396`, but moves the fix closer to the root cause.

atbenmurray (approved):
This seems much cleaner. It would be worth changing the docstring for `check_transforms_match` as the ID isn't used when running multiprocessing, and it might be worth adding a docstring to _transforms_match, but code wise it looks good

eclipse0922:
Hi @ericspod, the PR has been reworked based on @atbenmurray's suggestion (dropped the group approach, 
moved the fix into inverse stack matching). @atbenmurray has approved the updated version. 
Could you take another look when you get a chance? Thanks

ericspod (approved):
I think we're good now if both me and @atbenmurray have had a look. Thanks!

PR #8652, opened 2025-12-02
8627 perceptual loss errors out after hitting the maximum number of downloads

### LINKED ISSUE #8627: Perceptual Loss errors out after hitting the maximum number of downloads
**Describe the bug**
Currently, models are hosted in Google Drive and once the number of downloads hits the limit, the networks cannot be pulled.

**To Reproduce**
Depending on number of downloads, errors show up when instancing perceptual loss.

**Expected behavior**
No error.
We have  discussed porting these networks (python files defining them and artifacts) to Huggingface or similar.

**Additional context**
This has already been discussed internally.

virginiafdez:
Fixes #8627.
Moves the perceptual loss code to MONAI repository https://github.com/Project-MONAI/perceptual-models and the checkpoints to Huggingface. 

### Description
This PR changes and simplifies the torch.hub loading process and gets the models from Huggingface lirbary. 

### Types of changes

- [N/A] New tests added to cover the changes.
- [N/A] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [x]  In-line docstrings updated.


monai/losses/perceptual.py
```diff
@@ -205,11 +211,17 @@ class MedicalNetPerceptualSimilarity(nn.Module):
     """
 
     def __init__(
-        self, net: str = "medicalnet_resnet10_23datasets", verbose: bool = False, channel_wise: bool = False
+        self,
+        net: str = "medicalnet_resnet_10_23datasets",
```
ericspod: What we should do to improve security is to move "medicalnet_resnet10_23datasets" and any other valid name into constants then check `net` is one of them.

monai/losses/perceptual.py
```diff
@@ -205,11 +211,17 @@ class MedicalNetPerceptualSimilarity(nn.Module):
...
+        verbose: bool = False,
+        channel_wise: bool = False,
+        cache_dir: str | None = None,
     ) -> None:
         super().__init__()
         torch.hub._validate_not_a_forked_repo = lambda a, b, c: True
-        self.model = torch.hub.load("warvito/MedicalNet-models", model=net, verbose=verbose, trust_repo=True)
+        self.model = torch.hub.load(
```
ericspod, line 241: Perhaps this should have `trust_repo=True` as well since we do trust this repo and it's name is hardcoded here.

monai/losses/perceptual.py
```diff
@@ -205,11 +211,17 @@ class MedicalNetPerceptualSimilarity(nn.Module):
     """
 
     def __init__(
```
ericspod, line 230: We might need to check that `huggingface_hub` is installed as well, the `torch.hub.load` call will fail if so.

monai/losses/perceptual.py
virginiafdez: Done.

monai/losses/perceptual.py
virginiafdez: Done.

monai/losses/perceptual.py
virginiafdez: Done.

.github/workflows/pythonapp.yml
KumoLiu, line 1: - Removed the cache weekly timestamp step from the packaging job.
- Removed the heavyweight dependency installs that were triggering OSError: [Errno 28] No space left on device:
- Stopped installing the source archive with "[all]" extras (avoids large optional deps like itk, mlflow, clearml, lpips, etc.).
- Stopped installing requirements-dev.txt in the packaging quick-test phase; switched to requirements-min.txt + tests/min_tests.py.
- Switched packaging to CPU-only PyTorch on Ubuntu using the PyTorch CPU index, and installed it with --no-cache-dir to avoid caching huge wheels.
- Reintroduced a controlled pip cache for packaging:
     Cache path is pip cache dir.
     Cache key uses hashFiles('requirements*.txt', 'setup.cfg', 'pyproject.toml') so it updates when dependencies change.
      Kept torch installs --no-cache-dir to reduce disk pressure, while allowing smaller deps (e.g., requirements-min.txt) to benefit from caching.

cc @ericspod

ericspod: This file has been renamed and revamped with many of these changes integrated. It works now to resolve the issues with running out of space so we should delete this file to allow merging.

ericspod:
Hi @virginiafdez we should be able to get this through now, I removed a conflicting file but otherwise there's a few minor things to address. Would you be able to get back to this?

ericspod (dismissed):

ericspod (approved):

ericspod:
Thanks! I'll set the DCO to pass as you said that was fine to do as it's all your code.

PR #8665, opened 2025-12-19
Weights in alpha for FocalLoss

### LINKED ISSUE #8601: Weights in alpha for FocalLoss
Currently, MONAI’s FocalLoss accepts alpha as a single scalar to balance background vs. foreground. This prevents specifying different weights for each class in multi-class segmentation, which is problematic for highly imbalanced datasets.
I would like FocalLoss to support alpha as a tensor or list of floats, allowing a separate weight for each class (e.g., [0.1, 0.3, 0.6] for a 3-class problem). Allowing per-class alpha directly would provide more precise control and better alignment with the original focal loss formulation.

ytl0623:
Fixes #8601

### Description

Support alpha as a list, tuple, or tensor of floats, in addition to the existing scalar support.

### Types of changes

### DIFF
--- a/monai/losses/focal_loss.py
+++ b/monai/losses/focal_loss.py
@@ -168,7 +168,9 @@ class FocalLoss(_Loss):
         if self.use_softmax:
             if not self.include_background and self.alpha is not None:
                 if isinstance(self.alpha, (float, int)):
-                    warnings.warn("`include_background=False`, scalar `alpha` ignored when using softmax.")
+                    warnings.warn(
+                        "`include_background=False`, scalar `alpha` ignored when using softmax.", stacklevel=2
+                    )
             loss = softmax_focal_loss(input, target, self.gamma, self.alpha)
         else:
             loss = sigmoid_focal_loss(input, target, self.gamma, self.alpha)
--- a/monai/losses/focal_loss.py
+++ b/monai/losses/focal_loss.py
@@ -112,12 +112,17 @@ class FocalLoss(_Loss):
         self.include_background = include_background
         self.to_onehot_y = to_onehot_y
         self.gamma = gamma
-        self.alpha = alpha
         self.weight = weight
         self.use_softmax = use_softmax
         weight = torch.as_tensor(weight) if weight is not None else None
         self.register_buffer("class_weight", weight)
         self.class_weight: None | torch.Tensor
+        self.alpha: float | torch.Tensor | None
+
+        if isinstance(alpha, (list, tuple)):
+            self.alpha = torch.tensor(alpha)
+        else:
+            self.alpha = alpha
 
     def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
         """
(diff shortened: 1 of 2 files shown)


ytl0623:
Hi @ericspod @KumoLiu @Nic-Ma ,

sorry to bother u. the CI is failing with `[Errno 28] No space left on device` during the build-docs and packaging steps. This seems to be a server-side issue. Could u plz re-trigger the workflows?

Thanks in advance!

monai/losses/focal_loss.py
```diff
@@ -156,13 +164,19 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
         loss: Optional[torch.Tensor] = None
         input = input.float()
         target = target.float()
+
+        alpha_arg: float | torch.Tensor | None = self.alpha
+        if isinstance(alpha_arg, torch.Tensor):
+            alpha_arg = alpha_arg.to(input.device)
```
ericspod: If this necessary here if `sigmoid_focal_loss` and `softmax_focal_loss` do this conversion as well?

ericspod:
Hi @ytl0623 thanks for this, we should have the CICD issues sorted now. It looks good to me though I think we need to add tests for the class with a sequence `alpha` value to ensure this works correctly. Please add this then update the branch to incorporate the CICD fixes.

monai/losses/focal_loss.py
ytl0623: You are right. This block is redundant, so I will remove it.

ytl0623:
> Hi @ytl0623 thanks for this, we should have the CICD issues sorted now. It looks good to me though I think we need to add tests for the class with a sequence `alpha` value to ensure this works correctly. Please add this then update the branch to incorporate the CICD fixes.

Hi @ericspod, thanks for the feedback. I have removed the redundant conversion logic and added a new test case. It should be ready for another look. Thanks in advance!

tests/losses/test_focal_loss.py
```diff
@@ -374,6 +374,47 @@ def test_script(self):
...
+    def test_alpha_sequence_broadcasting(self):
+        """
+        Test FocalLoss with alpha as a sequence for proper broadcasting.
+        """
+        num_classes = 3
+        alpha_seq = [0.1, 0.5, 2.0]
+        batch_size = 2
+        spatial_dims = (4, 4)
+
+        devices = ["cpu", "cuda"] if torch.cuda.is_available() else ["cpu"]
```
ericspod: We have [this definition](https://github.com/Project-MONAI/MONAI/blob/01711cf7e3af991aabc92cc3fdf4c0b65949e65e/tests/test_utils.py#L874C1-L874C14) to use in place of this line if you want. Another approach is to have a `device` argument and use the `parameterized.expand` decorator [like here](https://github.com/Project-MONAI/MONAI/blob/01711cf7e3af991aabc92cc3fdf4c0b65949e65e/tests/data/meta_tensor/test_meta_tensor.py#L201).

tests/losses/test_focal_loss.py
```diff
@@ -374,6 +374,47 @@ def test_script(self):
             test_input = torch.ones(2, 2, 8, 8)
             test_script_save(loss, test_input, test_input)
 
+    def test_alpha_sequence_broadcasting(self):
```
ericspod: Here I think we do need another test, or use `parameterized.expand`, to test with and without the background.

ericspod:
Just about there, let's add a test for background=False and the stacklevel change for the warning and we'll be good to go.

tests/losses/test_focal_loss.py
ytl0623: Done.

ytl0623:
Thank you for the feedback. I have refactored the tests to use `@parameterized.expand`.

ericspod (approved):
Looks good now!

ericspod (approved):

PR #8666, opened 2025-12-19
[ADD] soft_label argument in DiceCELoss 

sifaoso:
Fixes # .

### Description

MONAI's DiceLoss supports soft_label in case the segmentation mask (i.e., the target) contains non-binary values. This functionality is already implemented. However, when using DiceCELoss it is not possible to specify the argument 'soft_label' for the DiceLoss part. 
This simple fix adds 'soft_label' as an argument of the DiceCELoss to be able to control whether the DiceLoss part should use the soft or hard label formulation.

### Types of changes

### DIFF
--- a/monai/losses/dice.py
+++ b/monai/losses/dice.py
@@ -674,6 +674,7 @@ class DiceCELoss(_Loss):
         lambda_dice: float = 1.0,
         lambda_ce: float = 1.0,
         label_smoothing: float = 0.0,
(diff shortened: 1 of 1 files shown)


ericspod (commented):
Hi @sifaoso thanks for this submission which looks good to me, but please add a simple unit test for this class with `soft_label=True` to be sure we're good.

PR #8669, opened 2025-12-22
Add sigmoid/softmax interface for AsymmetricUnifiedFocalLoss

### LINKED ISSUE #8603: Add sigmoid/softmax interface for `AsymmetricUnifiedFocalLoss`

ytl0623:
Fixes #8603 

### Description

This PR refactors `AsymmetricUnifiedFocalLoss` to extend support from binary-only to multi-class segmentation and aligns the interface with `FocalLoss`.

Key Changes:
1. Multi-class Support
2. Interface Standardization
3. Unit Tests

### Types of changes

### DIFF
--- a/monai/losses/unified_focal_loss.py
+++ b/monai/losses/unified_focal_loss.py
@@ -32,7 +32,7 @@ class AsymmetricUnifiedFocalLoss(_Loss):
 
     - "Unified Focal Loss: Generalising Dice and Cross Entropy-based Losses to Handle Class Imbalanced Medical Image Segmentation",
     Michael Yeung, Computerized Medical Imaging and Graphics
-    
+
     Example:
         >>> import torch
         >>> from monai.losses import AsymmetricUnifiedFocalLoss
--- a/monai/losses/unified_focal_loss.py
+++ b/monai/losses/unified_focal_loss.py
@@ -73,6 +73,12 @@ class AsymmetricUnifiedFocalLoss(_Loss):
             epsilon: Small value to prevent division by zero or log(0). Defaults to 1e-7.
         """
         super().__init__(reduction=LossReduction(reduction).value)
+        if not 0 <= weight <= 1:
+            raise ValueError(f"weight must be in [0, 1], got {weight}")
+        if not 0 <= delta <= 1:
+            raise ValueError(f"delta must be in [0, 1], got {delta}")
+        if gamma <= 0:
+            raise ValueError(f"gamma must be > 0, got {gamma}")
         self.weight = weight
         self.delta = delta
         self.gamma = gamma
@@ -97,7 +103,7 @@ class AsymmetricUnifiedFocalLoss(_Loss):
 
         if self.to_onehot_y:
             if n_pred_ch == 1:
-                warnings.warn("single channel prediction, `to_onehot_y=True` ignored.")
+                warnings.warn("single channel prediction, `to_onehot_y=True` ignored.", stacklevel=2)
             else:
                 if target.shape[1] == 1:
                     target = one_hot(target, num_classes=n_pred_ch)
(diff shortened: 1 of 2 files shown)


monai/losses/unified_focal_loss.py
```diff
@@ -162,79 +241,72 @@ def __init__(
...
         """
         Args:
-            to_onehot_y : whether to convert `y` into the one-hot format. Defaults to False.
-            num_classes : number of classes, it only supports 2 now. Defaults to 2.
-            delta : weight of the background. Defaults to 0.7.
-            gamma : value of the exponent gamma in the definition of the Focal loss. Defaults to 0.75.
-            epsilon : it defines a very small number each time. simmily smooth value. Defaults to 1e-7.
-            weight : weight for each loss function, if it's none it's 0.5. Defaults to None.
-
-        Example:
```
ericspod, line 175: We're loosing this example in the changes, I think it should be added back to the docstring as it's still useful to demonstrate something basic.

monai/losses/unified_focal_loss.py
```diff
@@ -162,79 +241,72 @@ def __init__(
...
         """
         Args:
-            y_pred : the shape should be BNH[WD], where N is the number of classes.
-                It only supports binary segmentation.
-                The input should be the original logits since it will be transformed by
-                    a sigmoid in the forward function.
-            y_true : the shape should be BNH[WD], where N is the number of classes.
-                It only supports binary segmentation.
-
-        Raises:
```
ericspod, line 203: We should still include what conditions raise exceptions, it's changed now but there's still raise statements in the code.

tests/losses/test_unified_focal_loss.py
```diff
@@ -19,47 +19,92 @@
...
 
     def test_with_cuda(self):
-        loss = AsymmetricUnifiedFocalLoss()
-        i = torch.tensor([[[[1.0, 0], [0, 1.0]]], [[[1.0, 0], [0, 1.0]]]])
-        j = torch.tensor([[[[1.0, 0], [0, 1.0]]], [[[1.0, 0], [0, 1.0]]]])
-        if torch.cuda.is_available():
-            i = i.cuda()
-            j = j.cuda()
+        if not torch.cuda.is_available():
+            print("CUDA not available, skipping test_with_cuda")
```
ericspod: Please don't print things in the tests, the output just gets very large when this happens. Other tests in MONAI do this but we need to clean those up as well.

ericspod (commented):
Hi @ytl0623 thanks for this contribution! Please look at the comments from Coderabbit, I think some of the logic does need to be double checked. I have gone over the code briefly and mentioned a few things, but once you've looked at the Coderabbit reviews I can revisit things.

ytl0623:
> Hi @ytl0623 thanks for this contribution! Please look at the comments from Coderabbit, I think some of the logic does need to be double checked. I have gone over the code briefly and mentioned a few things, but once you've looked at the Coderabbit reviews I can revisit things.

Hi @ericspod, Apologies for the confusion in the previous commit 😅. I've fixed the issues and all tests passed now. Ready for re-review. Thanks!

ericspod:
Hi @ytl0623 thanks for the addition here, however I think we should reconsider how this is refactored. The two other loss classes are removed in this refactor, although they weren't public (in that they appeared in the `__init__.py` file for `losses`) I still think they should be retained in case they are being used. If they were removed they should be replaced with stub classes with the `deprecated` decorator anyhow (#8680 has an example of doing this). 

Secondly, with the class that is present the constructor's arguments are reordered and some values removed, this would require using the `deprecated_arg` decorator to mark this as the case. I would instead suggest that you leave the order of arguments alone and add new ones at the end of the argument list to avoid needing to do this. If you do want to remove one argument you need to mark it as deprecated and put it at the end of the argument list, but just ignore its value in the constructor.

PR #8672, opened 2025-12-24
ENH: support additional dtypes in pad_nd

### LINKED ISSUE #7842: Support more dtypes in `pad_nd`
To be brief, I just have no idea why `torch.bool` (and other integer dtypes except for torch.int32) is considered to be not supported by PyTorch (it seems to work for me with PyTorch 2.3). I will be glad to add a PR.

https://github.com/Project-MONAI/MONAI/blob/4029c422f1da84cdf5e4ce546dbff33245cbe025/monai/transforms/croppad/functional.py#L99-L104

shubham-61969:
Prefer the PyTorch padding backend when supported and safely fall back
to NumPy on error. Add unit tests to validate backend selection and
ensure output dtype is preserved.

Fixes #7842

### Description

This pull request relaxes dtype restrictions in `pad_nd` and prefers
the PyTorch padding backend when supported, with a safe fallback to
NumPy on error. This enables support for additional dtypes (e.g. bool)
that are already handled correctly by recent PyTorch versions.

Unit tests are added to validate backend selection and ensure dtype
preservation.

### Types of changes


monai/transforms/croppad/functional.py
```diff
@@ -96,19 +96,24 @@ def pad_nd(
...
-            torch.bool,
-            torch.uint8,
-        }:
+        if mode in {"constant", "reflect", "edge", "replicate", "wrap", "circular"}:
+            # Try PyTorch pad for these modes; fallback to NumPy on error.
             _pad = _pt_pad
-        return _pad(img, pad_width=to_pad, mode=mode, **kwargs)
+        call_kwargs = dict(kwargs)
+        if mode != "constant":
+            call_kwargs.pop("value", None)
```
ericspod: If this condition isn't present, the effect of providing a `value` argument and not use "constant" mode is to raise an exception in the pad routine used, I think this is intended behaviour. Here if the `value` argument is removed this silently allows unintended arguments to be ignored, it's better to raise an exception instead.

tests/transforms/croppad/test_pad_nd_dtypes.py
```diff
@@ -0,0 +1,70 @@
+# Copyright (c) MONAI Consortium
```
ericspod, line 1: In MONAI we use `unittest` and `parameterized` packages for tests and not `pytest`, specifically we use test classes and methods for unit tests rather than functions. Please reformulate these tests to use these packages according the style of other tests that are present here. It might make sense to add your tests to an existing file rather than a new one, have a look at existing files to see if it does make sense that way.

ericspod (commented):
Hi @shubham-61969 thanks for the contribution. I think the change itself is fine with a few comments, the tests do need to be reformulated with `unittest` in particular. I think the previous implementation with its limitations was a result of older PyTorch versions so it's good to get this fix in. Please have a look again and then we can rereview.

shubham-61969:
Hi @ericspod , thanks for the feedback.

I’ve updated the implementation to explicitly raise an error when value is provided with modes other than "constant", so the original contract is preserved. I also reworked the tests to use unittest + parameterized and aligned them with MONAI’s existing test style.

All GitHub CI checks are now passing.

Could you please take another look? If there are any further changes you’d like, I’m happy to address them otherwise I believe this should be ready to merge.

monai/transforms/croppad/functional.py
```diff
@@ -96,19 +96,22 @@ def pad_nd(
...
-            torch.int16,
-            torch.int64,
-            torch.bool,
-            torch.uint8,
-        }:
+        if mode in {"constant", "reflect", "edge", "replicate", "wrap", "circular"}:
+            # Try PyTorch pad for these modes; fallback to NumPy on error.
             _pad = _pt_pad
-        return _pad(img, pad_width=to_pad, mode=mode, **kwargs)
+        call_kwargs = dict(kwargs)
```
ericspod: We don't actually need `call_kwargs` now as you can refer to `kwargs` directly.

tests/transforms/croppad/test_pad_nd_dtypes.py
```diff
@@ -0,0 +1,101 @@
...
+            patch.object(F, "_np_pad", wraps=F._np_pad) as mock_np,
+        ):
+            out = pad_nd(img, to_pad, mode="constant", value=0)
+
+        self.assertTrue(mock_pt.called)
+        self.assertTrue(mock_np.called)
+        self.assertEqual(out.dtype, img.dtype)
+        self.assertEqual(out.shape, (1, 6, 8))
+
+    @parameterized.expand([torch.bool, torch.int8, torch.int16, torch.int32, torch.int64, torch.uint8, torch.float32])
```
ericspod: Often it's cleaner to put arguments to `expand` into a variable at the top of the file, here and in the next usage it just is cleaner to read the tests.

tests/transforms/croppad/test_pad_nd_dtypes.py
```diff
@@ -0,0 +1,101 @@
...
+
+        self.assertEqual(out.shape, (1, 6, 8))
+        self.assertEqual(out.dtype, img.dtype)
+
+    def test_value_with_non_constant_mode_raises(self):
+        """Test that pad_nd raises ValueError when 'value' is provided with non-constant mode."""
+        img = torch.ones((1, 4, 4))
+        to_pad = [(0, 0), (1, 1), (2, 2)]
+        with self.assertRaises(ValueError):
+            pad_nd(img, to_pad, mode="reflect", **{"value": 0})
```
ericspod:
```suggestion
            pad_nd(img, to_pad, mode="reflect", value=0)
```

ericspod (approved):
Looks good to me now, I had some minor comments only. We'll try to get this merge shortly when other things have been resolved.

monai/transforms/croppad/functional.py
shubham-61969: My bad, I completely missed switching it back to kwargs during the refactor. Thanks for pointing that out!

tests/transforms/croppad/test_pad_nd_dtypes.py
shubham-61969: Noted!

shubham-61969:
Hello @ericspod, 
All suggested changes are now applied would appreciate a quick re-review when you have a moment.
Thanks!

ericspod (approved):
Hi @shubham-61969 sorry for the long delay, I think this looks good now, thanks!

PR #8690, opened 2026-01-12
Fix align_corners mismatch in AffineTransform

aymuos15:
## Summary
- Fixed inconsistent `align_corners` parameter in `AffineTransform`
- The `to_norm_affine` call was using hardcoded `align_corners=False` while `affine_grid` and `grid_sample` used `self.align_corners` (default `True`), causing half-pixel offset
- Changed to use `self.align_corners` consistently across the coordinate transformation pipeline

## Changes
- Modified `monai/networks/layers/spatial_transforms.py` to use `self.align_corners` instead of hardcoded `False`
- Updated test expected values to reflect correct behavior
- Added test cases to verify `align_corners` consistency

## Test Plan
- All existing tests pass with updated expected values
- New tests verify alignment behavior with both `align_corners=True` and `align_corners=False`
- Identity affine transforms now produce pixel-perfect outputs regardless of `align_corners` setting


atbenmurray:
Hi, taking a look at this

atbenmurray:
Sorry for the delay. On it now

aymuos15:
@atbenmurray 

I fixed the merge conflicts. Anything else to do here?

atbenmurray:
Reviewing now

monai/transforms/spatial/array.py
```diff
@@ -540,7 +540,8 @@ def __call__(
         if self.recompute_affine and isinstance(data_array, MetaTensor):
             if lazy_:
                 raise NotImplementedError("recompute_affine is not supported with lazy evaluation.")
-            a = scale_affine(original_spatial_shape, actual_shape)
+            ac = align_corners if align_corners is not None else False
```
atbenmurray, line 543: Looks like this change requires `Spacing` to set `align_corners` to self during `__init__` which it doesn't appear to do at the moment. Either that or it needs to fetch it from self.sp_resample.align_corners. Or am I wrong and is this `align_corners` independent from the one set during `__init__`?

aymuos15:
Hi @atbenmurray Do you think there is anything else left to do here? thanks :D

atbenmurray (commented):
Can you address the one comment? It might not need a change

aymuos15:
@atbenmurray thank you very much for pointing that out. That had slipped me. Have fixed it now properly

ericspod:
Hi @aymuos15 please check the Coderabbit comments to see if there's anything to address then marked as resolved. With that, if @atbenmurray can approve we can merge this soon.

aymuos15:
> Hi @aymuos15 please check the Coderabbit comments to see if there's anything to address then marked as resolved. With that, if @atbenmurray can approve we can merge this soon.

Everything should be addressed now, thank you!

atbenmurray (approved):
Everything looks resolved from here!

Issue #6753, opened 2023-07-21
PersistentDataset and CacheDataset hybrid


ibro45:
**Is your feature request related to a problem? Please describe.**
`CacheDataset` preprocesses the non-random transforms and loads the data into RAM.
`PersistentDataset` preprocesses the non-random transforms into pickled files on its first run, and any subsequent run reads them on the fly and applies the random transforms only.

However, when prototyping, you often rerun a setup with different hyper-parameters, and you end up waiting each time for the `CacheDataset` to preprocess the non-random transforms all over again. Using `PersistentDataset` on the other hand, won't require preprocessing them again at each run, but could still be slower than `CacheDataset` as it reads objects from the drive instead of RAM.

**Describe the solution you'd like**
I propose a combination of the two, that could also be framed as an extension to `PersistentDataset` that will allow loading of the pickled files into RAM. This way, the non-random transforms are only ever done once instead of always redoing them when loading the data into RAM.

ibro45:
This is a nice addition, thanks for pointing it out!

I was suggesting something a bit different. Maybe I should frame it this way - a `CacheDataset` that on the first run pickles the non-random transformed objects, and in subsequent runs just loads them directly into the RAM, without requiring to re-do the non-random transformation.

Or, a `PersistentDataset` whose pickled files are loaded into the RAM, instead of being read from the drive throughout the training. The current `PersistentDataset` behavior, if it had `cache_rate`, would be as if it were `cache_rate=0`. I'm basically suggesting a `PersistentData` with `cache_rate` like in `CacheDataset` specifying how much data should be kept in the RAM constantly.

wyli:
thanks @ibro45  that's a good idea, adding the feature request label here..

PR #8691, opened 2026-01-12
Add in_memory option to PersistentDataset for hybrid caching

aymuos15:
## Summary

Addresses #6753 

This PR adds an `in_memory` option to `PersistentDataset` that combines the benefits of persistent disk storage with fast RAM access:

- **First access**: Data is loaded from disk cache (or preprocessed if not cached) and stored in RAM
- **Subsequent accesses**: Data is served directly from RAM

This provides the best of both worlds:
- **vs default PersistentDataset**: No repeated disk reads during training epochs
- **vs CacheDataset**: No reprocessing on script restart

### Changes

- Add `in_memory: bool = False` parameter to `PersistentDataset`
- Add `memory_cache_size` property for inspecting cache state
- Optimize `_cachecheck()` to compute cache key once
- Clear memory cache when `set_data()` is called
- Add comprehensive unit tests including automatic hybrid caching test

### Usage

```python
from monai.data import PersistentDataset

# Hybrid mode: disk persistence + RAM caching
ds = PersistentDataset(
    data=data,
    transform=transform,
    cache_dir="./cache",
    in_memory=True  # NEW
)
```

## Test plan

- [x] Unit tests for `in_memory` cache functionality
- [x] Unit tests for `cache_dir=None` (pure RAM mode)
- [x] Unit tests for automatic hybrid caching behavior
- [x] Verified `set_data()` clears memory cache
- [x] Benchmarked against default PersistentDataset and CacheDataset


ericspod:
Hi @aymuos15 this looks good to me in general, though I wonder about the warning from Coderabbit regarding your tests. Can yoiu check that the results it's complaining about are actually working? I think it is correct that pi should be added to values. We also are waiting for a hotfix version to be released so it might be a while before we can merge this. Thanks!

aymuos15:
Thank you very much for the update. I have addressed the code rabbit comment.

ericspod:
> Thank you very much for the update. I have addressed the code rabbit comment.

I'm not sure I see anything different in the code. Do the commits you've made since my comment change things? Or was the Coderabbit comment spurious?

aymuos15:
I hadn't realised I did not add the change back then. Apologies. Have actually addressed the code rabbit comment now.

garciadias:
It seems this got stale. It's seems like an important issue to solve. Is there anything I can do to help?

aymuos15:
Hey @garciadias can you give this a review tomorrow? I will get to it by then? It shouldve ben okay as is, ill solve th conficts

garciadias:
> Hey @garciadias can you give this a review tomorrow? I will get to it by then? It shouldve ben okay as is, ill solve th conficts

Hi @aymuos15, there seems to be some work you need to do to solve the conflicts and run the CI/CD. I am fairly busy at the moment, so I can't promise, but if you solve these today, I could try to review it tomorrow.

Are you at BH? I think I sit behind you; there is a guy with your profile picture on their wallpaper. Say hi sometime if you see me.

monai/data/dataset.py
```diff
@@ -428,8 +442,11 @@ def _cachecheck(self, item_transformed):
         return _item_transformed
 
     def _transform(self, index: int):
-        pre_random_item = self._cachecheck(self.data[index])
-        return self._post_transform(pre_random_item)
+        if not self.in_memory:
+            return self._post_transform(self._cachecheck(self.data[index]))
+        if index not in self._memory_cache:
+            self._memory_cache[index] = self._cachecheck(self.data[index])
+        return self._post_transform(self._memory_cache[index])
```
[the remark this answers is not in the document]
aymuos15: latest commit resolves this

monai/data/dataset.py
```diff
@@ -428,8 +446,22 @@ def _cachecheck(self, item_transformed):
...
+        Fetch the pre-random-transform item for `index` and apply the random transforms to it.
+
+        Args:
+            index: index of the item in `self.data`.
+
+        Returns:
+            The fully transformed data element.
+
+        """
+        if not self.in_memory:
```
garciadias, line 462: The RAM cache stores whatever _cachecheck() returns on first access, but that's the raw pre-transform value -- it never goes through the convert_to_tensor(..., track_meta=self.track_meta) normalization that happens before torch.save (around line 414). Without in_memory, every access after the first goes through the disk-hit branch of _cachecheck and gets that normalized (tensor/MetaTensor) type back via torch.load. With in_memory=True, the first computed value is cached verbatim and reused for every later access in that process, so it can permanently stay as e.g. a plain numpy.ndarray even when track_meta=True is set (which is documented elsewhere as "converts to MetaTensor"). I verified this with a minimal repro: a pre-random transform that leaves a field as numpy.ndarray returns numpy.ndarray on every access once in_memory=True, vs. numpy.ndarray then torch.Tensor (1st vs 2nd+ access) when in_memory=False. Suggested fix: when first populating self._memory_cache[index], run it through the same convert_to_tensor(..., track_meta=self.track_meta) call used before torch.save, so the RAM-cached entry matches what a disk-cache hit would return.

monai/data/dataset.py
```diff
@@ -276,6 +277,11 @@ def __init__(
                 Setting to `False` should only be done if it's absolutely necessary to load unsafe pickled data,
                 eg. MetaTensor objects with unsafe objects in their metadata. Users must verify the safety of the data
                 they intend to load before doing so.
+            in_memory: if `True`, also keep the pre-processed data in RAM after first access, so that later
```
garciadias, line 280: _memory_cache is a plain per-process dict. With the common DataLoader(num_workers>0, persistent_workers=False) setup (the PyTorch default), worker processes are recreated every epoch, so the RAM cache is rebuilt from scratch each epoch and most of the "subsequent epochs skip disk reads" benefit is lost silently. This file already documents the equivalent caveat for CacheDataset/SmartCacheDataset (recommending persistent_workers=True); worth adding the same note here. Suggested change: add a line to the in_memory docstring noting that persistent_workers=True (or num_workers=0) is needed to actually retain the RAM cache across epochs.

garciadias (changes requested):
Thanks for this — the hybrid disk+RAM caching idea is solid and the mutation-isolation handling (deep-copying before post-transform) is correctly done, with good test coverage for the restart/rebuild scenario.

I found one issue worth addressing before merge: the RAM cache stores the raw pre-transform value on first access and never goes through the `convert_to_tensor(..., track_meta=self.track_meta)` normalization that the disk-cache round trip applies on every subsequent read. This means the returned type can silently differ (e.g. staying a plain `numpy.ndarray` instead of becoming `torch.Tensor`/`MetaTensor`) purely depending on whether `in_memory` is enabled — including when `track_meta=True` is set. See the inline comment for a concrete repro and suggested fix.

Also left a minor note about documenting the interaction with `DataLoader(num_workers>0, persistent_workers=False)`, where the RAM cache gets rebuilt every epoch and loses most of its benefit.

aymuos15:
Thanks for the detailed review. Both points are addressed in the latest commits:

- **`876b7998`** — the RAM-cached value now runs through the same `convert_to_tensor(..., track_meta=self.track_meta)` normalization used before `torch.save`, so the returned type matches a disk-cache hit (added a regression test for it).
- **`247a1d87`** — added the `persistent_workers=True` / `num_workers=0` caveat to the `in_memory` docstring.

PR #8695, opened 2026-01-13
Replace pickle with JSON in Auto3DSeg algo serialization

### LINKED ISSUE #8586: Update Auto3DSeg
Auto3DSeg uses pickle for communicating between processes. This is safe security-wise but it might be better to not use pickle for this purpose and to investigate an alternative.

aymuos15:
## Summary
- Replace `algo_to_pickle()` with `algo_to_json()` using compact JSON + MONAI's existing `_target_` ConfigParser pattern for truly pickle-free algo metadata serialization
- Add `_make_json_serializable()` helper to handle numpy arrays, tensors, Path objects
- Backward compatible: `algo_from_json()` can still load legacy `.pkl` files (with deprecation warning)
- Keep `algo_to_pickle()` / `algo_from_pickle()` as deprecated aliases

**Note:** Model weights still use `torch.save` separately—this PR focuses on the algo object serialization.

## Test plan
- [x] Linting passes (`./runtests.sh --codeformat`, `./runtests.sh --ruff`)
- [x] Unit tests for `_make_json_serializable()` and `_add_path_with_parent()` helpers
- [ ] Integration test with `tests/apps/test_auto3dseg_bundlegen.py` (requires optional deps)

Fixes #8586


monai/auto3dseg/utils.py
```diff
@@ -274,48 +278,94 @@ def verify_report_format(report: dict, report_format: dict) -> bool:
...
+        "mlflow_tracking_uri",
+        "mlflow_experiment_name",
+        "output_path",
+        "name",
+        "best_metric",
+        "fill_records",
+        "device_setting",
+    ]:
+        if hasattr(algo, attr):
+            state[attr] = _make_json_serializable(getattr(algo, attr))
```
ericspod:
```suggestion
    attrs = [
        "template_path",
        "data_stats_files",
        "data_list_file",
        "mlflow_tracking_uri",
        "mlflow_experiment_name",
        "output_path",
        "name",
        "best_metric",
        "fill_records",
        "device_setting"
    ]
    state = {a : _make_json_serializable(getattr(algo, a)) for a in attrs if hasattr(algo, a)}
```

Moreso than this it might make sense to add a `state_dict` method to `Algo` which would return a dictionary of values for its state. These values could then be run through `_make_json_serializable`, you wouldn't have to hard code the `attrs` list of attributes in this function, and allows `Algo` to be extended and modified in the future.

monai/auto3dseg/utils.py
```diff
@@ -330,51 +380,145 @@ def algo_from_pickle(pkl_filename: str, template_path: PathLike | None = None, *
...
+            FutureWarning,
+            stacklevel=2,
+        )
+        return _load_legacy_pickle(filename, template_path)
+
+    with open(filename) as f:
+        data = json.load(f)
+
+    if not isinstance(data, dict):
+        raise ValueError(f"the data object is {data.__class__}. Dict is expected.")
```
ericspod:
```suggestion
        raise ValueError(f"The data object type is {type(data)}, type dict is expected.")
```

monai/auto3dseg/utils.py
```diff
@@ -274,48 +278,94 @@ def verify_report_format(report: dict, report_format: dict) -> bool:
...
+        "fill_records",
+        "device_setting",
+    ]:
+        if hasattr(algo, attr):
+            state[attr] = _make_json_serializable(getattr(algo, attr))
+
+    # Build target string for dynamic class instantiation
+    cls = algo.__class__
+    target = f"{cls.__module__}.{cls.__name__}"
+
```
ericspod, line 330:
```suggestion

algo_meta_data= {str(k): _make_json_serializable(v) for k,v in algo_meta_data.items()}
```
I think this addresses the coderabbit comment.

monai/auto3dseg/utils.py
```diff
@@ -274,48 +278,94 @@ def verify_report_format(report: dict, report_format: dict) -> bool:
...
-def algo_from_pickle(pkl_filename: str, template_path: PathLike | None = None, **kwargs: Any) -> Any:
+    if value is None:
+        return None
+    if isinstance(value, (str, int, float, bool)):
+        return value
+    if isinstance(value, (list, tuple)):
+        return [_make_json_serializable(v) for v in value]
+    if isinstance(value, dict):
+        return {k: _make_json_serializable(v) for k, v in value.items()}
+    if hasattr(value, "__fspath__"):  # Path-like objects
```
ericspod: I'm not sure it's necessary to check this as `str(value)` should yield the path string for such objects. Alternative is to check that `value` is an instance of `os.PathLike`.

monai/auto3dseg/utils.py
```diff
@@ -330,51 +380,145 @@ def algo_from_pickle(pkl_filename: str, template_path: PathLike | None = None, *
...
     template_paths_candidates: list[str] = []
-
-    if os.path.isdir(str(template_path)):
-        template_paths_candidates.append(os.path.abspath(str(template_path)))
-        template_paths_candidates.append(os.path.abspath(os.path.join(str(template_path), "..")))
-
-    if os.path.isdir(str(algo_template_path)):
-        template_paths_candidates.append(os.path.abspath(algo_template_path))
-        template_paths_candidates.append(os.path.abspath(os.path.join(algo_template_path, "..")))
+    _add_path_with_parent(template_paths_candidates, str(template_path) if template_path else None)
```
ericspod:
```suggestion
    _add_path_with_parent(template_paths_candidates, template_path)
```
This seems adequate since the value of `template_path` is checked in the called function.

monai/auto3dseg/utils.py
```diff
@@ -330,51 +380,145 @@ def algo_from_pickle(pkl_filename: str, template_path: PathLike | None = None, *
...
+    algo = None
+    used_template_path: str | None = None
+    for path in paths_to_try:
+        path_added = False
+        try:
+            if path and path not in sys.path:
+                sys.path.insert(0, path)
+                path_added = True
+
+            algo_config: dict[str, Any] = {"_target_": target}
```
ericspod, line 474: Using this config will create a new object only if `target` refers to a class with a default constructor needing no arguments. I suppose it works for `Algo` for now but might not always.

ericspod (changes requested):
Hi @aymuos15 thanks for the work on this, I think it's going in the right direction. I would like others who have worked on Auto3DSeg to have a look as well, and I have a number of comments below. We're working on a CI issue still but will come back to this one soon.

monai/auto3dseg/utils.py
aymuos15: Ah okay. I just removed the check, I think that is cleaner then?

monai/auto3dseg/utils.py
aymuos15:
```python
  class Algo:                                                                                         
      def state_dict(self) -> dict:                                                                   
          """Return state for serialization."""                                                       
          return {}                                                                                   
                                                                                                      
      def load_state_dict(self, state: dict) -> None:                                                 
          """Restore state from dict."""                                                              
          pass                                                                                        
                                                                                                      
  # Then BundleAlgo overrides:                                                                          
  class BundleAlgo(Algo):                                                                             
      def state_dict(self) -> dict:                                                                   
          return {                                                                                    
              "template_path": self.template_path,                                                    
              "output_path": self.output_path,                                                        
              "name": self.name,                                                                      
              # ... etc                                                                               
          }                                                                                           
                                                                                                      
      def load_state_dict(self, state: dict) -> None:                                                 
          for key, value in state.items():                                                            
              if hasattr(self, key):                                                                  
                  setattr(self, key, value)                                                           
```

Would this be a better approach then?

monai/auto3dseg/utils.py
ericspod: Yes I'd just remove it here.

monai/auto3dseg/utils.py
ericspod: I think so. It's more in line with how PyTorch would do things. So long as any subclass of `Algo` has a default constructor you can create new instances the way you're doing now then restore the state with `load_state_dict`.

monai/auto3dseg/utils.py
```diff
@@ -330,51 +379,145 @@ def algo_from_pickle(pkl_filename: str, template_path: PathLike | None = None, *
...
+            setattr(algo, attr, value)
+
+    algo.template_path = used_template_path
+
+    if file_dir != os.path.abspath(algo.get_output_path()):
+        logging.debug(f"{algo.get_output_path()} is changed. Now override the Algo output_path with: {file_dir}.")
+        algo.output_path = file_dir
+
+    return algo, dict(data)
+
```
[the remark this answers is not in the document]
aymuos15: @ericspod Is e2e testing like this recommend/needed?

ericspod:
Hi @daguangxu this is one proposed solution for removing pickle from Auto3DSeg.

aymuos15:
@ericspod 

I've changed it so people can choose to keep using pickle through the env var `MONAI_ALLOW_PICKLE=1`, but it's also behind a deprecation (`FutureWarning`). Should we remove the deprecation and keep pickle as a permanent opt-in instead?

Ive also addressed all the previous comments.

monai/auto3dseg/utils.py
```diff
@@ -274,48 +278,79 @@ def verify_report_format(report: dict, report_format: dict) -> bool:
...
 
-def algo_from_pickle(pkl_filename: str, template_path: PathLike | None = None, **kwargs: Any) -> Any:
-    """
-    Import the Algo object from a pickle file.
+    algo_meta_data = {str(k): _make_json_serializable(v) for k, v in algo_meta_data.items()}
+    data: dict[str, Any] = {
+        "_target_": target,
+        "_state_": state,
+        "template_path": str(template_path) if template_path else None,
+        **algo_meta_data,
```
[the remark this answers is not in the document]
aymuos15: It should break?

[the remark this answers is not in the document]
aymuos15: Thinks its okay as is

monai/auto3dseg/utils.py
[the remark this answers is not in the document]
ericspod: I think if tests cover the usage of these functions already they don't need separate testing as they aren't used in any other context.

monai/auto3dseg/utils.py
```diff
@@ -274,48 +293,80 @@ def verify_report_format(report: dict, report_format: dict) -> bool:
...
+    Handles numpy arrays, Path objects, torch tensors, and other common types.
+    """
+    if value is None:
+        return None
+    if isinstance(value, (str, int, float, bool)):
+        return value
+    if isinstance(value, (list, tuple)):
+        return [_make_json_serializable(v) for v in value]
+    if isinstance(value, dict):
+        return {k: _make_json_serializable(v) for k, v in value.items()}
```
ericspod:
```suggestion
        return {str(k): _make_json_serializable(v) for k, v in value.items()}
```
This might be needed given what Coderabbit says below.

monai/auto3dseg/utils.py
```diff
@@ -274,48 +293,80 @@ def verify_report_format(report: dict, report_format: dict) -> bool:
...
     """
-    data = {"algo_bytes": pickle.dumps(algo), "template_path": str(template_path)}
-    pkl_filename = os.path.join(algo.get_output_path(), "algo_object.pkl")
-    for k, v in algo_meta_data.items():
-        data.update({k: v})
-    data_bytes = pickle.dumps(data)
-    with open(pkl_filename, "wb") as f_pi:
-        f_pi.write(data_bytes)
-    return pkl_filename
+    state = {k: _make_json_serializable(v) for k, v in algo.state_dict().items()}
```
ericspod:
```suggestion
    state = {str(k): _make_json_serializable(v) for k, v in algo.state_dict().items()}
```

ericspod (changes requested):
I had some minor changes again.

monai/auto3dseg/utils.py
```diff
@@ -368,13 +407,109 @@ def algo_from_pickle(pkl_filename: str, template_path: PathLike | None = None, *
...
+            used_template_path = path
+
+            if path and path in sys.path:
+                sys.path.remove(path)
+            break
+        except (ModuleNotFoundError, ImportError) as e:
+            if path and path in sys.path:
+                sys.path.remove(path)
+            logging.debug(f"Failed to instantiate {target} with path {path}: {e}")
+            continue
```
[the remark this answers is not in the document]
ericspod: @aymuos15 this was in the previous implementation so we'll leave it here as well, one solution in the future is to check that the function is called by the main thread only to prevent multithreading issues. I don't see how that would occur in this instance, so nothing need's doing here.

ericspod (approved):
Looks good to me, this addresses the security concern but we've left in the ability to use pickle for those that need it. We'll leave the deprecation warnings too as they are appropriate.

PR #8696, opened 2026-01-13
Fix GEGLU docstring: Sigmoid -> GELU

aymuos15:
## Summary
- Fixed GEGLU docstring which incorrectly stated the activation function was Sigmoid
- The code correctly uses GELU, as specified in the original GEGLU paper

## Details
- GLU uses Sigmoid: GLU(x) = σ(xW) ⊗ xV
- GEGLU uses GELU: GEGLU(x) = GELU(xW) ⊗ xV

Reference: https://arxiv.org/abs/2002.05202

### DIFF
--- a/monai/networks/blocks/activation.py
+++ b/monai/networks/blocks/activation.py
@@ -177,6 +177,14 @@ class GEGLU(nn.Module):
     Shape:
         - Input: :math:`(N, *, 2 * D)`
         - Output: :math:`(N, *, D)`, where `*` means, any number of additional dimensions
+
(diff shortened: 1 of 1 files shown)


aymuos15:
"Optional: For consistency with other activation classes in this file, consider adding an Examples section." -- Happy to add this as well if required.

ericspod:
Thanks @aymuos15 for this fix, please do add an example to be consistent with the other classes.

ericspod (approved):

PR #8699, opened 2026-01-15
Updating data_config.json with HF links and hashes

### LINKED ISSUE #8510: Moving Testing Data to HuggingFace
I have copied out testing data from [this release](https://github.com/Project-MONAI/MONAI-extra-test-data/releases/tag/0.8.1) into a HuggingFace dataset [here](https://huggingface.co/datasets/MONAI/testing_data). I propose that we change our URLs in the tests to use these locations instead, and in the future we can more easily add items to this dataset as well as avoid abusing the release mechanism Github provides. Second to that is to move data for tutorials to another new dataset and clean up the use of multiple storage locations such as gdrive accounts. Last is to possibly do the same for model zoo items and weights unless we migrate entirely to HuggingFace.

e-mny:
Fixes #8510

### Description

The data_config.json needs to be updated with HF links in view of migrating to HF.

### Types of changes

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -183,32 +183,35 @@ def skip_if_downloading_fails():
 
         raise rt_e
 
-
-SAMPLE_TIFF = (
-    https://huggingface.co/datasets/MONAI/testing_data/resolve/main/CMU-1.tiff
-)
-
+SAMPLE_TIFF = "https://huggingface.co/datasets/MONAI/testing_data/resolve/main/CMU-1.tiff"
 SAMPLE_TIFF_HASH = "73a7e89bc15576587c3d68e55d9bf92f09690280166240b48ff4b48230b13bcd"
 SAMPLE_TIFF_HASH_TYPE = "sha256"
 
-def test_download_url():
-    with tempfile.TemporaryDirectory() as tempdir:
-        # successful download with correct hash
-        download_url(
-            url=SAMPLE_TIFF,
-            filepath=os.path.join(tempdir, "model.tiff"),
-            hash_val=SAMPLE_TIFF_HASH,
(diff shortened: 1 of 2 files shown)


ericspod:
Hi @e-mny thanks for the help on this one! I think we're in the right direction, however the hash values mentioned in the file are for the file contents themselves and not the repository hashes on Huggingface. These shouldn't change then since the file contents shouldn't be different once downloaded. I think there isn't validation going on since the tests currently pass, specifically I think [here](https://github.com/Project-MONAI/MONAI/blob/57fdd594ac905ab2ea778aa4bb79ccd9a0a03b22/tests/test_utils.py#L796) and elsewhere the failing hash checks are being ignored (they produce an exception which gets tossed out by `skip_if_downloading_fails`). 

I would suggest changing the hashes back to their original values in the file but keep the new Huggingface URLs that you've got, and then check the files can be downloaded and hash checked correctly. This may need some digging around in the code to see how the file contents are used, the file is loaded in [testing_data_config](https://github.com/Project-MONAI/MONAI/blob/57fdd594ac905ab2ea778aa4bb79ccd9a0a03b22/tests/test_utils.py#L86) for example which is called in various places. You may want to add a test for hash download checking that uses this function and checks that `download_url` either fails entirely to download the file or successfully downloads then correctly validates the hash value against the file. 

One thing we may want to change later is to raise a different exception when the hash fails [here](https://github.com/Project-MONAI/MONAI/blob/57fdd594ac905ab2ea778aa4bb79ccd9a0a03b22/monai/apps/utils.py#L271). It should maybe raise something specific to this and so isn't suppressed by `skip_if_download_fails`. Let's not worry about that in this PR yet however.

PR #8704, opened 2026-01-16
Adjust execution order of activation and masking in MaskedDiceLoss

### LINKED ISSUE #8655: Potentially faulty usage of sigmoid argument in Masked Dice
Hi there! 

When using MaskedDice, my UNET doesn't train as well as when using the normal DICE. Upon closer inspection, I have noticed the following, which I believe to be an inconsistency in the usage of the "sigmoid" argument when initializing the loss (_please correct me if I'm wrong :)_).

The DICE loss, and thus also MaskedDiceloss, accept the argument "sigmoid=True", allowing the user to pass in model output _logits_ as opposed to model output _probabilities_.

When using DICE loss, the sigmoid is handled correctly. However, when using the MaskedDice, the sigmoid is applied **after** masking the input and target values. 

If I am reading the code correctly, MaskedDice applies the mask to the model outputs and targets by setting all to be masked values equal to zero, and then passes on these _masked_ model outputs and targets to the regular DICE loss. The sigmoid will then be regularly applied to the masked model outputs in the normal DICE loss's forward function, if the argument "sigmoid" has been set to "True" during initialization.

Notably, all model output values set to zero during masking **become 0.5 when forwarded through the sigmoid** in the normal DICE forward function. This is, if I understand correctly, not the behavior a user would expect.

To fix this, I'd suggest setting all values in the model output to _-inf_ when masking whenever the argument "sigmoid" is set to "True".  Or alternatively, apply the sigmoid in the forward function of the MaskedDice, and initialize the normal DICE with sigmoid=False.

I hope my explanation was understandable – maybe I am also misreading the code. Let me know. :) 

Kind regards!

ytl0623:
Fixes #8655

### Description

1. Modified `MaskedDiceLoss.forward` to ensure masked regions result in 0.0 probability instead of 0.5.
2. Updated unit tests with new expected values, as the fix mathematically changes the loss result by removing the erroneous background contribution.

### Types of changes

### DIFF
--- a/monai/losses/dice.py
+++ b/monai/losses/dice.py
@@ -13,7 +13,6 @@ from __future__ import annotations
 
 import warnings
 from collections.abc import Callable, Sequence
-from typing import Any
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


ericspod:
Hi @ytl0623 thanks for this fix, but I don't feel this is the right solution for it. The issue here is that the activation needs to happen before masking however the approach here creates a new `DiceLoss` object and so the members of the class inherited from `DiceLoss` aren't actually used. This isn't correct object-oriented practice and not consistent with other areas in MONAI. I would suggest that we need to keep new variables such as `self.pre_sigmoid` to track the original argument values passed in, but then pass along False to the `super().__init__` call that's been removed. I'd also suggest replacing `*args,**kwargs` in the constructor with the actual needed named arguments. The constructor would be something like this:

```python
def __init__(
        self,
        to_onehot_y: bool = False,
        sigmoid: bool = False,
        softmax: bool = False,
        other_act: Callable | None = None,
        squared_pred: bool = False,
        jaccard: bool = False,
        reduction: LossReduction | str = LossReduction.MEAN,
        smooth_nr: float = 1e-5,
        smooth_dr: float = 1e-5,
        batch: bool = False,
        weight: Sequence[float] | float | int | torch.Tensor | None = None,
        soft_label: bool = False,
    ) -> None:
        """
        Args follow :py:class:`monai.losses.DiceLoss`.
        """
        self.pre_sigmoid=sigmoid
        self.pre_softmax=softmax
        self.pre_other_act=other_act
        super().__init__(
            include_background=include_background,
            to_onehot_y=to_onehot_y,
            sigmoid=False,
            softmax=False,
            other_act=None,
            squared_pred=squared_pred,
            jaccard=jaccard,
            reduction=reduction,
            smooth_nr=smooth_nr,
            smooth_dr=smooth_dr,
            batch=batch,
            weight=class_weight,
            soft_label=soft_label
        )
        self.spatial_weighted = MaskedLoss(loss=super().forward)
```

Your `forward` method would be broadly the same though you'd change the names of the members you'd use to the `pre_` ones. This allows the inherited `DiceLoss` members to be used and will be compatible with future inheritance of this class if it happens. We're changing the constructor arguments to be what they should be for `DiceLoss`, I don't consider this a deprecating change so no need to use the `deprecated` decorator.

ytl0623:
Hi @ericspod, thanks for the guidance! Could you check the updated PR? Appreciate your help!

ericspod (approved):

PR #8705, opened 2026-01-16
Improve CuPy type hinting in ImageReader

ytl0623:
Related to #8188 

### Description

This PR addresses the [comment](https://github.com/Project-MONAI/MONAI/blob/57fdd594ac905ab2ea778aa4bb79ccd9a0a03b22/monai/data/image_reader.py#L666) in `monai/data/image_reader.py` regarding the correct type annotation for image arrays that can be either `np.ndarray` or `cp.ndarray`.

Previously, the type hint was restricted to `np.ndarray` to avoid ImportError for users without CuPy installed. This PR introduces a robust Ndarray type alias that supports static type checking for `cupy.ndarray` while maintaining runtime compatibility for CPU-only environments.

### Types of changes

### DIFF
--- a/monai/data/image_reader.py
+++ b/monai/data/image_reader.py
@@ -22,7 +22,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable, Iterator, Sequence
 from dataclasses import dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Union
 
 import numpy as np
 from torch.utils.data._utils.collate import np_str_obj_array_pattern
@@ -38,6 +38,7 @@ from monai.data.utils import (
 from monai.utils import MetaKeys, SpaceKeys, TraceKeys, ensure_tuple, optional_import, require_pkg
 
 if TYPE_CHECKING:
+    import cupy as cp
     import itk
     import nibabel as nib
     import nrrd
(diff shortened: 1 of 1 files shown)


monai/data/image_reader.py
```diff
@@ -841,7 +866,7 @@ def _get_seg_data(self, img, filename):
         if self.label_dict is not None:
             metadata["labels"] = self.label_dict
             if self.to_gpu:
-                all_segs = cp.zeros([*spatial_shape, len(self.label_dict)], dtype=array_data.dtype)
+                all_segs: Ndarray = cp.zeros([*spatial_shape, len(self.label_dict)], dtype=array_data.dtype)
```
ericspod: I'm not sure this is needed, the type could just be `cp.ndarray`. The definition of `all_segs` in the else condition should have a type as well.

ericspod (approved):
Thanks for these improvements, I had just a minor comment. We should later update typing in MONAI to consolidate this sort of optional type mechanism into one place as there's other areas with similar issues, but that would be for later work.

monai/data/image_reader.py
ytl0623: Done. I changed the type to `cp.ndarray` and added the `np.ndarray` type definition to the else condition.

ytl0623:
Hi @ericspod, Thanks for commenting.
Agreed. Maybe we can use a `Union` type definition (like `Union[np.ndarray, cp.ndarray]`)  to solve this.

monai/data/image_reader.py
ytl0623: You are right, we don't need this, Because `cp.zeros` and `np.zeros`  will already return objects with type.

PR #8710, opened 2026-01-19
Make y optional in ignore_background function

ytl0623:
### Description

This PR addresses a [TODO](https://github.com/Project-MONAI/MONAI/blob/57fdd594ac905ab2ea778aa4bb79ccd9a0a03b22/monai/metrics/active_learning_metrics.py#L133)  by refactoring the `ignore_background` utility function.

### Types of changes

### DIFF
--- a/monai/metrics/active_learning_metrics.py
+++ b/monai/metrics/active_learning_metrics.py
@@ -129,9 +129,7 @@ def compute_variance(
     y_pred = y_pred.float()
 
     if not include_background:
-        y = y_pred
-        # TODO If this utils is made to be optional for 'y' it would be nice
-        y_pred, y = ignore_background(y_pred=y_pred, y=y)
+        y_pred = ignore_background(y_pred=y_pred)
 
     # Set any values below 0 to threshold
     y_pred[y_pred <= 0] = threshold
--- a/monai/metrics/utils.py
+++ b/monai/metrics/utils.py
@@ -15,7 +15,7 @@ import warnings
 from collections.abc import Iterable, Sequence
 from functools import cache, partial
 from types import ModuleType
-from typing import Any
+from typing import Any, overload
 
 import numpy as np
 import torch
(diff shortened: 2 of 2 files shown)


monai/metrics/active_learning_metrics.py
```diff
@@ -130,7 +130,6 @@ def compute_variance(
 
     if not include_background:
         y = y_pred
-        # TODO If this utils is made to be optional for 'y' it would be nice
         y_pred, y = ignore_background(y_pred=y_pred, y=y)
```
ericspod:
```suggestion
        y_pred = ignore_background(y_pred=y_pred)
```
You can remove the `y` parameter now that it's optional (and remove the previous line).

monai/metrics/utils.py
```diff
@@ -51,20 +51,35 @@
...
+        y: ground truth, the first dim is batch. (Optional)
+
+    Returns:
+        Tuple of background-removed predictions and ground truth. `y` is None if not provided.
 
     """
 
-    y = y[:, 1:] if y.shape[1] > 1 else y  # type: ignore[assignment]
+    if y is not None:
+        y = y[:, 1:] if y.shape[1] > 1 else y  # type: ignore[assignment]
```
ericspod: Can we get rid of the "type: ignore[assignment]" tags now with the override signature definition?

ericspod (approved):
Hi @ytl0623 thanks for this fix, I think it's fine as a minor change with a couple of comments.

monai/metrics/utils.py
ytl0623: I tried removing them, but the CI fails. The issue stems from `mypy`'s strict inference on slicing `Union` types within the implementation, which the overloads don't address. So, we have to keep the tags.

ytl0623: Should I switch to `typing.cast` to handle the type narrowing explicitly instead of using the ignore tag?

ericspod: Let's keep the tags then for now, possibly mypy will allow removing them in the future.

PR #8715, opened 2026-01-22
Generalize TestTimeAugmentation to non-spatial predictions

### LINKED ISSUE #8276: Generalize `TestTimeAugmentation` to non-spatial predictions
Currently `monai.data.TestTimeAugmentation` seemingly [assumes the inverse transformations can be applied to the predictions](https://github.com/Project-MONAI/MONAI/blob/46a5272196a6c2590ca2589029eed8e4d56ff008/monai/data/test_time_augmentation.py#L199), which necessarily means the predictions must be spatial.

Seems fairly straightforward - would be willing to write up a PR if there's interest and no reason this doesn't make sense.

ytl0623:
Fixes #8276

### Description

- Added a new argument `apply_inverse_to_pred`. Defaults to `True` to preserve backward compatibility. When set to `False`, it skips the inverse transformation step and aggregates the model predictions directly.
- Added a new unit test to simulate a classification task with spatial augmentation, verifying that the aggregation works correctly without spatial inversion.

### Types of changes

### DIFF
--- a/monai/data/test_time_augmentation.py
+++ b/monai/data/test_time_augmentation.py
@@ -67,7 +67,7 @@ class TestTimeAugmentation:
     Args:
         transform: transform (or composed) to be applied to each realization. At least one transform must be of type
         `RandomizableTrait` (i.e. `Randomizable`, `RandomizableTransform`, or `RandomizableTrait`).
-        All random transforms must be of type `InvertibleTransform`.
+        When `apply_inverse_to_pred` is True, all random transforms must be of type `InvertibleTransform`.
         batch_size: number of realizations to infer at once.
         num_workers: how many subprocesses to use for data.
         inferrer_fn: function to use to perform inference.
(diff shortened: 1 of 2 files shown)


ericspod:
hi @ytl0623 thanks for this change, I think it's fine in principle. The `_check_transforms` method should be changed to account for when the new argument is `False`, in which case it doesn't need to check for invertibility of transforms. I noticed other issues with the original version of this method so I'd propose something like the following (which I haven't tested):

```python
def _check_transforms(self):
    """Should be at least 1 random transform, and all random transforms should be invertible."""
    transforms = [self.transform] if not isinstance(self.transform, Compose) else self.transform.transforms
    warns=[]
    randoms=[]
    for idx, t in transforms:
        if isinstance(t, Randomizable):
            randoms.append(t)
            if self.apply_inverse_to_pred and not isinstance(t, InvertibleTransform):
                warns.append(f"Transform {idx} (type {type(t).__name__}) not invertible.")
                
    if len(randoms)==0:
        warns.append("TTA usually requires at least one `Randomizable` transform in the given transform sequence.")
        
    if len(warns)>0:
        warnings.warn("TTA has encountered issues with the given transforms:"+"\n  ".join(warns))
```

Please check this logic, it might be that we need to check all transforms for invertibility whether they're random or not, but what I have here is equivalent to the original.

ytl0623:
Hi @ericspod, thanks for the suggestion!

ericspod (approved):
Looks good to me, thanks!

PR #8719, opened 2026-01-25
4609: Add AUC-Margin Loss for AUROC optimization

### LINKED ISSUE #4609: Feature request: `AUCMLoss`
None

[Large-scale Robust Deep AUC Maximization: A New Surrogate Loss and Empirical Studies on Medical Image Classification](https://arxiv.org/abs/2012.03173)

Could you please implement this loss function?
I couldn't do it myself due to my lack of skills, and I thought it would be nice if it could be used in monai as it is related to medical image classification.

None

**Additional context**
None

shubham-61969:
Fixes #4609.

### Description

This PR adds an implementation of **AUC-Margin Loss (AUCM)** for direct AUROC optimization in MONAI, based on:

> [Yuan et al., *“Large-scale Robust Deep AUC Maximization: A New Surrogate Loss and Empirical Studies on Medical Image Classification”*, ICCV 2021.](https://arxiv.org/abs/2012.03173)

Implementation based on:  
[https://github.com/Optimization-AI/LibAUC/blob/1.4.0/libauc/losses/auc.py](https://github.com/Optimization-AI/LibAUC/blob/1.4.0/libauc/losses/auc.py)

The loss is designed for imbalanced classification problems, which are common in medical imaging, where AUROC is often the primary evaluation metric. The implementation follows MONAI’s loss conventions, is fully PyTorch-native, and does not introduce any new dependencies.

### Types of changes

### DIFF
--- a/monai/losses/aucm_loss.py
+++ b/monai/losses/aucm_loss.py
@@ -119,32 +119,83 @@ class AUCMLoss(_Loss):
         pos_mask = (target == 1).float()
         neg_mask = (target == 0).float()
 
+        mean_pos_sq = (input - self.a) ** 2
+        mean_neg_sq = (input - self.b) ** 2
+
+        # Note:
+        # v1 uses global expectations (normalized by total number of samples),
+        # following the original LibAUC implementation.
+        # v2 uses class-conditional expectations (normalized by number of samples
+        # in each class), implemented via non-zero averaging.
+        # These behaviors differ and should not be unified.
         if self.version == "v1":
             p = float(self.imratio) if self.imratio is not None else float(pos_mask.mean().item())
+            p1 = 1.0 - p
+
+            mean_pos = self._global_mean(mean_pos_sq, pos_mask)
+            mean_neg = self._global_mean(mean_neg_sq, neg_mask)
+
+            interaction = self._global_mean(p * input * neg_mask - p1 * input * pos_mask, pos_mask + neg_mask)
+
             loss = (
-                (1 - p) * self._safe_mean((input - self.a) ** 2, pos_mask)
-                + p * self._safe_mean((input - self.b) ** 2, neg_mask)
-                + 2
-                * self.alpha
-                * (
-                    p * (1 - p) * self.margin
-                    + self._safe_mean(p * input * neg_mask - (1 - p) * input * pos_mask, pos_mask + neg_mask)
-                )
-                - p * (1 - p) * self.alpha**2
+                p1 * mean_pos
+                + p * mean_neg
+                + 2 * self.alpha * (p * p1 * self.margin + interaction)
+                - p * p1 * self.alpha**2
             )
-        else:
+
+        else:  # v2
+            mean_pos = self._class_mean(mean_pos_sq, pos_mask)
+            mean_neg = self._class_mean(mean_neg_sq, neg_mask)
+
+            mean_input_pos = self._class_mean(input, pos_mask)
+            mean_input_neg = self._class_mean(input, neg_mask)
+
             loss = (
-                self._safe_mean((input - self.a) ** 2, pos_mask)
-                + self._safe_mean((input - self.b) ** 2, neg_mask)
-                + 2 * self.alpha * (self.margin + self._safe_mean(input, neg_mask) - self._safe_mean(input, pos_mask))
-                - self.alpha**2
+                mean_pos + mean_neg + 2 * self.alpha * (self.margin + mean_input_neg - mean_input_pos) - self.alpha**2
             )
 
         return loss
 
-    def _safe_mean(self, tensor: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
-        """Compute mean safely over masked elements."""
+    def _global_mean(self, tensor: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
+        """
+        Compute the global mean of a masked tensor.
+
+        This computes the mean over all elements, where values outside the mask
+        are zeroed out. The result is normalized by the total number of elements,
+        not by the number of masked elements.
+
+        This corresponds to a global expectation:
+            E[mask * tensor]
+
+        Args:
+            tensor: Input tensor.
+            mask: Binary mask tensor of the same shape as ``tensor``.
+
+        Returns:
+            Scalar tensor representing the global mean.
+        """
+        return (tensor * mask).mean()
+
+    def _class_mean(self, tensor: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
+        """
+        Compute the class-conditional mean of a masked tensor.
+
+        This computes the mean over only the masked (non-zero) elements, i.e.,
+        the result is normalized by the number of masked elements.
+
+        This corresponds to a class-conditional expectation:
+            E[tensor | mask]
+
+        Args:
+            tensor: Input tensor.
+            mask: Binary mask tensor of the same shape as ``tensor``.
+
+        Returns:
+            Scalar tensor representing the class-conditional mean.
+            Returns 0 if no elements are selected by the mask.
+        """
         denom = mask.sum()
-        if denom == 0:
-            return torch.tensor(0.0, device=tensor.device, dtype=tensor.dtype)
+        if denom.item() == 0:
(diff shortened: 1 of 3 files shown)


shubham-61969:
Hi @KumoLiu , thanks for triggering the build.
I see the vulnerability report refers to an existing issue in MONAI unrelated to this PR.

Are there any changes needed from my side for this PR?

KumoLiu:
> Are there any changes needed from my side for this PR?

Hi @shubham-61969, no need, we are releasing a minor version to fix the issue.

monai/losses/aucm_loss.py
```diff
@@ -0,0 +1,150 @@
...
+                )
+                - p * (1 - p) * self.alpha**2
+            )
+        else:
+            loss = (
+                self._safe_mean((input - self.a) ** 2, pos_mask)
+                + self._safe_mean((input - self.b) ** 2, neg_mask)
+                + 2 * self.alpha * (self.margin + self._safe_mean(input, neg_mask) - self._safe_mean(input, pos_mask))
+                - self.alpha**2
+            )
```
ericspod, line 159:
```suggestion
        if self.version == "v1":
            p = float(self.imratio if self.imratio is not None else pos_mask.mean().item())
            p1 = 1 - p
            safe_mean_pos_neg = self._safe_mean(p * input * neg_mask - p1 * input * pos_mask, pos_mask + neg_mask)
        else:
            p = p1 = 1.0  # positive sample ratio discounted in this version by setting coefficients to 1
            safe_mean_pos_neg = self._safe_mean(input, neg_mask) - self._safe_mean(input, pos_mask)

        mean_a = p1 * self._safe_mean((input - self.a) ** 2, pos_mask)
        mean_b = p * self._safe_mean((input - self.b) ** 2, neg_mask)
        loss = mean_a + mean_b + 2 * self.alpha * (p * p1 * self.margin + safe_mean_pos_neg) - p * p1 * self.alpha**2
```

If I haven't messed up the refactor of the loss equations, this would help simplify the code and reduce redundant calculations. With a bit more commentary this may be easier to understand.

tests/losses/test_aucm_loss.py
```diff
@@ -0,0 +1,111 @@
...
+        self.assertEqual(loss.ndim, 0)
+
+    def test_v2(self):
+        """Test AUCMLoss with version 'v2'."""
+        loss_fn = AUCMLoss(version="v2")
+        input = torch.randn(32, 1, requires_grad=True)
+        target = torch.randint(0, 2, (32, 1)).float()
+        loss = loss_fn(input, target)
+        self.assertIsInstance(loss, torch.Tensor)
+        self.assertEqual(loss.ndim, 0)
```
ericspod: These tests are a good start but I think we need a few which tests the output values themselves to ensure the calculation is what it should be. You can precompute some values and store them as globals at the top of this file, look at other test files to see how this is done with `parameterized`.

tests/losses/test_aucm_loss.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def test_invalid_version(self):
```
ericspod, line 74: The following tests are good too but could be condensed using `parameterized` instead if you wanted to do that.

ericspod (commented):
Hi @shubham-61969 thanks for this submission. I had some minor comments on some things, specifically I had refactored the loss calculation to reduce space and hopefully make it more understandable. I haven't tested it so please double check what I've done, if you prefer your explicit version we can leave that in. Any more commentary to help explain the loss wouldn't hurt either in the code. I had minor comments on the tests as well.

shubham-61969:
Hi @ericspod , thanks for the detailed feedback and the suggested refactor! Also, apologies for the delayed response on this PR.

I carefully review the formulation against both the paper and the original LibAUC implementation. One key point I found is that v1 and v2 use different types of expectations:

- v1 uses global expectations (normalized by total number of samples)
- v2 uses class-conditional expectations (normalized by the number of samples in each class)

This distinction is present both in the paper formulation and in the LibAUC reference implementation. Because of this, fully unifying the expressions for v1 and v2 would change the intended behavior of the loss. I’ve therefore kept the two paths semantically distinct while refactoring the code to improve readability and reduce repetition.

I also added comments in the implementation to clarify this difference so future readers don’t accidentally simplify it in a way that alters the behavior.

On the testing side:
- I refactored the version tests using `parameterized` to reduce duplication.
- I added deterministic test cases with manually verified expected values to ensure correctness.

At the moment, I’ve included two deterministic cases (one for v1 and one for v2) using small inputs where the expected values can be derived reliably. For more complex or higher-dimensional inputs, computing exact expected values becomes less straightforward.

Would you recommend adding more deterministic cases similar to other losses, or is this level of coverage sufficient given the complexity of the formulation?

tests/losses/test_aucm_loss.py
```diff
@@ -0,0 +1,142 @@
...
+import unittest
+
+import numpy as np
+import torch
+from parameterized import parameterized
+
+from monai.losses import AUCMLoss
+from tests.test_utils import test_script_save
+
+TEST_CASES = [
```
ericspod, line 23: I think these are in the right direction but there are other cases, such as totally blank inputs for both versions. Tests should also cover a range of constructor argument value combinations, so different positive tests for values of `imratio` or `margin`. You can look at [how dice is tested](https://github.com/Project-MONAI/MONAI/blob/dev/tests/losses/test_dice_loss.py) or other files to get an idea of how to do this, you can pre-compute results from your or other implementations of this loss to store here.

tests/losses/test_aucm_loss.py
```diff
@@ -0,0 +1,142 @@
...
+
+        pred = torch.randn(32, 1)
+        target = torch.zeros(32, 1)
+        target[0] = 1.0  # only one positive
+
+        loss = loss_fn(pred, target)
+
+        self.assertIsInstance(loss, torch.Tensor)
+
+    def test_invalid_version(self):
```
ericspod, line 74: With these tests what I mean to say with `parameterized` is to have a test like:

```python
@parameterized(BAD_ARGS):
def test_bad_args(self, kwargs):
        AUCMLoss(**kwargs)
```

The `BAD_ARGS` list would contain dictionaries for all the bad combinations of arguments you want to check raise an exception.

ericspod (commented):
Hi @shubham-61969 I had some comments regarding the tests but I'm fine with the refactored formulation of the loss itself. I think we should improve the test coverage to be a bit more robust but then we're good to go, thanks.

shubham-61969:
Hi @ericspod, thanks again for the earlier feedback and suggestions on improving the tests.

I’ve now updated the AUCM loss tests to add broader deterministic coverage and refactored them into parameterized test cases similar to other MONAI loss tests. The updated tests now include:

* deterministic forward-value checks for both `v1` and `v2`
* explicit `imratio` coverage
* explicit `margin` coverage
* combined `imratio + margin` cases
* higher-dimensional tensor inputs
* invalid argument and invalid shape tests
* backward pass checks
* reduction argument coverage
* edge-case tests for degenerate/blank inputs

For the deterministic forward tests, I used precomputed expected values with fixed internal optimization variables (`a`, `b`, `alpha`) to make margin-dependent behavior deterministic and testable.

One additional note regarding the implementation semantics:

* `v1` uses global expectations normalized over all samples
* `v2` uses class-conditional expectations

While validating the tests against the original LibAUC implementation, I noticed that the LibAUC `v2` implementation uses nested masked means internally, which leads to slightly different numerical behavior compared to the more explicit class-conditional expectation computation used here. The current tests therefore validate the MONAI implementation behavior directly rather than enforcing exact LibAUC numerical parity for `v2`.

I also observed that for certain degenerate inputs (e.g., blank predictions with mixed targets), `v2` can produce `NaN` values because of the class-conditional reduction behavior, whereas `v1` remains finite due to its global expectation formulation. I added edge-case coverage for this behavior without enforcing strict `NaN` equality semantics in the deterministic forward-value tests.

Could you please take another look at the updated tests and let me know if this level of coverage and the current testing approach look reasonable?

ericspod (approved):
Hi @shubham-61969 this looks good to me now. It's up to you if you want to match libAUC outputs exactly, we can always revise this class later.

PR #8725, opened 2026-01-30
Update coverage configuration to use `monai` as source

Borda:
### Description

This pull request makes a minor configuration update to the test coverage settings. The change restricts coverage measurement to only the `monai` package, rather than the entire source tree.

- Coverage configuration: Changed the `source` setting in `setup.cfg` from `.` (current directory) to `monai`, so coverage reports now only include files within the `monai` package.

This addresses misleading information about coverage since it combines the package and tests, see: https://app.codecov.io/gh/Project-MONAI/MONAI

image

### Types of changes

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -278,7 +278,7 @@ disallow_incomplete_defs = True
 
 [coverage:run]
 concurrency = multiprocessing
-source = .
+source = monai
 data_file = .coverage/.coverage
 omit = setup.py
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -280,7 +280,6 @@ disallow_incomplete_defs = True
 concurrency = multiprocessing
 source = monai
 data_file = .coverage/.coverage
-omit = setup.py
 
 [coverage:report]
 exclude_lines =


Borda:
@KumoLiu @ericspod friendly ping... 🦩

ericspod:
Hi @Borda thanks for the suggested change. I feel though that leaving tests in as something that's being covered is useful to identify when tests are missed accidentally, or when code becomes dead. We should have near perfect coverage in that directory so it does bias the results as an overall percentage, but we would inspect the `monai` results to see where coverage was poor manually anyway.

Borda:
> Hi Borda thanks for the suggested change. I feel though that leaving tests in as something that's being covered is useful to identify when tests are missed accidentally, or when code becomes dead. We should have near perfect coverage in that directory so it does bias the results as an overall percentage, but we would inspect the `monai` results to see where coverage was poor manually anyway.

tbh, not sure what your judgment is, so you want to keep the test folder in the report? Note that test running is the way the coverage is generated, which is why/how I found this dead code in https://github.com/Project-MONAI/MONAI/pull/8724

ericspod:
> > Hi Borda thanks for the suggested change. I feel though that leaving tests in as something that's being covered is useful to identify when tests are missed accidentally, or when code becomes dead. We should have near perfect coverage in that directory so it does bias the results as an overall percentage, but we would inspect the `monai` results to see where coverage was poor manually anyway.
> 
> tbh, not sure what your judgment is, so you want to keep the test folder in the report? Note that test running is the way the coverage is generated, which is why/how I found this dead code in #8724

I would leave the test coverage in the report, yes. It did reveal that the code you flagged wasn't being used so I think that is useful.

PR #8732, opened 2026-02-05
Fix using a non-tuple sequence for multidimensional indexing that wil…

aarpon:
Fix using a non-tuple sequence for multidimensional indexing that will result in error in pytorch 2.9.

Fixes # .

### Description

The `SlidingWindowInferer` complains that from PyTorch 2.9, using non-tuple sequences to slice into Tensors will result in an error (or, worse, misbehave). This patch targets all places in `monai/inferers/utils.py` where indexing may happen with lists or other types of sequences.

### Types of changes

### DIFF
--- a/monai/inferers/utils.py
+++ b/monai/inferers/utils.py
@@ -228,8 +228,10 @@ def sliding_window_inference(
             ).to(sw_device)
             if condition is not None:
                 win_condition = torch.cat(
-                    [condition[tuple(win_slice) if isinstance(win_slice, list) else win_slice] for win_slice in
-                     unravel_slice]
+                    [
+                        condition[tuple(win_slice) if isinstance(win_slice, list) else win_slice]
+                        for win_slice in unravel_slice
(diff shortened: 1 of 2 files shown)


ericspod:
Hi @aarpon thanks for these fixes, I know tuple indices are needed in places more strictly than in the past. Would it make sense to use `ensure_tuple` in these places however? I don't think we should be doing the same `tuple(p) if isinstance(p, list) else p` everywhere, this should be condensed to a function but I think `ensure_tuple` does this already. This also covers cases where `p` isn't a list but some other iterable that should be converted to a tuple. If you agree please update and we can rereview.

aarpon:
Hi, I replaced my explicit casts with calls to monai.utils.ensure_tuple and I re-ran the inferer tests with PyTorch 2.10 forcing errors for warnings.

ericspod (approved):
I think this looks good now, thanks!

PR #8741, opened 2026-02-14
Add missing channel_wise parameter to RandScaleIntensityFixedMean

### LINKED ISSUE #8363: channel_wise option missing in transforms.RandScaleIntensityFixedMean
Although official docs and the __init__'s comments state channel_wise as available args for transforms.RandScaleIntensityFixedMean,
it is currently not available.
Is it still in progress? or just missing by mistake?

```
class RandScaleIntensityFixedMean(RandomizableTransform):
    """
    Randomly scale the intensity of input image by ``v = v * (1 + factor)`` where the `factor`
    is randomly picked. Subtract the mean intensity before scaling with `factor`, then add the same value after scaling
    to ensure that the output has the same mean as the input.
    """

    backend = ScaleIntensityFixedMean.backend

    def __init__(
        self,
        prob: float = 0.1,
        factors: Sequence[float] | float = 0,
        fixed_mean: bool = True,
        preserve_range: bool = False,
        dtype: DtypeLike = np.float32,
    ) -> None:
        """
        Args:
            factors: factor range to randomly scale by ``v = v * (1 + factor)``.
                if single number, factor value is picked from (-factors, factors).
            preserve_range: clips the output array/tensor to the range of the input array/tensor
            fixed_mean: subtract the mean intensity before scaling with `factor`, then add the same value after scaling
                to ensure that the output has the same mean as the input.
            channel_wise: if True, scale on each channel separately. `preserve_range` and `fixed_mean` are also applied
            on each channel separately if `channel_wise` is True. Please ensure that the first dimension represents the
            channel of the image if True.
            dtype: output data type, if None, same as input image. defaults to float32.

        """
```

engmohamedsalah:
## Summary

- Adds the `channel_wise` parameter to `RandScaleIntensityFixedMean` and `RandScaleIntensityFixedMeand`, which was documented in docstrings but never implemented.
- When `channel_wise=True`, a separate random scale factor is generated per channel, and `preserve_range`/`fixed_mean` are applied per channel — following the existing pattern from `RandScaleIntensity`.
- Fixes docstring indentation for `channel_wise` in both array and dictionary transforms.

Fixes #8363

### `monai/transforms/intensity/array.py`
- Added `channel_wise` parameter to `RandScaleIntensityFixedMean.__init__`
- Updated `randomize()` to generate per-channel factors when `channel_wise=True`
- Updated `__call__` to apply per-channel scaling with individual random factors

### `monai/transforms/intensity/dictionary.py`
- Added `channel_wise` parameter to `RandScaleIntensityFixedMeand.__init__`
- Updated `__call__` to pass image data to `randomize()` (needed for channel count), following the pattern from `RandScaleIntensityd`

### Tests
- Added `test_channel_wise` and `test_channel_wise_preserve_range` to `test_rand_scale_intensity_fixed_mean.py`
- Added `test_channel_wise` to `test_rand_scale_intensity_fixed_meand.py`

## Test plan

- [x] All existing tests pass (no regressions)
- [x] New `channel_wise` tests pass for both array and dictionary transforms
- [x] Verified per-channel mean preservation with `fixed_mean=True`
- [x] Verified `preserve_range` clipping works per channel

### DIFF
--- a/monai/transforms/intensity/array.py
+++ b/monai/transforms/intensity/array.py
@@ -601,6 +601,7 @@ class RandScaleIntensityFixedMean(RandomizableTransform):
         factors: Sequence[float] | float = 0,
         fixed_mean: bool = True,
         preserve_range: bool = False,
+        channel_wise: bool = False,
         dtype: DtypeLike = np.float32,
     ) -> None:
         """
@@ -611,8 +612,8 @@ class RandScaleIntensityFixedMean(RandomizableTransform):
             fixed_mean: subtract the mean intensity before scaling with `factor`, then add the same value after scaling
                 to ensure that the output has the same mean as the input.
             channel_wise: if True, scale on each channel separately. `preserve_range` and `fixed_mean` are also applied
-            on each channel separately if `channel_wise` is True. Please ensure that the first dimension represents the
-            channel of the image if True.
+                on each channel separately if `channel_wise` is True. Please ensure that the first dimension represents the
+                channel of the image if True.
             dtype: output data type, if None, same as input image. defaults to float32.
 
         """
@@ -626,17 +627,25 @@ class RandScaleIntensityFixedMean(RandomizableTransform):
         self.factor = self.factors[0]
         self.fixed_mean = fixed_mean
         self.preserve_range = preserve_range
+        self.channel_wise = channel_wise
         self.dtype = dtype
 
         self.scaler = ScaleIntensityFixedMean(
-            factor=self.factor, fixed_mean=self.fixed_mean, preserve_range=self.preserve_range, dtype=self.dtype
+            factor=self.factor,
+            fixed_mean=self.fixed_mean,
+            preserve_range=self.preserve_range,
+            channel_wise=self.channel_wise,
+            dtype=self.dtype,
         )
 
     def randomize(self, data: Any | None = None) -> None:
         super().randomize(None)
         if not self._do_transform:
             return None
-        self.factor = self.R.uniform(low=self.factors[0], high=self.factors[1])
+        if self.channel_wise:
+            self.factor = [self.R.uniform(low=self.factors[0], high=self.factors[1]) for _ in range(data.shape[0])]  # type: ignore
+        else:
+            self.factor = self.R.uniform(low=self.factors[0], high=self.factors[1])
 
     def __call__(self, img: NdarrayOrTensor, randomize: bool = True) -> NdarrayOrTensor:
         """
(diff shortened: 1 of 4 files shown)


monai/transforms/intensity/array.py
```diff
@@ -601,6 +601,7 @@ def __init__(
         factors: Sequence[float] | float = 0,
         fixed_mean: bool = True,
         preserve_range: bool = False,
+        channel_wise: bool = False,
```
ericspod: Please add new arguments to the end of the argument list so that positional arguments aren't affected. I realise the argument was meant to be here given the docstring ordering, but if users call this constructor with positional arguments this will break compatibility.

monai/transforms/intensity/array.py
```diff
@@ -626,29 +627,51 @@ def __init__(
...
+            dtype=self.dtype,
         )
 
     def randomize(self, data: Any | None = None) -> None:
         super().randomize(None)
         if not self._do_transform:
             return None
-        self.factor = self.R.uniform(low=self.factors[0], high=self.factors[1])
+        if self.channel_wise:
+            self.factor = [self.R.uniform(low=self.factors[0], high=self.factors[1]) for _ in range(data.shape[0])]  # type: ignore
```
ericspod:
```suggestion
            self.factor = self.R.uniform(low=self.factors[0], high=self.factors[1], size=data.shape[:1])  # type: ignore
```
This might be slightly better.

monai/transforms/intensity/array.py
```diff
@@ -626,29 +627,51 @@ def __init__(
...
+                out_channel = ScaleIntensityFixedMean(
+                    factor=self.factor[i],  # type: ignore
+                    fixed_mean=self.fixed_mean,
+                    preserve_range=self.preserve_range,
+                    dtype=self.dtype,
+                )(d[None])[0]
+                out.append(out_channel)
+            ret: NdarrayOrTensor = torch.stack(out)
+            ret = convert_to_dst_type(ret, dst=img, dtype=self.dtype or img.dtype)[0]
+            return ret
```
ericspod, line 673:
```suggestion
        if self.channel_wise:
            out = []
            for i, d in enumerate(img):
                scale_trans = ScaleIntensityFixedMean(
                    factor=float(self.factor[i]),
                    fixed_mean=self.fixed_mean,
                    preserve_range=self.preserve_range,
                    dtype=self.dtype,
                )
                out.append(scale_trans(d[None]))
            ret: NdarrayOrTensor = torch.cat(out)
            ret = convert_to_dst_type(ret, dst=img, dtype=self.dtype or img.dtype)[0]
            return ret
```
I think this is a little more readable, the type ignore for `factor` may still be needed though.

monai/transforms/intensity/dictionary.py
```diff
@@ -669,6 +669,7 @@ def __init__(
         factors: Sequence[float] | float,
         fixed_mean: bool = True,
         preserve_range: bool = False,
+        channel_wise: bool = False,
```
ericspod: Same issue here, despite the docstring argument being incorrectly included in this order.

ericspod (changes requested):
Hi @engmohamedsalah I made a few comments below but it otherwise looks good to me with a few changes. Thanks!

engmohamedsalah:
@ericspod please check again after apply all your comments

ericspod (approved):

PR #8744, opened 2026-02-19
Fix batch size broadcasting bug in GeneralizedWassersteinDiceLoss

### LINKED ISSUE #4650: Bug in the generalized Wasserstein Dice loss
**Describe the bug**
The values returned by the loss are not correct for batch sizes larger than 1

**To Reproduce**
In the binary case and for M=[[1,0],[0,1]] the generalized Wasserstein Dice loss should be the same as the Dice loss.
For batch size > 1 it is not the case.

**Expected behavior**
See above. For this special case the GWDL and the DL should return the same values.

**Additional context**
I am the author of the generalized Wasserstein Dice loss paper and the bug has been already fixed in my git repository
https://github.com/LucasFidon/GeneralizedWassersteinDiceLoss
I will fix the bug in MONAI too and add unit tests.

hongjie-qiu:
**Fixes #4650**

### Description

When `batch_size > 1`, `GeneralizedWassersteinDiceLoss` produces incorrect loss values because of a tensor broadcasting issue in `_compute_generalized_true_positive` and `_compute_denominator`.

After `torch.gather`, `alpha_extended` has shape `(B, 1, S)` while `wasserstein_distance_map` has shape `(B, S)`. The element-wise multiply silently broadcasts to `(B, B, S)`, which mixes values across batch samples. This means the loss has always been wrong for any training run with `batch_size > 1`.

The fix follows the [reference implementation](https://github.com/LucasFidon/GeneralizedWassersteinDiceLoss) by the original paper's author — squeeze `dim=1` after the gather so both tensors are `(B, S)`, and reduce with `dim=1` instead of `dim=[1, 2]`.

I also noticed that `reduction="none"` was broken (never had test coverage) — it tried to reshape the per-sample loss `(B,)` into `(B, C, 1, ...)`, but GWDL aggregates over classes internally so the class dimension doesn't exist in the output. Fixed that as well.

### Changes

- `monai/losses/dice.py`: squeeze + dim fix in `_compute_generalized_true_positive` and `_compute_denominator`; fixed `reduction="none"` path
- `tests/losses/test_generalized_wasserstein_dice_loss.py`: two new regression tests for batch consistency

### Tests

All existing tests pass. The new regression tests fail on unpatched code and pass with the fix.

### DIFF
--- a/monai/losses/dice.py
+++ b/monai/losses/dice.py
@@ -548,10 +548,8 @@ class GeneralizedWassersteinDiceLoss(_Loss):
         elif self.reduction == LossReduction.SUM.value:
             wass_dice_loss = torch.sum(wass_dice_loss)  # sum over the batch and channel dims
         elif self.reduction == LossReduction.NONE.value:
-            # If we are not computing voxelwise loss components at least
-            # make sure a none reduction maintains a broadcastable shape
-            broadcast_shape = input.shape[0:2] + (1,) * (len(input.shape) - 2)
-            wass_dice_loss = wass_dice_loss.view(broadcast_shape)
+            # GWDL aggregates over classes internally, so wass_dice_loss has shape (B,)
+            pass
         else:
             raise ValueError(f'Unsupported reduction: {self.reduction}, available options are ["mean", "sum", "none"].')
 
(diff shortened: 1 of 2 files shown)


ericspod:
Hi @hongjie-qiu thanks for looking into this. Please look at the comments from Coderabbit and we can then review again.

hongjie-qiu:
Hey @ericspod I've updated the tests accordingly!

ericspod (approved):
Hi @hongjie-qiu thanks for this update, I think it's good to go now.

PR #1423, opened 2021-01-10
1404 Fixing a few code performance issues

de-sh:
part of #1404.

### Description
As discussed in the discussion linked above, these issues were detected using the [DeepSource](https://deepsource.io) static analyzer tool. These issues were specifically tagged and highlighted for their impact on the performance of MONAI, they are:
- Pass string format arguments as logging method parameters
- Remove unnecessary use of comprehension
- Use literal syntax instead of function calls to create data structure

### Status
**Ready**

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh --codeformat --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

PR #8747, opened 2026-02-22
Slightly improve performance in engines/utils.py

benediktjohannes:
### Description
This change probably improves performance slightly (unforetunately, I don't have a local instance of MONAI installed in order to test this, but I'm pretty sure from looking at the code that using [], {} and kwargs_[k] = _get_data(v) should be semantically the same and probably slightly better in performance, see also #1423 (credits to @de-sh for the nice fix 👍 ) where at least [] and {} were changed as a refactoring and kwargs_[k] = _get_data(v) just seems better to me becuase I think that this saves some time due to dictionary lookups and temporary objects, but I'm not quite sure because it wasn't tested, so please correct me if I'm mistaken).

### Types of changes

### DIFF
--- a/monai/engines/utils.py
+++ b/monai/engines/utils.py
@@ -235,7 +235,7 @@ class PrepareBatchExtraInput(PrepareBatch):
                 args_.append(_get_data(k))
         elif isinstance(self.extra_keys, dict):
             for k, v in self.extra_keys.items():
-                kwargs_[k] = _get_data(v)
+                kwargs_.update({k: _get_data(v)})
 
         return cast(torch.Tensor, image), cast(torch.Tensor, label), tuple(args_), kwargs_
 
--- a/monai/engines/utils.py
+++ b/monai/engines/utils.py
@@ -219,8 +219,8 @@ class PrepareBatchExtraInput(PrepareBatch):
         `kwargs` supports other args for `Tensor.to()` API.
         """
         image, label = default_prepare_batch(batchdata, device, non_blocking, **kwargs)
-        args_ = list()
-        kwargs_ = dict()
+        args_ = []
+        kwargs_ = {}
 
         def _get_data(key: str) -> torch.Tensor:
             data = batchdata[key]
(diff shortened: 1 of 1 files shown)


monai/engines/utils.py
```diff
@@ -219,8 +219,8 @@ def __call__(
         `kwargs` supports other args for `Tensor.to()` API.
         """
         image, label = default_prepare_batch(batchdata, device, non_blocking, **kwargs)
-        args_ = list()
-        kwargs_ = dict()
+        args_ = []
```
ericspod, line 222: Here this might make a slight improvement since this method is called a lot, but I would suggest the following would be more efficient still (You'd have to test this):

```python
def __call__(
    self,
    batchdata: dict[str, torch.Tensor],
    device: str | torch.device | None = None,
    non_blocking: bool = False,
    **kwargs: Any,
) -> tuple[torch.Tensor, torch.Tensor, tuple, dict]:
    """
    Args `batchdata`, `device`, `non_blocking` refer to the ignite API:
    https://pytorch.org/ignite/v0.4.8/generated/ignite.engine.create_supervised_trainer.html.
    `kwargs` supports other args for `Tensor.to()` API.
    """
    image, label = default_prepare_batch(batchdata, device, non_blocking, **kwargs)
    args_ = ()
    kwargs_ = {}

    def _get_data(key: str) -> torch.Tensor:
        data = batchdata[key]

            data = data.to(device=device, non_blocking=non_blocking, **kwargs)

        return data

    if isinstance(self.extra_keys, (str, list, tuple)):
        args_ = tuple(_get_data(k) for k in ensure_tuple(self.extra_keys))
    elif isinstance(self.extra_keys, dict):
        kwargs_ = {k: _get_data(v) for k, v in self.extra_keys.items()}

    return cast(torch.Tensor, image), cast(torch.Tensor, label), args_, kwargs_
```

benediktjohannes: Hi, thanks for suggesting this! I'm not quite sure about this yet and I don't have a local instance of MONAI installed at the moment, so testing this would take a while for me because I'm relatively stressed at the moment and I guess it would take some time for me to download and build MONAI locally.

But maybe we can put these suggestions in another PR lateron and for now merge the suggested changes for at least a little improvement, see here https://towardsdatascience.com/no-and-list-are-different-in-python-8940530168b0/

benediktjohannes:
@ericspod could you maybe open an extra PR for your additional suggestions and merge this for now, so that we can check on the more complex (and test requiring) changes lateron, but already have these safe slight improvements? Thanks!

ericspod (approved):
We can merge this now then if you want to come back to the suggestion later.

PR #8748, opened 2026-02-22
Use [] and {} instead of list() and dict()

benediktjohannes:
### Description
See also #8747 

### Types of changes

### DIFF
--- a/monai/transforms/inverse.py
+++ b/monai/transforms/inverse.py
@@ -277,8 +277,8 @@ class TraceableTransform(Transform):
                     msg += f" for key {key}"
 
                 pend = out_obj.pending_operations[-1]
-                statuses = pend.get(TraceKeys.STATUSES, dict())
-                messages = statuses.get(TraceStatusKeys.PENDING_DURING_APPLY, list())
+                statuses = pend.get(TraceKeys.STATUSES, {})
+                messages = statuses.get(TraceStatusKeys.PENDING_DURING_APPLY, [])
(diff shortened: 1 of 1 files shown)


benediktjohannes:
@coderabbitai do you agree with these changes?

benediktjohannes:
Wow! I am amazed how cool this tool is, I thought that this would definitely not work, just to chat here with the AI, very nice and helpful for workflows! Appreciate it! 👍

ericspod:
Hi @benediktjohannes I commented in another PR about replacing names like this with literal syntax. I suppose this saves a name lookup but I wonder if the interpreter doesn't replace these things automatically anyway. Do you actually see any speedup here? There's probably many other places in MONAI that should have this done as well.

benediktjohannes:
> Hi @benediktjohannes I commented in another PR about replacing names like this with literal syntax. I suppose this saves a name lookup but I wonder if the interpreter doesn't replace these things automatically anyway. Do you actually see any speedup here? There's probably many other places in MONAI that should have this done as well.

I think (back then (it's already been 4 weeks now, so I'm not quite sure), I've searched for all those list() things) and it should be worth merging, see here https://towardsdatascience.com/no-and-list-are-different-in-python-8940530168b0/

ericspod (approved):
We can merge this too, but as I said it's something that would only matter in very high frequency code.

benediktjohannes:
Thanks, See #8747 please

PR #8750, opened 2026-02-22
Follow-up for Fixing SSIM Test Method Names

benediktjohannes:
This fixes some other issues which were described in #8746 .

### Description
This is a follow-up for #8746, thanks and credits to @ericspod! 👍 

### Types of changes

### DIFF
--- a/tests/metrics/test_ssim_metric.py
+++ b/tests/metrics/test_ssim_metric.py
@@ -45,7 +45,7 @@ class TestSSIMMetric(unittest.TestCase):
         metric(preds, target)
         result = metric.aggregate()
         expected_value = 0.050103
-        self.assertTrue(expected_value - result.item() < 0.000001)
+        self.assertTrue(abs(expected_value - result.item()) < 0.000001)
 
     def test_3d_gaussian(self):
         set_determinism(0)
--- a/tests/metrics/test_ssim_metric.py
+++ b/tests/metrics/test_ssim_metric.py
@@ -21,7 +21,7 @@ from monai.utils import set_determinism
 
 class TestSSIMMetric(unittest.TestCase):
 
-    def test2d_gaussian(self):
+    def test_2d_gaussian(self):
         set_determinism(0)
         preds = torch.abs(torch.randn(2, 3, 16, 16))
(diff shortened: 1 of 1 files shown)


ericspod:
Hi @benediktjohannes, please look at what coderabbit as said here, the other method names also should be changed as they aren't getting run as tests it seems and line 35 should be `self.assertTrue(abs(expected_value - result.item()) < 1e-6)` . Thanks!

tests/metrics/test_ssim_metric.py
```diff
@@ -34,7 +34,7 @@ def test2d_gaussian(self):
         expected_value = 0.045415
         self.assertTrue(expected_value - result.item() < 0.000001)
```
ericspod:
```suggestion
        self.assertTrue(abs(expected_value - result.item()) < 0.000001)
```

tests/metrics/test_ssim_metric.py
```diff
@@ -47,7 +47,7 @@ def test2d_uniform(self):
         expected_value = 0.050103
         self.assertTrue(expected_value - result.item() < 0.000001)
```
ericspod:
```suggestion
        self.assertTrue(abs(expected_value - result.item()) < 0.000001)
```

ericspod (approved):

benediktjohannes:
Thanks for approving this! Unforetunately, we got an OS error in Lint

ericspod:
> Thanks for approving this! Unforetunately, we got an OS error in Lint

When the other checks pass try rerunning the job, this was an issue with a previous PR as well. I thought this was addressed earlier but I need to get back to my Actions overhaul PR.

benediktjohannes:
> > Thanks for approving this! Unforetunately, we got an OS error in Lint
> 
> When the other checks pass try rerunning the job, this was an issue with a previous PR as well. I thought this was addressed earlier but I need to get back to my Actions overhaul PR.

Thanks for the information! I think I don't have permissions to rerun jobs, but now all seem to pass except for one which is waiting for status to be reported, seems to be broken because it's been 3 hours already.

PR #8751, opened 2026-02-22
Use {} for better performance in dict()

benediktjohannes:
### Description
Follow-up for #8748 and so on (I'll combine these changes in future 👍)

### Types of changes

### DIFF
--- a/monai/apps/auto3dseg/auto_runner.py
+++ b/monai/apps/auto3dseg/auto_runner.py
@@ -229,7 +229,7 @@ class AutoRunner:
             input = os.path.join(os.path.abspath(work_dir), "input.yaml")
             logger.info(f"Input config is not provided, using the default {input}")
 
-        self.data_src_cfg = dict()
+        self.data_src_cfg = {}
         if isinstance(input, dict):
             self.data_src_cfg = input
         elif isinstance(input, str) and os.path.isfile(input):


ericspod:
Hi @benediktjohannes I not sure this change is really necessary. I know I merged the previous PR about replacing `tuple()` with `()` but I'm not sure that would actually do anything. Here I think `dict()` and `{}` should be the same thing, and regardless the constructor is called only the once typically. Do you actually see any speedup from this?

benediktjohannes:
> Hi @benediktjohannes I not sure this change is really necessary. I know I merged the previous PR about replacing `tuple()` with `()` but I'm not sure that would actually do anything. Here I think `dict()` and `{}` should be the same thing, and regardless the constructor is called only the once typically. Do you actually see any speedup from this?

We replaced this in #1423 e.g. in line "properties = dict()" to "properties = {}" as well, so I think it's better.

benediktjohannes:
cc @ericspod

ericspod (approved):

ericspod:
I can merge this now but the thing with these minor changes is that they would only matter in high frequency code anyway. The lookup time saved only matters if the line is executed a very large number of times.

benediktjohannes:
@ericspod thanks! Could you please also merge the one in the hot loop path? This should be way more important. Thanks!

PR #8752, opened 2026-02-22
fix: support more dtypes in pad_nd function

skdas20:
Removed hardcoded dtype exclusion list
Now supports bool and integer dtypes with PyTorch's optimized padding

### DIFF
--- a/docs/source/handlers.rst
+++ b/docs/source/handlers.rst
@@ -83,6 +83,12 @@ Panoptic Quality metrics handler
     :members:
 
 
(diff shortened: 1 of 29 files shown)


ericspod:
Hi @skdas20 we have a previous PR #8672 that includes this fix. I am looking at that one now as we've left it to sit for too long, sorry if this was a fix you were needing.

ericspod:
Sorry for overtaking your PR but this one doesn't introduce changes anymore so we'll close now.

PR #8753, opened 2026-02-22
Improve/writer error messages

skdas20:
Added helpful error checking for missing tensorboard package
Provides installation instructions when package not found

### DIFF
--- a/monai/handlers/tensorboard_handlers.py
+++ b/monai/handlers/tensorboard_handlers.py
@@ -27,11 +27,13 @@ if TYPE_CHECKING:
     from ignite.engine import Engine
     from tensorboardX import SummaryWriter as SummaryWriterX
     from torch.utils.tensorboard import SummaryWriter
+
+    _tb_available = True
 else:
     Engine, _ = optional_import(
         "ignite.engine", IgniteInfo.OPT_IMPORT_VERSION, min_version, "Engine", as_type="decorator"
     )
-    SummaryWriter, _ = optional_import("torch.utils.tensorboard", name="SummaryWriter")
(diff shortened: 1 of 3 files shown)


ericspod:
Hi @skdas20 I think this would be a good addition but please look at the comment from coderabbit, the proposed fix looks reasonable to me. Please also look at the DCO issue. Thanks!

skdas20:
Rebased onto current `dev` and force-pushed. Two things fixed:

- **DCO** was failing because an older commit in this branch's history (from the now-closed #8752) was unsigned. The branch is now a single signed commit containing only this PR's change — the diff is unchanged (`tensorboard_handlers.py` + its test).
- **mypy** was reporting `Name "_tb_available" is not defined`: the flag was only bound in the `else:` branch of `if TYPE_CHECKING:`, so type-checkers never saw it. It's now also bound in the `TYPE_CHECKING` branch.

The CodeRabbit findings from before are still addressed: the availability flag from `optional_import` is captured and checked (the old `SummaryWriter is None` guard was unreachable, since `optional_import` returns a lazy-raising stub rather than `None`), the `Raises:` section is documented, and there's a regression test that patches the flag to simulate tensorboard being absent.

skdas20:
@ericspod gentle nudge on this one — both things you asked for were sorted in the July update: the DCO failure (an unsigned commit inherited from the closed #8752; the branch is now a single signed commit) and the CodeRabbit points, which now report no actionable comments. All checks are green and it's still mergeable against `dev`.

No rush at all if it's queued behind other things — just flagging it in case it slipped off the list.

ericspod (approved):
Hi @skdas20 sorry for the long delay, this looks good now, thanks. I accepted a very minor change from Coderabbit in the test as well.

Issue #7562, opened 2024-03-20
Enhance `ExtractDataKeyFromMetaKeyd` to work with ` MetaTensor`


KumoLiu:
Try `ExtractDataKeyFromMetaKeyd` to try and pull from the metadata of a MetaTensor later to work with `image_only=True`

Kent-McPhail:
I would like to try and tackle this issue.  This would be a first issue for me and am just looking for some guidance on completing it

KumoLiu:
Hi @Kent-McPhail, please refer to this contribution guide: https://github.com/Project-MONAI/MONAI/blob/dev/CONTRIBUTING.md

eliottvalette:
Hi @KumoLiu,

I’d also like to help with this issue as it aligns well with my goal of practicing for my first contribution. Can multiple people collaborate on a single issue, or can it only be assigned to one person? I read the contribution guide but couldn't find this information.

Thanks!

eliottvalette:
I forked the repo and managed to handle MetaTensors, but I'm struggling to understand what needs to be done with image_only. I've tried to understand it by referring to the functioning of LoadImaged, but I wouldn't mind some help.

aylward:
It is great if multiple people collaborate on a single issue, there can be multiple assignees, and please keep posting your questions.

Perhaps @ericspod can provide more info on the intent.  In general, image_only is passed to a function/transform when it is returning an image and some other data.  For example, Affine() returns the tuple (image, affine) if image_only is false, and returns image otherwise...but I am not following Eric's comment in that PR.

eliottvalette:
Nice, that’s great news—thank you!

Also, thank you for explaining how image_only works. That’s roughly what I expected, but I was initially surprised that such a parameter would be part of a transformation focused on metadata. I hadn’t realized that when the input is a dictionary (rather than a MetaTensor), the “image” could end up being stored in what we call metadata.

Additionally, I have a question: since the upgraded version of ExtractDataKeyFromMetaKeyd can handle both dictionaries and MetaTensors, should the output always be a dictionary for consistency, or should it return an updated MetaTensor when a MetaTensor is provided? I implemented an inplace parameter in case the second option (returning the updated MetaTensor) is preferred.

ericspod:
[...]

PR #8754, opened 2026-02-22
Enhance/metatensor extraction

skdas20:
Enhanced [ExtractDataKeyFromMetaKeyd] with [image_only] parameter
Properly handles MetaTensor metadata without duplication


ericspod:
Hi @skdas20 thanks for the contribution. Please look at the error outputs and the coderabbit comments. We can then review later.

skdas20:
Looks like issue #7562 was already resolved upstream via a different approach (`ExtractDataKeyFromMetaKeyd` now handles the MetaTensor case natively on `dev` without an `image_only` flag). This PR's approach no longer applies cleanly on top of that and would just be extra API surface for something already solved, so closing it — same situation as #8752. Thanks for the look either way.

Issue #8667, opened 2025-12-21
Ignore Class


ManasChoudhary1:
**Is your feature request related to a problem? Please describe.**
In various datasets, background class or unlabeled class is index by some number, a simple ignore_class = IGNORE_CLASS can help.

**Describe the solution you'd like**
Add a ignore_class argument everywhere, which takes integer class value to be ignored in losses.

PR #8757, opened 2026-02-25
feat(losses/metrics): implement ignore_index support across dice and …

### LINKED ISSUE #8734: Feature Request: Support ignore_index / ignore_class in Losses and Metrics
In practical medical image segmentation, it is very common to have label values that should be ignored (e.g. padding, unlabeled regions, auxiliary classes).

Currently, MONAI losses and metrics (e.g. DiceLoss, DiceMetric, MeanIoU) do not natively support ignoring specific classes or label values. Users must manually mask predictions and targets, which is error-prone and inconsistent.

Request:
Add native support for an argument such as ignore_index or ignore_classes to losses and metrics.

Example:

`DiceCELoss(include_background=False, ignore_index=2)`
`DiceMetric(include_background=False, ignore_index=2)`

Expected behavior:

Voxels with ignored labels are excluded from numerator and denominator

Works consistently for binary and multi-class segmentation

Compatible with reduction="mean" and get_not_nans

This would improve correctness, reproducibility, and align MONAI with common PyTorch APIs (e.g. CrossEntropyLoss(ignore_index)).

Thanks for considering!

Rusheel86:
### Description
This PR introduces comprehensive native support for `ignore_index` in core MONAI losses and metrics, as requested in issue #8734. The `ignore_index` parameter allows specific label values (e.g., padding, unlabeled regions, or boundary artifacts) to be excluded from loss and metric calculations which is a critical feature for medical imaging workflows.

### Implementation Summary

**Losses with `ignore_index` support:**
- `DiceLoss` - masks ignored voxels before dice computation
- `FocalLoss` - applies masking after one-hot conversion
- `TverskyLoss` - consistent masking approach with DiceLoss
- `UnifiedFocalLoss` - handles ignore_index in binary target conversion

**Metrics with `ignore_index` support:**
- `MeanDice` & `GeneralizedDiceScore` - spatial masking before computation
- `MeanIoU` - excludes ignored classes from IoU calculation  
- `ConfusionMatrixMetric` - filters ignored indices from confusion matrix
- `HausdorffDistanceMetric` - masks boundary computation
- `SurfaceDiceMetric` & `SurfaceDistanceMetric` - handles ignore_index in surface calculations

**Key Fixes:**
- Added support for `ignore_index`
- Added comprehensive test coverage in `test_ignore_index_losses.py` and `test_ignore_index_metrics.py`
- Replaced hardcoded numeric stability constants with configurable `epsilon` parameter in `GeneralizedDiceScore`

### Types of changes
- [x] Non-breaking change (fix or new feature that would not break existing functionality)
- [ ] Breaking change (fix or new feature that would cause existing functionality to change)
- [x] New tests added to cover the changes (`test_ignore_index_losses.py`, `test_ignore_index_metrics.py`)
- [x] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`
- [x] Quick tests passed locally by running `./runtests.sh --quick --unittests --disttests`
- [x] In-line docstrings updated
- [x] Documentation updated, tested `make html` command in the `docs/` folder

### Related Issues
Could merge with or addresses similar requirements as #8667 (Ignore Class)


monai/metrics/generalized_dice.py
```diff
@@ -147,52 +155,82 @@ def compute_generalized_dice(
...
-    y_o = torch.sum(y, dim=reduce_axis)
-    y_pred_o = torch.sum(y_pred, dim=reduce_axis)
+    intersection = torch.sum(y[:, channels_to_use, ...] * y_pred[:, channels_to_use, ...], dim=reduce_axis)
+    y_o = torch.sum(y[:, channels_to_use, ...], dim=reduce_axis)
+    y_pred_o = torch.sum(y_pred[:, channels_to_use, ...], dim=reduce_axis)
+
     denominator = y_o + y_pred_o
 
     # Set the class weights
+    # Set the class weights (computed from scored channels only)
```
aymuos15: Duplicated

aymuos15:
Just curious, wont a robust helper specifically for this be cleaner and be passed around to the loss/metric funcs?

Rusheel86:
> Just curious, wont a robust helper specifically for this be cleaner and be passed around to the loss/metric funcs?

Ah yes, thanks for the suggestion, that should be less work than this and cleaner. I initially kept the logic scoped to the loss function to avoid over-engineering while I was verifying the math, buta robust helper in the utils would be much cleaner.
I’ll refactor and update the PR shortly.

monai/losses/dice.py
```diff
@@ -100,6 +102,7 @@ def __init__(
                 The value/values should be no less than 0. Defaults to None.
             soft_label: whether the target contains non-binary values (soft labels) or not.
                 If True a soft label formulation of the loss will be used.
+            ignore_index: class index to ignore from the loss computation.
```
ericspod: There needs to be more details on what this parameter means in terms of the implications to the output loss value. It's not clear what happens when this parameter is set to something, what valid values are, etc. It's not clear why one would use parameter.

monai/losses/dice.py
```diff
@@ -163,12 +167,15 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
...
 
+        original_target = target if self.ignore_index is not None else None
+
         if self.to_onehot_y:
             if n_pred_ch == 1:
                 warnings.warn("single channel prediction, `to_onehot_y=True` ignored.")
             else:
                 target = one_hot(target, num_classes=n_pred_ch)
 
+        mask = create_ignore_mask(original_target if original_target is not None else target, self.ignore_index)
```
ericspod:
```suggestion
        mask = create_ignore_mask(target, self.ignore_index)
```

I think the logic of `original_target if original_target is not None else target` isn't correct here.
From this logic, when `self.ignore_index` is not `None` then `original_target` is `target` (and not `None`), so the expression here evaluates to `target`. If `self.ignore_index` is `None` then so is `original_target`, so the expression evaluates to `target` again.

monai/losses/dice.py
```diff
@@ -180,6 +187,10 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
         if target.shape != input.shape:
             raise AssertionError(f"ground truth has different shape ({target.shape}) from input ({input.shape})")
 
+        if mask is not None:
+            input = input * mask
+            target = target * mask
+
```
ericspod: This pattern appears in the loss functions you've modified. Why not turn `create_ignore_mask` into a function (`mask_loss_inputs` or something like that) which takes `input` and `target` and returns them either directly if `ignore_mask` is None or their masked versions otherwise. This would reduce code duplication in a lot of places and capture the intended purpose in a function.

monai/losses/tversky.py
```diff
@@ -129,8 +133,16 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
             if n_pred_ch == 1:
                 warnings.warn("single channel prediction, `to_onehot_y=True` ignored.")
             else:
+                original_target = target
                 target = one_hot(target, num_classes=n_pred_ch)
 
+        if self.ignore_index is not None:
+            mask_src = original_target if self.to_onehot_y and n_pred_ch > 1 else target
```
ericspod: Similarly as the above logic question, this value appears to evaluate to `target` no matter what.

monai/losses/unified_focal_loss.py
```diff
@@ -39,48 +40,69 @@ def __init__(
...
         dice_class = (tp + self.epsilon) / (tp + self.delta * fn + (1 - self.delta) * fp + self.epsilon)
 
         # Calculate losses separately for each class, enhancing both classes
         back_dice = 1 - dice_class[:, 0]
-        fore_dice = (1 - dice_class[:, 1]) * torch.pow(1 - dice_class[:, 1], -self.gamma)
+        fore_dice = torch.pow(1 - dice_class[:, 1], 1 - self.gamma)
 
         # Average class scores
-        loss = torch.mean(torch.stack([back_dice, fore_dice], dim=-1))
+        loss = torch.stack([back_dice, fore_dice], dim=-1)
```
ericspod, line 93: Is this a bug fix where we shouldn't have been applying `mean` before? If so, please check (if you haven't already) that the original version of the loss gave erroneous values when reduciton was something other than "mean", that's where I'd expect differences. Please also remove the comment above this line which isn't correct now.

monai/losses/unified_focal_loss.py
```diff
@@ -103,42 +125,66 @@ def __init__(
...
             else:
+                if self.ignore_index is not None:
+                    # Replace ignore_index with valid class before one_hot
+                    y_true = torch.where(y_true == self.ignore_index, torch.tensor(0, device=y_true.device), y_true)
                 y_true = one_hot(y_true, num_classes=n_pred_ch)
 
         if y_true.shape != y_pred.shape:
             raise ValueError(f"ground truth has different shape ({y_true.shape}) from input ({y_pred.shape})")
 
+        mask = create_ignore_mask(original_y_true if original_y_true is not None else y_true, self.ignore_index)
```
ericspod: Same logic question. I'm probably missing a few other places where this pattern occurs as well.

monai/losses/unified_focal_loss.py
```diff
@@ -103,42 +125,66 @@ def __init__(
...
         fore_ce = cross_entropy[:, 1]
         fore_ce = self.delta * fore_ce
 
-        loss = torch.mean(torch.sum(torch.stack([back_ce, fore_ce], dim=1), dim=1))
+        loss = torch.stack([back_ce, fore_ce], dim=1)  # [B, 2, H, W]
+
+        if self.reduction == LossReduction.MEAN.value:
+            if mask is not None:
+                masked_loss = loss * mask
+                return masked_loss.sum() / mask.expand_as(loss).sum().clamp(min=1e-5)
```
ericspod: In general we need to avoid using fixed constants like `1e-5`, every now and again some case will arise needing a different value. Can we use `self.epsilon` instead? If not use need another member to use here.

monai/losses/unified_focal_loss.py
```diff
@@ -207,28 +257,41 @@ def forward(self, y_pred: torch.Tensor, y_true: torch.Tensor) -> torch.Tensor:
             ValueError: When num_classes
             ValueError: When the number of classes entered does not match the expected number
         """
-        if y_pred.shape != y_true.shape:
```
ericspod, line 210: I think this check should be here still since nothing between here and where it was moved will change the shape of either tensor. It just avoids doing useless work before raising the exception.

monai/losses/unified_focal_loss.py
```diff
@@ -207,28 +257,41 @@ def forward(self, y_pred: torch.Tensor, y_true: torch.Tensor) -> torch.Tensor:
...
 
-        if torch.max(y_true) != self.num_classes - 1:
-            raise ValueError(f"Please make sure the number of classes is {self.num_classes - 1}")
+        # Save original for masking before one-hot conversion
+        original_y_true = y_true if self.ignore_index is not None else None
 
-        n_pred_ch = y_pred.shape[1]
         if self.to_onehot_y:
-            if n_pred_ch == 1:
-                warnings.warn("single channel prediction, `to_onehot_y=True` ignored.")
```
ericspod: We should keep this pattern of warning for single channel images and not try to one-hot those. Should masking also not be used with single channel images?

monai/metrics/generalized_dice.py
```diff
@@ -147,52 +155,77 @@ def compute_generalized_dice(
...
+        numer = 2.0 * intersection
+        denom = denominator
     else:
         numer = 2.0 * (intersection * w)
         denom = denominator * w
-        y_pred_o = y_pred_o
 
     # Compute the score
-    generalized_dice_score = numer / denom
+    generalized_dice_score = numer / (denom + 1e-6)
```
ericspod: Also constant here should be a member with a default value, possibly much smaller too.

monai/metrics/hausdorff_distance.py
```diff
@@ -179,17 +192,20 @@ def compute_hausdorff_distance(
...
+        yp = y_pred[b, c]
+        yt = y[b, c]
+
         _, distances, _ = get_edge_surface_distance(
-            y_pred[b, c],
-            y[b, c],
-            distance_metric=distance_metric,
-            spacing=spacing_list[b],
-            symmetric=not directed,
-            class_index=c,
```
ericspod, line 188: The change here omits this parameter, is there something else we can use instead? It's used for warning about empty ground truths/predictions, I think it should work correctly when `c` is not `self.ignore_index`?

monai/metrics/meandice.py
```diff
@@ -277,16 +284,22 @@ def __init__(
...
         Compute the dice metric for binary inputs which have only spatial dimensions. This method is called separately
         for each batch item and for each channel of those items.
 
         Args:
             y_pred: input predictions with shape HW[D].
             y: ground truth with shape HW[D].
+            mask: binary mask where 0 indicates voxels to ignore.
         """
+        if mask is not None:
+            y_pred = y_pred & mask.bool()
```
ericspod: Why cast `mask` to `bool` here and use `&`? I don't think this was done elsewhere, perhaps `mask` should be converted to the type of `y_pred` then multiplied? I guess I'm missing something you've encountered.

monai/metrics/surface_dice.py
```diff
@@ -264,12 +282,21 @@ def compute_surface_dice(
...
+            # Handle areas being returned as a single item or a tuple
+            if isinstance(areas, (list, tuple)):
+                if len(areas) == 2:
+                    areas_pred, areas_gt = areas
+                elif len(areas) == 1:
+                    areas_pred = areas_gt = areas[0]
+                else:
+                    areas_pred = areas_gt = torch.tensor([], device=y_pred.device)
+            else:
+                areas_pred = areas_gt = areas
```
ericspod: I'm not sure this is necessary from the return value of `get_edge_surface_distance` but the docstring there isn't precise enough. This could be done as:


```suggestion
            areas = ensure_tuple(areas)
            if len(areas) in {1, 2}:
                areas_pred, areas_gt = areas[0], areas[-1]
            else:
                areas_pred = areas_gt = torch.tensor([], device=y_pred.device)
```

monai/metrics/utils.py
```diff
@@ -320,28 +365,40 @@ def get_edge_surface_distance(
     edges_spacing = None
     if use_subvoxels:
         edges_spacing = spacing if spacing is not None else ([1] * len(y_pred.shape))
-    (edges_pred, edges_gt, *areas) = get_mask_edges(
-        y_pred, y, crop=True, spacing=edges_spacing, always_return_as_numpy=False
-    )
-    if not edges_gt.any():
```
ericspod, line 364: Are these warnings not needed now? If they interfere with masking then still something should be adapted to keep them if possible.

monai/metrics/utils.py
```diff
@@ -320,28 +365,40 @@ def get_edge_surface_distance(
...
         )  # type: ignore
     else:
-        distances = (get_surface_distance(edges_pred, edges_gt, distance_metric, spacing),)  # type: ignore
-    return convert_to_tensor(((edges_pred, edges_gt), distances, tuple(areas)), device=y_pred.device)  # type: ignore[no-any-return]
+        distances_raw = (get_surface_distance(edges_pred, edges_gt, distance_metric, spacing),)  # type: ignore
+
+    distances_list = list(distances_raw)
+    distances: tuple[torch.Tensor, torch.Tensor] | tuple[torch.Tensor] = (
+        tuple(distances_list) if len(distances_list) == 2 else (distances_list[0],)  # type: ignore[assignment]
+    )
```
ericspod: `ensure_tuple` could be used here.

monai/metrics/utils.py
```diff
@@ -320,28 +365,40 @@ def get_edge_surface_distance(
...
+
+    areas = edge_results[2:] if use_subvoxels else ()
+
+    # Ensure areas is always a tuple of 2 when use_subvoxels=True
+    if use_subvoxels and isinstance(areas, (list, tuple)):
+        if len(areas) == 1:
+            areas = (areas[0], areas[0])
+        elif len(areas) != 2:
+            # Unexpected length, create empty tensors
+            areas = (torch.tensor([], device=y_pred.device), torch.tensor([], device=y_pred.device))
```
ericspod: Here also `ensure_tuple` could reduce the amount of code.

ericspod (changes requested):
Hi @Rusheel86 sorry for the late response on this PR. I have a number of comments, some changes and some questions. I think this is good overall though I need someone else to review as well to be sure I've caught everything as it's a large PR. We're preparing for a release so it's unlikely to get through in time, but I'll try to keep things moving along to get it in relatively soon.

ericspod:
@Rusheel86 there are conflict I'll leave to you to resolve as they're a little complex but then we can run tests again. @garciadias @atbenmurray if either of you could have a look at this one, please, that would help double checking things. Thanks!

Rusheel86:
Hey @ericspod, thanks a lot for the detailed review, really appreciate you going through such a large PR.
I'm on summer vacation right now so I'll have plenty of time to work on this and  go through all the comments properly and get everything addressed. I'll try to have it all fixed and ready before the release goes out if that's still possible.
Will push updates soon :)

monai/metrics/meandice.py
```diff
@@ -414,7 +427,10 @@ def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor | tupl
                         f"Got y_pred={tuple(y_pred.shape)}, y={tuple(y.shape)}."
                     )
 
-        first_ch = 0 if self.include_background and not self.per_component else 1
+        # Create global mask for ignored voxels if ignore_index is set
```
garciadias, line 441: The `first_ch` expression changed from `first_ch = 0 if self.include_background and not self.per_component else 1` to `first_ch = 0 if self.include_background else 1`. This changes per-component behavior when `include_background=True` — previously the background channel (0) was excluded from iteration, now it is included. Is this intentional? If so, please document the behavioral change in the PR description. If not, please restore the `not self.per_component` guard.

monai/losses/focal_loss.py
```diff
@@ -204,12 +221,18 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
...
         elif self.reduction == LossReduction.MEAN.value:
-            loss = loss.mean()
+            if mask is not None:
+                # Ensure we only sum the loss where the mask is 1
+                # Then divide by the actual number of 1s in the mask
+                loss = (loss * mask).sum() / mask.sum().clamp(min=1e-5)
+            else:
+                loss = loss.mean()
+
         elif self.reduction == LossReduction.NONE.value:
```
garciadias, line 209: The `else: raise ValueError(...)` for unsupported `reduction` values was removed. Previously an invalid reduction would raise; now it silently falls through to `return loss` (equivalent to `"none"`). Please restore the error guard:

```python
elif self.reduction == LossReduction.NONE.value:
    pass
else:
    raise ValueError(
        f"Unsupported reduction: {self.reduction}, available options are [\"mean\", \"sum\", \"none\"]."
    )
```

monai/metrics/generalized_dice.py
garciadias: The score is computed as `numer / (denom + 1e-6)` but explicit `denom_zeros` handling exists below (lines 219+). Using both a small epsilon and an explicit zero-division guard is redundant and can conflict. Recommend either removing the `+ 1e-6` and relying solely on the explicit guard, or removing the `denom_zeros` guard and keeping only the epsilon.

monai/metrics/hausdorff_distance.py
```diff
@@ -179,17 +192,20 @@ def compute_hausdorff_distance(
     spacing_list = prepare_spacing(spacing=spacing, batch_size=batch_size, img_dim=img_dim)
 
     for b, c in np.ndindex(batch_size, n_class):
+        yp = y_pred[b, c]
+        yt = y[b, c]
+
         _, distances, _ = get_edge_surface_distance(
```
garciadias, line 182: The `class_index=c` kwarg was removed from the `get_edge_surface_distance` call. This means the "all zero" warning in that function will now say "Unknown" instead of the actual class index. Minor diagnostic regression — please restore `class_index=c`.

tests/metrics/test_ignore_index_metrics.py
```diff
@@ -0,0 +1,91 @@
+# Copyright (c) MONAI Consortium
```
garciadias, line 1: All tests use `ignore_index=255` which exercises only the sentinel code path in `create_ignore_mask` (where `ignore_index >= num_classes`). The class-index path (`0 <= ignore_index < num_classes`) — which masks a specific channel — has zero test coverage for both losses and metrics. This path is used by `GeneralizedDiceScore` (channel removal via `channels_to_use.remove(ignore_index)`) and by `get_confusion_matrix`.

Please add test cases with `ignore_index` values within the valid class range (e.g., `ignore_index=0`, `ignore_index=1`) for both losses and metrics to cover this code path.

monai/losses/focal_loss.py
```diff
@@ -161,6 +165,11 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
         if target.shape != input.shape:
             raise ValueError(f"ground truth has different shape ({target.shape}) from input ({input.shape})")
 
+        mask = create_ignore_mask(target, self.ignore_index)
```
garciadias: The `create_ignore_mask(target, self.ignore_index)` call at this point receives `target` that may have already been clipped by `include_background=False` (lines 157-163 above). For a class-index-based `ignore_index` (e.g., `ignore_index=2`), the mask would target the wrong channel because the tensor has been sliced.

In contrast, `DiceLoss` and `TverskyLoss` correctly save the original target before one-hot conversion (`original_target = target if self.ignore_index is not None else None`) and pass that to `create_ignore_mask` before the `include_background` clipping.

Please save target before `include_background` clipping and pass it to `create_ignore_mask` instead of the already-clipped `target`.

monai/metrics/meandice.py
```diff
@@ -128,6 +128,7 @@ def __init__(
         num_classes: int | None = None,
         return_with_label: bool | list[str] = False,
         per_component: bool = False,
+        ignore_index: int | None = None,
```
garciadias, line 135: **FL Compatibility — Heterogeneous `ignore_index` Breaks Metric Comparability (FL1)**

In MONAI federated learning ([`monai/fl/client/monai_algo.py`](https://github.com/Project-MONAI/MONAI/blob/dev/monai/fl/client/monai_algo.py)), each client runs its own evaluator with its own bundle config, and reports metrics via `ExchangeObject(metrics=self.evaluator.state.metrics)`. The server aggregates these metrics — typically by averaging — with **no metadata about the metric configuration used**.

If different FL clients use different `ignore_index` values (or one uses `ignore_index` and another does not), their reported Dice/IoU values operate on **different effective pixel sets** and are **not comparable**. Averaging them produces a meaningless number.

This is a pre-existing architectural gap (the same applies to `include_background`), but `ignore_index` makes it easier to accidentally create heterogeneous configurations across clients.

**Recommendation (out of scope for this PR but worth noting):** Add a docstring note in the `ignore_index` parameter docs for all affected losses and metrics, warning FL users that `ignore_index` must be consistent across all clients for meaningful metric aggregation.

monai/metrics/generalized_dice.py
```diff
@@ -147,52 +155,77 @@ def compute_generalized_dice(
     if y.shape != y_pred.shape:
         raise ValueError(f"y_pred - {y_pred.shape} - and y - {y.shape} - should have the same shapes.")
 
-    # Ignore background, if needed
+    # Apply ignore_index masking
+    mask = create_ignore_mask(y, ignore_index)
```
garciadias, line 171: **FL Compatibility — GeneralizedDiceScore Channel Count Mismatch (FL2)**

In `compute_generalized_dice`, when `ignore_index` is a valid class index (0 <= ignore_index < n_channels), that channel is removed from the scored set via `channels_to_use.remove(ignore_index)`. This means the output tensor shape changes: `[B, C]` becomes `[B, C-1]`.

If FL clients disagree on `ignore_index`:
- Client A (ignore_index=3) → output shape `[B, 9]`
- Client B (ignore_index=None) → output shape `[B, 10]`

These would be reported as differently-shaped tensors inside `ExchangeObject.metrics`. While unlikely to crash (metrics are typically flattened to a dict), the inconsistency would go undetected by the FL server.

**Recommendation:** Consider documenting in the `GeneralizedDiceScore` docstring that `ignore_index` changes the output channel count, or compute the score for the ignored class as NaN instead of removing it — whichever is more appropriate for your use case.

monai/metrics/utils.py
```diff
@@ -83,6 +84,44 @@ def ignore_background(
...
+        Returns None if ignore_index is None.
+    """
+    if ignore_index is None:
+        return None
+
+    if y.shape[1] == 1:
+        # Label-encoded: direct comparison
+        return (y != ignore_index).float()
+
+    # One-hot encoded
```
garciadias, line 125: **Documentation — Negative `ignore_index` Asymmetry (DOC1)**

In `create_ignore_mask`, a negative `ignore_index` (e.g., `ignore_index=-1`) behaves differently depending on the target encoding:

- **Label-encoded** (`y.shape[1] == 1`): `(y != ignore_index).float()` correctly masks out `-1` labels — a common convention for void/unlabeled pixels.
- **One-hot encoded** (`y.shape[1] > 1`): Since `0 <= -1` is False, the code falls through to the sentinel path, masking all-zero spatial positions instead.

This asymmetry could surprise users who switch between label-encoded and one-hot targets expecting the same `ignore_index` behavior. Worth noting in the `create_ignore_mask` docstring that negative values are only meaningful for label-encoded inputs.

garciadias:
Great work on this PR, @Rusheel86. Thank you! 🙏                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                                                  I've left a handful of inline comments with suggestions — mostly minor inconsistencies and some thoughts around Federated Learning compatibility.
None of them are blockers on the core idea. All looks great to me.

Thanks again for pushing this forward. 
👏🙌

Rusheel86:
Hey @garciadias, thank you so much for the kind words and the detailed review, I really appreciate you taking the time to go through it.
I'll address all the suggestions and get all of this sorted shortly. Always glad to help and its been a great learning experience for me too :)

Rusheel86:
Hey @ericspod @garciadias, sorry for the wait, I've pushed the changes after I went through all the review comments. Here's a basic overview:

### Masking

- Moved `create_ignore_mask` to `utils.py` so all callers import from one place
- In `focal_loss.py`, the mask is now built from the raw target before any clipping happens, fixes the wrong-channel bug
- Added a `mask_loss_inputs` helper that replaces the repeated inline masking pattern across all loss files

### Losses

- `FocalLoss` now correctly saves the original target before `include_background` slicing and passes that to `create_ignore_mask`
- `DiceLoss`, `TverskyLoss`, and `UnifiedFocalLoss` all use the new `mask_loss_inputs` helper consistently
- Restored the `else: raise ValueError(...)` guard in `focal_loss.py` for unsupported reduction values
- Restored `not self.per_component` in the `first_ch` expression in `meandice.py`, that was an accidental drop

### Metrics

- `class_index=c` restored in the `get_edge_surface_distance` call in `hausdorff_distance.py` so warnings show the actual class index
- Hardcoded `1e-5` and `1e-6` constants replaced with `self.epsilon` across `unified_focal_loss.py` and `generalized_dice.py`
- Redundant zero-division guard in `generalized_dice.py` removed, keeping only `self.epsilon`

### Tests

- Added test cases with `ignore_index=0` and `ignore_index=1` for both losses and metrics, the sentinel-only (`ignore_index=255`) coverage gap flagged
- New cases cover the channel-masking path in `GeneralizedDiceScore` and `get_confusion_matrix`

### Docs

- Added FL compatibility warning to `ignore_index` docstrings across all affected losses and metrics
- Noted in `GeneralizedDiceScore` that output shape changes from `[B, C]` to `[B, C-1]` when a valid class index is passed
- Noted in `create_ignore_mask` that negative values only behave as expected for label-encoded inputs

I'll push an empty commit later so all CI checks run properly but everything should be good for another look, let me know if anything needs more work :)

Rusheel86:
All 3 failing checks appear to be CI infrastructure issues unrelated to this PR's changes, all other checks are passing.

1. **static mypy check and full-dep (ubuntu latest)** : fails with `ModuleNotFoundError: No module named 'pytest'` inside `cupy/testing/_random.py`, outside the scope of this PR.

2. **packaging test** : fails inside `monai/utils/type_conversion.py` and `monai/transforms/`, none of which are touched by this PR.

monai/losses/focal_loss.py
```diff
@@ -161,6 +164,11 @@ def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
         if target.shape != input.shape:
             raise ValueError(f"ground truth has different shape ({target.shape}) from input ({input.shape})")
 
+        if self.ignore_index is not None:
+            mask = (target != self.ignore_index).float()
+            input = input * mask
+            target = target * mask
+
```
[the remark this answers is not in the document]
garciadias: Hi @Rusheel86, this one doesn't seem resolved. It would be useful to understand more about your decision to resolve this issue.

monai/losses/unified_focal_loss.py
```diff
@@ -132,13 +154,24 @@ def forward(self, y_pred: torch.Tensor, y_true: torch.Tensor) -> torch.Tensor:
...
         fore_ce = cross_entropy[:, 1]
         fore_ce = self.delta * fore_ce
 
-        loss = torch.mean(torch.sum(torch.stack([back_ce, fore_ce], dim=1), dim=1))
+        loss = torch.stack([back_ce, fore_ce], dim=1)  # [B, 2, H, W]
+        if self.reduction == LossReduction.MEAN.value:
+            if self.ignore_index is not None:
+                # Normalize by the number of non-ignored pixels
+                return loss.sum() / spatial_mask.sum().clamp(min=1e-5)
+            return loss.mean()
```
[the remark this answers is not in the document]
garciadias: @Rusheel86 same. This looks relevant to me.

garciadias (dismissed):
Thank you, @Rusheel86! 🚀 🥇

Rusheel86:
> Thank you, @Rusheel86! 🚀 🥇

Always happy to contribute :)

tests/losses/test_ignore_index_losses.py
```diff
@@ -0,0 +1,68 @@
...
+from parameterized import parameterized
+
+from monai.losses import AsymmetricUnifiedFocalLoss, DiceLoss, FocalLoss, TverskyLoss
+
+# Defining test cases: (LossClass, args)
+TEST_CASES = [
+    (DiceLoss, {"sigmoid": True}),
+    (FocalLoss, {"use_softmax": False}),
+    (TverskyLoss, {"sigmoid": True}),
+    (AsymmetricUnifiedFocalLoss, {}),
```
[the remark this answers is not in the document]
ericspod: @coderabbitai has this comment been addressed?

tests/metrics/test_ignore_index_metrics.py
ericspod: @coderabbitai has this comment been addressed adequately in the current tests?

tests/metrics/test_ignore_index_metrics.py
```diff
@@ -0,0 +1,163 @@
...
+        metric.reset()
+        metric(y_pred=y_pred2, y=y)
+        res2 = metric.aggregate()
+        if isinstance(res2, list):
+            res2 = res2[0]
+
+        torch.testing.assert_close(res1, res2, msg=f"Failed for {metric_class.__name__}")
+
+    @parameterized.expand(NO_BACKGROUND_METRICS)
+    def test_metric_ignore_class_index_without_background(self, metric_class, kwargs):
```
ericspod, line 117: This could be put into a separate class so that it's not guarded by the `skipUnless` to skip when scipy isn't present since these don't need scipy. We always test with scipy however.

tests/metrics/test_ignore_index_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+SCIPY_METRICS = [
+    (HausdorffDistanceMetric, {"include_background": True}),
+    (SurfaceDistanceMetric, {"include_background": True}),
+    (SurfaceDiceMetric, {"class_thresholds": [0.5, 0.5], "include_background": True}),
+]
+
+
+@unittest.skipUnless(has_scipy, "Scipy required for surface metrics")
+class TestIgnoreIndexMetrics(unittest.TestCase):
+    @parameterized.expand(TEST_METRICS + SCIPY_METRICS)
```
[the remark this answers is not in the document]
ericspod: @Rusheel86 we do always test with scipy so this is technically still an issue. We can refactor this some other time.

ericspod:
Hi @Rusheel86 sorry for the delay on this one again. There's still a number of open comments that haven't been resolved from myself and @garciadias. A number of these appear obsolete so please go through them and those as resolved. Other comments have not been answered, and coderabbit comments are still present about issues that should be addressed. This PR got busy so some comments get hidden sometimes by the interface. Please have a look through again, we do want this to be integrated into MONAI. Thanks!

Rusheel86:
> Hi @Rusheel86 sorry for the delay on this one again. There's still a number of open comments that haven't been resolved from myself and @garciadias. A number of these appear obsolete so please go through them and those as resolved. Other comments have not been answered, and coderabbit comments are still present about issues that should be addressed. This PR got busy so some comments get hidden sometimes by the interface. Please have a look through again, we do want this to be integrated into MONAI. Thanks!

Hi, thanks for following up and I'm really glad to hear this is something you'd like to see integrated!
I'll go through all the open comments and work on it this weekend and update the PR accordingly.

Rusheel86:
Hello @ericspod @garciadias, thanks again for the reviews, everything has been properly addressed now so feel free to take another look. I reviewed everything a couple of times so it took a bit of time but should be good now :)

monai/metrics/surface_dice.py
```diff
@@ -255,6 +283,7 @@ def compute_surface_dice(
             use_subvoxels=use_subvoxels,
             symmetric=True,
             class_index=c,
+            warn_empty=warn_empty,
         )
         boundary_correct: int | torch.Tensor | float
         boundary_complete: int | torch.Tensor | float
```
garciadias, line 289: The new code computes absolute_c = c + class_offset on the line above, but the class_index argument passed to get_edge_surface_distance is still c, not absolute_c.

**What this affects:** class_index is used inside get_edge_surface_distance only in warning messages, e.g. "the ground truth of class X is all 0, this may result in nan/inf distance." When include_background=False, channels are shifted after background removal: c=0 means actual class 1, c=1 means actual class 2, etc. Passing c instead of absolute_c causes the warning to name the wrong class.

**Concrete example:** A 3-class model with include_background=False and ignore_index=None. After ignoring background, the remaining channels are [class 1, class 2], indexed as c=0,1. If class 2 (c=1, absolute_c=2) has no ground truth, the warning would say "class 1" instead of "class 2." If ignore_index=2, the warning is suppressed (correctly via warn_empty), but when ignore_index is None or a different class, the wrong class name appears in the warning.

**Inconsistency:** Both compute_hausdorff_distance and compute_average_surface_distance in this same PR correctly pass class_index=absolute_c to their get_edge_surface_distance calls. compute_surface_dice is the only one still passing c.

**Fix:** Change class_index=c to class_index=absolute_c at this call site.

monai/losses/unified_focal_loss.py
```diff
@@ -187,10 +225,14 @@ def __init__(
         self.gamma = gamma
         self.delta = delta
         self.weight: float = weight
-        self.asy_focal_loss = AsymmetricFocalLoss(gamma=self.gamma, delta=self.delta)
-        self.asy_focal_tversky_loss = AsymmetricFocalTverskyLoss(gamma=self.gamma, delta=self.delta)
+        self.asy_focal_loss = AsymmetricFocalLoss(
+            to_onehot_y=False, gamma=self.gamma, delta=self.delta, ignore_index=ignore_index
```
garciadias: When ignore_index is a sentinel value (e.g., 255), the parent AsymmetricUnifiedFocalLoss.forward applies mask_loss_inputs to the data, then delegates to AsymmetricFocalLoss.forward and AsymmetricFocalTverskyLoss.forward — which each independently call create_ignore_mask and mask_loss_inputs again on the already-masked input.

The double-masking is mathematically idempotent (mask values are 0/1, so applying it twice produces the same result), so there is no correctness bug. But it:
- Wastes compute (extra mask creation and element-wise multiplication)
- Confuses ownership — it is unclear whether the parent or sub-loss is responsible for masking
- Creates a latent maintenance risk: if mask semantics change to non-binary weighting in the future, double-masking becomes incorrect

For valid class indices (e.g., ignore_index=0), use_mask is False in the parent and the sub-losses handle masking correctly through their own create_ignore_mask path. So the issue is only on the sentinel path.

Suggested fix: construct the sub-losses with ignore_index=None so masking is handled exclusively at the parent level.

monai/metrics/utils.py
```diff
@@ -84,6 +86,47 @@ def ignore_background(
     return y_pred, y
```
garciadias, line 88: For one-hot encoded inputs with a valid class index (0 <= ignore_index < num_classes), the mask is computed as 1.0 - y[:, ignore_index:ignore_index+1]. This returns a spatial mask of shape (B, 1, H, W, ...) that zeroes all channels at locations where the ignored class is present.

This is spatial masking ("ignore pixels belonging to class X everywhere"), not per-channel exclusion ("ignore the score for class X only"). Both are valid design choices, but users coming from other frameworks (e.g., PyTorch's CrossEntropyLoss where ignore_index only affects the loss for that class) might expect channel-level exclusion.

The behavior is correct given the docstring's description of "Voxels with this label are excluded from the score," and the tests confirm it. Consider adding a brief note in the docstring clarifying that for one-hot inputs, the masking is spatial across all channels, not per-channel, to help users understand the intended semantics.

monai/metrics/utils.py
```diff
@@ -358,30 +411,63 @@ def get_edge_surface_distance(
...
+        edges_gt = edges_gt & mask
+        if edges_pred_any and not edges_pred.any():
+            warn_empty_pred = False
+        if edges_gt_any and not edges_gt.any():
+            warn_empty_gt = False
+        if not mask.any():
+            warn_empty_pred = False
+            warn_empty_gt = False
+
+    if warn_empty_gt and not edges_gt.any():
```
garciadias, line 437: When use_subvoxels=True and the areas tuple has an unexpected length (!= 2), the code silently creates empty tensors as fallback. This could conceal upstream issues in get_mask_edges where an unexpected return tuple size might indicate a bug.

Consider adding a warnings.warn here to surface this condition for diagnosis, e.g.:

  warnings.warn(f"Unexpected number of area tensors from get_mask_edges: {len(areas)}, expected 2. Falling back to empty tensors.")

The fallback itself is a reasonable defensive measure — the suggestion is just to make it visible when it triggers.

monai/metrics/generalized_dice.py
```diff
@@ -51,11 +56,15 @@ def __init__(
         include_background: bool = True,
         reduction: MetricReduction | str = MetricReduction.MEAN,
         weight_type: Weight | str = Weight.SQUARE,
+        ignore_index: int | None = None,
+        epsilon: float = 1e-8,
```
garciadias, line 60: A new epsilon parameter (1e-8) was added to GeneralizedDiceScore and compute_generalized_dice. This is a numeric stability improvement and appears unrelated to the ignore_index feature this PR targets. The change is fine to include, but it is not mentioned in the PR description.

For reviewer and future-reader clarity, consider either mentioning the epsilon addition in the PR body, or splitting numeric stability improvements into a separate focused PR.

garciadias (changes requested):

garciadias:
> Hello @ericspod @garciadias, thanks again for the reviews, everything has been properly addressed now so feel free to take another look. I reviewed everything a couple of times so it took a bit of time but should be good now :)

I am sorry, @Rusheel86, but I think the latest changes have introduced some issues that need to be solved before merging. I didn't want to add more friction to this, but there are a couple of important things to address.

Many thanks for your contribution, patience, and endurance. 🙏🏾

garciadias (changes requested):
Thanks for this PR — well-structured implementation with good test coverage. Two findings to address before merge:

1. **compute_surface_dice** passes class_index=c instead of absolute_c (inconsistent with hausdorff/surface_distance in same PR)
2. **UnifiedFocalLoss** double-masks inputs when ignore_index is a sentinel value (sub-losses constructed with ignore_index=ignore_index)

Minor items are suggestions only. See inline comments for details.

Rusheel86:
@garciadias thanks for your patience too, this pr has been a great learning experience for me and I would like it to be upto the standards required so all reviews are greatly appreciated :)
I've added the epsilon change in the pr description and other code changes as suggested, feel free to let me know anything else too.

garciadias (approved):
Thank you! 🚀

ericspod:
@coderabbitai Summarize all the open comments in this PR and indicate whether they have been addressed in subsequent commits and so could be closed.

ericspod (approved):

PR #8761, opened 2026-03-01
fix: print install hint when no suitable ImageWriter found

### LINKED ISSUE #7980: Print what package should be installed when suitable writer is missing
I used `SaveImage(..., output_ext=".png")` but got this error message

```
monai.utils.module.OptionalImportError: No ImageWriter backend found for png.
```
This was after `pip install monai` in a clean environment.

It would be great if the error message would include a description of how to fix the issue, e..g. "Please install pillow". Initially I installed simpleitk which also didn't have the png writer apparently, and I got the same error.

n/a

**Additional context**
Happened with `monai==1.3.2`

yashnaiduu:
Fixes #7980

## What changed
When [resolve_writer()](cci:1://file:///Users/yashnaidu/.gemini/antigravity/scratch/opensource/MONAI/monai/data/image_writer.py:91:0-137:23) fails to find a backend for a given file extension,
the error was generic and unhelpful:

    OptionalImportError: No ImageWriter backend found for nii.gz.

Users had no idea which package was missing.

## Fix
Added a format-to-package lookup in [resolve_writer()](cci:1://file:///Users/yashnaidu/.gemini/antigravity/scratch/opensource/MONAI/monai/data/image_writer.py:91:0-137:23) that appends
an install hint for common formats:

| Format | Install |
|--------|---------|
| `.nii`, `.nii.gz` | `pip install nibabel` |
| `.mha`, `.mhd`, `.nrrd` | `pip install itk` |
| [.png](cci:7://file:///Users/yashnaidu/.gemini/antigravity/brain/be97a37c-5b6e-4a0e-bf27-4acea75f5bbe/ragas_fork_success_1772387250858.png:0:0-0:0), `.jpg`, `.tif`, `.bmp` | `pip install pillow` |

## After this fix
    OptionalImportError: No ImageWriter backend found for 'nii.gz'.
    Try installing the required package: pip install nibabel

No behaviour change for users who already have the packages installed.

### DIFF
--- a/monai/data/image_writer.py
+++ b/monai/data/image_writer.py
@@ -63,6 +63,21 @@ __all__ = [
 
 SUPPORTED_WRITERS: dict = {}
 
+# Maps common file extensions to the package needed by ImageWriter backends.
+FILETYPE_HINT: dict[str, str] = {
+    "nii": "nibabel",
+    "nii.gz": "nibabel",
+    "mha": "itk",
+    "mhd": "itk",
+    "nrrd": "itk",
+    "png": "pillow",
+    "jpg": "pillow",
+    "jpeg": "pillow",
+    "tif": "pillow",
+    "tiff": "pillow",
+    "bmp": "pillow",
+}
+
 
 def register_writer(ext_name, *im_writers):
     """
@@ -116,21 +131,8 @@ def resolve_writer(ext_name, error_if_not_found=True) -> Sequence:
         except Exception:  # other writer init errors indicating it exists
             avail_writers.append(_writer)
     if not avail_writers and error_if_not_found:
-        install_hints: dict = {
-            "nii": "nibabel",
-            "nii.gz": "nibabel",
-            "mha": "itk",
-            "mhd": "itk",
-            "nrrd": "itk",
-            "png": "pillow",
-            "jpg": "pillow",
-            "jpeg": "pillow",
-            "tif": "pillow",
-            "tiff": "pillow",
-            "bmp": "pillow",
-        }
-        hint = install_hints.get(fmt)
-        extra = f" Try installing the required package: pip install {hint}" if hint else ""
+        hint = FILETYPE_HINT.get(fmt)
+        extra = f" Try: pip install {hint}" if hint else ""
         raise OptionalImportError(f"No ImageWriter backend found for '{fmt}'.{extra}")
     writer_tuple = ensure_tuple(avail_writers)
     SUPPORTED_WRITERS[fmt] = writer_tuple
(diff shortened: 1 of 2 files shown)


monai/data/image_writer.py
```diff
@@ -116,7 +116,22 @@ def resolve_writer(ext_name, error_if_not_found=True) -> Sequence:
         except Exception:  # other writer init errors indicating it exists
             avail_writers.append(_writer)
     if not avail_writers and error_if_not_found:
-        raise OptionalImportError(f"No ImageWriter backend found for {fmt}.")
+        install_hints: dict = {
```
ericspod: I would put this dictionary into a global variable at the top of this file. A name like `FILETYPE_HINT` or something else sticking to the all-caps convention would work. This keeps the dictionary easily found and modifiable at runtime.

ericspod (changes requested):
Hi @yashnaiduu thanks for this update. I think it's fine as is with my suggested comment, but we should add at least one test to ensure the behaviour works. You would need only a test to check that the text in the exception message shows up to suggest a package when not installed, and doesn't when the file extension isn't known. You may need to mock the dictionary to emulate a package being missing (which you can do if it's a global variable).

yashnaiduu:
Thanks for the review @ericspod! I've moved the dictionary to a module-level constant `FILETYPE_HINT` as suggested and also added the requested unit tests in `tests/data/test_image_rw.py`.

monai/data/image_writer.py
```diff
@@ -116,7 +116,23 @@ def resolve_writer(ext_name, error_if_not_found=True) -> Sequence:
...
             avail_writers.append(_writer)
     if not avail_writers and error_if_not_found:
-        raise OptionalImportError(f"No ImageWriter backend found for {fmt}.")
+        # map common extensions to their required package
+        install_hints: dict = {
+            "nii": "nibabel",
+            "nii.gz": "nibabel",
+            "mha": "itk",
+            "mhd": "itk",
+            "nrrd": "itk",
```
dzenanz, line 125: This list is not exhaustive. A longer list is available [here](https://docs.itk.org/projects/wasm/en/latest/introduction/file_formats/images.html). Also, NRRD format can have `.nhdr` extension.

ericspod: Hi @yashnaiduu please have a look into adding more to the dictionary.

ericspod:
@yashnaiduu you should also have a look at what contributors to #7437 are doing for improving the messages with `LoadImage`. It's a similar problem but they have slightly different approaches. I'm trying to coordinate getting their PRs through after neglecting that and this one for too long, sorry.

PR #8762, opened 2026-03-02
Fft cleanup/update

aymuos15:
Remove legacy PyTorch 1.8.0 compatibility code from FFT utilities. MONAI now requires PyTorch ≥ 2.4.1, so the version checks and NumPy fallbacks are no longer needed.

### Types of changes

### DIFF
--- a/monai/networks/blocks/fft_utils_t.py
+++ b/monai/networks/blocks/fft_utils_t.py
@@ -28,9 +28,6 @@ def roll_1d(x: Tensor, shift: int, shift_dim: int) -> Tensor:
 
     Returns:
         1d-shifted version of x
-
-    Note:
-        This function is called when fftshift and ifftshift are not available in the running pytorch version
     """
     shift = shift % x.size(shift_dim)
     if shift == 0:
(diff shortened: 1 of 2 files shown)


aymuos15:
I guess the coderabbit comment is preexisting and not in the scope of this PR? I can make that change separately after the decision on this change?

ericspod (approved):
Thanks for the cleanup!

ericspod:
> I guess the coderabbit comment is preexisting and not in the scope of this PR? I can make that change separately after the decision on this change?

It appears so, yes. It should be removed with a deprecation warning which can be done as a separate PR, though it seems to be low priority.

aymuos15:
> > I guess the coderabbit comment is preexisting and not in the scope of this PR? I can make that change separately after the decision on this change?
> 
> It appears so, yes. It should be removed with a deprecation warning which can be done as a separate PR, though it seems to be low priority.

Sounds good. Thank you very much. Ill make one then after this is merged. I haven't looked into it to really verify what all that bit of code is touching.

PR #8764, opened 2026-03-03
Vectorize DiceHelper.__call__()

aymuos15:
## Summary
- Replace nested batch/channel Python loops in `DiceHelper.__call__()` with vectorized torch operations
- Remove dead `compute_channel` method (no longer called after vectorization)

## Note on memory
The current implementation vectorizes across both batch and channel dimensions simultaneously, which increases peak memory for large 3D volumes. If this becomes an issue, we can switch to looping over batch while keeping channels vectorized — this retains most of the speedup while keeping memory proportional to a single sample.

## Test plan
- [x] All existing `test_compute_meandice` tests pass
- [x] New test cases added covering mixed-format edge cases:
  - Single-channel `y` (class indices) + multi-channel `y_pred` (one-hot)
  - Single-channel `y_pred` (argmaxed, with `num_classes`) + multi-channel `y` (one-hot)
  - Same as above with `include_background=False` and batched inputs

### DIFF
--- a/monai/metrics/meandice.py
+++ b/monai/metrics/meandice.py
@@ -486,16 +486,16 @@ class DiceHelper:
                 return (f, not_nans) if self.get_not_nans else f
 
         first_ch = 0 if self.include_background else 1
-        data = []
+        data_list = []
         for b in range(y_pred.shape[0]):
             c_list = []
             for c in range(first_ch, n_pred_ch) if n_pred_ch > 1 else [1]:
                 x_pred = (y_pred[b, 0] == c) if (y_pred.shape[1] == 1) else y_pred[b, c].bool()
                 x = (y[b, 0] == c) if (y.shape[1] == 1) else y[b, c]
                 c_list.append(self.compute_channel(x_pred, x))
-            data.append(torch.stack(c_list))
+            data_list.append(torch.stack(c_list))
 
-        data = torch.stack(data, dim=0).contiguous()  # type: ignore
+        data = torch.stack(data_list, dim=0).contiguous()
 
         f, not_nans = do_metric_reduction(data, self.reduction)  # type: ignore
         return (f, not_nans) if self.get_not_nans else f
--- a/monai/metrics/meandice.py
+++ b/monai/metrics/meandice.py
@@ -374,6 +374,53 @@ class DiceHelper:
             return torch.tensor(1.0, device=y_o.device)
         return torch.tensor(0.0, device=y_o.device)
 
+    def compute_confusion_dice(self, y_pred: torch.Tensor, y: torch.Tensor, n_pred_ch: int) -> torch.Tensor:
+        """
+        Fast Dice for multi-class label maps. Both ``y_pred`` and ``y`` are single-channel integer class-index maps
+        with values in ``[0, n_pred_ch)``. A single batched confusion matrix is accumulated with ``torch.bincount``
+        (one pass over the voxels, independent of the number of classes), and per-class true-positive, prediction and
+        ground-truth counts are read off its diagonal and margins. This matches :meth:`compute_channel` exactly while
+        avoiding both the per-channel Python loop and the materialization of one-hot tensors.
+
+        Args:
+            y_pred: predicted class indices with shape (batch_size, 1, spatial_dims...).
+            y: ground-truth class indices with shape (batch_size, 1, spatial_dims...).
+            n_pred_ch: number of classes (background included).
+
+        Returns:
+            Per-(batch, class) Dice with shape (batch_size, num_classes_selected), background excluded when
+            ``include_background`` is False, following the same ``ignore_empty`` convention as the per-channel path.
+        """
+        batch_size = y_pred.shape[0]
+        device = y_pred.device
+        y_pred_flat = y_pred.reshape(batch_size, -1).long()
+        y_flat = y.reshape(batch_size, -1).long()
+        batch_offset = torch.arange(batch_size, device=device).unsqueeze(1) * (n_pred_ch * n_pred_ch)
+        indices = (batch_offset + y_pred_flat * n_pred_ch + y_flat).reshape(-1)
+        confusion = torch.bincount(indices, minlength=batch_size * n_pred_ch * n_pred_ch)
+        confusion = confusion.reshape(batch_size, n_pred_ch, n_pred_ch).to(torch.float32)
+
+        tp = torch.diagonal(confusion, dim1=1, dim2=2)
+        pred_sum = confusion.sum(dim=2)
+        y_sum = confusion.sum(dim=1)
+        dice = (2.0 * tp) / (pred_sum + y_sum)
+
+        if self.ignore_empty:
+            dice = torch.where(y_sum > 0, dice, torch.tensor(float("nan"), device=device, dtype=dice.dtype))
+        else:
+            dice = torch.where(
+                y_sum == 0,
+                torch.where(
+                    pred_sum == 0,
+                    torch.tensor(1.0, device=device, dtype=dice.dtype),
+                    torch.tensor(0.0, device=device, dtype=dice.dtype),
+                ),
+                dice,
+            )
+
+        first_ch = 0 if self.include_background else 1
+        return dice[:, first_ch:]
+
     def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
         """
         Compute the metric for the given prediction and ground truth.
@@ -382,6 +429,9 @@ class DiceHelper:
             y_pred: input predictions with shape (batch_size, num_classes or 1, spatial_dims...).
                 the number of channels is inferred from ``y_pred.shape[1]`` when ``num_classes is None``.
             y: ground truth with shape (batch_size, num_classes or 1, spatial_dims...).
+
+        Returns:
+            Per-(batch, class) Dice scores, or ``(scores, not_nans)`` when ``get_not_nans`` is True.
 
         Raises:
             ValueError: when the shapes of `y_pred` and `y` are not compatible for the per-component computation.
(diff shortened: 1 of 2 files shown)


monai/metrics/meandice.py
```diff
@@ -413,21 +394,72 @@ def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor | tupl
...
+
+        # Vectorized computation (replaces nested loops for better performance)
+        batch_size = y_pred.shape[0]
+        device = y_pred.device
+
+        # Convert y_pred to boolean (handle single-channel class indices vs multi-channel one-hot independently)
+        if y_pred.shape[1] == 1 and n_pred_ch > 1:
+            y_pred_bool = torch.zeros(batch_size, n_pred_ch, *y_pred.shape[2:], dtype=torch.bool, device=device)
+            for c in range(n_pred_ch):
+                y_pred_bool[:, c] = y_pred[:, 0] == c
```
ericspod: Could this be done instead with `torch.nn.functional.one_hot` without the loop?

monai/metrics/meandice.py
```diff
@@ -413,21 +394,72 @@ def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor | tupl
...
+        else:
+            y_pred_bool = y_pred.bool()
+
+        # Convert y: single-channel class indices → one-hot bool; multi-channel → preserve raw values
+        if y.shape[1] == 1 and n_pred_ch > 1:
+            y_expanded = torch.zeros(batch_size, n_pred_ch, *y.shape[2:], dtype=torch.float32, device=device)
+            for c in range(n_pred_ch):
+                y_expanded[:, c] = (y[:, 0] == c).float()
```
ericspod: Same here possibly.

monai/metrics/meandice.py
```diff
@@ -413,21 +394,72 @@ def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor | tupl
...
-                c_list.append(self.compute_channel(x_pred, x))
-            data.append(torch.stack(c_list))
-
-        data = torch.stack(data, dim=0).contiguous()  # type: ignore
+        # Flatten spatial dimensions for vectorized computation: (batch, channels, -1)
+        y_pred_flat = y_pred_bool.reshape(batch_size, n_pred_ch, -1).float()
+        y_flat = y_expanded.reshape(batch_size, n_pred_ch, -1).float()
+
+        # Compute Dice per (batch, channel) vectorized: all reductions at once
+        intersection = torch.sum(y_pred_flat * y_flat, dim=-1)  # (batch, n_pred_ch)
```
ericspod: The original method used `torch.masked_select`, would that not be better/faster somehow that just multiplication? I don't know what the reasoning was in the original.

monai/metrics/meandice.py
```diff
@@ -413,21 +394,72 @@ def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor | tupl
                         "(B, 2, H, W) or (B, 2, D, H, W). "
                         f"Got y_pred={tuple(y_pred.shape)}, y={tuple(y.shape)}."
                     )
+            data = torch.stack(
+                [
+                    self.compute_cc_dice(y_pred=y_pred[b].unsqueeze(0), y=y[b].unsqueeze(0)).reshape(-1)
+                    for b in range(y_pred.shape[0])
+                ],
+                dim=0,
+            ).contiguous()
```
ericspod, line 472:
```suggestion
            batch_cc = [self.compute_cc_dice(y_pred[b : b + 1], y[b : b + 1]).flatten() for b in range(y_pred.shape[0])]
            data = torch.stack(batch_cc, dim=0).contiguous()
```

I feel this is more readable, but check it's equivalent if you accept this change.

ericspod (changes requested):
Hi @aymuos15 I have some minor comments but it looks fine overall. I am worried a bit about memory consumption going up with this, if you have any numbers of how much we can have a look but typically the amount of data without gradients is pretty low when processing metrics. I have some minor comments to look at as well, mostly suggestion only.

aymuos15:
Post rebasing, checking your comments and doing a very deep benchmarking, I have taken a different approach.

Let me know if this is okay (since there is a if statement to cover the minor cons of the proposed approach), happy to close otherwise. 

The old path is the per channel Python loop, which is light on memory but gets slow as the number of classes grows. The vectorized path I had earlier (one-hot plus multiply) is fast in 2D but its peak memory balloons with class count, up to about 1 GB on big volumes. The new path builds a single `bincount` confusion matrix for label maps with 3+ classes and falls back to the loop otherwise, so it's fast and the memory stays flat. The dispatch sends `C<3` to the loop (no regression) and `C≥3` to the confusion matrix.

**Speedup by class count** (3D 128³, B1, old/confmat):

| C | 2 | 3 | 4 | 5 | 6 | 8 | 14 | 20 | 32 | 50 | 105 |
|---|---|---|---|---|---|---|---|---|---|---|---|
| ×faster | 0.7 | 1.5 | 2.4 | 3.3 | 3.8 | 5.3 | 10.1 | 15.0 | 24.3 | 37.5 | 73.1 |

Cost is class-independent (one pass), so improvement grows with C. `C=2` is the only case where the loop wins, which is why the gate is at `C≥3`.

**Dispatched vs old, time / peak GPU mem:**

| Case | old loop | dispatched | speedup |
|---|---|---|---|
| 2D B1 C2 256² | 0.21 ms / 1 M | 0.21 ms / 1 M | 1.0× |
| 2D B8 C2 256² | 1.30 ms / 1 M | 1.30 ms / 1 M | 1.0× |
| 2D B8 C9 256² | 7.10 ms / 1 M | 0.56 ms / 8 M | 12.7× |
| 2D B4 C21 512² | 13.85 ms / 3 M | 0.83 ms / 16 M | 16.7× |
| 3D B1 C2 128³ | 1.36 ms / 29 M | 1.36 ms / 29 M | 1.0× |
| 3D B2 C3 128³ | 4.70 ms / 21 M | 3.45 ms / 64 M | 1.4× |
| 3D B2 C6 128³ | 10.02 ms / 20 M | 2.99 ms / 64 M | 3.4× |
| 3D B2 C14 128³ | 23.30 ms / 20 M | 2.72 ms / 64 M | 8.6× |
| 3D B4 C14 96³ | 22.24 ms / 8 M | 2.32 ms / 54 M | 9.6× |
| 3D B2 C8 160×160×128 | 20.26 ms / 31 M | 4.52 ms / 100 M | 4.5× |
| 3D B1 C32 96³ | 12.71 ms / 8 M | 0.73 ms / 14 M | 17.4× |
| 3D B1 C105 96³ | 41.97 ms / 8 M | 0.76 ms / 14 M | 55.2× |

<sub>RTX A1000 Laptop GPU · driver 580.126.20 · CUDA 12.8 · torch 2.10.0+cu128 · i7-12800H</sub>

tests/metrics/test_compute_meandice.py
```diff
@@ -359,6 +410,33 @@ def test_channel_dimensions(self):
         with self.assertRaises(ValueError):
             DiceMetric(per_component=True)(torch.ones([3, 3, 144, 144]), torch.ones([3, 3, 144, 144]))
 
+    def test_label_map_fast_path(self):
+        torch.manual_seed(0)
+        for n in (3, 8):
+            for include_background in (True, False):
+                for ignore_empty in (True, False):
```
ericspod, line 417: It's better to use parameterized with a list of values created with `dict_product` rather than a loop. This will exit on the first assert fail and mask further failures. The alternative is use `subTest` around the asserts.

tests/metrics/test_compute_meandice.py
```diff
@@ -359,6 +410,33 @@ def test_channel_dimensions(self):
...
+                    y = torch.randint(0, n, (2, 1, 8, 8, 8))
+                    fast = DiceHelper(**kwargs)(y_pred, y)
+                    ref = DiceHelper(**kwargs)(torch.nn.functional.one_hot(y_pred[:, 0], n).movedim(-1, 1), y)
+                    np.testing.assert_array_equal(torch.isnan(fast).cpu().numpy(), torch.isnan(ref).cpu().numpy())
+                    np.testing.assert_allclose(
+                        torch.nan_to_num(fast).cpu().numpy(), torch.nan_to_num(ref).cpu().numpy(), atol=1e-4
+                    )
+
+    def test_label_map_out_of_range_fallback(self):
+        for y_pred in (torch.tensor([[[[0, 1, 2, 3]]]]), torch.tensor([[[[-1, 1, 2, 0]]]])):
```
ericspod, line 435: Same

monai/metrics/meandice.py
```diff
@@ -413,21 +463,39 @@ def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor | tupl
...
+            f, not_nans = do_metric_reduction(data, self.reduction)  # type: ignore
+            return (f, not_nans) if self.get_not_nans else f
+
+        # multi-class label maps use the single-pass confusion matrix; all else uses the per-channel loop
+        label_maps = (
+            y_pred.shape[1] == 1 and y.shape[1] == 1 and not y_pred.is_floating_point() and not y.is_floating_point()
+        )
+        if label_maps and n_pred_ch >= 3:
```
ericspod, line 480: The choice of 3 is an optimisation heuristic in that larger numbers of classes should use the old channel loop approach. This should be documented here with a comment about this and the motivation.

ericspod (changes requested):
Hi @aymuos15 I see this is updated to use a different technique. From the results table you have the time cost is much less but the memory is much higher. For larger volumes with many channels this can be 3x or more, which can cause an OOM in an unexpected way and end a training run. The cost of slower code is worse since one just needs to wait longer, and for a section of code that's a small component of the training time this isn't as impactful as a raised exception stopping everything. I've made some comments again below that are minor, but I feel this isn't in the right direction for what we should be optimising for.

ericspod:
@atbenmurray @garciadias @virginiafdez @mingxin-zheng @vikashg any input on this change would be appreciated, my concern is the time<->space tradeoff.

aymuos15:
Hi @ericspod I agree and I think for now we can simply close this. Will wait for a day for someone else tagged to reply. Thank you very much for your time on this.

ericspod:
> Hi @ericspod I agree and I think for now we can simply close this. Will wait for a day for someone else tagged to reply. Thank you very much for your time on this.

Thanks for the effort, we'll keep this in our back pocket anyway if we want to come back around.

PR #8765, opened 2026-03-03
Add explicit spatial_ndim tracking to MetaTensor

### LINKED ISSUE #6397: metatensor einops rearrange
**Describe the bug**
```py
import torch
from monai.data import MetaTensor
from einops import rearrange
from monai.transforms import Resize

# create a metatensor in batch mode
x = MetaTensor(torch.randn((1, 1, 481, 404, 3))).to("cuda:0")
x.is_batch = True
x.meta['affine'] = torch.eye(4)[None]

# select the first element in the batch and resize
out = Resize(spatial_size=(1024, 860, -1), align_corners=True, mode="trilinear")(x[0])
print(out.shape)  # works fine

# rearrange using einops
x_ = rearrange(x, "b c h w d -> (b c) h w d")
out = Resize(spatial_size=(1024, 860, -1), align_corners=True, mode="trilinear")(x_)
print(out.shape)  # error

```

error:
```
    affine = orig_affine @ to_affine_nd(len(orig_affine) - 1, affine, dtype=torch.float64)
ValueError: matmul: Input operand 1 has a mismatch in its core dimension 0, with gufunc signature (n?,k),(k,m?)->(n?,m?) (size 1 is different from 4)
```

workaround (convert to a regular tensor `as_tensor()`):
```py
x_ = rearrange(x, "b c h w d -> (b c) h w d")
if isinstance(x_, MetaTensor):
    x_ = x_.as_tensor()
out = Resize(spatial_size=(1024, 860, -1), align_corners=True, mode="trilinear")(x_)
print(out.shape)
```

aymuos15:
## Summary
Fixes #6397

- Adds a `_spatial_ndim: int` attribute to `MetaTensor` that explicitly tracks the number of spatial dimensions, preventing dimension-mismatch crashes when `einops.rearrange()` or other reshape operations change `ndim`
- The attribute propagates through `copy_meta_from` (via `__dict__` copy) and is preserved through arbitrary torch operations
- Updates transforms (`Resize`, `Rotate`, `Zoom`, `Flip`, `Affine`, `SplitDim`, `AddCoordinateChannels`, etc.) and lazy resampling to use `spatial_ndim` instead of hardcoded 3

### Key design decisions
- **Constructor**: `spatial_ndim = min(affine.shape[-1] - 1, ndim - 1)` — clamped by actual tensor dims
- **Affine setter**: `spatial_ndim = affine.shape[-1] - 1` — no clamping (user is explicit)
- **`peek_pending_affine`**: uses affine's inner matrix shape (fixes batched `(1,4,4)` case)
- **`spatial_resample`**: `min(spatial_ndim, ndim - 1, 3)` — adds ndim-1 constraint as safety net

### Files changed (16)
- `monai/data/meta_obj.py`, `meta_tensor.py`, `utils.py`, `__init__.py` — core MetaTensor changes
- `monai/transforms/` — spatial, croppad, intensity, inverse, lazy, post, utility transforms updated
- `tests/data/meta_tensor/test_spatial_ndim.py` — 18 new tests
- Existing test files updated with `spatial_ndim` assertions

## Test plan
- [x] 18 new unit tests for `spatial_ndim` property (construction, affine sync, propagation, einops reshape, transforms)
- [x] Existing MetaTensor tests pass (162 tests)
- [x] SqueezeDim and SplitDim tests pass with new assertions
- [x] Total: 216 tests verified passing


monai/data/utils.py
```diff
@@ -432,6 +432,7 @@ def collate_meta_tensor_fn(batch, *, collate_fn_map=None):
     collated.meta = default_collate(meta_dicts)
     collated.applied_operations = [i.applied_operations or TraceKeys.NONE for i in batch]
     collated.is_batch = True
+    collated.spatial_ndim = getattr(batch[0], "spatial_ndim", 3)  # assumes uniform spatial_ndim
```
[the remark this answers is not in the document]
aymuos15: This would crash much before if that was the case, Unnecessary I feel

monai/transforms/lazy/functional.py
```diff
@@ -256,9 +256,11 @@ def apply_pending(data: torch.Tensor | MetaTensor, pending: list | None = None,
     if not pending:
         return data, []
 
+    _rank = data.spatial_ndim if isinstance(data, MetaTensor) else 3
+
     cumulative_xform = affine_from_pending(pending[0])
-    if cumulative_xform.shape[0] == 3:
-        cumulative_xform = to_affine_nd(3, cumulative_xform)
+    if cumulative_xform.shape[0] < _rank + 1:
+        cumulative_xform = to_affine_nd(_rank, cumulative_xform)
```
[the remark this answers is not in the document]
aymuos15: This is logically wrong, previous code has that to 3 for a reason.

monai/transforms/utility/array.py
```diff
@@ -314,23 +314,28 @@ def __call__(self, img: torch.Tensor) -> list[torch.Tensor]:
         """
         Apply the transform to `img`.
         """
-        n_out = img.shape[self.dim]
+        dim = self.dim if self.dim >= 0 else self.dim + img.ndim
+        n_out = img.shape[dim]
```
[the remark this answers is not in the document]
aymuos15: `torch.split` already raises a good error

monai/transforms/utility/array.py
```diff
@@ -314,23 +314,28 @@ def __call__(self, img: torch.Tensor) -> list[torch.Tensor]:
...
-                ndim = len(item.affine)
-                shift = torch.eye(ndim, device=item.affine.device, dtype=item.affine.dtype)
-                shift[self.dim - 1, -1] = idx
-                item.affine = item.affine @ shift
+                ndim = len(out.affine)
+                shift = torch.eye(ndim, device=out.affine.device, dtype=out.affine.dtype)
+                shift[dim - 1, -1] = idx
+                out.affine = out.affine @ shift
+                if not self.keepdim:
+                    out.spatial_ndim = max(1, out.spatial_ndim - 1)
```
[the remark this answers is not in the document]
aymuos15: That scenario will not come up unless a major logic change in the affine updates

aymuos15:
Really do apologise for so many triggers. Was a bit confused.

atbenmurray:
I'd like to take a look at this one @ericspod, @KumoLiu, @Nic-Ma

aymuos15:
Thanks @atbenmurray 

I am still kinda wrapping my head around things over here to make sure there arent any edge cases I've missed. For now I do not think I have but will continue checking.

monai/data/meta_tensor.py
```diff
@@ -467,15 +519,40 @@ def affine(self) -> torch.Tensor:
...
+
+        This value is cached for hot-path performance and is kept in sync with the affine matrix
+        via the affine setter. The affine matrix is the source of truth for spatial dimensions.
+        """
+        return getattr(self, "_spatial_ndim", _DEFAULT_SPATIAL_NDIM)
+
+    @spatial_ndim.setter
+    def spatial_ndim(self, val: int) -> None:
+        """Set the number of spatial dimensions."""
+        if val < 1:
```
atbenmurray, line 549:
```suggestion
        if not isinstance(val, Integral):
            raise TypeError(f"'val' must be an numbers.Integral type; got {type(val)}.")
        if val < 1:
```

monai/data/meta_tensor.py
```diff
@@ -13,22 +13,59 @@
 
 import functools
 import warnings
-from collections.abc import Sequence
+from collections.abc import Mapping, Sequence
 from copy import deepcopy
 from typing import Any
```
atbenmurray, line 19:
```suggestion
from typing import Any
from numbers import Integral
```

monai/data/meta_tensor.py
```diff
@@ -467,15 +519,40 @@ def affine(self) -> torch.Tensor:
...
+        via the affine setter. The affine matrix is the source of truth for spatial dimensions.
+        """
+        return getattr(self, "_spatial_ndim", _DEFAULT_SPATIAL_NDIM)
+
+    @spatial_ndim.setter
+    def spatial_ndim(self, val: int) -> None:
+        """Set the number of spatial dimensions."""
+        if val < 1:
+            raise ValueError(f"spatial_ndim must be >= 1, got {val}")
+        self._spatial_ndim = val
```
atbenmurray:
```suggestion
        self._spatial_ndim = int(val)
```

atbenmurray (changes requested):

aymuos15:
Thank you very much @atbenmurray Have applied those changes.

aymuos15:
image

atbenmurray (dismissed):
Sorry, I thought I had already submitted these

monai/transforms/utility/array.py
```diff
@@ -314,23 +314,26 @@ def __call__(self, img: torch.Tensor) -> list[torch.Tensor]:
...
+                outputs[idx] = item.squeeze(dim)
             if self.update_meta and isinstance(img, MetaTensor):
-                if not isinstance(item, MetaTensor):
-                    item = MetaTensor(item, meta=img.meta)
-                if self.dim == 0:  # don't update affine if channel dim
+                out = outputs[idx]
+                if not isinstance(out, MetaTensor):
+                    out = MetaTensor(out, meta=img.meta)
+                    outputs[idx] = out
+                if dim == 0:  # don't update affine if channel dim
```
garciadias, line 331: **Stale `spatial_ndim` after `SplitDim` on channel dim with `keepdim=False`** (Major)

When `dim==0` (channel dim), the affine update is skipped (`if dim == 0: continue`), so the output tensor's `_spatial_ndim` remains unchanged. After `SplitDim(dim=0, keepdim=False)` on a `(3, 8, 7)` tensor, each output is `(8, 7)` — in channel-first format this has 1 spatial dim, but `spatial_ndim` stays at 2.

The test at `tests/transforms/utility/test_splitdim.py:72` asserts `spatial_ndim == 2`, which reflects the stale cached value rather than the correct spatial rank.

**Suggested fix:** After splitting on channel dim with `keepdim=False`, recalculate `spatial_ndim` via `_normalize_spatial_ndim(self.spatial_ndim, out.ndim)`. Update the test to expect `spatial_ndim == 1`.

monai/transforms/spatial/array.py
```diff
@@ -848,12 +848,14 @@ def __call__(
...
 
-        input_ndim = img.ndim - 1  # spatial ndim
+        input_ndim = get_spatial_ndim(img)
         if self.size_mode == "all":
             output_ndim = len(ensure_tuple(self.spatial_size))
             if output_ndim > input_ndim:
                 input_shape = ensure_tuple_size(img.shape, output_ndim + 1, 1)
                 img = img.reshape(input_shape)
+                if isinstance(img, MetaTensor):
+                    img.spatial_ndim = output_ndim
```
garciadias, line 860: **`Resize` with `size_mode="all"` mutates input tensor's `spatial_ndim`** (Major)

When `output_ndim > input_ndim`, the code does `img.spatial_ndim = output_ndim` directly on the input `img` (line 858). This mutates the caller's tensor, which is surprising — MONAI transforms generally avoid modifying inputs in place.

Even though the `.reshape()` on the previous line may copy data, the `spatial_ndim` write is an independent side-effect that persists on the original tensor.

**Suggested fix:** Set `spatial_ndim` on the reshaped result rather than the input tensor:
```python
img = img.reshape(input_shape)
if isinstance(img, MetaTensor):
    img.spatial_ndim = output_ndim  # now on the new tensor after reshape
```
(If `.reshape()` returns `self` when shape doesn't change, `.clone()` or explicit reconstruction may be safer.)

monai/data/meta_tensor.py
```diff
@@ -13,22 +13,60 @@
...
 import torch
 
 import monai
-from monai.config.type_definitions import NdarrayTensor
-from monai.data.meta_obj import MetaObj, get_track_meta
-from monai.data.utils import affine_to_spacing, decollate_batch, list_data_collate, remove_extra_metadata
+from monai.config.type_definitions import NdarrayOrTensor, NdarrayTensor
+from monai.data.meta_obj import _DEFAULT_SPATIAL_NDIM, MetaObj, get_track_meta
+from monai.data.utils import affine_to_spacing, decollate_batch, is_no_channel, list_data_collate, remove_extra_metadata
 from monai.utils import look_up_option
```
garciadias, line 28: **`_normalize_spatial_ndim` floor at 1 silently masks 0-spatial-dim data** (Major)

```python
limit = max(int(tensor_ndim), 1) if no_channel else max(int(tensor_ndim) - 1, 1)
return max(1, min(int(spatial_ndim), limit))
```

The `max(1, ...)` ensures `spatial_ndim >= 1` always. For a 1D tensor `(10,)` with `ndim=1`, this gives `limit = max(0, 1) = 1`, so `spatial_ndim` is clamped to 1 — but a 1D channel-first tensor has 0 spatial dimensions.

If 1D (non-spatial) data is not a supported use case, `_normalize_spatial_ndim` should raise an explicit error rather than silently returning 1. If 1D should be supported, `spatial_ndim=0` needs to be allowed and callers audited.

**Suggested fix:** Either:
a) Add `if tensor_ndim < 2: raise ValueError(...)` to reject sub-2D tensors with a clear message, or
b) Allow `spatial_ndim=0` by removing the `max(1, ...)` floor and handling zero in all callers, with appropriate documentation on what `spatial_ndim=0` means.

monai/data/utils.py
```diff
@@ -432,6 +432,7 @@ def collate_meta_tensor_fn(batch, *, collate_fn_map=None):
     collated.meta = default_collate(meta_dicts)
     collated.applied_operations = [i.applied_operations or TraceKeys.NONE for i in batch]
     collated.is_batch = True
+    collated.spatial_ndim = min(getattr(batch[0], "spatial_ndim", _DEFAULT_SPATIAL_NDIM), max(collated.ndim - 1, 1))
```
garciadias: **`collate_meta_tensor_fn` only inspects `batch[0]` for `spatial_ndim`** (Minor)

```python
collated.spatial_ndim = min(getattr(batch[0], "spatial_ndim", _DEFAULT_SPATIAL_NDIM), max(collated.ndim - 1, 1))
```

This takes `batch[0]`'s value as representative of the whole batch. If batch items have different `spatial_ndim` values (e.g., mixed 2D/3D data in the batch), the collated result may be incorrect.

**Suggested fix:** Use a comprehension to find the minimum across all items:
```python
collated.spatial_ndim = min(
    min(getattr(t, "spatial_ndim", _DEFAULT_SPATIAL_NDIM) for t in batch),
    max(collated.ndim - 1, 1)
)
```
This ensures the batch spatial rank is never larger than any individual item's rank.

monai/transforms/spatial/array.py
aymuos15: `reshape` returns a new tensor, so the `spatial_ndim` write doesn't affect the caller right?

monai/data/meta_tensor.py
aymuos15: 1D channel-first isn't a supported shape elsewhere in MONAI. Do we need this?

aymuos15:
@garciadias Thank you so much for the review. Have adressed a few and add a comments back on the ones I felt did not need a change.

monai/data/meta_tensor.py
garciadias: My point is that if it is not supported, it should raise an error, otherwise it will fail in an unexpected way and may be difficult to debug.

monai/transforms/spatial/array.py
garciadias: Good point — you're totally right actually 👍
reshape() does return a new MetaTensor object, and _spatial_ndim is just a Python attribute on the instance, not part of the tensor storage. So even if the underlying data happens to be shared (which is unlikely here since we're adding a dimension), writing spatial_ndim on the result has no effect on the caller's original tensor. I was a bit confused thinking what would happen when it returned a view  — my bad!

monai/data/meta_tensor.py
```diff
@@ -13,22 +13,60 @@
...
 from monai.utils.type_conversion import convert_data_type, convert_to_dst_type, convert_to_numpy, convert_to_tensor
 
-__all__ = ["MetaTensor"]
+__all__ = ["MetaTensor", "get_spatial_ndim"]
+
+
+def _normalize_spatial_ndim(spatial_ndim: int, tensor_ndim: int, no_channel: bool = False) -> int:
+    """Clamp spatial dims to a valid range for the current tensor shape."""
+    limit = max(int(tensor_ndim), 1) if no_channel else max(int(tensor_ndim) - 1, 1)
+    return max(1, min(int(spatial_ndim), limit))
```
garciadias, line 38: This has not been fixed. I will not hold the PR over this, but if you need to do any more commits to solve the CI, please add this one-line change. Sorry, I commented in the wrong line before.
```python                                                                                                                            
   def _normalize_spatial_ndim(spatial_ndim: int, tensor_ndim: int, no_channel: bool = False) -> int:                                 
       """Clamp spatial dims to a valid range for the current tensor shape."""                                                        
       if tensor_ndim < 2 and not no_channel:                                                                                         
           raise ValueError(                                                                                                          
               f"Channel-first tensor must have at least 2 dims (channel + 1 spatial), got ndim={tensor_ndim}"                        
           )                                                                                                                          
       limit = max(int(tensor_ndim), 1) if no_channel else max(int(tensor_ndim) - 1, 1)                                               
       return max(1, min(int(spatial_ndim), limit))                                                                                   
 ```

**Review by garciadias — DISMISSED** [review]
Thank you, @aymuos15! I have left a small comment, but the PR is good to go.

File: monai/data/meta_tensor.py (hunk 12 above)
Comment (aymuos15) [reply monai/data/meta_tensor.py:38]: Thanks! I have added this in.

Comment (aymuos15) [reply monai/data/meta_tensor.py:38]: So I had a deeper look into this and this basically breaks: `tests/utils/type_conversion/test_convert_data_type.py` ? Specifically the `MetaTensor(np.array(1.0))` (0-dim) and `MetaTensor([1.0,1.0]) (1-dim)` values in the test?

**Comment by ericspod** [comment]
Don't worry about the CodeQL fails, that's something I've disabled again until I can get it working.

File: monai/data/meta_tensor.py (hunk 12 above)
Comment (garciadias) [reply monai/data/meta_tensor.py:38]: Sorry, I missed that case. I was investigating, and I think the real issue is that `_normalize_spatial_ndim` gets called unconditionally in `__init__` for every MetaTensor, including ones that are never used spatially. So I think the fix belongs at the call site rather than inside `_normalize_spatial_ndim itself`. Something like:

```python
  # Sub-2D tensors (scalars, 1D) are not spatial volumes; skip normalization and keep the default.
  no_channel = _has_explicit_no_channel(self.meta)
  min_ndim = 1 if no_channel else 2
  if self.ndim >= min_ndim:
      if spatial_ndim is not None:
          self.spatial_ndim = _normalize_spatial_ndim(spatial_ndim, self.ndim, no_channel=no_channel)
      elif self.affine.ndim == 2:
          self.spatial_ndim = _normalize_spatial_ndim(self.affine.shape[-1] - 1, self.ndim, no_channel=no_channel)
```
Sub-2D tensors skip the call entirely and keep _DEFAULT_SPATIAL_NDIM as a placeholder — which is fine since no spatial op will ever read their spatial_ndim. Spatial tensors (ndim >= 2) behave exactly as before.

I tested this: 136 type-conversion tests + 119 MetaTensor tests all pass, including the 18 new `test_spatial_ndim.py` tests.

Again, sorry for my confusion.

garciadias (approved):

ericspod (approved):

PR #8772, opened 2026-03-11
Enhance ExtractDataKeyFromMetaKeyd to work with MetaTensor

### LINKED ISSUE #7562: Enhance `ExtractDataKeyFromMetaKeyd` to work with ` MetaTensor`

haoyu-haoyu:
### Description

Enhances `ExtractDataKeyFromMetaKeyd` to support extracting metadata from `MetaTensor` objects, in addition to plain metadata dictionaries.

**Before:** Only worked with metadata dictionaries (`image_only=False`):
```python
li = LoadImaged(image_only=False)
dat = li({"image": "image.nii"})  # {"image": tensor, "image_meta_dict": dict}
e = ExtractDataKeyFromMetaKeyd("filename_or_obj", meta_key="image_meta_dict")
dat = e(dat)  # extracts from dict
```

**After:** Also works with MetaTensor (`image_only=True`, the default):
```python
li = LoadImaged()  # image_only=True by default
dat = li({"image": "image.nii"})  # {"image": MetaTensor}
e = ExtractDataKeyFromMetaKeyd("filename_or_obj", meta_key="image")
dat = e(dat)  # extracts from MetaTensor.meta
assert dat["image"].meta["filename_or_obj"] == dat["filename_or_obj"]
```

### Changes

1. **`monai/apps/reconstruction/transforms/dictionary.py`**:
   - Added `MetaTensor` import
   - Modified `ExtractDataKeyFromMetaKeyd.__call__()` to detect if `meta_key` references a `MetaTensor` and extract from its `.meta` attribute
   - Updated docstring with both usage modes and examples

2. **`tests/apps/reconstruction/transforms/test_extract_data_key_from_meta_keyd.py`** (new):
   - 8 test cases covering: dict extraction, MetaTensor extraction, multiple keys, missing keys (with/without `allow_missing_keys`), and data preservation

### Testing

- [x] New unit tests for both dict-based and MetaTensor-based extraction
- [x] Tests for edge cases (missing keys, allow_missing_keys)
- [x] Backward compatible — existing dict-based usage unchanged

Signed-off-by: haoyu-haoyu <haoyu-haoyu@users.noreply.github.com>

### DIFF
--- a/monai/apps/reconstruction/transforms/dictionary.py
+++ b/monai/apps/reconstruction/transforms/dictionary.py
@@ -20,6 +20,7 @@ from torch import Tensor
 from monai.apps.reconstruction.transforms.array import EquispacedKspaceMask, RandomKspaceMask
 from monai.config import DtypeLike, KeysCollection
 from monai.config.type_definitions import NdarrayOrTensor
+from monai.data.meta_tensor import MetaTensor
 from monai.transforms import InvertibleTransform
 from monai.transforms.croppad.array import SpatialCrop
 from monai.transforms.intensity.array import NormalizeIntensity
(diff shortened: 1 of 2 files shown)


aymuos15:
Would this not be a cleaner approach where you add the keys into the actual meta tensor itself: [commit](https://github.com/aymuos15/MONAI/commit/74c3e7bb277c0ed83e4c3b6ed7f33a554cf17ab7)

haoyu-haoyu:
Thanks @aymuos15 — that's a cleaner approach! Adding `__contains__` and `__getitem__` delegation on `MetaTensor` itself means **all** transforms that access metadata via string keys would work transparently, not just `ExtractDataKeyFromMetaKeyd`.

The trade-off is that it touches the core `MetaTensor` class (wider blast radius), whereas this PR's approach is scoped to a single transform. But for long-term maintainability your approach is better since it avoids per-transform `isinstance(meta_obj, MetaTensor)` checks everywhere.

Happy to defer to the maintainers on which direction to go — or I can update this PR to adopt your approach if preferred.

haoyu-haoyu:
Updated — adopted @aymuos15's approach:

**Changes in this update:**
1. Added `__contains__` and `__getitem__` to `MetaTensor` (`monai/data/meta_tensor.py`) so string keys delegate to `.meta` automatically
2. Simplified `ExtractDataKeyFromMetaKeyd.__call__` — removed the `isinstance(meta_obj, MetaTensor)` branch; now uses `key in meta_obj` / `meta_obj[key]` directly, which works for both `MetaTensor` and plain dict
3. Removed the unused `MetaTensor` import from `dictionary.py`

Credit to @aymuos15 for the cleaner design.

haoyu-haoyu:
Reverted back to the `isinstance` approach (Approach A). 

Adding `__getitem__` to `MetaTensor` causes infinite recursion in `torch.as_tensor()` — PyTorch internally calls `__len__` and `__getitem__` on inputs, which triggers `__torch_function__` dispatch, creating a cycle:

```
torch.as_tensor() → __len__ → __torch_function__ → torch.as_tensor() → ...
```

The `__contains__`/`__getitem__` delegation idea is elegant but would need deeper integration with `__torch_function__` to avoid breaking tensor operations. That's a larger refactor best done as a separate PR with thorough tensor-operation test coverage.

For now, the scoped `isinstance` check in `ExtractDataKeyFromMetaKeyd` is the safe fix.

aymuos15:
Ah my bad, should've run the CI once. 

I do not think the `__getitem__` approach causes infinite recursion. The actual CI failure is `TypeError: len() of a 0-d tensor`.

It was making the `torch.as_tensor` treat list items as sequences instead of tensors. I've made another fix to this which I still think is more appropriate. But I think its best for a maintainer to decide which is better. Thanks for pointing it out!

https://github.com/aymuos15/MONAI/commit/c8312e1f

haoyu-haoyu:
Thanks for the correction @aymuos15 — you're right, it's `TypeError: len() of a 0-d tensor`, not infinite recursion. I misread the traceback. The root cause is that defining `__getitem__` makes `torch.as_tensor` treat `MetaTensor` instances inside lists as generic sequences instead of tensors.

Your fix in c8312e1 (stripping MetaTensor wrappers before `torch.as_tensor`) is a clean solution. Agreed — let's leave it to the maintainers to decide which approach is better. Happy to update this PR either way.

ericspod (approved):
Looks good to me, thanks!

PR #8773, opened 2026-03-12
Fix #8340: Remove incorrect minus sign in rotate_range documentation

### LINKED ISSUE #8340: Doc for `rotate_range` indicates the wrong sign
Hello,

I think there is an extra minus sign in the doc for the first value of the `rotate_range` parameter.

https://github.com/Project-MONAI/MONAI/blob/e538f7fd9e55c0fd0e67a4d4af2f69c6eb5b947e/monai/transforms/spatial/array.py#L1809

The method `_get_rand_param` uses the range without any minus sign added to the first argument.

https://github.com/Project-MONAI/MONAI/blob/e538f7fd9e55c0fd0e67a4d4af2f69c6eb5b947e/monai/transforms/spatial/array.py#L1864

If you agree that this is just a documentation issue, I'd be happy to search for other transform having the same issue (RandAffine and RandAffined seem to have the same issue) and create a PR to remove the extra signs.

Best

yunhaoli24:
## Summary
- Fixed incorrect minus sign in `rotate_range` parameter documentation
- The documentation incorrectly showed `uniform[-rotate_range[i][0], rotate_range[i][1])` 
- The actual implementation uses `uniform(f[0], f[1])` without negation on the first parameter
- Fixed across 7 locations in both array and dictionary transforms

## Changes
- Updated documentation in `monai/transforms/spatial/array.py`:
  - RandAffineGrid (line 1851)
  - RandAffine (line 2389)
  - Rand2DElastic (line 2659)
  - Rand3DElastic (line 2827)
- Updated documentation in `monai/transforms/spatial/dictionary.py`:
  - RandAffined (line 1064)
  - Rand2DElasticd (line 1249)
  - Rand3DElasticd (line 1399)

## Types of changes
- [x] Documentation update

Fixes #8340

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -1848,7 +1848,7 @@ class RandAffineGrid(Randomizable, LazyTransform):
         """
         Args:
             rotate_range: angle range in radians. If element `i` is a pair of (min, max) values, then
-                `uniform[-rotate_range[i][0], rotate_range[i][1])` will be used to generate the rotation parameter
+                `uniform[rotate_range[i][0], rotate_range[i][1])` will be used to generate the rotation parameter
                 for the `i`th spatial dimension. If not, `uniform[-rotate_range[i], rotate_range[i])` will be used.
                 This can be altered on a per-dimension basis. E.g., `((0,3), 1, ...)`: for dim0, rotation will be
                 in range `[0, 3]`, and for dim1 `[-1, 1]` will be used. Setting a single value will use `[-x, x]`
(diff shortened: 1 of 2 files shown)


ericspod:
Hi @yunhaoli24 thanks for the update. Please check that the line below many of your changes should be updated too, eg. `If not, uniform[-rotate_range[i], rotate_range[i])` which I think should not have the negation either

yunhaoli24:
Hi @ericspod, thanks for the review!

I've checked the second line and the negative sign there is actually correct. Looking at the implementation in `_get_rand_param` (line 1908):

```python
elif f is not None:
    out_param.append(self.R.uniform(-f, f))  # single value case
```

When `rotate_range` is a single value (e.g., `rotate_range=0.3`), the code uses `uniform(-f, f)`, so the negative sign in the second line is correct and should be kept.

The first line's negative sign was wrong because it describes the pair case:
```python
if issequenceiterable(f):
    out_param.append(self.R.uniform(f[0], f[1]))  # pair case - no negation
```

My fix only removed the incorrect negative sign from the first line (pair case), while keeping the correct negative sign in the second line (single value case).

ericspod (approved):

PR #8774, opened 2026-03-13
Add parameter to DiceMetric and DiceHelper classes

### LINKED ISSUE #8733: Feature Request: Evaluation of Semantic Segmentation Metrics on a per-component basis

Current MONAI segmentation metrics aggregate scores globally. In multi-instance medical segmentation tasks this can mask clinically relevant failure modes, e.g. missing small lesions while achieving high Dice scores due to dominant large structures. There is no native support for connected component aware evaluation except Panoptic Quality that treats each ground truth object as an individual evaluation unit.

Add connected component aware variants of existing MONAI semantic segmentation metrics. The proposed CC (Connected-Component-wise) Metrics operate as thin wrappers around existing MONAI metrics and:

- Decompose the ground truth foreground into connected components
- Assign each voxel to its nearest ground truth component via a Voronoi partition based on a single Euclidean distance/feature transform.
- Evaluate standard MONAI metrics inside each component specific region of interest.
- Supports aggregation modes (patient: mean over components per case &  overall: flat aggregation over all components)
- Integrate seamlessly with existing MONAI metrics such as DiceMetric, HausdorffDistanceMetric, SurfaceDistanceMetric, and SurfaceDiceMetric.

The implementation may be binary segmentation only first but coule be extended to multi-class approaches.

- Instance segmentation metrics. These require different model outputs and are not applicable to semantic segmentation pipelines. 
- Panoptic Quality offers a part of the solution but relies on a fixed and threshold dependant overlap-based matching and only consideres IoU as a metric.

**Additional context**

Reference implementation exists as a standalone library CC Metrics (https://github.com/alexanderjaus/CC-Metrics), published with peer reviewed validation.

The approach is described in “Every Component Counts: Rethinking the Measure of Success for Medical Semantic Segmentation in Multi Instance Segmentation Tasks”, AAAI 2025.

VijayVignesh1:
Fixes #8733 

### Description

This PR adds support for connected component-based Dice metric calculation to the existing DiceMetric and DiceHelper classes.

### Changes
* Added per_component: bool = False to both DiceMetric and DiceHelper constructors
* Implemented compute_cc_dice method that calculates Dice scores for each connected component individually
* Voronoi regions: Added compute_voronoi_regions_fast method for efficient connected component assignment without external cc3d dependency
* Added input shape validation requiring 5D binary segmentation with 2 channels (background + foreground) when per_component=True
* Updated first_ch calculation to properly exclude background channel when using per-component mode

### Reference
* https://arxiv.org/abs/2410.18684
* https://github.com/alexanderjaus/CC-Metrics 

### Types of changes


monai/metrics/meandice.py
```diff
@@ -322,15 +418,25 @@ def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor | tupl
...
 
-        first_ch = 0 if self.include_background else 1
+        if self.per_component:
+            if len(y_pred.shape) != 5 or len(y.shape) != 5 or y_pred.shape[1] != 2 or y.shape[1] != 2:
+                raise ValueError(
+                    "per_component requires both y_pred and y to be 5D binary segmentations "
+                    f"with 2 channels. Got y_pred={tuple(y_pred.shape)}, y={tuple(y.shape)}."
+                )
+
+        first_ch = 0 if self.include_background and not self.per_component else 1
```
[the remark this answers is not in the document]
VijayVignesh1: This is the expected behavior.

aymuos15:
Would be nice to extend this to both 2d and 3d within this PR itself.

aymuos15:
MONAI also entertains cupy, e.g: `https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/image_reader.py` so having the option to do the cc step through that would be cleaner in gpus.

VijayVignesh1:
> MONAI also entertains cupy, e.g: `https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/image_reader.py` so having the option to do the cc step through that would be cleaner in gpus.

I've added both 2d capabilities as well as cupy option.

monai/metrics/meandice.py
```diff
@@ -277,6 +304,117 @@ def __init__(
...
+        """
+        if isinstance(labels, torch.Tensor) and labels.is_cuda and has_cupy and has_cupy_ndimage:
+            xp = cupy
+            nd_distance_transform_edt = cupy_ndimage.distance_transform_edt
+            nd_generate_binary_structure = cupy_ndimage.generate_binary_structure
+            nd_label = cupy_ndimage.label
+            x = cupy.asarray(labels.detach())
+        else:
+            xp = np
+            nd_distance_transform_edt = scipy_ndimage.distance_transform_edt
```
ericspod: Note that we have a [distance_transform_edt](https://github.com/Project-MONAI/MONAI/blob/a8176f1530ef813d7585bab04fc7ae46bbefd871/monai/transforms/utils.py#L2416) function in MONAI, we should probably update this later to use cupyx if cucim isn't present.

monai/metrics/meandice.py
```diff
@@ -277,6 +304,117 @@ def __init__(
...
+        rank = x.ndim
+        if rank == 3:
+            conn_map = {6: 1, 18: 2, 26: 3}
+            connectivity = 26
+        elif rank == 2:
+            conn_map = {4: 1, 8: 2}
+            connectivity = 8
+        else:
+            raise ValueError("Only 2D or 3D inputs supported")
+        conn_rank = conn_map.get(connectivity, max(conn_map.values()))
```
ericspod: I'm not following the logic in this block. The value for `connectivity`  will always be 26 or 8 and then the value for `conn_rank` will always be 3 or 2, ie. the same value as `rank`.

monai/metrics/meandice.py
```diff
@@ -277,6 +304,117 @@ def __init__(
...
+            dice_scores = torch.where(
+                denom > 0, (2 * tp).float() / denom.float(), torch.tensor(1.0, device=denom.device)
+            )
+            data.append(dice_scores.unsqueeze(-1))
+            data = [
+                torch.where(torch.isinf(x), torch.tensor(0.0, dtype=torch.float32, device=x.device), x) for x in data
+            ]
+            data = [
+                torch.where(torch.isnan(x), torch.tensor(0.0, dtype=torch.float32, device=x.device), x) for x in data
+            ]
```
ericspod: I would recommend using [nan_to_num](https://docs.pytorch.org/docs/2.8/generated/torch.nan_to_num.html) if this works for what you're wanting to achieve here.

monai/metrics/meandice.py
```diff
@@ -322,15 +463,33 @@ def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor | tupl
...
+            if self.per_component:
+                data.append(self.compute_cc_dice(y_pred=y_pred[b].unsqueeze(0), y=y[b].unsqueeze(0)).reshape(-1))
+                continue
             c_list = []
             for c in range(first_ch, n_pred_ch) if n_pred_ch > 1 else [1]:
                 x_pred = (y_pred[b, 0] == c) if (y_pred.shape[1] == 1) else y_pred[b, c].bool()
                 x = (y[b, 0] == c) if (y.shape[1] == 1) else y[b, c]
                 c_list.append(self.compute_channel(x_pred, x))
+            # if self.per_component:
+            #     c_list = [self.compute_cc_dice(y_pred=y_pred[b].unsqueeze(0), y=y[b].unsqueeze(0))]
```
ericspod:
```suggestion
            else:
                c_list = []
                for c in range(first_ch, n_pred_ch) if n_pred_ch > 1 else [1]:
                    x_pred = (y_pred[b, 0] == c) if (y_pred.shape[1] == 1) else y_pred[b, c].bool()
                    x = (y[b, 0] == c) if (y.shape[1] == 1) else y[b, c]
                    c_list.append(self.compute_channel(x_pred, x))
```

monai/metrics/meandice.py
```diff
@@ -277,6 +304,117 @@ def __init__(
         self.activate = activate
         self.ignore_empty = ignore_empty
         self.num_classes = num_classes
+        self.per_component = per_component
+
+    def compute_voronoi_regions_fast(self, labels):
```
ericspod: This method doesn't depend on the state of `self` so should be a function defined external to this class.

monai/metrics/meandice.py
```diff
@@ -277,6 +304,117 @@ def __init__(
...
+        edt_input = xp.ones(cc.shape, dtype=xp.uint8)
+        edt_input[cc > 0] = 0
+        indices = nd_distance_transform_edt(edt_input, sampling=None, return_distances=False, return_indices=True)
+        voronoi = cc[tuple(indices)]
+        if xp is cupy:
+            return torch.as_tensor(cupy.asnumpy(voronoi), dtype=torch.int32)
+        else:
+            return torch.as_tensor(voronoi, dtype=torch.int32)
+
+    def compute_cc_dice(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
```
ericspod, line 306: Similar to the above, `self` is used to access `ignore_empty` which could be passed as a argument instead with this method turned into a function external to this class.

monai/metrics/meandice.py
```diff
@@ -277,6 +304,117 @@ def __init__(
...
+        Compute per-component Dice for a single batch item.
+
+        Args:
+            y_pred (torch.Tensor): Predictions with shape (1, 2, D, H, W) or (1, 2, H, W).
+            y (torch.Tensor): Ground truth with shape (1, 2, D, H, W) or (1, 2, H, W).
+
+        Returns:
+            torch.Tensor: Mean Dice over connected components.
+        """
+        data = []
```
ericspod: In this method `data` only ever gets one item placed into it?

ericspod (changes requested):
Hi @VijayVignesh1 thanks for this contribution. I've had a look through with a few comments, I think the added methods need to be reworked as functions and their internal logic looked at again. We can then rereview things.

monai/metrics/meandice.py
VijayVignesh1: MONAI's distance_transform_edt gives different results that scipy's and cupy's, as given in the docstring.
```
Input gets passed channel-wise to the distance-transform, thus results from this function will differ from directly calling ``distance_transform_edt()`` in CuPy or SciPy.
```

monai/metrics/meandice.py
VijayVignesh1: So, the connectivity can be 1 dimensional, 2 dimensional or 3 dimensional as given in conn_map. The deafult is 26 for 3D input and 8 for 2D input. I could add it as a default parameter to the function, for the future. Would that work?

monai/metrics/meandice.py
VijayVignesh1: Yes, the dice_scores if all the connected components.

monai/metrics/meandice.py
VijayVignesh1: It seems that the compute_channel function, which is similar to compute_cc_dice, resides in the same class. To maintain consistency and encapsulate related functionality, it would make sense to keep both functions within the same class, right?

monai/metrics/meandice.py
VijayVignesh1: I've removed data = [], and added direct assignment.

aymuos15:
https://github.com/BrainLesion/panoptica
https://github.com/alexanderjaus/CC-Metrics

There are quite a few nuances to something like this. After having a thought, I would be in the ground of not porting this directly on dice and having it seperate (calling it cc-metrics maybe?)

This is more akin to panoptic quality and that has a standalone func.

VijayVignesh1:
> https://github.com/BrainLesion/panoptica https://github.com/alexanderjaus/CC-Metrics
> 
> There are quite a few nuances to something like this. After having a thought, I would be in the ground of not porting this directly on dice and having it seperate (calling it cc-metrics maybe?)
> 
> This is more akin to panoptic quality and that has a standalone func.

That makes sense. However, my concern is that it’s essentially a variant of the Dice Metric, and adding a separate class for it might be a bit confusing for the users. @aymuos15 @ericspod thoughts?

aymuos15:
In my opinion, it's not a variant of the dice metric. 

The goal of that paper is to "match instances" and it's more of an instance aware metric. You can't even use that metric without first running connected components on your ground truth data. Like panoptic quality.

tests/metrics/test_compute_meandice.py
```diff
@@ -301,6 +341,30 @@ def test_nans_class(self, params, input_data, expected_value):
         else:
             np.testing.assert_allclose(result.cpu().numpy(), expected_value, atol=1e-4)
 
+    # CC DiceMetric  tests
+    @parameterized.expand([TEST_CASE_16, TEST_CASE_17])
+    @unittest.skipUnless(has_ndimage, "Requires scipy.ndimage.")
+    def test_cc_dice_value_nogpu(self, params, input_data, expected_value):
+        dice_metric = DiceMetric(**params)
+        cpu_inputs = {"y": input_data["y"].cpu(), "y_pred": input_data["y_pred"].cpu()}
```
ericspod:
```suggestion
        if not has_cupy_ndimage:
            cpu_inputs = {"y": input_data["y"].cpu(), "y_pred": input_data["y_pred"].cpu()}
```
I think if we add this we can avoid having to duplicate the test below.

monai/metrics/meandice.py
```diff
@@ -41,6 +47,9 @@ class DiceMetric(CumulativeIterationMetric):
     image size they can get overwhelmed by the signal from the background. This assumes the shape of both prediction
     and ground truth is BCHW[D].
 
+    The ``per_component`` parameter can be set to `True` to compute the Dice metric per connected component in the ground truth
+    , and then average. This requires binary segmentations with 2 channels (background + foreground) as input.
```
ericspod: I think this requires more explanation to know how to use this feature, and why one would want to. The input prediction and ground truth must both have 2 channels (foreground/background) where the ground truth is a binary segmentation, so this can't be combined with multiclass segmentation. These sorts of details should be mentioned so the tensor contents are clear.

monai/metrics/meandice.py
```diff
@@ -277,6 +301,50 @@ def __init__(
         self.activate = activate
         self.ignore_empty = ignore_empty
         self.num_classes = num_classes
+        self.per_component = per_component
+
+    def compute_cc_dice(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
```
ericspod, line 315: I would have put `compute_channel` outside this class as well unless there was expectation that this might be overridden in a subclass, but that's unlikely here. I'm fine leaving this here as just a minor thing I saw.

monai/metrics/meandice.py
```diff
@@ -322,15 +393,31 @@ def __call__(self, y_pred: torch.Tensor, y: torch.Tensor) -> torch.Tensor | tupl
...
+                        "(B, 2, H, W) or (B, 2, D, H, W). "
+                        f"Got y_pred={tuple(y_pred.shape)}, y={tuple(y.shape)}."
+                    )
+
+        first_ch = 0 if self.include_background and not self.per_component else 1
         data = []
         for b in range(y_pred.shape[0]):
+            if self.per_component:
+                data.append(self.compute_cc_dice(y_pred=y_pred[b].unsqueeze(0), y=y[b].unsqueeze(0)).reshape(-1))
+                continue
```
ericspod, line 422: Stylistically I'd use `else` here instead, but no need to change it.

ericspod (changes requested):
I had some very minor comments to a few things that require adding some commentary only, other things were minor.

ericspod:
> > https://github.com/BrainLesion/panoptica https://github.com/alexanderjaus/CC-Metrics
> > There are quite a few nuances to something like this. After having a thought, I would be in the ground of not porting this directly on dice and having it seperate (calling it cc-metrics maybe?)
> > This is more akin to panoptic quality and that has a standalone func.
> 
> That makes sense. However, my concern is that it’s essentially a variant of the Dice Metric, and adding a separate class for it might be a bit confusing for the users. @aymuos15 @ericspod thoughts?

I get the argument for and against this work going into a separate class as you've discussed with @aymuos15. I'm no familiar with these metrics as much but it obstensibly looks similar to a dice metric, but if it was moved to a separate class this wouldn't hurt either. Adding this to the code hasn't overcomplicated the class definition, that would be the point that implementing separate classes would make sense. I will leave it up to you but we can come back to this in the future of course if needed. I've also marked outdated comments of mine as resolved as I think the code has been changed to not have those concerns anymore.

VijayVignesh1:
Hi @aymuos15 @ericspod 
I believe adding this functionality directly within the current Dice metric class is a good approach for now. This keeps the API simple and avoids introducing unnecessary complexity with separate classes.
I’ll go ahead and push the changes now and close this PR. If we need to revisit the structure or split it out later, we can always adjust. :)

VijayVignesh1:
Hi @ericspod 
I’ve updated the docstring with more details about the metric, 2-channel input and ground truth requirements and have left the other two comments as is.

ericspod (approved):

PR #8778, opened 2026-03-17
Fix multi-axis shear transform to compose individual shear matrices

aymuos15:
Fix Shear transform


ericspod (approved):
Thanks @aymuos15 I think this looks fine, I'd like @atbenmurray to have a look as well if possible. We need to merge a fix shortly then can come back to this one.

PR #8779, opened 2026-03-17
Add support for configurable GD-enhancing tumor label in ConvertToMul…

AbdoMis:
**Description**
This PR introduces an `et_label` parameter to `ConvertToMultiChannelBasedOnBratsClasses` to easily support the BraTS 2023 dataset conventions without breaking backwards compatibility.

**Motivation and Context**
Historically (BraTS 18-21), the Enhancing Tumor (ET) label was 4. However, starting with the BraTS 2023 Adult Glioma (GLI) dataset, the ET label was shifted to 3. Because the original transform hardcoded the ET label as 4, it fails to recognize the ET channel in newer datasets and generates empty masks.

By parameterizing `et_label` and defaulting it to 4, existing pipelines remain 100% unaffected, while BraTS 23 users can now simply pass `et_label=3` to utilize the standard sub-region extraction (TC, WT, ET).

**Types of changes**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)


monai/transforms/utility/dictionary.py
```diff
@@ -1297,19 +1297,27 @@ def __call__(self, data: Mapping[Hashable, Any]):
...
+            - Use 4 for BraTS 2018-2022.
+            - Use 3 for BraTS 2023.
+            Defaults to 4.
+        allow_missing_keys: don't raise exception if key is missing.
     """
 
     backend = ConvertToMultiChannelBasedOnBratsClasses.backend
 
-    def __init__(self, keys: KeysCollection, allow_missing_keys: bool = False):
+    def __init__(self, keys: KeysCollection, et_label: int = 4, allow_missing_keys: bool = False):
```
ericspod:
```suggestion
    def __init__(self, keys: KeysCollection, allow_missing_keys: bool = False, et_label: int = 4):
```
This ensures we don't break wherever positional arguments areused to construct the transform.

ericspod (changes requested):
Hi @AbdoMis thanks for this adaptation to the new standard. I had a minor suggestion but otherwise we're good to go. We have a fix we need to put through shortly but then can hopefully merge this soon.

AbdoMis:
@ericspod Thanks for the review and the explanation. I have applied the suggested change to the parameter order.

ericspod (approved):

Issue #7727, opened 2024-04-29
Memory leak in monai.transforms.distance_transform_edt


aaronkujawa:
**Describe the bug**
There seems to be a memory leak in monai.transforms.distance_transform_edt when cuCIM is not available. 

**To Reproduce**
cuCIM must not be available to reproduce
```
import torch
from monai.transforms import distance_transform_edt

nb_interations = 2000
for i in range(nb_interations):
    mask = torch.randint(low=0, high=2, size=(10, 10, 10,), device='cuda')
    distance_transform_edt(mask)

print(torch.cuda.memory_summary(device=None, abbreviated=True))
```

**Expected behavior**
The GPU memory footprint increases with nb_iterations. This should not happen. This does not happen when cuCIM is installed.

**Additional context**
The problem seems to be an [exception leak](https://cosmicpercolator.com/2016/01/13/exception-leaks-in-python-2-and-3/) in optional_import:

https://github.com/Project-MONAI/MONAI/blob/56508992c32e328667740e18f99b82a6a19f2c2a/monai/utils/module.py#L398-L408

where tb does not get cleared. tb holds references that prevent the mask variable from being deleted. I think this could be a problem with other optional_import calls, too.

johnzielke:
Most likely related to: https://github.com/Project-MONAI/MONAI/issues/7480

PR #8782, opened 2026-03-18
Fix memory leak in optional_import traceback handling

### LINKED ISSUE #7480: HausdorffDTLoss leads to GPU memory leak.
**Describe the bug**
Using this loss method with Trainer from transformers library (Pytorch) and YOLOv8 (Pytorch) leads to crash training shortly after start due to cuda out of memory. 
16 gb gpu memory, batch size is 1 with 128*128 image. Training crash after ~ 100 iterations.

**Environment**
 
Kaggle Notebook, python 3.10.12, last monai version from pip.

aymuos15:
## Summary
- Fix memory leak in `optional_import` where stored traceback objects retain references to entire call stacks, preventing garbage collection
- Format traceback to string immediately and clear `import_exception.__traceback__ = None`
- Hoist 3 `optional_import` calls for cucim to module level in `monai/transforms/utils.py`

Fixes #7480, #7727

## Details

**`monai/utils/module.py`:**
- Replace storing live `__traceback__` object with `traceback.format_exception()` string
- Clear `import_exception.__traceback__ = None` to break reference chain
- Embed formatted traceback in error message under "Original traceback:" section

**`monai/transforms/utils.py`:**
- Move cucim `optional_import` calls to module level (consistent with existing skimage/scipy/cupy pattern)
- Fixes incidental name shadowing in `distance_transform_edt`

## Test plan
- [x] `test_no_traceback_leak` — weakref regression test for the leak
- [x] `test_failed_import_shows_traceback_string` — verifies traceback in error message
- [x] All 12 `test_optional_import.py` tests pass
- [x] pre-commit, black, ruff, pytype, mypy — all clean

### DIFF
--- a/monai/transforms/utils.py
+++ b/monai/transforms/utils.py
@@ -85,10 +85,9 @@ ndimage, has_ndimage = optional_import("scipy.ndimage")
 cp, has_cp = optional_import("cupy")
 cp_ndarray, _ = optional_import("cupy", name="ndarray")
 exposure, has_skimage = optional_import("skimage.exposure")
-_cucim_skimage, _has_cucim_skimage = optional_import("cucim.skimage")
-_cucim_morphology_edt, _has_cucim_morphology = optional_import(
-    "cucim.core.operations.morphology", name="distance_transform_edt"
-)
+# NOTE: cucim is deliberately NOT imported at module level.
+# Module-level cucim imports caused very slow import times and other buggy behaviour.
+# Keep cucim imports inside the functions that need them.
 
 __all__ = [
     "allow_missing_keys_mode",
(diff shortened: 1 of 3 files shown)


monai/transforms/utils.py
```diff
@@ -85,6 +85,10 @@
 cp, has_cp = optional_import("cupy")
 cp_ndarray, _ = optional_import("cupy", name="ndarray")
 exposure, has_skimage = optional_import("skimage.exposure")
+_cucim_skimage, _has_cucim_skimage = optional_import("cucim.skimage")
```
ericspod: I believe we had strange import issues if we had cucim at the module level, such as very slow import or some other buggy behaviour. Our solution was to put importation into functions so that this happened only when the library was needed, which was very rarely. Please undo this change but replace it with a note that cucim is deliberately not imported at the module level, there should have been one already we neglected to add in the past.

ericspod (changes requested):
Hi @aymuos15 thanks for these fixes. I commented on the cucim change which mostly should be reverted, the changes for the traceback leaking look fine to me so I hope this solves the problems the issues have raised.

aymuos15:
Thank you very much for the review. Will get to this soon!

tests/utils/test_optional_import.py
```diff
@@ -75,6 +77,34 @@ def versioning(module, ver, a):
...
+        ref = _do_import()
+        gc.collect()
+        self.assertIsNone(ref(), "optional_import is leaking frame references via traceback")
+
+    def test_failed_import_shows_traceback_string(self):
+        """Verify the error message includes the original traceback as a string."""
+        mod, flag = optional_import("nonexistent_module_for_tb_test")
+        self.assertFalse(flag)
+        with self.assertRaises(OptionalImportError) as ctx:
+            mod.something
```
ericspod:
```suggestion
            _ = getattr(mod, "something")
```

I think this would satisfy the issue.

ericspod (approved):
The pre-commit undid the `getattr` change in the test but I think it's fine now anyway, thanks.

Issue #7565, opened 2024-03-22
Parent `Compose`'s `map_items` is passed down to a child `Compose` even though it specifies its own


ibro45:
**Is your feature request related to a problem? Please describe.**
I tried defining a `Compose` within a `Compose` with a different `map_items` argument, but the one from the parent `Compose` doesn't get overridden. 

Only once I defined the parent `Compose` to be a `torchvision` one instead of `monai`, it worked, as the `torchvision` one doesn't have a `map_items` argument to pass down.

**Describe the solution you'd like**
A nested `Compose` should override the parent `Compose`'s arguments.

**Describe alternatives you've considered**

**Additional context**

Monai Bundle Config style example of what I'm doing. In this example, I need to wrap the final list into a dict to indicate that it's an input only, and not an input and target.

```yaml
system#datasets#train#transform:
    # Parent torchvision Compose so that the monai Compose children can have different map_items.
    # A parent monai Compose would force all children to have its map_items.
    _target_: torchvision.transforms.Compose
    transforms:
        - _target_: monai.transforms.Compose
          # `map_items=True` - When input to a Transform is a list, apply the transform to each element of the list.
          map_items: True
          transforms:
              - _target_: monai.transforms.LoadImage
                image_only: True
              - _target_: monai.transforms.EnsureChannelFirst
              - _target_: monai.transforms.Orientation
                axcodes: SPL
                lazy: True
              - _target_: monai.transforms.Spacing
                pixdim: "%CONSTANTS#SPACING"
                mode: bilinear
                lazy: True
              - _target_: monai.transforms.CropForeground
                lazy: True
              - _target_: monai.transforms.SpatialPad
                spatial_size: "%CONSTANTS#PATCH_SIZE_VOXELS"
                value: -1024
                lazy: True
              - _target_: monai.transforms.ScaleIntensityRange
                a_min: -1024
                a_max: 2048
                b_min: 0
                b_max: 1
                clip: True
              # Turn into a dict of views
              - _target_: monai.transforms.Lambda
                func: '$lambda x: {"view_0": x}'
              # From this point it's not a Tensor anymore, but a list of NUM_CONTRASTIVE_CROPS Tensors
              - _target_: monai.transforms.RandSpatialCropSamplesd
                keys: view_0
                roi_size: "%CONSTANTS#PATCH_SIZE_VOXELS"
                num_samples: "%CONSTANTS#NUM_CONTRASTIVE_CROPS"
                lazy: True
[...]

PR #8787, opened 2026-03-20
Fix nested Compose map_items in forward and inverse paths

### LINKED ISSUE #7932: Option to Disable `Compose` list/tuple Expansion
[Here](https://github.com/Project-MONAI/MONAI/blob/46e2b0eb9a7c8249177df775047fa098af26f6a6/monai/transforms/lazy/functional.py#L115) `Compose` will expand any data item which is a list or tuple and apply transforms individually to each element. This is used when a transform produces multiple outputs. This isn't always wanted if one wants to pass something iterable to transform as-is. 

A constructor argument to `Compose` to disable this behaviour. 

Data needs to be wrapped as a dictionary or something not a list/tuple to route around this issue.

**Additional context**
This is using transforms for data processing rather than training so a somewhat niche use case.

aymuos15:
## Summary
Fixes #7932, #7565

When a child `Compose` has a different `map_items` setting than its parent, the parent's `apply_transform` would expand list/tuple data before the child ever sees it — silently overriding the child's `map_items`.

This PR makes three coordinated changes so the child's `map_items` is respected:

- **Forward path** (`apply_transform`): Skip list expansion when the transform is a `Compose` instance, letting it handle expansion via its own `map_items` in `execute_compose`.
- **Inverse path** (`_inverse_one` helper): Delegate directly to `Compose.inverse()` for nested `Compose` objects (including `RandomOrder` and `SomeOf`) instead of routing through `apply_transform(t.inverse, ...)`.
- **`flatten()`**: Only inline nested `Compose` objects that share the same `map_items` as the parent. Children with a different `map_items` are preserved as-is.

## Test plan
- [x] `test_child_map_items_false_receives_list` — parent `map_items=True`, child `map_items=False`: child receives list as-is
- [x] `test_inverse_respects_child_map_items` — inverse roundtrip with nested Compose
- [x] `test_parent_no_map_child_map` — parent `map_items=False`, child `map_items=True`: child maps over items
- [x] `test_flatten_preserves_different_map_items` — `flatten()` does not merge children with different `map_items`


atbenmurray:
I'd like to have a careful think about this. There are a few ways in which nested Composes in their current form have problems, and they could do with being addressed holistically. At least, I'd like to make sure it doesn't potentially affect a broader fix.

atbenmurray (changes requested):
My initial feeling is that this should just work and doesn't get in the way of any future refactors.

The change to flatten is an ancillary fix that you made while changing this, yes? Might be worth expanding the tests around this for outer and inner composes having different combinations of flags, and perhaps two internal composes, one of which has it and one of which doesn't.

aymuos15:
@atbenmurray thank you very much for pointing out the expanded tests. Have added them now.

atbenmurray (approved):
Looks good from here

PR #8794, opened 2026-03-28
Enable global coordinates in spatial crop transforms

### LINKED ISSUE #8206: Enable global coordinates in spatial crop transforms
Currently, image coordinates are passed as roi center in the spatial crop transforms. It would be nice to have the ability to pass physical coordinates as all the information required to generate the image coordinates exists in the metadata. This would allow using coordinates without worrying about things like spacing changes. 

I've got an implementation lying around in a custom package but would like if I can use just MONAI moving forward
https://github.com/AIM-Harvard/foundation-cancer-image-biomarker/blob/e447a6f69f5240b658f1a6e874bdbd4f4f45df99/fmcib/preprocessing/seed_based_crop.py#L121

emanuilo:
I ran into this while building an injury classification pipeline on MRI. My annotations come in physical coordinates and I kept having to manually convert them to voxel space whenever I changed the preprocessing spacing. Found this issue and saw the design was already agreed on, so I went ahead and implemented it.

Follows the approach proposed by @ericspod in #8206.

Resolves #8206

## Changes

- Adds `TransformPointsWorldToImaged` and `TransformPointsImageToWorldd` — thin subclasses of `ApplyTransformToPointsd` that hardcode the `invert_affine` direction
- `SpatialCropd` now accepts string dictionary keys for `roi_center`, `roi_size`, `roi_start`, and `roi_end`. When a string is passed, the actual values are read from the data dict at call time instead of at init
- This means you can pipe world-space annotations straight through coordinate conversion and into cropping without manual recalculation

```python
from monai.transforms import Compose, TransformPointsWorldToImaged, SpatialCropd

pipeline = Compose([
    TransformPointsWorldToImaged(keys="roi_start", refer_keys="image"),
    TransformPointsWorldToImaged(keys="roi_end", refer_keys="image"),
    SpatialCropd(keys="image", roi_start="roi_start", roi_end="roi_end"),
])
```

## Design notes

- When no string keys are passed, `SpatialCropd` takes the original code path — zero overhead for existing usage
- The string-key path recreates a `SpatialCrop` on each `__call__` (stored on `self.cropper`), since slice computation must be deferred until the data dict is available. `inverse()` is overridden for this path to use `pop_transform(check=False)`, since the cropper identity changes between calls — the actual crop info is read from the MetaTensor's transform stack
- Tensors from `ApplyTransformToPoints` (shape `(C, N, dims)`) get flattened and rounded to int via `torch.round` (banker's rounding) to avoid systematic bias

## Tests

- 13 tests for `TransformPointsWorldToImaged` (correctness, equivalence with base class, inverse, error cases)
- 7 tests for `TransformPointsImageToWorldd`
- 11 tests for `SpatialCropd` string-key support (start/end, center/size, mixed params, tensor shapes, float rounding, missing keys, `requires_current_data`, multi-key, inverse)
- 1 end-to-end integration test: `TransformPointsWorldToImaged` → `SpatialCropd` with world-space ROI
- All 12 existing `SpatialCropd` tests pass unchanged
- All 21 existing `ApplyTransformToPointsd` tests pass unchanged


monai/transforms/croppad/dictionary.py
```diff
@@ -450,19 +466,134 @@ def __init__(
...
-        cropper = SpatialCrop(roi_center, roi_size, roi_start, roi_end, roi_slices, lazy=lazy)
+        self._roi_center = roi_center
+        self._roi_size = roi_size
+        self._roi_start = roi_start
+        self._roi_end = roi_end
+        self._roi_slices = roi_slices
+        self._has_str_roi = any(isinstance(v, str) for v in [roi_center, roi_size, roi_start, roi_end])
+
+        if not self._has_str_roi:
+            _roi_t = Optional[Union[Sequence[int], int]]
```
ericspod, line 489: Was using this and `cast` necessary? I would expect mypy to pick up the type of `roi_*` correctly. This might be a Python version issue, you should have been able to use `Sequence[int] | int | str | None` as the value here.

monai/transforms/croppad/dictionary.py
```diff
@@ -450,19 +466,134 @@ def __init__(
...
+        if isinstance(resolved, np.ndarray):
+            resolved = torch.from_numpy(resolved)
+        if isinstance(resolved, torch.Tensor):
+            resolved = torch.round(resolved.flatten()).to(torch.int64)
+        return resolved
+
+    @property
+    def requires_current_data(self):
+        """bool: Whether this transform requires the current data dictionary to resolve ROI parameters."""
+        return self._has_str_roi
```
ericspod:
```suggestion
    def requires_current_data(self) -> bool:
        """Returns True if ROI values are derived from dictionary members, False if constant members."""
        return self._has_str_roi
```

monai/transforms/croppad/dictionary.py
```diff
@@ -450,19 +466,134 @@ def __init__(
...
+    def __call__(self, data: Mapping[Hashable, torch.Tensor], lazy: bool | None = None) -> dict[Hashable, torch.Tensor]:
+        """
+        Args:
+            data: dictionary of data items to be transformed.
+            lazy: whether to execute lazily. If ``None``, uses the instance default.
+
+        Returns:
+            Dictionary with cropped data for each key.
+        """
+        if not self._has_str_roi:
```
ericspod:
```suggestion
        if not self.requires_current_data:
```

monai/transforms/croppad/dictionary.py
```diff
@@ -450,19 +466,134 @@ def __init__(
...
+            return super().__call__(data, lazy=lazy)
+
+        d = dict(data)
+        roi_center = self._resolve_roi_param(self._roi_center, d)
+        roi_size = self._resolve_roi_param(self._roi_size, d)
+        roi_start = self._resolve_roi_param(self._roi_start, d)
+        roi_end = self._resolve_roi_param(self._roi_end, d)
+
+        lazy_ = self.lazy if lazy is None else lazy
+        self.cropper = SpatialCrop(
```
ericspod: If you're creating a cropper for each invocation this will not be thread-safe if stored in a member like this. I would just use a local `cropper` variable to avoid that as we do use threading instead of processes at times for transforms.

ericspod (changes requested):
Hi @emanuilo thanks for looking into this issue. I had some minor comments below but it looks good to me. The changes are additive so there's little worry with merging this soon. Please have a look at the comments and I'll try to keep tabs on this.

monai/transforms/croppad/dictionary.py
emanuilo: Hi @ericspod, thanks for the review!

I tried that originally. The first commit on this branch had the plain call. But mypy can't narrow str away from the runtime `any(isinstance ...)` check, so all four args fail with `arg-type` (`SpatialCrop` accepts `Sequence[int] | int | NdarrayOrTensor | None`, no str). I added the casts in a follow-up just to silence that.

Would you be OK if I drop the `cast` + `_roi_t` alias and use a single `# type: ignore[arg-type]` on the `SpatialCrop(...)` line instead? That also lets me remove the `Optional/Union/cast`imports.

monai/transforms/croppad/dictionary.py
emanuilo: Done!

monai/transforms/croppad/dictionary.py
emanuilo: Done!

monai/transforms/croppad/dictionary.py
emanuilo: Fixed!

monai/transforms/croppad/dictionary.py
ericspod: Let's stick with the casts then as it is. It's something mypy might be better at in the future, but we should try to avoid ignoring things where we can (we do that too much already).

ericspod (approved):
I think this is good now, thanks, I can set the DCO to pass if you're ok with that rather than doing the remediation commit.

PR #8796, opened 2026-03-30
Fix: raise OptionalImportError when user-specified reader package is not installed

### LINKED ISSUE #7437: Raise the exception when LoadImage has a reader specified but it is not installed

Talhax55z:
Fixes #7437

### Description
When a user specifies a reader in LoadImage that requires a package 
which is not installed, the code previously only issued a warnings.warn() 
which could easily be missed.

This change raises an OptionalImportError instead, giving the user 
a clear, visible error message telling them exactly which reader 
package needs to be installed.

### Types of changes
- [ ] Breaking change
- [ ] New tests added
- [ ] Integration tests passed
- [ ] Quick tests passed
- [ ] In-line docstrings updated
- [ ] Documentation updated

### DIFF
--- a/monai/transforms/io/array.py
+++ b/monai/transforms/io/array.py
@@ -165,6 +165,10 @@ class LoadImage(Transform):
             args: additional parameters for reader if providing a reader name.
             kwargs: additional parameters for reader if providing a reader name.
 
+        Raises:
+            OptionalImportError: if a user-specified reader requires a package
+                that is not installed or has an incompatible version.
+                
         Note:
 
             - The transform returns a MetaTensor, unless `set_track_meta(False)` has been used, in which case, a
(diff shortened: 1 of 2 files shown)


ericspod:
Hi @Talhax55z thanks for the contribution, but I think there's three other PRs working on this with similar solutions. If you refer to my most recent comment on #7437 I was coordinating with the other contributors, if there's something specific in this PR you want to add to that conversation we can keep it here or close and integrate into someone else's work.

Talhax55z:
Hi @ericspod, thanks for the heads up! 

I checked issue #7437 and I'm happy to defer to the other contributors 
if the solutions are similar. Feel free to close this PR and integrate 
any useful parts into the existing work.

Thanks again for reviewing and for the kind coordination!

ericspod:
Thanks for understanding @Talhax55z. I almost certainly commented in the issue after you'd already started work on this PR so a little bad timing there. I'll leave this PR open and when we have progress with the others we'll circle back to this one.

Issue #6888, opened 2023-08-18
consistent usages of  `np.RandomState` in the code


wyli:
1. There are some usages of np.random.* in the transforms (see below). ..

> ./monai/transforms/signal/array.py:276:np.random.choice
> ./monai/transforms/signal/array.py:357:np.random.choice
> ./monai/transforms/spatial/array.py:1214:np.random.randint
> ./monai/transforms/spatial/dictionary.py:2347:np.random.randint
> ./monai/transforms/spatial/dictionary.py:698:np.random.randint
> ./monai/transforms/utils.py:1220:np.random.random
> ./monai/transforms/utils.py:246:np.random.randint
> ./monai/transforms/utils.py:510:np.random.randint
> ./monai/transforms/utils.py:552:np.random.random
> ./monai/transforms/utils.py:607:np.random.random

2. There are some usages in other parts of the main code. Same question on what to do with those remains:

> ./monai/apps/deepedit/interaction.py:65:np.random.choice
> ./monai/apps/deepedit/transforms.py:159:np.random.choice
> ./monai/apps/deepedit/transforms.py:60:np.random.choice
> ./monai/apps/detection/transforms/dictionary.py:1305:np.random.randint
> ./monai/apps/nuclick/transforms.py:370:np.random.choice
> ./monai/apps/nuclick/transforms.py:377:np.random.choice
> ./monai/auto3dseg/analyzer.py:190:np.random.rand
> ./monai/auto3dseg/analyzer.py:191:np.random.rand
> ./monai/auto3dseg/analyzer.py:292:np.random.rand
> ./monai/auto3dseg/analyzer.py:374:np.random.rand
> ./monai/auto3dseg/analyzer.py:862:np.random.rand
> ./monai/auto3dseg/operations.py:119:np.random.rand
> ./monai/auto3dseg/operations.py:120:np.random.rand
> ./monai/auto3dseg/operations.py:121:np.random.rand
> ./monai/auto3dseg/operations.py:122:np.random.rand
> ./monai/auto3dseg/operations.py:62:np.random.rand
> ./monai/data/synthetic.py:142:np.random.random
> ./monai/data/synthetic.py:65:np.random.random
> ./monai/data/utils.py:125:np.random.randint
> ./monai/utils/misc.py:353:np.random.seed
> ./monai/visualize/utils.py:92:np.random.rand
> ./monai/visualize/utils.py:97:np.random.rand

[...]

PR #8798, opened 2026-03-31
Replace direct np.random.* calls with np.random.RandomState instances

haoyu-haoyu:
### Description

Replace direct `np.random.*` global function calls with proper `np.random.RandomState` instances for reproducibility, as requested in #6888.

### Changes

| File | Call sites | Fix |
|------|-----------|-----|
| `monai/transforms/utils.py` | 3 × `np.random.random.__self__` | → `np.random.RandomState()` |
| `monai/transforms/signal/array.py` | 2 × `np.random.choice` | → `self.R.choice` (classes already Randomizable) |
| `monai/data/synthetic.py` | 2 × `np.random.random.__self__` | → `np.random.RandomState()` |
| `monai/data/utils.py` | 1 × `np.random.randint` fallback | → `np.random.RandomState().randint` |
| `monai/utils/ordering.py` | 1 × `np.random.shuffle` | → `np.random.RandomState().shuffle` |

### Scope

This PR covers 9 functional call sites across 5 files. The remaining `np.random.*` calls in the codebase are either:
- **Docstring examples** (auto3dseg, visualize) — educational, not functional
- **`np.random.seed` in `set_determinism()`** — intentionally sets global state
- **apps/deepedit/ and apps/nuclick/** — would require adding `Randomizable` inheritance, left for a follow-up

### Checks

- [x] `Signed-off-by` included (DCO)
- [x] No new global random state usage introduced
- [x] Signal transforms use existing `self.R` from `Randomizable` base class

Ref #6888

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -82,7 +82,7 @@ jobs:
         find  /opt/hostedtoolcache/* -maxdepth 0 ! -name 'Python' -exec rm -rf {} \;
     - name: Install the dependencies
       run: |
-        python -m pip install --user --upgrade pip wheel
+        python -m pip install --user --upgrade pip wheel pybind11
         python -m pip install torch==2.5.1 torchvision==0.20.1
         cat "requirements-dev.txt"
         python -m pip install --no-build-isolation -r requirements-dev.txt
--- a/docs/source/losses.rst
+++ b/docs/source/losses.rst
@@ -98,6 +98,11 @@ Segmentation Losses
 .. autoclass:: NACLLoss
     :members:
 
+`MCCLoss`
+~~~~~~~~~
+.. autoclass:: MCCLoss
+    :members:
+
 Registration Losses
 -------------------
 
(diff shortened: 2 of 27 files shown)


haoyu-haoyu:
Thank you for the review, @coderabbitai! You're absolutely right — replacing the global RNG with unseeded `np.random.RandomState()` breaks reproducibility for callers relying on `np.random.seed()`.

I've pushed a follow-up commit that **reverts** the changes to:
- `transforms/utils.py` (3 sites)
- `data/synthetic.py` (2 sites)
- `data/utils.py` (1 site)
- `utils/ordering.py` (1 site)

These utility functions accept an optional `rand_state` parameter and intentionally fall back to the global RNG when `None` — this is the correct design for public APIs that need to respect `np.random.seed()`.

The only change **retained** is in `signal/array.py`: replacing `np.random.choice` with `self.R.choice` in `SignalRandAddSine` and `SignalRandAddSquarePulsePartial`. These classes already inherit from `Randomizable` and use `self.R` for all other random operations — the `np.random.choice` call was an inconsistency.

atharvajoshi01 (commented):
Clean fix. Using self.R (the RandomState instance) instead of np.random.choice ensures reproducibility when a seed is set. Both call sites in SignalFillEmpty and the square pulse transform had the same issue.

ericspod (approved):
Hi @haoyu-haoyu I see there was more work here that has been reverted, this seems like the right call as pointed out and the remaining changes are fine.

PR #8801, opened 2026-04-02
8800 potentially wrong device using cuda variable status in monaiauto3dseganalyzerpy

### LINKED ISSUE #8800: Potentially wrong device using_cuda variable status in monai/auto3dseg/analyzer.py
We set the label_tensor device to the device of the image_tensor here, but we already determined whether cuda is set or not (using_cuda) above by checking whether any of both tensors uses cuda. So in case we got label_tensor to use cuda and image_tensor not, then we would first set using_cuda to True, but then we check whether if label_tensor.device is != image_tensor.device which is True and therefore we set the device of label_tensor to non-cuda (e.g. cpu) which means that now both tensors use cpu while using_cuda is still wrongly set to True.

(see the detailed explaination in #8708, I just opened this issue here so that we don't forget about this since #8708 is already merged, but still got this potential issue open)

garciadias:
Fixes #8800.

### Description

Fix bug introduced by me on @benediktjohannes' PR#8708

Fixed a device synchronization bug in `LabelStats.__call__()` where the `using_cuda` was being ignored. When image and label tensors were on different devices, the code would:
1. Set `using_cuda` to True if one is on GPU
2. Ignore using_cuda and move tensors to CPU if there was a mismatch

Now `using_cuda` is calculated, and when devices don't match, both tensors are moved to CUDA (when either is on CUDA) or CPU otherwise. 

Replaced `# type: ignore` comments with proper `cast()` for type safety.

### Types of changes


monai/auto3dseg/analyzer.py
```diff
@@ -479,8 +481,15 @@ def __call__(self, data: Mapping[Hashable, MetaTensor]) -> dict[Hashable, MetaTe
...
+            # If there's a device mismatch, move both to CUDA if either is on CUDA, otherwise sync to image device
             if label_tensor.device != image_tensor.device:
-                label_tensor = label_tensor.to(image_tensor.device)  # type: ignore
+                if using_cuda:
+                    # Prefer CUDA for performance when there's a mix
+                    cuda_device = image_tensor.device if image_tensor.device.type == "cuda" else label_tensor.device
+                    image_tensor = cast(MetaTensor, image_tensor.to(cuda_device))
+                    label_tensor = cast(MetaTensor, label_tensor.to(cuda_device))
+                else:
+                    label_tensor = cast(MetaTensor, label_tensor.to(image_tensor.device))
```
benediktjohannes, line 490: Is there a reason why it is better to sync to image device instead of label device or is it just that we want to sync it and it doesn't matter to which one? Or should we add additional checks here, e.g. sync to ROCm if available if the other device was CPU?

benediktjohannes (commented):
Overall LGTM, just one question

benediktjohannes:
And thanks for addressing this!

benediktjohannes:
And we need to address lint please

monai/auto3dseg/analyzer.py
```diff
@@ -479,8 +481,15 @@ def __call__(self, data: Mapping[Hashable, MetaTensor]) -> dict[Hashable, MetaTe
...
```
benediktjohannes, line 491: And one additional question: Why is cast required here?

monai/auto3dseg/analyzer.py
garciadias: The choice is meant to sync to the image device — the image is the primary input, so aligning to it is intuitive. That said, the main goal is ensuring device synchronization. Adding ROCm or other device-type detection could be relevant, but it's beyond the scope of this fix for #8800 (CUDA/CPU mismatch), and beyond my knowledge, to be honest.

monai/auto3dseg/analyzer.py
garciadias: The cast() is necessary because .to() returns a generic Tensor, not a MetaTensor. This is actually better than the previous # type: ignore comment. It's explicit type safety that helps type checkers verify correctness.

benediktjohannes: Thats Great, thank You!

monai/auto3dseg/analyzer.py
benediktjohannes: Nice, thanks! No matter, maybe we‘ll add other device checks in later PRs!

benediktjohannes (approved):
Approving this (only Lint fix left, but in General LGTM, thanks)!

benediktjohannes:
I don‘t have permissions to approve / merge, so could you please do this @ericspod? Thanks!

ericspod (approved):
I think this looks good now, you can fix your DCO issue with a remedial commit or I just mark it passed.

benediktjohannes:
> I think this looks good now, you can fix your DCO issue with a remedial commit or I just mark it passed.

Thanks!

benediktjohannes:
Great work, keep it up! 👍

benediktjohannes:
And thank you for helping out so quick, very nice! I've worked for several other projects which took way longer for those things, so very nice that you directly addressed this, thank you!

PR #8803, opened 2026-04-02
Fix: handle precomputed crops and safe no-grad cleanup in auto3dseg

### LINKED ISSUE #8802: [BUG] ImageStats crashes when precomputed 'nda_croppeds' is provided
**Describe the bug**
`monai.auto3dseg.ImageStats.__call__()` crashes when precomputed 'nda_croppeds' is provided in the input mapping.

The current implementation only defines the local 'nda_croppeds' variable when the key is not present, but later always uses 'nda_croppeds' to populate 'cropped_shape' and intensity statistics. As a result, passing precomputed cropped foregrounds raises 'UnboundLocalError'.

**To Reproduce**

```python
import torch
from monai.auto3dseg.analyzer import ImageStats

analyzer = ImageStats(image_key="image")
analyzer(
    {
        "image": torch.arange(64.0, dtype=torch.float32).reshape(1, 4, 4, 4),
        "nda_croppeds": [torch.ones((2, 2, 2), dtype=torch.float32)],
    }
)
```

**Expected behavior**
```UnboundLocalError local variable 'nda_croppeds' referenced before assignment.```
If nda_croppeds is already provided, ImageStats should use it directly instead of recomputing foreground crops or crashing.

**Screenshots**

**Environment**
MONAI version: 1.5.2+87.g853f7023
MONAI rev id: 853f70236e42fe96fd690363fafd3a2d1e924c9a
Python version: 3.9.12
Pytorch version: 2.8.0+cpu
Numpy version: 1.24.3
Platform: Windows-10-10.0.22631-SP0
Has CUDA: False

**Additional context**
While investigating this path, I also confirmed that the analyzer-side grad-state leak discussed in #5889 is still reproducible on the current dev branch when exceptions are raised before the restore line.

I have a local fix prepared with regression tests for both the nda_croppeds crash and the grad state leak.

### PR DESCRIPTION
**Pull request by bluehyena**
Fixes #8802
Also addresses #5889

### Description

This pull request fixes two issues in `monai.auto3dseg.analyzer`.

- `ImageStats.__call__()` now correctly handles the optional precomputed `nda_croppeds` input path instead of raising `UnboundLocalError`.

- `ImageStats`, `FgImageStats`, and `LabelStats` now use an exception-safe no-grad context so analyzer failures do not leak the global PyTorch grad-enabled state, which is related to #5889.

I also added tests covering both behaviors.

### Types of changes

### CODE REVIEW

**Comment by williams145** [comment]
I noticed a couple of technical differences worth flagging between this approach and the fixes in #8809 and #8810.

On nda_croppeds handling: The ternary here has no validation on the pre-computed value , if a caller passes a wrong type or a list with the wrong number of entries, the error surfaces downstream as a confusing shape mismatch rather than a clear message. #8809 adds explicit type and length validation with a descriptive ValueError.

On grad state: torch.no_grad() always restores grad to enabled after the block, regardless of what state the caller was in before. If the caller had grad disabled before calling the analyzer, this silently re-enables it , which is a subtle semantic change. The try/finally pattern in #8810 uses torch.is_grad_enabled() to capture and restore the exact pre-call state, so both enabled→enabled and disabled→disabled are handled correctly.

File: monai/auto3dseg/analyzer.py
Comment (ericspod) [inline monai/auto3dseg/analyzer.py:221]: This docstring comment appears spurious here, although it was in the original code please place it at the top of the method and check it's contents are correct.

File: monai/auto3dseg/analyzer.py [hunk 1]
```diff
@@ -252,39 +252,35 @@ def __call__(self, data):
...
-
-        report[ImageStatsKeys.SHAPE] = [list(nda.shape) for nda in ndas]
-        report[ImageStatsKeys.CHANNELS] = len(ndas)
-        report[ImageStatsKeys.CROPPED_SHAPE] = [list(nda_c.shape) for nda_c in nda_croppeds]
-        report[ImageStatsKeys.SPACING] = (
-            affine_to_spacing(data[self.image_key].affine).tolist()
-            if isinstance(data[self.image_key], MetaTensor)
-            else [1.0] * min(3, data[self.image_key].ndim)
-        )
+        with torch.no_grad():
```
ericspod: One other thing to note is that `no_grad` could be used as a method decorator as well. This might be cleaner for this code but you would have to check it is working as intended.

ericspod (changes requested):
Hi @bluehyena thanks for looking into this. I have minor comments to make, we should also look at @williams145 's PRs though I commented in there that using `no_grad` should work correctly and so I'm waiting to hear back from them.

bluehyena:
Hi @ericspod thanks for the review! 

I moved the `ImageStats.__call__` docstring to the top of the method and switched the analyzer `__call__` methods to use `@torch.no_grad()` so it is clearer that the whole method is intended to run without autograd tracking.

I also checked the current PyTorch behavior and local tests to make sure this works as intended: `no_grad` restores the caller's previous grad state on exit, including exception paths. 

In addition, I incorporated the `nda_croppeds` validation suggested in @williams145's PR so that malformed pre computed values now raise a clear `ValueError`, and I expanded the tests to cover valid pre-computed crops, invalid inputs, and grad state preservation on both normal and exception paths.

williams145:
I want to correct something from my earlier comment. I said that torch.no_grad() unconditionally re-enables gradients on exit, that was wrong. I looked at the implementation and __exit__ calls torch.set_grad_enabled(self.prev), so if gradients were already off going in, they stay off, it doesn't force them back on. I apologise for the inaccuracy.

I can see the nda_croppeds validation has since been added. That addresses the concern I raised earlier, good stuff.

ericspod (commented):
Thanks @bluehyena for this update, please look at the Coderabbit comment but I think @williams145 may want to address that instead, also please look at the DCO issue.

ericspod:
> I want to correct something from my earlier comment. I said that torch.no_grad() unconditionally re-enables gradients on exit, that was wrong. I looked at the implementation and **exit** calls torch.set_grad_enabled(self.prev), so if gradients were already off going in, they stay off, it doesn't force them back on. I apologise for the inaccuracy.
> 
> I can see the nda_croppeds validation has since been added. That addresses the concern I raised earlier, good stuff.

Thanks for contributing! I think we'll merge this one but then have a look at your PRs if there's anything further you want to address.

ericspod (approved):

bluehyena:
Thanks very much for the review.

I have rebased the branch with sign-offs and force-pushed the updated history, so DCO is green now.

Also, regarding the earlier CodeRabbit follow-up: @williams145, are you planning to handle that as part of the related follow-up work, or would you prefer that I address it here?

ericspod:
> Thanks very much for the review.
> 
> I have rebased the branch with sign-offs and force-pushed the updated history, so DCO is green now.

I see there are some CI issues now, I should be able to just rerun jobs to fix these.

PR #8809, opened 2026-04-08
Fix #8802: read nda_croppeds from input dict when pre-computed

### LINKED ISSUE #8802: [BUG] ImageStats crashes when precomputed 'nda_croppeds' is provided

williams145:
## Problem
`ImageStats.__call__` crashes with `UnboundLocalError: local variable 'nda_croppeds' referenced before assignment` when the caller pre-populates `"nda_croppeds"` in the input dict.

## Root Cause
`nda_croppeds` was only assigned inside the `if "nda_croppeds" not in d:` branch, but used unconditionally at lines 267 and 279 regardless of which branch was taken.

## Fix
Added the missing `else` branch to read the pre-computed value from the dict. Also wrapped the method body in `try/finally` to guarantee `torch.set_grad_enabled` is always restored on exit, even if an exception is raised mid-computation.

## Testing
```python
analyzer = ImageStats(image_key="image")
data = {"image": torch.rand(1, 10, 10, 10), "nda_croppeds": [np.random.rand(8, 8, 8)]}
result = analyzer(data)  # previously raised UnboundLocalError
````
### Types of changes
- [x] Non-breaking change (bug fix or new feature that would not break existing functionality)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] New tests added
- [ ] Integration tests run locally
- [x] Quick tests run locally
- [x] Docstrings updated
- [ ] Documentation updated


atharvajoshi01 (commented):
Two fixes in one: reading precomputed nda_croppeds from the dict (the actual bug) and wrapping the whole block in try/finally to restore grad state on exception. Both are correct. The else branch on line ~261 was the missing piece from the original code.

williams145:

williams145:
> The quick-py3 (macOS-latest) failure is a pre-existing infrastructure issue, pytype 2024.4.11 fails to install on the macOS runner due to a missing pybind11 dependency, which is unrelated to the changes in this PR. The same failure appears on other open PRs against dev for the same reason.

Note: PR #8803 by @bluehyena covers overlapping ground on issue #8802. The key difference in this PR is explicit validation of the pre-computed nda_croppeds value, wrong type or mismatched channel count raises a clear ValueError rather than propagating silently to a confusing downstream error.

ericspod:
Hi @williams145 please look at my comment on #8810 as this needs to be revised if I'm correct about `no_grad` behaviour.

williams145:
> Hi @williams145 please look at my comment on #8810 as this needs to be revised if I'm correct about `no_grad` behaviour.

I've addressed your point on #8810 @ericspod , the no_grad() characterisation in my earlier comment was wrong and I've acknowledged that there. Worth noting this PR also fixes an unrelated issue: the UnboundLocalError when nda_croppeds is pre-computed in the input dict, which is independent of the grad state discussion.

williams145:
Closing this in favour of #8803, which was merged and covers the same fix. Thanks for the review.

PR #8810, opened 2026-04-08
Fix #5889: restore grad state on exception in FgImageStats and LabelStats

### LINKED ISSUE #5889: data analyzer set_grad_enabled
**Describe the bug**
the `torch.set_grad_enabled(False)` in the data analyzer may disable the pytorch autograd completely, and reenable it later:

https://github.com/Project-MONAI/MONAI/blob/ae7f86ef187e1de7a32ca3e8589511a180f76e89/monai/auto3dseg/analyzer.py#L240
https://github.com/Project-MONAI/MONAI/blob/ae7f86ef187e1de7a32ca3e8589511a180f76e89/monai/auto3dseg/analyzer.py#L266

for the unit test cases, if the program exits somewhere between these two commands, it may leave the autograd 'off' and break the other test cases.

williams145:
## Problem
`FgImageStats.__call__` and `LabelStats.__call__` disable PyTorch gradient computation on entry but only restore it via a plain assignment before `return`. If any exception is raised mid-execution (e.g. shape mismatch, `RuntimeError` from `verify_report_format`), the restore line is skipped and `torch.is_grad_enabled()` stays `False` for the rest of the process, silently breaking all subsequent gradient computations.

## Root Cause
The grad state restore had no exception safety, `torch.set_grad_enabled(restore_grad_state)` was a bare statement that would be skipped entirely if anything above it raised.

## Fix
Wrapped the computation body in `try/finally` in both `FgImageStats.__call__` and `LabelStats.__call__` so the restore is guaranteed regardless of how the function exits.

## Testing
```python
import torch
from monai.auto3dseg import FgImageStats

analyzer = FgImageStats(image_key="image", label_key="label")
torch.set_grad_enabled(True)
try:
    analyzer({"image": MetaTensor(torch.rand(1,4,4,4)), "label": MetaTensor(torch.rand(5,5,5))})
except Exception:
    pass
assert torch.is_grad_enabled()  # was False before this fix


williams145:

williams145:
> The quick-py3 (macOS-latest) failure is a pre-existing infrastructure issue, pytype 2024.4.11 fails to install on the macOS runner due to a missing pybind11 dependency, which is unrelated to the changes in this PR. The same failure appears on other open PRs against dev for the same reason.

Note: PR #8803 by @bluehyena covers overlapping ground on issue #5889 using torch.no_grad(). The distinction in this PR is semantic: torch.no_grad() unconditionally re-enables grad after the block, which silently mutates caller state if grad was disabled before the call. The try/finally + restore_grad_state pattern here preserves the exact pre-call state in both directions.

ericspod:
Hi @williams145 thanks for looking into this. From my reading of the code for `no_grad` [here](https://github.com/pytorch/pytorch/blob/70d99e998b4955e0049d13a98d77ae1b14db1f45/torch/autograd/grad_mode.py#L80), the decorator seems to do what you want and doesn't unconditionally restore gradient calculation. This appears to be old code so it's not that this was fixed recently. Could you please double check that you're seeing the behaviour in `no_grad` you're trying to fix?

williams145:
> Hi @williams145 thanks for looking into this. From my reading of the code for `no_grad` [here](https://github.com/pytorch/pytorch/blob/70d99e998b4955e0049d13a98d77ae1b14db1f45/torch/autograd/grad_mode.py#L80), the decorator seems to do what you want and doesn't unconditionally restore gradient calculation. This appears to be old code so it's not that this was fixed recently. Could you please double check that you're seeing the behaviour in `no_grad` you're trying to fix?

After further inspection, you are actually correct, I apologise for the inaccuracy. I looked at the PyTorch source and torch.no_grad().__exit__ calls torch.set_grad_enabled(self.prev), so it does properly restore the caller's pre-entry state, it doesn't unconditionally re-enable. That part of my comment was wrong.

The actual bug being fixed is different: the original code used a raw torch.set_grad_enabled(False) call with the restore on the happy path only. No context manager, no try/finally. So any exception raised between the two calls leaves grad permanently disabled for the rest of that process. torch.no_grad() would fix this equally well. The try/finally pattern I used is equivalent; I just chose it to make the save/restore intent explicit, but there's no semantic advantage over with torch.no_grad().

Happy to refactor to use torch.no_grad() if you'd prefer the more idiomatic form.

ericspod:
> > Hi @williams145 thanks for looking into this. From my reading of the code for `no_grad` [here](https://github.com/pytorch/pytorch/blob/70d99e998b4955e0049d13a98d77ae1b14db1f45/torch/autograd/grad_mode.py#L80), the decorator seems to do what you want and doesn't unconditionally restore gradient calculation. This appears to be old code so it's not that this was fixed recently. Could you please double check that you're seeing the behaviour in `no_grad` you're trying to fix?
> 
> After further inspection, you are actually correct, I apologise for the inaccuracy. I looked at the PyTorch source and torch.no_grad().**exit** calls torch.set_grad_enabled(self.prev), so it does properly restore the caller's pre-entry state, it doesn't unconditionally re-enable. That part of my comment was wrong.
> 
> The actual bug being fixed is different: the original code used a raw torch.set_grad_enabled(False) call with the restore on the happy path only. No context manager, no try/finally. So any exception raised between the two calls leaves grad permanently disabled for the rest of that process. torch.no_grad() would fix this equally well. The try/finally pattern I used is equivalent; I just chose it to make the save/restore intent explicit, but there's no semantic advantage over with torch.no_grad().
> 
> Happy to refactor to use torch.no_grad() if you'd prefer the more idiomatic form.

Thanks for checking I'm not going crazy! As I said in #8803 we'll merge that one then come back to your PRs to see what to additionally change.

williams145:
> > > Hi @williams145 thanks for looking into this. From my reading of the code for `no_grad` [here](https://github.com/pytorch/pytorch/blob/70d99e998b4955e0049d13a98d77ae1b14db1f45/torch/autograd/grad_mode.py#L80), the decorator seems to do what you want and doesn't unconditionally restore gradient calculation. This appears to be old code so it's not that this was fixed recently. Could you please double check that you're seeing the behaviour in `no_grad` you're trying to fix?
> > 
> > 
> > After further inspection, you are actually correct, I apologise for the inaccuracy. I looked at the PyTorch source and torch.no_grad().**exit** calls torch.set_grad_enabled(self.prev), so it does properly restore the caller's pre-entry state, it doesn't unconditionally re-enable. That part of my comment was wrong.
> > The actual bug being fixed is different: the original code used a raw torch.set_grad_enabled(False) call with the restore on the happy path only. No context manager, no try/finally. So any exception raised between the two calls leaves grad permanently disabled for the rest of that process. torch.no_grad() would fix this equally well. The try/finally pattern I used is equivalent; I just chose it to make the save/restore intent explicit, but there's no semantic advantage over with torch.no_grad().
> > Happy to refactor to use torch.no_grad() if you'd prefer the more idiomatic form.
> 
> Thanks for checking I'm not going crazy! As I said in #8803 we'll merge that one then come back to your PRs to see what to additionally change.

Understood, happy to wait for #8803 to be resolved. Just flagging that this PR currently has 100% docstring coverage and all pre-merge checks green, so it should be straightforward to review whenever you're ready.

williams145:
Closing this as #8803, which was merged, also addresses #5889. Thanks for the patience while this was under review.

PR #8776, opened 2026-03-17
Fix compute_shape_offset non-tuple indexing for PyTorch >=2.9

Harrisaint:
Fixes #8775  .

### Description

This PR resolves issue #8775 by casting spatial_shape to a tuple inside compute_shape_offset to prevent breaking changes in PyTorch >= 2.9.

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick --unittests  --disttests`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

PR #8812, opened 2026-04-09
Fix #8775: handle torch.Tensor input in compute_shape_offset

### LINKED ISSUE #8775: [BUG] compute_shape_offset triggers PyTorch ≥2.9 breaking change for non-tuple sequence indexing
## Describe the bug
`compute_shape_offset` in `monai/data/utils.py` (line 883) passes `spatial_shape` directly to `np.array()`. When `spatial_shape` is a PyTorch tensor, this triggers the non-tuple sequence multidimensional indexing deprecation that became a breaking change in PyTorch 2.9+.

## To reproduce
Run any pipeline that goes through `SaveImaged` → `NiftiWriter.set_metadata` → `resample_if_needed` → `spatial_resample` → `compute_shape_offset` with PyTorch ≥ 2.9.

## Expected behavior
No error. `spatial_shape` should be converted to a plain Python tuple before being passed to `np.array`.

## Actual behavior
On PyTorch ≥ 2.9, this either raises an error or produces silently wrong results.
On PyTorch 2.8 it was a deprecation warning.

## Environment
- MONAI: 1.5.2
- PyTorch: 2.10.0+cu129
- Python: 3.11
- OS: Windows 11

## Suggested fix
```python
# monai/data/utils.py, line 883
# Before:
shape = np.array(spatial_shape, copy=True, dtype=float)
# After:
shape = np.array(tuple(spatial_shape), copy=True, dtype=float)

williams145:
## Description

`compute_shape_offset` in `monai/data/utils.py` passes `spatial_shape` directly to `np.array()`. When `spatial_shape` is a `torch.Tensor`, this relies on the non-tuple sequence indexing protocol, which PyTorch removed in version 2.9. The call raises a hard error on PyTorch ≥ 2.9.

The fix is to wrap `spatial_shape` in `tuple()` before passing it to `np.array()`. This routes through `__iter__`, which is stable across all PyTorch versions, and produces 0-d scalar tensors that NumPy consumes correctly.

## Root cause

The direct caller in `monai/transforms/spatial/functional.py` (line 115) constructs an `in_spatial_size` as a `torch.Tensor` and passes it straight to `compute_shape_offset`. This path has been broken since PyTorch 2.9.

## Changes

- `monai/data/utils.py` — one-character change: `np.array(spatial_shape, ...)` → `np.array(tuple(spatial_shape), ...)`
- `tests/data/utils/test_compute_shape_offset.py`: new unit tests covering `torch.Tensor`, `np.ndarray`, and plain list inputs

Fixes #8775

### DIFF
--- a/monai/data/utils.py
+++ b/monai/data/utils.py
@@ -881,7 +881,7 @@ def compute_shape_offset(
             Default is False, using option 1 to compute the shape and offset.
 
     """
-    shape = np.array(spatial_shape, copy=True, dtype=float)
+    shape = np.array(tuple(spatial_shape), copy=True, dtype=float)
     sr = len(shape)
     in_affine_ = convert_data_type(to_affine_nd(sr, in_affine), np.ndarray)[0]
     out_affine_ = convert_data_type(to_affine_nd(sr, out_affine), np.ndarray)[0]
--- a/tests/data/utils/test_compute_shape_offset.py
+++ b/tests/data/utils/test_compute_shape_offset.py
@@ -21,7 +21,7 @@ from monai.data.utils import compute_shape_offset
 
 class TestComputeShapeOffset(unittest.TestCase):
     """Unit tests for :func:`monai.data.utils.compute_shape_offset`."""
-    
+
     def test_pytorch_size_input(self):
         """Validate `torch.Size` input produces expected shape and offset.
 
--- a/tests/data/utils/test_compute_shape_offset.py
+++ b/tests/data/utils/test_compute_shape_offset.py
@@ -1,0 +1,49 @@
+# Copyright (c) MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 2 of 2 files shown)


williams145:

ericspod:
Hi @williams145 thanks for this fix, this was already in the works in #8776 however. What I'd suggest is to merge that one first then merge in your extended test set which I think are good to add. We have to merge #8814 before this can be merged however which should address the pybind11 issue.

williams145:
> Hi @williams145 thanks for this fix, this was already in the works in #8776 however. What I'd suggest is to merge that one first then merge in your extended test set which I think are good to add. We have to merge #8814 before this can be merged however which should address the pybind11 issue.

PR #8814 has been merged. The pybind11 blocker should now be resolved, this should mean that CI can be re-triggered when convenient.

atharvajoshi01 (commented):
Converting to tuple before passing to np.array is a good fix for PyTorch >= 2.9 which deprecated non-tuple sequence indexing. Test coverage looks solid — covers numpy, list, tuple, and torch.Tensor inputs.

ericspod (approved):

PR #8818, opened 2026-04-11
fix: use register_buffer for kernel and kernel_vol in LocalNormalizedCrossCorrelationLoss

Zeesejo:
### Description

This PR fixes an incorrect kernel initialization pattern in `LocalNormalizedCrossCorrelationLoss` introduced in `monai/losses/image_dissimilarity.py`.

**Bug found:** The original code used plain attribute assignment with a typo:
```python
self.kernel = _kernel(self.kernel_size)
self.kernel.require_grads = False  # typo: 'require_grads' is NOT a valid tensor attribute
self.kernel_vol = self.get_kernel_vol()
```

`require_grads` (with an 's') is not a valid PyTorch tensor attribute. The correct attribute is `requires_grad`. This means the kernel was **silently tracking gradients** throughout every forward pass, wasting memory and computation.

Furthermore, assigning the kernel as a plain attribute (`self.kernel = ...`) rather than registering it as a buffer means:
- The kernel tensor does **not** automatically move to the correct device when `.to(device)` / `.cuda()` / `.half()` is called on the loss module
- The kernel is **not saved/restored** correctly via `state_dict()` / `load_state_dict()`

**Fix:** Replace both assignments with `register_buffer`, which is the correct PyTorch `nn.Module` pattern for constant tensors:
```python
self.register_buffer("kernel", _kernel(self.kernel_size))
self.register_buffer("kernel_vol", self.get_kernel_vol())
```

This removes the need for any manual `requires_grad` management and ensures proper device placement.

### Types of changes

Signed-off-by: Zeesejo <92383127+Zeesejo@users.noreply.github.com>


Zeesejo:
Related issue documenting the full bug analysis and 1→10→100 chain of improvements: #8819

The issue also identifies two follow-up areas to audit:
1. `GlobalMutualInformationLoss` — `bin_centers` should also be a registered buffer
2. Broader test coverage for device movement of all loss modules

Fixes #8819

ericspod (approved):
Hi @Zeesejo thanks for this fix. I don't think that more tests are needed as the behaviour shouldn't be change drastically and it's unlikely anyone is storing the state dict for this loss function.

monai/losses/image_dissimilarity.py
ericspod, line 125:
```suggestion

def get_kernel_vol(self):
    vol0: torch.Tensor = self.kernel
    vol: torch.Tensor = vol0
    for _ in range(self.ndim - 1):
        vol = torch.matmul(vol.unsqueeze(-1), vol0.unsqueeze(0))
    return torch.sum(vol)
```
Now that `self.kernel` is a registered buffer it's type seems to be `Module|Tensor` so may need to be cast so that `vol` has the correct type.

ericspod, line 143: Similarly `kernel` and `kernel_vol` will need `torch.Tensor` as a type annotation and possibly casting the assigned values.

monai/losses/image_dissimilarity.py
```diff
@@ -111,14 +111,14 @@ def __init__(
             raise ValueError(f"kernel_size must be odd, got {self.kernel_size}")
 
         _kernel = look_up_option(kernel_type, kernel_dict)
-        self.kernel = _kernel(self.kernel_size)
-        self.kernel.require_grads = False
-        self.kernel_vol = self.get_kernel_vol()
+        self.register_buffer("kernel", _kernel(self.kernel_size))
+        self.register_buffer("kernel_vol", self.get_kernel_vol())
```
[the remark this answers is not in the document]
Zeesejo: Good catch! You're right — since `kernel` and `kernel_vol` are deterministically computed from `kernel_size`, there's no need to save them in checkpoints. I'll add `persistent=False` to both `register_buffer` calls to avoid breaking old checkpoints with `strict=True` loading.

monai/losses/image_dissimilarity.py
```diff
@@ -111,14 +111,14 @@ def __init__(
             raise ValueError(f"kernel_size must be odd, got {self.kernel_size}")
 
         _kernel = look_up_option(kernel_type, kernel_dict)
-        self.kernel = _kernel(self.kernel_size)
-        self.kernel.require_grads = False
-        self.kernel_vol = self.get_kernel_vol()
+        self.register_buffer("kernel", _kernel(self.kernel_size), persistent=False)
```
ericspod, line 116:
```suggestion
        self.kernel: torch.Tensor
        self.kernel_vol: torch.Tensor
        self.register_buffer("kernel", _kernel(self.kernel_size), persistent=False)
```

I think this will address the mypy issue being confused about what members get created with register_buffer.

PR #8821, opened 2026-04-11
Fix: swap input_amplitude and target_amplitude in JukeboxLoss.forward

### LINKED ISSUE #8820: [Bug] `JukeboxLoss.forward`: `input_amplitude` and `target_amplitude` variable names are swapped
## Describe the bug

In `monai/losses/spectral_loss.py`, `JukeboxLoss.forward()` has the variable names `input_amplitude` and `target_amplitude` **swapped**:

```python
def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
    input_amplitude = self._get_fft_amplitude(target)   # ← BUG: should be _get_fft_amplitude(input)
    target_amplitude = self._get_fft_amplitude(input)   # ← BUG: should be _get_fft_amplitude(target)

    loss = F.mse_loss(target_amplitude, input_amplitude, reduction="none")
```

## Impact

While this does **not** affect the numerical output of the loss (since MSE is symmetric: `mse(a, b) == mse(b, a)`), the naming is **semantically inverted** which:

1. Makes the code **misleading and confusing** to anyone reading, extending, or debugging it
2. Creates a **silent maintenance hazard** — if anyone adds asymmetric processing (e.g. normalization, weighting, gradient masking) based on input vs. target, the swap will cause subtle, hard-to-debug training errors
3. Violates the principle of least surprise for the standard `forward(input, target)` contract that all PyTorch loss functions follow

## Root Cause

In `forward(self, input, target)`:
- `input_amplitude` is computed from `target` (wrong)
- `target_amplitude` is computed from `input` (wrong)

## Expected behavior

```python
def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
    input_amplitude = self._get_fft_amplitude(input)    # correct
    target_amplitude = self._get_fft_amplitude(target)  # correct
    loss = F.mse_loss(target_amplitude, input_amplitude, reduction="none")
```

## Location

`monai/losses/spectral_loss.py` lines 55-56

Affects all MONAI versions containing `JukeboxLoss`. Reproducible on `dev` branch as of 2026-04-11.

Zeesejo:
Fixes #8820

### Description

In `JukeboxLoss.forward()`, the variable names `input_amplitude` and `target_amplitude` were swapped:
- `input_amplitude` was computed from `target` (should be from `input`)
- `target_amplitude` was computed from `input` (should be from `target`)

This fix corrects the assignments to match semantic meaning and the standard `forward(input, target)` PyTorch convention.

### Types of changes

### DIFF
--- a/monai/losses/spectral_loss.py
+++ b/monai/losses/spectral_loss.py
@@ -55,8 +55,8 @@ class JukeboxLoss(_Loss):
         self.fft_norm = fft_norm
 
     def forward(self, input: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
-        input_amplitude = self._get_fft_amplitude(target)
-        target_amplitude = self._get_fft_amplitude(input)
+        input_amplitude = self._get_fft_amplitude(input)
+        target_amplitude = self._get_fft_amplitude(target)
 
         # Compute distance between amplitude of frequency components
         # See Section 3.3 from https://arxiv.org/abs/2005.00341
--- a/monai/losses/ssim_loss.py
+++ b/monai/losses/ssim_loss.py
@@ -111,17 +111,17 @@ class SSIMLoss(_Loss):
                 # 2D data
                 x = torch.ones([1,1,10,10])/2
                 y = torch.ones([1,1,10,10])/2
-                print(1-SSIMLoss(spatial_dims=2)(x,y))
+                print(SSIMLoss(spatial_dims=2)(x,y))
 
                 # pseudo-3D data
                 x = torch.ones([1,5,10,10])/2  # 5 could represent number of slices
                 y = torch.ones([1,5,10,10])/2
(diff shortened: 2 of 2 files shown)


Zeesejo:
**Update**: This PR now also includes a second commit (`4417d79`) that fixes a docstring bug in `SSIMLoss.forward()` (tracked in #8822).

The `forward()` docstring examples showed `print(1-SSIMLoss()(x,y))`, but since `SSIMLoss` already computes `1 - ssim_value` internally, those examples would actually return the raw SSIM value (not the loss). This fix removes the incorrect `1-` prefix from all three docstring print examples.

ericspod (approved):
Hi @Zeesejo thanks for this change, I think it's fine and of course the behaviour hasn't changed since `mse_loss` was symmetric. I don't think the Copilot comments need to be addressed.

PR #8823, opened 2026-04-12
fix(LoadImage): raise OptionalImportError when user-specified reader is not installed

### LINKED ISSUE #7437: Raise the exception when LoadImage has a reader specified but it is not installed

KamayaniR:
## What does this PR do?

Fixes #7437

When a user explicitly specifies a reader in `LoadImage` (e.g. `LoadImage(reader="ITKReader")`),
and the required package is not installed, the previous behavior was to silently warn and
continue with fallback readers. This masked real configuration errors.

## Changes

- `monai/transforms/io/array.py`: Changed `warnings.warn` to `raise OptionalImportError`
  in the user-specified reader path inside `LoadImage.__init__`
- `tests/transforms/test_load_image.py`: Added `TestLoadImageReaderNotInstalled` to verify
  the exception is raised correctly

## Behavior

**Before:**
LoadImage(reader="ITKReader")  # itk not installed
# UserWarning: silently falls back to another reader — user has no idea!

**After:**
LoadImage(reader="ITKReader")  # itk not installed
# OptionalImportError: required package for reader ITKReader is not installed

## Checklist
- [x] Fixes the reported issue
- [x] Backward compatible (only affects explicit reader= usage)
- [x] Linting passes (ruff check)
- [x] Test added
- [x] Signed-off (git commit -s)

### DIFF
--- a/monai/transforms/io/array.py
+++ b/monai/transforms/io/array.py
@@ -210,10 +210,10 @@ class LoadImage(Transform):
                     the_reader = look_up_option(_r.lower(), SUPPORTED_READERS)
                 try:
                     self.register(the_reader(*args, **kwargs))
-                except OptionalImportError:
-                    warnings.warn(
+                except OptionalImportError as e:
+                    raise OptionalImportError(
                         f"required package for reader {_r} is not installed, or the version doesn't match requirement."
-                    )
+                    ) from e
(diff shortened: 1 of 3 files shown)


dzenanz (approved):
LGTM

PR #8833, opened 2026-04-18
Validate downloaded file integrity and raise ValueError on hash mismatch

### LINKED ISSUE #8832: Raise different exception when hash value of download is wrong

e-mny:
Fixes #8832 .

### Description

Adds a check on downloaded files to verify their hash against the expected value and raises a `ValueError` if there is a mismatch, indicating possible corruption or tampering.

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality)..


tests/apps/test_download_and_extract.py
```diff
@@ -42,10 +42,10 @@ def test_actions(self):
...
             try:
                 download_url(url, filepath, wrong_md5)
-            except (ContentTooShortError, HTTPError, RuntimeError) as e:
-                if isinstance(e, RuntimeError):
+            except (ContentTooShortError, HTTPError, ValueError, RuntimeError) as e:
+                if isinstance(e, ValueError):
                     # FIXME: skip MD5 check as current downloading method may fail
-                    self.assertTrue(str(e).startswith("md5 check"))
+                    self.assertIn("hash check", str(e))
                 return  # skipping this test due the network connection errors
```
ericspod:
```suggestion
            except ValueError as e:
                # FIXME: skip MD5 check as current downloading method may fail
                self.assertIn("hash check", str(e))
                return 
            except (ContentTooShortError, HTTPError, RuntimeError) as e:
                return  # skipping this test due the network connection errors
```
I think a separate clause makes sense despite what the original was.

tests/test_utils.py
```diff
@@ -182,6 +182,39 @@ def skip_if_downloading_fails():
...
+
+        raise v_e
+
+
+SAMPLE_TIFF = "https://huggingface.co/datasets/MONAI/testing_data/resolve/main/CMU-1.tiff"
+SAMPLE_TIFF_HASH = "73a7e89bc15576587c3d68e55d9bf92f09690280166240b48ff4b48230b13bcd"
+SAMPLE_TIFF_HASH_TYPE = "sha256"
+
+
+class TestDownloadUrl(unittest.TestCase):
```
ericspod: We have a conflict now from your previous PR being merged. These two classes possibly need to be merged as the two tests are different.

tests/apps/test_download_and_extract.py
e-mny: resolved in https://github.com/Project-MONAI/MONAI/pull/8833/commits/88f83727b3d75e8daf81c66a2a82a153de3d2ce9

tests/test_utils.py
e-mny: resolved in https://github.com/Project-MONAI/MONAI/pull/8833/commits/64b8e090b5bf999302c732c93f121b05ca90aae7

monai/apps/utils.py
```diff
@@ -220,17 +220,16 @@ def download_url(
...
     if not filepath:
         filepath = Path(".", _basename(url)).resolve()
         logger.info(f"Default downloading to '{filepath}'")
     filepath = Path(filepath)
     if filepath.exists():
         if not check_hash(filepath, hash_val, hash_type):
-            raise RuntimeError(
-                f"{hash_type} check of existing file failed: filepath={filepath}, expected {hash_type}={hash_val}."
+            raise ValueError(
+                f"{hash_type} hash check of existing file failed: filepath={filepath}, expected {hash_type}={hash_val}."
```
ericspod:
```suggestion
                f"{hash_type} hash check of existing file failed: {filepath=}, expected {hash_type}={hash_val}."
```
There's newish syntax to make this a little cleaner if you wanted to use it.

tests/test_utils.py
```diff
@@ -182,37 +182,11 @@ def skip_if_downloading_fails():
             raise unittest.SkipTest(f"Error while downloading: {rt_e}") from rt_e  # incomplete download
 
         raise rt_e
+    except ValueError as v_e:
```
ericspod: I think this shouldn't be here as it confuses what's being tested, my original comment was about not ignoring failed hash checks. We want to consider whether the download succeeds separately from whether the hash check succeeds. If the hash check fails then `skip_if_downloading_fails` should allow that error to be raised since the download succeeded. If the download fails we skip the test because there's external factors, but if the hash check fails we want that to be a fail since something's wrong with a successfully download file. Sorry I should have caught this earlier.

ericspod (changes requested):
Hi @e-mny I think this just about good now, but I had a comment about the change to `skip_if_downloading_fails`. 

I thought about creating a new exception class to raise, eg. `HashCheckError` or something like that inheriting from `ValueError`. This would make it easier to catch that specific case, if you agree go ahead and make that change as well but I'm fine leaving `ValueError` here. Thanks!

monai/apps/utils.py
e-mny: That's cool! Never knew it was something f-strings could do, thanks for letting me learn something new today!

monai/apps/utils.py
```diff
@@ -220,17 +220,16 @@ def download_url(
...
     """
     if not filepath:
         filepath = Path(".", _basename(url)).resolve()
         logger.info(f"Default downloading to '{filepath}'")
     filepath = Path(filepath)
     if filepath.exists():
         if not check_hash(filepath, hash_val, hash_type):
-            raise RuntimeError(
-                f"{hash_type} check of existing file failed: filepath={filepath}, expected {hash_type}={hash_val}."
+            raise ValueError(
```
ericspod: Hi @e-mny sorry for the delay. What I would suggest here is to put `class HashCheckError(ValueError): pass` at the top of this file, then raise that exception. This would make it explicit the error came from a hash check fail and it's something that can be caught specifically in an except block.

e-mny: Noted, have amended the error and edited relevant lines which utilize the HashCheckError!

I'll be looking out for other issues to work on concurrently, but if you have any high-priority ones or need my help on, do direct me! Cheers

monai/apps/utils.py
```diff
@@ -220,18 +224,15 @@ def download_url(
         HTTPError: See urllib.request.urlretrieve.
         ContentTooShortError: See urllib.request.urlretrieve.
         IOError: See urllib.request.urlretrieve.
-        RuntimeError: When the hash validation of the ``url`` downloaded file fails.
-
+        ValueError: When the hash validation of the ``url`` downloaded file fails.
```
ericspod:
```suggestion
        HashCheckError: When the hash validation of the ``url`` downloaded file fails.
```

monai/apps/utils.py
```diff
@@ -325,10 +328,15 @@ def extractall(
             be False.
 
     Raises:
-        RuntimeError: When the hash validation of the ``filepath`` compressed file fails.
+        ValueError: When the hash validation of the ``filepath`` compressed file fails.
```
ericspod:
```suggestion
        HashCheckError: When the hash validation of the ``filepath`` compressed file fails.
```

monai/apps/utils.py
```diff
@@ -48,6 +48,10 @@
 SUPPORTED_HASH_TYPES = {"md5": hashlib.md5, "sha1": hashlib.sha1, "sha256": hashlib.sha256, "sha512": hashlib.sha512}
 
 
+class HashCheckError(ValueError):
```
ericspod, line 59: This class should be added to __all__ in this file and imported in the monai/apps/__init__.py file.

tests/apps/test_download_and_extract.py
```diff
@@ -26,39 +25,62 @@
...
+    def test_download_url_hash_mismatch(self):
+        """download_url should raise ValueError on hash mismatch."""
+        filepath = self.testing_dir / "MedNIST.tar.gz"
+
+        with skip_if_downloading_fails():
+            # First ensure file is downloaded correctly
+            download_url(self.url, filepath, hash_val=self.hash_val, hash_type=self.hash_type)
+
+        # Now test incorrect hash
+        with self.assertRaises(ValueError) as ctx:
```
ericspod:
```suggestion
        with self.assertRaises(HashCheckError) as ctx:
```

ericspod:
Hi @e-mny I have some minor changes to make here but then we can merge. We should merge this one before your most recent PR which includes many of these changes a well. Thanks for being patient!

e-mny:
Thanks @ericspod , have addressed the issues that you have mentioned. Not sure why the DCO check is failing even though I have tried committing twice using the -s flag... Any tips on how I can try to submit a new commit to sign off with no changes to my files?

And thank YOU for being so patient with me! I really appreciate it 🥹🙏

e-mny:
Disregard the previous issue on DCO - have fixed it

ericspod (approved):
Looks good now, thanks.

PR #8834, opened 2026-04-18
Fix #8462: embed patch sizes in einops pattern for einops >= 0.8 compatibility

### LINKED ISSUE #8462: [BUG]einops.layer.torch.Rearrange got an unexpected keyword argument
Describe the bug
When I am building ViT from monai

```
self.patch_embedding = PatchEmbeddingBlock(
            in_channels=in_channels,
            img_size=img_size,
            patch_size=patch_size,
            hidden_size=hidden_size,
            num_heads=num_heads,
            proj_type=proj_type,
            dropout_rate=dropout_rate,
            spatial_dims=spatial_dims,
        )
```
the issue happened here in monai.networks.blocks.patchembedding.py
```
elif self.proj_type == "perceptron":
        # for 3d: "b c (h p1) (w p2) (d p3)-> b (h w d) (p1 p2 p3 c)"
        chars = (("h", "p1"), ("w", "p2"), ("d", "p3"))[:spatial_dims]
        from_chars = "b c " + " ".join(f"({k} {v})" for k, v in chars)
        to_chars = f"b ({' '.join([c[0] for c in chars])}) ({' '.join([c[1] for c in chars])} c)"
        axes_len = {f"p{i+1}": p for i, p in enumerate(patch_size)}
        self.patch_embeddings = nn.Sequential(
            Rearrange(f"{from_chars} -> {to_chars}", **axes_len), nn.Linear(self.patch_dim, hidden_size)
        )
```
the error appers as
TypeError: Rearrange.init() got an unexpected keyword argument 'p1'

Image

Reproduction steps

Expected behavior

Your platform
Version of einops, python and DL package that you used
accelerate==1.3.0
aiofiles==23.2.1
aiohttp==3.11.12
alembic==1.14.1
anyio==4.8.0
beautifulsoup4==4.13.3
bert-score==0.3.13
cachetools==5.5.1
databricks-sdk==0.44.0
deepspeed==0.16.3
docker==7.1.0
einops==0.8.1
evaluate==0.4.3
fastapi==0.115.11
Flask==3.1.0
fonttools==4.56.0
fsspec==2024.12.0
gradio==5.20.1
grpcio==1.70.0
gunicorn==23.0.0
huggingface-hub==0.28.1
imageio==2.37.0
Jinja2==3.1.5
joblib==1.4.2
lmdb==1.6.2
matplotlib==3.10.0
mlflow==2.20.1
monai==1.4.0
nibabel==5.3.2
nltk==3.9.1
numpy==1.26.4
opencv-python-headless==4.9.0.80
pandas==2.2.3
peft==0.14.0
Pillow==11.1.0
protobuf==5.29.3
psutil==6.1.1
pyarrow==18.1.0
pydantic==2.10.6
pydicom==3.0.1
python-multipart==0.0.20
scikit-learn==1.6.1
scipy==1.15.1
SimpleITK==2.4.1
tensorboard==2.19.0
timm==1.0.14
torch==2.6.0
torchvision==0.21.0
tqdm==4.67.1
transformers==4.48.3
triton==3.2.0
Unidecode==1.3.8
uvicorn==0.34.0
wandb==0.19.6

williams145:
Fixes #8462

### Description

`PatchEmbeddingBlock` with `proj_type="perceptron"` raises `TypeError` on einops >= 0.8.x because `Rearrange.__init__()` no longer accepts arbitrary keyword arguments.

The current code builds an `axes_len` dict and passes it as `**axes_len` to `Rearrange(pattern, **axes_len)`. einops 0.8.x removed support for this.

**Fix:** embed the patch size values directly as integer literals in the einops pattern string — e.g. `"b c (h 16) (w 16) (d 16)"` instead of using kwargs. einops supports integer literals in patterns across all versions including 0.8.x. The change is semantically identical.

### Types of changes
- [ ] Breaking change
- [ ] New tests added

### DIFF
--- a/monai/networks/blocks/patchembedding.py
+++ b/monai/networks/blocks/patchembedding.py
@@ -38,10 +38,10 @@ class _PatchRearrange(nn.Module):
         self.patch_size = patch_size
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
-        B, C = x.shape[0], x.shape[1]
+        batch, channels = x.shape[0], x.shape[1]
         sp = x.shape[2:]
         g = tuple(s // p for s, p in zip(sp, self.patch_size))
-        v: list[int] = [B, C]
+        v: list[int] = [batch, channels]
         for gi, pi in zip(g, self.patch_size):
             v += [gi, pi]
         x = x.view(*v)
@@ -52,7 +52,7 @@ class _PatchRearrange(nn.Module):
         n_patches = 1
         for gi in g:
             n_patches *= gi
-        return x.reshape(B, n_patches, -1)
+        return x.reshape(batch, n_patches, -1)
 
 
 class PatchEmbeddingBlock(nn.Module):
--- a/monai/networks/blocks/patchembedding.py
+++ b/monai/networks/blocks/patchembedding.py
@@ -34,12 +34,26 @@ class _RearrangeFn(nn.Module):
     """Thin wrapper around einops.rearrange function, used as a fallback when the
     Rearrange layer rejects axis-size kwargs in some einops builds."""
 
-    def __init__(self, pattern: str, axes_lengths: dict) -> None:
+    def __init__(self, pattern: str, axes_lengths: dict[str, int]) -> None:
+        """Initialize the functional-rearrange fallback module.
+
+        Args:
+            pattern: Einops rearrangement pattern.
+            axes_lengths: Named axis lengths consumed by ``einops.rearrange``.
+        """
         super().__init__()
         self.pattern = pattern
         self.axes_lengths = axes_lengths
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Rearrange an input tensor using the stored einops pattern.
+
+        Args:
+            x: Input tensor.
+
+        Returns:
+            Rearranged tensor produced by ``einops.rearrange``.
+        """
         out: torch.Tensor = einops_rearrange(x, self.pattern, **self.axes_lengths)
         return out
 
--- a/monai/networks/blocks/patchembedding.py
+++ b/monai/networks/blocks/patchembedding.py
@@ -40,7 +40,8 @@ class _RearrangeFn(nn.Module):
         self.axes_lengths = axes_lengths
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
-        return einops_rearrange(x, self.pattern, **self.axes_lengths)
+        out: torch.Tensor = einops_rearrange(x, self.pattern, **self.axes_lengths)
+        return out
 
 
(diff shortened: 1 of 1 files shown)


williams145:
@ericspod: CI is green and the branch is up to date. Happy to answer any questions if helpful.

ericspod:
Hi @williams145 the following works for me with einops 0.8.1:

```python
import torch
from einops.layers.torch      import Rearrange
images = torch.rand(3, 30, 40, 3)
kwargs={"h1":2, "w1":2}
r=Rearrange('b (h h1) (w w1) c -> b h w (c h1 w1)', **kwargs)
print(r(images).shape)  # torch.Size([3, 15, 20, 12])
```

I think this replicates the use case in which you're seeing an error, can you check this code for yourself? Which version of einops specifically are you using? 

Even if this is a real issue needing to be fixed I would not recommend re-implementing a rearranging algorithm which is going to be error-prone and hard to test as the Coderabbit comment suggests above. I would look into using the `einops.rearrange` function in your class regardless, we should use the einops function which has been well tested.

williams145:
Thanks for taking a close look. You're right — I tried reproducing on my end with a fresh `pip install einops==0.8.1` and the kwargs path works fine, just like your example. The reporter listed `einops==0.8.1` in their environment but the failure mode (`Rearrange.init() got an unexpected keyword argument 'p1'` — note `init` not `__init__`) suggests something non-standard in their setup, possibly a custom build or namespace conflict.

Your point about avoiding a re-implementation is well taken. I've dropped the pure-torch fallback and replaced it with a thin wrapper around `einops.rearrange` (the function), so even the fallback path uses einops' tested rearrangement engine. The primary path still uses `Rearrange` as before — the wrapper only kicks in if `Rearrange.__init__` raises `TypeError` on a particular build.

Let me know if you'd prefer a different approach, or if you'd rather close this since the bug isn't reliably reproducible.

While you're here — would also appreciate a look at #8835 when you get a chance. It's a small, green, focused fix for a `LoadImage` crash when `set_track_meta(False)` is active.

ericspod:
Hi @williams145 honestly I think we should just close this if you agree with me the issue can't be reproduced with a clean environment. We can close the original issue as well. 

It's possible the original issue is related to the environment as you say or previous versions of einops. The implementation is better as it uses existing components, but would be a little hard to test as the Coderabbit comment states. We'd have to monkey-patch einops to force an exception raise, but another issue is that exception raises aren't specific enough to use a fallback, ie. `TypeError` could be raised for other issues other than the one at hand. I'd say you should be checking the einops version instead, but we don't know what version is the issue so we can't do that anyway.

williams145:
Closing per maintainer's suggestion (see ericspod's comment on May 5 — could not reproduce the issue in a clean environment).

PR #8835, opened 2026-04-18
Fix #8442: pass meta_data to EnsureChannelFirst when track_meta is False

### LINKED ISSUE #8442: Using `set_track_meta(False)` causes crash in `LoadImage` if `ensure_channel_first=True`
**Describe the bug**
When following the training optimization tips, I used `monai.data.set_track_meta(False)` as suggested.
However, this causes a `ValueError` in my `LoadImaged` transform, which has en `ensure_channel_first` argument set to `True`.

```shell
ValueError: Metadata not available and channel_dim=None, EnsureChannelFirst is not in use.
```

I haven't done extensive digging, but it seems to me that the conversion to `MetaTensor` at `monai.transforms.io.array:299` does not successfully generate a `MetaTensor`. It seems that this function call exits early at `monai.data.meta_tensor:557`, due to the afforementioned lack of meta tracking. In turn, this is probably the reason that the check at `monai.transforms.utility.array:198` passes, while it would be expected that the output from `LoadImaged` is actually a `MetaTensor`.

**To Reproduce**
1. Set track meta to `False` globally, using `set_track_meta(False)`
2. Attempt to use `LoadImage` or `LoadImaged` with `ensure_channel_first=True`

**Expected behavior**
No error

**Environment**

```
Printing MONAI config...
MONAI version: 1.4.0
Numpy version: 1.26.4
Pytorch version: 2.6.0+cu124
MONAI rev id: 46a5272196a6c2590ca2589029eed8e4d56ff008
MONAI __file__: /usr/local/lib/python3.12/dist-packages/monai/__init__.py

Optional dependencies:
ITK version: 5.4.3
Nibabel version: 5.3.2
scikit-image version: NOT INSTALLED or UNKNOWN VERSION.
scipy version: 1.14.1
Pillow version: 11.1.0
Tensorboard version: 2.16.2
TorchVision version: 0.21.0+cu124
tqdm version: 4.67.1
psutil version: 6.1.0
pandas version: 2.2.3
einops version: 0.8.0
mlflow version: 2.22.0
pynrrd version: NOT INSTALLED or UNKNOWN VERSION.
clearml version: NOT INSTALLED or UNKNOWN VERSION.

Printing system config...
System: Linux
Linux version: Ubuntu 24.04.2 LTS
Platform: Linux-6.8.0-55-generic-x86_64-with-glibc2.39
Processor: x86_64
Machine: x86_64
Python version: 3.12.3
Process name: python
Open files: [popenfile(path='[removed]', fd=1, position=1276, mode='w', flags=32769)]
Num physical CPUs: 40
Num logical CPUs: 80
Num usable CPUs: 80
CPU usage (%): [15.1, 10.0, 16.5, 13.4, 12.7, 14.3, 16.5, 13.4, 14.3, 5.6, 14.1, 13.8, 14.1, 14.1, 13.8, 4.4, 14.4, 11.6, 11.2, 14.3, 10.7, 3.4, 2.9, 5.6, 4.4, 2.9, 3.2, 2.9, 13.6, 3.1, 13.4, 2.9, 2.9, 2.9, 2.7, 2.9, 2.9, 5.3, 13.4, 2.9, 11.4, 14.0, 11.2, 24.8, 29.4, 14.3, 11.2, 14.1, 11.6, 10.7, 14.3, 14.1, 14.1, 13.9, 13.8, 13.4, 11.4, 14.1, 93.0, 14.1, 10.9, 3.2, 3.2, 0.0, 0.0, 2.7, 0.7, 2.7, 0.2, 2.9, 2.9, 2.7, 2.7, 2.9, 2.9, 2.9, 2.9, 10.9, 3.6, 98.5]
CPU freq. (MHz): 957
Load avg. in last 1, 5, 15 mins (%): [5.8, 9.7, 12.0]
Disk usage (%): 51.4
Total physical memory (GB): 503.5
Available memory (GB): 399.4
Used memory (GB): 100.3

Printing GPU config...
Num GPUs: 6
Has CUDA: True
CUDA version: 12.4
cuDNN enabled: True
NVIDIA_TF32_OVERRIDE: None
TORCH_ALLOW_TF32_CUBLAS_OVERRIDE: 1
cuDNN version: 90100
Current device: 0
GPU 0 Name: NVIDIA RTX A5000
GPU 0 Is integrated: False
GPU 0 Multi processor count: 64
GPU 0 Total memory (GB): 23.7
GPU 0 CUDA capability (maj.min): 8.6
GPU 1 Name: NVIDIA RTX A5000
GPU 1 Is integrated: False
GPU 1 Multi processor count: 64
GPU 1 Total memory (GB): 23.7
GPU 1 CUDA capability (maj.min): 8.6
GPU 2 Name: NVIDIA A100-PCIE-40GB
GPU 2 Is integrated: False
GPU 2 Is multi GPU board: False
GPU 2 Multi processor count: 108
GPU 2 Total memory (GB): 39.4
GPU 2 CUDA capability (maj.min): 8.0
GPU 3 Name: NVIDIA GeForce RTX 3090
GPU 3 Is integrated: False
GPU 3 Is multi GPU board: False
GPU 3 Multi processor count: 82
GPU 3 Total memory (GB): 23.7
GPU 3 CUDA capability (maj.min): 8.6
GPU 4 Name: NVIDIA RTX A5000
GPU 4 Is integrated: False
GPU 4 Is multi GPU board: False
GPU 4 Multi processor count: 64
GPU 4 Total memory (GB): 23.7
GPU 4 CUDA capability (maj.min): 8.6
GPU 5 Name: NVIDIA RTX A5000
GPU 5 Is integrated: False
GPU 5 Is multi GPU board: False
GPU 5 Multi processor count: 64
GPU 5 Total memory (GB): 23.7
GPU 5 CUDA capability (maj.min): 8.6
```

williams145:
Fixes #8442

### Description

`LoadImage` with `ensure_channel_first=True` raises `ValueError` when `set_track_meta(False)` is active.

**Root cause chain:**
1. `set_track_meta(False)` causes `MetaTensor.ensure_torch_and_prune_meta()` to return a plain tensor instead of a `MetaTensor`
2. `EnsureChannelFirst()(img)` then has no metadata source (img is not a MetaTensor and no `meta_dict` is passed)
3. This hits the `ValueError` branch in `EnsureChannelFirst.__call__` at line 205 of `utility/array.py`

**Fix:** pass `meta_data` explicitly to `EnsureChannelFirst` when `img` is not a `MetaTensor`. `EnsureChannelFirst.__call__` already accepts a `meta_dict` parameter; `meta_data` is always a validated dict at this point in `LoadImage.__call__`.

### Types of changes

### DIFF
--- a/monai/transforms/io/array.py
+++ b/monai/transforms/io/array.py
@@ -299,7 +299,10 @@ class LoadImage(Transform):
             img_array, meta_data, self.simple_keys, pattern=self.pattern, sep=self.sep
         )
         if self.ensure_channel_first:
-            img = EnsureChannelFirst()(img)
+            if not isinstance(img, MetaTensor):
+                img = EnsureChannelFirst()(img, meta_dict=meta_data)
+            else:
+                img = EnsureChannelFirst()(img)
         if self.image_only:
             return img
         return img, img.meta if isinstance(img, MetaTensor) else meta_data
--- a/monai/transforms/io/array.py
+++ b/monai/transforms/io/array.py
@@ -299,10 +299,7 @@ class LoadImage(Transform):
             img_array, meta_data, self.simple_keys, pattern=self.pattern, sep=self.sep
         )
         if self.ensure_channel_first:
-            if not isinstance(img, MetaTensor):
-                img = EnsureChannelFirst()(img, meta_dict=meta_data)
-            else:
-                img = EnsureChannelFirst()(img)
+            img = EnsureChannelFirst()(img, meta_dict=meta_data)
         if self.image_only:
             return img
         return img, img.meta if isinstance(img, MetaTensor) else meta_data
--- a/tests/transforms/test_load_image.py
+++ b/tests/transforms/test_load_image.py
@@ -497,6 +497,16 @@ class TestLoadImageMeta(unittest.TestCase):
             self.assertNotIsInstance(r, MetaTensor)
             self.assertFalse(hasattr(r, "affine"))
 
+    def test_track_meta_false_ensure_channel_first(self):
+        try:
+            set_track_meta(False)
+            r = LoadImage(image_only=True, ensure_channel_first=True)(self.test_data)
+            self.assertTupleEqual(r.shape, (1, 128, 128, 128))
+            self.assertIsInstance(r, torch.Tensor)
+            self.assertNotIsInstance(r, MetaTensor)
+        finally:
+            set_track_meta(True)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/tests/transforms/test_load_image.py
+++ b/tests/transforms/test_load_image.py
@@ -25,7 +25,7 @@ from PIL import Image
 
 from monai.apps import download_and_extract
 from monai.data import NibabelReader, PydicomReader
-from monai.data.meta_obj import set_track_meta
+from monai.data.meta_obj import get_track_meta, set_track_meta
 from monai.data.meta_tensor import MetaTensor
 from monai.transforms import LoadImage
 from monai.utils import optional_import
@@ -498,6 +498,7 @@ class TestLoadImageMeta(unittest.TestCase):
             self.assertFalse(hasattr(r, "affine"))
 
     def test_track_meta_false_ensure_channel_first(self):
+        _previous_meta = get_track_meta()
         try:
             set_track_meta(False)
             r = LoadImage(image_only=True, ensure_channel_first=True)(self.test_data)
@@ -505,7 +506,7 @@ class TestLoadImageMeta(unittest.TestCase):
             self.assertIsInstance(r, torch.Tensor)
             self.assertNotIsInstance(r, MetaTensor)
         finally:
-            set_track_meta(True)
+            set_track_meta(_previous_meta)
 
 
 if __name__ == "__main__":


williams145:
@ericspod: CI is green and the branch is up to date. Happy to answer any questions if helpful.

monai/transforms/io/array.py
```diff
@@ -300,7 +300,10 @@ def __call__(self, filename: Sequence[PathLike] | PathLike, reader: ImageReader
             img_array, meta_data, self.simple_keys, pattern=self.pattern, sep=self.sep
         )
         if self.ensure_channel_first:
-            img = EnsureChannelFirst()(img)
+            if not isinstance(img, MetaTensor):
+                img = EnsureChannelFirst()(img, meta_dict=meta_data)
+            else:
+                img = EnsureChannelFirst()(img)
```
ericspod:
```suggestion
            ctrans = EnsureChannelFirst()
            img = ctrans(img, meta_dict=None if isinstance(img, MetaTensor) else meta_data)
```
I'd like to simplify this and remove the poltergeist at the same time. Is it possible to pass `meta_data` unconditionally to the call, that is second line becomes `img = ctrans(img, meta_dict=meta_data)`?

ericspod (changes requested):
Hi @williams145 thanks for this fix. I think it's correct but we should have a regression test for the class as suggested. I suggested a change, please have a look at that as well.

williams145:
Thanks for the review. Verified the simplification works. Looking at `EnsureChannelFirst.__call__`, when `img` is a `MetaTensor` it does `meta_dict = img.meta` (line 212 of `monai/transforms/utility/array.py`), which overrides whatever was passed in. So passing `meta_data` unconditionally is safe in both cases. Applied your suggestion and pushed.

Also added a regression test (`test_track_meta_false_ensure_channel_first`) in `tests/transforms/test_load_image.py` that covers the exact scenario this PR fixes: `set_track_meta(False)` combined with `ensure_channel_first=True`. It asserts the result is a plain `torch.Tensor` (not `MetaTensor`) with the channel dim correctly added.

CI re-running now.

tests/transforms/test_load_image.py
```diff
@@ -497,6 +497,16 @@ def test_correct(self, input_param, expected_shape, track_meta):
...
 
+    def test_track_meta_false_ensure_channel_first(self):
+        try:
+            set_track_meta(False)
+            r = LoadImage(image_only=True, ensure_channel_first=True)(self.test_data)
+            self.assertTupleEqual(r.shape, (1, 128, 128, 128))
+            self.assertIsInstance(r, torch.Tensor)
+            self.assertNotIsInstance(r, MetaTensor)
+        finally:
+            set_track_meta(True)
```
ericspod:
```suggestion
        _previous_meta = get_track_meta()
        try:
            set_track_meta(False)
            r = LoadImage(image_only=True, ensure_channel_first=True)(self.test_data)
            self.assertTupleEqual(r.shape, (1, 128, 128, 128))
            self.assertIsInstance(r, torch.Tensor)
            self.assertNotIsInstance(r, MetaTensor)
        finally:
            set_track_meta(_previous_meta)
```

This ensures the previous state is restored which could be False. This isn't thread-safe by the way but every other usage of these functions isn't either so we will have to accept this for now until that's fixed.

ericspod: This is unlikely to be an issue in the tests honestly but we should do this anyway just in case.

ericspod (dismissed):

williams145:
Thanks for the suggestion; applied and pushed. Captured the previous state via get_track_meta() and restored it in the finally block so the test won't clobber whatever the suite had set before.

ericspod (approved):

ericspod (approved):

PR #8837, opened 2026-04-29
fix: replace set_track_meta with F.interpolate in RandSimulateLowResolution (#8409)

### LINKED ISSUE #8409: RandSimulateLowResolution race condition with set_track_meta()
**Describe the bug**
RandSimulateLowResolution currently calls [set_track_meta()](https://github.com/Project-MONAI/MONAI/blob/4986d7ffd2d351c9d66de0e0329884b1a26d5500/monai/transforms/spatial/array.py#L3533) which affects the global value. In case you are processing data in multiple threads, Python could release the GIL/switch the thread between the deactivation and reactivation of the track_meta setting. This can lead to 
errors if another thread e.g. calls MetaTensor.ensure_torch_and_prune_meta() at the same time.

The relevant piece of code is this:
```python
resize_tfm_downsample = Resize(
                spatial_size=target_shape, size_mode="all", mode=self.downsample_mode, anti_aliasing=False
            )

            resize_tfm_upsample = Resize(
                spatial_size=input_shape,
                size_mode="all",
                mode=self.upsample_mode,
                anti_aliasing=False,
                align_corners=self.align_corners,
            )
            # temporarily disable metadata tracking, since we do not want to invert the two Resize functions during
            # post-processing
            original_tack_meta_value = get_track_meta()
            set_track_meta(False)

            img_downsampled = resize_tfm_downsample(img)
            img_upsampled = resize_tfm_upsample(img_downsampled)

            # reset metadata tracking to original value
            set_track_meta(original_tack_meta_value)

            # copy metadata from original image to down-and-upsampled image
            img_upsampled = MetaTensor(img_upsampled)
            img_upsampled.copy_meta_from(img)

```
**To Reproduce**
Have one thread call RandSimulateLowResolution() while another uses MetaTensor.ensure_torch_and_prune_meta()

**Expected behavior**
RandSimulateLowResolution does not affect other threads.

Conceptually, I am not sure if the set_track_meta is even necessary. If only the voxel information is copied from the modified tensors, there is no need to turn off the metadata tracking, since it is discarded afterwards, right?

chhayankjain:
Fixes #8409

  ### Description

  `RandSimulateLowResolution` internally performs a downsample → upsample cycle using two
  `Resize` transforms. To prevent these from being recorded in the invertible transform
  stack, the previous implementation temporarily toggled the global `set_track_meta(False)`
  flag and restored it afterward.

  This is not thread-safe: in multi-threaded data loading (e.g. `ThreadDataLoader`), another
  thread calling `get_track_meta()` between the toggle and the restore would silently receive
  the wrong value, causing incorrect metadata tracking behaviour.

  Fix: replace the `Resize` transforms with direct `torch.nn.functional.interpolate` calls on
  a plain tensor obtained via `convert_to_tensor(img, track_meta=False)`. This avoids any
  global state mutation entirely. Output dtype (float32) and metadata-copy behaviour are
  preserved from the original implementation. `set_track_meta` is also removed from the import
  since it is no longer used anywhere in the file.

  ### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -26,7 +26,7 @@ import torch
 from monai.config import USE_COMPILED, DtypeLike
 from monai.config.type_definitions import NdarrayOrTensor
 from monai.data.box_utils import BoxMode, StandardMode
-from monai.data.meta_obj import get_track_meta, set_track_meta
+from monai.data.meta_obj import get_track_meta
 from monai.data.meta_tensor import MetaTensor
 from monai.data.utils import AFFINE_TOL, affine_to_spacing, compute_shape_offset, iter_patch, to_affine_nd, zoom_affine
 from monai.networks.layers import AffineTransform, GaussianFilter, grid_pull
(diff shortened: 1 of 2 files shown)


chhayankjain:
Hello Team, 
@KumoLiu @ericspod @Nic-Ma 
I’m currently waiting for a review on this branch, and this will be my first contribution. Could you please let me know if things looks good and if it’s ready for a review?
Thank you!!

ericspod (approved):
Looks good to me!

PR #8846, opened 2026-05-08
perf: skip redundant convert_to_tensor for MetaTensor strip

aymuos15:
Several transforms invoke ``convert_to_tensor`` twice on the same input: once with ``track_meta=get_track_meta()`` to keep the MetaTensor wrapper, then again with ``track_meta=False`` to obtain a plain-tensor view for the actual computation. The second call routes through ``_convert_tensor(data).to(dtype=..., device=..., memory_format=contiguous_format)`` (``monai/utils/type_conversion.py``) and is functionally equivalent to calling ``MetaTensor.as_tensor()`` for the purpose intended here, since the first call already enforced ``contiguous_format``.

Replace the second call with an explicit ``as_tensor()`` strip:

    img_t = img.as_tensor() if isinstance(img, MetaTensor) else img

This avoids a redundant ``convert_to_tensor`` dispatch and a no-op ``Tensor.to(memory_format=contiguous_format)`` per ``__call__``. Sites touched (all in transform ``__call__``, so per-sample per-epoch):

* ``KeepLargestConnectedComponent`` (post/array.py)
* ``RemoveSmallObjects`` (post/array.py)
* ``LabelToContour`` (post/array.py)
* ``ScaleIntensity`` (intensity/array.py)
* ``ScaleIntensityFixedMean`` (intensity/array.py)
* ``ClipIntensityPercentiles`` (intensity/array.py)
* ``NormalizeIntensity`` (intensity/array.py)
* ``SavitzkyGolaySmooth`` (intensity/array.py)
* ``RandHistogramShift`` (intensity/array.py)
* ``KSpaceSpikeNoise`` (intensity/array.py)
* ``IntensityRemap`` (intensity/array.py)

No behavioral change: the resulting tensor is the same underlying data (``as_tensor()`` returns the wrapped ``torch.Tensor`` without copying).

### DIFF
--- a/monai/transforms/intensity/array.py
+++ b/monai/transforms/intensity/array.py
@@ -484,7 +484,7 @@ class ScaleIntensity(Transform):
 
         """
         img = convert_to_tensor(img, track_meta=get_track_meta())
-        img_t = img.as_tensor() if isinstance(img, MetaTensor) else cast(torch.Tensor, img)
+        img_t: NdarrayOrTensor = img.as_tensor() if isinstance(img, MetaTensor) else img
         ret: NdarrayOrTensor
         if self.minv is not None or self.maxv is not None:
             if self.channel_wise:
@@ -543,7 +543,7 @@ class ScaleIntensityFixedMean(Transform):
         factor = factor if factor is not None else self.factor
 
         img = convert_to_tensor(img, track_meta=get_track_meta())
-        img_t = img.as_tensor() if isinstance(img, MetaTensor) else cast(torch.Tensor, img)
+        img_t: NdarrayOrTensor = img.as_tensor() if isinstance(img, MetaTensor) else img
         ret: NdarrayOrTensor
         if self.channel_wise:
             out = []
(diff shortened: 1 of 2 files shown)


ericspod:
Hi @aymuos15 these changes are correct in that they are equivalent, but I don't really see the need to make them. The overhead of a few calls isn't great especially for code which isn't used extremely frequently, you'd need batch sizes on the order of thousands before optimisations like this in transforms are measurable. Furthermore, this doesn't adhere to the DRY principle in that the `if-else` expression is copied everywhere, this should be in a function instead, but honestly that's what `convert_to_tensor` is what we mean to use in these situations as the single source of truth. Perhaps a better idea is to improve `convert_to_tensor` but I still don't see the need here without some proof of performance gain.

aymuos15:
Hello. I definitely agree with that. Should've benchmarked this properly before opening the PR. Sorry about that. After benchmarking I do not think its worth changing anything. Will close this. Thank you.

PR #8848, opened 2026-05-12
fix: only instantiate CrossAttentionBlock when with_cross_attention=True

### LINKED ISSUE #8845: TransformerBlock instantiates CrossAttentionBlock even when with_cross_attention=False
**Describe the bug**
Hi ,

I noticed that TransformerBlock always instantiates cross-attention layers even when with_cross_attention=False.

Current implementation:

```

    hidden_size=hidden_size,
    num_heads=num_heads,
    dropout_rate=dropout_rate,
    qkv_bias=qkv_bias,
    causal=False,
)
```

but in forward:

```
if self.with_cross_attention:
    x = x + self.cross_attn(self.norm_cross_attn(x), context=context)
```

So when with_cross_attention=False the cross-attention modules are never used but their parameters still appear ( registered)  in model.parameters()

**To Reproduce**
With self.with_cross_attention =False  , 
```
 
    for name, p in model.named_parameters():
        if p.grad is None:
            print("[NO GRAD] ", name)
```
[NO GRAD] encoder.vit.vit.blocks.1.norm_cross_attn.weight
[NO GRAD] encoder.vit.vit.blocks.1.norm_cross_attn.bias
[NO GRAD] encoder.vit.vit.blocks.1.cross_attn.out_proj.weight
[NO GRAD] encoder.vit.vit.blocks.1.cross_attn.out_proj.bias
[NO GRAD] encoder.vit.vit.blocks.1.cross_attn.to_q.weight
[NO GRAD] encoder.vit.vit.blocks.1.cross_attn.to_q.bias
[NO GRAD] encoder.vit.vit.blocks.1.cross_attn.to_k.weight
[NO GRAD] encoder.vit.vit.blocks.1.cross_attn.to_k.bias
[NO GRAD] encoder.vit.vit.blocks.1.cross_attn.to_v.weight
[NO GRAD] encoder.vit.vit.blocks.1.cross_attn.to_v.bias

**Expected behavior**
There should be no params registered for the cross attention module , if the transformer block doesnt use cross attention.

chhayankjain:
Fixes #8845.                                                                                                                    
                                         
  ### Description                                                                                                                 
                                                                                                                                  
`TransformerBlock` previously instantiated `norm_cross_attn` and `cross_attn` unconditionally in `__init__`, even when          `with_cross_attention=False`. These unused modules registered dead parameters in `model.parameters()`, consuming memory without contributing to computation. The `forward()` method already had the correct guard (`if self.with_cross_attention:`), so the instantiation and the usage were inconsistent.

This fix wraps both instantiations in `if with_cross_attention:`, so the modules are only created when actually needed.

### Types of changes

### DIFF
--- a/monai/networks/blocks/transformerblock.py
+++ b/monai/networks/blocks/transformerblock.py
@@ -46,10 +46,19 @@ class TransformerBlock(nn.Module):
             dropout_rate (float, optional): fraction of the input units to drop. Defaults to 0.0.
             qkv_bias(bool, optional): apply bias term for the qkv linear layer. Defaults to False.
             save_attn (bool, optional): to make accessible the attention matrix. Defaults to False.
+            causal (bool, optional): whether to apply causal masking in self-attention. Defaults to False.
+            sequence_length (int | None, optional): sequence length required for causal masking. Defaults to None.
+            with_cross_attention (bool, optional): whether to include cross-attention layers that attend to an
+                external context tensor. When False, cross_attn is set to nn.Identity() so that the attribute
+                always exists for typing and checkpoint compatibility. Defaults to False.
             use_flash_attention: if True, use Pytorch's inbuilt flash attention for a memory efficient attention mechanism
                 (see https://pytorch.org/docs/2.2/generated/torch.nn.functional.scaled_dot_product_attention.html).
             include_fc: whether to include the final linear layer. Default to True.
             use_combined_linear: whether to use a single linear layer for qkv projection, default to True.
+
+        Raises:
+            ValueError: if dropout_rate is not in [0, 1].
+            ValueError: if hidden_size is not divisible by num_heads.
 
         """
 
@@ -79,14 +88,18 @@ class TransformerBlock(nn.Module):
         self.with_cross_attention = with_cross_attention
 
         self.norm_cross_attn = nn.LayerNorm(hidden_size)
-        self.cross_attn = CrossAttentionBlock(
-            hidden_size=hidden_size,
-            num_heads=num_heads,
-            dropout_rate=dropout_rate,
-            qkv_bias=qkv_bias,
-            causal=False,
-            use_flash_attention=use_flash_attention,
-        )
+        self.cross_attn: CrossAttentionBlock | nn.Identity
+        if with_cross_attention:
+            self.cross_attn = CrossAttentionBlock(
+                hidden_size=hidden_size,
+                num_heads=num_heads,
+                dropout_rate=dropout_rate,
+                qkv_bias=qkv_bias,
+                causal=False,
+                use_flash_attention=use_flash_attention,
+            )
+        else:
+            self.cross_attn = nn.Identity()
 
     def forward(
         self, x: torch.Tensor, context: torch.Tensor | None = None, attn_mask: torch.Tensor | None = None
--- a/monai/networks/nets/vit.py
+++ b/monai/networks/nets/vit.py
@@ -131,3 +131,41 @@ class ViT(nn.Module):
         if hasattr(self, "classification_head"):
             x = self.classification_head(x[:, 0])
         return x, hidden_states_out
+
+    def load_old_state_dict(self, old_state_dict: dict, verbose: bool = False) -> None:
+        """
+        Load a state dict from a ViT model trained with an older version of MONAI where
+        ``CrossAttentionBlock`` was unconditionally instantiated in ``TransformerBlock``
+        even when ``with_cross_attention=False``. Old checkpoints contain stale
+        ``blocks.{i}.cross_attn.*`` keys that are not present in the current model and
+        are automatically dropped.
+
+        Args:
+            old_state_dict: state dict from the older ViT model.
+            verbose: if True, print keys that are missing or unmatched. Defaults to False.
+        """
+        new_state_dict = self.state_dict()
+        if all(k in new_state_dict for k in old_state_dict):
+            if verbose:
+                print("All keys match, loading state dict.")
+            self.load_state_dict(old_state_dict)
+            return
+
+        if verbose:
+            for k in new_state_dict:
+                if k not in old_state_dict:
+                    print(f"key {k} not found in old state dict")
+            print("----------------------------------------------")
+            for k in old_state_dict:
+                if k not in new_state_dict:
+                    print(f"key {k} not found in new state dict")
+
+        # copy over all matching keys; stale cross_attn.* keys are left as unmatched
+        # leftovers and are not inserted into new_state_dict
+        for k in new_state_dict:
+            if k in old_state_dict:
+                new_state_dict[k] = old_state_dict.pop(k)
+
+        if verbose:
+            print("remaining keys in old_state_dict:", old_state_dict.keys())
+        self.load_state_dict(new_state_dict)
--- a/monai/networks/nets/vitautoenc.py
+++ b/monai/networks/nets/vitautoenc.py
@@ -133,3 +133,41 @@ class ViTAutoEnc(nn.Module):
         x = self.conv3d_transpose(x)
         x = self.conv3d_transpose_1(x)
         return x, hidden_states_out
+
+    def load_old_state_dict(self, old_state_dict: dict, verbose: bool = False) -> None:
+        """
+        Load a state dict from a ViTAutoEnc model trained with an older version of MONAI where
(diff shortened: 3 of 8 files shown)


monai/networks/blocks/transformerblock.py
```diff
@@ -78,15 +87,16 @@ def __init__(
...
-        self.norm_cross_attn = nn.LayerNorm(hidden_size)
-        self.cross_attn = CrossAttentionBlock(
-            hidden_size=hidden_size,
-            num_heads=num_heads,
-            dropout_rate=dropout_rate,
-            qkv_bias=qkv_bias,
-            causal=False,
-            use_flash_attention=use_flash_attention,
-        )
+        if with_cross_attention:
```
ericspod, line 92: We've had this issue with other classes with conditional components. If we don't take this branch then the members aren't created which causes issues with typing, Torchscript (though this isn't so much a concern anymore), and loading weights. The saved weights for this class will have those for `cross_attn` even if it's not used, so loading it with this updated version of the class will raise exceptions about unused keys. We've had to work around this with methods to [load old state dicts like this](https://github.com/Project-MONAI/MONAI/blob/586dea1556932b1d2214d61f28b6e139635ddcbe/monai/networks/nets/controlnet.py#L425). 

You would need to look at where this class is used and see if such adaptation is needed, but either way the `norm_cross_attn` and `cross_attn` members should always exist. Since `norm_cross_attn` is pretty lightweight I'd instantiate it always, and `cross_attn` should be `nn.Identity`.

chhayankjain: @ericspod  Thank you so much for the detailed explanation!                    

I have updated the implementation based on your feedback:                                                                         
                                                                                                                                  
  - norm_cross_attn is now always instantiated (lightweight, keeps state dict consistent)
  - cross_attn is set to nn.Identity() when with_cross_attention=False so the attribute always exists for typing and checkpoint compatibility

I also looked at DecoderOnlyTransformer in transformer.py (the only caller that uses with_cross_attention) and updated its load_old_state_dict method to guard the norm2 → norm_cross_attn key remapping so it only fires when the target key exists in the new state dict.

ericspod: Hi @chhayankjain this class is now what I'd suggest to do, however your fixes for loading the state dictionaries won't resolve the issue in other places where `TransformerBlock` is used. For example, [in ViT](https://github.com/Project-MONAI/MONAI/blob/8b785896b83f63628c46471d0545657eb3a7398a/monai/networks/nets/vit.py#L109) the `TransformerBlock` instantiation will have `with_cross_attention` with its default value of False, so your change to this class means `self.cross_attn` will be missing from the state dict. If you attempt to load old weights you'll get an exception saying there's unused weights. There would need to be a `load_old_state_dict` added for this class and others like it which would check if a key with `cross_attn` is missing from the new state dict and skip it if present in the state dict being loaded. That's what these methods do in existing classes. There would need to be a lot more work done in a number of places to make this change work, I agree it is better to not instantiate things you don't need, but perhaps it's too much work.

monai/networks/nets/transformer.py
```diff
@@ -147,7 +147,7 @@ def load_old_state_dict(self, old_state_dict: dict, verbose=False) -> None:
 
         # fix the renamed norm blocks first  norm2 -> norm_cross_attention , norm3 -> norm2
         for k in list(old_state_dict.keys()):
-            if "norm2" in k:
+            if "norm2" in k and k.replace("norm2", "norm_cross_attn") in new_state_dict:
```
ericspod: I'm not sure this fixes the issue. `norm_cross_attn` isn't modified in your PR, only `cross_attn` is sometimes `nn.Identity`. I think in the case of loading an old state dict with members for `cross_attn` you don't want this method will work as needed. The old state dict will have leftover members but these get ignored.

monai/networks/blocks/transformerblock.py
chhayankjain: @ericspod Thanks for the thorough explanation! I think I understand the scope of work needed.          
                                                                                                                                                               
  **The core issue**: Old checkpoints saved with the pre-fix code will always contain cross_attn.* keys (e.g. cross_attn.to_q.weight, cross_attn.to_k.weight, etc.) even for models where with_cross_attention=False, because CrossAttentionBlock was unconditionally instantiated. With the fix applied, those same models now
set cross_attn = nn.Identity() (no learnable parameters), so attempting to load old weights will raise an unexpected keys error.

  Classes that need a load_old_state_dict:
  - ViT (vit.py) — self.blocks is an nn.ModuleList
  - ViTAutoEnc (vitautoenc.py) — self.blocks is an nn.ModuleList
  - MaskedAutoEncoderViT (masked_autoencoder_vit.py) — both self.blocks (encoder) and self.decoder_blocks (decoder) are nn.Sequential

  DecoderOnlyTransformer is already handled in this PR.

  **My plan for each**: Follow the same pattern as the existing load_old_state_dict in DecoderOnlyTransformer, copy all matching keys first, then for any remaining old keys that match cross_attn.*, skip/drop them since these three classes always use with_cross_attention=False. norm_cross_attn.* doesn't need special handling since it is always instantiated in both old and new code and loads directly.

I'll extend this PR with those changes across all three classes.

Does this approach sound right to you, and have I missed any other affected classes?

ericspod: @chhayankjain this sounds like a good plan. I can review again when you're ready. Thanks!

chhayankjain: @ericspod Could you please take another look? I've made all the necessary changes:
                                                                                                                                                                                                                                                 
  - Added load_old_state_dict to ViT, ViTAutoEnc, and MaskedAutoEncoderViT to drop stale cross_attn.* keys from old checkpoints
  - Reverted the unnecessary guard in DecoderOnlyTransformer.load_old_state_dict
  - Added tests for all three new methods

Also seems like One CI test is failing (test_load_image) due to a transient network error in the GitHub Actions Windows runner,  it's unrelated to this PR. [Please correct me if I'm am wrong]

Thank you!

ericspod: I'm looking now, the CI failures are something I'm working on in another PR as well.

monai/networks/nets/vit.py
```diff
@@ -131,3 +131,40 @@ def forward(self, x):
...
+        ``blocks.{i}.cross_attn.*`` keys that are not present in the current model and
+        are automatically dropped.
+
+        Args:
+            old_state_dict: state dict from the older ViT model.
+            verbose: if True, print keys that are missing or unmatched. Defaults to False.
+        """
+        new_state_dict = self.state_dict()
+        if all(k in new_state_dict for k in old_state_dict):
+            print("All keys match, loading state dict.")
```
ericspod: This should print only in verbose mode.

monai/networks/nets/vitautoenc.py
```diff
@@ -133,3 +133,40 @@ def forward(self, x):
...
+        ``blocks.{i}.cross_attn.*`` keys that are not present in the current model and
+        are automatically dropped.
+
+        Args:
+            old_state_dict: state dict from the older ViTAutoEnc model.
+            verbose: if True, print keys that are missing or unmatched. Defaults to False.
+        """
+        new_state_dict = self.state_dict()
+        if all(k in new_state_dict for k in old_state_dict):
+            print("All keys match, loading state dict.")
```
ericspod: Same here.

monai/networks/blocks/transformerblock.py
chhayankjain: @ericspod  Thank you!

monai/networks/nets/transformer.py
chhayankjain: Addressed!

ericspod (changes requested):
Hi @chhayankjain I had some minor comments again, we're having to wait for the other PR I have to fix the tests, I'm on it now.

monai/networks/nets/vit.py
chhayankjain: Thank you!
Updated!

monai/networks/nets/vitautoenc.py
chhayankjain: Thank you!
Updated!

ericspod (approved):
Thanks, we have to wait for the fix however.

chhayankjain:
> Thanks, we have to wait for the fix however.

Will keep an eye on the dev branch and rebase when needed.
Thank you!

PR #8858, opened 2026-05-18
Add nested dot-notation access to ConfigParser (#6837)

### LINKED ISSUE #6837: Configparser to access nested attributes with dot notation
Follow up of https://github.com/Project-MONAI/MONAI/issues/5804, since v1.2 PR https://github.com/Project-MONAI/MONAI/pull/5813
```
data_file_base_dir = parser.data_file_base_dir
```
is an equivalent command of
```
data_file_base_dir = parser.get_parsed_content("data_file_base_dir")
```
which improves the code readability. However, the current implementation doesnt' support nested structures such as `parser.foo[0].bar.x`.

This is a feature request for extending the dot notation for accessing the nested config structures.

please see also:
- a possible implementation to integrate: easydict https://github.com/makinacorpus/easydict/blob/master/easydict/__init__.py
- configparser usage https://github.com/Project-MONAI/MONAI/blob/dev/docs/source/config_syntax.md

aymuos15:
Fixes #6837.

### Description

Follow-up to #5813: extends dot/bracket notation to nested config. Dict/list results from `ConfigParser.__getattr__` are wrapped in a `_ConfigProxy` that chains via `get_parsed_content`, so nested `$@ref`/`_target_` still resolve:

    parser.training.trainer.max_epochs  # = get_parsed_content("training::trainer::max_epochs")

`._raw` exposes the raw container; `len`/iter/`in`/`bool` and missing-key access delegate to it. Config keys take precedence over dict/list methods on attribute access (consistent with #5813).

### Types of changes
- [x] Non-breaking (caveat: `parser.x` dict/list is now `_ConfigProxy`; `isinstance(parser.x, dict)` no longer matches).
- [x] New tests added.

### DIFF
--- a/monai/bundle/config_parser.py
+++ b/monai/bundle/config_parser.py
@@ -113,6 +113,15 @@ class _ConfigProxy:
     def _child_id(self, key: str | int) -> str:
         return f"{self._id}{ID_SEP_KEY}{key}"
 
+    def _backing_id(self) -> str:
+        """Return the real config id this proxy writes to, resolving a single ``$@ref`` hop if needed."""
+        raw = self._parser[self._id]
+        if isinstance(raw, str):
+            refs = ReferenceResolver.match_refs_pattern(raw)
+            if refs:
+                return next(iter(refs))
+        return self._id
+
     def _chain(self, key: str) -> Any:
         """
         Resolve ``key`` as a nested config id.
@@ -156,12 +165,13 @@ class _ConfigProxy:
             return self._value[key]
 
     def __setitem__(self, key: str | int, value: Any) -> None:
-        # write through to the config source and reset the resolver so the change is
-        # visible from both ``parser.<id>`` and ``get_parsed_content``.
-        self._parser[self._child_id(key)] = value
+        # Resolve a potential $@ref so we write to the real backing config node.
+        backing = self._backing_id()
+        self._parser[f"{backing}{ID_SEP_KEY}{key}"] = value
 
     def __delitem__(self, key: str | int) -> None:
-        node = self._parser[self._id]
+        backing = self._backing_id()
+        node = self._parser[backing]
         del node[key if isinstance(node, dict) else int(key)]
         self._parser.ref_resolver.reset()
 
--- a/monai/bundle/config_parser.py
+++ b/monai/bundle/config_parser.py
@@ -114,13 +114,21 @@ class _ConfigProxy:
         return f"{self._id}{ID_SEP_KEY}{key}"
 
     def _backing_id(self) -> str:
-        """Return the real config id this proxy writes to, resolving a single ``$@ref`` hop if needed."""
-        raw = self._parser[self._id]
-        if isinstance(raw, str):
+        """Return the real config id this proxy writes to, resolving all ``$@ref`` hops transitively."""
+        current = self._id
+        seen: set[str] = set()
+        while True:
+            if current in seen:
+                break
+            seen.add(current)
+            raw = self._parser[current]
+            if not isinstance(raw, str):
+                break
             refs = ReferenceResolver.match_refs_pattern(raw)
-            if refs:
-                return next(iter(refs))
-        return self._id
+            if not refs:
+                break
+            current = next(iter(refs))
+        return current
 
     def _chain(self, key: str) -> Any:
         """
@@ -179,9 +187,13 @@ class _ConfigProxy:
         if key in _ConfigProxy._INTERNAL:
             object.__setattr__(self, key, value)
             return
+        if key == "_raw":
+            raise AttributeError("_raw is read-only")
         self[key] = value
 
     def __delattr__(self, key: str) -> None:
+        if key == "_raw":
+            raise AttributeError("_raw is read-only")
         del self[key]
 
     def __len__(self) -> int:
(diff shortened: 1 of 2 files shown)


tests/bundle/test_config_parser.py
```diff
@@ -326,6 +326,56 @@ def test_get_via_attributes(self):
         result = trans(np.ones(64))
         self.assertTupleEqual(result.shape, (1, 8, 8))
 
+    def test_nested_dot_notation(self):
```
ericspod: I think this should be broken up into multiple tests focused on specific areas. Checking exception raises should be separate, for example. You can put `config` into a `setUp` method to avoid copying it.

tests/bundle/test_config_parser.py
```diff
@@ -326,6 +326,56 @@ def test_get_via_attributes(self):
...
+            _ = parser.A.B.D[5]
+        with self.assertRaises(KeyError):
+            _ = parser.A.B["nonexistent"]
+
+        with self.assertRaises(AttributeError):
+            _ = parser.A.nonexistent
+
+        # item assignment/deletion writes through to the underlying container,
+        # preserving the pre-proxy behaviour where ``parser.x`` was the raw dict/list.
+        parser.A.B["C"] = 99
```
ericspod: I think `_ConfigProxy.__setattr__`  may need to be implemented or something else fixed. If I do `parser.A.X=[2,3]` this should create an item `X` in the dictionary `A` as defined in your test here, and I can see that dictionary changed in the return value for `parser.get_parsed_content("A")`. However, the expression `parser.A.X` raises `AttributeError: 'dict' object has no attribute 'X'`. I don't think the assignment is updating the underlying config structure correctly. You should add tests that attempt to add members this way then extract the value again, and also check against `get_parsed_content`.

ericspod (changes requested):
Hi @aymuos15 I think this solution generally works. It has the side-effect of changing the type of extracted values from a `ConfigParser` instance, eg. `type(parser.A)` is `_ConfigProxy` rather than `dict`. This would possibly change some behaviour for those checking the member type. The docstring for `__getattr__` mentions this behaviour but it should be explained in the docstring for the class along with an added example code of the notation. I had another question about behaviour in a comment below as well.

tests/bundle/test_config_parser.py
aymuos15: The nested-access tests are now in a dedicated `TestConfigProxy` class.

tests/bundle/test_config_parser.py
aymuos15: Thank you very much for pointing that out. Have addressed this now.

monai/bundle/config_parser.py
```diff
@@ -35,6 +36,179 @@
...
+        Raises:
+            AttributeError: if ``key`` is neither a config key nor an attribute of the
+                underlying ``dict``/``list``.
+        """
+        if key.startswith("__") and key.endswith("__"):
+            raise AttributeError(key)
+        try:
+            return self._chain(key)
+        except KeyError:
+            return getattr(self._value, key)
```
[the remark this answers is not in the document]
aymuos15: I dont think this happens in practice?

monai/bundle/config_parser.py
```diff
@@ -35,6 +36,189 @@
...
+    def __deepcopy__(self, memo: Any) -> Any:
+        return deepcopy(self._value, memo)
+
+    def __reduce__(self) -> Any:
+        return (_identity, (self._value,))
+
+    @property
+    def _raw(self) -> Any:
+        """The underlying ``dict``/``list`` (read-only view; mutate via the proxy or the parser)."""
+        return self._value
```
ericspod, line 234: As you return the actual `_value` object and not some read-only object or a deep copy, this or children objects can still be modified. The Coderabbit comment is about this docstring claiming it's read-only when it's only the variable `self._value` you can't change, but that's assumed for a `@property`.

ericspod:
My comments were addressed here and it looks better, but you should look at the Coderabbit remarks and check what should be addressed. The one about raw mutator methods maybe is fine to leave however, but either way resolve comments once they're addressed then I can review again.

monai/bundle/config_parser.py
```diff
@@ -35,6 +36,189 @@
...
+        self._parser.ref_resolver.reset()
+
+    def __setattr__(self, key: str, value: Any) -> None:
+        if key in _ConfigProxy._INTERNAL:
+            object.__setattr__(self, key, value)
+            return
+        self[key] = value
+
+    def __delattr__(self, key: str) -> None:
+        del self[key]
```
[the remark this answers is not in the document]
aymuos15: Seems to be okay based on the below discussion.

monai/bundle/config_parser.py
aymuos15: Thank you. Did this in the most recent commit

monai/bundle/config_parser.py
```diff
@@ -35,6 +36,201 @@
...
+            return self._chain(str(key))
+        except KeyError:
+            # no config key of that name: defer to the underlying dict/list so normal
+            # indexing semantics apply (IndexError, negative indices, dict KeyError).
+            return self._value[key]
+
+    def __setitem__(self, key: str | int, value: Any) -> None:
+        # Resolve a potential $@ref so we write to the real backing config node.
+        backing = self._backing_id()
+        self._parser[f"{backing}{ID_SEP_KEY}{key}"] = value
```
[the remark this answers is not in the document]
aymuos15: addressed in latest commit

ericspod (approved):
Looks good to me now!

aymuos15:
Thank you so much for the guidance and detailed review!

PR #8860, opened 2026-05-18
Only instantiate cross-attention layers when with_cross_attention=True

### LINKED ISSUE #8845: TransformerBlock instantiates CrossAttentionBlock even when with_cross_attention=False

AlexanderSanin:
## Summary
Fixes #8845

- `TransformerBlock` unconditionally instantiated `CrossAttentionBlock` and its `LayerNorm` even when `with_cross_attention=False`
- This registered unused parameters in the model, wasting memory and causing no-gradient warnings during training
- Wrapped the cross-attention layer creation in a conditional so the modules are only created when actually used

## Test plan
- [ ] Existing parameterized tests in `test_transformerblock.py` already cover both `with_cross_attention=False` and `True`
- [ ] Verify that `model.parameters()` no longer includes cross-attention weights when `with_cross_attention=False`
- [ ] Verify forward pass still works correctly with `with_cross_attention=True`

Signed-off-by: Oleksandr Sanin <alexaaander.sanin@gmail.com>

### DIFF
--- a/monai/__init__.py
+++ b/monai/__init__.py
@@ -131,7 +131,7 @@ try:
     # workaround related to https://github.com/Project-MONAI/MONAI/issues/7575
     if hasattr(torch.cuda.device_count, "cache_clear"):
         torch.cuda.device_count.cache_clear()
-except BaseException:
+except Exception:
     from .utils.misc import MONAIEnvVars
 
     if MONAIEnvVars.debug():
--- a/monai/apps/auto3dseg/data_analyzer.py
+++ b/monai/apps/auto3dseg/data_analyzer.py
@@ -341,7 +341,7 @@ class DataAnalyzer:
                     _label_argmax = True  # track if label is argmaxed
                     batch_data[self.label_key] = label.to(device)
                 d = summarizer(batch_data)
-            except BaseException as err:
+            except Exception as err:
                 if "image_meta_dict" in batch_data.keys():
                     filename = batch_data["image_meta_dict"][ImageMetaKey.FILENAME_OR_OBJ]
                 else:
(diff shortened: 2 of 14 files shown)


ericspod:
Hi @AlexanderSanin thanks for this contribution, but we have a PR #8848 addressing this issue already. We're still in the process of working out the issues with it, specifically to do with loading weights. If we don't create the cross-attention object then saved state dicts from unfixed versions of networks will have the weights for it and then raise an error. I think we should focus effort on the earlier PR as that's farther along and close this one.

PR #8861, opened 2026-05-18
Add HashMismatchError for download hash validation failures

### LINKED ISSUE #8832: Raise different exception when hash value of download is wrong

AlexanderSanin:
## Summary
Fixes #8832

As suggested by @ericspod in #8699, the hash check failure in `download_url` should raise a specific exception rather than a generic `RuntimeError`, so it isn't accidentally suppressed by `skip_if_downloading_fails`.

- Added `HashMismatchError(RuntimeError)` in `monai/apps/utils.py`
- `download_url` now raises `HashMismatchError` when `check_hash` fails
- `skip_if_downloading_fails` re-raises `HashMismatchError` immediately (hash mismatch = real failure, not network issue)
- Removed `"md5 check"` from `DOWNLOAD_FAIL_MSGS` since hash failures are now a distinct exception type
- Backward compatible: `HashMismatchError` inherits from `RuntimeError`, so existing `except RuntimeError` handlers still work

## Test plan
- [ ] Existing `TestDownloadUrl` test updated to expect `HashMismatchError` for bad hash values
- [ ] Network errors still correctly skip tests via `skip_if_downloading_fails`
- [ ] Hash mismatches no longer silently skipped in CI

Signed-off-by: Oleksandr Sanin <alexaaander.sanin@gmail.com>


monai/apps/utils.py
ericspod:
```suggestion
        HashMismatchError: When the hash validation of the ``filepath`` existing file fails.
```

Issue #8865, opened 2026-05-20
Pytype Deprecation


ericspod:
As [stated on pytype's repo](https://github.com/google/pytype/issues/1925) the focus of work is now elsewhere and the library won't support Python versions after 3.12. This is essentially deprecated then, and so we should look to remove it from our CI pipeline in place of alternatives (or nothing). The alternatives they provide to consider are:

mypy - https://github.com/python/mypy 
Pyright - https://github.com/microsoft/pyright 

Pyrefly - https://github.com/facebook/pyrefly 
ty - https://github.com/astral-sh/ty

aymuos15:
Would vote for pyrefly here as they are out of beta and have tensor shape support.

Although I do not find that much of an issue with ty either. However, still beta.

These are also much faster than mypy/pyrighy.

ericspod:
My quick experiment with ty was that it was very fast and found a number of things that should be addressed. It's still beta but I wonder about adding it as a non-required test anyway since the cost is so low.

aymuos15:
Agreed!

ericspod:
@garciadias has started on #8868, my feeling was to use mypy AND ty together, but we can discuss further. If ty is fast and accurate maybe it can be a precommit hook instead.

aymuos15:
Just out of curiosity, why not pyrefly? Especially because it's not out of beta. 

The reason Would help for my personal stuff too :)

ericspod:
We haven't really evaluated things entirely so there's no reason why not, I just hadn't looked at it. ty was just really fast and the first I got to.

aymuos15:
Okay, thanks! Ty sounds good to me. It's my default for prototyping.

garciadias:
Thank you for the suggestion, @aymuos15. I will take a look at Pyrefly and see how it fits with respect to the other tools. Thank you for your comments, @ericspod. I will go through that PR today. 🙏🏾

PR #8867, opened 2026-05-21
Further CICD Cleanup

### LINKED ISSUE #8864: macOS Testing Extremely slow
**Describe the bug**
Full dep tests for macOS are extremely slow and hampering PR reviewing turnaround times. On a recent PR the full-dep macOS reported a test runtime of 5822.766s, Windows 2415.079s, and Ubuntu 1418.648s. The min-dep tests showed a similar ratio. 

I would suggest dropping the macOS full-dep test but keep the min-dep tests. We haven't had significant issues caught by this test, anything reported in the future can be added to a min-dep test.

ericspod:
Fixes #8864 #8865.

### Description

This PR cleans up the CI scripts a little more and removes the full-dep test for macOS (for being very slow) and pytype checking (for being very slow and now deprecated for later Python versions). Some other fixes were included to get failing jobs running. 

Further work to clean up the actions used for testing should be done. Coverage should be included in the PR so need to get Codecov to report again. The `setupapp.yml` should probably be consolidated into `cicd_tests.yml` with reduced duplication of steps. The `cron.yml`, `pythonapp-gpu.yml`, and `blossom-ci.yml` perhaps should be removed as deprecated actions.

This also upgrades the download action version per dependabot, and adds `pytest` as a temporary dependency to get around a CuPy bug: https://github.com/Project-MONAI/MONAI/pull/8873#issuecomment-4541743975

Also a future PR should drop PyTorch 2.8 support from the CI actions since version 2.12 is out now.

### Types of changes

### DIFF
--- a/.github/workflows/pythonapp-gpu.yml
+++ b/.github/workflows/pythonapp-gpu.yml
@@ -3,12 +3,12 @@ name: premerge-gpu
 
 on:
   # quick tests for pull requests and the releasing branches
-  push:
-    branches:
-      - main
-      - releasing/*
-  pull_request:
-    types: [opened, synchronize, closed]
+  # push:
+  #   branches:
+  #     - main
+  #     - releasing/*
+  # pull_request:
+  #   types: [opened, synchronize, closed]
 
 concurrency:
   # automatically cancel the previously triggered workflows when there's a newer version
--- a/.github/workflows/pythonapp-gpu.yml
+++ b/.github/workflows/pythonapp-gpu.yml
@@ -1,14 +1,14 @@
 # Jenkinsfile.monai-premerge
 name: premerge-gpu
 
-on:
-  # quick tests for pull requests and the releasing branches
-  # push:
-  #   branches:
-  #     - main
-  #     - releasing/*
-  # pull_request:
-  #   types: [opened, synchronize, closed]
+# on:
+#   # quick tests for pull requests and the releasing branches
+#   push:
+#     branches:
+#       - main
+#       - releasing/*
+#   pull_request:
+#     types: [opened, synchronize, closed]
 
 concurrency:
   # automatically cancel the previously triggered workflows when there's a newer version
--- a/.github/workflows/weekly-preview.yml
+++ b/.github/workflows/weekly-preview.yml
@@ -5,38 +5,36 @@ on:
   - cron: "0 2 * * 0"  # 02:00 of every Sunday
 
 jobs:
-  flake8-py3:
+  static-checks:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         opt: ["codeformat", "pytype", "mypy"]
     steps:
+    - name: Clean unused tools
+      run: |
+        find  /opt/hostedtoolcache/* -maxdepth 0 ! -name 'Python' -exec rm -rf {} \;
+        sudo rm -rf /usr/share/dotnet
+        sudo rm -rf /usr/local/lib/android
+        sudo rm -rf /opt/ghc /usr/local/.ghcup
+        sudo docker system prune -f
+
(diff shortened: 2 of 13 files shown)


.github/workflows/weekly-preview.yml
```diff
@@ -5,38 +5,35 @@ on:
...
-      id: pip-cache
-      run: |
-        echo "datew=$(date '+%Y-%V')" >> $GITHUB_OUTPUT
-    - name: cache for pip
-      uses: actions/cache@v5
-      id: cache
-      with:
-        path: ~/.cache/pip
-        key: ${{ runner.os }}-pip-${{ steps.pip-cache.outputs.datew }}
+        cache: 'pip'
```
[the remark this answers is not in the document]
ericspod: @coderabbitai propose a commitable suggestion to address this issue.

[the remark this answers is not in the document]
ericspod: The actions versions were suggested by dependabot rather than using hashes like this, it seems that is the correct usage rather than pinning to a specific version with hashes.

.github/workflows/pythonapp-gpu.yml
```diff
@@ -1,14 +1,14 @@
...
 
-on:
-  # quick tests for pull requests and the releasing branches
-  push:
-    branches:
-      - main
-      - releasing/*
-  pull_request:
-    types: [opened, synchronize, closed]
+# on:
```
aylward, line 4: @ericspod  - why are the triggers for pythonapp-gui.yml and setupapp..yml being disabled?   Sorry, missing context.

tests/data/test_lmdbdataset.py
```diff
@@ -89,38 +88,17 @@ def __call__(self, data):
...
-                hash_func=json_hashing,
-            )
-            self.assertEqual(list(ds1), list(ds))
-            self.assertEqual(items, [[[]], [[0]], [[0, 1]], [[0, 1, 2]], [[0, 1, 2, 3]]])
-
-        self.assertTrue(isinstance(ds1.info(), dict))
+            ds = LMDBDataset(items, transform=_InplaceXform(), cache_dir=tempdir, hash_func=hash_func, progress=False)
+            self.assertEqual(items, [[[]], [[0]], [[0, 1]], [[0, 1, 2]], [[0, 1, 2, 3]]], "Input dataset mutated.")
+            self.assertNotEqual(items, list(ds), "Output data unmodified by transform.")
+            self.assertIsInstance(ds.info(), dict)
```
aylward, line 101: Drops test of repeated run of downloading the data.  Do you have a history of why ds and ds1 were created and compared previously?

aylward (approved):
Looks good - just two open questions, arising from my limited background knowledge.

Am approving, assuming questions won't result in chaanges.

.github/workflows/pythonapp-gpu.yml
ericspod: These tests were never actually being run, there's a `if: ${{ false }} ` clause below. I've commented these out so they don't appear in PRs anymore, but I wanted to preserve the file for future use with nightly GPU resources.

tests/data/test_lmdbdataset.py
ericspod: I don't actually, these were very old tests that worked until recently. I think they were always unsafe in that internal pointers in LMDB could have become stale, LMDB now restricts opening databases twice for this reason. Either way, with the new restriction these can't be run and testing that a database exists between openings is something that LMDB should be testing for rather than us.

ericspod:
Thanks so much for the help @aylward! If you've approved of these changes, I can disable the branch rules requiring pytype and macOS-latest to pass then integrate.

PR #8870, opened 2026-05-25
Fix diffusion inferers to support diffusers-style schedulers

### LINKED ISSUE #8597: Make MONAI generative  features compatible with huggingface diffusers library
The diffusers library has now a lot of new diffusion schedulers that might improve training quality. MONAI offers a reduced number of schedulers and schedulers features.
Of course, it would make sense to try incorporating those, but as this diffusers library has grown widely, and the code in MONAI was made relatively compatible with that from hugging face, I wonder whether we could make our DiffusionInferer and LatentDiffusionInferer classes compatible with the schedulers from the diffusers library. 

That you can use the schedulers from diffusers with MONAI inferer same as you use the MONAI schedulers. Example:
`
import diffusers
from monai.inferers import LatentDiffusionInferer
scheduler = diffusers.DDPMScheduler(
            num_train_timesteps=1000,
            beta_schedule='scaled_linear',
            beta_start=0.0015, beta_end=0.0205,
            prediction_type='epsilon',
            timestep_spacing='trailing',
        )

inferer = LatentDiffusionInferer(scheduler = scheduler, scale_factor=0.5)
`

I've only had to do a few tweaks to make it work. For instance, the 'step' function of the DDPMScheduler is by default returning a special class dictionary as output, which gives an error, so when calling this function on the diffusers scheduler you have to set parameter return_dict = False (argument that doesn't exist in the MONAI version).

Not use MONAI inferers at all. 

**Things to consider**
Perhaps we need to list which schedulers can be made compatible. I've only tried it with a simple DDPM one, I assume there are others that require more modifications.

ugbotueferhire:
Fixes #8597.

### Description

This PR adds compatibility for `diffusers`-style schedulers in MONAI's diffusion inferers (`DiffusionInferer` and `LatentDiffusionInferer`). 

Specifically, this includes:
- **Scheduler Step Normalization**: The sampling path in `inferer.py` now detects `return_dict` support on `scheduler.step`, requests tuple output when available, and normalizes both MONAI tuple returns and `diffusers`-style `prev_sample` outputs. This shared helper also benefits the ControlNet diffusion inferers.
- **Likelihood Path Fixes**: Fixed the likelihood paths to consistently use the specific `scheduler` instance passed into the call.
- **Regression Testing**: Added regression coverage in `test_diffusion_inferer.py` and `test_latent_diffusion_inferer.py` using a small diffusers-style DDPM shim that returns a `prev_sample` output object by default. Verified locally (`31 passed` and `46 passed`).

### Types of changes

### DIFF
--- a/monai/inferers/inferer.py
+++ b/monai/inferers/inferer.py
@@ -879,6 +879,60 @@ class DiffusionInferer(Inferer):
             return step_output.prev_sample
         raise TypeError("Unsupported scheduler.step output. Expected a tuple or an object with `prev_sample`.")
 
+    @staticmethod
+    def _get_scheduler_name(scheduler: Scheduler) -> str:
+        if hasattr(scheduler, "_get_name"):
+            return scheduler._get_name()
+        return scheduler.__class__.__name__
+
+    @staticmethod
+    def _get_scheduler_config_value(scheduler: Scheduler, name: str, default: Any = None) -> Any:
+        config = getattr(scheduler, "config", None)
+        if isinstance(config, Mapping):
+            if name in config:
+                return config[name]
+        elif config is not None and hasattr(config, name):
+            return getattr(config, name)
+
+        if hasattr(scheduler, name):
+            return getattr(scheduler, name)
+        return default
+
+    @staticmethod
+    def _get_posterior_mean(
+        scheduler: Scheduler, timestep: int | torch.Tensor, x_0: torch.Tensor, x_t: torch.Tensor
+    ) -> torch.Tensor:
+        alpha_t = scheduler.alphas[timestep]
+        alpha_prod_t = scheduler.alphas_cumprod[timestep]
+        alpha_prod_t_prev = scheduler.alphas_cumprod[timestep - 1] if timestep > 0 else scheduler.one
+
+        x_0_coefficient = alpha_prod_t_prev.sqrt() * scheduler.betas[timestep] / (1 - alpha_prod_t)
+        x_t_coefficient = alpha_t.sqrt() * (1 - alpha_prod_t_prev) / (1 - alpha_prod_t)
+
+        return x_0_coefficient * x_0 + x_t_coefficient * x_t
+
+    def _get_posterior_variance(
+        self, scheduler: Scheduler, timestep: int | torch.Tensor, predicted_variance: torch.Tensor | None = None
+    ) -> torch.Tensor:
+        alpha_prod_t = scheduler.alphas_cumprod[timestep]
+        alpha_prod_t_prev = scheduler.alphas_cumprod[timestep - 1] if timestep > 0 else scheduler.one
+        variance = (1 - alpha_prod_t_prev) / (1 - alpha_prod_t) * scheduler.betas[timestep]
+        variance_type = self._get_scheduler_config_value(scheduler, "variance_type")
+
+        if variance_type == "fixed_small":
+            variance = torch.clamp(variance, min=1e-20)
+        elif variance_type == "fixed_large":
+            variance = scheduler.betas[timestep]
+        elif variance_type == "learned" and predicted_variance is not None:
+            return predicted_variance
+        elif variance_type == "learned_range" and predicted_variance is not None:
+            min_log = variance
+            max_log = scheduler.betas[timestep]
+            frac = (predicted_variance + 1) / 2
+            variance = frac * max_log + (1 - frac) * min_log
+
+        return variance
+
     def _scheduler_step(
         self,
         scheduler: Scheduler,
@@ -1069,10 +1123,10 @@ class DiffusionInferer(Inferer):
 
         if not scheduler:
             scheduler = self.scheduler
-        if scheduler._get_name() != "DDPMScheduler":
+        scheduler_name = self._get_scheduler_name(scheduler)
+        if scheduler_name != "DDPMScheduler":
             raise NotImplementedError(
-                f"Likelihood computation is only compatible with DDPMScheduler,"
-                f" you are using {scheduler._get_name()}"
+                f"Likelihood computation is only compatible with DDPMScheduler," f" you are using {scheduler_name}"
             )
         if mode not in ["crossattn", "concat"]:
             raise NotImplementedError(f"{mode} condition is not supported")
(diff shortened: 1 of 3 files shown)


ericspod:
@virginiafdez could you please have a look at this one? Thanks!

tests/inferers/test_diffusion_inferer.py
```diff
@@ -55,6 +55,31 @@
...
+        prev_sample, pred_original_sample = super().step(
+            model_output=model_output, timestep=timestep, sample=sample, generator=generator
+        )
+        if return_dict:
+            return DiffusersLikeSchedulerOutput(
+                prev_sample=prev_sample, pred_original_sample=pred_original_sample
+            )
+        return prev_sample, pred_original_sample
+
+
```
virginiafdez: For a single method, I don't think this class is necessary. We can have this as a method on the main test class. But not 100% sure @ericspod

tests/inferers/test_latent_diffusion_inferer.py
```diff
@@ -313,6 +313,33 @@
...
+    ):
+        prev_sample, pred_original_sample = super().step(
+            model_output=model_output, timestep=timestep, sample=sample, generator=generator
+        )
+        if return_dict:
+            return DiffusersLikeSchedulerOutput(
+                prev_sample=prev_sample, pred_original_sample=pred_original_sample
+            )
+        return prev_sample, pred_original_sample
+
```
virginiafdez: Same as comment for test_diffusion_inferer.

virginiafdez (dismissed):
- The changes to `inferer` look good and make sense. 
- I would make a modification on the tests to make the changes less disruptive. 
- In addition, since the main point of this issue was to make MONAI inferer class compatible with the diffusers library, I would add a method on both testing files that queries an example scheduler from the `diffusers` library.

e.g.


```
from diffusers import DDPMScheduler as DiffusersDDPMScheduler
model = DiffusionModelUNet(
        spatial_dims=2,
        in_channels=1,
        out_channels=1,
        channels=[32, 64],
        attention_levels=[False, True],
        num_res_blocks=1,
        num_head_channels=32,
    ).to(device)

scheduler = DiffusersDDPMScheduler(
        num_train_timesteps=1000,
        beta_schedule="linear",
        prediction_type="epsilon",
    )
    scheduler.set_timesteps(num_inference_steps=50)
    inferer = DiffusionInferer(scheduler=scheduler)
    batch_size = 2
    image_size = 32
    inputs = torch.randn(batch_size, 1, image_size, image_size).to(device)
    noise = torch.randn_like(inputs)
    timesteps = torch.randint(0, scheduler.config.num_train_timesteps, (batch_size,)).long().to(device)
    
    with torch.no_grad():
        prediction = inferer(
            inputs=inputs,
            diffusion_model=model,
            noise=noise,
            timesteps=timesteps
        )
```

Another test: what happens if you import `DDPMScheduler` from the `diffusers` library and call `get_likelihood`? Will it work? Or something needs to do with renaming

tests/inferers/test_diffusion_inferer.py
ericspod: It hasn't been our style or practice with MONAI to create output classes like this, which here are essentially like namedtuple or dataclass. Was there a specific reason for this usage?

ugbotueferhire:
Hi @virginiafdez @ericspod , thanks for the review.

I removed the custom diffusers-style scheduler test shims and replaced them with optional tests using the real `diffusers.DDPMScheduler`. I also added coverage for `get_likelihood` with a diffusers scheduler and updated the inferer compatibility helpers so scheduler config values work with both MONAI and diffusers schedulers.

I also addressed the CodeRabbit comment by adding `DiffusionInferer.sample(...)` coverage for the real diffusers scheduler path.

Verified with:

```bash
python -m ruff check monai/inferers/inferer.py tests/inferers/test_diffusion_inferer.py tests/inferers/test_latent_diffusion_inferer.py
python -m pytest tests/inferers/test_diffusion_inferer.py -q
python -m pytest tests/inferers/test_latent_diffusion_inferer.py -q

File: tests/inferers/test_diffusion_inferer.py [hunk 3]
```diff
@@ -126,6 +127,68 @@ def test_ddpm_sampler(self, model_params, input_shape):
...
+        batch_size = 2
+        image_size = 32
+        inputs = torch.randn(batch_size, 1, image_size, image_size).to(device)
+        noise = torch.randn_like(inputs)
+        timesteps = torch.randint(0, scheduler.config.num_train_timesteps, (batch_size,)).long().to(device)
+        with torch.no_grad():
+            prediction = inferer(inputs=inputs, diffusion_model=model, noise=noise, timesteps=timesteps)
+
+        self.assertEqual(prediction.shape, inputs.shape)
```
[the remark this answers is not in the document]
ugbotueferhire: Resolved. The test now exercises both paths: `DiffusionInferer.__call__()` and `DiffusionInferer.sample()` with a real `diffusers.DDPMScheduler`, so the scheduler-step compatibility logic is covered.

monai/inferers/inferer.py
```diff
@@ -1029,10 +1124,10 @@ def get_likelihood(
 
         if not scheduler:
             scheduler = self.scheduler
-        if scheduler._get_name() != "DDPMScheduler":
+        scheduler_name = self._get_scheduler_name(scheduler)
+        if scheduler_name != "DDPMScheduler":
             raise NotImplementedError(
-                f"Likelihood computation is only compatible with DDPMScheduler,"
-                f" you are using {scheduler._get_name()}"
+                f"Likelihood computation is only compatible with DDPMScheduler," f" you are using {scheduler_name}"
```
ericspod:
```suggestion
                f"Likelihood computation is only compatible with DDPMScheduler, you are using {scheduler_name},"
```

monai/inferers/inferer.py
```diff
@@ -1632,10 +1738,10 @@ def get_likelihood(  # type: ignore[override]
...
```
ericspod:
```suggestion
                f"Likelihood computation is only compatible with DDPMScheduler, you are using {scheduler_name}."
```

ericspod (dismissed):
Hi @ugbotueferhire thanks for this update, sorry for the wait! It looks good now with some minor fixes.

ericspod (approved):

ericspod (approved):

PR #8872, opened 2026-05-25
fix(losses): register buffers in GlobalMutualInformationLoss

### LINKED ISSUE #8819: [Bug] `LocalNormalizedCrossCorrelationLoss`: kernel not registered as buffer — silent gradient tracking + wrong device placement
## Describe the bug

In `monai/losses/image_dissimilarity.py`, `LocalNormalizedCrossCorrelationLoss.__init__` contains two related problems in kernel initialization:

### Problem 1: Typo in `require_grads` (silent no-op)

```python
self.kernel = _kernel(self.kernel_size)
self.kernel.require_grads = False  # BUG: 'require_grads' is NOT a valid tensor attribute!
self.kernel_vol = self.get_kernel_vol()
```

`require_grads` (plural, with trailing `s`) is **not** a valid PyTorch tensor attribute. The correct attribute is `requires_grad`. This line silently does nothing — it creates a new Python attribute called `require_grads` on the tensor object instead of controlling gradient tracking. As a result, **the kernel tensor silently tracks gradients** in every forward pass, consuming unnecessary memory in the computation graph.

### Problem 2: Plain attribute assignment instead of `register_buffer`

Using `self.kernel = ...` (plain attribute) instead of `self.register_buffer("kernel", ...)` means:
- When the user calls `loss.to("cuda")`, `loss.cuda()`, or `loss.half()` on the loss module, **the kernel does NOT move to the target device** — it stays on CPU. This causes a device mismatch at runtime.
- The kernel is **not included** in `state_dict()` / `load_state_dict()` which leads to silent inconsistencies if checkpointing the loss object.

## To Reproduce

```python
import torch
from monai.losses import LocalNormalizedCrossCorrelationLoss

loss = LocalNormalizedCrossCorrelationLoss(spatial_dims=3, kernel_type="gaussian")

# Bug 1: require_grads typo silently does nothing
print(loss.kernel.requires_grad)   # True! Not False as intended
print(hasattr(loss.kernel, 'require_grads'))  # True -- spurious attribute created

# Bug 2: kernel not a registered buffer
print(dict(loss.named_buffers()))  # {} -- kernel is NOT here!
loss.cuda()
print(loss.kernel.device)  # cpu -- kernel did NOT move to GPU!
```

## Expected behavior

- `loss.kernel.requires_grad` should be `False`
- `loss.kernel` and `loss.kernel_vol` should appear in `loss.named_buffers()`
- After `loss.cuda()`, `loss.kernel.device` should be `cuda:0`

## Fix

Replace both assignments with `register_buffer`:

```python
self.register_buffer("kernel", _kernel(self.kernel_size))
self.register_buffer("kernel_vol", self.get_kernel_vol())
```

This is tracked in PR #8818.

## Environment

Affects all versions. Reproducible on MONAI `dev` branch as of 2026-04-11.

## Related chain of issues

This bug reveals a broader pattern worth auditing across the MONAI losses module:
1. `GlobalMutualInformationLoss` — check if `bin_centers` is properly registered as a buffer
2. Other custom loss classes that use constant tensors initialized in `__init__`
3. Test coverage for device movement of loss modules (`loss.cuda()` should not cause device mismatch)

AlexanderSanin:
## Summary

- `GlobalMutualInformationLoss` stored `preterm` and `bin_centers` as plain tensor attributes when `kernel_type="gaussian"`, so calling `loss.to("cuda")` or `loss.cuda()` did **not** move them to the target device
- Replace the plain assignments with `register_buffer(..., persistent=False)`, consistent with the pattern already applied to `LocalNormalizedCrossCorrelationLoss` in #8818
- The `.to(img)` calls in `parzen_windowing_gaussian` are retained for dtype coercion (e.g. float16 inference)

## Test plan

- [x] `python -m pytest tests/losses/image_dissimilarity/test_global_mutual_information_loss.py -v` — all existing tests still pass
- [x] `TestGlobalMutualInformationLossBuffers::test_gaussian_kernel_registers_buffers` — `preterm` and `bin_centers` are in `_buffers` and have `requires_grad=False`
- [x] `TestGlobalMutualInformationLossBuffers::test_bspline_kernel_has_no_gaussian_buffers` — b-spline mode is unaffected
- [x] `TestGlobalMutualInformationLossBuffers::test_gaussian_kernel_forward_correct` — forward pass returns a scalar loss

Closes #8819


AlexanderSanin:
Hey @ericspod @aymuos15. Could you, please, have a look at this?

aymuos15:
Happy to go through this. Any idea why the CI is failing?

AlexanderSanin:
Hi @aymuos15! The CI failures are pre-existing on the `dev` branch itself and are not related to this PR.

The root cause is that `monai/networks/trt_compiler.py` imports `polygraphy.backend.common`, which transitively imports `cupy.testing._random`, which does `import pytest`. In the `packaging` / `full-dep (ubuntu-latest)` CI jobs, `pytest` is not installed in the target environment (it's only in `requirements-dev.txt`), so any import of `monai.networks` triggers a `ModuleNotFoundError` and causes cascade failures across ~6759 tests.

You can verify by checking the most recent `dev` branch CI run — it shows the exact same `FAILED (failures=17, errors=6759, skipped=1020)` result on the same jobs.

All tests specific to this PR pass in CI (confirmed in the `packaging` job log):
```
test_bspline_kernel_has_no_gaussian_buffers ... ok
test_gaussian_buffers_move_with_module      ... ok
test_gaussian_kernel_forward_correct        ... ok
test_gaussian_kernel_registers_buffers      ... ok
test_ill_opts_{0..3}                        ... ok
test_ill_shape_{0..1}                       ... ok
```

I've also rebased the branch on the latest `dev` just now.

aymuos15:
Okay! Thanks a lot for the detailed reply.

Let's wait for that to get an upstream fix then? I think that is being tracked and worked on in real time.

AlexanderSanin:
> Okay! Thanks a lot for the detailed reply.
> 
> Let's wait for that to get an upstream fix then? I think that is being tracked and worked on in real time.

Hi @aymuos15 ,

I've addressed that in another PR:  https://github.com/Project-MONAI/MONAI/pull/8873
Please review , thanks

AlexanderSanin:
## Status & Blockers

**All code changes are complete and tests pass.** Two things are blocking merge:

### 1. CI Failures (external regression — not our code)

The failing CI jobs (`full-dep (ubuntu-latest)`, `static-checks (pytype)`, `mypy`, `codeformat`) are caused by a **`cupy-cuda12x` 14.1.0 regression** released on May 26, 2026. That version introduced `import pytest` at module load time in `cupy/testing/_random.py`, which breaks any environment where `pytest` is not installed.

The failure chain is:
`monai.networks` → `trt_compiler` → `polygraphy.backend.common` → `polygraphy.util.util` → `cupy.testing._random` → `import pytest` → `ModuleNotFoundError`

This is **pre-existing on the `dev` branch** — the last fully green dev CI run (#26286623163) was on May 22 with cupy 14.0.1. All our PR-specific tests pass in the packaging job logs.

### 2. Awaiting Required Review

This PR needs at least one approving review from a code owner before it can be merged.

@KumoLiu @ericspod @Nic-Ma — could one of you take a look when you get a chance? The change is in `monai/losses/image_dissimilarity.py` and registers `preterm` / `bin_centers` as non-persistent buffers so they move correctly with `.to(device)` / `.cuda()`. Thanks!

aymuos15:
Hey, thanks for the fix there. But I think that is a deeper fix and one the maintainers should do. Will get to both when they are done.

ericspod:
Hi @AlexanderSanin a previous PR #8869 was looking at the same issue. Yours does something slightly different but we should merge that one first then integrate your changes and tests. I've updated so the tests should now run.

AlexanderSanin:
Thanks @ericspod, that plan works for me — happy to have #8869 go in first.

I've rebased this branch onto the latest `dev` (so it now picks up the CI fixes) and resolved the one remaining `static-checks (mypy)` failure: `register_buffer` left mypy inferring the broad `Tensor | Module` union for `preterm`/`bin_centers`, which broke the arithmetic in `parzen_windowing_gaussian`. Added explicit `torch.Tensor` annotations to fix it.

A couple of things this PR does on top of #8869, in case you'd like to fold them in after merging it:

- Registers **both** `preterm` and `bin_centers` as non-persistent buffers (#8869 registers only `bin_centers`), so the whole gaussian-kernel state moves consistently with `.to(device)` / `.cuda()`.
- Adds a focused test class (`TestGlobalMutualInformationLossBuffers`) covering: the non-persistent contract (buffers present in `_buffers` but excluded from `state_dict()`), the b-spline kernel registering no gaussian buffers, forward correctness, and device movement.

I'm glad to rebase these tests/changes directly on top of #8869 once it lands so they apply cleanly — just let me know your preference.

aymuos15:
Since `.to(img)` has to stay for the dtype coercion (it handles device too) and the results are bitwise accurate irrespective of the change, maybe this change is not required because we cant get rid of the .to(img) anyways?

AlexanderSanin:
Hi @aymuos15, great question! You're right that `.to(img)` handles device too — but `register_buffer` is still needed for correct module semantics, and here's why:

**Without `register_buffer`:**
```python
loss = GlobalMutualInformationLoss(kernel_type="gaussian")
loss = loss.to("cuda")      # moves all registered parameters/buffers — but preterm/bin_centers are plain attributes → stay on CPU
pred = pred.cuda()
target = target.cuda()
result = loss(pred, target)  # .to(img) allocates a new CUDA copy on every forward call
```
Every forward pass silently allocates a fresh copy of `preterm`/`bin_centers` on GPU. For repeated inference this is wasteful, and `loss.preterm.device` would still report `cpu` even after `loss.to("cuda")` — misleading for anyone inspecting the module's state.

**With `register_buffer(..., persistent=False)`:**
```python
loss = GlobalMutualInformationLoss(kernel_type="gaussian")
loss = loss.to("cuda")      # moves preterm and bin_centers to CUDA once
result = loss(pred, target)  # .to(img) is now a device no-op; only coerces dtype if needed (e.g. float16)
```
`loss.preterm.device` correctly shows `cuda`, and the buffers are allocated once — not once per forward call.

**Why keep `.to(img)` at all?**
Exactly as you noted — for dtype coercion. If the user feeds `float16` inputs the buffers need to match. After this change `.to(img)` only does a dtype cast (cheap), never a device transfer (because device already matches after `loss.to("cuda")`).

**Summary:** `register_buffer` upholds the contract that `.to(device)` fully moves a module's non-trainable state, while `.to(img)` remains as a safety net for dtype. Both are needed but for different reasons.

Also pushed a small follow-up commit: replaced the bare `return` in `test_gaussian_buffers_move_with_module` with `self.skipTest("CUDA not available")` so pytest correctly marks the test as skipped rather than emitting a warning.

aymuos15:
I do agree with the unnecessary calls, but that is completley neglible if you benchmark it right?

AlexanderSanin:
You're right — the overhead is negligible in practice. A few bytes re-pinned to GPU on each forward call won't show up in any real benchmark.

The real reason for `register_buffer` isn't performance, it's **API correctness**: PyTorch's contract for `nn.Module` is that `module.to(device)` moves *all* non-trainable module state. Without `register_buffer`, a user who writes:

```python
loss = GlobalMutualInformationLoss(kernel_type='gaussian').to('cuda')
```

is left with `loss.preterm` still reporting `cpu` — inconsistent with how every other `nn.Module` buffer behaves, and likely confusing to anyone inspecting the module or writing device-agnostic code that introspects `loss.buffers()`.

`register_buffer` is the documented PyTorch mechanism for exactly this case: a non-trainable tensor that is conceptually part of the module's state and should move with it. The non-persistent flag keeps it out of `state_dict()` since it's deterministically derived from the constructor args and doesn't need to be saved.

Happy to defer to maintainer preference on whether that level of correctness is worth the diff, but that's the motivation behind the change.

monai/losses/image_dissimilarity.py
```diff
@@ -233,10 +233,14 @@ def __init__(
         self.kernel_type = look_up_option(kernel_type, ["gaussian", "b-spline"])
         self.num_bins = num_bins
         self.kernel_type = kernel_type
-        self.bin_centers: torch.Tensor | None
+        # declared as buffers so they move with the module (e.g. ``.to(device)``); only populated for the
+        # gaussian kernel, hence the ``Tensor`` annotation reflects the type at the use sites in that path.
+        self.preterm: torch.Tensor
+        self.bin_centers: torch.Tensor
```
ericspod:
```suggestion
        self.preterm: torch.Tensor | None
        self.bin_centers: torch.Tensor | None
```

ericspod (dismissed):
Hi @AlexanderSanin I sorted a merge conflict and made a minor annotated type change, it looks good now if you're happy with that. I will merge now if tests pass but we can make later changes.

monai/losses/image_dissimilarity.py
ericspod:
```suggestion
        if self.bin_centers is None or self.preterm is None:
            raise ValueError("bin_centers and preterm must be defined for gaussian parzen windowing.")
```

monai/losses/image_dissimilarity.py
```diff
@@ -317,7 +321,8 @@ def parzen_windowing_gaussian(self, img: torch.Tensor) -> tuple[torch.Tensor, to
         img = torch.clamp(img, 0, 1)
         img = img.reshape(img.shape[0], -1, 1)  # (batch, num_sample, 1)
         if self.bin_centers is None:
```
ericspod:
```suggestion
```

ericspod (approved):

PR #8875, opened 2026-05-27
Adding allow_pickle argument

ericspod:
Addresses [GHSA-wg9g-w2j2-8pgr](https://github.com/Project-MONAI/MONAI/security/advisories/GHSA-wg9g-w2j2-8pgr). 

### Description

This adds `allow_pickle=False` argument to `NumpyReader`. This addresses the security concern but allows users to override the setting to load pickled data if needed.

The version of mlflow was also fixed to exclude version 3.13 temporarily until #8891 is addressed.

### Types of changes

### DIFF
--- a/monai/data/image_reader.py
+++ b/monai/data/image_reader.py
@@ -1232,13 +1232,18 @@ class NumpyReader(ImageReader):
             stack the loaded items together to construct a new first dimension.
         channel_dim: if not None, explicitly specify the channel dim, otherwise, treat the array as no channel.
         allow_pickle: if True, allows loading pickled contents from NPY/NPZ files. Note that the default value of False
-            prevents the risk of remote code execution, set this to True only for loading known trusted data.
+            prevents the risk of remote code execution, set this to True only for loading known trusted data. If this
+            argument is False and pickled data is loaded, a ValueError will be raised.
         kwargs: additional args for `numpy.load` API except `allow_pickle`. more details about available args:
             https://numpy.org/doc/stable/reference/generated/numpy.load.html
     """
 
     def __init__(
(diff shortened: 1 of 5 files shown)


monai/data/image_reader.py
```diff
@@ -1231,17 +1231,26 @@ class NumpyReader(ImageReader):
...
             https://numpy.org/doc/stable/reference/generated/numpy.load.html
-
     """
 
-    def __init__(self, npz_keys: KeysCollection | None = None, channel_dim: str | int | None = None, **kwargs):
+    def __init__(
+        self,
+        npz_keys: KeysCollection | None = None,
+        channel_dim: str | int | None = None,
+        allow_pickle: bool = False,
```
garciadias, line 1245: The default change from the implicit `allow_pickle=True` (hardcoded in the previous `np.load` call) to `allow_pickle=False` is a **breaking change** for any caller who loads `.npy`/`.npz` files containing Python objects. The PR template marks this as non-breaking, and the commit message does not include a migration note. This will silently raise `ValueError` for existing users — they need to be told to pass `allow_pickle=True` explicitly, and CHANGELOG.MD, I think, would be worth mentioning in the release notes.

garciadias (dismissed):
I have just one small comment. Thank you, Eric!

monai/data/image_reader.py
ericspod: I've added an exception raise that will note this has been changed. I think it's very unlikely anyone is loading pickled data this way but the exception doesn't hurt to add. Since this wasn't an existing argument to the class I couldn't rightly add a deprecation decorator.

garciadias (approved):
Perfect. This is a very clear message.

PR #8890, opened 2026-06-01
Fix 3D decomposed relative position embedding for anisotropic inputs

aymuos15:
The 3D branch of add_decomposed_rel_pos was broken for non-cubic spatial sizes in two ways:

1. rel_d contracted the depth-axis embedding against the width axis ('bhwdc,wkc->bhwdk' instead of 'bhwdc,dkc->bhwdk').
2. The broadcast-add into the (b, qh, qw, qd, kh, kw, kd) attention tensor used 4 leading colons instead of 5 for each rel_* term, so the key axis landed in the wrong slot (e.g. rel_h's kh fell into the kd position).

Both were masked by the only 3D test using a cubic input_size (8, 8, 8): with kh == kw == kd every misplaced axis still broadcasts. For anisotropic volumes the forward pass raised a shape error (or, where conformable, added the wrong positional bias) in SABlock/CrossAttentionBlock with rel_pos_embedding='decomposed'.

Adds a regression test parametrized over several anisotropic 3D input sizes, including permutations of distinct dims so each spatial axis is exercised in a different position.

### Types of changes


tests/networks/blocks/test_selfattention.py
ericspod:
```suggestion
            "use_combined_linear": False,
```
This is something Coderabbit mentioned as an existing bug, it would be good to fix it here as well.

ericspod (commented):
Hi @aymuos15 this looks good to me, I commented on one minor change we can integrate here.

PR #8893, opened 2026-06-01
ci: mute dynamic __warningregistry__ logs noise in pytest suite

### LINKED ISSUE #8892: Excessive Warning Printing During Tests
When running tests, thousands of lines like this appear in the logs:

```
Accessing `__warningregistry__` from `.models.video_llava.image_processing_video_llava`. Returning `__warningregistry__` instead. Behavior may be different and this alias will be removed in future versions.
```

This appears unrelated to MONAI or transformers but related to using `unittest.TestCase.assertWarns` method. This method using a context manager that messes around with `__warningregistry__` in places, but it's unclear why these get printed now and it's not something seen before.

Priyanshu31102003:
Fixes #8892

### Description
During automated test execution, the logs were getting flooded with thousands of verbose lines when `unittest.TestCase.assertWarns` internally triggered property accessors for `__warningregistry__` on dynamic configurations.

This PR globally configures `pytest` inside `setup.cfg` to ignore and suppress regex pattern filter hits matching `.*__warningregistry__.*`, significantly cleaning up the CI/CD pipeline output logs.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -64,6 +64,8 @@ hf_local_entry_error, _has_hf_local = optional_import("huggingface_hub.errors", 
 quick_test_var = "QUICKTEST"
 _tf32_enabled = None
 _test_data_config: dict = {}
+# Fix dynamic warningregistry logs noise in python unit/pytest configurations
+warnings.filterwarnings("ignore", message="Accessing.*__warningregistry__")
 
 MODULE_PATH = Path(__file__).resolve().parents[1]
 


ericspod:
Hi @Priyanshu31102003 thanks for looking into this. Was there no indication of what was causing this issue? I imagine the warning suppression done here will work but I don't see the messages now in logs from recent commits (nor this one). I would be interesting in fixing the underlying issue if it was something on our side. I'm good with this fix now however.

ericspod (approved):

PR #8896, opened 2026-06-03
tests: remove duplicate and dead test cases

aymuos15:
### Description

Test-only cleanup. Removes byte-identical duplicate cases and one permanently-skipped test. No production code or test logic changes.

### Changes

Duplicate within-file parametrized cases (identical entries that asserted the same thing twice, so removal changes nothing covered):

- `tests/losses/test_unified_focal_loss.py`: both `TEST_CASES` entries were identical
- `tests/inferers/test_sliding_window_inference.py`: duplicate `3D small roi` case
- `tests/losses/test_dice_loss.py`: `sigmoid` case repeated
- `tests/losses/test_generalized_dice_loss.py`: `sigmoid` case repeated
- `tests/transforms/utility/test_apply_transform_to_pointsd.py`: duplicate entry

Dead test removed:

- `tests/apps/test_download_url_yandex.py`: removed `test_verify` and its now-unused `YANDEX_MODEL_URL`. It was permanently `@unittest.skip`-ed ("data source unstable") and hits an external Yandex URL, so it never runs in CI. The error path stays covered by `test_verify_error`.

### Types of changes
- [x] Non-breaking change (test-only cleanup)
- [x] In-line documentation / comments updated as needed
- [x] All tests passing locally


ericspod:
@coderabbitai where else in the unit tests are there duplicate test cases like those removed in this PR?

ericspod (approved):
Hi @aymuos15 thanks for the cleanup, I'm going to merge now but from the Coderabbit response there may be other similar places that can be improved in #8942.

aymuos15:
I'll make a follow up PR for this, Thank you for pointing that out!

PR #8897, opened 2026-06-04
Minor Fixes From Testing 1.6

ericspod:
### Description

This fixes the install issue in the Docker image seen earlier with MetricsReloaded, and fixes a serialisation issue related to moving from pickle to json for Auto3DSeg.

### Types of changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,8 +30,8 @@ WORKDIR /opt/monai
 COPY requirements.txt requirements-min.txt requirements-dev.txt /tmp/
 RUN cp /tmp/requirements.txt /tmp/req.bak \
   && awk '!/torch/' /tmp/requirements.txt > /tmp/tmp && mv /tmp/tmp /tmp/requirements.txt \
-  && python -m pip install --upgrade --no-cache-dir pip \
-  && python -m pip install --no-cache-dir -r /tmp/requirements-dev.txt
+  && python -m pip install --upgrade --no-cache-dir --no-build-isolation pip wheel wheel-stub \
+  && python -m pip install --no-cache-dir --no-build-isolation -r /tmp/requirements-dev.txt
 
 # compile ext and remove temp files
 # TODO: remark for issue [revise the dockerfile #1276](https://github.com/Project-MONAI/MONAI/issues/1276)
--- a/monai/apps/auto3dseg/bundle_gen.py
+++ b/monai/apps/auto3dseg/bundle_gen.py
@@ -379,6 +379,7 @@ class BundleAlgo(Algo):
             based on which path successfully imports the Algo class.
         """
         return {
+            "template_path": self.template_path,
             "data_stats_files": self.data_stats_files,
             "data_list_file": self.data_list_file,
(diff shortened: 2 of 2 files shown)


monai/apps/auto3dseg/bundle_gen.py
```diff
@@ -378,6 +379,7 @@ def state_dict(self) -> dict:
             based on which path successfully imports the Algo class.
```
garciadias, line 379: Adding template_path to state_dict is the correct fix for the JSON serialization issue described in the PR.

Just flagging: is there a corresponding load_state_dict path that reads template_path back? If the deserialization uses a generic loop over state_dict keys (e.g. via ConfigParser), this should work automatically. But if it relies on method-based deserialization, the loading side would need an update too. Worth a quick verification.

Dockerfile
```diff
@@ -30,8 +30,8 @@ WORKDIR /opt/monai
 COPY requirements.txt requirements-min.txt requirements-dev.txt /tmp/
 RUN cp /tmp/requirements.txt /tmp/req.bak \
   && awk '!/torch/' /tmp/requirements.txt > /tmp/tmp && mv /tmp/tmp /tmp/requirements.txt \
-  && python -m pip install --upgrade --no-cache-dir pip \
-  && python -m pip install --no-cache-dir -r /tmp/requirements-dev.txt
+  && python -m pip install --upgrade --no-cache-dir --no-build-isolation pip wheel wheel-stub \
+  && python -m pip install --no-cache-dir --no-build-isolation -r /tmp/requirements-dev.txt
 
 # compile ext and remove temp files
```
garciadias, line 36: wheel-stub is a PEP 517 build backend for external wheel hosting (PEP 759) that redirects pip to download wheels from a 3rd-party index instead of PyPI. This is unrelated to building MONAI or MetricsReloaded from source.

--no-build-isolation and wheel both make sense here, but wheel-stub is an unnecessary extra dependency that increases build time and image size, and could fail if transitive dependencies conflict.

Recommendation: Remove wheel-stub unless there is a specific documented reason. If the intent was just to ensure wheel is present, wheel alone suffices.

garciadias: I think I remember seeing this before, and I agree it may be required. A little more documentation could help.

garciadias (approved):

Dockerfile
ericspod: I had to add it somewhere to get around the issue with MetricsReloaded, so I think it is needed. It's something we'll revisit for the next release for sure, along with a revamp of the build process for MONAI.

monai/apps/auto3dseg/bundle_gen.py
ericspod: It would be [this method](https://github.com/Project-MONAI/MONAI/blob/2a7d0cf05c82dde97fd4b26f5fe9b9ab4e2100fe/monai/apps/auto3dseg/bundle_gen.py#L392), which loops through values and uses `setattr` to set state. This could be more robust but state dicts are only expected to be generated by the same class and used immediately, this isn't about loading stored data from an external source.

PR #8898, opened 2026-06-05
Feat/channel wise transforms

### LINKED ISSUE #8311: Implementing Channel-Wise Transforms

It is common practice to concatenate different images along the channel axis before feeding them into a model (early fusion). However (to the best of my knowledge), applying data augmentation channel-wise in the current setup is not straightforward. Introducing a helper class that wraps a MONAI transform and applies it along a specified axis could be helpful.

In my case, I had different 3D volumes concatenated along the channel axis, resulting in a shape of `C x H x W x D`. My solution was to create a wrapper transform as shown below.

**P.S.: Apologies for not adhering to the MONAI coding guidelines—this was a quick prototype.**

```python
from monai.transforms import MapTransform, RandomizableTransform


class RandChannelWiseApply(RandomizableTransform):
    
    def __init__(self, transform_to_wrap, prob=1):
        RandomizableTransform.__init__(self, prob)
        self.transform_to_wrap = transform_to_wrap
        
    def __call__(self, x):
        x = x.clone()
        if len(x.shape) != 4:
            raise Exception("Input tensor must be of shape C x H x W x D")
        for ch in range(x.shape[0]):
            x[ch] = self.transform_to_wrap(x[ch])
        return x
            

class RandChannelWiseApplyD(MapTransform, RandomizableTransform):
    
    def __init__(self,
                 keys, 
                 transform_to_wrap, 
                 prob=1,
                 allow_missing_keys=False):
        
        MapTransform.__init__(self, keys, allow_missing_keys)
        RandomizableTransform.__init__(self, 1)
        self.transform = RandChannelWiseApply(transform_to_wrap, prob)

    def __call__(self, data):
        for key in self.keys:
            if key in data:
                data[key] = self.transform(data[key])
            elif self.allow_missing_keys:
                continue
            else:
                raise Exception(f'Key {key} is missing.') 
        return data
```

Let me know if you think implementing this in MONAI would be useful. I’d be happy to contribute to the library.

1. An alternative is to concatenate all the images after applying the augmentation. However, this approach becomes challenging when the images are already saved in a concatenated format.

2. It’s quite possible that MONAI already offers an alternative way to achieve this that I’m not aware of.

ugbotueferhire:
Fixes #8311.

### Description

Adds new wrapper transforms `ChannelWise`, `RandChannelWise`, `ChannelWised`, and `RandChannelWised` to independently apply an array-based transform to each channel of an input array. This resolves issues surrounding applying data augmentations channel-wise, which is a common requirement for early fusion models where different 3D volumes or modalities are concatenated along the channel axis. 

The `ChannelWise` transform ensures the inner transform receives slices with a singleton channel dimension to maintain expected shape invariants, and successfully maintains independent PRNG states for random augmentations per-channel.

### Types of changes


ugbotueferhire:
Hi @ericspod please can you check this out?

aymuos15:
`ChannelWise(ScaleIntensity())` duplicates the existing `channel_wise=True` flag already on most intensity transforms (`ScaleIntensity`, `NormalizeIntensity`, `RandShiftIntensity`, etc.)

I think that should be the base (its also more performant than the loop approach here)

ugbotueferhire:
> `ChannelWise(ScaleIntensity())` duplicates the existing `channel_wise=True` flag already on most intensity transforms (`ScaleIntensity`, `NormalizeIntensity`, `RandShiftIntensity`, etc.)
> 
> I think that should be the base (its also more performant than the loop approach here)


Thanks, that’s a fair point. For transforms that already support `channel_wise=True`, that should remain the preferred path since it avoids the Python-level loop and is more efficient.

The intent of this wrapper is to cover the more general case: arbitrary array transforms, custom callables, or transforms that don’t expose a `channel_wise` option but still need to be applied independently per channel. `ScaleIntensity` is only used in the tests as a simple deterministic transform, not as the primary motivating use case.

I can add wording to the docstring to make that distinction explicit: use native `channel_wise=True` when available; use `ChannelWise`/`RandChannelWise` as a generic fallback wrapper.

aymuos15:
Why would we want a fallback there? And is it not better to just expand what already exists?

PR #8905, opened 2026-06-08
Make NormalizeIntensity invertible (#5647)

### LINKED ISSUE #5647: Add the possibility to compute the NormalizeIntensity inverse
### Discussed in https://github.com/Project-MONAI/MONAI/discussions/5645

<sup>Originally posted by **sh3rlock14** December  5, 2022</sup>
When applying the NormalizeIntensity transform (w/o passing subtrahend and divisor), the computed `_sub `and `_div ` should be added to the meta information. This would be helpful in getting back the original intensity values: not only this could be useful for recon. problems, but also for logging.
</div>

azrabano23:
Fixes #5647

### Description

`NormalizeIntensity` (and its dict wrapper `NormalizeIntensityd`) is now invertible. As requested in #5647, the subtrahend and divisor actually used — whether passed in or computed from the image mean/std — are stored in the transform's meta information, so the normalization can be reversed (`img * divisor + subtrahend`). This is useful for reconstruction problems and for logging/recovering original intensities.

Implementation follows the maintainer's suggestion in the issue thread (store the stats via `push_transform(..., extra_info=...)`):

- `NormalizeIntensity` now subclasses `InvertibleTransform`. `_normalize` returns the `sub`/`div` it used; `__call__` collects them (per channel when `channel_wise=True`, once otherwise) and pushes them onto the output MetaTensor's transform stack. A new `inverse()` reverses the operation.
- `NormalizeIntensityd` subclasses `InvertibleTransform` and delegates `inverse()` to the array transform per key.
- Transform tracking is only performed when the output is a `MetaTensor` and `get_track_meta()` is True, so behaviour is unchanged when meta tracking is off.

**Scope note (`nonzero=True`):** inversion is intentionally not supported when `nonzero=True`, because reversing the masked normalization exactly would require storing the per-voxel zero mask. In that case `inverse()` raises a clear `NotImplementedError` rather than returning an incorrect result. Happy to extend this (e.g. by optionally storing the mask) if maintainers prefer.

### Types of changes
- [x] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`. *(ran the relevant unit tests + formatting/lint locally — see below)*
- [x] Quick tests passed locally by running `./runtests.sh --quick --unittests --disttests`. *(ran the affected suites)*

### Testing

```
python -m pytest tests/transforms/test_normalize_intensity.py tests/transforms/test_normalize_intensityd.py
# 106 passed (99 existing + 7 new inverse tests)
```

New tests assert round-trip `inverse(transform(x)) ≈ x` for global and channel-wise modes with both computed and explicitly-provided sub/div, and that `nonzero=True` raises `NotImplementedError` on inverse. `isort`, `black`, and `ruff check` all pass on the changed files.


aymuos15:
What about the `nonzero=True` case? it'll always raise inside `Compose.inverse`/`Invertd` right.

azrabano23:
Yes, you were right. `Compose([NormalizeIntensity(nonzero=True)]).inverse(...)` and `Invertd` both raised (`RuntimeError: applying transform ... NormalizeIntensity.inverse` wrapping the `NotImplementedError`), which made the transform unusable in any invertible pipeline for the most common config. Fixed in 8e53cfb7 (branch merged with `dev` first, which is also why the diff now shows the `get_spatial_ndim` import).

What changed:
- With `nonzero=True` the forward pass leaves zero voxels at zero, so the inverse rebuilds the forward mask as `out != 0`. The one ambiguity is a non-zero voxel whose value equals the subtrahend: it becomes exactly 0 and is indistinguishable from a skipped voxel. `_normalize` now records the flat indices of those voxels in `extra_info["zeroed_idx"]` (normally an empty tensor; it only grows in the degenerate integer-valued case) and `inverse()` adds them back to the mask before applying `out * div + sub`. Inversion is exact, including that case — `test_inverse_nonzero_value_equal_to_mean` uses `[[0,1,2,3],[0,0,0,2]]`, whose non-zero mean is 2 globally and per channel, so two voxels hit 0 and are restored to 2 with `atol=0`.
- The all-zero-input path stores identity stats (`0.0`/`1.0`) instead of `None`, per the CodeRabbit note, so `applied_operations` stays collate-safe.
- `NormalizeIntensityd.inverse` is typed on `torch.Tensor` like the spatial dictionary transforms; that was the `static-checks (mypy)` failure on the previous head.
- Google-style docstrings on the new methods; the tests restore `track_meta` via `addCleanup`.

Tests added: `nonzero`, `channelwise_nonzero`, `nonzero_explicit` in `test_inverse` (image has zero voxels and an all-zero channel), the value-equals-mean case above, `Compose(...).inverse` with `nonzero=True`, and `Invertd` over `NormalizeIntensityd(nonzero=True)`.

```
python -m pytest tests/transforms/test_normalize_intensity.py tests/transforms/test_normalize_intensityd.py   # 114 passed
python -m pytest tests/transforms/inverse/test_invertd.py tests/transforms/test_inverse_collation.py tests/transforms/compose/test_compose.py   # 191 passed, 32 skipped (needs scipy/nibabel installed)
mypy monai/transforms/intensity/array.py monai/transforms/intensity/dictionary.py   # no findings in the changed code
black/isort/ruff on the four touched files: clean
```

One design note in case you prefer otherwise: I chose to store only the collision indices rather than the full boolean mask, since a 512^3 mask would add ~134 MB to every `MetaTensor`'s applied operations. Happy to switch if you would rather have the mask.

Issue #8857, opened 2026-05-18
Discussion: optional RankSEG-style decoding support in MONAI post-processing


statmlben:
Hi MONAI maintainers,

I would like to ask for your feedback on whether [RankSEG](https://github.com/rankseg/rankseg)-style decoding could be a good fit for MONAI post-processing. I ran a small smoke test using RankSEG on top of a pretrained MONAI bundle, and the result looked encouraging, especially for the tumor class. This issue is meant as an initial discussion rather than a request to merge a specific API.

## Preliminary Smoke Test

I used the pretrained MONAI `pancreas_ct_dints_segmentation` bundle on 6 MSD Pancreas cases. This is not intended as a full benchmark, and the sample size is small. The goal was only to check whether RankSEG can be inserted into a standard MONAI bundle workflow and whether the result is worth further discussion.

Setup:

- MONAI: `1.5.2`
- PyTorch: `2.12.0+cu130`
- Bundle: `pancreas_ct_dints_segmentation`
- Task: MSD Pancreas, 3 classes: background, pancreas, tumor
- No retraining
- No checkpoint modification
- Post-processing only

Results:

| Method | Mean Dice | Pancreas Dice | Tumor Dice |
| --- | ---: | ---: | ---: |
| MONAI argmax | 0.6584 | 0.6979 | 0.6188 |
| MONAI largest connected component | 0.6584 | 0.6979 | 0.6188 |
| RankSEG Dice | 0.6923 | 0.7193 | 0.6654 |

In this small test, the largest visible change was on the tumor class, where Dice improved from `0.6188` with argmax to `0.6654` with RankSEG.

Example visualization:

<img width="2340" height="2088" alt="Image" src="https://github.com/user-attachments/assets/5a2c0846-0662-494e-aa2e-db3820487b67" />

The visualization focuses on tumor segmentation. Error maps use white for true positive, red for false positive, and cyan for false negative.

## Broader Evidence From RankSEG Papers

The MONAI experiment above is only a small smoke test for this possible integration point. RankSEG has also been studied more systematically in two papers:

- [RankSEG: A Consistent Ranking-based Framework for Segmentation](https://www.jmlr.org/papers/v24/22-0712.html),
  Dai and Li, JMLR 2023.
- [RankSEG-RMA: An Efficient Segmentation Algorithm via Reciprocal Moment Approximation](https://neurips.cc/virtual/2025/poster/119909),
  Wang and Dai, NeurIPS 2025.

[...]

PR #8908, opened 2026-06-09
Add optional RankSEG decoding to AsDiscrete

ZixunWang:
Related to #8857.

### Description

This PR adds optional RankSEG decoding support to MONAI post-processing.

RankSEG is integrated as an extension of the existing discrete post-processing API:

```python
AsDiscrete(rankseg=True)
AsDiscreted(keys="pred", rankseg=True)
```

`rankseg=False` remains the default, so existing behavior is unchanged.

### Types of changes

### Changes

- Add `rankseg` support to `AsDiscrete` and `AsDiscreted`.
- Import RankSEG through MONAI's optional dependency mechanism.
- Add optional dependency config, documentation, and tests.

Many thanks to @statmlben and @Leev1s for their collaboration and helpful discussions on the RankSEG integration.

### Feedback Requested

This PR is intended to start from a conservative integration point. We would especially appreciate feedback on:

- whether `AsDiscrete(rankseg=True)` / `AsDiscreted(..., rankseg=True)` is the preferred API shape;
- whether combining `argmax=True` and `rankseg=True` should raise a `ValueError` as implemented, or only emit a warning.


aymuos15:
Really cool! To me, the only changes would be adding a 3D test and extending out a metric arg so that people can choose something other than dice?

ZixunWang:
@aymuos15 Thanks for your review and suggestions! I have added a 3D test case:
```python
[[[[0.3, 0.6]]], [[[0.7, 0.4]]]]  # shape: [C, H, W, D] = [2, 1, 1, 2]
```
RankSEG internally handles inputs with shape `[B, C, *spatial]`, where `spatial` has no restriction on the number of dimensions.

The `metric` argument is also exposed via `AsDiscrete(rankseg=True, metric=...)` / `AsDiscreted(..., rankseg=True, metric=...)`

aymuos15:
Thanks! Yeah, just wanted to make sure the internal handle is obvious here and the metric is an actual choice. LGTM!

ZixunWang:
Hi @ericspod @Nic-Ma @KumoLiu, gentle ping on this PR.

@aymuos15's feedback has been addressed in `f08231b`: we added the 3D test case and exposed the `metric` argument for RankSEG decoding. @aymuos15 has since said LGTM.

The checks are green and the PR appears to be mergeable cleanly. Would one of you be able to take a code-owner review, or let us know if there are any remaining API concerns before merging?

Happy to make follow-up changes if needed.

ZixunWang:
Hi @ericspod @Nic-Ma @KumoLiu, gentle follow-up on this PR.

We have synced the branch with the latest MONAI dev branch and addressed the previous feedback.

The intended usage follows the existing `AsDiscrete` / `AsDiscreted` API shape. After model logits are converted to probabilities, RankSEG can be enabled as an alternative discrete post-processing step:

```python
probs = Activations(softmax=True, dim=1)(logits)

post_argmax = AsDiscrete(argmax=True)
post_rankseg = AsDiscrete(rankseg=True, metric="dice")

pred_argmax = post_argmax(probs)
pred_rankseg = post_rankseg(probs)
```

If the AsDiscrete(rankseg=True) / AsDiscreted(..., rankseg=True) API shape looks acceptable, would one of you be able to take a code-owner review? If you prefer a different API direction, happy to adjust.

monai/transforms/post/array.py
```diff
@@ -155,9 +160,17 @@ class AsDiscrete(Transform):
             Defaults to ``None``.
         rounding: if not None, round the data according to the specified option,
             available options: ["torchrounding"].
+        rankseg: whether to apply RankSEG decoding. Requires installing the optional ``rankseg`` package.
```
ericspod: Probably this is going to be the first place users will encounter any reference to RankSEG, there should be some description of what it does and what it's for. It would help to reduce the length of the existing text as well.

monai/transforms/post/array.py
```diff
@@ -183,9 +202,13 @@ def __init__(
         to_onehot: int | None = None,
         threshold: float | None = None,
         rounding: str | None = None,
+        rankseg: bool = False,
         **kwargs,
     ) -> None:
+        if argmax and rankseg:
+            raise ValueError("`rankseg=True` is incompatible with `argmax=True`.")
         self.argmax = argmax
+        self.rankseg = rankseg
```
ericspod, line 211: If rankseg is True but the package isn't present, it would help to raise the `OptionalImportError` here as well.

monai/transforms/post/dictionary.py
```diff
@@ -182,14 +183,21 @@ def __init__(
             rounding: if not None, round the data according to the specified option,
                 available options: ["torchrounding"]. it also can be a sequence of str or None,
                 each element corresponds to a key in ``keys``.
+            rankseg: whether to apply RankSEG decoding. Requires installing the optional ``rankseg`` package.
```
ericspod: Similarly this is likely where new users will encounter this so a comment to see the description in `AsDiscrete` would help.

ericspod (commented):
Thanks @ZixunWang, I've looked over the PR with only some minor comments. I'm hesitant to approve this now as we want to limit the number of externald dependencies if we can, there are too many now as it is for a variety of reasons. I would like to have more discussion from other maintainers first. We also need to resolve merge conflicts.

monai/transforms/post/array.py
statmlben: Thanks for the suggestion. I shortened the `rankseg` documentation and added a compact description clarifying that RankSEG is an inference-time decoder that converts class probability maps into segmentation label maps while maximizing the expected samplewise Dice or IoU.

monai/transforms/post/array.py
statmlben: Thanks. I added the dependency check to `AsDiscrete.__init__()`, so `AsDiscrete(rankseg=True)` now raises `OptionalImportError` immediately when the `rankseg` package is unavailable. I retained the check in `__call__()` for cases where RankSEG is enabled per call, and added test coverage for both paths.

monai/transforms/post/dictionary.py
statmlben: Thanks. I shortened the `AsDiscreted` parameter documentation and added a cross-reference to `AsDiscrete` for the full RankSEG description, while retaining the dictionary-specific sequence behavior.

statmlben:
Thank you @ericspod for the careful review! The inline comments and merge conflict have now been addressed, and the checks are green.

I understand the concern about adding another external dependency. To provide more concrete evidence for the value of this optional integration, we performed a reproducible comparison between the existing argmax post-processing and RankSEG using the same pretrained model probability outputs.

### Experimental setup

- **Dataset:** Medical Segmentation Decathlon Task07 Pancreas. We reproduced the MONAI Bundle's `prepare_datalist.py` split (`seed=123`, 196 training cases from 281 sorted labeled cases), and fixed the first 20 validation IDs in numerical order before evaluation.
- **Model:** unmodified MONAI `pancreas_ct_dints_segmentation` Bundle v0.5.2, `models/model.pt` (SHA-256: `a18ae8b837f6affe778d7e9f130e6045c04a6f7d5b5dd8470155b9a18b6bcb65`).
- **Number of evaluated cases:** 20 3D volumes.
- **Input to all methods:** identical cached float32 softmax probability maps.
- **Compared methods:**
  - `AsDiscrete(argmax=True)`
  - `AsDiscrete(rankseg=True, metric="dice")`
  - `AsDiscrete(rankseg=True, metric="iou")`
- No retraining, fine-tuning, threshold tuning, or method-specific preprocessing was used.
- Dice and IoU were computed per volume, then macro-averaged over the foreground classes (pancreas and tumor). Background was excluded. If both masks for a class were empty, the class score was defined as 1.
- **Hardware:** NVIDIA GeForce RTX 3090 (24 GB); PyTorch 2.12.0+cu130; RankSEG 0.0.5; MONAI PR head `fcee761a3280cd740753331a913a65fa0d90ba89`.

### Quantitative results

| Method | Mean samplewise Dice | Mean samplewise IoU | Target-metric win rate vs. argmax | Median post-processing time |
| --- | ---: | ---: | ---: | ---: |
| Argmax | 0.6631 ± 0.1836 | 0.5409 ± 0.1762 | — | 1.65 ms |
| RankSEG (Dice) | **0.6840 ± 0.1840** | **0.5653 ± 0.1778** | 90.0% (18/20 Dice) | 145.73 ms |
| RankSEG (IoU) | 0.6838 ± 0.1840 | 0.5651 ± 0.1777 | 95.0% (19/20 IoU) | 142.62 ms |

Compared with argmax:

- RankSEG (Dice) changed mean samplewise Dice by **+0.0209** (paired 95% bootstrap CI **[+0.0124, +0.0307]**) and mean samplewise IoU by **+0.0243** (**[+0.0155, +0.0352]**).
- RankSEG (IoU) changed mean samplewise Dice by **+0.0207** (**[+0.0124, +0.0303]**) and mean samplewise IoU by **+0.0242** (**[+0.0154, +0.0348]**).
- Per-class mean Dice for argmax -> RankSEG (Dice) was **0.7531 -> 0.7721** for pancreas and **0.5731 -> 0.5960** for tumor.
- RankSEG was not uniformly better, but the few target-metric losses were small in absolute magnitude. For RankSEG (Dice), the two Dice losses were `pancreas_092` (**0.3758 ->  0.3748**, Δ **-0.0011**) and `pancreas_117` (**0.4752 -> 0.4749**, Δ **-0.0003**). For RankSEG (IoU), the only IoU loss was `pancreas_092` (**0.3011 -> 0.2980**, Δ **-0.0030**).

### Qualitative examples

The figure below shows the input, ground truth, argmax prediction, RankSEG (Dice) prediction, and RankSEG error overlay. Pancreas is cyan and tumor is orange; in the error overlay, white is true positive, red is false positive, cyan is false negative, and yellow denotes a foreground class mismatch.

The cases were selected using the 25th, 50th, and 75th percentiles of the paired Dice improvement rather than selecting only the largest improvements. For each 3D volume, the displayed slice maximizes ground-truth foreground area, and the crop is based only on the ground-truth foreground bounds; neither rule depends on either decoder's result.

MSD Pancreas argmax and RankSEG qualitative comparison


The evaluation code, fixed configuration, per-case results, summary, and raw timing samples are available at:

[**monai-pr-8908-reproducibility**](https://github.com/rankseg/rankseg/tree/main/experiments/monai-pr-8908-reproducibility)

The intended MONAI usage remains a small change to the standard decollated post-processing workflow:

```python
post_pred = Compose(
    [
        Activations(softmax=True),
        AsDiscrete(rankseg=True, metric="dice"),
    ]
)

predictions = [post_pred(x) for x in decollate_batch(logits)]
```

RankSEG remains fully optional: `rankseg=False` is the default, it is loaded through MONAI's optional import mechanism, and existing behavior is unchanged unless explicitly enabled.

For additional context, I maintain RankSEG and supervised (my students) on this work. RankSEG is a **long-term project that we intend to actively maintain**, and I would be happy to help address future compatibility or MONAI integration issues.

I appreciate the maintainers taking the time to consider the dependency trade-off, and I am happy to adjust the packaging or integration boundary if a different approach would be preferable.

garciadias:
Thanks for all the work on this, for iterating through the feedback so thoroughly, and congratulations on creating this algorithm — the integration itself is clean and well-tested, and it clearly follows the existing `AsDiscrete/argmax`.
  
Picking up on @ericspod's point about the extras list already being large: I looked into rankseg itself a bit more (it's still pre-1.0, ~8 months since its first release, and its commit history is concentrated in the two of you) — not a knock on the project at all, just a consideration on the long term maintainance of the repo that worth weighing alongside the dependency-count concern. My concern is that if your repo gets stale with an old version of scipy or other dependencies, it could silently break and accumulate technical debt in MONAI. Maybe one of these two options could get `RankSEG` in front of MONAI users without adding another entry to the optional-extras list, in case either is of interest:
  
Option A — vendor the core algorithm into MONAI directly. This would remove the dependency question entirely (no new PyPI package in setup.cfg), and it's actually a well-precedented pattern in this codebase already, monai/networks/nets/hovernet.py, mednext.py/mednext_block.py, monai/losses/cldice.py, and the monai/apps/detection/ subsystem all vendor external algorithm code wholesale, each isolated in its own file with a header attribution block (source URL + license) rather than a dependency entry. rankseg's BSD-3 license is compatible with that pattern.
  
Concretely, this would probably look like a new monai/transforms/utils_rankseg.py (sibling to the existing utils_morphological_ops.py, which plays the same "tensor-algorithm helper behind a transform" role), carrying functional.py + _rankseg_algo.py + the RefinedNormalPB/RefinedNormal distribution classes (~650-700 lines total — the SAM/Transformers integration helpers wouldn't be needed here since MONAI would only call the core rankseg() function). The scipy use in the distribution code would reuse MONAI's existing optional_import("scipy...") pattern (already used the same way in apps/pathology, apps/deepgrow), so no new dependency-handling mechanism is needed either.

I recognise this is a meaningfully bigger ask than the current PR, but given you already know this implementation best, I think this would be a feasible task for you. After your implementation is sufficiently tested, we could maintain the code and update it as needed.
  
Option B — a tutorial in Project-MONAI/tutorials (https://github.com/Project-MONAI/tutorials). Lower lift, and not mutually exclusive with Option A: a short notebook showing how to wrap rankseg.functional.rankseg in a small custom Transform/MapTransform and drop it into a Compose pipeline alongside standard MONAI post-processing. This keeps MONAI core dependency-free while still giving users a documented, discoverable on-ramp to `RankSEG`, and could ship quickly as a near-term win regardless of which way the core-integration question goes.

I am happy to help think through either path further. @ericspod, does this framing match what you had in mind re: the dependency-count concern?

ericspod:
Thanks @ZixunWang and @statmlben for the work here. I think @garciadias has gotten to the root of the issue here, we've had issues with other packages before causing installation problems and/or breaking CI for MONAI itself. My feeling is that we not consider merging this PR for the time being but possibly consider it again in the future. If you want to consider Option A (and it's feasible) we can re-review once those changes are made. I would suggest doing option B either way, we can put things into a tutorial with specialised transform classes (or subclass AsDiscreted) that illustrate how to use the proposed method. Users can take that and use your library easily without needing to integrate into MONAI directly, and you'll get the visibility for your methodology as well. I hope our reasoning makes sense for you, if you want to update this PR or create the tutorial we can definitely help and review again.

statmlben:
Thanks @garciadias and @ericspod for the thoughtful feedback and for explaining the maintenance concerns. The reasoning makes sense to us.

We will proceed with Option B first and should be able to prepare a tutorial demonstrating how to use RankSEG in a MONAI pipeline fairly quickly.

We are also very interested in Option A. We are now continuing to improve the quality and maturity of the RankSEG algorithm and codebase. As a next step, we will put together an initial prototype and draft structure, and then seek your guidance on the preferred API and expected scope before proceeding with the full vendoring effort. We would be happy to provide long-term maintenance support if the core algorithm is eventually integrated into MONAI. 

We hope that making RankSEG accessible to MONAI users, will also help us gather broader real-world feedback and further improve both the method and its implementation. Thanks again for offering a constructive path forward. We will follow up with the tutorial PR soon:)

statmlben:
Hi @garciadias and @ericspod,

Following up on Option B, we have opened a tutorial PR:

https://github.com/Project-MONAI/tutorials/pull/2068

The tutorial demonstrates how to use `RankSEG` and `RankSEGd` as optional third-party post-processing transforms in a MONAI workflow. It uses a pretrained MONAI pancreas segmentation Bundle and a real MSD Task07 Pancreas case to:

- compare RankSEG with conventional argmax decoding;
- demonstrate the array-based `RankSEG` transform;
- demonstrate `RankSEGd` in a dictionary-based MONAI `Compose` pipeline;
- report per-class Dice scores and visualize the resulting segmentations.

The notebook is self-contained and has passed the local MONAI tutorial runner. We would appreciate any feedback on the tutorial structure and integration approach.

To be clear, this tutorial PR is our Option B deliverable. We also remain committed to **Option A** and will continue following it up here in #8908. We plan to prepare an initial prototype and proposed API/design scope, then bring them back here for discussion and guidance before proceeding with the full vendoring effort.

Thanks again for suggesting this constructive path!

ericspod:
Hi @statmlben I have reviewed the tutorial now so please look there for comments. We're happy with you continuing with option A, if you're still working on the algorithm then I'd suggest doing what improvements you can first in your own repo then we look into MONAI integration here. Continued support from your side would be definitely appreciated as well, thanks.

garciadias:
Hi all, now that we have approved [tutorials:2068](https://github.com/Project-MONAI/tutorials/pull/2068), can we close this one?

statmlben:
Thanks everyone for the reviews and discussion. I’m okay with closing this PR for now, given that the approved tutorial provides a working optional integration. We may revisit a core integration in the future if there is broader user demand or a more suitable general abstraction. Thanks again!

> Hi all, now that we have approved [tutorials:2068](https://github.com/Project-MONAI/tutorials/pull/2068), can we close this one?

ericspod:
Hi @statmlben thanks for the tutorial, that's now merged. I'm going to close this PR but we can reopen at any time. Thanks for the contribution!

Issue #8909, opened 2026-06-09
Performance: `get_surface_distance` should use a KDTree for euclidian distance on CPU


jberg5:
https://github.com/Project-MONAI/MONAI/blob/eccefc57550b111ed781d82249dfe77872a0e918/monai/metrics/utils.py#L294

Currently it does a full euclidian distance transform which computes all the distances between all of the points. On a GPU this is an embarrassingly parallel computation, but on a CPU we can get significant speedups from using a [KDTree](https://docs.scipy.org/doc//scipy-1.9.3/reference/generated/scipy.spatial.KDTree.html) to just compute the distances to the nearest edge voxel.

I'll raise a PR - just creating this issue for tracking purposes.

jberg5:
Fixed by https://github.com/Project-MONAI/MONAI/pull/8910

PR #8910, opened 2026-06-10
Perf: Use a KDTree on CPU instead of full EDT for 1.5x to 16x faster metric computation

jberg5:
Fixes # 8909.
https://github.com/Project-MONAI/MONAI/issues/8909

### Description

Pretty straightforward: we don't need to compute the distance between every conceivable voxel, just the edges, so we can use a KDTree on CPU and compute `HausdorffDistanceMetric` and `SurfaceDistanceMetric` etc significantly faster.

GPU implementations of KDTrees exist, but I did a little benchmarking and found that they are slower than the full EDT since the distance computations are embarrassingly parallel and well-suited to the hardware (gpu goes brrr), so I left that path unchanged.

Measured speedups (on my M3 mac, and Intel Cascade Lake) range from 1.5x for small inputs to 16x for larger volumes and noisier data.

I think existing test coverage is good enough that we don't need more here - all pass for me, and I've spot checked a few problems to ensure identical output metrics. Edit: have added some more tests per the coderabbit's suggestions.

### Types of changes


monai/metrics/utils.py
```diff
@@ -291,6 +293,26 @@ def get_surface_distance(
...
+                gt_coords = np.argwhere(convert_to_numpy(seg_gt)).astype(np.float64)
+                pred_coords = np.argwhere(convert_to_numpy(seg_pred)).astype(np.float64)
+                if spacing is not None:
+                    scale = np.asarray(spacing, dtype=np.float64)
+                    gt_coords *= scale
+                    pred_coords *= scale
+                # leafsize larger than the default (16) is faster here: we build the tree
+                # for a single batched query rather than amortizing it over many queries.
+                surface_distance = KDTree(gt_coords, leafsize=32).query(pred_coords, k=1)[0]
+                return convert_to_dst_type(surface_distance, seg_pred, dtype=lib.float32)[0]
```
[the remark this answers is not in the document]
jberg5: Sure, added some tests.

jberg5:
Can any humans confirm that tests need docstrings? I don't see this as a pattern generally. Happy to add them if that's the guidance but feels a little odd.

monai/metrics/utils.py
```diff
@@ -291,6 +293,26 @@ def get_surface_distance(
...
+            # The euclidean surface distance only needs the distance from each `seg_pred`
+            # edge voxel to the nearest `seg_gt` edge voxel. CPU and GPU favour different
+            # algorithms for this:
+            #   * On CPU, a KD-tree over the (sparse) edge-voxel coordinates avoids the dense
+            #     full-volume distance transform, and handles outlier points that expand the
+            #     bounding box.
+            #   * On GPU, the dense EDT is embarrassingly parallel and significantly faster than
+            #     cupy's KDTree (as of this writing anyway)
+            on_gpu = isinstance(seg_gt, torch.Tensor) and seg_gt.device.type == "cuda"
+            if not on_gpu:
```
ericspod: You will have to check here that the scipy module is present, so your optional import will need the second return value to do that. If the module isn't present the code should default to the existing behaviour.

tests/metrics/test_surface_distance.py
```diff
@@ -16,8 +16,10 @@
 import numpy as np
 import torch
 from parameterized import parameterized
+from scipy.ndimage import distance_transform_edt
```
ericspod: This will have to be an optional import as well.

tests/metrics/test_surface_distance.py
```diff
@@ -182,5 +184,43 @@ def test_nans(self, input_data):
...
+    gt = create_spherical_seg_3d(radius=20, centre=(30, 30, 30))
+    pred = create_spherical_seg_3d(radius=20, centre=(32, 31, 30))
+    rng = np.random.RandomState(seed)
+    for _ in range(5):
+        pred[tuple(rng.randint(0, s) for s in pred.shape)] = 1
+    edges_pred, edges_gt = get_mask_edges(pred, gt)
+    return np.asarray(edges_pred, dtype=bool), np.asarray(edges_gt, dtype=bool)
+
+
+class TestSurfaceDistanceKDTreeMatchesEDT(unittest.TestCase):
```
ericspod, line 205: The tests in this class will have to have `skipUnless` guards to skip the tests when scipy isn't present. Have a look at other tests using scipy or other libraries to see what the pattern is.

ericspod (commented):
Hi @jberg5 thanks for this optimisation. I had minor comments about the need to make scipy optional but otherwise it looks good. You might want to also add a flag to the function `use_scipy = True` that will skip the conditional block you've added if set to False, in case anyone has reason not to use scipy.

jberg5:
Thanks for the review @ericspod ! The one thing I didn't do is add a `use_scipy` bool to `get_surface_distance` since it already hard-depends on scipy.ndimage via `distance_transform_cdt` and `distance_transform_edt`, and indeed I believe we would even throw a runtime error if ndimage isn't available:

https://github.com/Project-MONAI/MONAI/blob/4ba89bd818a44395842f11f3316b5aaad48353a6/monai/transforms/utils.py#L2542

But happy to treat KDTree as an optional import and leave the fallback in place.

ericspod (approved):
Looks good to me now, thanks.

PR #8912, opened 2026-06-11
fix: update Dockerfile and requirements-dev.txt for MONAI 1.6 tutorial compatibility

garciadias:
## Summary

- Rebase Docker image from `nvcr.io/nvidia/pytorch:24.10-py3` to `25.03-py3` to support RTX 5090 (Blackwell, SM_120/CUDA 12.8); remove the now-obsolete `torch.patch` ONNX revert which was specific to 24.10
- Pin `mlflow<3.0` — mlflow 3.x is broken on Python 3.12 due to a relative import in `mlflow.utils.uv_utils` (`from .. import zipp` fails at top-level scope); this caused 4 tutorial notebooks to fail in our CI run
- Pin `transformers<5.0` — transformers 5.x references `torch.float8_e8m0fnu` which does not exist in the nv25.03 build of PyTorch 2.7; this caused the HuggingFace tutorial to fail
- Add `aim` and `lightning>=2.0` as declared dependencies in `requirements-dev.txt` (were previously undeclared but required by tutorial notebooks)
- Rebuild the NVIDIA pip constraint file to retain `numpy==1.26.4` (nv25.03 PyTorch compiled against NumPy 1.x) and add `setuptools<71` (newer setuptools dropped `pkg_resources` needed by legacy `setup.py` in git-sourced packages like MetricsReloaded and segment-anything)
- Remove `python_version <= '3.10'` caps from `cucim`, `onnxruntime`, and `transformers` — these restrictions were keeping packages out of the Python 3.12 image unnecessarily
- Install `papermill`, `jupytext`, `autopep8`, `autoflake`, and `ipywidgets` directly in the Dockerfile so the tutorial runner is self-contained

## Context

These changes were identified by running the full MONAI tutorial test suite in a fresh Docker build against a MONAI 1.6 dev branch and comparing results with a native conda reference run (Eric's run, `eccefc57`). The rerun with stderr captured (`runner_output_our_only.logs`) confirmed the specific error for each notebook group.

## Type of change
- [x] Bug fix
- [ ] Feature
- [ ] Hotfix
- [ ] Spike / exploration
- [ ] Documentation
- [ ] Refactor

## Test plan
- [ ] Rebuild Docker image with `docker build -t monai_1_6:latest .`
- [ ] Re-run `bash run_our_only.sh 2>&1 | tee runner_output_v2.logs` inside the container
- [ ] Verify mlflow notebooks pass (R1: 4 notebooks)
- [ ] Verify `hugging_face/hugging_face_pipeline_for_monai.ipynb` passes (R3)
- [ ] Verify `experiment_management/spleen_segmentation_aim.ipynb` passes (R6)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,12 +26,21 @@ RUN if [[ $(uname -m) =~ "aarch64" ]]; then \
 
 WORKDIR /opt/monai
 
+# Patch NVIDIA's pip constraint file:
+#   - keep numpy==1.26.4 pin (PyTorch nv25.12 was compiled against NumPy 1.x)
+#   - add setuptools<71 (newer setuptools removed pkg_resources needed by legacy setup.py)
+#   - remove jupytext/isort pins so the tutorial runner can install its required versions
+RUN grep '^numpy==' /etc/pip/constraint.txt > /tmp/new_constraints.txt \
+  && printf 'setuptools<71\n' >> /tmp/new_constraints.txt \
+  && cp /tmp/new_constraints.txt /etc/pip/constraint.txt
+
 # install full deps
 COPY requirements.txt requirements-min.txt requirements-dev.txt /tmp/
 RUN cp /tmp/requirements.txt /tmp/req.bak \
   && awk '!/torch/' /tmp/requirements.txt > /tmp/tmp && mv /tmp/tmp /tmp/requirements.txt \
   && python -m pip install --upgrade --no-cache-dir --no-build-isolation pip wheel wheel-stub \
-  && python -m pip install --no-cache-dir --no-build-isolation -r /tmp/requirements-dev.txt
+  && python -m pip install --no-cache-dir --no-build-isolation -r /tmp/requirements-dev.txt \
+  && python -m pip install --no-cache-dir --no-build-isolation papermill jupytext autopep8 autoflake ipywidgets
 
 # compile ext and remove temp files
 # TODO: remark for issue [revise the dockerfile #1276](https://github.com/Project-MONAI/MONAI/issues/1276)
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,11 +27,15 @@ RUN if [[ $(uname -m) =~ "aarch64" ]]; then \
 WORKDIR /opt/monai
 
 # Patch NVIDIA's pip constraint file:
-#   - keep numpy==1.26.4 pin (PyTorch nv25.12 was compiled against NumPy 1.x)
+#   - keep the base image's numpy pin if it has one (nv25.03 pins numpy==1.26.4 as its
+#     torch was compiled against NumPy 1.x; nv25.12 ships an empty constraint.txt)
 #   - add setuptools<71 (newer setuptools removed pkg_resources needed by legacy setup.py)
+#   - pin urllib3>=2 so that notebook cells doing !pip install legacy packages (e.g.
+#     bentoml==0.13.1) cannot downgrade urllib3 to 1.x, which would break requests,
+#     huggingface_hub, gdown, transformers, mlflow, etc. for every subsequent notebook
 #   - remove jupytext/isort pins so the tutorial runner can install its required versions
-RUN grep '^numpy==' /etc/pip/constraint.txt > /tmp/new_constraints.txt \
-  && printf 'setuptools<71\n' >> /tmp/new_constraints.txt \
+RUN (grep '^numpy' /etc/pip/constraint.txt || true) > /tmp/new_constraints.txt \
+  && printf 'setuptools<71\nurllib3>=2\n' >> /tmp/new_constraints.txt \
   && cp /tmp/new_constraints.txt /etc/pip/constraint.txt
 
 # install full deps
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -26,12 +26,25 @@ RUN if [[ $(uname -m) =~ "aarch64" ]]; then \
...
+  && cp /tmp/new_constraints.txt /etc/pip/constraint.txt
+
 # install full deps
 COPY requirements.txt requirements-min.txt requirements-dev.txt /tmp/
 RUN cp /tmp/requirements.txt /tmp/req.bak \
   && awk '!/torch/' /tmp/requirements.txt > /tmp/tmp && mv /tmp/tmp /tmp/requirements.txt \
   && python -m pip install --upgrade --no-cache-dir --no-build-isolation pip wheel wheel-stub \
-  && python -m pip install --no-cache-dir --no-build-isolation -r /tmp/requirements-dev.txt
+  && python -m pip install --no-cache-dir --no-build-isolation -r /tmp/requirements-dev.txt \
+  && python -m pip install --no-cache-dir --no-build-isolation papermill jupytext autopep8 autoflake ipywidgets
```
ericspod: We haven't really expected the Dockerfile to have everything for tutorials. Many of the notebooks mention extra requirements that are added through pip calls in cells, so that would suffice for things in `requirements-dev.txt`. I wouldn't add the requirements here for running Jupyter, users who want to use this image for tutorials should install them separately themselves. We can add information in the tutorials repo on how to do that.

requirements-dev.txt
```diff
@@ -33,9 +34,11 @@ tifffile; platform_system == "Linux" or platform_system == "Darwin"
 pandas
 requests
 einops
-transformers>=4.53.0
-mlflow>=2.12.2,<3.13
+transformers>=4.53.0, <5.0  # 5.x references torch.float8_e8m0fnu absent in older PyTorch builds
+mlflow>=2.12.2, <3.0  # 3.x broken on Python 3.12 (relative import in mlflow.utils.uv_utils)
 clearml>=1.10.0rc0
+aim; platform_system == "Linux"  # experiment tracking backend for spleen_segmentation_aim tutorial
+lightning>=2.0  # PyTorch Lightning for lightning-based training tutorials
```
ericspod: Related to my previous comment, I wouldn't add requirements here for specific notebooks, those should be installed within the notebook or the tutorial documentation should mention they're needed. We have too many requirements in this file as it is!

Dockerfile
```diff
@@ -26,12 +26,25 @@ RUN if [[ $(uname -m) =~ "aarch64" ]]; then \
...
 
+# Patch NVIDIA's pip constraint file:
+#   - keep the base image's numpy pin if it has one (nv25.03 pins numpy==1.26.4 as its
+#     torch was compiled against NumPy 1.x; nv25.12 ships an empty constraint.txt)
+#   - add setuptools<71 (newer setuptools removed pkg_resources needed by legacy setup.py)
+#   - pin urllib3>=2 so that notebook cells doing !pip install legacy packages (e.g.
+#     bentoml==0.13.1) cannot downgrade urllib3 to 1.x, which would break requests,
+#     huggingface_hub, gdown, transformers, mlflow, etc. for every subsequent notebook
+#   - remove jupytext/isort pins so the tutorial runner can install its required versions
+RUN (grep '^numpy' /etc/pip/constraint.txt || true) > /tmp/new_constraints.txt \
```
ericspod, line 34: Thanks for looking into the image. Is this RUN needed for the image in general or specifically for the tutorials? If the latter we shouldn't have it perhaps.

Dockerfile
garciadias: You're right. I've removed the  line from the Dockerfile in the latest commit. Those are tutorial runner tools, not MONAI image dependencies — the notebooks already handle their own extra deps via `pip install` cells. I'll make sure any notebooks that need additional setup have the appropriate install cells in the tutorials repository.

requirements-dev.txt
garciadias: Agreed — removed both `aim` and `lightning>=2.0` in the latest commit. The notebooks that need them already self-install via `pip install` cells, so there is no need to carry them in the MONAI dev requirements.

Dockerfile
garciadias: The constraint patching is general-purpose rather than tutorial-specific:

- `setuptools<71` — `MetricsReloaded` (in `requirements-dev.txt`) uses a legacy `setup.py` that imports `pkg_resources`, which was removed from `setuptools>=71`. Without this pin the `requirements-dev.txt` install step fails on any fresh image build.
- `urllib3>=2` — prevents `pip install` of legacy packages (e.g. `bentoml==0.13.1`) from downgrading `urllib3` to 1.x, which would silently break `requests`, `gdown`, `huggingface_hub` and others already installed by the base image.
- `numpy` — just preserves whatever pin the base image ships with (nv25.12 ships an empty constraint file, so this is a no-op there; older base images had `numpy==1.26.4` to match their PyTorch build).

I have updated the comment to drop the tutorial-specific bullet and clarify the general-purpose intent.

ericspod: Sounds good, we should look into [the MetricsReloaded issue](https://github.com/Project-MONAI/MetricsReloaded/issues/59) in the short term.

ericspod (approved):
Looks good to me now. We need to do a Dockerhub release at some point soon, that was the next thing to sort for me. We can look at the slim image later if it needs any updates along these lines, but we can merge this now.

PR #8915, opened 2026-06-12
Raise exception for uninstalled LoadImage reader and print writer pkg…

VedhSontha:
… suggestion

Fixes # .

### Description

### Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ If you have used MONAI in your research, please cite us! The citation can be exp
 
 ## Model Zoo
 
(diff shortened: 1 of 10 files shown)


ericspod:
Hi @VedhSontha thanks for the contribution. There seems to be a number of unrelated changes in this PR that would be better separated out. We are also tracking the image writer install hint fix with #7980, specifically #8989 is looking like the best solution to me. I would rather continue with its approach then we can return to others like this to adapt.

PR #8918, opened 2026-06-15
Use compact [1,-2,1] kernel for BendingEnergyLoss second derivatives

### LINKED ISSUE #5939: BendingEnergyLoss

I know this function comes from the translation of the function from the DeepReg library, but the computation of double derivatives using central difference two times is very unstable numerically. The finite difference template would be too large.

The correct template kernel to use for the second difference would be [1 -2 1] for double derivative wrt same variable.
and for mixed partials, you can use forward difference approx. for example here
http://www.holoborodko.com/pavel/2014/11/04/computing-mixed-derivatives-by-finite-differences/

vishnukannaujia:
Fixes #5939.

### Summary

`BendingEnergyLoss` previously computed second-order derivatives by applying the central first-order helper `spatial_gradient` twice, which yields a wide `[1, 0, -2, 0, 1] / 4` stencil for pure second derivatives that spans four voxels per axis. That stencil is less accurate than the standard compact one and forced the validation "all spatial dims > 4".

This PR replaces the second-order computation with compact central stencils evaluated directly on `pred`:

- **Pure** ``d^2/dx_i^2``: ``x[i+1] - 2 * x[i] + x[i-1]`` (the standard `[1, -2, 1]` kernel).
- **Mixed** ``d^2/(dx_i dx_j)``: ``(x[i+1,j+1] - x[i+1,j-1] - x[i-1,j+1] + x[i-1,j-1]) / 4`` (compact 4-point central scheme).

Both span three voxels per axis, so the spatial-size validation is relaxed from `> 4` to `> 2`, matching `DiffusionLoss`. The public API (`__init__(normalize, reduction)`, `forward(pred)`) and `normalize` semantics are unchanged. The existing `spatial_gradient` helper is left untouched because `DiffusionLoss` still uses it.

### Why TEST_CASES expected values do not change

For ``f(x) = x^2``, the analytical second derivative is the constant `2`. Both the previous central-of-central stencil ``(f[i+2] - 2*f[i] + f[i-2]) / 4`` and the new compact ``[1, -2, 1]`` stencil ``f[i+1] - 2*f[i] + f[i-1]`` are exact for quadratics, so both return `2` at every interior voxel. Mixed-partial test inputs are constant in at least one of the two axes, so both stencils return `0` for mixed terms on these cases. Squared and reduced by ``mean``, the existing `TEST_CASES` expected values (``0.0``, ``4.0``, ``100.0``) are therefore invariant under this change.

What does change in the tests:

- `test_ill_shape` is updated to trigger on shape `2` (was `4`) so it still exercises the spatial-size guard.
- A new `TEST_CASES` row covers shape `(1, 3, 3, 3, 3)` of ones (previously rejected by the `> 4` guard) → expected `0.0`, validating the relaxed guard.

Reference for the compact mixed-partial scheme: Pavel Holoborodko's finite-difference notes cited in the original issue.


ericspod (approved):
Hi @vishnukannaujia thanks for this optimisation, I think it looks good and the tests pass just fine.

PR #8919, opened 2026-06-15
Force GC in WSIReader tests to suppress ResourceWarning for unclosed files

### LINKED ISSUE #5461: tests.test_wsireader unclosed file

vishnukannaujia:
Fixes #5461.

### Summary

WSI reader tests emit `ResourceWarning: unclosed file <_io.FileIO ...>` / `BufferedReader` for the temp TIFF inputs they exercise (e.g. the CMU-1 generic TIFF). Investigation:

- Every direct `reader.read(...)` call site in ``tests/utils/enums/test_wsireader.py`` already either uses ``with reader.read(...) as obj:`` or explicitly calls ``obj.close()`` on the returned WSI object. Those sites are not the leak.
- The remaining leak comes from ``LoadImage.__call__`` in ``monai/transforms/io/array.py`` (around L256–289), which does:
  ```python
  img = reader.read(filename)
  img_array, meta_data = reader.get_data(img)
  # ... img_array is wrapped into MetaTensor and returned
  # the reader-returned `img` is never closed
  ```
  The two ``test_with_dataloader*`` tests in this module use ``LoadImaged(reader=WSIReader, backend=..., ...)`` and inherit that leak. The temp TIFF handles only get closed when the garbage collector eventually runs, frequently after Python has already emitted the warning.

This PR keeps the fix at test-hygiene scope: add ``gc.collect()`` to ``tearDown`` of the shared ``WSIReaderTests.Tests`` base class. Each of ``TiffFile.__del__`` / ``OpenSlide.__del__`` / ``CuImage.__del__`` closes the underlying file descriptor, so running gc explicitly at the end of every test invokes those finalizers deterministically and eliminates the warning.

### Why not change ``LoadImage`` / ``BaseWSIReader``

A reader-level or ``LoadImage``-level fix is feasible (e.g. closing ``img`` after ``get_data`` returns) but is broader in scope, affects all readers, and would need to land alongside changes to the public contract of ``BaseWSIReader.read`` (currently documented to return an open WSI object). Happy to follow up with that if a maintainer prefers; this PR was scoped narrowly to the test symptom that the issue raises.

### Verify

```bash
python -m pytest tests/utils/enums/test_wsireader.py -W error::ResourceWarning -v
```
Expected: tests pass and no ``ResourceWarning`` escalations.


ericspod (approved):
Hi @vishnukannaujia thanks for this change, it looks good to me.

PR #8921, opened 2026-06-18
fix: return mu (not mu+std) in VarAutoEncoder reparameterize at eval time (Fixes #8413)

rtmalikian:
Fixes #8413

## Problem

In `VarAutoEncoder.reparameterize()` and `VarFullyConnectedNet.reparameterize()`, the method always returns `std + mu`:

```python
std = torch.exp(0.5 * logvar)
if self.training:
    std = torch.randn_like(std).mul(std)
return std.add_(mu)  # BUG: at eval time, this returns mu + σ instead of mu
```

At **training** time, `std` is correctly replaced with `ε · σ` (random noise scaled by standard deviation). But at **eval** time, the raw standard deviation `σ = exp(0.5 · logvar)` is still added to the mean, producing `μ + σ` instead of just `μ`.

The reparameterization trick (Kingma & Welling, 2014) specifies:
- **Training:** `z = μ + ε · σ` where `ε ~ N(0,1)`
- **Inference:** `z = μ` (no stochastic component)

Adding the standard deviation to the mean at inference time introduces systematic bias — the latent code is shifted away from the learned mean by an amount proportional to the variance.

## Solution

Restructured the `reparameterize` method to return `mu` directly when not in training mode, avoiding the unnecessary `exp(0.5 * logvar)` computation:

```python
def reparameterize(self, mu, logvar):
    if self.training:
        std = torch.exp(0.5 * logvar)
        return mu + torch.randn_like(std) * std
    return mu
```

Applied to both:
- `monai/networks/nets/varautoencoder.py` — `VarAutoEncoder.reparameterize`
- `monai/networks/nets/fullyconnectednet.py` — `VarFullyConnectedNet.reparameterize`

Note: `spade_network.py` has a different implementation that always samples noise (intentional for GAN-style use), so it was left unchanged.

## Verification

**Manual verification:**
```
Eval mode: mu=tensor([[1., 2.]]), z=tensor([[1., 2.]])
PASS: In eval mode, reparameterize returns mu

Train mode: z1=tensor([[-0.6354,  2.5485]]), z2=tensor([[0.2493, 2.8168]])
PASS: In train mode, reparameterize returns mu + noise * std

Old buggy result would have been: tensor([[2.2840, 2.7788]])
```

**Existing tests pass (7/7 + 3/3):**
```
tests/networks/nets/test_varautoencoder.py::TestVarAutoEncoder::test_shape_0 PASSED
[...]

PR #8906, opened 2026-06-08
docs: add Windows BUILD_MONAI install instructions

Shizoqua:
Fixes #6119 .

### Description

`docs/source/installation.md` only documents the POSIX inline form `BUILD_MONAI=1 pip install ...` for building the MONAI C++/CUDA extensions. This `VAR=value command` syntax is not supported by Windows `cmd.exe` or PowerShell, so the documented commands fail out of the box on Windows.

This adds short cmd.exe and PowerShell snippets to both install flows (Option 1 system-wide and Option 2 editable): set the environment variable first, then run the existing `pip install` command. The `set BUILD_MONAI=1` form is the one confirmed working on Windows 11 in the issue thread.

### Types of changes
- [x] Non-breaking change (documentation only).
- [x] Documentation updated.

### How tested
- Verified the new `bat`/`powershell` fenced blocks use valid Pygments lexers and lex without error tokens, so the strict docs build (`build_docs.yml`) emits no new warnings.
- pre-commit markdown hooks (end-of-file, trailing-whitespace, mixed-line-ending) pass on the changed file.

PR #8933, opened 2026-06-19
Fix VarAutoEncoder reparameterize returning mu+std at inference (#8413)

### LINKED ISSUE #8413: Error in VarAutoEncoder class
I'm wondering if there isn't a bug in the implementation of the VarAutoEncoder class.
For the reparameterisation trick (lines 144-150) https://github.com/Project-MONAI/MONAI/blob/46a5272196a6c2590ca2589029eed8e4d56ff008/monai/networks/nets/varautoencoder.py#L28-L155

def reparameterize(self, mu: torch.Tensor, logvar: torch.Tensor) -> torch.Tensor:
    std = torch.exp(0.5 * logvar)

    if self.training:  # multiply random noise with std only during training
        std = torch.randn_like(std).mul(std)

    return std.add_(mu)

The code is correct for training. At test/inference time, we don't multiple std by a random epsilon, which is correct. However, the function still returns mu+std, which doesn't make sense to me. It should be returning mu only at test time, it makes no sense to take mu+std. Can someone check if I'm correct?

Shizoqua:
### Description
Fixes #8413. `VarAutoEncoder.reparameterize` computed `std = exp(0.5*logvar)`, added the random noise term only when training, and then returned `std.add_(mu)`. At inference (`eval()`), `std` is still the standard deviation, so the method returned `mu + std` instead of the posterior mean `mu`. The reparameterization trick's stochastic term should apply during training only; at inference the latent code should be `mu`.

The fix returns `mu` directly when not training, and computes `mu + eps * std` out-of-place otherwise (also avoids the in-place `add_`).

`VarFullyConnectedNet.reparameterize` (`monai/networks/nets/fullyconnectednet.py`) contained the identical bug, so it is fixed the same way in this PR.

### Types of changes

### Testing
Added regression tests asserting that in eval mode the returned latent equals `mu` and is deterministic, while training stays stochastic, for both `VarAutoEncoder` and `VarFullyConnectedNet`.

```
python -m unittest tests.networks.nets.test_varautoencoder tests.networks.nets.test_fullyconnectednet
# Ran 12 tests ... OK
```

### DIFF
--- a/monai/networks/nets/fullyconnectednet.py
+++ b/monai/networks/nets/fullyconnectednet.py
@@ -103,6 +103,8 @@ class VarFullyConnectedNet(nn.Module):
         act: activation type and arguments. Defaults to PReLU.
         bias: whether to have a bias term in linear units. Defaults to True.
         adn_ordering: order of operations in :py:class:`monai.networks.blocks.ADN`.
+        use_mean_at_inference: whether to return the posterior mean (rather than ``mu + std``)
+            as the latent code during inference. Defaults to False for backward compatibility.
 
     Examples::
 
@@ -122,11 +124,13 @@ class VarFullyConnectedNet(nn.Module):
         act: tuple | str | None = Act.PRELU,
         bias: bool = True,
         adn_ordering: str | None = None,
+        use_mean_at_inference: bool = False,
     ) -> None:
         super().__init__()
         self.in_channels = in_channels
         self.out_channels = out_channels
         self.latent_size = latent_size
+        self.use_mean_at_inference = use_mean_at_inference
 
         self.encode = nn.Sequential()
         self.decode = nn.Sequential()
@@ -172,11 +176,27 @@ class VarFullyConnectedNet(nn.Module):
         return x
 
     def reparameterize(self, mu: torch.Tensor, logvar: torch.Tensor) -> torch.Tensor:
+        """Sample a latent code using the reparameterization trick.
+
+        At inference (eval mode), if ``use_mean_at_inference`` is enabled, the posterior
+        mean is returned directly. Otherwise, ``mu + std`` is returned, matching the
+        original behaviour. During training, returns ``mu + eps * std`` with
+        ``eps ~ N(0, I)``.
+
+        Args:
+            mu: Posterior mean, shape ``(batch, latent_size)``.
+            logvar: Log-variance of the posterior, same shape as ``mu``.
+
+        Returns:
+            Sampled latent code, same shape as ``mu``.
+        """
+        if not self.training and self.use_mean_at_inference:
+            # At inference the latent code is the posterior mean; the random
+            # term is only added during training (the reparameterization trick).
+            return mu
         std = torch.exp(0.5 * logvar)
-
-        if self.training:  # multiply random noise with std only during training
-            std = torch.randn_like(std).mul(std)
-
+        if self.training:
+            std = torch.randn_like(std).mul_(std)
         return std.add_(mu)
 
     def forward(self, x: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
--- a/monai/networks/nets/varautoencoder.py
+++ b/monai/networks/nets/varautoencoder.py
@@ -51,6 +51,8 @@ class VarAutoEncoder(AutoEncoder):
             According to `Performance Tuning Guide <https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html>`_,
             if a conv layer is directly followed by a batch norm layer, bias should be False.
         use_sigmoid: whether to use the sigmoid function on final output. Defaults to True.
+        use_mean_at_inference: whether to return the posterior mean (rather than ``mu + std``)
+            as the latent code during inference. Defaults to False for backward compatibility.
 
     Examples::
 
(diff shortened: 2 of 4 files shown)


Shizoqua:
@ericspod Since #8906 (Windows docs) was merged, this one's still waiting. Fixes #8413 — VarAutoEncoder returning mu+std at inference during eval. Rebased, CI all green, issue still open. Would appreciate a review when you get a chance.

Shizoqua:
@ericspod — This has been ready for review for 27 days. Could you take a look when you have a chance?

Shizoqua:
Rebased onto current \dev\ and addressed the review feedback — added Google-style docstrings to the \
eparameterize\ methods in \ullyconnectednet.py\ and \arautoencoder.py\, and converted the test-method inline comments to proper docstrings in both test files.

vikashg (dismissed):
PR #8921 (by rtmalikian) is the same fix but without tests. Closing #8921 in favor of this one. Thanks @Shizoqua

ericspod:
Hi @Shizoqua @vikashg I'm not sure this is correct in fact, and is a breaking change that isn't accounted for. 

When doing inference with a trained VAE, a sample can be generated by feeding a random variable through the decoder to produce a sample. Here though the variable is from encoding inputs, and the goal is to reconstruct the input. Supposedly adding the standard deviation as in the original formulation produces less blurry results.

Either way, this PR represents a breaking change that would impact other users. For those that want the original behaviour, this should be left as the default so that backward compatibility is maintained, then both classes should have a new constructor argument added to enable the new behaviour. 

Thoughts on this? I really think this should be added before merging, thanks.

vikashg:
@ericspod Let me have a look again... today thanks

Shizoqua:
@ericspod Thanks for the detailed review — you're right that this is a breaking change. I've updated the PR to preserve backward compatibility:

- **Default behaviour is unchanged**: at inference, both classes still return `mu + std` (as before the PR).
- A new `use_mean_at_inference=True` constructor argument opts into returning the posterior mean at eval, which is the behaviour that issue #8413 asked for.
- Added to both `VarAutoEncoder` and `VarFullyConnectedNet`, with regression tests covering the opt-in behaviour and the preserved default.

Commit a0b471d06. Happy to adjust the parameter name if you'd prefer something else.

vikashg:
Thanks @Shizoqua. I will let @ericspod look at this. I think he is more fluent with this than I am. Thank you

monai/networks/nets/fullyconnectednet.py
```diff
@@ -172,12 +176,27 @@ def decode_forward(self, z: torch.Tensor, use_sigmoid: bool = True) -> torch.Ten
...
+        Returns:
+            Sampled latent code, same shape as ``mu``.
+        """
+        if not self.training and self.use_mean_at_inference:
+            # At inference the latent code is the posterior mean; the random
+            # term is only added during training (the reparameterization trick).
+            return mu
+        std = torch.exp(0.5 * logvar)
+        eps = torch.randn_like(std) if self.training else torch.ones_like(std)
+        return mu + eps * std
```
ericspod:
```suggestion
        std = logvar.mul(0.5).exp_()
        if self.training:
            std = torch.randn_like(std).mul_(std)
        return std.add_(mu)
```
I'd like to make a change back to something like the original to avoid making a large `ones_like` tensor just to do a no-op multiplication, and to use inline operators to avoid further copies.

monai/networks/nets/varautoencoder.py
```diff
@@ -142,12 +146,27 @@ def decode_forward(self, z: torch.Tensor, use_sigmoid: bool = True) -> torch.Ten
...
+            # term is only added during training (the reparameterization trick).
+            return mu
         std = torch.exp(0.5 * logvar)
-
-        if self.training:  # multiply random noise with std only during training
-            std = torch.randn_like(std).mul(std)
-
-        return std.add_(mu)
+        eps = torch.randn_like(std) if self.training else torch.ones_like(std)
+        return mu + eps * std
```
ericspod: Same change here as the above to avoid `ones_like` and use inline ops where possible.

ericspod (changes requested):
Thanks @Shizoqua I had some minor change for the formulation to improve memory efficiency, otherwise it looks fine to me. Please have a look at this and the formatting issue but we should be good to go then.

Shizoqua:
@ericspod Thanks for the review. I have cleaned up the branch:

- Rebased onto the latest dev.
- Removed the stray monai/utils/safeeval.py and tests/utils/test_safe_eval.py files that had accidentally leaked into this branch.
- Kept the memory-efficiency tweak: at inference with the default use_mean_at_inference=False, the code now returns mu + std directly instead of allocating torch.ones_like(std).
- Ran black on the two test files to fix the formatting failure.

CI is re-running now. Should be good for another look.

Shizoqua:
Hi @ericspod, thanks for the feedback. I've updated both `VarAutoEncoder` and `VarFullyConnectedNet` to use the more memory-efficient formulation:
- training: `std = torch.randn_like(std).mul_(std); return std.add_(mu)`
- inference with `use_mean_at_inference=True`: returns `mu` directly
- default inference: still returns `mu + std` for backward compatibility

I also rebased onto the latest `upstream/dev` and the diff is now clean (no stray files). Could you take another look when you have a moment?

Shizoqua:
Hi @ericspod, I have rebased this onto the latest upstream/dev (now

ericspod (approved):
Hi @Shizoqua it looks good now, thanks.

ericspod:
Thanks @vikashg for help on this one as well!

PR #8936, opened 2026-06-20
Safe eval

ericspod:
Addresses [GHSA-h89g-r5pc-wxfm](https://github.com/Project-MONAI/MONAI/security/advisories/GHSA-h89g-r5pc-wxfm).

### Description

This introduces a `safe_eval` function to evaluate known safe expressions which do not contain member access, calls, indexing, or other expressions which could be used for code injection. Use of `eval` is replaced where appropriate.

### Types of changes


monai/utils/safeeval.py
```diff
@@ -0,0 +1,73 @@
...
+    ast.Load,
+    ast.Constant,
+    ast.BinOp,
+    ast.UnaryOp,
+    ast.Add,
+    ast.Sub,
+    ast.Mult,
+    ast.Div,
+    ast.FloorDiv,
+    ast.Pow,
```
garciadias, line 35: Tried this locally: safe_eval("9**9**9**9**9**9**9") just hangs forever computing a huge integer -- no timeout, no size cap on Pow. Since this function is called with attacker-influenced strings from bundle metadata in _get_fake_spatial_shape (scripts.py:165), that's an easy DoS even after this fix closes the RCE-style bypass. Could we bound Pow/FloorDiv operand size or overall eval cost, either generally in safe_eval or at the _get_fake_spatial_shape call site, and add a regression test?

monai/bundle/scripts.py
```diff
@@ -161,7 +162,7 @@ def _get_fake_spatial_shape(shape: Sequence[str | int], p: int = 1, n: int = 1,
                 for c in _get_var_names(i):
                     if c not in ["p", "n"]:
                         raise ValueError(f"only support variables 'p' and 'n' so far, but got: {c}.")
-                ret.append(eval(i, {"p": p, "n": n}))
+                ret.append(safe_eval(i, {"p": p, "n": n}))
```
garciadias: This is exactly the sink the GHSA advisory called out, and the safe_eval swap correctly blocks the disclosed Attribute/Subscript/Call bypass -- nice. But i here comes straight from bundle metadata (attacker-controlled per the advisory's own reachability note), and safe_eval still allows unbounded ** chains, so a shape entry like "9**9**9**9**9**9**9" will hang this call indefinitely. Worth bounding it here too, not just relying on safe_eval's docstring claim of being safe and secure.

monai/utils/safeeval.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Evaluate the Python expression `expr` using `eval`, but only if it is a safe expression in that its parsed AST
+    contains nodes whose types are given in `allowed_types`. This ensures unsafe node types are excluded, if these
+    are present in the AST a ValueError is raised. The default set of such types in `SAFE_TYPES` ensures only
+    expressions with constants and names can be evaluated, so excludes attribute access, indexing, and calls. Code
+    injection is infeasible through such expressions, so this is a safe and secure way of evaluating simple expressions.
+
+    Args:
+        expr: expression to evaluate, this will be stripped before parsing to avoid indentation complaints
+        globals_vars: global variable mapping, this will be treated as read-only for this function, unlike `eval`
```
garciadias, line 81: The docstring says globals_vars is "treated as read-only... unlike eval", but the real protection is that the AST allowlist blocks Assign/Call nodes entirely -- the dict() copy isn't really what's saving us here. If allowed_types is ever widened for a specific use case, this docstring could give false confidence.

garciadias (commented):
Every other monai.utils.* submodule has an automodule entry in docs/source/utils.rst, but the new safeeval module doesn't (that file isn't touched by this PR). Since safe_eval/SAFE_TYPES are exported as public API from monai.utils, it'd be worth adding an automodule:: monai.utils.safeeval entry so people find this instead of reaching for raw eval() next time.

garciadias (changes requested):
The core fix correctly closes the disclosed attribute/subscript/call bypass from GHSA-h89g-r5pc-wxfm -- verified against the advisory's PoC strings and the new tests. However, the replacement safe_eval primitive still allows unbounded exponentiation (ast.Pow), and is called on the identical attacker-influenced bundle metadata string that the original vulnerability came from -- so a short expression like 9**9**9**9**9**9**9 hangs the process indefinitely. That's a real availability gap in the exact path this PR is meant to secure. Details in the inline comments; everything else looks solid.

monai/utils/safeeval.py
ericspod: That expression is going to hang the Python interpreter no matter where it's used, we want to include the `**` operator for spatial shape calculations so it's not easy to avoid this. There's probably many other ways of abusing expressions like this, to fix this one we'd require that an `BinOp` have a constant right hand side if the operator is `Pow`, but that will catch only this one case. Integer math in Python doesn't overflow, it would help here if it could be turned on. Another approach is to use Numpy which can evaluate these expressions to a value (not always a useful one), this would be leveraged by wrapping numbers in the expression strings:

```python
import numpy as np
import re

pat=re.compile("([\+\-\d\.eE]+)")

def replace(matchobj):
    m=str(matchobj.group(1))
    nptype='np.int32' if str.isdigit(m) else 'np.float32'
    return f'{nptype}({m})'


e=pat.sub(replace, "9**+9.0**9e1")
print(e)  # np.int32(9)**np.float32(+9.0)**np.float32(9e1)
print(eval(e,{"np":np}))  # inf
```

Unless there's a way of doing something like this in Numpy directly. Other approaches using `ast` get more complex very quickly [like this one](https://stackoverflow.com/a/9558001).

monai/utils/safeeval.py
ericspod: What `save_eval` does is make a shallow copy of `global_vars` if it's provided, it's worth noting here that the behaviour is different from `eval` in case anyone thought to use it for getting values out.

monai/bundle/scripts.py
ericspod: Bounding expressions is still going to be an ad-hoc approach, we can restrict power chains like your example but then what about `((88**88)**88)**88` or other constructions? The left hand side is something else while the right hand side is just `88`, so this would be allowed. For this particular application we know the range of plausible expressions is constrained, but rules based filtering will eventually prevent something that should be valid.

monai/utils/safeeval.py
ericspod: Actually this approach is going to be too naive, the parsed expression needs to be transformed to make this work. I've committed an attempt at this.

garciadias (approved):

PR #8938, opened 2026-06-20
Add ConformalPredictor and ConformalCalibrator for split-conformal (LAC) prediction sets

txmed82:
Fixes #8935 (part 1 of 2).

### Description

Brings split-conformal prediction to MONAI via the LAC score
`1 - softmax[y]` (Sadinle et al. 2019). First of two PRs for #8935; the
second will add image-level loss-bounded calibration and a per-voxel
uncertainty mask to `monai/metrics/`.

`monai/inferers/conformal_predictor.py`:
- `ConformalCalibrator`: collect scores on a held-out split, return the
  threshold `qhat` with a `1 - alpha` marginal coverage guarantee. Works
  for classification `(B, C)` and per-voxel segmentation `(B, C, spatial...)`;
  `include_background=False` drops background-labeled samples so the
  threshold isn't dominated by easy background.
- `ConformalPredictor`: an `Inferer` that wraps a network + `qhat` and
  returns the prediction-set mask `{ y : 1 - softmax[y] <= qhat }`. Accepts
  a pre-calibrated `qhat` or calibrates in-band from a loader.

API docs: added `ConformalPredictor` and `ConformalCalibrator` autoclass
entries to `docs/source/inferers.rst`.

### Verification run locally

- `pytest tests/inferers/test_conformal_predictor.py` — 15 passed.
- `pytest tests/inferers/` — 246 passed, 176 skipped (excluding
  `test_zarr_avg_merger.py`, which fails collection due to a missing optional
  `zarr` dep, unrelated to this PR).
- `isort` / `black` / `ruff` — clean on the changed files.
- `mypy monai/inferers/conformal_predictor.py` — no errors in this module.

I have **not** run the full `./runtests.sh --net --coverage` integration suite
or the whole-repo `--quick --unittests --disttests` pass locally, so those
boxes are left unchecked for CI to confirm. `make html` was likewise not run.

### Types of changes
- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [x] New tests added to cover the changes.
- [ ] Integration tests passed locally by running `./runtests.sh -f -u --net --coverage`.
- [ ] Quick tests passed locally by running `./runtests.sh --quick --unittests  --disttests`.
- [x] In-line docstrings updated.
- [ ] Documentation updated, tested `make html` command in the `docs/` folder.

## Summary by CodeRabbit

* **New Features**
[...]

PR #8939, opened 2026-06-21
Add ConformalRiskCalibrator and ConformalRiskPredictor for conformal risk control prediction sets

### LINKED ISSUE #8935: Feat Request: Conformal prediction

I want to be able to give a prediction set with a coverage guarantee beyond the current calibration and TTA.

-- Collect softmax on a held out calibration, compute non conformity scores (LAC:1-softmax), take the empirical quintile as threshold and at inference return score(y) <= threshold. 
-- Then calibrate a single threshold that bounds an image level loss, put a per voxel uncertainty mask

Potentially: 
inferers/conformal_predictor.py
metrics/conformal_risk.py

External libraries like TorchCP

**Additional context**
https://arxiv.org/pdf/2208.02814

txmed82:
Fixes #8935 (part 2 of 2).

### Description

Adds image-level conformal risk control to `monai/metrics/`, the loss-bounded
counterpart to the marginal-coverage `ConformalPredictor` in #8938.

`monai/metrics/conformal_risk.py`:
- `ConformalRiskCalibrator`: calibrate one global threshold `lambda_hat` that
  bounds an image-level loss (`miscoverage` or `false_negative`) on a held-out
  split via the finite-sample selection of Angelopoulos et al. 2022, giving
  `E[L] <= alpha`. Handles classification and variable-size segmentation;
  `include_background=False` drops background voxels.
- `ConformalRiskPredictor`: apply `lambda_hat` and return the prediction-set
  mask plus a per-voxel uncertainty mask (set size > 1).
- `Coverage` / `SetSize`: `CumulativeIterationMetric`s for the coverage vs.
  set-size trade-off.

API docs added to `docs/source/metrics.rst`.

### Types of changes


Whatsonyourmind:
Reviewed this alongside #8938 — the risk-control math is correct, including the parts that are easy to get subtly wrong:

- the finite-sample selection `(n*R_hat(lambda) + B)/(n+1) <= alpha` is implemented exactly as `R_hat <= ((n+1)*alpha - B)/n`;
- `within[0]` is the right infimum, given the empirical risk is non-increasing in `lambda`;
- the grid discretization is conservative — the selected `lambda` is `>=` the true continuous infimum, and since risk is non-increasing a larger `lambda` only lowers risk, so realized risk stays `<= alpha`;
- the `alpha < 1/(n+1)` -> full-set fallback is right.

Controlling an *image-level* loss (rather than per-voxel) is also the principled choice here: image exchangeability is exactly what the CRC theorem needs, so within-image voxel correlation doesn't break the guarantee (unlike a per-voxel marginal treatment).

One robustness item and minor nits:

**1. A custom non-monotone `loss` silently breaks the guarantee.** CRC (Angelopoulos et al. 2022, Thm 1) requires the loss to be non-increasing in `lambda`; both built-ins satisfy that, but the `loss: Callable` escape hatch lets a user pass one that doesn't — and then `within[0]` is no longer an infimum and `E[L] <= alpha` no longer holds, with no error raised. Two cheap options: (a) state the "must be non-increasing in `lambda`" precondition prominently on the `loss` arg, and/or (b) after building `emp_risk` over the grid, assert/warn if it isn't non-increasing (`(emp_risk[1:] <= emp_risk[:-1] + tol).all()`), which catches a bad custom loss essentially for free.

**2. Nits.**
- `B = 1` is the correct/tight bound for both built-ins (their range is exactly `[0, 1]`); only worth exposing as a parameter if you later add a loss with a smaller range.
- The all-background `continue` (the image still counts in `n` but contributes 0 risk) is a reasonable definitional choice — worth a one-line comment so it isn't read as a missed accumulation.
- Same `# ponytail:` placeholder comments as #8938.

Solid PR — glad to see CRC going into MONAI.

txmed82:
@Whatsonyourmind  Appreciate the deep read. Conformal is something I'd really like to see. Took both of your options on non-monotone loss and the loss docstring now states the non increasing in lambda (precondition) and made calibrate() warn if the risk over the grin isn't non increasing. Added one nit but kept B=1 hardcoded since both built-in loses are bounded to 1. Anyway appreciate you.

Whatsonyourmind:
Thanks — and I went back through both PRs:

- **#8938 small-n guard:** the `+inf` fallback when `ceil((n+1)(1-α)) > n` is in and reads exactly right — that's the regime where no finite calibration score can deliver the guarantee, so admitting the full set is the honest move rather than under-covering at the max score. Both correctness items are now closed. 👍
- **#8939 non-monotone warn:** documenting the "non-increasing in λ" precondition *and* warning when the empirical risk isn't monotone over the grid is the right belt-and-suspenders — the docstring alone is too easy to skip past. And keeping `B = 1` hardcoded is correct while both built-in losses are `[0, 1]`-bounded.

One forward idea, since you mentioned wanting more conformal in MONAI and you already ship `Coverage` / `SetSize`: a cheap **size-stratified coverage** readout would add real diagnostic value on top of marginal coverage. LAC (and image-level CRC) control the *marginal* / average quantity, but both can sit at 90% marginal while systematically under-covering the hard cases — e.g. the large-set (ambiguous) stratum covering at ~80% while singletons over-cover. Bucketing realized coverage by prediction-set size surfaces exactly that adaptivity gap, and it's only a few lines on top of what `SetSize` already computes (Angelopoulos & Bates call it size-stratified coverage). Happy to sketch it if useful — it pairs naturally with the coverage/set-size trade-off you're already reporting.

Nice to see CRC + LAC landing in MONAI properly.

Whatsonyourmind:
Glad both loss fixes landed cleanly — and thanks for the `B=1` note, agreed that's the right call while both built-in losses are bounded to 1.

Here's the size-stratified-coverage diagnostic I offered, sketched in your `conformal_risk.py` idiom. The motivation: `Coverage` reports the *marginal* rate, but a predictor can hit its target on average while systematically under-covering exactly the voxels where it's most confident (size-1 sets) and "buying back" the average with near-full sets on ambiguous voxels. That's the failure mode a segmentation reviewer cares about, and it's invisible to marginal coverage. Size-Stratified Coverage (Angelopoulos & Bates, *A Gentle Introduction to Conformal Prediction*, §3.2) stratifies by prediction-set size and reports the **worst-stratum** coverage.

```python
def compute_size_stratified_coverage(
    sets: torch.Tensor, labels: torch.Tensor
) -> tuple[dict[int, tuple[float, int]], float, int]:
    """Coverage stratified by prediction-set size (Angelopoulos & Bates, SSC).

    Args:
        sets: bool tensor ``(B, C, spatial...)``.
        labels: integer class indices ``(B, 1, spatial...)`` or ``(B, spatial...)`` in [0, C).

    Returns:
        ``(per_size, worst_coverage, worst_size)`` where ``per_size`` maps each
        observed set size ``k`` -> ``(coverage_k, n_voxels_k)``, pooled over all
        voxels with that size. ``worst_coverage`` is the minimum over strata and
        ``worst_size`` the size that attains it. Empty strata are omitted.
    """
    sets_flat, labels_flat = _flatten_spatial(sets, labels)              # (N, C), (N,)
    covered = sets_flat.gather(1, labels_flat.unsqueeze(1)).squeeze(1).bool()
    sizes = sets_flat.sum(dim=1)                                         # (N,) in [0, C]

    per_size: dict[int, tuple[float, int]] = {}
    worst_coverage, worst_size = 1.0, -1
    for k in torch.unique(sizes).tolist():
        mask = sizes == k
        cov = covered[mask].float().mean().item()
        per_size[int(k)] = (cov, int(mask.sum()))
        if cov < worst_coverage:
            worst_coverage, worst_size = cov, int(k)
    return per_size, worst_coverage, worst_size
```

Toy check (3 classes, 1000 voxels): size-1 sets deliberately under-covered at 0.80, size-2/3 sets at ~1.0.

```
marginal coverage : 0.898          <- Coverage() alone: looks fine
per-size (cov, n)  : {1: (0.796, 500), 2: (1.0, 300), 3: (1.0, 200)}
worst stratum      : size 1 -> 0.796   <- SSC: the confident voxels are the failure
```

One wiring subtlety if you want a `CumulativeIterationMetric` wrapper (`SizeStratifiedCoverage`) rather than a free function: coverage has to be **pooled within each stratum across the whole test set**, not averaged over per-image SSC values. Strata are unevenly populated per image (some images have no size-3 voxels at all), so a per-image mean would weight a 5-voxel stratum in one image equally with a 5000-voxel stratum in another. The clean way is to buffer per-stratum `(covered_count, total_count)` tensors of length `C+1` in `_compute_tensor` and divide only in `aggregate()` — i.e. carry sufficient statistics, reduce at the end. Happy to draft that class too if the free function looks like the right shape to you.

No rush — this is purely additive on top of what you already have. Nice work on the CRC PR.

txmed82:
Rebased onto latest `dev` (Aug 28, head `71f8661b`). All feedback from @Whatsonyourmind is addressed — the non-monotone custom-loss docstring and runtime warning are in. One unrelated Windows `full-dep` flake (`tests.test_timedcall_dist.test_good_call` timeout) on the prior head; retriggered and required CI is green now. Ready for maintainer review when you have a moment; thanks!

PR #8940, opened 2026-06-21
Enabling MetaTensor Persistent Caching

ericspod:
Addresses [GHSA-636w-j999-g7x5](https://github.com/Project-MONAI/MONAI/security/advisories/GHSA-636w-j999-g7x5).

### Description

This modifies the `PersistentDataset` class to permit storing `MetaTensor` objects. This is done by relying on the `torch.load` functionality to load only safe object types and those white-listed with `torch.serialization.add_safe_globals`. This also uses sha256 hashing in place of md5 in case of security concerns. 

### Types of changes


tests/data/test_persistentdataset.py
```diff
@@ -200,6 +203,134 @@ def test_track_meta_and_weights_only(self, track_meta, weights_only, expected_er
...
+        """
+        Thorough test of metadata loading correctly with MetaTensor. This will store a MetaTensor with safe object types
+        in its metadata dictionary, test the cache file exists and can be safely loaded with weights only, and that the
+        loaded object is another MetaTensor with the correct information
+        """
+        meta = {"test_meta": 123, "foo": "bar", "test_tuple": (1, 2, 3)}
+        imt = MetaTensor(torch.rand(1, 128, 128, 128), meta=dict(meta), affine=torch.rand(4, 4))
+
+        with tempfile.TemporaryDirectory() as tempdir:
+            # cache_dir = os.path.join(os.path.join(tempdir, "cache"), "data")
```
garciadias: Leftover commented-out line from refactoring. The old os.path.join approach is replaced by Path usage on the next line, so this is dead code.

garciadias (commented):
Minor observation: the local variable `data_item_md5` in `_cachecheck` (dataset.py:387) and `CacheNTransDataset` (dataset.py:1621) is now misleading — this PR changes the hash function from md5 to sha256 in `utils.py`, so these variable names should be `data_item_hash` instead. A future reader might assume md5 is still in use.

garciadias (dismissed):
Approved with minor comments. Core change is correct — removing the artificial restriction on track_meta=True + weights_only=True is the right fix since MetaTensor is already registered as a safe global. Tests are thorough, covering both the happy path and unsafe-object rejection. Hash algorithm switch to sha256 is a security improvement. One item to address in release notes: the hash change invalidates all existing cache files, so users with large cached datasets should be warned to expect recomputation.

ericspod:
Thanks @garciadias I've addressed the comments, you'll have to rereview I'm afraid though.

garciadias (approved):

PR #8887, opened 2026-05-31
Fix nnUNet test directory leakage into current working directory

aymuos15:
Fixes #8886 

### Types of changes
- [x] Non-breaking change (fix or new feature that would not break existing functionality).

PR #8944, opened 2026-06-24
Fix nnUNet runner store_true flag command construction (#8941)

### LINKED ISSUE #8941: Flagged nnUNet Runner Command Construction Issue
_⚠️ Potential issue_ | _🟠 Major_ | _⚡ Quick win_

<details>
<summary>🧩 Analysis chain</summary>

🌐 Web query:

`nnUNetv2_train argparse flags store_true --c --val --use_compressed --disable_checkpointing`

💡 Result:

In nnU-Net v2, the command line utility nnUNetv2_train utilizes several argparse flags to control the training and validation process. These flags are implemented as store_true options, meaning they do not require an associated value and serve as toggles [1][2]. The requested flags perform the following functions: --c (Continue training): This flag instructs the trainer to resume training from the most recent checkpoint. It is essential if a training run was interrupted or if you wish to extend a previously stopped training session [2]. --val (Validation only): This flag sets the trainer to skip the training process and instead run only the final validation on the specified fold [2][3]. This is often used after a training run has completed or to re-run validation without re-training [4][3]. --use_compressed: This flag directs the system to use compressed data during training. By default, nnU-Net may decompress data; enabling this flag prevents decompression. It is noted that reading compressed data is significantly more CPU and potentially RAM intensive, and it should only be used when necessary [2]. --disable_checkpointing: This flag disables the saving of model checkpoints during training. This is useful in scenarios where you are testing a configuration or running quick experiments and do not want to consume disk space with multiple large model checkpoint files [2]. In addition to these, the --npz flag is frequently used alongside these commands. Although not explicitly in your list, it is critical to note that --npz is required if you intend to use nnUNetv2_find_best_configuration later, as it forces the model to save the softmax probabilities from the final validation, which are necessary for ensemble creation and model selection [4][3].

Citations:

- 1: https://github.com/MIC-DKFZ/nnUNet/discussions/1543
- 2: https://hackmd.io/n2NJtLqVT1iITYIBpzCLXA
- 3: https://github.com/MIC-DKFZ/nnUNet/blob/master/documentation/how-to/train-models.md
- 4: https://github.com/MIC-DKFZ/nnUNet/blob/master/documentation/how_to_use_nnunet.md

---

</details>

**Fix argv construction for `store_true` kwargs (e.g., `--c`, `--val`)** in `monai/apps/nnunet/nnunetv2_runner.py` (lines 585-605): the `kwargs` loop always appends `str(_value)`, so documented boolean flags (`c`, `val`, `use_compressed`, `disable_checkpointing`) become `--c True` / `--val True` instead of being emitted as bare flags (the same pattern as the special-case `--npz`). Update the builder to emit `--<flag>` only when the value is `True` and omit the flag when `False` (and similarly only include `pretrained_weights` when it’s a real path, not `False`).

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@monai/apps/nnunet/nnunetv2_runner.py` around lines 585 - 605, The argv
builder in the kwargs loop currently always appends str(_value), causing
store_true options to appear as "--flag True" instead of a bare "--flag" and
including false/empty pretrained_weights; update the loop in nnunetv2_runner.py
where cmd is assembled so that for boolean flags (e.g., keys "c", "val",
"use_compressed", "disable_checkpointing" and any other store_true options) you
only append the flag name (prefix + _key) when _value is True and skip it when
False, and for "pretrained_weights" (or "-p") only append the option and its
value when _value is truthy (a non-empty path); keep the existing prefix
selection logic for short vs long options and continue converting items to
strings later.
```

</details>

_Originally posted by @coderabbitai[bot] in https://github.com/Project-MONAI/MONAI/pull/8887#discussion_r3329940183_

aymuos15:
### Description

`nnUNetV2Runner.train_single_model_command` builds the `nnUNetv2_train` argv by iterating `kwargs` and always appending `str(_value)`, so documented `store_true` flags were emitted with a value instead of bare. Passing `c=True` produced `--c True`, `val=True` produced `--val True`, and likewise for `use_compressed` and `disable_checkpointing`. `nnUNetv2_train` declares these as `store_true`, so the trailing `True` is parsed as a positional argument and the command fails. A falsy `pretrained_weights` was similarly emitted as `-pretrained_weights False` instead of being omitted.

The builder now appends `store_true` flags only when their value is truthy and skips them otherwise, and includes `pretrained_weights`/`-p` only when given a real path. Regular value kwargs and the existing `--npz` handling are unchanged.

Fixes #8941, originally flagged in a review thread on #8887.

### Types of changes


aymuos15:
Also fixed the same bug in the `validate` path: `validate_single_model` passed `only_run_validation=True`, which emitted the invalid `--only_run_validation True` (nnUNetv2_train's validation-only flag is the `store_true` `--val`). Now passes `val=True`, routing through the new handling. Added a regression test for this as well.

monai/apps/nnunet/nnunetv2_runner.py
```diff
@@ -596,9 +596,17 @@ def train_single_model_command(
...
             cmd.append("--npz")
 
+        store_true_flags = {"c", "val", "use_compressed", "disable_checkpointing"}
         for _key, _value in kwargs.items():
             prefix = "-" if _key in {"p", "pretrained_weights"} else "--"
-            cmd += [f"{prefix}{_key}", str(_value)]
+            if _key in store_true_flags:
+                if _value:
+                    cmd.append(f"{prefix}{_key}")
+            elif _key in {"p", "pretrained_weights"}:
```
ericspod, line 604: If you're changing to have this `elif` block then there's no need to use the `prefix` variable, just put the prefix directly into the strings you're constructing here.

monai/apps/nnunet/nnunetv2_runner.py
```diff
@@ -758,7 +766,7 @@ def validate_single_model(self, config: str, fold: int, **kwargs: Any) -> None:
             kwargs: this optional parameter allows you to specify additional arguments defined in the
                 ``train_single_model`` method.
         """
-        self.train_single_model(config=config, fold=fold, only_run_validation=True, **kwargs)
+        self.train_single_model(config=config, fold=fold, val=True, **kwargs)
```
ericspod, line 768: Why was this changed?

ericspod (changes requested):
Hi @aymuos15 I had minor comments to make but it looks fine otherwise.

monai/apps/nnunet/nnunetv2_runner.py
aymuos15: Should've stressed this more: --only_run_validation doesn't exist in nnUNet repo, the actual flag is --val

monai/apps/nnunet/nnunetv2_runner.py
aymuos15: committed

ericspod (approved):

aymuos15:
Thank you very much!

PR #8945, opened 2026-06-24
Add AbsoluteVolumeDifferenceMetric to monai.metrics

### LINKED ISSUE #4009: Requesting new metric to calculate volume difference for small fluid volume --> Absolute Volume Difference (AVD)
In 3D Image segmentation tasks, Dice Score is known to be overly sensitive to cases with small fluid volume. Therefore it is not the right metric to do performance analysis. Whereas AVD directly reflects volume size discrepancies.

Reference: [RETOUCH: The Retinal OCT Fluid Detection and Segmentation Benchmark and Challenge ](https://ieeexplore.ieee.org/document/8653407)

For small fluid volume, there should be a metric that directly gives absolute volume difference.
`AVD = abs(|X| - |Y|), where X = Segmentation, and Y= reference`

Currently, I need to manually implement it

**Additional context**

MDSALMANSHAMS:
## Summary

Closes #4009.

Adds AbsoluteVolumeDifferenceMetric and a standalone compute_absolute_volume_difference() function to monai.metrics.

**Why AVD?**  
Dice score is known to be overly sensitive for small-object segmentation (e.g. retinal fluid sub-types in OCT volumes - SRF, IRF, PED), because small volume differences produce large Dice swings. AVD directly reflects volume size discrepancy, making it the standard evaluation metric in the RETOUCH retinal OCT fluid benchmark (Bogunovic et al., IEEE TMI 2019).

## Changes

- **monai/metrics/absolute_volume_difference.py** - new file:
  - AbsoluteVolumeDifferenceMetric(CumulativeIterationMetric) - cumulative class matching the DiceMetric / MeanIoU interface. Supports include_background, 
eduction, get_not_nans, ignore_empty.
  - compute_absolute_volume_difference() - standalone function returning shape [B, C].
- **monai/metrics/__init__.py** - exports both new symbols.
- **	ests/metrics/test_absolute_volume_difference.py** - 14 unit tests.

## Test plan

- [x] Perfect prediction returns zero
- [x] Known volume difference (hand-verified)
- [x] ignore_empty=True sets NaN for empty ground-truth channels
- [x] ignore_empty=False returns raw absolute difference
- [x] include_background=False strips channel 0
- [x] 3D spatial volumes (BCDHW)
- [x] Multi-class output shape
- [x] Cumulative accumulation across batches
- [x] Buffer reset
- [x] Shape mismatch raises ValueError
- [x] Fewer than 3 dimensions raises ValueError
- [x] Top-level import from monai.metrics

All 14 tests pass (python -m unittest tests.metrics.test_absolute_volume_difference -v).


MDSALMANSHAMS:
Hi @ericspod @Nic-Ma @KumoLiu - gentle reminder on this PR adding `AbsoluteVolumeDifferenceMetric` to `monai.metrics`. Would appreciate a review when you get a moment; happy to make any changes. Thanks!

monai/metrics/absolute_volume_difference.py
```diff
@@ -0,0 +1,183 @@
...
+
+from monai.metrics.utils import do_metric_reduction, ignore_background
+from monai.utils import MetricReduction
+
+from .metric import CumulativeIterationMetric
+
+__all__ = ["AbsoluteVolumeDifferenceMetric", "compute_absolute_volume_difference"]
+
+
+class AbsoluteVolumeDifferenceMetric(CumulativeIterationMetric):
```
ericspod, line 24: This class needs to be mentioned in `docs/source/metrics.rst`.

monai/metrics/absolute_volume_difference.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+    Compute the Absolute Volume Difference (AVD) between predicted and ground-truth
```
ericspod, line 26: It's worth noting that for 2D cases this will compute surface areas instead. The values are also not scaled by the voxel/pixel size, so the values may not be in the units of the original images.

monai/metrics/absolute_volume_difference.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+
+    def __init__(
+        self,
+        include_background: bool = True,
+        reduction: MetricReduction | str = MetricReduction.MEAN,
+        get_not_nans: bool = False,
+        ignore_empty: bool = True,
+    ) -> None:
+        """Initialize AbsoluteVolumeDifferenceMetric. See class docstring for argument descriptions."""
```
ericspod: This docstring isn't really necessary.

ericspod (changes requested):
Hi @MDSALMANSHAMS I had some minor comments, mainly needing the new class to be documented. I think we're good otherwise, thanks!

ericspod:
Please also look at the ruff and format comments. Your import test may not be necessary so you can remove that if it's easier.

MDSALMANSHAMS:
Thanks for the review @ericspod! Addressed all points in the latest commit:

- **Docs:** added `AbsoluteVolumeDifferenceMetric` and `compute_absolute_volume_difference` to `docs/source/metrics.rst`.
- **Docstring:** added a note that 2D inputs yield surface areas, and that values are raw voxel/pixel counts (not scaled by voxel/pixel spacing, so not in physical units).
- Removed the unnecessary `__init__` docstring.
- Removed the redundant top-level import test.
- Fixed the `ruff`/format failure (ran `black` + `isort`); `codeformat` is green now.

Also asserted `ValueError` explicitly (instead of a bare `Exception`) on `aggregate()` after `reset()`, per the earlier CodeRabbit note. Let me know if anything else is needed!

ericspod (approved):

PR #8946, opened 2026-06-24
Fix division by zero in Warp for singleton spatial dimensions

VenkateswarluNagineni:
### Description

`Warp.forward` (native `grid_sample` path) normalizes grid coordinates with:

```python
grid[..., i] = grid[..., i] * 2 / (dim - 1) - 1
```

When a spatial dimension has size 1 — a **single-slice volume** `(B, C, 1, H, W)` or a single-row/column image `(B, C, 1, W)` / `(B, C, H, 1)` — `dim - 1 == 0`, so the normalization divides by zero and produces `inf`/`nan` coordinates. With `padding_mode="zeros"`, even a **zero displacement field** then yields:

- `nan` output for 2D single-row/column inputs, and
- all-zeros output for 3D single-slice inputs,

instead of returning the input image unchanged.

#### Fix

Clamp the denominator to 1 so the lone voxel maps to `-1`. This mirrors the guard MONAI already applies in `monai.networks.utils.normalize_transform`, which performs the identical `align_corners=True` normalization and clamps the size with `norm[norm <= 1.0] = 2.0` to avoid exactly this division by zero. Non-singleton dimensions are numerically unchanged.

```python
grid[..., i] = grid[..., i] * 2 / max(dim - 1, 1) - 1
```

#### Verification

Added `test_singleton_spatial_dim` (2D single-row, 2D single-column, 3D single-slice). With a zero displacement field the warped output now equals the input and contains no `nan`. The test fails on the current code and passes with the fix; the existing `Warp` tests are unaffected.

### Types of changes

### DIFF
--- a/monai/networks/blocks/warp.py
+++ b/monai/networks/blocks/warp.py
@@ -155,7 +155,10 @@ class Warp(nn.Module):
 
         if not _use_compiled:  # pytorch native grid_sample
             for i, dim in enumerate(grid.shape[1:-1]):
-                grid[..., i] = grid[..., i] * 2 / (dim - 1) - 1
+                # guard against a singleton spatial dim (e.g. a single-slice volume), where
+                # ``dim - 1 == 0`` would divide by zero; clamp the denominator to 1 so the lone
+                # voxel maps to -1, matching ``monai.networks.utils.normalize_transform``.
+                grid[..., i] = grid[..., i] * 2 / max(dim - 1, 1) - 1
             index_ordering: list[int] = list(range(spatial_dims - 1, -1, -1))
             grid = grid[..., index_ordering]  # z, y, x -> x, y, z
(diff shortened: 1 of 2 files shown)


VenkateswarluNagineni:
Thanks for the review. Addressed both points in the latest commit:

- The test now pins `USE_COMPILED=False` so it always runs through the native `grid_sample` branch — the only path that normalizes the grid (the csrc `grid_pull` path doesn't, so it's unaffected by this fix).
- Added a Google-style docstring to `test_singleton_spatial_dim`.

Verified locally: the test fails on the pre-fix code and passes with the fix; `isort`/`black`/`flake8` are clean on both changed files.

ericspod (approved):
Hi @VenkateswarluNagineni this looks good, thanks!

PR #8951, opened 2026-06-25
Fix FROC num_targets miscount when labels_to_exclude is set

aymuos15:
### Description

`compute_fp_tp_probs_nd` computed `num_targets = max_label - len(labels_to_exclude)`, which only holds when every excluded label is a distinct value present in `[1, max_label]`. An absent, out-of-range, or duplicated entry subtracts targets that were never counted, leaving `num_targets` too small. Since `compute_froc_curve_data` divides cumulative true positives by `num_targets`, this inflates the reported sensitivity.

The count is now the labels in `[1, max_label]` that are not excluded, computed in the existing loop. A regression test covering an out-of-range and a duplicated exclusion is included; both undercount before the fix.

### Types of changes

### DIFF
--- a/monai/metrics/froc.py
+++ b/monai/metrics/froc.py
@@ -11,7 +11,7 @@
 
 from __future__ import annotations
 
-from typing import Any, cast
+from typing import Any
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


ericspod (approved):
Looks good to me, thanks!

PR #8952, opened 2026-06-25
Remove dead and redundant code across monai/

aymuos15:
### Description

Ran a code agent for dead code. Vetted them myself thereafter. 

Removes dead statements, no-ops, and leftover commented-out code. No runtime behavior changes.

| File | Dead/redundant code removed |
|---|---|
| `networks/nets/basic_unet.py` | Stray `print(f"BasicUNet features: ...")` debug call in constructor |
| `networks/nets/basic_unetplusplus.py` | Stray `print(f"BasicUNetPlusPlus features: ...")` debug call in constructor |
| `networks/blocks/text_embedding.py` | Stray `print(self.text_embedding)` debug call in `TextEncoder.forward` |
| `metrics/generalized_dice.py` | No-op self-assignment `y_pred_o = y_pred_o` |
| `networks/layers/simplelayers.py` | No-op self-assignment `filter = filter` in `MeanFilter` |
| `losses/nacl_loss.py` | Redundant `.abs_()` after `.pow_(2)` (operand already non-negative) in L2 branch |
| `losses/image_dissimilarity.py` | Overwrite discarding the `look_up_option`-validated `kernel_type` in `GlobalMutualInformationLoss` |
| `data/ultrasound_confidence_map.py` | Dead `elif` branch with discarded bare `s.shape[0]` expression |
| `inferers/merger.py` | Duplicated recomputation of `is_zarr_v3` in `ZarrAvgMerger` |
| `utils/profiling.py` | Unused module-level `pandas` optional-import (only consumer re-imports locally) |
| `apps/nnunet/utils.py` | Three blocks of commented-out code in `create_new_dataset_json` |
| `apps/vista3d/transforms.py` | Commented-out `AsDiscrete` alternative |
| `transforms/utils.py` | Commented-out `torch.zeros` alternative |
| `networks/layers/filtering.py` | Unreachable commented-out body after `raise` in `PHLFilter.backward` |

### Types of changes

### DIFF
--- a/monai/data/ultrasound_confidence_map.py
+++ b/monai/data/ultrasound_confidence_map.py
@@ -244,6 +244,8 @@ class UltrasoundConfidenceMap:
 
             if iter_idx == 1:
                 vertical_end = s.shape[0]  # Vertical edges length
+            elif iter_idx == 5:
+                s.shape[0]  # Diagonal edges length
 
         # Normalize weights
         s = self.normalize(s)
--- a/monai/data/ultrasound_confidence_map.py
+++ b/monai/data/ultrasound_confidence_map.py
@@ -244,8 +244,6 @@ class UltrasoundConfidenceMap:
 
             if iter_idx == 1:
                 vertical_end = s.shape[0]  # Vertical edges length
-            elif iter_idx == 5:
-                s.shape[0]  # Diagonal edges length
 
         # Normalize weights
         s = self.normalize(s)
--- a/monai/utils/profiling.py
+++ b/monai/utils/profiling.py
@@ -35,6 +35,8 @@ if TYPE_CHECKING:
 else:
     Events = optional_import("ignite.engine", name="Events")
 
+pd, has_pandas = optional_import("pandas")
+
 __all__ = [
     "torch_profiler_full",
     "torch_profiler_time_cpu_gpu",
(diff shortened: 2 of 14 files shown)


monai/data/ultrasound_confidence_map.py
```diff
@@ -244,8 +244,6 @@ def confidence_laplacian(self, padded_index: NDArray, padded_image: NDArray, bet
 
             if iter_idx == 1:
                 vertical_end = s.shape[0]  # Vertical edges length
-            elif iter_idx == 5:
-                s.shape[0]  # Diagonal edges length
```
ericspod, line 248: Should this actually be `vertical_end = s.shape[1]`? I suspect this is really meant to do something.

monai/utils/profiling.py
ericspod: I think this should be removed instead of the `optional_import`.

ericspod (changes requested):
Looks good thanks, I had some minor comments below.

monai/utils/profiling.py
aymuos15: Addressed thank you!

monai/data/ultrasound_confidence_map.py
aymuos15: Have removed this change. I think this is a bug rather than dead code, will investigate this on a later date then. Thank you very much for pointing this out.

ericspod (approved):
Looks good now, thanks.

PR #8953, opened 2026-06-26
Fix non-functional jitter in Warp.get_reference_grid

aymuos15:
### Description

Little messy description as trying to fix multiple things but the jist is: `Warp.get_reference_grid` never applied the `jitter` it advertises and crashed whenever `jitter=True`. 

The grid is built from `torch.arange` (integer dtype) and `self.ref_grid` was assigned `grid.to(ddf)` before the jitter block, so `grid += torch.rand_like(grid)` mutated a local that was never returned, and `torch.rand_like` raises `NotImplementedError` on an integer tensor anyway. Separately, `fork_rng(enabled=seed)` disabled RNG forking when `seed` took its default of `0`, leaking the seeded state into the global RNG.

The grid is now cast to `ddf` before jittering, the jittered tensor is assigned to `self.ref_grid`, and `fork_rng()` isolates the seeded draw. 

The non-jitter path is unchanged. A regression test covers the float/non-integer jittered grid, the integer un-jittered grid, and per-seed reproducibility; it fails before this change with `NotImplementedError`.

### Types of changes

### DIFF
--- a/tests/networks/blocks/warp/test_warp.py
+++ b/tests/networks/blocks/warp/test_warp.py
@@ -145,7 +145,7 @@ class TestWarp(unittest.TestCase):
         self.assertFalse(torch.equal(grid, grid.round()))
 
         ref = Warp().get_reference_grid(ddf, jitter=False)
-        self.assertTrue(bool((ref == ref.round()).all()))
+        self.assertTrue(torch.equal(grid, grid.round()))
 
         same = Warp().get_reference_grid(ddf, jitter=True, seed=7)
         repeat = Warp().get_reference_grid(ddf, jitter=True, seed=7)
--- a/tests/networks/blocks/warp/test_warp.py
+++ b/tests/networks/blocks/warp/test_warp.py
@@ -142,7 +142,7 @@ class TestWarp(unittest.TestCase):
         ddf = torch.zeros(1, 2, 4, 5)
         grid = Warp(jitter=True).get_reference_grid(ddf, jitter=True, seed=0)
         self.assertTrue(grid.is_floating_point())
-        self.assertTrue(bool((grid != grid.round()).any()))
+        self.assertFalse(torch.equal(grid, grid.round()))
 
         ref = Warp().get_reference_grid(ddf, jitter=False)
         self.assertTrue(bool((ref == ref.round()).all()))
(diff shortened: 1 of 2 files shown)


tests/networks/blocks/warp/test_warp.py
```diff
@@ -138,6 +138,21 @@ def test_ill_shape(self):
         with self.assertRaisesRegex(ValueError, ""):
             warp_layer(image=torch.arange(4).reshape((1, 1, 2, 2)).to(dtype=torch.float), ddf=torch.zeros(1, 2, 3, 3))
 
+    def test_jitter(self):
+        ddf = torch.zeros(1, 2, 4, 5)
+        grid = Warp(jitter=True).get_reference_grid(ddf, jitter=True, seed=0)
+        self.assertTrue(grid.is_floating_point())
+        self.assertTrue(bool((grid != grid.round()).any()))
```
ericspod:
```suggestion
        self.assertFalse(torch.equal(grid, grid.round()))
```

tests/networks/blocks/warp/test_warp.py
```diff
@@ -138,6 +138,21 @@ def test_ill_shape(self):
...
+
+        ref = Warp().get_reference_grid(ddf, jitter=False)
+        self.assertTrue(bool((ref == ref.round()).all()))
```
ericspod:
```suggestion
        self.assertTrue(torch.equal(grid, grid.round()))
```

ericspod (dismissed):
Hi @aymuos15 super minor changes you can do if you feel needed but good to go as it is.

aymuos15:
Thanks for pointing those out. Will remember it for the next one!

tests/networks/blocks/warp/test_warp.py
```diff
@@ -139,6 +139,21 @@ def test_ill_shape(self):
         with self.assertRaisesRegex(ValueError, ""):
             warp_layer(image=torch.arange(4).reshape((1, 1, 2, 2)).to(dtype=torch.float), ddf=torch.zeros(1, 2, 3, 3))
 
+    def test_jitter(self):
+        ddf = torch.zeros(1, 2, 4, 5)
+        grid = Warp(jitter=True).get_reference_grid(ddf, jitter=True, seed=0)
+        self.assertTrue(grid.is_floating_point())
+        self.assertFalse(torch.equal(grid, grid.round()))
+
+        ref = Warp().get_reference_grid(ddf, jitter=False)
```
ericspod:
```suggestion
        grid = Warp().get_reference_grid(ddf, jitter=False)
```

ericspod (approved):

PR #8956, opened 2026-06-27
fix(data): avoid divide-by-zero in pydicom affine for single-slice volumes

### LINKED ISSUE #8925: fix: division by zero in DICOMReader._get_affine for single-slice 3D volumes
## Problem

In `monai/data/image_reader.py`, the `DICOMReader._get_affine` method at line 764 computes:

```python
n = metadata[MetaKeys.SPATIAL_SHAPE][-1]
k1, k2, k3 = (t1n - sx) / (n - 1), (t2n - sy) / (n - 1), (t3n - sz) / (n - 1)
```

When `n == 1` (single-slice 3D DICOM segmentation), `n - 1 = 0` causes a **ZeroDivisionError**.

This occurs in the segmentation code path (`_get_seg_data`) at line 898, where `lastImagePositionPatient` is always set from the frame metadata regardless of the number of frames. For a single-frame segmentation, the first and last positions are identical, so `t1n - sx = 0`, `t2n - sy = 0`, `t3n - sz = 0`, and `n - 1 = 0`, resulting in `0 / 0`.

## Proposed Fix

Guard the `lastImagePositionPatient` computation in `_get_affine` with `n > 1`:

```python
if "lastImagePositionPatient" in metadata and n > 1:
```

## Affected Files

- `monai/data/image_reader.py` (line 761)

SID-6921:
Summary:
- fix PydicomReader._get_affine to skip lastImagePositionPatient z-step derivation when there is only one slice
- add regression tests for single-slice and multi-slice metadata paths

Why:
For single-slice 3D DICOM segmentation metadata, n == 1 caused division by zero in affine z-step computation.

Validation:
- python -m pytest tests/data/test_init_reader.py -k pydicom_reader_get_affine -q
- python -m pre_commit run --files monai/data/image_reader.py tests/data/test_init_reader.py

Closes #8925

### DIFF
--- a/tests/data/test_init_reader.py
+++ b/tests/data/test_init_reader.py
@@ -19,6 +19,7 @@ import numpy as np
 
 from monai.data import ITKReader, NibabelReader, NrrdReader, NumpyReader, PILReader, PydicomReader
 from monai.transforms import LoadImage, LoadImaged
+from monai.utils import MetaKeys
 from tests.test_utils import SkipIfNoModule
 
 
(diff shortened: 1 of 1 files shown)


monai/data/image_reader.py
```diff
@@ -758,7 +758,7 @@ def _get_affine(self, metadata: dict, lps_to_ras: bool = True):
         affine[2, 3] = sz
 
         # 3d
-        if "lastImagePositionPatient" in metadata:
+        if "lastImagePositionPatient" in metadata and metadata[MetaKeys.SPATIAL_SHAPE][-1] > 1:
```
ericspod: Hi @SID-6921 I'm afraid we just merged another PR that addressed this issue. Could you please look at the current state of the repo and see if that satisfies the issue? If so we can drop the change in this file but keep any of your tests that add value. Thanks!

SID-6921:
Thanks for the heads-up. I synced with current dev and removed the duplicate change in monai/data/image_reader.py from this PR. I kept the regression tests in 	ests/data/test_init_reader.py (single-slice and multi-slice affine metadata coverage), which still add value. Local validation passes for the updated PR scope.

SID-6921:
Update done: I rebased the PR branch onto current dev and force-pushed it so the duplicate image_reader.py fix is no longer part of this PR. The PR now keeps only the added regression coverage in 	ests/data/test_init_reader.py (single-slice and multi-slice affine metadata cases, guarded by optional Pydicom availability).

tests/data/test_init_reader.py
```diff
@@ -100,6 +101,40 @@ def test_nibabel_reader_avoids_eager_c_order_copy(self):
                     # (F-order) layout from nibabel should be preserved here.
                     self.assertFalse(data.flags.c_contiguous)
 
+    @SkipIfNoModule("Pydicom")
```
ericspod:
```suggestion
    @SkipIfNoModule("pydicom") 
```
Module names are case sensitive, same in the next test.

tests/data/test_init_reader.py
```diff
@@ -100,6 +101,38 @@ def test_nibabel_reader_avoids_eager_c_order_copy(self):
...
+            "00280030": {"Value": [1.0, 1.0]},
+            "lastImagePositionPatient": np.array([0.0, 0.0, 4.0]),
+            MetaKeys.SPATIAL_SHAPE: np.array([8, 8, 5]),
+        }
+
+        affine = reader._get_affine(metadata, lps_to_ras=False)
+
+        np.testing.assert_allclose(affine[0, 2], 0.0)
+        np.testing.assert_allclose(affine[1, 2], 0.0)
+        np.testing.assert_allclose(affine[2, 2], 1.0)
```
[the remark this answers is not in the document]
ericspod: @SID-6921 please have a look at this comment, I think the change makes sense. It's possible the tests weren't being run if the module name was "Pydicom" instead of "pydicom".

ericspod:
Thanks @SID-6921 I had minor comments about the module name and the Coderabbit comment is worth looking at.

SID-6921:
Thanks for the review comments. I applied the module-name fix and force-pushed: all SkipIfNoModule uses for pydicom in this test file are now lowercase (pydicom), including the two new affine regression tests. Local pytest and pre-commit passed after this change.

ericspod (approved):
Looks good now, thanks!

PR #8957, opened 2026-06-29
Add affine-aware landmark heatmap generation

mustafamm072:
### Description

Adds opt-in world-coordinate landmark support to `GenerateHeatmapd`.

By default, behavior is unchanged. With `coordinate_space="world"`, landmarks are transformed into the reference image voxel space using `ref_image_keys` before heatmap generation. The transform can also emit per-landmark visibility masks via `visibility_keys`.

Related to #7486.

### Testing

- Added tests for translated, rotated, anisotropic, and scaled reference affines.
- Added tests for `MetaTensor` point affines and visibility masks.
- Ran `py_compile` and `git diff --check`.


docs/source/whatsnew_1_6.md
```diff
@@ -9,6 +9,7 @@
 - Nested dot-notation key access in `ConfigParser`.
 - Auto3DSeg algo serialization migrated from pickle to JSON for improved security and portability.
 - Global coordinates support in spatial crop transforms. These now support global coordinate mode, allowing crops to be specified in world/global coordinates rather than local image indices, improving interoperability with physical-space annotations.
+- `GenerateHeatmapd` can convert world-coordinate landmarks to reference-image voxel space and emit landmark visibility masks.
```
ericspod: Version 1.6 has been released so this shouldn't be included here.

monai/transforms/post/dictionary.py
```diff
@@ -590,13 +609,20 @@ class GenerateHeatmapd(MapTransform):
...
+    )
+    _SUPPORTED_COORDINATE_SPACES = {"voxel", "world"}
 
     def __init__(
         self,
         keys: KeysCollection,
         sigma: Sequence[float] | float = 5.0,
         heatmap_keys: KeysCollection | None = None,
         ref_image_keys: KeysCollection | None = None,
+        coordinate_space: str | Sequence[str] = "voxel",
```
ericspod: New arguments should be at the end of the argument list so that any existing invocation of this constructor using positional arguments isn't broken. The order of the docstrings should match as well.

monai/transforms/post/dictionary.py
```diff
@@ -711,6 +777,46 @@ def _update_spatial_metadata(self, heatmap: MetaTensor, spatial_shape: tuple[int
...
+
+    def _get_reference_affine(self, reference: Any) -> torch.Tensor:
+        if reference is None:
+            raise ValueError("coordinate_space='world' requires ref_image_keys or a reference affine.")
+        affine = getattr(reference, "affine", None)
+        if affine is not None:
+            return affine
+        if isinstance(reference, (torch.Tensor, np.ndarray)) and reference.shape in ((3, 3), (4, 4)):
+            return reference
+        raise ValueError("coordinate_space='world' requires reference data with an affine matrix.")
```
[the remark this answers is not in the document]
ericspod: I think this is a valid concern. If you want `reference` to possibly be the affine itself you will need some other mechanism to indicate this.

monai/transforms/post/dictionary.py
```diff
@@ -711,6 +772,46 @@ def _update_spatial_metadata(self, heatmap: MetaTensor, spatial_shape: tuple[int
...
+        if isinstance(reference, (torch.Tensor, np.ndarray)) and reference.shape in ((3, 3), (4, 4)):
+            return cast(torch.Tensor, convert_to_tensor(reference, dtype=torch.float32, track_meta=False))
+        raise ValueError("coordinate_space='world' requires reference data with an affine matrix.")
+
+    def _compute_visibility(self, points: Any, spatial_shape: tuple[int, ...]) -> torch.Tensor:
+        points_t = cast(torch.Tensor, convert_to_tensor(points, dtype=torch.float32, track_meta=False))
+        if points_t.ndim != 2:
+            raise ValueError(f"{self._ERR_INVALID_POINTS} Got {points_t.ndim}D tensor.")
+        bounds = torch.as_tensor(spatial_shape, dtype=points_t.dtype, device=points_t.device)
+        return torch.isfinite(points_t).all(dim=1) & (points_t >= 0).all(dim=1) & (points_t < bounds).all(dim=1)
```
ericspod:
```suggestion
        return points_t.mul(torch.isfinite(points_t)).clamp_(0, bounds - 1).all(dim=1)
```
My concern is that the original expression involves a lot of tensor copies that's not efficient. Please check this is correct if you think it's worth changing.

monai/transforms/post/dictionary.py
```diff
@@ -632,6 +663,8 @@ def __call__(self, data: Mapping[Hashable, Any]) -> dict[Hashable, Any]:
                     heatmap.affine = reference.affine
                     self._update_spatial_metadata(heatmap, shape)
             d[out_key] = heatmap
+            if visibility_key is not None:
```
ericspod, line 668: If `visibility_key` is None then the computation for `visiblity` is done needlessly, can line 652 be placed in this condition block?

ericspod (changes requested):
hi @mustafamm072 thanks for this addition, I had some comments but with changes we should be good to go.

monai/transforms/post/dictionary.py
mustafamm072: Done. Moved coordinate_space and visibility_keys to the end of the argument list and reordered the docstring to match.

docs/source/whatsnew_1_6.md
mustafamm072: Removed. thanks, hadn't caught that 1.6 was already out.

monai/transforms/post/dictionary.py
mustafamm072: Good catch. _compute_visibility is now only called inside the if visibility_key is not None: block.

monai/transforms/post/dictionary.py
mustafamm072: Reduced the copies by combining the checks into a single boolean mask and doing one .all(dim=1) reduction. I didn't use the exact points_t.mul(...).clamp_(0, bounds-1) form because it changes the semantics. clamp_ forces every point in bounds so all points would report visible, and multiplying by the finite-mask doesn't drop NaNs (0.0 * nan == nan). The new version keeps correctness while removing the redundant reductions. Let me know if you'd prefer a different approach.

vikashg (approved):

monai/transforms/post/dictionary.py
mustafamm072: @ericspod Can you please review this change?

ericspod (approved):

PR #8963, opened 2026-06-30
Support configurable rotation order in create_rotate

### LINKED ISSUE #6029: support different ordering of rotating about x, y, z

https://github.com/Project-MONAI/MONAI/blob/f5708ea89af86ca74f88e93dabb0ec94c4944ac1/monai/transforms/utils.py#L689-L693

the current implementation is hard-coded to Rx @ Ry @ Rz

aymuos15:
Fixes #6029 .

### Description

`create_rotate` hard-coded 3D rotations to the intrinsic `Rx @ Ry @ Rz` composition. This adds a `rotate_order` parameter following the convention of `scipy.spatial.transform.Rotation.from_euler`: a string of up to three axes from `{x, y, z}`, where lower case selects extrinsic rotations (about the fixed world axes) and upper case selects intrinsic rotations (about the moving body axes). The default `"XYZ"` reproduces the previous behaviour exactly, so existing pipelines are unaffected.

The name avoids collision with the spline interpolation order already selected via `mode`. The parameter is threaded through `functional.rotate`, `Rotate`, `RandRotate`, `AffineGrid`, `RandAffineGrid`, `Affine`, `RandAffine` and their dictionary variants. Invalid sequences raise `ValueError`, and 2D inputs ignore it.

A new test module checks that the default matches the legacy matrix, that every supported axis sequence matches scipy for both the numpy and torch backends, that invalid sequences raise, that 2D inputs ignore the order, and that the `Rotate` transform honours it while remaining invertible.

### Types of changes

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -936,6 +936,9 @@ class Rotate(InvertibleTransform, LazyTransform):
             the output data type is always ``float32``.
         lazy: a flag to indicate whether this transform should execute lazily or not.
             Defaults to False
+        rotate_order: for 3D inputs, the order in which the axes are rotated about, following the convention of
+            :py:func:`scipy.spatial.transform.Rotation.from_euler`. See
+            :py:func:`monai.transforms.utils.create_rotate`. Defaults to ``"XYZ"`` (the legacy behaviour).
     """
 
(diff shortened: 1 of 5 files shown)


ericspod (approved):
Hi @aymuos15 this looks good to me thanks!

PR #8967, opened 2026-07-01
Performence: implement KDTree for CPU surface distance calculation

qwepablo12:
### Description
Addresses #8909. 

This PR optimizes the `get_surface_distance` computation on CPU by utilizing `scipy.spatial.KDTree` instead of calculating the full exact distance transform field (`monai_distance_transform_edt`) and slicing it afterwards. 

By extracting contour coordinates via `np.argwhere` and building a spatial KDTree, we completely avoid heavy background distance calculations. This significantly improves computation speed and reduces memory overhead when calculating surface metrics on large 3D medical volumes on CPU.

### Types of changes

### DIFF
--- a/monai/metrics/utils.py
+++ b/monai/metrics/utils.py
@@ -291,45 +291,38 @@ def get_surface_distance(
             dis = dis[seg_gt]
             return convert_to_dst_type(dis, seg_pred, dtype=dis.dtype)[0]
         if distance_metric == "euclidean":
-            # Проверяем, находятся ли данные на CPU (numpy или torch на cpu)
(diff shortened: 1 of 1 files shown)


ericspod:
Hi @qwepablo12 thanks for the contribution but a previous PR was already looking into this problem. I think your changes are redundant now but have a look at the current state of `dev` to see if there's anything in that implementation you think should be changed, we can then reconsider this PR.

qwepablo12:
> Hi @qwepablo12 thanks for the contribution but a previous PR was already looking into this problem. I think your changes are redundant now but have a look at the current state of `dev` to see if there's anything in that implementation you think should be changed, we can then reconsider this PR.

Hi @ericspod,
Thank you for the review and for letting me know!

PR #8969, opened 2026-07-01
Feat: add ignore_index support to DiceLoss

qwepablo12:
### Description
Addresses #8734.

This PR introduces `ignore_index` support for `DiceLoss`. When `ignore_index` is specified, the target regions matching this index are masked out and completely excluded from the Dice coefficient computation (both numerator and denominator). This ensures that ignored regions/classes do not affect the loss values or gradients during training.

### Types of changes

### DIFF
--- a/monai/losses/dice.py
+++ b/monai/losses/dice.py
@@ -67,45 +67,11 @@ class DiceLoss(_Loss):
         batch: bool = False,
         weight: Sequence[float] | float | int | torch.Tensor | None = None,
         soft_label: bool = False,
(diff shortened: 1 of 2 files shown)


ericspod:
Hi @qwepablo12 we have an existing PR https://github.com/Project-MONAI/MONAI/pull/8757 to implement this feature, it's been a bit neglected but we'll get it through eventually. I think this covers what this PR intends to do so I don't feel we'd merge this one, but thanks for the effort.

PR #8975, opened 2026-07-02
Perf: skip redundant full-image mask on StdShiftIntensity nonzero=False path

### LINKED ISSUE #8972: StdShiftIntensity does redundant full-image mask + gather + scatter on the default nonzero=False path
### Describe the bug

`StdShiftIntensity._stdshift` (`monai/transforms/intensity/array.py`) always builds a full-image boolean mask, gathers every voxel through it, computes the std on that copy, and scatters back:

```python
slices = (img != 0) if self.nonzero else ones(img.shape, dtype=bool)
if slices.any():
    offset = self.factor * std(img[slices])
    img[slices] = img[slices] + offset
```

On the default `nonzero=False` path (also the default for `RandStdShiftIntensity`), `slices` is all-`True`, so `std(img[slices]) == std(img)` and the scatter is just `img + offset`. The mask allocation, `.any()`, gather and scatter are extra.

aymuos15:
Fixes #8972 .

### Description

On the default `nonzero=False` path `StdShiftIntensity._stdshift` built an all-True boolean mask the size of the image and shifted through it (`img[slices] = img[slices] + offset`), forcing a full advanced-index gather and scatter plus the mask allocation and `.any()` scan even though every voxel is selected. That is equivalent to shifting the image directly, so the `nonzero=False` branch now does `img + factor * std(img)`. Output is bit-for-bit identical and the `nonzero=True` path is untouched. `RandStdShiftIntensity` benefits too, since it also defaults to `nonzero=False`. Measured across 2D/3D, single and multi channel, float32 and float64 on both backends (best-of-3, CPU); output verified equal in every configuration:

| shape | dtype | backend | current (ms) | proposed (ms) | speedup |
|---|---|---|---|---|---|
| 1x256x256 | f32 | numpy | 0.123 | 0.048 | 2.58x |
| 1x256x256 | f32 | torch | 0.563 | 0.039 | 14.47x |
| 1x1024x1024 | f32 | numpy | 3.016 | 1.294 | 2.33x |
| 1x1024x1024 | f32 | torch | 7.040 | 0.274 | 25.66x |
| 1x1024x1024 | f64 | torch | 9.644 | 0.601 | 16.04x |
| 1x64x64x64 | f32 | numpy | 0.637 | 0.251 | 2.54x |
| 1x64x64x64 | f64 | torch | 2.646 | 0.095 | 27.84x |
| 1x128x128x128 | f32 | numpy | 7.019 | 2.914 | 2.41x |
| 1x128x128x128 | f32 | torch | 41.618 | 0.630 | 66.09x |
| 1x128x128x128 | f64 | torch | 44.534 | 1.825 | 24.40x |
| 4x96x96x96 | f32 | torch | 68.152 | 1.357 | 50.23x |
| 4x160x160x160 | f32 | numpy | 128.429 | 82.662 | 1.55x |
| 4x160x160x160 | f32 | torch | 342.641 | 16.727 | 20.48x |
| 4x160x160x160 | f64 | torch | 354.009 | 41.652 | 8.50x |

numpy ranges 1.55x to 2.65x and torch 8.5x to 66x across the full sweep; the largest gains are on torch, where all-True boolean-mask indexing is especially costly.

### Types of changes

### DIFF
--- a/monai/transforms/intensity/array.py
+++ b/monai/transforms/intensity/array.py
@@ -354,16 +354,16 @@ class StdShiftIntensity(Transform):
         self.dtype = dtype
 
     def _stdshift(self, img: NdarrayOrTensor) -> NdarrayOrTensor:
-        ones: Callable
         std: Callable
         if isinstance(img, torch.Tensor):
-            ones = torch.ones
             std = partial(torch.std, unbiased=False)
         else:
-            ones = np.ones
             std = np.std
 
-        slices = (img != 0) if self.nonzero else ones(img.shape, dtype=bool)
+        if not self.nonzero:
+            shifted: NdarrayOrTensor = img + self.factor * std(img)
+            return shifted
+        slices = img != 0
         if slices.any():
             offset = self.factor * std(img[slices])
             img[slices] = img[slices] + offset
--- a/monai/transforms/intensity/array.py
+++ b/monai/transforms/intensity/array.py
@@ -360,11 +360,8 @@ class StdShiftIntensity(Transform):
         else:
             std = np.std
 
-        if not self.nonzero:
-            shifted: NdarrayOrTensor = img + self.factor * std(img)
-            return shifted
-        slices = img != 0
-        if slices.any():
+        slices = (img != 0) if self.nonzero else ()
+        if not self.nonzero or slices.any():
             offset = self.factor * std(img[slices])
             img[slices] = img[slices] + offset
         return img
--- a/monai/transforms/intensity/array.py
+++ b/monai/transforms/intensity/array.py
@@ -361,7 +361,7 @@ class StdShiftIntensity(Transform):
             std = np.std
 
         slices = (img != 0) if self.nonzero else ()
-        if not self.nonzero or slices.any():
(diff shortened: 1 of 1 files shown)


monai/transforms/intensity/array.py
```diff
@@ -354,16 +354,16 @@ def __init__(
...
             std = partial(torch.std, unbiased=False)
         else:
-            ones = np.ones
             std = np.std
 
-        slices = (img != 0) if self.nonzero else ones(img.shape, dtype=bool)
+        if not self.nonzero:
+            shifted: NdarrayOrTensor = img + self.factor * std(img)
+            return shifted
+        slices = img != 0
```
ericspod:
```suggestion
        slices = (img != 0) if self.nonzero else ()
```
You can use an empty tuple to select everything in a tensor/array. This would give you a view which should have very minimal overhead.

aymuos15: Thank you very much for pointing it out. Committed.

ericspod (approved):

PR #8976, opened 2026-07-02
Enable macOS Install Test Only

ericspod:
### Description

macOS tests are really slow and disabled for now in the CI system, see #8864. This change will enable most of the test again except for the actual unit tests, so this will test the build and install only. Subsequent changes will look into re-enabling a subset of tests for macOS. 

### Types of changes


MMelQin (approved):
Thanks!

Issue #8971, opened 2026-07-02
Skip CI Tests Based on Checks, Not Action File


ericspod:
**Is your feature request related to a problem? Please describe.**
An attempt was made to avoid running tests when only documentation was changed by adding patterns to `paths-ignore` in the actions files. This has the effect of making checks not run which blocks PRs, the hope was instead that these would show up as successfully ignored but it's not to be. 

The alternative is to conditions the steps of jobs within actions on a check against changes. My idea is to use checking code in one job to detect if there's code change in Python files or changes in any other non-documentation file, then condition steps on that result. This would cause documentation only changes to not trigger tests, so changes to documentation files are ignored, and Python changes affecting only docstrings or comments are ignored.

**Describe the solution you'd like**
A prototype solution is here: https://github.com/ericspod/ActionsTest/blob/768a042fe2ae0693b528486be438b74081d77406/.github/workflows/test.yml 

This uses a Python script to parse source files and compare them at the AST level, while also ignoring documentation files. Any other file changes trigger tests just in case. This uses the standard library so no dependencies exist other than Python. 

The first job uses this script inside a bash loop going through every changed file. The output is then used by subsequent jobs to skip the test steps if no changes are found, so the job runs successfully even if nothing is done.

**Describe alternatives you've considered**
Using `paths-ignore` won't work in conjunction with require job passes for PRs to be mergeable. There isn't another actions mechanism to do this it seems.

aymuos15:
Comments never reach the AST ([`compare_sources.py#L32-L33`](https://github.com/ericspod/ActionsTest/blob/768a042fe2ae0693b528486be438b74081d77406/compare_sources.py#L32-L33)), so deleting a `# type: ignore` or `# noqa` would count as docs-only and skip lint/mypy. Maybe we should always run the static-checks?

```python
>>> import ast
>>> ast.dump(ast.parse("x = 1  # type: ignore")) == ast.dump(ast.parse("x = 1"))
True
```

ericspod:
> Comments never reach the AST ([`compare_sources.py#L32-L33`](https://github.com/ericspod/ActionsTest/blob/768a042fe2ae0693b528486be438b74081d77406/compare_sources.py#L32-L33)), so deleting a `# type: ignore` or `# noqa` would count as docs-only and skip lint/mypy. Maybe we should always run the static-checks?
> 
> >>> import ast
[...]

PR #8979, opened 2026-07-03
Gate CI test jobs via change scanning instead of workflow-level paths-ignore

SID-6921:
## Summary
- remove workflow-level paths-ignore from cicd_tests.yml so required CI checks always register on PRs
- add a lightweight changes job that computes whether non-doc files changed
- gate heavy CI jobs (static-checks, min-dep, ull-dep, packaging) on 
eeds.changes.outputs.run_tests

## Why
Workflow-level paths-ignore can prevent required checks from appearing at all for doc-only PRs, which blocks mergeability. With this change, checks are always created, and heavy jobs are skipped only when changes are docs-only.

## Change Detection Logic
- treats docs/**, *.md, and *.rst as documentation-only changes
- runs tests when any other file changes
- handles both pull_request and push events

Related: #8971


ericspod (changes requested):
Hi @SID-6921 thanks for this contribution but I don't see this helping enough for what I had wanted in the issue. From the script it looks like this will only detect when appropriate files are changed, what I had experimented with [here](https://github.com/ericspod/ActionsTest/blob/768a042fe2ae0693b528486be438b74081d77406/.github/workflows/test.yml#L50) was checking that Python files have actual code changes, such that docstring or comment changes don't trigger testing. I think that's important, and the issues from Copilot and Coderabbit are relevant to look at also. I think my implementation could use work to prevent failure allowing tests to be skipped as well, but skipping tests for docstring changes is also desired. If you wanted to update from these comments I can re-review, otherwise I'd prefer to close this PR.

PR #8983, opened 2026-07-06
Clarify CUDA/CuPy install requirements in docs (#8106)

### LINKED ISSUE #8106: Documentation is not entirely clear about CUDA/`cupy` installation requirements

In one of our processing scripts, we currently perform inference using a MONAI model, and set `torch.device` to `'cuda'` to perform inference on GPU. We use `device` when loading the models weights, as well as for the `sw_device` argument of the `sliding_window_inference` function. 

Initially, GPU usage was working OK, I think? But later, it seems we forgot to install the necessary optional dependencies, because we ran into the following error:

```python
  File "python3.9/site-packages/monai/transforms/spatial/array.py", line 2091, in __call__
    img_np = (convert_to_cupy if is_cuda else convert_to_numpy)(img_t, wrap_sequence=True)
  File "python3.9/site-packages/monai/utils/type_conversion.py", line 246, in convert_to_cupy
    data = cp.asarray(data, dtype)
  File "python3.9/site-packages/monai/utils/module.py", line 441, in __getattr__
    raise self._exception
  File "python3.9/site-packages/monai/utils/module.py", line 399, in optional_import
    pkg = __import__(module)  # top level module
monai.utils.module.OptionalImportError: import cupy (No module named 'cupy').

```

Naturally, the thought is to `pip install cupy`, but this fails, because the actual package is called `cupy-cuda12x`. 

Looking at the list of potential "extras" options, it's not entirely clear how to pick the correct extra to solve this issue. We started with `monai[all]` just to troubleshoot, and this does fix the problem, but it seems like overkill, installing dependencies we do not actually need. With some guesswork, we learned that `monai[cucim]` will install `cupy`. But, even then, it is not 100% clear from the documentation alone if this will install all of the necessary dependencies to perform inference without error.

I think it would make sense if the installation section had an "installing MONAI for GPU usage" section that explains that, when `is_cuda` is true, MONAI will try to use CuPY, and thus [insert blank] dependencies are necessary (`monai[cucim]`)?

It might also help to have an extra specifically called `monai[cuda]` to make it more discoverable/intuitive. But, if this turns out to just be an alias for `monai[cucim]`, then maybe it's not necessary.

Thank you!

Shizoqua:
Fixes #8106

### Description

The installation docs did not make it clear that some MONAI transforms require CuPy for GPU-accelerated array operations. Users who ran GPU inference hit `OptionalImportError: import cupy (No module named 'cupy')` with no obvious guidance — `pip install cupy` fails because the real package is CUDA-version specific (`cupy-cuda12x`), and it was not documented that the `cucim` extra provides CuPy.

This PR updates `docs/source/installation.md` to:
- Promote the GPU/CUDA setup steps into their own `## GPU-enabled installation (CUDA and CuPy)` section (with a Table of Contents entry) instead of an unlabeled paragraph.
- Explain that some transforms use CuPy (e.g. via `convert_to_cupy`), what error appears when it is missing, and that CuPy is installed through the `cucim` extra (`pip install 'monai[cucim]'`).
- Note that installing CuPy directly requires the CUDA-specific package name (`cupy-cuda12x` / `cupy-cuda11x`), linking to the CuPy installation guide.
- Add a cross-reference from the recommended-dependencies list.

Docs-only change; no code or public API is modified.

### Types of changes

### DIFF
--- a/docs/source/installation.md
+++ b/docs/source/installation.md
@@ -38,7 +38,7 @@ Ignite](https://pytorch.org/ignite/), please follow the instructions:
 The installation commands below usually end up installing the CPU variant of PyTorch. To install GPU-enabled PyTorch:
 
 1. Install the latest NVIDIA driver.
-1. Check [PyTorch Official Guide](https://pytorch.org/get-started/locally/) for the recommended CUDA versions. For Pip package, the user needs to download the CUDA manually, install it on the system, and ensure CUDA_PATH is set properly.
+1. Check the [PyTorch Official Guide](https://pytorch.org/get-started/locally/) for the recommended CUDA versions. For Pip packages, PyTorch wheels already bundle the CUDA runtime, so you only need to pick the CUDA version matching your driver from the selector and install with the provided command. You do not need to manually download CUDA or set `CUDA_PATH`.
 1. Continue to follow the guide and install PyTorch.
 1. Install MONAI using one of the ways described below.
 
@@ -47,15 +47,18 @@ however, additionally use [CuPy](https://cupy.dev/) for GPU-accelerated array op
 when a transform converts a CUDA tensor via `convert_to_cupy`). If CuPy is not installed, these code
 paths raise `OptionalImportError: import cupy (No module named 'cupy')`.
 
-CuPy is provided through the `cucim` extra, so installing MONAI with that extra pulls in a compatible
-CuPy build:
+MONAI provides a dedicated `cupy` extra that installs a compatible CuPy build:
 
 ```bash
-pip install 'monai[cucim]'
+pip install 'monai[cupy]'
 ```
 
+The `cucim` extra installs [cuCIM](https://github.com/rapidsai/cucim) (`cucim-cu12` or `cucim-cu13`
+depending on your Python version), which is a separate GPU image-processing library and does not
+install CuPy.
+
 If you prefer to install CuPy directly, note that the PyPI package name is CUDA-version specific
-(e.g. `cupy-cuda12x` for CUDA 12.x, `cupy-cuda11x` for CUDA 11.x) rather than plain `cupy`. See the
+(e.g. `cupy-cuda12x` for CUDA 12.x, `cupy-cuda13x` for CUDA 13.x) rather than plain `cupy`. See the
 [CuPy installation guide](https://docs.cupy.dev/en/stable/install.html) for the correct package for
 your CUDA toolkit.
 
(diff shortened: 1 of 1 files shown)


vikashg:
@Shizoqua Can you resubmit it with DCO and I can approve.

Shizoqua:
@vikashg This was resubmitted with DCO sign-off (commit 4784bf3fd, pushed 08-18) and all checks are now green (DCO, DeepSource, build-docs, pre-commit, CodeRabbit all pass). Per your earlier note, this should be ready for your approval. Thanks!

ericspod:
Hi @Shizoqua please do look at the Coderabbit comments, they are relevant I think. Note that we have also updated the pyproject.toml file recently, [CuPy isn't installed through the cucim extra now](https://github.com/Project-MONAI/MONAI/blob/56f0bd90d969c18c60abc3411c560d678aa0295e/pyproject.toml#L114) since it's not associated exclusively with it. Thanks!

Shizoqua:
@ericspod Thanks for the nudge. I've rebased onto the latest dev so the branch is current. The Coderabbit comments from the earlier review have been addressed in commit 15a636bbd:

- PyTorch/CUDA guidance now states that pip wheels bundle the CUDA runtime and that users do not need to manually download CUDA or set CUDA_PATH.
- CuPy installation is documented through the dedicated monai[cupy] extra, and the cucim extra is described as installing only cuCIM.
- CuPy examples use cupy-cuda13x for CUDA 13.x.
- The editable-install extras are quoted and the --no-build-isolation typo is fixed.

Please let me know if any of the Coderabbit items still need adjustment.

Shizoqua:
Hi @ericspod, I've rebased this onto the latest `upstream/dev`. The Coderabbit items you mentioned are still addressed in the current diff:
- PyTorch wheels bundle the CUDA runtime; no manual CUDA/CUDA_PATH needed
- CuPy is installed via `monai[cupy]`; the `cucim` extra only installs cuCIM
- examples use `cupy-cuda13x` for CUDA 13.x
- editable-install extras are quoted and `--no-build-isolation` is spelled correctly

Please let me know if anything else needs adjusting.

Shizoqua:
Hi @ericspod, this branch is now rebased onto the latest upstream/dev (3dc1fe24a) and the Coderabbit items remain addressed. Please let me know if anything else is needed.

ericspod (approved):
Hi @Shizoqua thanks for this update, looks good to me!

PR #8986, opened 2026-07-09
Make Fuzzy Gdown Argument Version-dependent

ericspod:
### Description

Gdown has dropped the "fuzzy" argument in version 6.0.0. This PR makes the argument dependent on the Gdown version being below this, and removes it where it's not needed.

### Types of changes

### DIFF
--- a/monai/apps/utils.py
+++ b/monai/apps/utils.py
@@ -240,7 +240,7 @@ def download_url(
             if urlparse(url).netloc == "drive.google.com":
                 if not has_gdown:
                     raise RuntimeError("To download files from Google Drive, please install the gdown dependency.")
-                if "fuzzy" not in gdown_kwargs and not min_version(gdown,"6.0.0"):  # "fuzzy" dropped in gdown 6.0.0
+                if "fuzzy" not in gdown_kwargs and not min_version(gdown, "6.0.0"):  # "fuzzy" dropped in gdown 6.0.0
                     gdown_kwargs["fuzzy"] = True  # default to true for flexible url
                 gdown.download(url, f"{tmp_name}", quiet=not progress, **gdown_kwargs)
             elif urlparse(url).netloc == "cloud-api.yandex.net":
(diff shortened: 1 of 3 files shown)


monai/apps/utils.py
```diff
@@ -240,7 +240,7 @@ def download_url(
             if urlparse(url).netloc == "drive.google.com":
                 if not has_gdown:
                     raise RuntimeError("To download files from Google Drive, please install the gdown dependency.")
-                if "fuzzy" not in gdown_kwargs:
+                if "fuzzy" not in gdown_kwargs and not min_version(gdown, "6.0.0"):  # "fuzzy" dropped in gdown 6.0.0
```
garciadias, line 243: This guard only injects fuzzy=True when the key is absent and gdown is old enough to support it. If a caller explicitly passes fuzzy=... (e.g. a hand-authored large_files.yaml, or a direct caller of this public API), it still gets forwarded to gdown.download() unchanged, which will raise TypeError on gdown >= 6.0.0. Not a regression from this diff, but since the goal here is gdown 6.0.0 compatibility, it may be worth stripping/ignoring an explicit fuzzy when it's unsupported, or at least documenting the version cutoff in the gdown_kwargs docstring above. Suggested: if not min_version(gdown, "6.0.0"): gdown_kwargs.setdefault("fuzzy", True), else: gdown_kwargs.pop("fuzzy", None); or a one-line docstring note.

garciadias, line 243: No unit test exercises both branches of this version check (gdown < 6.0.0 vs >= 6.0.0). Understandable given it requires mocking gdown.__version__, but a small test would lock in the fix against future refactors. Optional: a test mocking gdown.__version__ above/below 6.0.0 and asserting whether fuzzy ends up in the kwargs passed to gdown.download.

garciadias (approved):
Small, correct, well-scoped compatibility fix for gdown 6.0.0 dropping the fuzzy argument. Left two optional minor follow-up suggestions inline (an edge case with explicit caller-supplied fuzzy on gdown>=6.0.0, and optional test coverage for the version-conditional branch) — neither is blocking.

monai/apps/utils.py
ericspod: This is just to make it compatible with 6.0, so long as it's not regression I'd rather leave such an exception to prompt users to remove the unused flag.

ericspod: We don't need to support older gdown explicitly and `min_version` is well tested as it is.

PR #8991, opened 2026-07-13
Adding Merge Queue Support

### LINKED ISSUE #8990: Use Merge Queues
Github [merge queues](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) might be a useful thing to use for accelerating our PR merging. Currently we have to wait for one PR to finish its checks before merging the next, we can't automate the process of updating branches for outdated PRs that are approved and have auto-merge enabled.

Updates to the CI actions to permit running in the queue and use of this feature.

A possible alternative is an action, triggered on a merge to `dev`, which updates any approved PR with auto-merge enabled. This would automate the merging process such that reviewers can approve multiple PRs and set auto-merge and not have to monitor them for updating.

ericspod:
Fixes #8990.

### Description

This modifies the actions to permit running in merge queues. This is something which could be used to accelerate PR integration and automate updating PR branches when other PRs merge into `dev`. This should be discussed in the issue before merging.

This also removes the `paths-ignore` component of the main CI script as this was blocking documentation-only PRs.

### Types of changes


ericspod:
> `merge_group` runs in the base-repository context and can receive repository secrets. The workflow checks out the merge-group revision before running `./runtests.sh --min`, so queued pull-request code can read and exfiltrate `NGC_API_KEY`, `NGC_ORG`, and `NGC_TEAM`. Gate the secret-dependent step with `github.event_name != 'merge_group'`, or isolate it in a trusted job.

@coderabbitai how would any untrusted code which exposes these keys get into the merge queue? No code should get to the queue that hasn't been approved in a PR.

.github/workflows/cicd_tests.yml
```diff
@@ -7,19 +7,13 @@ on:
...
   pull_request:
     branches:
       - dev
       - main
       - releasing/*
-    paths-ignore:  # skip if only docs are modified
-      - '**.md'
-      - '**.rst'
-      - 'docs/**'
+  merge_group:  # needed for merge queue
```
garciadias, line 15: This adds the `merge_group` trigger for the GitHub-Actions-hosted checks, but the branch protection ruleset (ID 16034365) also requires third-party checks: DCO, DeepSource (Docker/Python/Shell), and pre-commit.ci. A merge queue entry only completes when every required check reports on the `gh-readonly-queue/...` branch. Have these external apps been verified to support `merge_group` events, or will they be excluded from the merge queue's required checks? If not, enqueued PRs will wait until the queue times out.

.github/workflows/cicd_tests.yml
```diff
@@ -7,19 +7,13 @@ on:
       - dev
       - main
       - releasing/*
-    paths-ignore:  # skip if only docs are modified
-      - '**.md'
-      - '**.rst'
-      - 'docs/**'
   pull_request:
```
garciadias, line 10: Removing `paths-ignore` means every documentation-only PR now triggers the full test matrix (min-dep across OS/Python/PyTorch combinations, full-dep, and packaging). Note that the `merge_group` trigger does not use path filters, so this removal is not required for merge queue support. Could you confirm the CI cost increase is intentional, and whether a lighter-weight always-required job could preserve the doc-skip behavior while still satisfying the required check?

.github/workflows/codeql-analysis.yml
```diff
@@ -17,6 +17,8 @@ on:
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ dev ]
+  merge_group:  # needed for merge queue
```
garciadias, line 20: This workflow is currently disabled on the repository (`disabled_manually`), so the `merge_group` trigger here will not take effect until CodeQL is re-enabled.

.github/workflows/cicd_tests.yml
ericspod: When the doc-only PRs come through the actions don't run at all so don't get counted as passing either. This is to re-enable these until I work on a better solution involving a script to filter things and predicate actions on its output.

.github/workflows/codeql-analysis.yml
ericspod: This is in another PR I'm working on.

.github/workflows/cicd_tests.yml
ericspod: I don't honestly know, we might have to see if it goes through or adjust our branch rules. The DCO, pre-commit, and other checks don't need to be run again in the queue, only code that's gotten through a PR is in the queue, so it would be good to exclude them.

.github/workflows/codeql-analysis.yml
ericspod: This has been addressed now with #9079.

PR #8995, opened 2026-07-16
Add ConvNeXt for 1D, 2D and 3D inputs

### LINKED ISSUE #4798: Add ConvNeXt model architecture
Hi,

Is there interest in adding the ConvNeXt architecture to MONAI? It is a modern version of a ConvNet using current best practices based on this paper: https://arxiv.org/abs/2201.03545.

The code for ConvNeXt are available here: https://github.com/facebookresearch/ConvNeXt

If there is interest I can work on contributing a PR for this architecture.

VenkateswarluNagineni:
Fixes #4798.

### Description

Adds ConvNeXt (`A ConvNet for the 2020s`) as a classification backbone. MONAI has no ConvNeXt today, even though `MedNeXt` derives from it — MedNeXt is a segmentation architecture, so the modern convolutional *backbone* is still missing alongside `resnet`/`densenet`/`senet`/`efficientnet`.

This follows the direction @wyli gave on the issue:

> I think from monai's perspective we can add value to it by supporting both 2d and 3d, and making use of existing monai network layers/blocks whenever possible.

**Supports 1D, 2D and 3D.** The reference implementation is 2D-only; every layer here is built through MONAI's dimension-agnostic factories, so volumetric inputs work with the same code path. The reference applies the pointwise convolutions as `Linear` layers on a permuted channels-last tensor; using 1x1 convolutions instead is mathematically equivalent and keeps the block agnostic to the number of spatial dimensions.

**Reuses existing MONAI layers** rather than re-implementing them:
- `Conv` / `Pool` factories for all dimension-agnostic ops
- `DropPath` (`monai.networks.layers`) for stochastic depth — it is already shape-agnostic, and the `DropPath(p) if p > 0 else nn.Identity()` guard and `torch.linspace` schedule follow `swin_unetr.py`
- `get_act_layer` for the `act` argument
- `trunc_normal_` for the reference's `std=0.02` initialisation

The structure follows `densenet.py` (the `features` / `class_layers` split, `spatial_dims, in_channels, out_channels` ordering, `OrderedDict` module naming, real subclasses for variants, alias block).

### `LayerNormNd`

ConvNeXt normalises over the channel dimension of a channels-first feature map. `torch.nn.LayerNorm` normalises over the *trailing* dimensions, so it expects channels-last and cannot be used directly, and MONAI has no channels-first equivalent (`Norm.LAYER` is a plain `nn.LayerNorm` alias). So this PR adds `LayerNormNd`, which normalises dimension 1 of a `(batch, channel, *spatial)` tensor for any number of spatial dimensions.

It computes the statistics explicitly over `dim=1` instead of permuting to channels-last, which keeps it uniform across ranks and TorchScript-friendly. A test asserts it equals `nn.LayerNorm` applied to the equivalent permuted tensor.

I've kept it in `convnext.py` to keep this PR's surface small — happy to promote it to `monai/networks/layers/` as a public layer if you'd prefer, since nothing else in MONAI currently offers this.

### Types of changes

- [x] New feature (non-breaking change which adds functionality).

Docs *are* updated (`docs/source/networks.rst`, following the `DenseNet` entry pattern — base class with `:members:`, variants without), but I haven't run `make html` locally, and I'm on Windows so I ran the test files directly with `unittest` and the linters individually rather than through `runtests.sh`. Details below — happy to run anything else you'd like to see.

### Testing

New file `tests/networks/nets/test_convnext.py`, 26 tests, all passing (`python -m unittest tests.networks.nets.test_convnext`):

- **Faithfulness to the reference.** The default variants match the parameter counts published by [facebookresearch/ConvNeXt](https://github.com/facebookresearch/ConvNeXt) exactly for ImageNet-1k (Tiny 28,589,128 / Small 50,223,688 / Base 88,591,464; Large 197,767,336 and XLarge 350,196,968 also verified locally). This checks the architecture is correct rather than merely shape-correct.
- **Shapes** for 1D, 2D (non-cubic input) and 3D, over all five variants, plus a case with stochastic depth and layer scale disabled.
- **`LayerNormNd` equivalence** to `nn.LayerNorm` on the permuted tensor, for 2D and 3D (max difference ~3e-07).
- **TorchScript** via `test_script_save` for every case, including both the layer-scale and no-layer-scale paths.
- **Drop path schedule** ascends linearly from 0 to `drop_path_rate` across the whole network.
- **Ill-args** raise `ValueError`.

Also ran `tests/networks/nets/test_densenet.py` and `test_mednext.py` (56 tests) to confirm the `nets/__init__.py` addition doesn't regress anything.

Checks run on Windows with torch 2.13.0+cpu: `ruff check`, `black --check`, `isort --check-only` all clean on the changed files; `mypy monai/networks/nets/convnext.py` reports no errors for this file. (I ran the test files directly with `unittest` rather than the full `./runtests.sh`, which expects a POSIX shell.)

### Notes

- **No `pretrained` support in this PR.** The reference 2D weights could be mapped onto this implementation in a follow-up (the 1x1-conv vs `Linear` pointwise weights differ only by a reshape), but that felt like a separate change; `MedNeXt` likewise ships without pretrained weights.
- Every spatial dimension of the input should be divisible by 32 (patchify stem of 4 plus three downsamplings), which is documented in the class docstring.

### DIFF
--- a/docs/source/networks.rst
+++ b/docs/source/networks.rst
@@ -326,11 +326,6 @@ Layers
 .. automodule:: monai.networks.layers.Norm
   :members:
 
-`LayerNormNd`
-~~~~~~~~~~~~~
-.. autoclass:: LayerNormNd
-  :members:
-
 `Conv`
 ~~~~~~
 .. automodule:: monai.networks.layers.Conv
@@ -347,6 +342,11 @@ Layers
   :members:
 
 .. currentmodule:: monai.networks.layers
+
+`LayerNormNd`
+~~~~~~~~~~~~~
+.. autoclass:: LayerNormNd
+    :members:
 
 `ChannelPad`
 ~~~~~~~~~~~~
--- a/docs/source/networks.rst
+++ b/docs/source/networks.rst
@@ -326,6 +326,11 @@ Layers
 .. automodule:: monai.networks.layers.Norm
   :members:
 
+`LayerNormNd`
+~~~~~~~~~~~~~
+.. autoclass:: LayerNormNd
+  :members:
+
 `Conv`
 ~~~~~~
 .. automodule:: monai.networks.layers.Conv
--- a/docs/source/networks.rst
+++ b/docs/source/networks.rst
@@ -467,6 +467,31 @@ Nets
 .. autoclass:: AHNet
   :members:
 
+`ConvNeXt`
+~~~~~~~~~~
+.. autoclass:: ConvNeXt
+  :members:
+
+`ConvNeXtTiny`
+~~~~~~~~~~~~~~
+.. autoclass:: ConvNeXtTiny
+
+`ConvNeXtSmall`
+~~~~~~~~~~~~~~~
+.. autoclass:: ConvNeXtSmall
+
+`ConvNeXtBase`
+~~~~~~~~~~~~~~
+.. autoclass:: ConvNeXtBase
+
+`ConvNeXtLarge`
+~~~~~~~~~~~~~~~
+.. autoclass:: ConvNeXtLarge
+
+`ConvNeXtXLarge`
+~~~~~~~~~~~~~~~~
+.. autoclass:: ConvNeXtXLarge
+
 `DenseNet`
 ~~~~~~~~~~
 .. autoclass:: DenseNet
--- a/monai/networks/layers/__init__.py
+++ b/monai/networks/layers/__init__.py
@@ -17,6 +17,7 @@ from .drop_path import DropPath
 from .factories import Act, Conv, Dropout, LayerFactory, Norm, Pad, Pool, RelPosEmbedding, split_args
 from .filtering import BilateralFilter, PHLFilter, TrainableBilateralFilter, TrainableJointBilateralFilter
 from .gmm import GaussianMixtureModel
+from .layer_norm_nd import LayerNormNd
 from .simplelayers import (
     LLTM,
     ApplyFilter,
--- a/monai/networks/layers/layer_norm_nd.py
+++ b/monai/networks/layers/layer_norm_nd.py
@@ -1,0 +1,46 @@
+# Copyright (c) MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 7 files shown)


VenkateswarluNagineni:
Thanks for the review. Addressed in the latest commit:

- **Input validation** — added. I confirmed the `kernel_size` case is a real trap: an even kernel doesn't preserve the spatial size at stride 1, so the residual add fails with an opaque `RuntimeError: The size of tensor a (8) must match tensor b (9)` deep in `forward`. It's now rejected in `__init__` with a clear message, along with empty/mismatched/non-positive `depths`/`features` and an out-of-range `drop_path_rate`. All are covered by `test_ill_arg` and documented in the class `Raises`.
- **Alias coverage** — fixed. `TEST_VARIANT_CASES` now mixes the exported lowercase/mixed-case aliases (`convnext_tiny`, `Convnext_base`, …) with the canonical classes, so a broken package export is caught too.

I've left the per-`forward` docstrings out on purpose: the peer classification nets in this package (`densenet.py`, `senet.py`) document behavior at the class level and don't docstring `forward`/private helpers, and the `ConvNeXt`/`_ConvNeXtBlock`/`LayerNormNd` class docstrings already cover the tensor in/out contract. Following the existing house style there rather than adding docstrings the neighbouring modules don't have.

aymuos15:
> happy to promote it to monai/networks/layers/ as a public layer if you'd prefer,

Would definitely do this

aymuos15:
Out of curiosity why not Port the V2 version instead? (Cool contribution!)

VenkateswarluNagineni:
Thanks @aymuos15!

**Public `LayerNormNd`** — done, just pushed. It now lives in `monai/networks/layers/layer_norm_nd.py` (mirroring `drop_path.py`), is exported from `monai.networks.layers`, has a Layers docs entry and its own equivalence-vs-`nn.LayerNorm` test, and `convnext.py` imports it from there — so any channels-first backbone can reuse it instead of it being buried in this net.

**ConvNeXt V2** — good call, and I'd like to add it, but as a focused follow-up rather than swapping it in here:
- #4798 asked for ConvNeXt, and V1 is still the most widely used variant with the broadest set of released weights, so it's the natural first backbone to land.
- This PR pins architectural fidelity by asserting exact parameter-count parity with the official V1 release (Tiny 28,589,128 / Small 50,223,688 / Base 88,591,464).
- V2's changes sit cleanly on exactly this scaffolding: it drops LayerScale and adds a Global Response Normalization (GRN) layer per block (the FCMAE pretraining recipe is orthogonal to the modules). I can add a `GRN` layer + V2 variant presets in a follow-up that reuses this block/stage code directly.

Happy to fold V1 + V2 into this single PR instead if you'd prefer — just say the word.

aymuos15:
>  and V1 is still the most widely used variant with the broadest set of released weights

Wasnt aware pytorch doesnt have the official weights for it. My bad. This LGTM

PR #8997, opened 2026-07-16
Fix 1/N floor in effective-rank collapse score after mean-centering

### LINKED ISSUE #8996: EmbeddingCollapseMetric: effective-rank score has a 1/N floor when N <= D
## Describe the bug

`_effective_rank_score` (`monai/metrics/embedding_collapse.py:298`) mean-centers the embedding matrix, which caps its attainable rank at `min(N-1, D)`, but normalizes the effective rank by `min(N, D)`.

Since the score is `1 - eff_rank/max_rank`, this imposes a hard floor of `1/N` whenever `N <= D` - so the metric structurally cannot report "no collapse" in the regime it was designed for (high-dimensional embeddings, modest sample counts).

`_per_class_rank` is affected more, since it partitions N by class: the smaller per-class N makes the floor larger.

## To Reproduce

Isotropic embeddings - no collapse is present, so the correct score is ~0:

| N | D | score now | expected |
|---|---|---|---|
| 2 | 768 | **0.5000** | 0.0 |
| 4 | 768 | **0.2503** | ~0.0004 |
| 8 | 768 | **0.1262** | ~0.0014 |
| 64 | 768 | **0.0261** | ~0.0106 |
| 1024 | 768 | 0.1234 | 0.1234 (N > D: unaffected) |

Two maximally-spread points score exactly 0.50 ("50% collapsed") when they span the only subspace two samples can span:

```python
import torch
from monai.metrics.embedding_collapse import _effective_rank_score

emb = torch.zeros(2, 768)
emb[0, 0], emb[1, 0] = 1.0, -1.0
print(float(_effective_rank_score(emb)))  # 0.5, expected 0.0
```

The observed floor matches `1 - (N-1)/N = 1/N` exactly.

## Expected behavior

Normalize by `min(N-1, D)`, the attainable maximum after centering. This is a no-op for `N > D`, where both expressions equal `D`.

## Environment

Introduced in #8815 (merged 2026-04-21), shipped in 1.6.0.

I have a fix with tests ready and will open a PR shortly.

One design question: for `N = 2` the centered matrix is always rank-1, so the score is a constant regardless of the data. Is returning `0.0` right there, or would `None` be preferable so it is excluded from `aggregate()`, as `_class_collapse` does for a single class?

rubenG1009:
Fixes #8996

### Description

`_effective_rank_score` mean-centers the embedding matrix (L290), which caps its attainable rank at `min(N-1, D)`, but normalized the effective rank by `min(N, D)` (L298).

Since the score is `1 - eff_rank/max_rank`, that imposed a hard floor of `1/N` whenever `N <= D`: isotropic embeddings at N=4, D=768 scored **0.2503**, and two maximally-spread points scored **exactly 0.50** ("50% collapsed"), where the correct answer is ~0. The metric could not report "no collapse" in precisely the regime it was designed for. `_per_class_rank` is affected more, since per-class N is smaller and the floor is therefore larger.

The observed floor matches `1 - (N-1)/N = 1/N` exactly.

### Why `min(N-1, D)`

Mean-centering forces the rows to sum to zero, which is a linear dependency: if you know N-1 rows, the last one is determined. So `rank(centered) <= min(N-1, D)`. Normalizing by the attainable maximum removes the artifact.

**The fix is a no-op for `N > D`**, where `min(N-1, D)` and `min(N, D)` both equal `D`. Verified: a 1024x768 matrix scores 0.1234 before and after.

### Tests

Three tests added to `TestEffectiveRankScore`, all RED on `dev` and GREEN after the fix (except the third, which is a regression guard and passes both ways):

- `test_effective_rank_no_collapse_when_n_leq_d` - 4x768 isotropic: 0.2503 -> ~0.0004
- `test_effective_rank_two_distinct_samples_report_no_collapse` - 2x768: 0.5 -> 0.0
- `test_effective_rank_unchanged_when_n_gt_d` - 1024x768 unchanged at 0.1234 (proves no regression)

Full suite: 48 passed, 2 skipped (sklearn not installed locally).

### Note on the modified test

`test_formula_matches_manual` mirrored the old normalization inline (`min(20, 16)`). Its fixture is N=20 > D=16, so it passes either way, but it would otherwise encode the old formula. Updated to `min(20 - 1, 16)` - flagging it explicitly rather than changing an assertion silently.

### Types of changes


vikashg (approved):
Mathematically correct fix. Mean-centering introduces one linear dependency (rows sum to 0), so rank(centered) \u2264 min(N-1, D) \u2014 normalizing by min(N, D) was imposing a spurious 1/N floor. The tests clearly demonstrate the bug (0.25 and 0.50 for isotropic inputs) and the fix. No-op for N > D confirmed. LGTM.

vikashg (approved):
Mathematically correct fix. Mean-centering introduces one linear dependency (rows sum to 0), so rank(centered) \u2264 min(N-1, D) \u2014 normalizing by min(N, D) was imposing a spurious 1/N floor. The tests clearly demonstrate the bug (0.25 and 0.50 for isotropic inputs) and the fix. No-op for N > D confirmed. LGTM.

PR #8999, opened 2026-07-17
docs: correct pip extras in installation guide to match setup.cfg

ymxlx:
### Description

The extras list in the installation guide (`docs/source/installation.md`) has drifted from the actual `[options.extras_require]` section of `setup.cfg`:

1. **`onnxruntime` is documented as an extra, but no such extra exists.** `onnxruntime` is a dependency *installed by* the `onnx` extra. As a result, `pip install 'monai[onnxruntime]'` currently succeeds but silently installs nothing (pip only warns that the extra is unknown), which is misleading for users following the docs.
2. **Four real, functional extras are undocumented:** `torchio`, `packaging`, `polygraphy`, and `pyamg`.
3. The "which correspond to" package list was misaligned with the bracket list (40 vs 39 entries, out of step from `transformers` onward).

This PR updates only the documentation so the bracket list and the corresponding package list match `setup.cfg` exactly (42 extras, one prose entry per extra, same order), and notes that the `onnx` extra also installs `onnxruntime`. Verified programmatically: 0 missing, 0 extra, 0 nonexistent after the edit. No changes to `setup.cfg`.

### Types of changes

- [x] Non-breaking change (documentation only)

DCO: commit is signed off (`Signed-off-by: ymxlx`).

### DIFF
--- a/docs/source/installation.md
+++ b/docs/source/installation.md
@@ -286,11 +286,11 @@ Since MONAI v0.2.0, the extras syntax such as `pip install 'monai[nibabel]'` is 
 - The options are
 
 ```
-[nibabel, skimage, scipy, pillow, tensorboard, gdown, ignite, torchvision, itk, tqdm, lmdb, psutil, cucim, openslide, pandas, einops, transformers, mlflow, clearml, matplotlib, tensorboardX, tifffile, imagecodecs, pyyaml, fire, jsonschema, ninja, pynrrd, pydicom, h5py, nni, optuna, onnx, onnxruntime, zarr, lpips, pynvml, huggingface_hub, hyena]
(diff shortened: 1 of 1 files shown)


vikashg (changes requested):
The documentation change is accurate — verified all claims against `setup.cfg`:

- `onnxruntime` correctly removed (not a standalone extra; it's a dependency of the `onnx` extra)
- `torchio`, `packaging`, `polygraphy`, `pyamg` correctly added (all exist in `[options.extras_require]`)
- Bracket list and "which correspond to" package list are now aligned (42 extras, 1:1 mapping)

**Issue blocking merge:** The DCO check is failing because commit `7ddd5ea` is missing a `Signed-off-by` line. Please fix with:

```bash
git commit --amend -s
git push --force-with-lease
```

Once DCO passes, this is good to go.

vikashg:
Minor nit: the `onnx` extra only installs `onnxruntime` on Python ≤ 3.10 (there's a `python_version <= '3.10'` marker in `setup.cfg`). Consider updating the text to:

```
`onnx` (which also installs `onnxruntime` on Python ≤ 3.10)
```

Otherwise LGTM — this correctly syncs all 42 extras with `setup.cfg`. Thanks for catching the drift.

ericspod:
Hi @ymxlx @vikashg I'm working on updating the whole build process with https://github.com/Project-MONAI/MONAI/pull/9010. This will redo this section of the documentation with updated dependency lists, so I think this PR will not be needed. We can leave this open for now and come back to it later to see if you wanted to update anything on top of those changes.

PR #9001, opened 2026-07-18
Raise dedicated HashValidationError on download/extract hash mismatch

### LINKED ISSUE #8832: Raise different exception when hash value of download is wrong

vishnukannaujia:
Fixes #8832.

### Summary

Follow-up to the note left in #8699: a hash-validation failure should raise something specific so it is not silently suppressed by `skip_if_downloading_fails`.

Today `download_url` and `extractall` raise a plain `RuntimeError` on a hash mismatch, with a message containing `"md5 check"`. The test helper `skip_if_downloading_fails` (`tests/test_utils.py`) matches that substring in `DOWNLOAD_FAIL_MSGS` and converts the failure into a **skipped** test — the same treatment as transient network errors. As a result a genuine hash mismatch (data corruption, or the remote content having changed) is quietly skipped instead of surfaced.

### Changes

- Add `HashValidationError(RuntimeError)` in `monai/apps/utils.py` (exported in `__all__`). It is a subclass of `RuntimeError`, so existing `except RuntimeError` handlers keep working and no message text changes.
- Raise `HashValidationError` at all three hash checks: existing file and downloaded file (`download_url`), and compressed file (`extractall`). Docstring `Raises:` entries updated accordingly.
- `skip_if_downloading_fails` now re-raises `HashValidationError` (before the generic `RuntimeError` handling) instead of skipping.
- Remove the now-redundant `"md5 check"` entry from `DOWNLOAD_FAIL_MSGS` — hash failures are represented by the dedicated type, and this entry would otherwise still swallow them.

### Backward compatibility

- `HashValidationError` subclasses `RuntimeError`; callers catching `RuntimeError` are unaffected.
- Message strings are unchanged, so existing assertions like `str(e).startswith("md5 check")` in `test_download_and_extract.py` still hold.

### Tests

New offline tests (no network required) in `tests/apps/test_download_and_extract.py`:

- existing on-disk file with a wrong hash → `HashValidationError`;
- `extractall` with a wrong hash → `HashValidationError`;
- `HashValidationError` is a `RuntimeError` subclass;
- `skip_if_downloading_fails` propagates `HashValidationError` rather than skipping.

### Note for reviewers

Removing `"md5 check"` from `DOWNLOAD_FAIL_MSGS` is a deliberate behavior change: a fully-downloaded file whose hash does not match will now **fail** rather than skip. This is the intent of #8832 (distinguishing corruption from transient/incomplete downloads, which still surface as `"unexpected EOF"`/`"network issue"`). Happy to adjust if maintainers prefer to keep skipping in a specific CI path.

### DIFF
--- a/monai/apps/utils.py
+++ b/monai/apps/utils.py
@@ -42,10 +42,29 @@ if TYPE_CHECKING:
 else:
     tqdm, has_tqdm = optional_import("tqdm", "4.47.0", min_version, "tqdm")
 
-__all__ = ["check_hash", "download_url", "extractall", "download_and_extract", "get_logger", "SUPPORTED_HASH_TYPES"]
+__all__ = [
+    "check_hash",
+    "download_url",
+    "extractall",
+    "download_and_extract",
+    "get_logger",
+    "SUPPORTED_HASH_TYPES",
+    "HashValidationError",
+]
 
 DEFAULT_FMT = "%(asctime)s - %(levelname)s - %(message)s"
 SUPPORTED_HASH_TYPES = {"md5": hashlib.md5, "sha1": hashlib.sha1, "sha256": hashlib.sha256, "sha512": hashlib.sha512}
+
+
+class HashValidationError(RuntimeError):
+    """
+    Raised when a downloaded or existing file does not match its expected hash value.
+
+    This is a dedicated subclass of :class:`RuntimeError` (so existing ``except RuntimeError``
+    handlers keep working) that specifically signals data corruption or a content mismatch,
+    as opposed to a transient/incomplete download. It lets callers distinguish a genuine hash
+    failure from recoverable network errors and avoid silently skipping it.
+    """
 
 
 def get_logger(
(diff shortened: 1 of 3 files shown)


vikashg (approved):
Well-designed: `HashValidationError(RuntimeError)` gives programmatic catching while preserving backward compat with existing `except RuntimeError` handlers. The behavioral change (hash failures surface instead of being skipped) is the correct fix for #8832. Offline tests cover all three raise sites plus the skip_if_downloading_fails propagation. LGTM.

vishnukannaujia:
Thanks for the review @vikashg! All checks are green and CodeRabbit's pre-merge checks pass. Is there anything else needed before this can be merged, or is it good to go?

vishnukannaujia:
Closing this in favor of #8833, which merged first and addresses #8832 the same way — a dedicated hash-mismatch exception (`HashCheckError`) replacing the bare `RuntimeError` across `download_url` and `extractall`. Thanks @e-mny for the fix, and @vikashg for the review here.

One small note for whoever's watching this area: `HashCheckError` subclasses `ValueError`, so code that previously caught these as `RuntimeError` will no longer catch them — worth a mention in the release notes if backward-compat there matters. Happy to send a tiny follow-up if that's a concern.

vikashg:
Great Thank you

PR #9002, opened 2026-07-18
Run train_parallel commands without shell=True in nnUNetV2Runner

### LINKED ISSUE #8992: Avoid shell=True in nnUNetV2Runner.train_parallel while preserving per-device command order
### Is your feature request related to a problem?

`nnUNetV2Runner.train_parallel` currently joins each device's argument-list commands with `;` and starts the resulting string with `subprocess.Popen(..., shell=True)`.

The commands appear to be constructed internally, so I am not reporting this as a confirmed command-injection vulnerability. However, invoking a shell is unnecessary here and increases the attack surface if command construction changes in the future.

### Describe the solution you'd like

Run each device's commands sequentially with `shell=False`, while retaining parallel execution across devices. One possible implementation is:

- one worker per active device;
- commands within a device are executed in their original order;
- different devices run concurrently;
- each command keeps its associated environment;
- empty stages are skipped.

I prepared and locally checked a proof-of-concept patch. The checks cover command order, list arguments, `shell=False`, empty-stage handling, Python compilation, and patch applicability. I have not run MONAI's complete nnUNet integration test suite, so I would appreciate maintainer guidance on the preferred test location and whether this change fits the current design.

### Alternatives considered

Splitting the semicolon-joined string with `shlex.split` was rejected because it changes command semantics. Launching every command at once was also rejected because commands assigned to the same GPU must remain sequential.

### Additional context

This suggestion came from a static-analysis validation exercise against MONAI. The initial rule finding was manually reviewed before filing this request, and this issue is intentionally framed as defense-in-depth rather than a confirmed vulnerability.

vishnukannaujia:
Fixes #8992.

### Summary

`nnUNetV2Runner.train_parallel` joined each device's argument-list commands with `"; "` and launched the resulting string with `subprocess.Popen(cmd_str, shell=True, ...)`. The commands are constructed internally, so this is not a confirmed injection vulnerability, but invoking a shell is unnecessary here and enlarges the attack surface if command construction ever changes.

This PR runs the commands with `shell=False`, preserving the original execution semantics:

- one worker thread per active device;
- commands within a device run **sequentially, in their original order**;
- different devices run **concurrently**;
- each stage still completes (all threads joined) before the next stage starts;
- empty stages/devices are skipped.

The per-device execution is extracted into a small static helper `_run_device_commands`, which runs each `(cmd, env)` pair via `subprocess.run(cmd, env=env, stdout=subprocess.DEVNULL, check=False)`.

### Behavioral notes

- **No shell** is spawned; each command is passed as an argument list.
- **Per-command environment**: each command now runs with its own `env`. Previously the joined shell string used only the *first* command's env for the whole device; since all commands for a device target the same GPU their envs are identical, so this is equivalent but more correct.
- **Failure handling**: `check=False` means a non-zero exit does not stop the remaining commands for that device, matching the previous `"; "`-join behavior (`;` continues regardless of exit status).

### Tests

Added unit tests in `tests/apps/nnunet/test_nnunetv2_runner_command.py` (no `nnunetv2` dependency required; they mock `subprocess.run`):

- `test_runs_in_order_without_shell`: commands run in order, each as an argument **list** (not a joined string), with `shell` unset/falsy and each command's own `env`.
- `test_continues_after_nonzero_exit`: `check=False` is used and all commands are attempted.

Full nnUNet integration training is not exercised here as it requires the `nnunetv2` package and GPU data; happy to add an integration-level check if maintainers prefer a specific location.

### DIFF
--- a/monai/apps/nnunet/nnunetv2_runner.py
+++ b/monai/apps/nnunet/nnunetv2_runner.py
@@ -17,6 +17,7 @@ import os
 import re
 import shlex
 import subprocess
+import threading
 from typing import Any
 
 import monai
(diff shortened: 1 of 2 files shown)


vishnukannaujia:
Friendly ping — this one is green across all checks and follows the same pattern as #9001 (typed `HashValidationError` / robust subprocess handling). Would a maintainer be able to take a look when there's a chance? Happy to adjust anything.

ericspod:
Hi @vishnukannaujia unfortunately this was overtaken with another PR we were working on for security fixes. Please have a look at the current dev to verify we've addressed the problem, if there's anything to change or your tests should be included still please update this PR and we can review. Thanks for the effort!

PR #9003, opened 2026-07-18
Fix ClipIntensityPercentiles metadata accumulation

engmohamedsalah:
### Description

`ClipIntensityPercentiles` stored returned clipping values on the transform instance. Reusing the same transform therefore accumulated values from earlier calls and also mutated the metadata of earlier outputs.

This change keeps the clipping-value accumulator local to each `__call__`. Each result now receives only its own clipping values, and later calls cannot change metadata already returned to a caller.

Regression tests cover repeated channel-wise and non-channel-wise calls.

### Types of changes

### Testing

- `python -m unittest tests.transforms.test_clip_intensity_percentiles tests.transforms.test_clip_intensity_percentilesd` — 96 tests passed
- `python -m ruff check monai/transforms/intensity/array.py tests/transforms/test_clip_intensity_percentiles.py`
- `python -m black --check monai/transforms/intensity/array.py tests/transforms/test_clip_intensity_percentiles.py`


vikashg (approved):
Clean fix for a classic mutable-state-on-instance bug. Moving the clipping_values accumulator from `self` to local scope in `__call__` is the right pattern — prevents cross-call contamination and metadata mutation. Regression tests cover both channel-wise and non-channel-wise paths. All CI green. LGTM.

vikashg (approved):

engmohamedsalah:
@vikashg, thanks for the approval. Can we merge this PR

PR #9005, opened 2026-07-19
fix(transforms): preserve device and dtype in squarepulse

aymuos15:
### Description

`squarepulse` allocated its output via `torch.zeros(t.shape)`, which defaults to a CPU float32 tensor regardless of the input tensor's device or dtype. This caused CUDA callers to silently receive a CPU result, breaking downstream transforms when GPU tensors were passed.

This fix allocates the output on the input tensor's device, and preserves the input's floating-point dtype. Integer inputs still promote to the default float dtype (matching `torch.zeros` behavior).

### Diff

```diff
-    y = torch.zeros(t.shape)
+    y = torch.zeros(t.shape, device=t.device, dtype=t.dtype if t.is_floating_point() else None)
```

### Types of changes


monai/transforms/utils.py
```diff
@@ -2281,7 +2281,7 @@ def squarepulse(sig, duty: float = 0.5):
     w = convert_to_tensor(w)
     t = convert_to_tensor(t)
 
-    y = torch.zeros(t.shape)
+    y = torch.zeros(t.shape, device=t.device, dtype=t.dtype if t.is_floating_point() else None)
```
ericspod:
```suggestion
    y = torch.zeros_like(t, dtype=torch.float32)
```
Wouldn't this work? Fixing the dtype to float32 preserves existing behaviour as well.

PR #8978, opened 2026-07-02
Perf: faster get_largest_connected_component_mask (bincount + LUT gather)

aymuos15:
Fixes #8974 .

### Description

`get_largest_connected_component_mask` (`monai/transforms/utils.py`) ranked component sizes by gathering every non-background voxel through `lib.nonzero(features)` and then built the mask with `lib.isin` over the whole label field. Both allocate large transient arrays. This counts labels with a single full-field `lib.bincount` (zeroing index 0 to drop background), and builds the mask with a boolean lookup-table gather (`keep[features]`) instead of `isin`. Output is bit-identical, verified across 2D and 3D at several foreground fractions and component counts, and the numpy and cupy/cucim paths are covered unchanged.

The `isin` replacement is the dominant win on both time and peak memory; the `bincount` change removes the redundant `nonzero` index arrays and gathered copy on top.

| case | metric | before | after | improvement |
|---|---|---|---|---|
| 3D 192^3, fg 50% | time | 82.0 ms | 8.6 ms | 9.5x |
| 3D 192^3, fg 50% | peak transient mem | 108.0 MB | 7.2 MB | 15x |
| 2D 1024^2, fg 60% | time | 21.5 ms | 1.8 ms | 11.7x |
| 2D 1024^2, fg 60% | peak transient mem | 18.0 MB | 1.4 MB | 12.9x |
| 3D 128^3, fg 5% (48k comps) | time | 10.7 ms | 6.2 ms | 1.7x |

### Types of changes
- [x] Non-breaking change (fix or new feature that would not break existing functionality).

PR #9009, opened 2026-07-23
Perf: single-pass remap_instance_id (unique + bincount + LUT gather)

aymuos15:
### Description

Replaces the per-instance loops in `remap_instance_id` (which scanned the full volume once per instance id — twice with `by_size=True` — for `O(instances * volume)` work) with a single `torch.unique(return_inverse=True)` + `bincount` + LUT-gather pass, computing the same relabeling in `O(volume)`. Same approach as #8978; `by_size` tie-breaking (stable, ascending-id) and the int32 output dtype are preserved exactly.

### Benchmark

Synthetic instance masks (random sparse ids, `by_size=True`), median of repeated runs, `OMP_NUM_THREADS=8`.

**2D**

| Input (H×W, #inst) | CPU speedup | CUDA speedup |
| --- | --- | --- |
| 128×128, 40 | ~4.2× | ~8.2× |
| 256×256, 80 | ~10.8× | ~10.4× |
| 512×512, 150 | ~9.1× | ~17.9× |
| 512×512, 300 | ~18× | ~36× |
| 1024×1024, 150 | ~5.7× | ~16.6× |
| 1024×1024, 500 | ~18.7× | ~54.6× |

**3D**

| Input (D×H×W, #inst) | CPU speedup | CUDA speedup |
| --- | --- | --- |
| 64×64×64, 30 | ~2.4× | ~3.8× |
| 96×96×96, 50 | ~2.8× | ~5.8× |
| 128×128×128, 80 | ~3.6× | ~8.9× |
| 128×128×128, 200 | ~8.6× | ~21.6× |
| 160×160×160, 120 | ~5.8× | ~13× |

Numerics are bit-identical (`torch.equal`) at every size. The old implementation was slowest on GPU (e.g. 2720 ms for 1024×1024 with 500 instances, vs 262 ms on CPU) because the per-instance masked scans serialize; that inversion is gone (49.8 ms).

System: 12th Gen Intel Core i7-12800H (20 threads, `OMP_NUM_THREADS=8`); NVIDIA RTX A1000 Laptop GPU (4 GB); Linux 6.8.0-124-generic x86_64; Python 3.10.12; PyTorch 2.12.1+cu130.

### Types of changes
- [x] The metrics tests passed locally (490 passed, including new parameterized cases for non-contiguous ids, size ties, no-background and empty inputs, and randomized 2D/3D equivalence against the previous implementation), along with `black`/`isort`/`ruff`/`mypy`.


monai/metrics/utils.py
```diff
@@ -419,23 +419,22 @@ def remap_instance_id(pred: torch.Tensor, by_size: bool = False) -> torch.Tensor
...
+    Returns:
+        tensor of the same shape as ``pred`` with ids remapped to ``1..K`` (``torch.int`` dtype),
+        or ``pred`` unchanged (original dtype) when it contains no foreground ids.
     """
-    pred_id: Iterable[Any] = list(pred.unique())
-    # the original implementation has the limitation that if there is no 0 in pred, error will happen
-    pred_id = [i for i in pred_id if i != 0]
-
-    if not pred_id:
+    uniq, inverse = torch.unique(pred, return_inverse=True)
```
ericspod, line 426: If `by_size` is False couldn't you just return `inverse` as the result?

tests/metrics/test_remap_instance_id.py
```diff
@@ -0,0 +1,93 @@
...
+        """Remapping produces the hand-computed contiguous ids, including `by_size` ordering and ties."""
+        result = remap_instance_id(torch.as_tensor(pred, device=_device), by_size=by_size)
+        torch.testing.assert_close(result.cpu(), torch.as_tensor(expected, dtype=torch.int), check_dtype=False)
+
+    @parameterized.expand([["all_background_2d", (4, 4)], ["all_background_3d", (2, 3, 4)], ["empty", (0,)]])
+    def test_passthrough(self, _, shape):
+        """Inputs without foreground ids are returned unchanged, keeping their original dtype."""
+        pred = torch.zeros(shape, dtype=torch.int64, device=_device)
+        result = remap_instance_id(pred, by_size=True)
+        self.assertEqual(result.dtype, pred.dtype)
```
ericspod: I think `assert_close` will check this so this line isn't needed.

ericspod (dismissed):
Hi @aymuos15 super minor things but otherwise looks good.

monai/metrics/utils.py
aymuos15: I guess here inverse is a position in the sort and not a label technically so this fails th no_background tests?

tests/metrics/test_remap_instance_id.py
aymuos15: pushed in recent commit!

PR #9010, opened 2026-07-23
Modernising Build Process

### LINKED ISSUE #8980: Modernise Building Process
MONAI uses requirement files, a cfg file, and a setup.py file for module definition and building. Many actions use different building processes to setup the environment for testing and install MONAI rather than just using `pip install monai[...]`. 

To modernise things, MONAI should be updated with:

- [ ] Package information consolidated in the `pyprojects.toml` file. (#9010)
- [ ] Actions updated to use `pip` for installation only. (#9010)
- [ ] Requirements files removed in favour of different dependencies sections like in the cfg file, this should be accompanied with a script to extract dependencies from `pyprojects.toml` to recreate these if needed. (#9010)
- [ ] Installation instructions updated to match. (#9010)
- [ ] Verify these work for tutorials and bundles.

ericspod:
Fixes #8980.

### Description

This updates the way MONAI is built to be more modern, relying on `pip` for everything with all information consolidated into the `pyproject.toml` file. A script `print_dependencies.py` is provided to reconstruct a requirements file from the toml file when needed, such as installing dependencies before installing MONAI for technical reasons. Highlights:

* Move everything build related into `pyproject.toml`.
* Remove the requirements file and the `setup.cfg` file.
* Add a script to recreate the requirements files if needed. 
* Update the version of Versioneer used.
* Update actions to use the new installation process in a uniform manner.
* Update the Docker files to use the process.
* Update installation docs to reflect these changes and clarify some parts.
* Adds a test in packaging for uv.

### Types of changes


Dockerfile.slim
```diff
@@ -46,18 +46,14 @@ RUN wget -q ${NGC_CLI_URI} && unzip ngccli_linux.zip && chmod u+x ngc-cli/ngc &&
...
 COPY tests ./tests
 COPY monai ./monai
 
-# install full deps
-RUN python -m pip install --no-cache-dir --no-build-isolation -U wheel wheel-stub
-RUN python -m pip install --no-cache-dir --no-build-isolation "torch>=2.8.0,<2.11" -r requirements-dev.txt
-
-# compile ext
-RUN CUDA_HOME=/usr/local/cuda FORCE_CUDA=1 USE_COMPILED=1 BUILD_MONAI=1 python setup.py develop
+RUN BUILD_MONAI=1 FORCE_CUDA=1 pip install --no-cache-dir -e .[all,testing]
```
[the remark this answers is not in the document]
ericspod: @coderabbitai This should use the same PyTorch version for the isolated environment as the install environment as setuptools is in the "dependencies" explicitly. This was the library that causes some of the PyTorch version mismatches. Is there any way of forcing pip to use the same versions of dependencies in the isolated environment as it would install into the target environment without discarding build isolation?

monai/config/print_dependencies.py
```diff
@@ -0,0 +1,64 @@
...
+PROJ_KEY = "project"
+OPTS_KEY = "optional-dependencies"
+DEP_KEY = "dependencies"
+REQ_KEY = "requires"
+TOML_FILE = "pyproject.toml"
+
+
+def parse_dependencies(filename=TOML_FILE, sections=[]):
+    # these imports should be here to avoid attempting to import when MONAI is imported and both packages are missing
+    if sys.version_info.minor >= 11:
```
ericspod, line 47: @coderabbitai What comment do I put here to make isort ignore this whole if-else block?

ericspod:
@coderabbitai review again.

.github/workflows/cicd_tests.yml
```diff
@@ -138,15 +138,14 @@ jobs:
...
-        python -m pip install --upgrade pip wheel
-        python -m pip install --user more-itertools>=8.0
+        python -m pip install --upgrade pip wheel tomli
     - name: Install the minimum dependencies
       run: |
         # min. requirements
         python -m pip install torch==${{ matrix.pytorch-version }}
-        python -m pip install -r requirements-min.txt
+        python monai/config/print_dependencies.py build-system | xargs pip install --no-build-isolation
+        python -m pip install --no-build-isolation .[testing]
```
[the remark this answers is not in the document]
ericspod: How did the previous requirements-min.txt file constrain pandas?

Dockerfile.slim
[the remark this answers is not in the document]
ericspod: @coderabbitai How do I generate the constraint files? I don't know the version of PyTorch that should be installed. That should be resolved by pip when making the build environment and also the target environment. It should resolve the same version here, but in case it doesn't I need to extract the version that it would choose and put that into a constraint file.

pyproject.toml
```diff
@@ -1,12 +1,264 @@
...
+  "typeguard<3",  # https://github.com/microsoft/nni/issues/5457
+  "types-PyYAML",
+  "types-setuptools"
+]
+
+[tool.setuptools]
+license-files = ["LICENSE"]
+
+[tool.setuptools.dynamic]
+version = {attr = "monai.__version__"}
```
garciadias, line 184: Resolving version via attr = "monai.__version__" means setuptools has to import monai (which imports torch at module level) just to read the version number -- for every setup.py invocation, not just full builds. Reproduced this directly: in a clean venv with only setuptools/wheel/packaging installed (exactly what release.yml's packaging and versioning jobs install), running python setup.py build crashes with ModuleNotFoundError: No module named 'torch'. This breaks release.yml's packaging job (line 37, sdist/bdist_wheel) and versioning job (line 109, setup.py build), and cron.yml's cron-gpu job (lines 154-156, which explicitly uninstalls torch right before calling setup.py). All three run on real release/cron triggers.

.github/workflows/cicd_tests.yml
```diff
@@ -219,16 +218,15 @@ jobs:
...
+        python monai/config/print_dependencies.py build-system | xargs pip install --no-build-isolation
+        python -m pip install --no-build-isolation .[all,testing]
         python -m pip list
-        python -m pip install --no-build-isolation -e .  # test no compile installation
       shell: bash
     - name: Run compiled (${{ runner.os }})
       run: |
         python -m pip uninstall -y monai
-        BUILD_MONAI=1 python -m pip install --no-build-isolation -e .  # compile the cpp extensions
+        BUILD_MONAI=1 python -m pip install --no-build-isolation .  # compile the cpp extensions
```
garciadias: This dropped the -e (editable) flag from the compiled-extension install step. Editable installs build C extensions in-place in the local source tree; a regular install doesn't. The next step runs python -m unittest -v from the repo root, which still has the local uncompiled monai/ source dir -- if that shadows the site-packages copy on sys.path, tests could silently exercise the uncompiled fallback path instead of the real compiled extension, with no visible CI failure (monai degrades gracefully when monai._C isn't importable). Could you confirm the compiled path is still being tested, e.g. by asserting monai._C is importable right after this step?

pyproject.toml
```diff
@@ -1,12 +1,264 @@
...
+  "monai.optimizers.*",
+  "monai.losses.*",
+  "monai.inferers.*",
+  "monai.config.*",
+  "monai._extensions.*",
+  "monai.fl.*",
+  "monai.engines.*",
+  "monai.handlers.*",
+  "monai.auto3dseg.*",
+  "monai.bundle.*",
```
garciadias, line 231: The old setup.cfg's mypy override list for disallow_incomplete_defs included monai.metrics.* and monai.apps.* -- both are missing from this migrated block. Looks like an unintentional drop during the setup.cfg -> pyproject.toml migration; silently weakens mypy strictness for those two packages.

garciadias (commented):
A few places outside this PR's diff still reference the requirements*.txt/setup.cfg files this PR deletes, and will break or mislead once this merges: CONTRIBUTING.md (around lines 60, 135, 223-224) still tells contributors to pip install -U -r requirements-dev.txt and links directly to that file; tests/min_tests.py's docstring (around lines 23, 26) still references requirements-min.txt; and .github/workflows/setupapp.yml (around line 168) still does pip install -r requirements-min.txt (workflow_dispatch-only, so it won't break automatically, but will fail the next time someone runs it manually). Worth a repo-wide grep for requirements-dev.txt / requirements-min.txt / requirements.txt / setup.cfg before merging, and a follow-up commit updating CONTRIBUTING.md and setupapp.yml to the pip install -e .[all,testing] / print_dependencies.py pattern used elsewhere in this PR.

garciadias (commented):
monai/config/print_dependencies.py is now load-bearing across nearly every workflow and both Dockerfiles, but has no unit tests. Worth a small test covering its dependency-parsing behavior for "*", "build-system", an explicit extra name, and the no-args case.

garciadias (commented):
Follow-up to the version-resolution comment on pyproject.toml -- a few of the concrete breakage points aren't reachable as inline comments because the affected lines aren't part of this PR's diff (they're pre-existing commands that this PR's change to version resolution breaks indirectly), so flagging them here: (1) setup.py line 160, version=versioneer.get_version() is now dead/conflicting with the pyproject.toml dynamic version -- pick one source of truth. (2) .github/workflows/release.yml line 62, pip install -r requirements-min.txt in the packaging job's Quick test installed step -- that file is deleted by this PR, so this fails on the next tagged release. (3) .github/workflows/cron.yml line 175, cp $root_dir/requirements*.txt "$tmp_dir" -- same issue, and since set -e is on, this aborts the whole cron-gpu job. All three are straightforward once the root-cause version-resolution fix lands, but worth checking off explicitly before merge.

.github/workflows/cicd_tests.yml
ericspod: I'm going to put -e back in and add a test to ensure extensions were compiled.

pyproject.toml
ericspod: I'll add these back in.

pyproject.toml
ericspod: Please check that these are resolved now, there's a few places that haven't been changed in commented jobs which will be refactored away in later updates.

.github/workflows/cicd_tests.yml
```diff
@@ -274,14 +274,15 @@ jobs:
...
-        python -m pip install --no-build-isolation -r requirements-dev.txt
-        python -m pip install -e .
-        # nvsubquadratic runtime imports need only torch + einops + omegaconf; install
-        # the package itself without its core dependency tree (see job comment above).
-        python -m pip install omegaconf
-        python -m pip install --no-deps 'nvsubquadratic>=0.1.1'
+        python -m pip install --upgrade pip wheel tomli pytest
+        # need a specific version of torch for nvsubquadratic
+        python monai/config/print_dependencies.py build-system | \
+          xargs -d '\n' pip install --no-build-isolation torch==2.10.0 torchvision==0.25.0
```
KumoLiu: What's the reason for specify the specific version of torch here instead of using `PYTORCH_VER1`? nvsubquadratic not compatible with other version?

garciadias (approved):

PR #9029, opened 2026-07-28
Adding Temporary Directory Routine

ericspod:
### Description

This adds a temporary directory creation routine to replace the following cell in many tutorial notebooks:

```python
# Set data directory
directory = os.environ.get("MONAI_DATA_DIRECTORY")
if directory is not None:
    os.makedirs(directory, exist_ok=True)
root_dir = tempfile.mkdtemp() if directory is None else directory
print(root_dir)
```

This should be functionally equivalent, but can also add a finaliser function to delete the directory when the notebook closes. The notebooks can then be updated to reduce the amount of code.

### Types of changes

### DIFF
--- a/monai/apps/utils.py
+++ b/monai/apps/utils.py
@@ -459,6 +459,6 @@ def create_temp_dir(directory: str | None = None, delete_on_finalise: bool = Fal
     os.makedirs(directory, exist_ok=True)
 
     if delete_on_finalise:
-        atexit.register(shutil.rmtree, directory)
+        atexit.register(shutil.rmtree, directory, ignore_errors=True)
 
     return directory
--- a/monai/apps/utils.py
+++ b/monai/apps/utils.py
@@ -11,6 +11,7 @@
 
 from __future__ import annotations
 
+import atexit
 import hashlib
 import json
 import logging
@@ -29,7 +30,7 @@ from urllib.parse import urlparse
 from urllib.request import urlopen, urlretrieve
 
 from monai.config.type_definitions import PathLike
-from monai.utils import look_up_option, min_version, optional_import
+from monai.utils import MONAIEnvVars, look_up_option, min_version, optional_import
 
 requests, has_requests = optional_import("requests")
 gdown, has_gdown = optional_import("gdown", "4.7.3")
@@ -42,7 +43,15 @@ if TYPE_CHECKING:
 else:
     tqdm, has_tqdm = optional_import("tqdm", "4.47.0", min_version, "tqdm")
 
-__all__ = ["check_hash", "download_url", "extractall", "download_and_extract", "get_logger", "SUPPORTED_HASH_TYPES"]
+__all__ = [
+    "check_hash",
+    "download_url",
+    "extractall",
+    "download_and_extract",
+    "get_logger",
+    "SUPPORTED_HASH_TYPES",
+    "create_temp_dir",
+]
 
 DEFAULT_FMT = "%(asctime)s - %(levelname)s - %(message)s"
 SUPPORTED_HASH_TYPES = {"md5": hashlib.md5, "sha1": hashlib.sha1, "sha256": hashlib.sha256, "sha512": hashlib.sha512}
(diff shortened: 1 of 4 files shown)


tests/apps/test_create_temp_dir.py
```diff
@@ -0,0 +1,84 @@
...
+        """Test giving a directory to the function, ensuring it creates the directory."""
+        with tempfile.TemporaryDirectory() as temp_dir:
+            selected_dir = f"{temp_dir}/test_inner_dir"
+            test_dir = create_temp_dir(selected_dir)
+
+            self.assertTrue(os.path.isdir(selected_dir))
+            self.assertEqual(test_dir, selected_dir)
+
+    def test_finalisation(self):
+        """Test the temporary directory is deleted by finalisation using a subprocess."""
```
garciadias: This docstring still says "using a subprocess," but the earlier subprocess-based check was replaced with a monkeypatched atexit.register that the test then calls manually (see lines 64-65, 77-78). That proves shutil.rmtree works, but not that create_temp_dir is actually wired up correctly with the real atexit module or that cleanup fires on process exit. Could you fix the docstring to describe what's actually tested, and/or restore a lightweight subprocess smoke test (guarded with skip_if_quick like before) so there's still one real end-to-end check?

tests/apps/test_create_temp_dir.py
```diff
@@ -0,0 +1,84 @@
...
+from unittest.mock import patch
+
+from monai.apps import create_temp_dir
+from monai.utils import MONAIEnvVars
+
+MONAI_DATA_DIRECTORY = "MONAI_DATA_DIRECTORY"
+
+
+class TestCreateTempDir(unittest.TestCase):
+    def test_basic_use(self):
```
garciadias, line 28: test_basic_use exercises the implicit auto-cleanup path (no directory, no MONAI_DATA_DIRECTORY), where delete_on_finalise gets forced to True in monai/apps/utils.py. That's the exact path every notebook migrating to this helper will hit, but this test never asserts atexit.register was called, so a regression here would go unnoticed. Could you patch atexit.register in this test too and assert it gets called when no directory or env var is provided?

monai/apps/utils.py
```diff
@@ -422,3 +431,34 @@ def download_and_extract(
         filename = filepath or Path(tmp_dir, get_filename_from_url(url)).resolve()
         download_url(url=url, filepath=filename, hash_val=hash_val, hash_type=hash_type, progress=progress)
         extractall(filepath=filename, output_dir=output_dir, file_type=file_type, has_base=has_base)
+
+
+def create_temp_dir(directory: str | None = None, delete_on_finalise: bool = False) -> str:
```
garciadias: Every other path-like parameter in this file (check_hash, download_and_extract, etc.) is typed PathLike, but this one is str | None. Minor consistency nit: directory: PathLike | None = None.

garciadias (commented):
create_temp_dir isn't added to docs/source/apps.rst alongside check_hash, download_url, extractall, and download_and_extract, so it won't show up in the built API docs. Consider adding an autofunction entry for it.

garciadias (changes requested):
Two Major findings around test rigor for the new create_temp_dir utility: the finalisation test's docstring is stale (still claims subprocess-based verification, but now just mocks atexit.register and calls it manually), and the more common implicit auto-cleanup path isn't asserted at all. The implementation itself looks correct and low-risk. Details in the inline comments.

tests/apps/test_create_temp_dir.py
ericspod: I've changed the docstring, we don't need to test atexit itself so I feel this test is adequate. The process based test was just really too slow, even when running full tests we need to be paring things down.

tests/apps/test_create_temp_dir.py
ericspod: This is updated now.

monai/apps/utils.py
ericspod: I've updated this now.

PR #8894, opened 2026-06-02
Migrate core MLflow integration from file store to SQLite

Lawson-Darrow:
Closes #8891. Implements Option 2 from the issue discussion.

MLflow 3.13+ raises for the filesystem (file store) tracking backend unless `MLFLOW_ALLOW_FILE_STORE=true` is set. This migrates MONAI core MLflow usage off the file store to the recommended SQLite backend.

- `MLFlowHandler` defaults to `sqlite:///<cwd>/mlruns.db` when no `tracking_uri` is given and `MLFLOW_TRACKING_URI` is unset. For any local SQLite `tracking_uri` it derives an `mlruns` artifact directory next to the database file, mirroring the previous layout. Remote and other backends are left to MLflow to decide.
- Adds an optional `artifact_location` argument to `MLFlowHandler` to override the derived location. Any explicit `tracking_uri` (including a `file://` URI) or env var is passed through unchanged.
- `DEFAULT_MLFLOW_SETTINGS` now uses a SQLite `tracking_uri` under `<output_dir>/mlruns.db`, keeping artifacts under `<output_dir>/mlruns`.
- Adds `monai.utils.path_to_sqlite_uri` to build a cross-platform SQLite URI.
- Migrates the bundle-run and federated-learning tests off file-store tracking URIs and cleans up the derived artifact directory.

The model zoo bundles that pass an explicit file URI (e.g. vista3d) are left for a separate PR as discussed in the issue.

Unit tests cover the SQLite default, explicit `tracking_uri` passthrough, explicit and derived `artifact_location`, remote `tracking_uri` (no local artifact root injected), and a basic run flow.

PR #9032, opened 2026-07-30
fix: address Dependabot alerts for mlflow, transformers, setuptools

garciadias:
## Summary

Closes the open Dependabot alerts for `mlflow`, `transformers`, and `setuptools`, and migrates MONAI's MLflow integration off the deprecated filesystem (file store) backend onto the recommended SQLite backend.

### Fixed

| Package | Change |
|---|---|
| `mlflow` | Bumped floor to `>=3.15.2`, closing the 2.x/early-3.x CVEs and the unauthenticated webhook SSRF (CVE-2026-64849, fixed in 3.15.0). Since mlflow>=3.13 hard-errors on the filesystem tracking backend, `MLFlowHandler` now defaults to a local SQLite backend (`sqlite:///<cwd>/mlruns.db`, artifacts under `<cwd>/mlruns`) and rejects explicit local-path / `file://` tracking URIs with an actionable error. Adds `monai.utils.path_to_sqlite_uri`, an `artifact_location` argument, and SQLite engine disposal on `close()`. |
| `transformers` | Bumped floor to `>=5.5.0`, closing two HIGH severity CVEs. `MultiModal` now builds a real `transformers.BertConfig` with `_attn_implementation="eager"` for transformers>=4.48's attention dispatch. |
| `setuptools` | Bumped the build-system floor to `>=78.1.1`, closing one HIGH severity CVE. The old `<=79.0.1` cap (#8439) is no longer needed since the legacy `fetch_build_eggs` CLI usage is gone from `setup.py`. |

This PR supersedes #8894 (the standalone SQLite migration), which I have proposed closing.

### Test plan

- `tests/handlers/test_handler_mlflow.py` — SQLite default, artifact co-location, file-store rejection, remote URI, and a full run flow.
- `tests/fl/monai_algo/test_fl_monai_algo.py` and `tests/integration/test_integration_bundle_run.py` — updated to SQLite tracking URIs.
- `tests/networks/nets/test_transchex.py` — passes against transformers 4.36-4.40 and 5.5+.
- `tests/utils/misc/test_monai_utils_misc.py::TestPathToSqliteUri` — SQLite URI construction and escaping.

### Please re-verify before merging

The `transformers>=5.5.0` bump: the original `<5.0` cap (#8912) cited `torch.float8_e8m0fnu` missing from the NGC Docker image's PyTorch 2.7 build. This was not reproducible against PyPI `torch>=2.8.0`, but please re-run the Docker/tutorial CI against the current NGC base image.

### DIFF
--- a/monai/handlers/mlflow_handler.py
+++ b/monai/handlers/mlflow_handler.py
@@ -22,7 +22,16 @@ import torch
 from torch.utils.data import Dataset
 
 from monai.apps.utils import get_logger
-from monai.utils import CommonKeys, IgniteInfo, ensure_tuple, flatten_dict, min_version, optional_import
+from monai.utils import (
+    CommonKeys,
+    IgniteInfo,
+    ensure_tuple,
+    flatten_dict,
+    min_version,
+    optional_import,
+    path_to_sqlite_uri,
+    path_to_uri,
+)
 
 Events, _ = optional_import("ignite.engine", IgniteInfo.OPT_IMPORT_VERSION, min_version, "Events")
 mlflow, _ = optional_import("mlflow", descriptor="Please install mlflow before using MLFlowHandler.")
@@ -68,7 +77,14 @@ class MLFlowHandler:
         tracking_uri: connects to a tracking URI. can also set the `MLFLOW_TRACKING_URI` environment
             variable to have MLflow find a URI from there. in both cases, the URI can either be
             an HTTP/HTTPS URI for a remote server, a database connection string, or a local path
-            to log data to a directory. The URI defaults to path `mlruns`.
+            to log data to a directory. When no ``tracking_uri`` is provided and the
+            ``MLFLOW_TRACKING_URI`` environment variable is unset, the handler now
+            defaults to a local SQLite database backend at ``sqlite:///<cwd>/mlruns.db`` with
+            artifacts stored under ``<cwd>/mlruns``. The default was changed from the filesystem
+            (file store) backend because MLflow 3.13+ raises an exception for the file store unless
+            ``MLFLOW_ALLOW_FILE_STORE=true`` is set; SQLite is the backend MLflow recommends and it
+            does not raise. Any explicitly provided ``tracking_uri`` (including a local file path or
+            ``file://`` URI) is passed through unchanged.
             for more details: https://mlflow.org/docs/latest/python_api/mlflow.html#mlflow.set_tracking_uri.
         iteration_log: whether to log data to MLFlow when iteration completed, default to `True`.
             ``iteration_log`` can be also a function and it will be interpreted as an event filter
@@ -113,6 +129,11 @@ class MLFlowHandler:
         optimizer_param_names: parameter names in the optimizer that need to be recorded during running the
             workflow, default to `'lr'`.
         close_on_complete: whether to close the mlflow run in `complete` phase in workflow, default to False.
+        artifact_location: the location to store run artifacts in, passed to MLflow when the experiment is
+            created. When ``None`` and a local SQLite backend is used (from the ``tracking_uri`` argument
+            or the ``MLFLOW_TRACKING_URI`` environment variable), it defaults to an ``mlruns`` directory
+            next to the database file; for other backends ``None`` lets MLflow decide based on the
+            ``tracking_uri``. Has no effect if the experiment already exists.
 
     For more details of MLFlow usage, please refer to: https://mlflow.org/docs/latest/index.html.
 
@@ -141,6 +162,7 @@ class MLFlowHandler:
         artifacts: str | Sequence[Path] | None = None,
         optimizer_param_names: str | Sequence[str] = "lr",
         close_on_complete: bool = False,
+        artifact_location: str | None = None,
     ) -> None:
         self.iteration_log = iteration_log
         self.epoch_log = epoch_log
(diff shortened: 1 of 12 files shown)


ericspod:
I think we still need to resolve https://github.com/Project-MONAI/MONAI/pull/8894 first?

monai/networks/nets/transchex.py
```diff
@@ -219,7 +220,11 @@ def __init__(
 
         """
         super().__init__()
-        self.config = type("obj", (object,), bert_config)
+        self.config = BertConfig(**bert_config)
+        # explicitly select the eager attention path: transformers>=4.48 dispatches attention
+        # implementations via `config._attn_implementation`, which is otherwise left unset since
+        # `bert_config` above does not come from a `from_pretrained` call.
+        self.config._attn_implementation = "eager"
```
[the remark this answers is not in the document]
garciadias: This is a false positive. `MultiModal` subclasses the local `BertPreTrainedModel(nn.Module)` shim defined in this same file (line 33), not `transformers.PreTrainedModel`. The HF base class's requirement to pass `config` through `super().__init__()` does not apply here, and `nn.Module.__init__` accepts no arguments, so the suggested change would raise a `TypeError`. The existing `tests/networks/nets/test_transchex.py` suite (3 tests) passes unchanged against transformers 5.5.0, which confirms the current construction is correct. No change needed.

tests/handlers/test_handler_mlflow.py
ericspod: I'd suggest a top-level use of TemporaryDirectory here, all other temp directories are then rooted in this one which can be safely deleted in tearDown. I'm worried about doing rmtree in places like it is (and like the original was) rather than everything going into one directory object that has it's own cleanup method.

ericspod: This is an obsolete guard now I think.

pyproject.toml
```diff
@@ -71,7 +71,7 @@ all = [
   "lpips==0.1.4",
   "matplotlib>=3.6.3",
   "MetricsReloaded @ git+https://github.com/Project-MONAI/MetricsReloaded@monai-support",
-  "mlflow>=2.12.2,<3.13",
+  "mlflow>=3.14.0",
```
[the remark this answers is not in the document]
ericspod: Current version is 3.15.2, so we can say `>3.15.0` now.

garciadias:
Resolved — this PR now folds in the SQLite migration from #8894 (MLFlowHandler SQLite default, `path_to_sqlite_uri`, bundle defaults, and matching tests), so #8894 is superseded and I have proposed closing it there. Happy to flip the ordering if you would prefer #8894 merge first.

monai/handlers/mlflow_handler.py
```diff
@@ -156,6 +179,41 @@ def __init__(
...
+            self.artifact_location = path_to_uri(db_path.parent / "mlruns")
+        # MLflow 3.13+ refuses the filesystem (file store) tracking backend, and 3.14+ resolves
+        # the store eagerly at client construction, so a local path or ``file://`` URI would raise
+        # an opaque MlflowException. Reject those here with an actionable message instead.
+        if effective_tracking_uri.startswith("file://") or "://" not in effective_tracking_uri:
+            raise ValueError(
+                "MLflow no longer supports the filesystem (file store) tracking backend; got "
+                f"tracking_uri={effective_tracking_uri!r}. Use a SQLite URI "
+                "(sqlite:///<path>/mlruns.db) or a remote tracking URI, or set "
+                "MLFLOW_ALLOW_FILE_STORE=true to opt back in."
```
ericspod: If this environment variable is set, won't this exception still be raised? Either we need to check this or, if we're going to disallow filestore entirely, omit mentioning the environment variable as it won't opt back in anymore.

monai/handlers/mlflow_handler.py
```diff
@@ -156,6 +179,41 @@ def __init__(
...
+        # When no tracking_uri is provided, default to a local SQLite backend instead of the
+        # filesystem (file store) backend. MLflow 3.13+ raises for the file store unless
+        # `MLFLOW_ALLOW_FILE_STORE=true` is set, while SQLite is the recommended backend and does
+        # not raise. Artifacts cannot live inside a database, so by default they are stored under
+        # the `./mlruns` directory (where the previous file store default kept them) via the
+        # experiment `artifact_location`. Any explicitly provided tracking_uri is left unchanged.
+        self.artifact_location = artifact_location
+        # Resolve the effective tracking URI from the argument or the `MLFLOW_TRACKING_URI`
+        # environment variable, so both configure the artifact location the same way.
+        effective_tracking_uri = tracking_uri or os.environ.get("MLFLOW_TRACKING_URI")
```
ericspod:
```suggestion
        effective_tracking_uri = os.environ.get("MLFLOW_TRACKING_URI", tracking_uri)
```
Perhaps the environment variable should have priority? The idea I would think is to override something hard-coded in a piece of code, it has to be in this order to do that.

monai/handlers/mlflow_handler.py
```diff
@@ -156,6 +179,41 @@ def __init__(
...
+        # For a local SQLite backend, keep run artifacts in an `mlruns` directory next to the
+        # database file (mirroring the previous file-store layout) unless the caller set
+        # `artifact_location`. Other backends (e.g. a remote server) are left to MLflow to decide.
+        # Only `tracking_uri` is passed to the client, so an `MLFLOW_TRACKING_URI` env var is
+        # still resolved by MLflow itself.
+        if (
+            self.artifact_location is None
+            and effective_tracking_uri
+            and effective_tracking_uri.startswith("sqlite:///")
+        ):
```
ericspod:
```suggestion
        if self.artifact_location is None and effective_tracking_uri.startswith("sqlite:///"):
```
`effective_tracking_uri` should definitely be not None after the previous if clause.

monai/handlers/mlflow_handler.py
garciadias: Done — the file store is now disallowed outright, so the misleading `MLFLOW_ALLOW_FILE_STORE` mention is removed from the error message.

monai/handlers/mlflow_handler.py
garciadias: Done — `MLFLOW_TRACKING_URI` now takes precedence over the `tracking_uri` argument, and the client receives `None` so MLflow resolves the env var itself. Added a test for the override.

monai/handlers/mlflow_handler.py
garciadias: Done — removed the redundant truthiness check; `effective_tracking_uri` is guaranteed set after the default fallback.

ericspod (approved):

ericspod:
Thanks @garciadias let's merge this one and close or revise #8894.

PR #9038, opened 2026-08-01
Check Environment Script

ericspod:
Fixes # .

### Description

This adds a script to check the environment for installed libraries, if PyTorch works, if MONAI works, and print platform information. This script is meant to be useable without anything being installed other than Python itself, so it can be used to inspect an environment before attempting to install MONAI or produce diagnostic information when some environmental failure occurs. The script can be run remotely with `curl` as it only relies on the standard library. Minor tweaks are made to the CI action to use this script rather than the copy-pasted short Python lines.

### Types of changes


monai/config/check_env.py
```diff
@@ -0,0 +1,181 @@
...
+    Check MONAI by importing it then printing its debug info.
+    """
+    try:
+        import monai
+
+        monai.config.print_debug_info()
+        return True
+    except ImportError:
+        efprint("MONAI not installed")
+        return False
```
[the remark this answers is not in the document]
ericspod: This is intentional, the only caught issues are the ones that would be expected if PyTorch or MONAI were absent.

monai/config/check_env.py
```diff
@@ -0,0 +1,181 @@
...
+    fprint("CPU:", platform.processor(), "Count:", multiprocessing.cpu_count())
+    fprint("Python:", sys.executable, platform.python_implementation(), platform.python_version())
+
+
+def print_environment_vars():
+    """
+    Print all environment variables other than a few known pointless ones.
+    """
+    fprint("Environment:")
+    for k, v in os.environ.items():
```
garciadias, line 84: This prints every environment variable except LS_COLORS/PS1/PS2. Since the module docstring recommends "python check_env.py --env --envvars --monai" as the full-diagnostics command and the script is meant to be run and shared for remote troubleshooting (including the curl-pipe-to-python usage), this risks a user pasting secrets (API keys, tokens, credentials) that happen to be set in their shell into a shared diagnostic dump. Consider redacting values for keys matching common secret patterns (TOKEN, KEY, SECRET, PASSWORD, AUTH, etc.) before printing, or adding a warning when --envvars is used.

monai/config/check_env.py
```diff
@@ -0,0 +1,181 @@
...
+        return False
+
+
+def check_torch_cuda():
+    """
+    Check CUDA capability in PyTorch by moving a tensor to each available device.
+    """
+    import torch
+
+    dcount = torch.cuda.device_count()
```
garciadias: dcount = torch.cuda.device_count() is called before the try block starts. If this call itself raises (e.g. a broken or mismatched CUDA driver -- exactly the kind of environment problem this script exists to diagnose), the script will crash with a raw traceback instead of producing the intended graceful "PyTorch encountered CUDA error" message. Consider moving the device_count() call inside the try block below it.

garciadias (commented):
The PR checklist claims "New tests added to cover the changes," but the new check_env.py script (argument parsing, check_torch, check_torch_cuda, print_environment, print_environment_vars) has no test coverage at all. The only test change updates the pre-existing print_debug_info test to capture output via StringIO. Since this script is also now invoked directly from CI (cicd_tests.yml), consider adding tests/config/test_check_env.py covering at least the CLI flags and the anonymization behavior of fprint.

garciadias (changes requested):
Two Major findings worth addressing before merge: a potential secret-leak via --envvars, and no test coverage for the new check_env.py script despite the checklist claim. Details in inline/PR comments. Everything else looks solid.

monai/config/check_env.py
ericspod: I've added a warning, checking for possible unsafe patterns isn't going to be reliable enough, so the advice is not to use this unless sure.

garciadias (approved):

PR #9042, opened 2026-08-01
tests: remove the temporary directory make_nifti_image creates

### LINKED ISSUE #9041: Test suite leaks temp files: make_nifti_image never removes the directory it creates
### Describe the problem

`make_nifti_image` in `tests/test_utils.py` creates a NIfTI file inside a fresh `tempfile.mkdtemp()` directory when no `dir=` is given, but returns only the file path. Its docstring assigns cleanup to the caller:

> Create a temporary nifti image on the disk and return the image name.
> **User is responsible for deleting the temporary file when done with it.**

11 of the 15 call sites across 9 test files discard the returned path entirely. The other four, in `tests/data/test_nifti_rw.py` and `tests/integration/test_integration_sliding_window.py`, do try to honour the contract, but they `os.remove` the *file* and never remove the directory holding it. Running `tests/data/test_nifti_rw.py` alone leaves **65 empty directories** behind.

So every call site leaks a directory. The helper creates two things and hands back one, and both callers who tried to clean up got it wrong the same way.

`tests/data/test_make_nifti.py:30` additionally calls `tempfile.mkdtemp()` inside a loop at module scope, leaking three more directories at import time, before any test runs.

### To reproduce

```bash
before=$(ls /tmp | wc -l)
python -m pytest tests/data/test_nifti_rw.py tests/data/test_make_nifti.py \
                 tests/transforms/inverse/test_invert.py \
                 tests/transforms/utility/test_splitdimd.py -q
after=$(ls /tmp | wc -l); echo $((after - before))
```

77 entries left behind: a mix of empty directories (callers that removed the file but not the directory) and populated ones (callers that removed nothing).

### Expected behavior

A test run leaves no new entries under the system temp directory.

### Context

Surfaced by https://github.com/Project-MONAI/MONAI/pull/9040#discussion_r3696162437, where fixing it was out of scope.

aymuos15:
Fixes #9041

### Description

`make_nifti_image` in `tests/test_utils.py` creates a `tempfile.mkdtemp()` directory to hold the image it writes, but returns only the file path. Its docstring made the caller responsible for cleanup:

> User is responsible for deleting the temporary file when done with it.

Of the 15 call sites across 9 test files, 11 discard the returned path entirely. The remaining four (`tests/data/test_nifti_rw.py`, `tests/integration/test_integration_sliding_window.py`) do try to honour the contract, but they `os.remove` the *file* and never remove the directory holding it, so they leak an empty directory per call instead of a populated one. Running `tests/data/test_nifti_rw.py` on its own left 65 directories in the system temp directory.

Both callers that tried to clean up got it wrong the same way, because the helper creates two things and hands back one. Rather than editing 15 call sites, this makes the helper own what it created: when `dir` is not supplied, the directory it makes is registered for removal at interpreter exit. A caller-supplied `dir` is untouched; that one still belongs to the caller.

```python
    if dir is None:
        dir = tempfile.mkdtemp()
        atexit.register(shutil.rmtree, dir, ignore_errors=True)
    else:
        os.makedirs(dir, exist_ok=True)
```

`ignore_errors=True` is deliberate: several callers already delete the file inside the directory, and a test is free to delete what it was handed. Without it, an exit-time handler would turn a passing run into a traceback at shutdown over a directory that is already in the desired state. `tempfile.TemporaryDirectory` makes the same call in its own finalizer.

`tests/data/test_make_nifti.py` also built its own temp directory at module scope, inside a loop that ran three times, leaking three directories at import, before any test ran. That is hoisted to a single directory with the same exit-time cleanup.

### Changes

**`tests/test_utils.py`**: `make_nifti_image` registers the directory it creates for removal at exit; docstring updated to state who owns what. Adds `atexit`/`shutil` imports.

**`tests/data/test_make_nifti.py`**: hoists the module-level `mkdtemp()` out of the parameter loop (3 calls → 1) and registers it for cleanup. Adds two tests:
- `test_temp_dir_removed_at_exit`: runs the helper in a subprocess and asserts the directory is gone once that process exits. This is the regression test for the leak; it fails without the change.
- `test_caller_owned_dir_kept`: asserts a caller-supplied `dir` is *not* removed, guarding the other direction.

### Verification

Counting entries left in the system temp directory by `test_nifti_rw.py`, `test_make_nifti.py`, `test_invert.py` and `test_splitdimd.py`:

| | temp entries left behind |
|---|---|
| before | 77 |
| after | 0 |

The leaked directories were a mix of empty ones (callers that removed the file but not the directory) and populated ones (callers that removed nothing). Both are gone.

Tests over every file that calls the helper (`tests/data/test_make_nifti.py`, `tests/data/test_nifti_rw.py`, `tests/data/utils/test_decollate.py`, `tests/transforms/inverse/`, `tests/transforms/test_inverse_collation.py`, `tests/transforms/utility/test_splitdimd.py`):

```
2 failed, 430 passed, 1 skipped in 74.85s
```

The two failures are `test_nifti_rw.py::test_write_2d` and `::test_write_3d`, which fail identically on an unmodified `dev` checkout in this environment (a `TypeError: '>=' not supported between instances of 'NoneType' and 'int'` and a shape mismatch, a nibabel version issue). They are unrelated to this change and are left alone. `tests/integration/test_integration_sliding_window.py` was also checked separately for the subprocess path: it leaves nothing behind, and its one failure likewise predates this change.

```
./runtests.sh --ruff     All checks passed!
./runtests.sh --mypy     Success: no issues found in 1318 source files
black --check            2 files would be left unchanged
isort --check-only       clean
```

### Notes

Cleanup happens at interpreter exit, so directories still accumulate within a single long-running process and a hard-killed process (`SIGKILL`, `os._exit`) still leaks. What changes is that a normally-terminating test run no longer leaves anything behind, and a new caller cannot reintroduce the leak by forgetting to clean up.

### Types of changes


ericspod (approved):
Hi @aymuos15 thanks for the fix!

PR #9047, opened 2026-08-03
Fix clipping boxes with large coordinates

### LINKED ISSUE #9045: clip_boxes_to_image can drop valid boxes for large coordinates
**Describe the bug**
`spatial_crop_boxes` converts `roi_start` and `roi_end` to `torch.int16` before clamping boxes. For coordinates >= 32768, the ROI bounds overflow/wrap, so boxes that are fully inside the requested crop can be clamped to zero width and dropped when `remove_empty=True`.

This is also reachable through the public `clip_boxes_to_image` API for large images.

**To Reproduce**
```python
import torch
from monai.data.box_utils import clip_boxes_to_image, spatial_crop_boxes

boxes = torch.tensor([[41000.0, 5000.0, 45000.0, 15000.0]], dtype=torch.float32)

cropped, keep = spatial_crop_boxes(
    boxes,
    roi_start=[40000, 0],
    roi_end=[50000, 20000],
    remove_empty=True,
)
print(keep.tolist(), cropped.tolist())

clipped, keep = clip_boxes_to_image(boxes, spatial_size=[50000, 50000], remove_empty=True)
print(keep.tolist(), clipped.tolist())
```

Current output:
```text
[False] []
[False] []
```

**Expected behavior**
The box is inside the crop/image and should be kept. For the crop example, the expected cropped box is:

```text
[True] [[1000.0, 5000.0, 5000.0, 15000.0]]
```

For the `clip_boxes_to_image` example, the expected clipped box is unchanged and kept.

**Environment**
```text
MONAI version: 1.6.0rc1+48.g8690ae74
Numpy version: 1.26.4
Pytorch version: 2.10.0+cpu
MONAI rev id: 8690ae74a8a489d31fe1f9ac8ef0bff63165383e
System: Windows-11-10.0.26200-SP0
Python version: 3.12.1
```

**Additional context**
The overflow appears to come from the `.to(torch.int16)` casts on the converted ROI bounds in `monai/data/box_utils.py`. Whole-slide pathology and detection workflows can use image coordinates larger than 32767, so this can silently remove valid detections instead of raising an error.

Rajioba1:
Fixes #9045.

### Description

`spatial_crop_boxes` converted crop ROI bounds to `torch.int16`, which overflows for coordinates above 32767. This could cause valid boxes in large images to be clamped incorrectly and silently removed when `remove_empty=True`.

This PR keeps the ROI bounds in the same tensor dtype as the boxes during clipping and adds a regression test covering both `spatial_crop_boxes` and the public `clip_boxes_to_image` path for large coordinates.

### Types of changes

Local validation:
- `python -m tests.data.test_box_utils`
- `python -m ruff check monai/data/box_utils.py tests/data/test_box_utils.py`

`black` and `isort` were not installed in my local Python environment, so I could not run those checks directly.

### DIFF
--- a/monai/data/box_utils.py
+++ b/monai/data/box_utils.py
@@ -1035,8 +1035,8 @@ def spatial_crop_boxes(
     # convert to float32 since torch.clamp_ does not support float16
     boxes_t = boxes_t.to(dtype=COMPUTE_DTYPE)
 
-    roi_start_t = convert_to_dst_type(src=roi_start, dst=boxes_t, wrap_sequence=True)[0].to(torch.int16)
-    roi_end_t = convert_to_dst_type(src=roi_end, dst=boxes_t, wrap_sequence=True)[0].to(torch.int16)
+    roi_start_t = convert_to_dst_type(src=roi_start, dst=boxes_t, wrap_sequence=True)[0]
+    roi_end_t = convert_to_dst_type(src=roi_end, dst=boxes_t, wrap_sequence=True)[0]
     roi_end_t = torch.maximum(roi_end_t, roi_start_t)
 
     # makes sure the bounding boxes are within the patch
--- a/tests/data/test_box_utils.py
+++ b/tests/data/test_box_utils.py
@@ -271,6 +271,7 @@ class TestBoxUtilsDtype(unittest.TestCase):
         self.assertGreater(iou[0, 0], 0.0, "IoU should not be truncated to 0")
 
     def test_large_coordinates_are_not_dropped(self):
+        """Verify large-coordinate boxes are preserved by cropping and clipping."""
         boxes = torch.tensor([[41000.0, 5000.0, 45000.0, 15000.0]], dtype=torch.float32)
 
         cropped_boxes, keep = spatial_crop_boxes(
--- a/tests/data/test_box_utils.py
+++ b/tests/data/test_box_utils.py
@@ -35,6 +35,7 @@ from monai.data.box_utils import (
     convert_box_mode,
     convert_box_to_standard_mode,
     non_max_suppression,
+    spatial_crop_boxes,
 )
 from monai.utils.type_conversion import convert_data_type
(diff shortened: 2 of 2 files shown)


vikashg:
## Review

**Scope:** 2 files, +17/-2 lines. Clean, minimal fix.

### The Bug

`spatial_crop_boxes` casts ROI bounds to `torch.int16`, which overflows for coordinates > 32767:

```python
>>> torch.tensor([41000.0]).to(torch.int16)
tensor([-24536], dtype=torch.int16)  # silent overflow!
```

This causes valid boxes in large images (e.g., whole-slide pathology images) to be clipped to garbage values and silently dropped when `remove_empty=True`.

### The Fix

Removes the `.to(torch.int16)` cast, keeping `roi_start_t` and `roi_end_t` in the same dtype as `boxes_t` (which is already `float32` per `COMPUTE_DTYPE`). This is correct — `float32` handles coordinates up to ~3.4e38 with no overflow, and the subsequent `clamp`/arithmetic operations are all float-compatible.

### Verification

I confirmed:
- On dev: `spatial_crop_boxes(boxes=[[41000, 5000, 45000, 15000]], roi_start=[40000, 0], roi_end=[50000, 20000])` drops the box (`keep=False`) ✅ bug reproduced
- With fix: box correctly yields `[[1000, 5000, 5000, 15000]]` with `keep=True` ✅
- `clip_boxes_to_image` with large coords also works ✅
- Small coordinates regression pass ✅
- Partially overlapping large boxes clipped correctly ✅
- Boxes completely outside ROI still dropped ✅
- Full test suite (60 tests) passes ✅

### Notes

- DCO sign-off present on both commits ✅
- PR is MERGEABLE ✅
- Same author as #9048; both are well-scoped, independent fixes

### Verdict

**Approve.** This is a clear, well-motivated bugfix with appropriate test coverage. No concerns.

vikashg (approved):
Looks good to me.

PR #8911, opened 2026-06-10
fix(MetaTensor): astype with torch dtype now returns MetaTensor preserving metadata

AlexanderSanin:
## Summary

- `MetaTensor.astype()` called with a torch dtype (e.g. `torch.int32`, `torch.float16`) was silently returning a plain `torch.Tensor`, discarding all metadata (affine matrix, spacing, applied_operations, and any custom keys).
- Root cause: `out_type` was hardcoded to `torch.Tensor` instead of `type(self)` (`MetaTensor`), so `convert_data_type` set `track_meta=False` and stripped the metadata.
- Fix: use `out_type = type(self)` when `mod_str == "torch"`, so `convert_data_type` receives `output_type=MetaTensor`, sets `track_meta=True`, and the dtype cast is performed while preserving all metadata.
- The `auto3dseg/analyzer.py` module already annotated `label_tensor.astype(torch.int16)` as returning a `MetaTensor` (line 493), relying on this contract.

Closes #8202

## Test plan

- [ ] Existing `test_astype` test updated to assert `isinstance(result, MetaTensor)` and that metadata keys survive the cast.
- [ ] All 96 `tests/data/meta_tensor/` tests pass locally (0 failures).
- [ ] Manual verification:

```python
import torch
from monai.data import MetaTensor

t = MetaTensor(torch.tensor([1., 2., 3.]), meta={"fname": "scan.nii"})
result = t.astype(torch.int32)
assert isinstance(result, MetaTensor)          # was torch.Tensor before
assert result.meta["fname"] == "scan.nii"      # metadata preserved
assert result.dtype == torch.int32             # dtype correctly cast
```

PR #9048, opened 2026-08-03
Fix RandGridPatch whole-dimension patch sizes

### LINKED ISSUE #9046: RandGridPatch crashes for documented 0 or None patch_size entries
**Describe the bug**
`RandGridPatch` documents that `patch_size` entries of `0` or `None` select the whole dimension, matching `GridPatch`. However, when `max_offset` is not supplied, `RandGridPatch.__call__` computes `s % p` for each spatial dimension. This crashes for `p == 0` and `p is None` before patch generation starts.

The dictionary wrapper `RandGridPatchd` is affected through the same array transform.

**To Reproduce**
```python
import torch
from monai.transforms import GridPatch, RandGridPatch

img = torch.arange(1 * 8 * 8 * 8, dtype=torch.float32).reshape(1, 8, 8, 8)

for patch_size in [(0, 4, 4), (None, 4, 4)]:
    print("GridPatch", patch_size, GridPatch(patch_size=patch_size)(img).shape)
    print("RandGridPatch", patch_size, RandGridPatch(patch_size=patch_size)(img).shape)
```

Current behavior:
```text
GridPatch (0, 4, 4) torch.Size([4, 1, 8, 4, 4])
RandGridPatch (0, 4, 4) ZeroDivisionError: integer modulo by zero
GridPatch (None, 4, 4) torch.Size([4, 1, 8, 4, 4])
RandGridPatch (None, 4, 4) TypeError: unsupported operand type(s) for %: 'int' and 'NoneType'
```

**Expected behavior**
`RandGridPatch` should honor the documented `0` / `None` whole-dimension behavior and return patches like `GridPatch` for these inputs.

**Environment**
```text
MONAI version: 1.6.0rc1+48.g8690ae74
Numpy version: 1.26.4
Pytorch version: 2.10.0+cpu
MONAI rev id: 8690ae74a8a489d31fe1f9ac8ef0bff63165383e
System: Windows-11-10.0.26200-SP0
Python version: 3.12.1
```

**Additional context**
The likely failing expression is in `monai/transforms/spatial/array.py`, where default `max_offset` is computed as `tuple(s % p for s, p in zip(array.shape[1:], self.patch_size))`.

Rajioba1:
Fixes #9046.

### Description

`RandGridPatch` documents that `patch_size` entries of `0` or `None` select the whole dimension. When `max_offset` is omitted, the transform currently computes the default offset range with `s % p`, which raises for `0` and `None` patch dimensions.

This PR treats whole-dimension patch entries as having zero random offset for that dimension, matching the documented behavior and `GridPatch` behavior. It adds regression coverage for both `RandGridPatch` and `RandGridPatchd` with `patch_size=(0, 2)` and `patch_size=(None, 2)`.

### Types of changes

Local validation:
- `python -m tests.transforms.spatial.test_rand_grid_patch`
- `python -m tests.transforms.spatial.test_rand_grid_patchd`
- `git diff --check`

Notes:
- `black` and `isort` were not installed in my local Python environment, so I could not run those checks directly.
- `python -m ruff check monai/transforms/spatial/array.py tests/transforms/spatial/test_rand_grid_patch.py tests/transforms/spatial/test_rand_grid_patchd.py` reports an unrelated pre-existing `UP038` warning in `array.py` outside this PR's changed lines.

### DIFF
--- a/monai/transforms/spatial/array.py
+++ b/monai/transforms/spatial/array.py
@@ -3567,7 +3567,7 @@ class RandGridPatch(GridPatch, RandomizableTransform, MultiSampleTrait):
         else:
             min_offset = ensure_tuple_rep(self.min_offset, len(self.patch_size))
         if self.max_offset is None:
-            max_offset = tuple(s % p for s, p in zip(array.shape[1:], self.patch_size))
+            max_offset = tuple(s % p if p else 0 for s, p in zip(array.shape[1:], self.patch_size))
         else:
             max_offset = ensure_tuple_rep(self.max_offset, len(self.patch_size))
 
--- a/tests/transforms/spatial/test_rand_grid_patch.py
+++ b/tests/transforms/spatial/test_rand_grid_patch.py
@@ -20,7 +20,7 @@ from parameterized import parameterized
 from monai.data import MetaTensor, set_track_meta
 from monai.transforms.spatial.array import RandGridPatch
 from monai.utils import set_determinism
-from tests.test_utils import TEST_NDARRAYS, assert_allclose
+from tests.test_utils import TEST_NDARRAYS, TEST_NDARRAYS_NO_META_TENSOR, assert_allclose
 
 A = np.arange(16).repeat(3).reshape(4, 4, 3).transpose(2, 0, 1)
 A11 = A[:, :2, :2]
@@ -67,6 +67,7 @@ TEST_CASE_13 = [
     [A[:, 1:3, 1:3]],
 ]
 TEST_CASE_14 = [{"patch_size": (0, 2)}, A, [A[:, :, :2], A[:, :, 2:]]]
+TEST_CASE_15 = [{"patch_size": (None, 2)}, A, [A[:, :, :2], A[:, :, 2:]]]
 
 TEST_CASE_META_0 = [
     {"patch_size": (2, 2)},
@@ -104,6 +105,8 @@ for p in TEST_NDARRAYS:
     TEST_SINGLE.append([p, *TEST_CASE_12])
     TEST_SINGLE.append([p, *TEST_CASE_13])
     TEST_SINGLE.append([p, *TEST_CASE_14])
+for p in TEST_NDARRAYS_NO_META_TENSOR:
+    TEST_SINGLE.append([p, *TEST_CASE_15])
 
 
 class TestRandGridPatch(unittest.TestCase):
--- a/tests/transforms/spatial/test_rand_grid_patch.py
+++ b/tests/transforms/spatial/test_rand_grid_patch.py
@@ -66,6 +66,8 @@ TEST_CASE_13 = [
     A,
     [A[:, 1:3, 1:3]],
 ]
+TEST_CASE_14 = [{"patch_size": (0, 2)}, A, [A[:, :, :2], A[:, :, 2:]]]
+TEST_CASE_15 = [{"patch_size": (None, 2)}, A, [A[:, :, :2], A[:, :, 2:]]]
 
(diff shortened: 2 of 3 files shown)


vikashg:
## Review

**Effective diff against current `dev`:** 1 source line + 2 test lines (the MetaTensor.astype and write_metrics_reports changes were already merged via #8911 and #8902).

### The RandGridPatch fix

The one-line change is correct and minimal:

```python
# Before (crashes for p=0 or p=None):
max_offset = tuple(s % p for s, p in zip(array.shape[1:], self.patch_size))

# After:
max_offset = tuple(s % p if p else 0 for s, p in zip(array.shape[1:], self.patch_size))
```

I confirmed:
- `ZeroDivisionError` on current dev with `patch_size=(0, 2)` ✅
- Fix resolves it for both `0` and `None` (whole-dimension) entries ✅
- Normal `patch_size=(2, 2)` still works (regression pass) ✅
- Edge case `patch_size=(0, 0)` returns full image as single patch ✅
- All 62 tests in `test_rand_grid_patch.py` pass ✅

### Suggestions

1. **Add a `patch_size=(None, 2)` test case** alongside the `(0, 2)` case. The docs say both `0` and `None` select the whole dimension — covering both in tests makes the contract explicit.

2. (Nit) The PR description mentions this fixes three things, but #8911 and #8902 are already in dev. Consider rebasing so the PR history only shows the RandGridPatch commits.

### Verdict

**Approve** — clean, minimal, well-tested fix for a documented bug. The nit about adding a `None` test case is optional.

Rajioba1:
Thanks for the thorough review, @vikashg — really appreciate it.

Good call on covering `None` explicitly. I initially had a `patch_size=(None, 2)` case, but removed it because it exposed a separate metadata issue rather than just exercising this fix.

Unlike `0`, `None` makes `np.array(self.patch_size)` object dtype:

```python
np.array((None, 2)).dtype  # dtype("O")
```

That object-dtype `spatial_shape` flows into the patch metadata, and through the `MetaTensor` test path it can produce object-dtype metadata that breaks downstream collation in the shared `min-dep` test run.

I see two clean options:

1. Add a `patch_size=(None, 2)` test scoped to non-`MetaTensor` inputs in this PR, so the documented behavior is covered without expanding scope.
2. Normalize `None`/`0` patch-size entries to the actual dimension size before writing `spatial_shape`, so the `MetaTensor` case is also safe. That's stronger, but slightly broader and may be better as a follow-up.

Happy to do either — which would you prefer?

vikashg:
@Rajioba1 I think option 1 will be a cleaner merge. Please make the change and tag me so I can approve and merge

Rajioba1:
@vikashg Added option 1 in c0001781: `patch_size=(None, 2)` is now covered for both `RandGridPatch` and `RandGridPatchd`, scoped to `TEST_NDARRAYS_NO_META_TENSOR` so it exercises the documented `None` behavior without entering the MetaTensor metadata path.

Validation:
- `python -m pytest tests/transforms/spatial/test_rand_grid_patch.py tests/transforms/spatial/test_rand_grid_patchd.py -q` -> 96 passed
- `python -m pytest tests/transforms/spatial/test_rand_grid_patchd.py tests/inferers/test_sliding_window_inference.py::TestSlidingWindowInference::test_default_device_1 tests/inferers/test_sliding_window_inference.py::TestSlidingWindowInferenceCond::test_default_device_1 -q` -> 49 passed
- `python -m ruff check tests/transforms/spatial/test_rand_grid_patch.py tests/transforms/spatial/test_rand_grid_patchd.py` -> passed

Issue #7694, opened 2024-04-22
Error in test_integration_nnunetv2_runner


KumoLiu:
```
[2024-04-20T22:34:15.575Z] 2024-04-20 22:34:15.017497: do_dummy_2d_data_aug: False
[2024-04-20T22:34:15.575Z] 2024-04-20 22:34:15.017690: Creating new 5-fold cross-validation split...
[2024-04-20T22:34:15.575Z] 2024-04-20 22:34:15.018173: Desired fold for training: 0
[2024-04-20T22:34:15.575Z] 2024-04-20 22:34:15.018205: This split has 8 training and 2 validation cases.
[2024-04-20T22:34:15.575Z] using pin_memory on device 0
[2024-04-20T22:34:15.830Z] using pin_memory on device 0
[2024-04-20T22:34:15.830Z] 2024-04-20 22:34:15.437163: Using torch.compile...
[2024-04-20T22:34:16.389Z] /usr/local/lib/python3.10/dist-packages/torch/optim/lr_scheduler.py:28: UserWarning: The verbose parameter is deprecated. Please use get_last_lr() to access the learning rate.
[2024-04-20T22:34:16.390Z]   warnings.warn("The verbose parameter is deprecated. Please use get_last_lr() "
[2024-04-20T22:34:16.390Z] 
[2024-04-20T22:34:16.390Z] This is the configuration used by this training:
[2024-04-20T22:34:16.390Z] Configuration name: 3d_fullres
[...]

PR #9054, opened 2026-08-10
Report a missing nnunetv2 as an optional dependency in nnUNetV2Runner

Kirscher:
Found while investigating #7694.

### Description

`nnUNetV2Runner.__init__` imports `nnunetv2.configuration` directly:

https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/nnunet/nnunetv2_runner.py#L220

so constructing the runner without `nnunetv2` installed raises a bare `ModuleNotFoundError`, rather than the `OptionalImportError` MONAI raises for its other optional dependencies. Every other third-party import in that module goes through `optional_import`.

It is also actively misleading. The dataset lookup immediately above is wrapped in a broad `except Exception`, so it swallows that same `ImportError` first and logs:

```
WARNING - Dataset with name/ID: 123 cannot be found in the record. Please ignore the
message above if you are running the pipeline from a fresh start. But if the dataset is
expected to be found, please check your input_config.
```

A user whose only problem is that `nnunetv2` is not installed is told to check their configuration, and only then gets an unrelated `ModuleNotFoundError`.

This decorates the class with `@require_pkg(pkg_name="nnunetv2")`, the same pattern used for `ITKReader`, `NibabelReader`, `PILReader` and others. Construction now fails immediately with an `OptionalImportError` naming the package and pointing at the installation docs, and the misleading dataset warning is no longer emitted. Behaviour with `nnunetv2` installed is unchanged.

Two regression tests are added in `tests/apps/nnunet/test_nnunetv2_runner_optional_import.py`, one per symptom (wrong exception type, misleading warning). Both fail on `dev` and pass with the change. They simulate the missing package via `mock.patch`, so they run whether or not `nnunetv2` is installed — which matters here, since `nnunetv2` is not currently a declared test dependency.

### Types of changes


ericspod (approved):
Hi @Kirscher thanks for this fix, looks good to me.

PR #9057, opened 2026-08-11
Fix GHSA-873f-pvrv-4x83: warn before executing a bundle's config in load()/run()

garciadias:
## Summary
Fixes GHSA-873f-pvrv-4x83: https://github.com/Project-MONAI/MONAI/security/advisories/GHSA-873f-pvrv-4x83

`monai.bundle.load()`, with its default `model=None`, builds a bundle's network by parsing the bundle's own config through `create_workflow()`. That parsing resolves any `"_target_"` value to an importable callable with no allow list, and passes any `"$"`-prefixed value to Python `eval()`. `monai.bundle.run()` reaches the same code path via a caller-supplied `config_file`. Either way, this means loading or running a bundle whose config you haven't reviewed can execute arbitrary code.

### Design
An earlier version of this fix added an opt-in `trust_remote_code` flag to `load()`. Per review discussion, that was dropped: MONAI has no mechanism to actually establish whether a bundle is trustworthy (unlike, say, a per-repo "has custom code" check), so a flag like that mostly teaches people to set it once and forget about it, without giving them a real basis to decide.

Instead:
- `create_workflow()` — the shared path both `load()` and `run()` use to parse a config file — now raises a `UserWarning` immediately before doing so, spelling out exactly what `"_target_"`/`"$"`-expression content can do and linking this advisory.
- No behavior is blocked. Default behavior is unchanged other than the added warning: `load()`/`run()` still parse and execute the config exactly as before.
- The warning applies uniformly to every caller of `create_workflow()`, not just `load()`.

### Changes
- `monai/bundle/scripts.py`: warning added in `create_workflow()`; docstrings on `load()`, `run()`, and `create_workflow()` updated to describe the risk and point at the advisory.
- `tests/bundle/test_bundle_download.py`: `TestLoadWarnsOnConfigExecution` — default `load()` warns and still executes the config (no flag needed), explicit `model=` still skips config parsing entirely (and warns about nothing), and `run()` warns via the same `create_workflow()` path.

## Test plan
- [x] `python3 -m unittest tests.bundle.test_bundle_download.TestLoadWarnsOnConfigExecution -v`
- [x] Full `tests/bundle/test_bundle_download.py`, `tests/bundle/test_config_parser.py` — no new failures vs. `dev` (remaining failures are pre-existing environment gaps: missing `requests`/`nibabel`, one `pdb`/`bdb` quirk)


ericspod (dismissed):
Looks good to me, the Coderabbit comments look worthwhile however. I can rereview if you want to address

tests/bundle/test_bundle_download.py
```diff
@@ -488,5 +489,65 @@ def test_ngc_download_bundle(self, bundle_name, version, remove_prefix, download
...
+    def _stage_malicious_bundle(self, tempdir: str, marker: str) -> str:
+        name = "evil_bundle"
+        bundle_root = os.path.join(tempdir, name)
+        os.makedirs(os.path.join(bundle_root, "configs"))
+        os.makedirs(os.path.join(bundle_root, "models"))
+        torch.save({"state_dict": {}}, os.path.join(bundle_root, "models", "model.pt"))
+        # `marker` is embedded via `!r` (not raw-interpolated) since this string is itself later
+        # evaluated as Python source -- on Windows, a raw path's backslashes would otherwise be
+        # misparsed as escape sequences.
+        payload = f"$__import__('os').system({('echo pwned > ' + marker)!r})"
```
ericspod:
```suggestion
        payload = f"$__import__('os').system(\"echo pwned > '{marker}'\")"
```
I think the comment from Coderabbit was just about how a command string is put together. The issue is if `marker` contains spaces which will cause the path to be split by the shell when executed. Your solution appears vulnerable to that still. Consider `marker="/tmp/foo bar"`, what would `payload` be with that value? 

I think bash, cmd, and powershell will work with single quotes around the path, and I'd rather add them explicitly like this rather than rely on repr with `!r`. `shlex.join` doesn't work properly here either since it wants to quote the `>` operator.

tests/bundle/test_bundle_download.py
```diff
@@ -488,5 +489,65 @@ def test_ngc_download_bundle(self, bundle_name, version, remove_prefix, download
...
+                warnings.simplefilter("error", UserWarning)
+                load(name=name, model=model, bundle_dir=tempdir, source="github", repo="attacker/repo")
+            self.assertFalse(os.path.exists(marker))
+
+    def test_run_warns_on_config_execution(self):
+        with tempfile.TemporaryDirectory() as tempdir:
+            marker = os.path.join(tempdir, "PWNED")
+            config_file = os.path.join(tempdir, "train.json")
+            with open(config_file, "w") as f:
+                payload = f"$__import__('os').system({('echo pwned > ' + marker)!r})"
```
ericspod: Same. Perhaps just use `_stage_malicious_bundle` with `config_file = os.path.join(tempdir, "evil_bundle", "configs", "train.json")`?

tests/bundle/test_bundle_download.py
```diff
@@ -488,5 +489,65 @@ def test_ngc_download_bundle(self, bundle_name, version, remove_prefix, download
...
+    def test_default_warns_and_executes_config(self):
+        with tempfile.TemporaryDirectory() as tempdir:
+            marker = os.path.join(tempdir, "PWNED")
+            name = self._stage_malicious_bundle(tempdir, marker)
+            with self.assertWarnsRegex(UserWarning, r"GHSA-873f-pvrv-4x83"):
+                with self.assertRaises(AttributeError):
+                    # the malicious config is missing metadata.json and returns a plain `int` for
+                    # `network_def`, so the workflow construction fails after the payload has already
+                    # run -- this mirrors the advisory's own PoC, where the failure happens *after* RCE.
+                    load(name=name, bundle_dir=tempdir, source="github", repo="attacker/repo")
```
ericspod: The other values of `source` are "ngc", "monaihosting", "github", or "huggingface_hub". Is it worth testing with these as well using `parameterized` and appropriate values for `repo`, or are we confident these are independent factors? I think we're good here but it's worth mentioning.

ericspod (changes requested):
I think the payload construction needs to be looked at but otherwise it's good by me.

tests/bundle/test_bundle_download.py
garciadias: Fixed in a27a4ca9 — `marker` is now wrapped in single quotes inside the double-quoted shell command: `f"$__import__('os').system(\"echo pwned > '{marker}'\")"`. With `marker="/tmp/foo bar"` this now produces `echo pwned > '/tmp/foo bar'`, which bash/cmd/powershell all parse as a single path rather than splitting on the space.

tests/bundle/test_bundle_download.py
garciadias: Fixed in 83e64b32 — `test_run_warns_on_config_execution` now calls `self._stage_malicious_bundle(tempdir, marker)` and derives `config_file = os.path.join(tempdir, name, "configs", "train.json")` from it, removing the duplicated payload-construction logic.

tests/bundle/test_bundle_download.py
garciadias: Parameterized in 83e64b32 — `test_default_warns_and_executes_config` now runs over `TEST_CASE_SOURCE_GITHUB`, `TEST_CASE_SOURCE_MONAIHOSTING`, `TEST_CASE_SOURCE_NGC`, and `TEST_CASE_SOURCE_HUGGINGFACE_HUB` via `@parameterized.expand`, confirming the warning fires the same way regardless of `source`/`repo` (they only steer where `download()` would fetch from, which is never reached here since the bundle is already staged on disk).

ericspod (approved):

Issue #9058, opened 2026-08-12
CI Updates 2026


ericspod:
The CI process could benefit from updating. Some on-going work is moving this forward that should be tracked:

- [x] Replace pytype and mypy (#8868) (remember to update the ruleset to allow the PR to pass and add the pyrefly check)
- [x] Modernise building (#8980 #9010)
- [ ] Skip CI Tests Based on Checks, Not Action File (#8971)
- [ ] Add support for merge queues (#8991)
- [x] Move black and isort formatting into pre-commit (#9061) Regular pre-commit may just be fine. ~~using Pre-Commit Lite so that a action can be used which constructs the same environment the other actions run under, this ensures the versions of black and isort match what would be used locally.~~
- [x] Fix weekly-preview #9075 #9081 
- [x] Fix CodeQL #9079
- [ ] Re-instate coverage testing
- [ ] Update actions for uploading to PyPI, docker updating, and anything else that hasn't worked for a while, this will likely need custom runners as Docker building is too large.
- [ ] Update or remove action scripts that have become outdated, especially cron ones.
- [ ] Update dependency versions once everything is settled, some are getting old and can be updated when Python 3.10 support is dropped.
- [ ] Create tests for using `uv`, `pipx`, `conda/mamba`, `poetry`, and other tools that should work with MONAI currently.

PR #9061, opened 2026-08-18
Precommit Autofixing

ericspod:
Part of #9058

### Description

This adds black and isort pre-commit hooks. This should fix formatting in the same way as `runtests.sh --autofix` but automatically, and before other actions run for too long in a PR. Note that the versions of both need to be set in the `.pre-commit-config.yaml` file separately from wherever else they're specified, so when versions are changed they need to be synced between files. The version for isort is left unchanged but the `<6` restriction should be removed shortly. The pre-commit step for `pycln` was removed in favour of checking for unused imports with Ruff by enabling F401 in `pyproject.toml`. 

### Types of changes


pyproject.toml
```diff
@@ -104,6 +104,7 @@ all = [
   "torchvision",
   "tqdm>=4.47.0",
   "transformers>=4.53.0, <5.0",
```
garciadias: This branch is based on an older `dev`; GitHub currently reports the PR as conflicting. At this hunk, current `dev` has `transformers>=5.5.0` while this branch still has `transformers>=4.53.0, <5.0`, so the three-way merge conflicts here. The branch also still carries older `setuptools` and `mlflow>=2.12.2,<3.13` values. Please rebase or merge current `dev` and preserve the recent dependency bumps (`setuptools>=78.1.1`, `mlflow>=3.15.2`, `transformers>=5.5.0`) along with this new `typeguard<3` line.

.pre-commit-config.yaml
```diff
@@ -37,8 +40,25 @@ repos:
               ^monai/_version.py
           )
 
-  - repo: https://github.com/hadialqattan/pycln
-    rev: v2.6.0
+  - repo: https://github.com/psf/black-pre-commit-mirror
+    rev: 26.5.1  # Black version, keep synced with MONAI requirements
```
garciadias, line 44: These hook revisions are pinned, but the corresponding `testing` extra constraints in `pyproject.toml` are not: `black>=26.3.1` (line 164), `isort>=5.1,!=6.0.0` (line 166), and `ruff>=0.14.11` (line 175). `./runtests.sh --autofix` uses the environment-installed versions, so a local environment can format with a different Black, isort, or Ruff than pre-commit/CI, producing churn or failing checks. Please pin these three to the exact pre-commit revisions (`black==26.5.1`, `isort==8.0.1`, `ruff==0.16.3`) or otherwise keep them in sync.

.pre-commit-config.yaml
```diff
@@ -37,8 +40,25 @@ repos:
               ^monai/_version.py
           )
 
-  - repo: https://github.com/hadialqattan/pycln
-    rev: v2.6.0
+  - repo: https://github.com/psf/black-pre-commit-mirror
```
garciadias, line 43: The `pycln` hook is removed here, but `pyproject.toml` still contains a `[tool.pycln]` section (line 293) and no remaining code references pycln. Please remove the dead `[tool.pycln]` configuration as part of this change.

garciadias (commented):
The PR description states a Pyrefly check step was added, but the final `.pre-commit-config.yaml` contains only Black and isort additions, no Pyrefly hook (it was removed in commit `a1fc8eab`). Please update the description to note Pyrefly was intentionally dropped.

garciadias (changes requested):
This branch conflicts with the current `dev` and cannot be merged as-is. Please rebase or merge `dev`, preserving the recent dependency bumps (`setuptools>=78.1.1`, `mlflow>=3.15.2`, `transformers>=5.5.0`). Also left comments on the formatter version sync between pre-commit and the `testing` extra, and on the stale `[tool.pycln]` config and PR description.

.pre-commit-config.yaml
ericspod: I don't think we should pin these versions exactly in our requirements since we don't need specifically these versions, anything current will do as far as we know  so we don't want to force users to need these. There's no way of stating "latest" as a revision in these hooks. I'll update the versions in our requirements but leave them open ended.

garciadias (approved):

PR #9064, opened 2026-08-20
perf(visualize): batch SmoothGrad noise samples with chunked forward

aymuos15:
### Description

`SmoothGrad` ran `n_samples` forward/backward passes sequentially at `monai/visualize/gradient_based.py:132` and `VanillaGrad.get_grad` rejected batched input at `monai/visualize/gradient_based.py:92`.

This change adds `SmoothGrad(sample_batch_size=1)`. Default `1` keeps the current behavior. Larger values stack noisy copies on the batch dim and run one `get_grad` per chunk.

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality)
- [ ] Breaking change (fix or new feature that would cause existing functionality to change)
- [ ] New tests added to cover the changes
- [x] Integration tests passed locally by running `python -m pytest tests/integration/test_vis_gradbased.py` (20 passed)
- [x] Quick tests passed locally by running existing visualize tests
- [ ] In-line docstrings updated
- [ ] Documentation updated, tested `make html` command in the `docs/` folder

### DIFF
--- a/monai/visualize/gradient_based.py
+++ b/monai/visualize/gradient_based.py
@@ -140,15 +140,18 @@ class SmoothGrad(VanillaGrad):
     ) -> torch.Tensor | int | None:
         if index is not None:
             return index
-        # resolve argmax once on clean input
-        was_training = self._model.model.training
-        self._model.model.eval()
-        with torch.no_grad():
-            logits = self._model.model(x, **kwargs)
-            if isinstance(logits, (list, tuple)):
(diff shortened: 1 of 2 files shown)


aymuos15:
@ericspod 

image

Sorry to ping you but, by mistake a signed my preivous commit with something wrong. However the dco still passed. Any idea why this would happen?

PR #9074, opened 2026-08-25
Adding MedNeXt to documentation

id-b3:
MedNeXt implementation is missing from the documentation.
Small fix to add it to the docs so it's clear that it exists in the MONAI framework when using the API Documentation.

### Description

Small fix to the documentation to add the missing MedNeXt model architecture and the MedNeXt block.

### Types of changes

### DIFF
--- a/docs/source/networks.rst
+++ b/docs/source/networks.rst
@@ -295,6 +295,11 @@ N-Dim Fourier Transform
 .. autofunction:: monai.networks.blocks.fft_utils_t.fftshift
 .. autofunction:: monai.networks.blocks.fft_utils_t.ifftshift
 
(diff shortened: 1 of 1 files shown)


docs/source/networks.rst
```diff
@@ -795,6 +800,11 @@ Nets
 .. autoclass:: VoxelMorph
    :members:
 
+`MedNeXt`
+~~~~~~~~~~~~
```
ericspod:
```suggestion
`MedNeXt`
~~~~~~~~~
```

ericspod (approved):
Hi @id-b3 thanks for this docs fix, this looks good to me as is.

PR #9075, opened 2026-08-25
Weekly Preview Fix

ericspod:
### Description

The `weekly-preview.yml` action is not working due to some strange interaction with the updated build process. The solution is to install MONAI first before making the wheel.

### Types of changes

### DIFF
--- a/.github/workflows/weekly-preview.yml
+++ b/.github/workflows/weekly-preview.yml
@@ -82,7 +82,7 @@ jobs:
         ls -lh dist
     - name: Test Installation
       run: |
-        pip uninstall monai -y
+        pip uninstall monai-weekly -y
         pip install dist/*.whl
         pip list
     - name: Publish to PyPI
--- a/.github/workflows/weekly-preview.yml
+++ b/.github/workflows/weekly-preview.yml
@@ -11,7 +11,7 @@ on:
       - dev
 
 env:
-  PYTHON_VER: '3.11'
+  PYTHON_VER: '3.10'
   PYTORCH_VER: '2.8.0'
   PIP_EXTRA_INDEX_URL: "https://download.pytorch.org/whl/cpu"  # forces CPU PyTorch installation, should be faster
 
--- a/.github/workflows/weekly-preview.yml
+++ b/.github/workflows/weekly-preview.yml
@@ -6,6 +6,9 @@ permissions:
 on:
   schedule:
   - cron: "0 2 * * 0"  # 02:00 of every Sunday
+  pull_request:  # temporarily enable for PR
+    branches:
+      - dev
 
 jobs:
   static-checks:
@@ -55,7 +58,7 @@ jobs:
         python-version: '3.10'
     - name: Install setuptools
       run: |
-        python -m pip install --user --upgrade setuptools wheel packaging
+        python -m pip install --user --upgrade pip setuptools wheel packaging
     - name: Build distribution
       run: |
         export HEAD_COMMIT_ID=$(git rev-parse HEAD)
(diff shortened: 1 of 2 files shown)


.github/workflows/weekly-preview.yml
```diff
@@ -72,9 +77,14 @@ jobs:
         git tag "1.7.dev${YEAR_WEEK}"
         git log -1
         git tag --list
-        python setup.py sdist bdist_wheel
-
+        python -m build --config-setting="--extra-index-url=https://download.pytorch.org/whl/cpu"
```
garciadias: `--config-setting` is passed to the setuptools backend, not to pip. `setuptools.build_meta` only honors `--build-option` and `--global-option`, so the build-isolation step still installs `torch>=2.8.0` from the default index. The PR workflow run confirms this: the isolated build resolved `torch 2.13.0+cu130`, and the later wheel install pulled the CUDA toolkit packages. If the goal is CPU wheels, set the pip environment variable instead:

```yaml
        PIP_EXTRA_INDEX_URL=https://download.pytorch.org/whl/cpu python -m build
```

Apply the same environment to the Test Installation step as well.

.github/workflows/weekly-preview.yml
```diff
@@ -6,9 +6,13 @@ permissions:
 on:
   schedule:
   - cron: "0 2 * * 0"  # 02:00 of every Sunday
+  pull_request:
```
garciadias, line 9: The new `pull_request` trigger means this job runs for every PR targeting `dev`, but the checkout is pinned to `ref: dev`. That exercises the base branch rather than the PR head, so the job can pass without validating the PR's changes and adds an expensive packaging/install run to unrelated PRs. If PR validation is intended, check out the head SHA on pull requests:

```yaml
        ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || 'dev' }}
```

Otherwise consider using `workflow_dispatch` for manual validation instead of `pull_request`.

.github/workflows/weekly-preview.yml
```diff
@@ -72,9 +77,14 @@ jobs:
         git tag "1.7.dev${YEAR_WEEK}"
         git log -1
         git tag --list
-        python setup.py sdist bdist_wheel
-
+        python -m build --config-setting="--extra-index-url=https://download.pytorch.org/whl/cpu"
+        ls -lh dist
+    - name: Test Installation
+      run: |
+        pip install dist/*.whl
```
garciadias, line 90: `pip list` confirms installation metadata but does not verify the package imports. Add a smoke test after installation, ideally from a directory outside the checkout so the source tree cannot mask the installed wheel:

```yaml
        cd "$(mktemp -d)"
        python -c 'import monai; print(monai.__version__)'
```

.github/workflows/weekly-preview.yml
```diff
@@ -53,9 +57,10 @@ jobs:
       uses: actions/setup-python@v6
       with:
         python-version: '3.10'
+        cache: 'pip'
     - name: Install setuptools
```
garciadias: The step named "Install setuptools" runs `python -m pip install -U pip build`. Consider renaming it to "Install build tooling".

garciadias (changes requested):
Left inline comments on two issues that should be addressed before merge: (1) the `--config-setting` flag is passed to the setuptools backend rather than pip, so the CPU-wheel intent is not realized — set `PIP_EXTRA_INDEX_URL` instead; (2) the new `pull_request` trigger checks out `ref: dev`, so it exercises the base branch rather than the PR head. Also suggested a wheel import smoke test and a step-name cleanup.

ericspod:
Thanks @garciadias your comments should be addressed now.

garciadias (approved):

Issue #9069, opened 2026-08-22
Unit-test results depend on the Python version: unbounded upper pins let 3.12 resolve dependencies CI never sees


hjmjohnson:
### Describe the bug

Whether the unit-test suite passes depends on which Python version the environment was built with, because `[project.optional-dependencies].all` has no upper bounds. Several dependencies have moved to `requires-python >= 3.11` or `>= 3.12`, so a Python 3.10 environment simply cannot reach the releases that break MONAI, while a 3.12 environment installs them by default.

CI pins `PYTHON_VER1: '3.10'` and uses it for `static-checks` and `full-dep`, so **CI never resolves the versions that fail.** A contributor following `CONTRIBUTING.md` on a current interpreter does.

### To Reproduce

```bash
uv venv --python 3.12 /tmp/env312 && VIRTUAL_ENV=/tmp/env312 uv pip install -e ".[all,testing]"
uv venv --python 3.10 /tmp/env310 && VIRTUAL_ENV=/tmp/env310 uv pip install -e ".[all,testing]"
./runtests.sh -u --net --coverage
```

What each interpreter resolves, from the same unbounded specifiers:

| dependency | Python 3.12 | Python 3.10 | why 3.10 cannot reach it |
|---|---|---|---|
| `zarr` | 3.3.0 | 2.18.3 | `requires-python >= 3.12` |
| `scipy` | 1.18.1 | 1.15.3 | `requires-python >= 3.12` |
| `matplotlib` | 3.11.1 | 3.10.9 | `requires-python >= 3.11` |

Full suite, unmodified `dev` (`c1240a2d4`), Python 3.12: **18196 tests, 15 failures, 26 errors**. Rebuilding the same tree on Python 3.10 clears five of the seven affected modules.

<details>
<summary>Which modules fail on 3.12 and why</summary>

| module | cause | 3.10 |
|---|---|---|
| `tests.inferers.test_zarr_avg_merger` (20 errors) | zarr 3.3.0: `ValueError: True is not a valid chunk input. Use chunks=None or chunks="auto" ...` from `zarr/core/chunk_grids.py:781` | passes |
| `tests.metrics.test_compute_fid_metric` | scipy 1.18 removed `disp` from `scipy.linalg.sqrtm`; `monai/metrics/fid.py:85` calls `sqrtm(..., disp=False)` and unpacks two return values | passes |
[...]

PR #9076, opened 2026-08-28
Fix FID metric for scipy >= 1.18: sqrtm no longer accepts disp

dhillrigo:
Fixes part of #9069 (the `scipy` row of the failure table).

### Description

`scipy` 1.18 removed the `disp` parameter from `scipy.linalg.sqrtm`, and with it the 2-tuple `(matrix, errest)` return that `disp=False` produced. `monai/metrics/fid.py:85` still called it as:

```python
scipy_res, _ = scipy.linalg.sqrtm(..., disp=False)
```

so on `scipy >= 1.18` every FID computation fails:

```
TypeError: sqrtm() got an unexpected keyword argument 'disp'
```

This is not reachable in CI today because `full-dep` pins `PYTHON_VER1: '3.10'`, and `scipy >= 1.18` requires Python >= 3.12 — so CI never resolves the breaking version, while a contributor on a current interpreter hits it immediately. That resolution gap is the subject of #9069; this PR fixes just the `scipy` breakage itself.

The fix drops `disp` and uses the return value directly. That is correct across MONAI's entire supported range (`scipy>=1.12.0`): older versions return the matrix alone when `disp` is left at its default.

| scipy | signature | `sqrtm(A)` | `sqrtm(A, disp=False)` |
|---|---|---|---|
| 1.12.0 | `(A, disp=True, blocksize=64)` | ndarray | tuple of 2 |
| 1.18.1 | `(A)` | ndarray | `TypeError` |

### One behavioural note

Omitting `disp` means it reverts to its `True` default on `scipy < 1.18`, so a matrix with no computable square root now prints scipy's `"Failed to find a square root."` to stdout rather than being silent. On `scipy >= 1.18` the same case emits a `LinAlgWarning`.

Downstream behaviour is unchanged either way: both versions return non-finite values, and `compute_frechet_distance` already branches on `torch.isfinite(covmean).all()` and prints its own message for the singular case.

I judged that preferable to version-gating the call site, but happy to switch to an explicit `scipy` version check or to suppress the message if you'd rather keep that path quiet.

### Verification

Run on unmodified `dev` and with the fix, in two environments:

| | scipy 1.18.1 / py3.12 / torch 2.13.0 | scipy 1.12.0 / py3.11 / torch 2.13.0 |
|---|---|---|
| `dev`, unpatched | `test_results` **fails** (`TypeError`) | passes |
| with this PR | 3 passed | 3 passed |

- `tests/metrics/` in full on scipy 1.18.1: **383 passed, 41 skipped**.
- The added `test_sqrtm_returns_tensor` fails on unpatched `dev` under scipy 1.18.1 and passes with the fix, so it genuinely guards the regression. Under scipy 1.12 it passes either way — the bug is version-gated, and the test asserts the contract (`_sqrtm` returns a tensor, not a tuple) rather than the version.
- `black` and `isort` clean; `flake8 --max-line-length=120` clean on both changed files. The one remaining `E501` in `fid.py` is pre-existing on `dev` (line 66, a docstring) and left untouched.

### Types of changes

<sub>On the two unchecked test boxes: I ran `pytest tests/metrics/` in full on both scipy versions rather than the full `runtests.sh` suites, since the change is confined to one function. Happy to run either in full if you want it before merge.</sub>

### DIFF
--- a/monai/metrics/fid.py
+++ b/monai/metrics/fid.py
@@ -82,7 +82,7 @@ def _cov(input_data: torch.Tensor, rowvar: bool = True) -> torch.Tensor:
 
 def _sqrtm(input_data: torch.Tensor) -> torch.Tensor:
     """Compute the square root of a matrix."""
-    scipy_res, _ = scipy.linalg.sqrtm(input_data.detach().cpu().numpy().astype(np.float64), disp=False)
+    scipy_res = scipy.linalg.sqrtm(input_data.detach().cpu().numpy().astype(np.float64))
     return torch.from_numpy(scipy_res)
 
 
--- a/tests/metrics/test_compute_fid_metric.py
+++ b/tests/metrics/test_compute_fid_metric.py
@@ -17,6 +17,7 @@ import numpy as np
 import torch
 
 from monai.metrics import FIDMetric
+from monai.metrics.fid import _sqrtm
 from monai.utils import optional_import
 
 _, has_scipy = optional_import("scipy")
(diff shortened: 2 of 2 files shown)


markov12:
Independent verification on a different platform: macOS arm64, Python 3.12.14, torch 2.14.0, scipy 1.18.1.

- Unpatched `dev`: `test_compute_fid_metric.py` has 1 failed, 1 passed (`TypeError: sqrtm() got an unexpected keyword argument 'disp'`), reproducing the issue.
- With this PR: 3 passed.
- Full `tests/metrics` with this PR: 409 passed, 0 failed (41 skipped: MetricsReloaded / scikit-learn not installed).

LGTM from a testing perspective.

(Verification done with the help of an AI assistant; I reviewed the results.)

PR #9078, opened 2026-08-28
Fix GHSA-x6pr-233j-x5cw: warn before executing an FL-provisioned bundle config

garciadias:
## Summary

Fixes GHSA-x6pr-233j-x5cw: https://github.com/Project-MONAI/MONAI/security/advisories/GHSA-x6pr-233j-x5cw
Also closes GHSA-wvpx-5qmp-46g3: https://github.com/Project-MONAI/MONAI/security/advisories/GHSA-wvpx-5qmp-46g3

`MonaiAlgo`/`MonaiAlgoStats` run a bundle whose entire app directory is provisioned by the FL system. `initialize(extra)` resolves `bundle_root = os.path.join(extra[APP_ROOT], self.bundle_root)` — `APP_ROOT` is supplied by the aggregation server — then builds a `ConfigWorkflow` over `<app_root>/configs/train.json` and runs its `initialize` expressions. Because FL tasks are dispatched per round and executed with no human in the loop, a malicious or compromised server gets silent code execution on every participating client.

The `UserWarning` added in #9057 for GHSA-873f-pvrv-4x83 lives in `create_workflow()`. This path never calls it — `MonaiAlgo` constructs `ConfigWorkflow` directly — so nothing warned here at all.

### Design

Executing the config stays unblocked, for the same reason the `trust_remote_code` flag was dropped from #9057: MONAI has no mechanism to establish whether a bundle is trustworthy, so a flag mostly teaches operators to set it once and forget about it. Both `initialize()` methods now warn, naming the trust boundary (`extra[APP_ROOT]`) and the absence of per-round human interaction.

The one behaviour change is narrowly scoped, and it targets the sink with no functional role in FL. `ConfigWorkflow` defaults `logging_file` to the bundle's own `configs/logging.conf` and passes it to `logging.config.fileConfig`, which `eval()`s the INI's `class=`/`args=` fields. That is code execution at construction time, before any config is parsed, and it hides in a plain INI rather than the MONAI `$`-DSL — easy to miss when reviewing a bundle. The FL client now treats `extra[ExtraItems.LOGGING_FILE]` as `False` both when the key is absent and when it is explicitly `None`, so a server-written `logging.conf` is never applied. `None` needs the same treatment as absent because it was the pre-PR default and `ConfigWorkflow` reads it as "fall back to the bundle's own `configs/logging.conf`" — exactly the file this change exists to keep away from `fileConfig`. An FL system that wants bundle logging passes an explicit path, through the key that already exists for it.

The `fileConfig` warning sits inside the branch that actually calls it, not at the top of `__init__`. That keeps it truthful (nothing runs when the file is absent or logging is disabled, both common) and avoids double-warning callers who already got the `create_workflow()` warning, which is about `_target_`/`$` rather than the INI.

### Changes

- `monai/fl/client/monai_algo.py`: warning in both `initialize()` methods; `ExtraItems.LOGGING_FILE` treated as `False` when absent or explicitly `None`; security notes on both class docstrings; both `initialize()` docstrings rewritten for the new default (this also fixes a `diable` typo).
- `monai/bundle/workflows.py`: `_warn_logging_file_execution()` called immediately before each of the two `fileConfig` invocations; `logging_file` docstring entries updated on `BundleWorkflow`, `PythonicWorkflow` and `ConfigWorkflow`.
- `tests/fl/monai_algo/test_fl_monai_algo.py`: `TestFLMonaiAlgoWarnsOnProvisionedConfig` — stages an app whose `train.json` and `logging.conf` each drop a distinct marker, for both `MonaiAlgo` and `MonaiAlgoStats`. Asserts the config still executes with the advisory warning; that the server's `logging.conf` no longer does, whether the key is absent or explicitly `None`; that an explicit path opts back in; and that no `fileConfig` warning fires when nothing is executed.
- `tests/bundle/test_bundle_workflow.py`: `TestConfigWorkflowWarnsOnLoggingConf` — a bundle's default `configs/logging.conf` warns and still applies; `logging_file=False` neither warns nor applies it.

Both new test classes snapshot and restore the root logger, closing any handler `fileConfig` installs. The suite runs in one process, so without that they would leak a root handler and formatter into every test that follows.

## Test plan

- [x] `python -m unittest tests.fl.monai_algo.test_fl_monai_algo` — 17 passed
- [x] `python -m unittest tests.fl.test_fl_monai_algo_stats` — 3 passed
- [x] `python -m unittest tests.bundle.test_bundle_workflow.TestConfigWorkflowWarnsOnLoggingConf` — 2 passed
- [x] `python -m unittest tests.bundle.test_bundle_download.TestLoadWarnsOnConfigExecution` — 6 passed, no double-warn regression on the #9057 fix
- [x] Each new assertion checked against the unpatched code first — the advisory's payload writes its marker via the bundle config and via `logging.conf` before the change, and only via the bundle config after it
- [x] Root logger verified identical before and after both new test classes run
- [x] `black`, `isort`, `ruff` clean on the changed files

### Types of changes

The breaking-change box is for the `LOGGING_FILE` default only: an FL deployment relying on the bundle shipping its own `logging.conf` now has to pass the path explicitly. Everything else is additive.

### DIFF
--- a/tests/bundle/test_bundle_workflow.py
+++ b/tests/bundle/test_bundle_workflow.py
@@ -11,11 +11,13 @@
 
 from __future__ import annotations
 
+import json
 import os
 import shutil
 import sys
 import tempfile
 import unittest
+import warnings
 from copy import deepcopy
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


tests/bundle/test_bundle_workflow.py
```diff
@@ -268,5 +271,84 @@ def test_create_pythonic_workflow(self):
...
+        with tempfile.TemporaryDirectory() as tempdir:
+            configs = os.path.join(tempdir, "configs")
+            os.makedirs(configs)
+            marker = os.path.join(tempdir, "PWNED")
+            with open(os.path.join(configs, "train.json"), "w") as f:
+                json.dump({"initialize": []}, f)
+            # `fileConfig` eval()s the `class=` field, so the tuple subscript runs the payload and
+            # still yields a usable handler class.
+            with open(os.path.join(configs, "logging.conf"), "w") as f:
+                f.write(
```
ericspod, line 316: Another way of printing multi-line files is to put the text into a """ """ multi-line string and use [dedent to get the indentation right](https://github.com/Project-MONAI/MONAI/blob/fd8a819e7e550c1458cd66dda785d21be01b2521/monai/bundle/scripts.py#L1841).

ericspod (approved):
Looks good to me. Maybe we should think about a safe implementation of fileConfig? At least check a file for safe values along the lines of safe_eval.

PR #9079, opened 2026-08-28
Update CodeQL Action

ericspod:
Part of #9058.

### Description
This updates the CodeQL action to get this running again.

### Types of changes

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -9,67 +9,95 @@
 # the `language` matrix defined below to confirm you have the correct set of
 # supported CodeQL languages.
 #
-name: "CodeQL"
+name: "CodeQL Advanced"
 
 on:
   push:
-    branches: [ dev, main ]
+    branches: [ "dev" ]
   pull_request:
-    # The branches below must be a subset of the branches above
-    branches: [ dev ]
+    branches: [ "dev" ]
   schedule:
-    - cron: '18 1 * * 0'
+    - cron: '0 2 * * 1'  # 2AM Monday
 
 jobs:
   analyze:
-    name: Analyze
-    runs-on: ubuntu-latest
+    name: Analyze (${{ matrix.language }})
+    # Runner size impacts CodeQL analysis time. To learn more, please see:
+    #   - https://gh.io/recommended-hardware-resources-for-running-codeql
+    #   - https://gh.io/supported-runners-and-hardware-resources
+    #   - https://gh.io/using-larger-runners (GitHub.com only)
+    # Consider using larger runners or machines with greater resources for possible analysis time improvements.
+    runs-on: 'ubuntu-latest'
     permissions:
+      # required for all workflows
+      security-events: write
+
+      # required to fetch internal or private CodeQL packs
+      packages: read
+
+      # only required for workflows in private repositories
       actions: read
       contents: read
-      security-events: write
 
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'cpp', 'python' ]
-        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
-        # Learn more about CodeQL language support at https://git.io/codeql-language-support
-
+        include:
+        - language: actions
+          build-mode: none
+        - language: c-cpp
+          build-mode: none
+        - language: python
+          build-mode: none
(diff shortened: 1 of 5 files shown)


garciadias (dismissed):

ericspod:
I've had to make a small fix to dints.py with the arrival of PyTorch 2.14 in PyPI. This creates a deprecation warning we shouldn't ignore now, further removal of deprecated Torchscript components is being worked on in #8777.

ericspod:
Further minor fixes were done.

monai/utils/profiling.py
```diff
@@ -76,7 +76,7 @@ def torch_profiler_time_cpu_gpu(func):
 
     @wraps(func)
     def wrapper(*args, **kwargs):
-        with torch.autograd.profiler.profile(use_cuda=True) as prof:
+        with torch.autograd.profiler.profile() as prof:
```
garciadias, line 79: Removing use_cuda=True disables CUDA event collection, so gpu_time (the sum of evt.self_cuda_time_total on the following lines) will silently evaluate to 0. This decorator's documented purpose is to report both CPU and GPU time, so this is a regression for downstream users running on a GPU. The replacement that preserves the behavior is use_device="cuda" — the option the use_cuda deprecation warning itself points to — and it works on both torch 2.8 and 2.14. Suggest: with torch.autograd.profiler.profile(use_device="cuda") as prof: (the same applies at line 60 in torch_profiler_full).

monai/networks/nets/dints.py
```diff
@@ -36,20 +36,23 @@
 __all__ = ["DiNTS", "TopologyConstruction", "TopologyInstance", "TopologySearch"]
 
 
-@torch.jit.interface
-class CellInterface(torch.nn.Module):
-    """interface for torchscriptable Cell"""
+# TODO: added temporarily for PyTorch 2.14 warnings, remove when factoring out deprecated Torchscript components
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
```
garciadias, line 41: warnings.simplefilter("ignore") with no category suppresses every warning raised while these two classes are defined, not just the torch.jit.interface deprecation this is meant to silence. Suggest scoping the filter to the specific warning category and/or message so unrelated import-time warnings are not masked; a targeted filter also makes it clearer when this temporary workaround can be dropped.

.github/workflows/codeql-analysis.yml
```diff
@@ -9,67 +9,110 @@
...
+    - name: Set up Python ${{ env.PYTHON_VER }}
+      if: matrix.language == 'c-cpp' && matrix.build-mode == 'manual'
+      uses: actions/setup-python@v6
+      with:
+        python-version: ${{ env.PYTHON_VER }}
+        cache: 'pip'
 
-    - name: Build
+    - name: Run manual build steps
+      if: matrix.language == 'c-cpp' && matrix.build-mode == 'manual'
```
garciadias, line 103: These manual build steps can never execute: the c-cpp matrix entry sets build-mode: none (line 55), while both this step and the 'Set up Python' step above are guarded by matrix.build-mode == 'manual'. Since build-mode none is the accepted approach here, this is dead code. Suggest removing the 'Set up Python' and 'Run manual build steps' blocks (lines 95-111), together with the now-unused env vars BUILD_MONAI and PIP_EXTRA_INDEX_URL (PYTHON_VER and PYTORCH_VER are only consumed by these dead steps).

garciadias (approved):
Approved. Left three minor suggestions (profiler use_device, dead manual-build steps in the CodeQL workflow, and scoping the dints.py warning filter) — all optional, at your discretion.

monai/networks/nets/dints.py
ericspod: There's no other possible warnings that would be raised here other than jit related ones. No other code is executed to create warnings that we'd want to see, and this is going to be temporary anyhow.

.github/workflows/codeql-analysis.yml
ericspod: This was an attempt to get manual building working, I think we'll need to do this later so I want to leave this here for another round of fixing this script. The reasons why the manual build failed aren't clear but we should first focus on getting CodeQL to not complain in Github about its configuration being incorrect.

ericspod:
Thanks @garciadias I'm going to merge this now but the profile fix I shall put into the weekly build PR that we need to merge soon as well.

PR #9080, opened 2026-08-29
fix(transforms): resolve TypeError when spatial_size is None (#9068)

### LINKED ISSUE #9068: spatial_resample raises TypeError when spatial_size is None and spatial_rank is 1
### Describe the bug

`spatial_resample` passes `lambda x: x >= 0` to `fall_back_tuple`, but that helper's *default* predicate is `lambda x: x and x > 0`, whose `x and` short-circuits on `None`. The override does not, so a `None` element raises `TypeError` instead of falling back to the default — which is the documented behaviour of `fall_back_tuple`, and its own docstring says so:

```python
>>> fall_back_tuple((-1, None), (32, 32))
(32, 32)
```

`monai/transforms/spatial/functional.py:159`:

```python
spatial_size = torch.tensor(
    fall_back_tuple(ensure_tuple(spatial_size)[:spatial_rank], in_spatial_size, lambda x: x >= 0)
)
```

`spatial_size` reaches that line as `None` whenever the caller did not supply one and `spatial_rank <= 1`: the branch above it, `elif spatial_size is None and spatial_rank > 1`, is the only thing that replaces `None`, so with a rank of 1 the `None` survives into `ensure_tuple(None)` → `(None,)` → `None >= 0`.

### To Reproduce

```python
import numpy as np
from monai.data.image_writer import NibabelWriter

w = NibabelWriter()
w.set_data_array(np.random.rand(3, 5), channel_dim=None)
w.set_metadata({"affine": np.diag([1, 1, 1]), "original_affine": np.diag([1.4, 1, 1])})
```

```
File "monai/data/image_writer.py", line 604, in set_metadata
File "monai/data/image_writer.py", line 273, in resample_if_needed
File "monai/transforms/spatial/array.py", line 229, in __call__
File "monai/transforms/spatial/functional.py", line 159, in <lambda>
    fall_back_tuple(ensure_tuple(spatial_size)[:spatial_rank], in_spatial_size, lambda x: x >= 0)
TypeError: '>=' not supported between instances of 'NoneType' and 'int'
```

Instrumenting `spatial_resample` confirms it is entered with `spatial_size=None` and `img.shape=(1, 3, 5)`.

### Expected behavior

`None` means "no size given for this axis" and should fall back to the corresponding `in_spatial_size` entry, exactly as `fall_back_tuple`'s docstring describes.

### Screenshots / test impact

This is not a corner case reachable only by hand — it fails **8 tests on current `dev`**:

- `tests/data/test_nifti_rw.py` — `test_write_2d`, `test_write_3d`
- `tests/data/test_image_rw.py` — 4 failures + 4 errors across `TestRegRes` / writer round-trips

Reproduced on unmodified `dev` (`c1240a2d4`) in two independent environments:

| | |
|---|---|
| Python 3.12 / torch 2.13.0 / numpy 2.x | fails |
| Python 3.10 / torch 2.11.0 (highest version CI tests) / numpy 2.2.6 | fails identically |

So it is not a new-dependency artifact. `git log -L159,159` dates that line to #6068 (Feb 2023).

### Environment

<details>
<summary>Ensuring you use the relevant python executable, please paste the output of:</summary>

```
MONAI version: 1.6.0rc1+58.gc1240a2d4
Python: 3.10 and 3.12 (both affected)
PyTorch: 2.11.0 and 2.13.0 (both affected)
numpy: 2.2.6
nibabel: 5.4.2
```

</details>

### Additional context

The narrowest fix is to make the predicate `None`-safe at the call site, matching the helper's default:

```python
lambda x: x is not None and x >= 0
```

Worth checking the other `fall_back_tuple` callers that pass an explicit `func` for the same hazard — any predicate that does not short-circuit on `None` inherits it.

FinalSunFlower:
### Description
Fixes #9068.

This PR resolves a `TypeError` in `spatial_resample` when `spatial_size=None` (or contains `None` entries). The predicate now safely guards against `NoneType` comparisons when evaluating target spatial dimensions.

### Key Changes
- Updated `spatial_size` evaluation in `monai/transforms/spatial/functional.py` to be `None`-safe.
- Added dedicated CPU unit test cases in `tests/transforms/test_spatial_resample.py` testing explicit `spatial_size=None` and partial `None` dimensions.
- Aliased the imported lazy-resampler helper so pytest does not collect it as a standalone fixture-based test.

### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Unit tests added/updated

### DIFF
--- a/monai/transforms/spatial/functional.py
+++ b/monai/transforms/spatial/functional.py
@@ -137,6 +137,13 @@ def spatial_resample(
         dtype_pt: data `dtype` for resampling computation.
         lazy: a flag that indicates whether the operation should be performed lazily or not
         transform_info: a dictionary with the relevant information pertaining to an applied transform.
+
+    Returns:
+        torch.Tensor: The resampled output tensor, with metadata preserved when metadata tracking is enabled.
+
+    Raises:
+        ValueError: If the affine or spatial dimensions are invalid, or if the output spatial size cannot be
+            computed.
     """
     original_spatial_shape = img.peek_pending_shape() if isinstance(img, MetaTensor) else img.shape[1:]
     src_affine: torch.Tensor = img.peek_pending_affine() if isinstance(img, MetaTensor) else torch.eye(4)
--- a/monai/transforms/spatial/functional.py
+++ b/monai/transforms/spatial/functional.py
@@ -117,7 +117,9 @@ def spatial_resample(
     Args:
         img: data to be resampled, assuming `img` is channel-first.
         dst_affine: target affine matrix, if None, use the input affine matrix, effectively no resampling.
-        spatial_size: output spatial size, if the component is ``-1``, use the corresponding input spatial size.
+        spatial_size: output spatial size. Components set to ``-1`` or ``None`` use the corresponding input
+            spatial dimension. If the entire value is ``None``, the output size is computed automatically when
+            possible, otherwise the input spatial shape is used.
         mode: {``"bilinear"``, ``"nearest"``} or spline interpolation order 0-5 (integers).
             Interpolation mode to calculate output values.
             See also: https://pytorch.org/docs/stable/generated/torch.nn.functional.grid_sample.html
--- a/monai/transforms/spatial/functional.py
+++ b/monai/transforms/spatial/functional.py
@@ -156,7 +156,7 @@ def spatial_resample(
     elif spatial_size is None and spatial_rank > 1:  # auto spatial size
         spatial_size, _ = compute_shape_offset(in_spatial_size, src_affine, dst_affine)  # type: ignore
     spatial_size = torch.tensor(
-        fall_back_tuple(ensure_tuple(spatial_size)[:spatial_rank], in_spatial_size, lambda x: x >= 0)
+        fall_back_tuple(ensure_tuple(spatial_size)[:spatial_rank], in_spatial_size, lambda x: x is not None and x >= 0)
     )
     extra_info = {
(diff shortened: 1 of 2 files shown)


hjmjohnson:
A rebase now that https://github.com/Project-MONAI/MONAI/pull/9079 is merged may clear the failing test.

markov12:
Thanks for working on this! I tested the PR locally (macOS arm64, Python 3.12.14, torch 2.14.0, nibabel 5.4.2):

- The repro from #9068 no longer raises, and `test_spatial_resample.py` goes from 74 passed + 1 collection error to 76 passed. The rename of `test_resampler_lazy` fixes that collection error.
- However, `tests/data/test_nifti_rw.py` still has 2 failures with the PR: `test_write_2d` now fails with affine `1.0` vs expected `1.4` (no resampling happens), and `test_write_3d` fails as before (shape `(1, 1, 5)` vs `(1, 1, 3)`).

I traced this to the writer setting the affine before adding the channel dim, which makes `spatial_ndim` 1 after #8765. Details, bisect, and a small suggested change are in my comment on #9068. With that change applied, all 41 `test_nifti_rw` tests pass. Feel free to include it in this PR if it helps. I don't intend to open a separate one.

(Testing done with the help of an AI assistant; I reviewed the results.)

FinalSunFlower:
Thanks for the thorough investigation and local verification, @markov12!

I've integrated your suggested fix regarding the channel dimension ordering before setting the affine. Confirmed locally that all 41 tests in tests/data/test_nifti_rw.py as well as test_spatial_resample.py are now passing cleanly.

Pushed the update to this PR for review.

PR #9081, opened 2026-09-01
Fixing Git Dependency Weekly Build Issue

ericspod:
### Description

This fixes an issue with git dependencies in the requirements for MONAI. This was sorted earlier in #5887 but I missed this reference when retiring setup.cfg. This also adds a step to publish to the Test PyPI server but this is commented out for now until we have a working key.

### Types of changes


.github/workflows/weekly-preview.yml
```diff
@@ -90,8 +92,17 @@ jobs:
         pip install dist/*.whl
         pip list
         (cd "$(mktemp -d)" && python -c 'import monai; print(monai.__version__)')
+    - name: Publish to Test PyPI
+      if: github.event_name == 'pull_request' && github.repository == 'Project-MONAI/MONAI' # only test publish on PR
```
garciadias: This step fails because the PR is opened from your personal fork (`ericspod/MONAI:weekly_preview2`, i.e. `isCrossRepository: true`), so GitHub withholds `secrets.TEST_PYPI_TOKEN` from this `pull_request`-triggered run by design — forked PRs never get repo secrets, so a malicious fork can't add workflow code that exfiltrates them. That's why `TWINE_PASSWORD` resolves empty here and Test PyPI returns a 403.

No workflow change is actually needed to fix this safely and automatically: since you have write access to this repo, push `weekly_preview2` as a branch on `Project-MONAI/MONAI` itself (instead of your personal fork) and open the PR from there. A same-repo PR isn't cross-repository, so the `pull_request` trigger gets secrets automatically — the same way the `schedule`-triggered `Publish to PyPI` step already works with zero friction. No manual dispatch step, no relaxed security posture.

garciadias: Worth flagging: pushing to the base repo fixes it for you specifically because you have write access to `Project-MONAI/MONAI`. It won't help an external contributor who forks the repo without write access — they can never push a branch here directly, so this exact 403 will resurface for them. Since that'll come up again, it's worth deciding now between the two safe options for that case (ruling out `pull_request_target`, which would run untrusted fork code with secrets exposed — not worth the risk):

1. **Gate the step on same-repo PRs**: add `&& github.event.pull_request.head.repo.full_name == github.repository` to this `if:`. Fully automatic, no security trade-off — but the Test PyPI publish step silently skips for any fork PR, so it never validates the workflow for outside contributions touching this file.
2. **Move the step to `workflow_dispatch`**: drop it from the `pull_request` trigger entirely and let a maintainer manually run it (Actions tab or `gh workflow run`) against any branch, fork or not. Still safe (dispatch requires write access), and it does let someone validate a fork's changes — just not automatically on every push.

Given this step only really matters when someone is actively changing this workflow file (like you are now), and that's rare, I'd lean option 1 — the common case (fork PR that isn't touching CI) skips a step it doesn't need, and if someone specifically needs to validate a Test PyPI publish from a fork, they'd have a maintainer manually dispatch it (which is effectively option 2 used situationally rather than baked into the trigger). Let me know which way you want to go and I'm happy to open a follow-up PR.

garciadias (dismissed):

.github/workflows/weekly-preview.yml
ericspod: Thanks for looking into this. My intent was to test uploading to testpypi for every PR so that any changes contributors make elsewhere don't break the wheel, for example adding dependencies like direct git URLs which is what we have to filter out currently. If we can't test upload functionality for every PR I don't think either option is going to provide much value. We'll just end up fixing the issue once we notice the preview action failing, this would always be the case even if we did either option. I've added checking with twine but this is [known to not cover everything that would fail for PyPI](https://github.com/pypa/twine/issues/1152). Also, I had tried uploading bad packages to a local instance of [pypiserver](https://github.com/pypiserver/pypiserver) but these were accepted without complaint. 

I suggest we remove the test for now and come back to it in the future, I'll push changes for this shortly.

ericspod:
PyTorch 2.14 was just released and has caused the fails, the fix is in #9079.

KumoLiu (approved):

PR #9087, opened 2026-09-03
Reject non-finite DICOM affine metadata in PydicomReader

garciadias:
### Description

`PydicomReader._get_affine` builds the affine matrix from DICOM `PixelSpacing`, `ImagePositionPatient`, and `ImageOrientationPatient` values with no finite check. A crafted DICOM carrying `NaN`/`inf` in those DS tags produces a corrupted affine that propagates through spatial transforms and crashes MONAILabel inference or silently corrupts results.

Validate all affine inputs with `math.isfinite()` and raise `ValueError` naming the offending tag before building the matrix (GHSA-6hp3-vr39-rqw8).

### Types of changes
- [ ] Non-breaking change
- [x] Breaking change (non-finite DICOM geometry now raises instead of silently proceeding)

### DIFF
--- a/monai/data/image_reader.py
+++ b/monai/data/image_reader.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 import glob
 import gzip
 import io
+import math
 import os
 import re
 import tempfile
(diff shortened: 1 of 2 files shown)


monai/data/image_reader.py
```diff
@@ -752,13 +753,29 @@ def _get_affine(self, metadata: dict, lps_to_ras: bool = True):
                 stacklevel=2,
             )
             return affine
+
+        def _raise_if_not_finite(value: Any, tag: str) -> None:
+            if not math.isfinite(value):
+                raise ValueError(
+                    f"PydicomReader: cannot derive affine matrix because DICOM tag {tag} "
+                    f"has a non-finite value: {value}."
+                )
```
ericspod:
```suggestion
        def _raise_if_not_finite(values: Sequence[Any], tag: str) -> None:
            if not np.isfinite(tuple(values)).all():
                raise ValueError(
                    f"PydicomReader: cannot derive affine matrix because DICOM tag {tag} "
                    f"has a non-finite value: {value}."
                )
```
This should work for the values you're wanting to check, so you can do something like ` _raise_if_not_finite((rx, ry, rz, cx, cy, cz), "ImageOrientationPatient (0020,0037)")`.

monai/data/image_reader.py
ericspod, line 799:
```suggestion
        if not np.isfinie(affine).all():
            raise ValueError("PydicomReader: affine matrix not finite after composition.") 

        if lps_to_ras:
```
Something like this? Maybe the message should give more info about why it's not finite? You'd have to look into what computation above made a non-finite result from finite inputs I guess.

tests/data/test_pydicom_reader.py
```diff
@@ -39,6 +39,40 @@ def test_partial_orientation_tags_warns(self):
             affine = reader._get_affine(metadata)
         np.testing.assert_array_equal(affine, np.eye(4))
 
+    def test_non_finite_pixel_spacing_raises(self):
```
ericspod, line 43: All this looks good, we would need a test for a non-finite matrix getting created from finite values that cause an overflow as coderabbit flagged above.

ericspod (approved):

